Issue #6694, opened 2024-02-14
[FIRRTL] Failure with Nightly Chisel re: layers


seldridge:
The following circuit is currently failing to compile:

```
FIRRTL version 4.0.0
circuit Foo :
  layer A, bind :
    layer D, bind :
  module Foo :
    input port : UInt<1>

    layerblock A :
      wire a : UInt<1>
      connect a, port

      layerblock D :
        node _d_T = xor(port, a)
        wire d : UInt<1>
        connect d, _d_T
```

```
Website.fir:13:21: error: 'firrtl.strictconnect' op using value defined outside the region
        node _d_T = xor(port, a)
                    ^
Website.fir:13:21: note: see current operation: "firrtl.strictconnect"(%0#0, <<UNKNOWN SSA VALUE>>) : (!firrtl.uint<1>, !firrtl.uint<1>) -> ()
Website.fir:5:3: note: required by region isolation constraints
  module Foo :
```

rwy7:
Thanks for reporting!

rwy7:
Likely related failure:
```
// RUN: circt-opt -firrtl-lower-layers -split-input-file %s | FileCheck %s

firrtl.circuit "CaptureHardware" {
  firrtl.layer @A bind {
    firrtl.layer @B bind {}
  }
  firrtl.module @CaptureHardware() {
    %c0_ui1 = firrtl.constant 0 : !firrtl.uint<1>
    firrtl.layerblock @A {
      %0 = firrtl.add %c0_ui1, %c0_ui1 : (!firrtl.uint<1>, !firrtl.uint<1>) -> !firrtl.uint<2>
    }
  }
}

```

generates:

```
// -----// IR Dump After LowerLayers Failed (firrtl-lower-layers) //----- //
"firrtl.circuit"() ({
  "sv.verbatim"() {format_string = "`include \22groups_CaptureHardware_A.sv\22\0A`ifndef groups_CaptureHardware_A_B\0A`define groups_CaptureHardware_A_B", output_file = #hw.output_file<"groups_CaptureHardware_A_B.sv", excludeFromFileList>, symbols = []} : () -> ()
  "sv.verbatim"() {format_string = "`ifndef groups_CaptureHardware_A\0A`define groups_CaptureHardware_A", output_file = #hw.output_file<"groups_CaptureHardware_A.sv", excludeFromFileList>, symbols = []} : () -> ()
  "firrtl.module"() ({
[...]

PR #6699, opened 2024-02-14
[FIRRTL][Lower-Layers] do not capture uses multiple times

rwy7:
This PR fixes two issues.

First, when a captured value has been replaced by a block-argument on the layerblock, if we encounter a use of the block-argument, we would attempt to "recapture" the argument, because our check for "is this value defined inside the layerblock" was broken for block arguments. This PR fixes that check. Now a value can be captured more than once.

Second, when capturing a value as a "source/input-port", we should only replaces uses where the value is used as a source. Similarly, when a value is captured as a "sink/output-port", we should only replace uses where the value is used as the destination of a connect-like operation.

This means that if a value is used as both a source and a sink, we will create both an input and an output port. In this case, we may eventually want to create only a single output output port.

Fixes: #6694

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -24,6 +24,17 @@ using namespace circt;
 using namespace firrtl;
 
 //===----------------------------------------------------------------------===//
+// Helpers
+//===----------------------------------------------------------------------===//
+
+static bool isAncestor(Operation *op, Value value) {
+  if (auto result = dyn_cast<OpResult>(value))
+    return op->isAncestor(result.getOwner());
+  auto argument = cast<BlockArgument>(value);
+  return op->isAncestor(argument.getOwner()->getParentOp());
+}
+
+//===----------------------------------------------------------------------===//
 // Type Conversion
 //===----------------------------------------------------------------------===//
 
@@ -231,7 +242,6 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp) {
 
     // Create an input port for an operand that is captured from outside.
     auto createInputPort = [&](Value operand, Location loc) {
-      auto portNum = ports.size();
       auto operandName = getFieldName(FieldRef(operand, 0), true);
 
       // If the value is a ref, we must resolve the ref inside the parent,
@@ -247,15 +257,21 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp) {
       // Update the layer block's body with arguments as we will swap this body
       // into the module when we create it.  If this is a ref type, then add a
       // refsend to convert from the non-ref type input port.
-      body->addArgument(type, loc);
-      Value replacement = body->getArgument(portNum);
+      Value replacement = body->addArgument(type, loc);
       if (isa<RefType>(operand.getType())) {
         OpBuilder::InsertionGuard guard(builder);
         builder.setInsertionPointToStart(body);
         replacement = builder.create<RefSendOp>(loc, replacement);
       }
-      operand.replaceUsesWithIf(replacement, [&](OpOperand &operand) {
-        return layerBlock->isAncestor(operand.getOwner());
+      operand.replaceUsesWithIf(replacement, [&](OpOperand &use) {
+        auto *user = use.getOwner();
+        if (!layerBlock->isAncestor(user))
+          return false;
+        if (auto connectLike = dyn_cast<FConnectLike>(user)) {
+          if (use.getOperandNumber() == 0)
+            return false;
+        }
+        return true;
       });
 
       connectValues.push_back({operand, ConnectKind::NonRef});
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -23,6 +23,18 @@
...
+//===----------------------------------------------------------------------===//
+// Helpers
+//===----------------------------------------------------------------------===//
+
+static bool isAncestor(Operation *op, Value value) {
+  if (auto result = dyn_cast<OpResult>(value))
+    return op->isAncestor(result.getOwner());
+  if (auto argument = dyn_cast<BlockArgument>(value))
+    return op->isAncestor(argument.getOwner()->getParentOp());
+  return false;
```
seldridge, line 35: Could this be a blind cast (which would pull in an assert for debug-mode checking)? (Can this be anything else?)

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -294,12 +311,18 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp) {
...
-                               });
+        dest.replaceUsesWithIf(replacement, [&](OpOperand &use) {
+          auto user = use.getOwner();
+          if (!layerBlock->isAncestor(user))
+            return false;
+          if (auto connectLike = dyn_cast<FConnectLike>(user)) {
+            if (use.getOperandNumber() == 0)
+              return true;
+          }
+          return false;
```
seldridge, line 324: Consider factoring this to a common static function or lambda so that the PR can capture what it needs.

seldridge: This logic and it's prior use, makes sense to me!

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -339,39 +362,35 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp) {
...
           continue;
         }
         // Create an output port.
         if (!dstInLayerBlock) {
-          createOutputPort(connect.getDest(), connect.getSrc());
-          if (!connect.getDest().getType().isa<RefType>())
+          createOutputPort(dst, src);
+          if (!dst.getType().isa<RefType>())
             connect.erase();
           continue;
```
seldridge, line 395: This region is some great cleanup.

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -471,3 +471,82 @@ firrtl.circuit "ModuleNameConflict" {
...
+// https://github.com/llvm/circt/issues/6694
+
+firrtl.circuit "CaptureHardwareMultipleTimes" {
+  firrtl.layer @A bind {
+    firrtl.layer @B bind {}
+  }
+
+  // CHECK: firrtl.module private @[[A:.+]](in %[[p:.+]]: !firrtl.uint<1>) {
+  // CHECK:   %0 = firrtl.add %[[p]], %[[p]] : (!firrtl.uint<1>, !firrtl.uint<1>) -> !firrtl.uint<2>
+  // CHECK: }
```
seldridge, line 486: Awesome!

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -471,3 +471,82 @@ firrtl.circuit "ModuleNameConflict" {
...
+      // capture first as a source.
+      %2 = firrtl.ref.resolve %w : !firrtl.probe<uint<1>, @A>
+      // capture again, as a sink.
+      %c0_ui1 = firrtl.constant 0 : !firrtl.uint<1>
+      %0 = firrtl.ref.send %c0_ui1 : !firrtl.uint<1>
+      %1 = firrtl.ref.cast %0 : (!firrtl.probe<uint<1>>) -> !firrtl.probe<uint<1>, @A>
+      firrtl.ref.define %w, %1 : !firrtl.probe<uint<1>, @A>
+    }
+  }
+}
```
seldridge, line 552: Nits in new tests: 

The type information is really unimportant for these tests. Consider dropping it where unnecessary as it doesn't aid test readability.

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -471,3 +471,82 @@ firrtl.circuit "ModuleNameConflict" {
...
+  // CHECK:   firrtl.strictconnect %[[in]], %0 : !firrtl.uint<1>
+  // CHECK:   firrtl.ref.define %w, %[[out]] : !firrtl.probe<uint<1>>
+  // CHECK: }
+  firrtl.module @CaptureAsDstThenSrc() {
+    %w = firrtl.wire : !firrtl.probe<uint<1>, @A>
+    firrtl.layerblock @A {
+      // capture first as a sink.
+      %c0_ui1 = firrtl.constant 0 : !firrtl.uint<1>
+      %0 = firrtl.ref.send %c0_ui1 : !firrtl.uint<1>
+      %1 = firrtl.ref.cast %0 : (!firrtl.probe<uint<1>>) -> !firrtl.probe<uint<1>, @A>
```
seldridge, line 519: This could be simplified to a wire of the appropriate ref type.

Ditto in the dual of this test below.

seldridge (approved):
LGTM

This is also great in that it doesn't create unnecessary ports.

I tested locally and this fixes the nightly Chisel failure here: https://github.com/chipsalliance/chisel/actions/runs/7900229902/job/21561340132

PR #6701, opened 2024-02-14
[FIRRTL] Add integer addition parser support.

mikeurbach:
This adds parser support for IntegerAddOp by defining the primitive expression keyword. This is only supported in FIRRTL version 4.0.0 and up.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
@@ -1186,6 +1186,18 @@ class IntegerBinaryPrimOp<string mnemonic, list<Trait> traits = []> :
                 [Pure, SameOperandsAndResultType] # traits> {
   // We use the standard inferReturnTypes from SameOperandsAndResultType.
   let inferReturnTypesDecl = "";
+
+  // We use a static inferType that always returns FIntegerType.
+  let inferType = "getFIntegerType";
+
+  // Define the getFIntegerType inline in ODS.
+  let firrtlExtraClassDeclaration = [{
+    static FIntegerType getFIntegerType(FIRRTLType lhs,
+                                        FIRRTLType rhs,
+                                        std::optional<Location> loc) {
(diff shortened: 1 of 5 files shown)


mikeurbach:
Ah I just realized the changes in #6691 need to be taken into account, will fix this up.

mikeurbach:
The parser expects to be able to call inferReturnType (singular): https://github.com/llvm/circt/blob/6f1fea12cf0329d7a7c37450bfc6c4eecdda073c/lib/Dialect/FIRRTL/Import/FIRParser.cpp#L2281

Since we customized IntegerBinaryPrimOp to provide inferReturnTypes (plural) via SameOperandsAndResultType, nothing was defining inferReturnType (singular) for these ops. In the last commit I pushed, I changed IntegerBinaryPrimOp to a) define inferReturnType to call a custom `inferType` callback, and b) defined that callback in ODS to just return an FIntegerType. What do we think about this? I don't normally put code into TableGen files like this, but this seems like the simplest way to appease the existing ODS setup without adding boilerplate elsewhere.

uenoku (approved):
>  In the last commit I pushed, I changed IntegerBinaryPrimOp to a) define inferReturnType to call a custom inferType callback, and b) defined that callback in ODS to just return an FIntegerType

This approach looks good to me. 

Relevantly I'm wondering if we can remove the type inference manually implemented in the macro (not by this PR). I'm pretty sure that an equivalent builder is generated by ODS `void build(::mlir::OpBuilder &odsBuilder, ::mlir::OperationState &odsState, ::mlir::ValueRange operands, ::llvm::ArrayRef<::mlir::NamedAttribute> attributes = {});` and we can just use plain verifier today. Maybe `inferReturnType` stuff was added before MLIR provided `InferTypeOpInterface`?

mikeurbach:
Yes, I think you are right--this ODS business likely predates InferTypeOpInterface existing. I will go ahead and merge the current PR as-is, and open an issue about ripping this out to use the standard methods if we don't already have an issue tracking that.

PR #6702, opened 2024-02-14
[OM] Add rationale for expressions.

mikeurbach:
This adds some basic rationale for why it is valuable to be able to represent computation in the OM dialect, with a couple examples including arithmetic and container constructions.


uenoku (approved):
Looks great!

PR #6703, opened 2024-02-15
[OM] Add OpInterface for IntegerBinaryArithmeticOp.

mikeurbach:
As we start adding integer binary arithmetic expressions, this will
allow us to abstract over these ops in the evaluator. The interfaces
defines accessors for each operand and the result, which can be
trivially implemented by a shared helper in ODS. The interface also
defines one method to be used in the evaluator, which is to actually
evaluate the integer arithmetic operation. APSInts for each operand
are passed to the interface method, and an APSInt representing the
result or a failure must be returned.

### DIFF
--- a/include/circt/Dialect/OM/OMOpInterfaces.td
+++ b/include/circt/Dialect/OM/OMOpInterfaces.td
@@ -56,4 +56,20 @@ def ClassFieldLike : OpInterface<"ClassFieldLike"> {
   ];
 }
 
+def IntegerBinaryArithmeticInterface : OpInterface<"IntegerBinaryArithmeticOp"> {
+  let cppNamespace = "circt::om";
+  let description = "Common interface for integer binary arithmetic ops.";
+  let methods = [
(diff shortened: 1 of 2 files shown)


include/circt/Dialect/OM/OMOpInterfaces.td
```diff
@@ -56,4 +56,20 @@ def ClassFieldLike : OpInterface<"ClassFieldLike"> {
   ];
 }
 
+def IntegerArithmeticInterface : OpInterface<"IntegerArithmeticOp"> {
```
uenoku, line 59: Can we mention this is an interface for binary arithmetic operation?

uenoku (approved):

include/circt/Dialect/OM/OMOpInterfaces.td
mikeurbach: Yes, good point.

PR #6704, opened 2024-02-15
[OM] Add integer addition op.

mikeurbach:
This op adds two OMIntegerType operands to produce an OMIntegerType result. This defines the evaluateIntegerOperation interface method in terms of APSInt +.

### DIFF
--- a/lib/Dialect/OM/OMOps.cpp
+++ b/lib/Dialect/OM/OMOps.cpp
@@ -536,6 +536,16 @@ PathCreateOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
 }
 
 //===----------------------------------------------------------------------===//
+// IntegerAddOp
+//===----------------------------------------------------------------------===//
+
+FailureOr<llvm::APSInt>
+IntegerAddOp::evaluateIntegerOperation(const llvm::APSInt &lhs,
+                                       const llvm::APSInt &rhs) {
+  return success(lhs + rhs);
+}
+
+//===----------------------------------------------------------------------===//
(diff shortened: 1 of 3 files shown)


uenoku (approved):
LGTM

lib/Dialect/OM/OMOps.cpp
```diff
@@ -535,6 +535,15 @@ PathCreateOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// IntegerAddOp
+//===----------------------------------------------------------------------===//
+
+FailureOr<llvm::APSInt>
+IntegerAddOp::evaluateIntegerOperation(llvm::APSInt lhs, llvm::APSInt rhs) {
```
prithayan, line 543:
```suggestion
IntegerAddOp::evaluateIntegerOperation(const llvm::APSInt &lhs, const llvm::APSInt &rhs) {
```

prithayan (approved):

lib/Dialect/OM/OMOps.cpp
mikeurbach: Ah thanks for the suggestion, that's a good point. I don't know how expensive APSInt is to copy, but no reason this can't be const ref.

mikeurbach: I see.. clang-tidy is also enforcing this. Since this is an interface method, I'll send a patch to main to enforce this in the interface definition itself as well.

PR #6708, opened 2024-02-15
[ImportVerilog] Add translation, run Slang compilation

fabianschuiki:
Add an `--import-verilog` translation target for `circt-translate`. This facilitates writing single-file test cases that don't require configuration of the Slang language frontend.

Actually run the Slang elaboration and compilation from within `ImportContext::importVerilog`. This will essentially do the full frontend work to ingest Verilog, but the part that translates from the Slang AST to CIRCT IR ops is still left to be filled in. The import will always produce an empty MLIR module for now. However, reporting of syntax or other errors on the Verilog side is already covered and can be tested for.

### DIFF
--- a/lib/Conversion/ImportVerilog/ImportVerilog.cpp
+++ b/lib/Conversion/ImportVerilog/ImportVerilog.cpp
@@ -236,8 +236,26 @@ LogicalResult ImportContext::prepareDriver(SourceMgr &sourceMgr) {
 /// Parse and elaborate the prepared source files, and populate the given MLIR
 /// `module` with corresponding operations.
 LogicalResult ImportContext::importVerilog(ModuleOp module) {
-  // This is where the Slang AST to CIRCT op conversion will go.
-  return success();
+  // Parse the input.
+  auto parseTimer = ts.nest("Verilog parser");
+  bool parseSuccess = driver.parseAllSources();
+  parseTimer.stop();
+
+  // Elaborate the input.
+  auto compileTimer = ts.nest("Verilog elaboration");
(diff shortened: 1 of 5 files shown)


darthscsi (approved):

lib/Conversion/ImportVerilog/ImportVerilog.cpp
```diff
@@ -236,8 +236,26 @@ LogicalResult ImportContext::prepareDriver(SourceMgr &sourceMgr) {
...
+  auto parseTimer = ts.nest("Verilog parser");
+  bool parseSuccess = driver.parseAllSources();
+  parseTimer.stop();
+
+  // Elaborate the input.
+  auto compileTimer = ts.nest("Verilog elaboration");
+  auto compilation = driver.createCompilation();
+  for (auto &diag : compilation->getAllDiagnostics())
+    driver.diagEngine.issue(diag);
+  if (!parseSuccess || driver.diagEngine.getNumErrors() > 0)
```
dtzSiFive, line 249: Probably doesn't matter, but should the timer be stopped before returning failure?

dtzSiFive: Is there a scope-based version that'll auto-stop on exit paths? I forget (and may not fit well here/style, just a thought)

fabianschuiki: It does :slightly_smiling_face:. The `ts.nest()` call produces an RAII timer guard that you can either manually stop, or you just let it go out of scope to the same effect.

PR #6709, opened 2024-02-15
[LowerFirReg] Reimplement the mux reachability analysis

prithayan:
The PR implements a new heuristic to determine if a Mux is reachable from a FirReg.
In general an operation is reachable from a register if its in the fanout of the register.
For FirReg lowering, an if/else structure is required for proper enable inference, if a mux is within the fanout from the register.
The fanout path can only consist of `MuxOp`, `ArrayGetOp` or `ArrayCreateOp`.
Thus any ops other than `MuxOp`, `ArrayGetOp` or `ArrayCreateOp` block the reachability.
The analysis is built lazily when its queried and the result is cached to avoid redundant traversal of the IR.


lib/Conversion/SeqToSV/FirRegLowering.h
```diff
@@ -10,26 +10,127 @@
...
+// This analysis is built lazily on every query.
+// The query: is a mux is reachable from a reg, results in a DFS traversal
+// of the IR rooted at the register. This traversal is completed and the result
+// is cached in a Map, for faster retrieval on any future query of any op in
+// this subgraph.
+class ReachableMuxes {
+public:
+  ReachableMuxes(HWModuleOp m) : module(m) {}
+
+  bool isMuxReachableFrom(Operation *regOp, Operation *muxOp) {
```
mikeurbach, line 40: minor, but maybe instead of passing `Operation *`, pass the actual operation subclass, so we can call methods on it directly instead of things like `regOp->getResult(0)`.

lib/Conversion/SeqToSV/FirRegLowering.h
```diff
@@ -10,26 +10,127 @@
...
+
+using namespace hw;
+// This class computes the set of muxes that are reachable from an op.
+// The heuristic propagates the reachability only through the 3 ops, mux,
+// array_create and array_get. All other ops block the reachability.
+// This analysis is built lazily on every query.
+// The query: is a mux is reachable from a reg, results in a DFS traversal
+// of the IR rooted at the register. This traversal is completed and the result
+// is cached in a Map, for faster retrieval on any future query of any op in
+// this subgraph.
```
mikeurbach, line 35: This is the key thing that I didn't implement initially: "any future query of any op in this subgraph". I think this is a great way to approach this problem, so thanks for all the work pushing in this direction!

lib/Conversion/SeqToSV/FirRegLowering.h
```diff
@@ -10,26 +10,127 @@
...
+    // that doesnot block the reachability is a child node. Then, the ops that
+    // are reachable from a node is computed as the union of the Reachability of
+    // all its child nodes.
+    // for all child in the Children(node)
+    // Reachability(node) = node + Union{Reachability(child)}
+    if (visited.find(startNode) != visited.end())
+      return;
+    // The op and its users information that needs to be tracked on the stack
+    // for an iterative DFS traversal.
+    struct OpUserInfo {
```
mikeurbach, line 65: nit: is it necessary to define this in the middle of the method? I personally find it more readable to define this within ReachableMuxes but not within this method.

lib/Conversion/SeqToSV/FirRegLowering.h
```diff
@@ -10,26 +10,127 @@
...
+      auto getNextValid(mlir::ResultRange::user_iterator &iter) const {
+        for (; iter != userRange.end(); ++iter)
+          if (!opBlocksReachability(*iter))
+            return true;
+        return false;
+      }
+    };
+    // The stack to record enough information for an iterative post-order
+    // traversal.
+    std::stack<OpUserInfo> stk;
```
mikeurbach, line 84: Just curious, what's the motivation for stdlib data structures for the stack and visited ops when we often use LLVM data structures?

lib/Conversion/SeqToSV/FirRegLowering.h
```diff
@@ -10,26 +10,127 @@
...
+using namespace hw;
+// This class computes the set of muxes that are reachable from an op.
+// The heuristic propagates the reachability only through the 3 ops, mux,
+// array_create and array_get. All other ops block the reachability.
+// This analysis is built lazily on every query.
+// The query: is a mux is reachable from a reg, results in a DFS traversal
+// of the IR rooted at the register. This traversal is completed and the result
+// is cached in a Map, for faster retrieval on any future query of any op in
+// this subgraph.
+class ReachableMuxes {
```
mikeurbach, line 36: IMHO this could be declared here but defined in the cpp file. Does the whole implementation need to be defined in the header?

mikeurbach (commented):
I left some minor comments. In general, this iteration seems to be heading in a great direction. Thanks for taking the quick n dirty solution I put together and thinking hard about how to do it both correctly and efficiently.

lib/Conversion/SeqToSV/FirRegLowering.h
prithayan: I agree, moved the definition to cpp.

lib/Conversion/SeqToSV/FirRegLowering.h
prithayan: Was an oversight, switched to llvm data structures.

lib/Conversion/SeqToSV/FirRegLowering.h
prithayan: Moved it out of the method.

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -20,34 +20,64 @@ using llvm::MapVector;
...
+    auto &info = stk.back();
+    Operation *currentNode = info.op;
+
+    // Node is being visited for the first time.
+    if (info.userIter == info.userRange.begin())
+      visited.insert(currentNode);
+    if (info.getNextValid(info.userIter)) {
+      Operation *child = *info.userIter;
+      ++info.userIter;
+      if (visited.find(child) == visited.end())
```
uenoku, line 60: nit: `visited.contains/count`

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -20,34 +20,64 @@ using llvm::MapVector;
...
+      auto userIterator = info.userRange.begin();
+      while (info.getNextValid(userIterator)) {
+        Operation *childOp = *userIterator;
+        reachableMuxes[currentNode].insert(childOp);
+        // Propagate the reachability backwards from m to currentNode.
+        auto iter = reachableMuxes.find(childOp);
+
+        if (iter != reachableMuxes.end())
+          reachableMuxes[currentNode].insert(iter->getSecond().begin(),
+                                             iter->getSecond().end());
```
dtzSiFive, line 75: The propagation of reachability sets -- does this work in the presence of cycles?

(where the reachable set for each node in the cycle should be the same, with all nodes in the cycle plus any children not part of the cycle)

I think maybe the idea is that cycles are not possible here, is that right?

(that's why can do the dataflow union thing in a single "pass" (on the return of DFS'ing through it)?)

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -20,34 +20,64 @@ using llvm::MapVector;
...
+
+  stk.emplace_back(startNode);
+
+  while (!stk.empty()) {
+    auto &info = stk.back();
+    Operation *currentNode = info.op;
+
+    // Node is being visited for the first time.
+    if (info.userIter == info.userRange.begin())
+      visited.insert(currentNode);
```
dtzSiFive, line 56: By inserting it here, seems like could put the same node multiple times on the visit stack -- haven't thought it through but would it make sense to update the set at point adding to the stack? (maybe the `visited.insert(child).second` pattern?

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -70,6 +100,17 @@ void FirRegLowering::addToIfBlock(OpBuilder &builder, Value cond,
...
 
+FirRegLowering::FirRegLowering(TypeConverter &typeConverter,
+                               hw::HWModuleOp module,
+                               bool disableRegRandomization,
+                               bool emitSeparateAlwaysBlocks)
+    : typeConverter(typeConverter), module(module),
+      disableRegRandomization(disableRegRandomization),
+      emitSeparateAlwaysBlocks(emitSeparateAlwaysBlocks) {
+
+  reachableMuxes = std::make_unique<ReachableMuxes>(module);
```
dtzSiFive, line 111: What's going on here re:unique_ptr, could this just be initialized like the others and a simple member?

(might be missing something?)

Since went to look, okay so these are created for each module and used in parallel and dropped at end of each module's processing.

dtzSiFive (commented):
Thanks for tackling this! Left some feedback :+1: .

lib/Conversion/SeqToSV/FirRegLowering.h
```diff
@@ -10,26 +10,75 @@
...
+    Operation *op;
+    const mlir::ResultRange::user_range userRange;
+    mlir::ResultRange::user_iterator userIter;
+
+    OpUserInfo(Operation *op)
+        : op(op), userRange(op->getUsers()), userIter(userRange.begin()) {}
+
+    // Increments the itertor to the next valid user op and returns false if
+    // the iterator reaches the end of the range.
+    auto getNextValid(mlir::ResultRange::user_iterator &iter) const {
```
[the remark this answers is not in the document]
prithayan: Thanks for the suggestion, updated to use a filtered iterator.

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -20,34 +20,64 @@ using llvm::MapVector;
...
+      // Any op is reachable from itself.
+      reachableMuxes[currentNode].insert(currentNode);
+      auto userIterator = info.userRange.begin();
+      while (info.getNextValid(userIterator)) {
+        Operation *childOp = *userIterator;
+        reachableMuxes[currentNode].insert(childOp);
+        // Propagate the reachability backwards from m to currentNode.
+        auto iter = reachableMuxes.find(childOp);
+
+        if (iter != reachableMuxes.end())
```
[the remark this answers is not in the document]
prithayan: You are right, the set can never be empty and it must exist, added an assert and removed the check.

lib/Conversion/SeqToSV/FirRegLowering.cpp
prithayan: You are right, this implementation assumes absence of comb cycles. This PR actually implements the check: "Is a mux reachable from a register, given that the register is reachable from the register"

lib/Conversion/SeqToSV/FirRegLowering.cpp
prithayan: A node is added to stack only if the iterator is not at the end of the range, that's why this check is here, to ensure the visited is populated even if the users of an op are empty. I don't think a node can be added multiple times to the stack.

prithayan:
Thanks for all the feedback. Finally merged this PR and sorry for the delay, I just realized it was still open. 
Testing on some internal cores, `LowerSeqToSV` time reduced from `1.2%` to `0.5%` of the total runtime.

PR #6710, opened 2024-02-16
[FIRRTL] Add integer addition conversion to LowerClasses.

mikeurbach:
This is a straightforward 1:1 conversion from FIRRTL IntegerAddOp to OM IntegerAddOp, using the converted operand types from the OpAdaptor.


mikeurbach:
This is so simply I am tempted to use DRR, but I don't think those work with OpConversionPatterns.

uenoku (approved):

PR #6711, opened 2024-02-16
[OM] Support integer binary arithmetic in the Evaluator.

mikeurbach:
This adds support for the Evaluator to evaluate integer binary arithmetic. This is defined in terms of the IntegerBinaryArithmeticOp interface, so the Evaluator can generically handle all ops that implement this interface in terms of the interface methods.

At a high level, getOrCreateValue will create a partially evaluated value for the result of this op. Once the operands are fully evaluated, they are passed to the evaluateIntegerOperation interface method, and if successful, the result is used for the evaluator value.

Because these are the first kind of operation using AttributeValues that doesn't necessarily make the Attribute immediately available, the AttributeValue has been extended to support construction in a partially evaluated state to be filled in later.

The tests show example of straightforward evaluation where the operands are immediately available from constants, where the operands are references to object fields, and where the operands are references to object fields that are initially not evaluated.

This handles many cases, but there is currently no detection for dataflow cycles through an integer binary arithmetic operation. The evaluator would need significant changes in the current setup to detect this on the fly, so currently assumes cycles are prevented earlier in the pipeline.

### DIFF
--- a/lib/Dialect/OM/Evaluator/Evaluator.cpp
+++ b/lib/Dialect/OM/Evaluator/Evaluator.cpp
@@ -56,9 +56,7 @@ LogicalResult circt::om::evaluator::EvaluatorValue::finalize() {
 
 Type circt::om::evaluator::EvaluatorValue::getType() const {
   return llvm::TypeSwitch<const EvaluatorValue *, Type>(this)
-      .Case<AttributeValue>([](auto *attr) -> Type {
-        return cast<TypedAttr>(attr->getAttr()).getType();
-      })
+      .Case<AttributeValue>([](auto *attr) -> Type { return attr->getType(); })
       .Case<ObjectValue>([](auto *object) { return object->getObjectType(); })
       .Case<ListValue>([](auto *list) { return list->getListType(); })
       .Case<MapValue>([](auto *map) { return map->getMapType(); })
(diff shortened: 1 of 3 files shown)


lib/Dialect/OM/Evaluator/Evaluator.cpp
```diff
@@ -394,6 +403,74 @@ circt::om::Evaluator::evaluateConstant(ConstantOp op,
...
+
+  // Extract the integer attributes.
+  auto extractAttr = [](evaluator::EvaluatorValue *value) {
+    return std::move(
+        llvm::TypeSwitch<evaluator::EvaluatorValue *, om::IntegerAttr>(value)
+            .Case([](evaluator::AttributeValue *val) {
+              return val->getAs<om::IntegerAttr>();
+            })
+            .Case([](evaluator::ReferenceValue *val) {
+              return cast<evaluator::AttributeValue>(val->getValue().get())
```
uenoku, line 439: It's necessary to use `getStrippedValue` since `ReferenceValue` might point at another reference. 
```suggestion
              return cast<evaluator::AttributeValue>(val->getStrippedValue().get())
```

uenoku (approved):
LGTM 👍

Issue #6618, opened 2024-01-29
[LTL] Add operators to enhance expressive power


liuyic00:
Current LTL dialect provides a limited set of operators and cannot achieve the expressive power of LTL or SVA. This issue will discuss different levels of expressive power progressively and possible plans to add operators in the LTL dialect.

To differentiate between concepts, `ltl` will represent the LTL dialect, while "LTL" will denote the academic Linear Temporal Logic.

## LTL and Regular Expression

`Seqeunce` and `Property` in SVA could be consider as regular expression and LTL. In my point of view, `Sequence` and `Property` serve different purposes in SVA, although some properties can be expressed in both `Sequence` and `Property`. `Sequence` uses its operators to easily express a fragment that will match on the trace. `Property` uses its propositional logical operators and temporal modal operators to express the temporal property that we want to hold on the trace.

This part adds operators that could achieve the expressive power of LTL and regular expressions. This means `ltl` can achieve the expressive power of SVA, although the lack of syntactic sugar may result in larger formulas.

### Add `NextOp` and `UntilOp` for LTL

Academic LTL has two parts of operators, propositional logical operators and temporal modal operators. `ltl` dialect already includes operators `NotOp`,  `AndOp` and `OrOp`, which satisfies the needs of propositional logical. But temporal modal operators $X$ (next) and $U$ (until) are missing, which are the most basic operators in the academic LTL, consider the operators like $F$ (future/eventually) and $G$ (globally/always) can be expressed using $X$ and $U$.

Thus, by adding `NextOp` and `UntilOp` to `Property` in `ltl`, we can achieve the expressive power of LTL. Although `DelayOp` has a similar effect to `NextOp`, it can only act on `Sequence`, not on `Property`.

### Add `RepeatOp` for Regular Expression

Basic operators of regular expression are $union$ (or), $concatenation$ (connect) and $closure$ (unbounded repetition).  Add `RepeatOp` to `Sequence` in `ltl` will complete the missing semantics.

`DelayOp` on `Sequence` can be considered as the combination of connect and wildcard if we use a regular expression to explain.

[...]

PR #5255, opened 2023-05-24
Add LTL and Verif dialects

fabianschuiki:
Add two new dialects to CIRCT.

The `ltl` dialect represents Linear Temporal Logic expressions and borrows from SystemVerilog's rich assertion system (SVAs). The current implementation is fairly minimal, but is sufficient to capture basic liveness properties and something akin to "regular expressions over time". Since the mapping from SVAs to this dialect is non-trivial, there is an `LTL.md` doc accompanying the dialect to capture some of the design decisions.

The `verif` dialect captures various verification-related operations. It currently contains just a trivial flavor of assert, assume, and cover, without any labels, messages, or value interpolation. The intention is for these operations to be expanded in the future until they can replace the corresponding operations in the SV dialect. The `verif` dialect will also be a good place to introduce verification-specific concepts such as test programs, clock generators, pre- and post-condition checks.

The LTL operations have a few basic folds and canonicalizations defined on them, to be expanded later. Something like a "never" sequence or property would be useful to have unreachable/unmatachable sequences fold to something.

Support for emission in ExportVerilog will be done in a separate commit.

PR #6712, opened 2024-02-16
[LTL] Add fundamental operators

liuyic00:
This PR aims to add fundamental operators into LTL dialect, namely `until`, `next` and `repeat`, to complete expressive power according to #6618 .

The current plan is:
- Add `ltl.until` for properties.
- Reuse `ltl.delay` as `next` operator on property. Inspired by `ltl.or` can be applied to both sequences and properties.
- Add `ltl.repeat` for sequences.

The document for the LTL dialect has been updated. I will add code implementation later.

Please let me know if you have any advice.

### DIFF
--- a/docs/Dialects/LTL.md
+++ b/docs/Dialects/LTL.md
@@ -9,7 +9,7 @@ This dialect provides operations and types to model [Linear Temporal Logic](http
 
 The main goal of the `ltl` dialect is to capture the core formalism underpinning SystemVerilog Assertions (SVAs), the de facto standard for describing temporal logic sequences and properties in hardware verification. (See IEEE 1800-2017 section 16 "Assertions".) We expressly try *not* to model this dialect like an AST for SVAs, but instead try to strip away all the syntactic sugar and Verilog quirks, and distill out the core foundation as an IR. Within the CIRCT project, this dialect intends to enable emission of rich temporal assertions as part of the Verilog output, but also provide a foundation for formal tools built ontop of CIRCT.
 
-As a primary reference, the `ltl` dialect attempts to model SVAs after the [Linear Temporal Logic](https://en.wikipedia.org/wiki/Linear_temporal_logic) formalism as a way to distill SystemVerilog's syntactic sugar and quirks down to a core representation. However, most definitions of LTL tend to be rather academic in nature and may be lacking certain building blocks to make them useful in practice. (See section on [concatenation](#concatenation) below.) To inform some practical design decisions, the `ltl` dialect tries to think of temporal sequences as "regular expressions over time", borrowing from their wide applicability and usefulness.
+As a primary reference, the `ltl` dialect attempts to model SVAs after the [Linear Temporal Logic](https://en.wikipedia.org/wiki/Linear_temporal_logic) formalism as a way to distill SystemVerilog's syntactic sugar and quirks down to a core representation. However, most definitions of LTL tend to be rather academic in nature and may be lacking certain building blocks to make them useful in practice. (See section on [concatenation and repetition](#concatenation-and-repetition) below.) To inform some practical design decisions, the `ltl` dialect tries to think of temporal sequences as "regular expressions over time", borrowing from their wide applicability and usefulness.
 
 
 ### Sequences and Properties
@@ -30,6 +30,8 @@ The core building blocks for modeling temporal logic in the `ltl` dialect are *s
 
 - `always s` checks that the sequence `s` holds in every cycle. This is often referred to as the **G** (or "globally") operator in LTL.
 - `eventually s` checks that the sequence `s` will hold at some cycle now or in the future. This is often referred to as the **F** (or "finally") operator in LTL.
+- `next[n] p` checks that the property `p` will hold on the cycle n. This is often referred to as the **X** (or "next") operator in LTL.
+- `p until q` checks that the property `p` holds in every cycle before `q` holds. This is often referred to as the **U** (or "until") operator in LTL.
 - `s implies t` checks that whenever the sequence `s` is observed, it is immediately followed by sequence `t`.
 
 Traditional definitions of the LTL formalism do not make a distinction between sequences and properties. Most of their operators fall into the property category, for example, quantifiers like *globally*, *finally*, *release*, and *until*. The set of sequence operators is usually very small, since it is not necessary for academic treatment, consisting only of the *next* operator. The `ltl` dialect provides a richer set of operations to model sequences.
@@ -133,6 +135,30 @@ The `ltl.implication` op implements the overlapping case `|->`, such that the tw
 An important benefit of only modeling the overlapping `|->` implication operator is that it does not interact with a clock. The end point of the left-hand sequence is the starting point of the right-hand sequence. There is no notion of delay between the end of the left and the start of the right sequence. Compare this to the `|=>` operator in SVA, which implies that the right-hand sequence happens at "strictly the next clock tick", which requires the operator to have a notion of time and clocking. As described above, it is still possible to model this using an explicit `ltl.delay` op, which already has an established interaction with a clock.
 
 
+### Repetition
+
+Consecutive repetition is an important operator in SVA sequences. For example, `s[*3]` repeats the sequence `s` three times, which is equivalent to `s ##1 s ##1 s`. This also applies when the sequence `s` could match traces with different lengths. For example `(##[0:3] a)[*2]` is equivalent to the disjunction of all the combinations such as `a ##1 a`, `a ##1 (##3 a)`, `(##3 a) ##1 (##2 a)`.
+
+The definition of `ltl.repeat` is similar to that of `ltl.delay`. The mapping from SVA consecutive repetition to the LTL dialect is as follows:
+
+- `seqA[*]`. Repeats zero to unbounded times.
+  ```
+  ltl.repeat %seqA, 0
+  ```
+- `seqA[+]`. Repeats one to unbounded times.
+  ```
+  ltl.repeat %seqA, 1
+  ```
+- `seqA[*n]`. Repeats `n` times.
+  ```
+  ltl.repeat %seqA, n, 0
+  ```
+- `seqA[*m:n]`. Repeats `m` to `n` times.
+  ```
+  ltl.repeat %seqA, m, n-m
+  ```
+
+
 ### Clocking
 
 Sequence and property expressions in SVAs can specify a clock with respect to which all cycle delays are expressed. (See IEEE 1800-2017 section 16.16 "Clock resolution".) These map to the `ltl.clock` operation.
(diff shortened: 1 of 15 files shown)


docs/Dialects/LTL.md
```diff
@@ -175,6 +200,9 @@ The `ltl.delay` sequence operation represents various shorthands for the *next*/
 | `ltl.delay %a, 0`    | **F**a                      |
 | `ltl.delay %a, 2`    | **XXF**a                    |
 
+### Until, Always and Eventually
+
+`ltl.until` and `ltl.always` are *weak*, meaning the property will hold even if there isn't enough clock to evaluate the property. `ltl.eventually` is *strong*, where `ltl.eventually %p` means `p` must hold at some point in the trace.
```
Dobios, line 205: Rewrite: """ltl.until` and `ltl.always` are *weak*, meaning the property will hold even if the trace does not contain clock cycles to evaluate the property."""    
  
Maybe it should also be noted that the default case in LTL would be `ltl.always`.

dobios (commented):
Hi, just wanted to drop a few comments on this, as I've been staring at LTL for quite some time and have a background in verification and the theoretical aspects related to this.

These seem like really interesting additions, and I'm looking forward to see how you plan on mapping these concepts to SV. I am wondering though, for things such as indefinite repetition or delays, how do you plan on implementing that? Is this just being planned as something that is directly mapped to SV and not actually lowered to synthesizable hardware?

Thanks, and I'm looking forward to see how this evolves!

docs/Dialects/LTL.md
liuyic00: I am not sure what this "default case" means. Does it mean that `assert(p)` will check `always p` by default, or will it check `p` every clock?

I notice that SVA has a global `assert` and an `initial assert`. global `assert(s_eventually p)` checks `s_eventually p` every clock, and `initial assert(s_eventually p)` only checks once. It seems we don't have a corresponding feature.

docs/Dialects/LTL.md
```diff
@@ -133,6 +135,29 @@ The `ltl.implication` op implements the overlapping case `|->`, such that the tw
 An important benefit of only modeling the overlapping `|->` implication operator is that it does not interact with a clock. The end point of the left-hand sequence is the starting point of the right-hand sequence. There is no notion of delay between the end of the left and the start of the right sequence. Compare this to the `|=>` operator in SVA, which implies that the right-hand sequence happens at "strictly the next clock tick", which requires the operator to have a notion of time and clocking. As described above, it is still possible to model this using an explicit `ltl.delay` op, which already has an established interaction with a clock.
 
 
+### Repetition
+
+The consecutive repetition is an important operator in SVA sequences. For example, `s[*3]` repeats the sequence `s` three times, which is equivalent to `s ##1 s ##1 s`. This also works when the length of sequence s is unspecified, like `##[0:3] a`. In an intuitive sense, this consecutive repetitio imports the ability to repeat `ltl.delay` an unbounded number of times, but LTL does not have this capability.
```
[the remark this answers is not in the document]
liuyic00: Repetition imports some feature that theoretical LTL cannot describe at all. Therefore, I compared them to clarify the additions, considering that this dialect seems to be an extended version of LTL. Perhaps I could move this discussion to the 'Concatenation' section, where another operation without a direct equivalent in LTL is discussed.

liuyic00:
Thanks for the review. I fixed some of the problems you pointed out and left comments above regarding the parts I am not sure about.

> I am wondering though, for things such as indefinite repetition or delays, how do you plan on implementing that? Is this just being planned as something that is directly mapped to SV and not actually lowered to synthesizable hardware?

The current plan is to directly map the properties to SVA.
However, if synthesizable hardware is used for emulation to replace the simulation with SVA, indefinite repetition and delays can be implemented by an automaton with loops, considering that the sequence expresses a regular expression.

docs/Dialects/LTL.md
Dobios: The current way the ltl stuff is handled in CIRCT (from what I've seen through my experience playing around with it) is that is considers that an assertProperty will by default be an assert always. That is also how it's being treated in the ltl lowering to core.

Dobios:
> Thanks for the review. I fixed some of the problems you pointed out and left comments above regarding the parts I am not sure about.
> 
> > I am wondering though, for things such as indefinite repetition or delays, how do you plan on implementing that? Is this just being planned as something that is directly mapped to SV and not actually lowered to synthesizable hardware?
> 
> The current plan is to directly map the properties to SVA. However, if synthesizable hardware is used for emulation to replace the simulation with SVA, indefinite repetition and delays can be implemented by an automaton with loops, considering that the sequence expresses a regular expression.

My issue was more in how you would like to do this in the `always` case, where synthesizing the automata requires creating multiple copies of them for each cycle in your delays and repetitions. So in theory if you have a variable length repetition that can go to infinity, then implementing this in an `assert always` requires an infinite amount of registers. So either this can be something that is special cased out or maybe we need to figure out some approximation if we are given the max simulation time ahead of compilation. I might just be overthinking this, but I ran into this problem when testing out my first ideas for implementing the variable delays. They also mention this issue in the [BlueSpec Verilog SVA implementation paper](https://drive.google.com/file/d/1YT1t8aQEOYwGD9lzyaRHVmXv9UGXzYHj/view) from way back in 2005 if you want to learn about it.

liuyic00:
I added `ltl.repeat`, `ltl.until` and reused `ltl.delay` in LTL, FIRRTL and ExportVerilog.

Using `ltl.delay` on properties allows `ltl.delay` to express `nexttime[n]` and bounded weak eventually:

```
LTL dialect and SVA:
ltl.delay p, 0    == eventually p
ltl.delay p, n    == nexttime[n] eventually p
ltl.delay p, n, 0 == nexttime[n] p
ltl.delay p, n, m == eventually[n:n+m] p
                  == nexttime[n] eventually[0:m] p
```

We can also express all versions of bounded ` always ` and ` eventually `.

```
SVA on the left and SVA or LTL dailect on the right:

always p        == not s_eventually not p
always[m:n] p   == not s_eventually[m:n] not p 
always[m:$] p   == not s_eventually[m:$] not p
s_always[m:n] p == not eventually[m:n] not p

s_eventually p      == ltl.eventually p                                         -> OK
s_eventually[m:n] p == (not nexttime[n] false) and (eventually[m:n] p)
s_eventually[m:$] p == (not nexttime[m] false) and (nexttime[m] s_eventually p)
eventually[m:n] p   == ltl.delay p, m, n-m                                      -> OK

nexttime[m] p   == ltl.delay p, m, 0                                            -> OK
s_nexttime[m] p == not nexttime[m] not p

// (not nexttime[n] false) holds if there is 0 to n clock cycle.
```

I think this kind of implementation has some problems. If ["we'll want to have that always/globally quantifier eventually"]( https://github.com/llvm/circt/pull/5255#issuecomment-1563212094 ), maybe we will have bounded eventually at the same time, considering the symmetry between always and eventually. The function will be duplicated when using `ltl.delay` to express bounded weak eventually.
Another problem is we do not have an operator to cast a sequence or boolean expression to a property. I think this would be useful for clarifying where the property is. For example, avoiding `nexttime ##1 s` be folded to `ltl.delay s, 2, 0`. I tried to add a `CastToProp` operation, but it is not suitable because the existing code doesn't seem to solve the problem this way. We already have `ltl.delay i, 0, 0` to cast boolean expression to sequence, and `LTLAnyPropertyType` worked like implicit type conversion.

Therefore, maybe we should not reuse `ltl.dealy`. Add a `next[n]` for property like in SVA will be batter:

`ltl.next s, 0` could be used as cast to property.
Bounded `always` and `eventually` could be implemented with one parameter:
```
SVA on the left and possible LTL dailict on the right:

always p        == always[*] p
always[m:n] p   == next[m] always[n-m] p
always[m:$] p   == next[m] always[*] p
s_always[m:n] p == (next[m] always[n-m] p) and (not next[n] false)
                == strong(next[m] strong(always[n-m] p))           // or use `strong(...)` and `weak(...)` operator
                == next['s', m] always['s', n-m] p                 // or add 'strong' and 'weak' as parameter

s_eventually p      == eventually[*] p // here is strong eventually
s_eventually[m:n] p == (next[m] eventually[n-m] p) and (not next[n] false)
s_eventually[m:$] p == (next[m] eventually[*] p) and (not next[m] false)
eventually[m:n] p   == next[m] ((eventually[n-m] p) or (next[n-m] false))

nexttime[m] p   == next[m] p
s_nexttime[m] p == not next[m] not p
```

I mentioned bounded operators here because I believe they will influence how we implement the `next` operator. However, they should be discussed and added in another pull request.

> My issue was more in how you would like to do this in the always case, where synthesizing the automata requires creating multiple copies of them for each cycle in your delays and repetitions. ...

This is indeed a problem that I didn't consider thoroughly.

docs/Dialects/LTL.md
```diff
@@ -133,6 +136,30 @@ The `ltl.implication` op implements the overlapping case `|->`, such that the tw
...
 
 
+### Repetition
+
+Consecutive repetition is an important operator in SVA sequences. For example, `s[*3]` repeats the sequence `s` three times, which is equivalent to `s ##1 s ##1 s`. This also applies when the sequence `s` could match traces with different lengths. For example `(##[0:3] a)[*2]` is equivalent to the disjunction of all the combinations such as `a ##1 a`, `a ##1 (##3 a)`, `(##3 a) ##1 (##2 a)`.
+
+The definition of `ltl.repeat` is similar to that of `ltl.delay`. The mapping from SVA consecutive repetition to the LTL dialect is as follows:
+
+- `seqA[*]`. Repeats zero to unbounded times.
```
sequencer, line 145: IIRC, There are no `unbounded` in current CIRCT LTL dialect, while it will be a block to some specific lowering, e.g. BMC check, synthesis-able Verification.
I suggest mark all unbounded features with a special attribute, maybe `unbounded`?

sequencer (commented):
cc @fabianschuiki, maybe this is a good start to add Until for CIRCT, I'm confident w/ the C++ code, I think we need to care about the existence of unbounded check will lead to non-synthesis-able circuit which may introduce problems in FPGA lowering or LEC.

After this got reviewed and merge, I will start to draft the Chisel frontend and provide some use case for it.

liuyic00:
I have added `repeat`, `until` and `next` (without reusing `delay`). These should be good enough for review and passing CI. I changed the plan from reusing `delay` because the "weak unbounded eventually" introduced by `ltl.delay %p, 0` is unnecessary redundancy.

I think it's preferable to distinguish between sequences and properties rather than explaining them all through LTL. Particularly, sequences express some semantics that LTL cannot capture, and sequences evaluate differently on the left and right sides of `|->`.

sequencer:
I'm not so sure I'm a good person to review the LTL implementation in CIRCT, may I request review from @fabianschuiki, I think he was the original design to this dialect.
For implementations in chisel, I'm happy to review after this is merged.

fabianschuiki (commented):
The `RepeatOp` and `UntilOp` look very nice 😃 Feel free to land those in a separate PR!

I'm not so sure about `NextOp` -- that might warrant some more iteration. It feels like `NextOp` is essentially the same as `DelayOp`, with the only difference being whether the operand is a property or sequence. If that's the only difference, can we extend `DelayOp` to work with properties and sequences? Or could we name them `DelaySeqOp` and `DelayPropOp` instead, and make their definitions be essentially identical, except for the type of the operand?

liuyic00:
`NextOp` and `DelayOp` are similar. I tried using `DelayOp` on both properties and sequences before. You can see [this branch](https://github.com/liuyic00/circt/tree/legacy-add-fund-ltl-op) or previous comments.

There are two reasons I gave up this plan. The "weak unbounded eventually" or "weak unbounded next" introduced by `ltl.delay %p, n` (work on a property) is unnecessary redundancy, as it always evaluates to `true`. This redundancy will mislead the semantics of this operation and cause trouble when processing the expressions. Another reason is `ltl.delay %p, n, m` represents `eventually[n:n+m] p` in SVA. Considering that we are likely to add "bounded always" in some way in the future, it will be strange for "bounded eventually" to be merged into "next", where "always" and "eventually" are two very symmetrical operations.

`ltl.dealy %s, n` or `##[n:$] s` is acceptable because we evaluate the sequence differently when we treat it as a sequence or a property. `##[n:$] s` is like *strong* when finding a match on the trace at left side of `|->` as a sequence, but is like *weak* when we determining if the trace meets it as a property. ("strong" and "weak" are not used to discrube operations inside of sequences in SVA.)

liuyic00:
I have written a longer plan for adding operators [here](https://github.com/llvm/circt/issues/6618#issuecomment-2031841905), which may be more convincing about `NextOp`.

sequencer:
Any update on this PR? hopefully it won't be stall. :)

SpriteOvO (approved):
Just a nitpick. And I also somewhat prefer @fabianschuiki's idea of naming them `DelaySeqOp` and `DelayPropOp`. Others LGTM.

liuyic00:
The `repeat` and `until` are adding in #6989. `next/delayProp` needs further discussion.

PR #6726, opened 2024-02-20
[FIRRTL] Prevent divison by zero in CreateSiFiveMetadata

fzi-hielscher:
Fixes a crash when compiling a [Chisel read only memory ](https://scastie.scala-lang.org/GRNH6zh0TT6mgIq77mfyLw) with a `LoadMemoryFromFile` annotation using the `--repl-seq-mem` option.

The FMemoryModuleOp's `maskBits` attribute is set to zero in this case. While the result of the division is not used, it can still cause a trap.


fzi-hielscher:
Mildly interesting extra info: It crashed on a x86_64 gcc release build. It did not on an aarch64 clang debug build.

darthscsi (approved):
LGTM

PR #6727, opened 2024-02-21
[Emit] Organize output files using the `emit` dialect

nandor:


include/circt/Dialect/Emit/EmitDialect.td
```diff
@@ -0,0 +1,31 @@
...
+#ifndef CIRCT_DIALECT_EMIT_EMITDIALECT
+#define CIRCT_DIALECT_EMIT_EMITDIALECT
+
+def EmitDialect : Dialect {
+  let name = "emit";
+  let cppNamespace = "::circt::emit";
+
+  let summary = "Types and operations for the `emit` dialect";
+  let description = [{
+    The `emit` dialect is intented to model the structure of the emitted RTL.
```
mwachs5, line 22: Spelling : intended

include/circt/Dialect/Emit/EmitDialect.td
```diff
@@ -0,0 +1,31 @@
...
+
+  let summary = "Types and operations for the `emit` dialect";
+  let description = [{
+    The `emit` dialect is intended to model the structure of the emitted RTL.
+
+    It organizes the files, file lists, directories and collateral.
+  }];
+
+  let useDefaultTypePrinterParser = 0;
+  let useDefaultAttributePrinterParser = 0;
```
mikeurbach, line 28: Just curious, why not use these? If we want to use TypeDef and AttrDef I think they remove some boilerplate.

lib/Dialect/Emit/EmitOps.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+//===----------------------------------------------------------------------===//
+// FileListOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult FileListOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  auto *parent = getOperation()->getParentOp();
+  for (auto sym : getFiles()) {
+    Operation *op =
+        symbolTable.lookupSymbolIn(parent, sym.cast<FlatSymbolRefAttr>());
```
mikeurbach, line 45: It looks like FileOp and FileListOp are expected to be siblings that exist within the same level of SymbolTable. Should we write that down in the ODS description? Could we go farther and maybe use HasParent with the same kind of op for both? Or do we not care, as long as their siblings?

mikeurbach (commented):
The rationale and usual dialect boilerplate look good to me. A couple minor questions on the implementation, but looks like a great start.

include/circt/Dialect/Emit/EmitOps.td
```diff
@@ -0,0 +1,99 @@
...
+    SymbolNameAttr:$sym_name,
+    StrAttr:$file_name
+  );
+  let results = (outs);
+
+  let assemblyFormat = "$file_name `sym` $sym_name $body attr-dict";
+
+  let skipDefaultBuilders = 1;
+  let builders = [
+    OpBuilder<(ins "StringRef":$symName, "StringRef":$fileName,
```
rwy7, line 46: I think it's worthwhile to document that the builder's insertion point is moved for the duration of the callback.

include/circt/Dialect/Emit/EmitOps.td
```diff
@@ -0,0 +1,99 @@
...
+#define CIRCT_DIALECT_EMIT_EMITOPS_TD
+
+include "circt/Dialect/HW/HWTypes.td"
+include "circt/Dialect/Emit/EmitDialect.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+class EmitOp<string mnemonic, list<Trait> traits = []> :
+    Op<EmitDialect, mnemonic, traits>;
+
+def FileOp : EmitOp<"file", [
```
uenoku, line 23: We don't want to capture SSA values outside, correct? If so could you add `IsolatedFromAbove`?

lib/Dialect/Emit/EmitOps.cpp
nandor: HasParent is tricky since at the FIRRTL level these should live inside of a CircuitOp, but at the HW level they live inside a ModuleOp.

mikeurbach: Makes sense. I guess we should just document in FileListOp that it expects the FileOps it references to be in the same symbol table as the FileListOp itself.

include/circt/Dialect/Emit/EmitOps.td
```diff
@@ -0,0 +1,114 @@
...
+def VerbatimOp : EmitOp<"verbatim"> {
+  let summary = "Verbatim opaque text emitted inline.";
+  let description = [{
+    This operation produces opaque text inline in the file.
+
+    emit.verbatim allows symbol reference substitutions with {{0}} syntax.
+  }];
+  let arguments = (ins
+    StrAttr:$format_string,
+    OptionalAttr<NameRefArrayAttr>:$symbols
```
darthscsi, line 73: My worry is that what it means to substitute a symbol in a verbatim depends on the two dialects involved.  Does emit.verbatim know about sv symbols and ops and define all those substitutions, or do ops specify how to substitute themselves?  For both of these, what if the correct substitution depends on the dialect of the verbatim and the dialect of the op with the symbol?

Which is a long ways of saying I see how `sv.verbatim` knows what to do for sv.module symbols.  If the verbatim is for sv code, why not leave the verbatim there?  There are couplings between ops.

lib/Dialect/Emit/EmitOps.cpp
```diff
@@ -0,0 +1,70 @@
...
+// FileOp
+//===----------------------------------------------------------------------===//
+
+void FileOp::build(OpBuilder &builder, OperationState &result,
+                   StringRef symName, StringRef fileName,
+                   llvm::function_ref<void()> bodyCtor) {
+  MLIRContext *context = builder.getContext();
+  OpBuilder::InsertionGuard guard(builder);
+
+  result.addAttribute("sym_name", StringAttr::get(context, symName));
```
darthscsi, line 29: Can this dialect opt into properties instead of attributes since it's new?

lib/Dialect/Emit/EmitOps.cpp
darthscsi: It should fail if it's not.  The lookup should be structure to ensure that, which might mean not poking at random op parents to do lookup but following the symbol lookup rules.

darthscsi (commented):
Looks good.  Given the issues we've had, should be checking symbol nesting validity with explicit tests.
I'm not sold on emit.verbatim making sv.verbatim redundant.  A generic emit.verbatim does make sense, but I worry about dialect-specific substitution semantics.

lib/Dialect/Emit/EmitOps.cpp
nandor: We can use some properties, but I'll keep the filenames and text as attributes since string property handling seems to be quite rough (nonexistent?) in MLIR.

lib/Dialect/Emit/EmitOps.cpp
mikeurbach: This looks good to me now using `lookupNearestSymbolFrom(getOperation(), ...)`

include/circt/Dialect/Emit/EmitOps.td
```diff
@@ -0,0 +1,107 @@
...
+  let summary = "Represents the contents of an emitted file";
+
+  let description = [{
+    This operation groups a set of nested operations to be emitted to a file.
+  }];
+
+  let regions = (region SizedRegion<1>:$body);
+  let arguments = (ins
+    StrAttr:$file_name,
+    OptionalAttr<SymbolNameAttr>:$sym_name
```
mikeurbach, line 39: This optionality wasn't in the initial version. Is the idea that we can have a FileOp that doesn't define a symbol, and therefore can't appear in file lists, but otherwise does its usual job? Maybe worth explaining what the symbol is for in the op description.

include/circt/Dialect/Emit/EmitOps.td
```diff
@@ -0,0 +1,107 @@
...
+  let summary = "Represents a file list";
+
+  let description = [{
+    This operation emits a file list referencing a set of files.
+  }];
+
+  let arguments = (ins
+    StrAttr:$file_name,
+    FlatSymbolRefArrayAttr:$files,
+    OptionalAttr<SymbolNameAttr>:$sym_name
```
mikeurbach, line 95: This is also optional now, but I'm not sure when it would make sense to have a symbol on a FileListOp. Maybe I missed this, but what would reference a FileListOp by symbol?

mikeurbach (commented):
My previous concerns are addressed, just some questions about the optional symbols.

include/circt/Dialect/Emit/EmitOps.td
nandor: Symbol names are a bit of a pain to conjure up, as they need to be unique, so when we don't need to reference a file, it's best to not have to populate a name.

include/circt/Dialect/Emit/EmitOps.td
nandor: Other file list ops should be able to point to them.

lib/Dialect/Emit/EmitOps.cpp
```diff
@@ -0,0 +1,71 @@
...
+//===----------------------------------------------------------------------===//
+// FileListOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult FileListOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  for (auto sym : getFiles()) {
+    Operation *op = symbolTable.lookupNearestSymbolFrom(
+        getOperation(), sym.cast<FlatSymbolRefAttr>());
+    if (!op || !isa<emit::FileOp>(op))
+      return emitError("invalid file reference \"") << sym << '\"';
```
seldridge, line 60: Consider breaking these into two checks for a more specific error message.

include/circt/Dialect/Emit/EmitOps.td
```diff
@@ -0,0 +1,113 @@
...
+  );
+  let results = (outs);
+
+  let assemblyFormat = "$file_name `,` $files (`sym` $sym_name^)? attr-dict";
+
+  let extraClassDeclaration = [{
+    // SymbolOpInterface
+    static bool isOptionalSymbol() { return true; }
+  }];
+}
```
seldridge, line 111: I think it's fine to proceed with this filelist op. There is an alternative design point which deserves some thought/commentary, though:

The filelist op is, itself a file op. It just has some special handling of how it will emit its `FlatSymbolRefArrayAttr:$file` argument. Put differently, there is an alternative formulation of this that, instead of using a filelist op, puts _filename ops_ inside a file op and treats that as the filelist. This conserves existing concepts more (there is only one op which becomes a file), however, it does not decrease the total number of ops.

This is hinting at the different design points around who knows how to serialize a symbol? Does the op with the symbol know? Does the op that references the symbol know?

WDYT?

seldridge (approved):
Generally, I'm good with this. This will do wonders to help wrangle all the various files that come out of a build.

There are some design point considerations around if a filelist op is actually needed. I don't have a super strong opinion here.

include/circt/Dialect/Emit/EmitOps.td
nandor: Once we build out emission a bit it's going to be easier to build up the file lists. Subsequently, we could keep a builder and build a file op out of it.

However, I think that in its current form the op simplifies implementation and even testing, so it might be worth keeping.

PR #6728, opened 2024-02-21
[ExportVerilog][Emit] Export emitted files

nandor:


lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6026,6 +6027,79 @@ void ModuleEmitter::emitHWModule(HWModuleOp module) {
...
+    auto lhsRhs = string.split('\n');
+    if (isFirst)
+      isFirst = false;
+    else {
+      ps << PP::end << PP::newline << PP::neverbox;
+    }
+    ps << PPExtString(lhsRhs.first);
+    string = lhsRhs.second;
+  }
+
```
uenoku, line 6097: Can we avoid code duplication with SV verbatim op emission?

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6126,6 +6202,11 @@ void SharedEmitterState::gatherFiles(bool separateModules) {
     // Separate the operation into dedicated output file, or emit into the
     // root file, or replicate in all output files.
     TypeSwitch<Operation *>(&op)
+        .Case<emit::FileOp, emit::FileListOp>([&](auto file) {
+          // Emit file ops to their respective files.
+          symbolCache.addDefinition(file.getSymNameAttr(), file);
```
seldridge, line 6207: It seems weird to add this symbol to the symbolCache as I always thought that was more of a mechanism for generating unique names as opposed to something that is used for symbol lookup. Is this the right way to do this?

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6026,6 +6027,79 @@ void ModuleEmitter::emitHWModule(HWModuleOp module) {
...
+};
+
+void FileEmitter::emit(emit::FileOp op) {
+  for (Operation &op : *op.getBodyBlock()) {
+    TypeSwitch<Operation *, void>(&op)
+        .Case<emit::VerbatimOp>([&](auto op) { emitOp(op); })
+        .Default([&](auto op) {
+          op->emitError("cannot emit operation to file");
+          state.encounteredError = true;
+        });
```
seldridge, line 6052: The void return type on `TypeSwitch` may be implicit and not necessary to specify.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6026,6 +6027,79 @@ void ModuleEmitter::emitHWModule(HWModuleOp module) {
...
+private:
+  void emitOp(emit::VerbatimOp op);
+};
+
+void FileEmitter::emit(emit::FileOp op) {
+  for (Operation &op : *op.getBodyBlock()) {
+    TypeSwitch<Operation *, void>(&op)
+        .Case<emit::VerbatimOp>([&](auto op) { emitOp(op); })
+        .Default([&](auto op) {
+          op->emitError("cannot emit operation to file");
```
seldridge, line 6050: Mega-nit: `emitOpError` may be better with a message that works with that format, i.e., `cannot be emitted to a file`.

This is only illegal at this point because more full support is not implemented correct? I.e., I'd expect this to work HW Modules. If this is intended to be very restrictive, then this smells like a missing verifier on `FileOp`.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6026,6 +6027,79 @@ void ModuleEmitter::emitHWModule(HWModuleOp module) {
...
+}
+
+void FileEmitter::emit(emit::FileListOp op) {
+  // Find the associated file ops and write the paths on individual lines.
+  for (auto sym : op.getFiles()) {
+    auto fileName = cast<FlatSymbolRefAttr>(sym).getAttr();
+    auto file = cast<emit::FileOp>(state.symbolCache.getDefinition(fileName));
+    ps << PP::neverbox;
+    ps << PPExtString(file.getFileName());
+    ps << PP::end << PP::newline;
```
seldridge, line 6064: Can possibly all be combined as `ps << ... << ... << .. << ...;`.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
seldridge: Agree.

seldridge (approved):
This looks good.

Please address @uenoku's comments about code duplication before merging.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
nandor: Yes, we'll have to pull in module/statement emitters here.

nandor:
I didn't factor out the common functionality with verbatim since I wanted to avoid adding some rather complex helpers. Instead, I replaced the body in verbatim emission with a shorter implementation that works.

PR #6730, opened 2024-02-21
[FIRRTL][CAPI] Add function for importing annotations

SpriteOvO:
Added a new function `firrtlImportAnnotationsJSONRaw`, which accepts a JSON serialized annotations string from external programs and outputs a parsed `ArrayAttr`.


lib/CAPI/Dialect/FIRRTL.cpp
```diff
@@ -308,3 +312,23 @@ FIRRTLValueFlow firrtlValueFoldFlow(MlirValue value, FIRRTLValueFlow flow) {
...
+
+bool firrtlImportAnnotationsJSONRaw(MlirContext ctx,
+                                    MlirStringRef annotationsStr,
+                                    MlirAttribute *importedAnnotationsArray) {
+  auto annotations = json::parse(unwrap(annotationsStr));
+  if (!annotations) {
+    return false;
+  }
+
+  auto ctxUnwrapped = unwrap(ctx);
```
sequencer, line 324:
```
/__w/circt/circt/lib/CAPI/Dialect/FIRRTL.cpp:324:3: warning: 'auto ctxUnwrapped' can be declared as 'auto *ctxUnwrapped' [llvm-qualified-auto]
  auto ctxUnwrapped = unwrap(ctx);
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.

  ^~~~~
  auto *
/__w/circt/circt/lib/CAPI/Dialect/FIRRTL.cpp:324:3: note: FIX-IT applied suggested code changes
```
please make clang-tidy happy

SpriteOvO: Done.

sequencer (approved):
LGTM, I tested this PR w/ https://github.com/chipsalliance/chisel/pull/3842 under https://github.com/chipsalliance/t1/pull/425

MLIRBC elaborated successfully.

lib/CAPI/Dialect/FIRRTL.cpp
```diff
@@ -7,17 +7,21 @@
 //===----------------------------------------------------------------------===//
 
 #include "circt-c/Dialect/FIRRTL.h"
+#include "../../Dialect/FIRRTL/Import/FIRAnnotations.h"
```
mikeurbach, line 10: It's a little unusual to depend on a header in the lib directory. I'm not sure if it's worth refactoring this, or if it's fine to depend on the implementation in lib/Dialect/FIRRTL/Import from the CAPI.

mikeurbach (commented):
I have one question, but the main implementation looks good to me. Please add a small CAPI unit test with some small example JSON string.

lib/CAPI/Dialect/FIRRTL.cpp
dtzSiFive: This is not fine, and headers in `lib` are there because they're not part of the library's interface.

Please move the declaration to an appropriate header location.

(although one might reasonably wonder if perhaps there's a reason this wasn't already there....)

SpriteOvO: Added a simple test for the new function and moved `FIRAnnotations.h` file from `lib/Dialect/FIRRTL/Import/` to `include/circt/Dialect/FIRRTL/Import/`.

Thanks for the advice @dtzSiFive and the review from @mikeurbach :+1:

mikeurbach: Sorry, I don't think simply moving the header into the include directory is what @dtzSiFive meant. The original problem was you were depending on one C++ API that was an implementation detail. But to address that, you have made the entire set of C++ implementation details public APIs. Please don't do that either.

I think what you could do is revert the last commit, and think about in which header the declaration of fromJSONRaw should live, and in which library the definition should live. For example, one idea is to create a new `include/circt/Dialect/FIRRTL/Import/FIRAnnotations.h`, only define fromJSONRaw, and keep the definition in the CIRCTImportFIRFile library.

mikeurbach (changes requested):
Please don't wholesale move the FIRAnnotation.h header from private to public.

lib/CAPI/Dialect/FIRRTL.cpp
SpriteOvO: My bad, I misunderstood the comment. In the latest commit, I only moved the function `fromJSONRaw` to the new header `include/circt/Dialect/FIRRTL/Import/FIRAnnotations.h`, and I renamed it to `importAnnotationsFromJSONRaw` in order to make it cleaner in the public namespace.

Hopefully I'm on the right path this time.

mikeurbach: Thanks, yes, I think that is more in line with what I was thinking. @dtzSiFive does it make sense to you now to expose this one function from the existing library?

dtzSiFive: If we're exposing this functionality, this seems the way to do it.  Thanks! :+1:

SpriteOvO:
Final tiny change: Renamed the added CAPI `firrtlImportAnnotationsJSONRaw` to `firrtlImportAnnotationsFromJSONRaw` to align with the C++ function name.

Thanks to @mikeurbach @dtzSiFive for reviewing :) I'm gonna merge it when the CI passes.

test/CAPI/firrtl.c
```diff
@@ -79,10 +79,42 @@ void testValueFoldFlow(MlirContext ctx) {
...
+      mlirModuleCreateParse(ctx, mlirStringRefCreateFromCString(testFIR));
+  MlirBlock mlirModule = mlirModuleGetBody(module);
+  MlirOperation firCircuit = mlirBlockGetFirstOperation(mlirModule);
+
+  const char *rawAnnotationsJSON = "[{\
+    \"class\":\"firrtl.transforms.DontTouchAnnotation\",\
+    \"target\":\"~AnnoTest|AnnoTest>in\"\
+  }]";
+  MlirAttribute rawAnnotationsAttr;
+  assert(firrtlImportAnnotationsFromJSONRaw(
```
mikeurbach, line 99: FYI @SpriteOvO , some builds may disable asserts, and we do test with asserts disabled in nightly CI. I missed this in review: this call shouldn't be wrapped in assert since it does meaningful work needed for the test to function. I changed it here: https://github.com/llvm/circt/commit/ed19a32f865acf1055635310f143f8959dcec121

SpriteOvO: Oh! Once again, I forgot about this. 🤦 Thanks for your fix.

mikeurbach: No problem, I will try to keep a better look out for it in review as well.

PR #6731, opened 2024-02-21
[FIRRTL] Verify public and extmodule requirements about uninferred width and reset

dtzSiFive:
Tests need updating, lots of breakage.

"main module is public" pre-FIRRTL 4.0 does not have requirement that all ports have known width/reset.

Checking during parsing instead, gated on version (and explicit "public" keyword?) would have better compatibility.

extmodule's are not presently forbidden from abstract reset in spec AFAICT, although all FIRRTL-defined module definitions ("public") are required to have concrete reset types so maybe this can/will change.


lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
```diff
@@ -130,6 +130,23 @@ LogicalResult circt::firrtl::verifyModuleLikeOpInterface(FModuleLike module) {
...
 
+  // Enforce public and ext module requirements.
+  if (module.isPublic() || isa<FExtModuleOp>(module)) {
+    for (auto [idx, type] : llvm::enumerate(
+             module.getPortTypesAttr().getAsValueRange<TypeAttr>())) {
+      if (auto ftype = dyn_cast<FIRRTLType>(type)) {
+        if (ftype.hasUninferredWidth())
+          return emitError(module.getPortLocation(idx),
+                           "public or ext module port must have known width");
+        // Presently abstract reset on extmodule is not explicitly disallowed in
```
darthscsi, line 141: That seems like a deficiency in the spec, not the intended behavior.  I would be ok with updating the spec and enforcing this.

dtzSiFive: https://github.com/chipsalliance/firrtl-spec/pull/181

dtzSiFive: Adding the checks to the parse per current spec behavior, will extend once that is made illegal there.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -4964,6 +4964,15 @@ ParseResult FIRCircuitParser::parseExtModule(CircuitOp circuit,
   if (parseParameterList(parameters) || parseRefList(portList, internalPaths))
     return failure();
 
+  if (version >= FIRVersion{4, 0, 0}) {
+    for (auto [pi, loc] : llvm::zip_equal(portList, portLocs)) {
+      if (auto ftype = dyn_cast<FIRRTLType>(pi.type)) {
```
mikeurbach, line 4969: nit: do we need to use our type_dyn_cast here and below to be safe?

mikeurbach (approved):
Makes sense to me.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
dtzSiFive: Good catch, thanks!

Don't need to but I think I'd rather write it that way and let the constexpr thing just `dyn_cast` so no one has to ask themselves this or have subtle bugs if things change re:what is/can be an alias :+1: .

mikeurbach: I see, yeah I wasn't sure if that would even be legal here. Up to you if it's worth it, I just pattern match on "should this be our custom dyn_cast".

PR #6733, opened 2024-02-21
[FIRRTL][LowerLayers] Clean up names of artifacts generated by layers

rwy7:
- Layer module name: move into helper.
- Layer instance name:
  - don't put the parent module name in the instance name,
  - lowercase each part of the instance name, not just the first word.
- Layer file name: move into helper.
- Captured port names: don't prefix the port name with an underscore.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -35,10 +35,6 @@ static bool isAncestor(Operation *op, Value value) {
   return op->isAncestor(argument.getOwner()->getParentOp());
 }
 
-//===----------------------------------------------------------------------===//
-// Type Conversion
-//===----------------------------------------------------------------------===//
-
 namespace {
 
 /// Indicates the kind of reference that was captured.
@@ -62,6 +58,64 @@ struct ConnectInfo {
 // hierarchical path operations after layers are converted to modules.
 using InnerRefMap =
     DenseMap<hw::InnerRefAttr, std::pair<hw::InnerSymAttr, StringAttr>>;
+
+//===----------------------------------------------------------------------===//
+// Naming Helpers
+//===----------------------------------------------------------------------===//
+
+static void appendName(StringRef name, SmallString<32> &output,
+                       bool toLower = false) {
+  if (name.empty())
+    return;
+  if (!output.empty())
+    output.append("_");
+  output.append(name);
+  if (!toLower)
+    return;
+  auto i = output.size() - name.size();
+  output[i] = llvm::toLower(output[i]);
+}
+
+static void appendName(SymbolRefAttr name, SmallString<32> &output,
+                       bool toLower = false) {
+  appendName(name.getRootReference(), output, toLower);
+  for (auto nested : name.getNestedReferences())
+    appendName(nested.getValue(), output, toLower);
+}
+
+/// For a layer `@A::@B::@C` in module Module,
+/// the generated module is called `Module_A_B_C`.
+static SmallString<32> moduleNameForLayer(StringRef moduleName,
+                                          SymbolRefAttr layerName) {
+  SmallString<32> result;
+  appendName(moduleName, result);
+  appendName(layerName, result);
+  return result;
+}
+
+/// For a layerblock `@A::@B::@C`,
+/// the generated instance is called `a_b_c`.
+static SmallString<32> instanceNameForLayer(SymbolRefAttr layerName) {
+  SmallString<32> result;
+  appendName(layerName, result, /*toLower=*/true);
+  return result;
+}
+
+/// For all layerblocks `@A::@B::@C` in a circuit called Circuit,
+/// the output filename is `layers_Circuit_A_B_C.sv`.
+static SmallString<32> fileNameForLayer(StringRef circuitName,
+                                        SymbolRefAttr layerName) {
+  SmallString<32> result;
+  result.append("layers");
+  appendName(circuitName, result);
+  appendName(layerName, result);
+  result.append(".sv");
+  return result;
+}
+
+//===----------------------------------------------------------------------===//
+// LowerLayersPass
+//===----------------------------------------------------------------------===//
 
 class LowerLayersPass : public LowerLayersBase<LowerLayersPass> {
   /// Safely build a new module with a given namehint.  This handles geting a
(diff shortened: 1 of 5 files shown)


seldridge:
> We need a verifier that bans layernames which are empty or include an underscore.

Good catch. If this is going to be a problem with the ABI, then we should put it in the spec.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -63,6 +59,94 @@ struct ConnectInfo {
...
+  for (auto nested : layerName.getNestedReferences()) {
+    auto part = nested.getValue();
+    if (!part.empty()) {
+      result.append("_");
+      result.append(part);
+    }
+  }
+  result.append(".sv");
+  return result;
+}
```
seldridge, line 144: It seems like there is a missing, common utility that does the logic of converging the `SymbolRefAttr` to a string. Would this help clean things up?

``` c++
static void layerToString(SymbolRefAttr layerName, SmallString<32> &str);
```

It _might_ be cleaner to do this by defining the streaming operator `<<` on a symbol and using streaming to build up the string.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -63,6 +59,94 @@ struct ConnectInfo {
...
     DenseMap<hw::InnerRefAttr, std::pair<hw::InnerSymAttr, StringAttr>>;
 
+//===----------------------------------------------------------------------===//
+// Naming Helpers
+//===----------------------------------------------------------------------===//
+
+/// For a layer `@A::@B::@C` in module Module,
+/// the generated module is called `Module_A_B_C`.
+static SmallString<32> moduleNameForLayer(StringRef moduleName,
+                                          SymbolRefAttr layerName) {
```
seldridge, line 69: General nit: passing a `SmallString` reference that is mutated may be cleaner to do here.

seldridge (commented):
Output looks way better. It is totally reasonable to drop the leading `_` and to simplify the naming of the instances.

Some comments about how to further simplify things in the review.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
rwy7: Added some more helpers

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -254,7 +300,7 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
 
     // Create an input port for an operand that is captured from outside.
     auto createInputPort = [&](Value operand, Location loc) {
-      auto operandName = getFieldName(FieldRef(operand, 0), true);
+      const auto &[portName, _] = getFieldName(FieldRef(operand, 0), true);
```
seldridge, line 303: Nice cleanup.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -476,9 +522,7 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
     // of the pass.  This is done to avoid having to revisit and rewrite each
     // instance everytime it is moved into a parent layer.
     builder.setInsertionPointAfter(layerBlock);
-    auto moduleName = newModule.getModuleName();
-    auto instanceName =
-        (Twine((char)tolower(moduleName[0])) + moduleName.drop_front()).str();
+    auto instanceName = instanceNameForLayer(layerBlock.getLayerName());
```
seldridge, line 525: This is so much cleaner. 💯

seldridge (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -63,6 +59,64 @@ struct ConnectInfo {
...
+static void appendName(StringRef name, SmallString<32> &output,
+                       bool toLower = false) {
+  if (name.empty())
+    return;
+  if (!output.empty())
+    output.append("_");
+  output.append(name);
+  if (!toLower)
+    return;
+  auto i = output.size() - name.size();
```
dtzSiFive, line 75: What's this toLower business, by the way?

dtzSiFive: (for instance naming, looks like?)

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -63,6 +59,64 @@ struct ConnectInfo {
...
+/// the generated instance is called `a_b_c`.
+static SmallString<32> instanceNameForLayer(SymbolRefAttr layerName) {
+  SmallString<32> result;
+  appendName(layerName, result, /*toLower=*/true);
+  return result;
+}
+
+/// For all layerblocks `@A::@B::@C` in a circuit called Circuit,
+/// the output filename is `layers_Circuit_A_B_C.sv`.
+static SmallString<32> fileNameForLayer(StringRef circuitName,
```
dtzSiFive, line 106: Does this change for FIRRTL 4.0?

abi.md presently reads:

```
filename = "layers_" , module , "_", root , { "_" , nested } , ".sv" ;
```

(module not circuit)

Is this going to work for public modules that aren't the main module as-is?

dtzSiFive (approved):
LGTM, nice work!

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
rwy7: Yes it's just for naming the instance of a layer-turned-module.

fabianschuiki (approved):

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: I think this pass does not do the right thing for public modules. This happens to work now for the main module. The pass should be updated to emit the collateral for all public modules in a follow-on. Good catch, Will.

rwy7: Opened an issue: https://github.com/llvm/circt/issues/6748

PR #6734, opened 2024-02-21
[FIRRTL] Support PropertyType in emitConnect.

mikeurbach:
We have a centralized emitConnect helper that knows how to create the right kind of connections for various special, non-base types, like RefType and AnalogType. This adds another special case for PropertyType, which emits a PropAssignOp.

This is used in various places, but the pass where I noticed this was missing is InjectDUTHierarchy, so I've added a test there similar to the test for RefType.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLUtils.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLUtils.cpp
@@ -43,8 +43,14 @@ void circt::firrtl::emitConnect(ImplicitLocOpBuilder &builder, Value dst,
       if (dstFType != srcFType)
         src = builder.create<RefCastOp>(dstFType, src);
       builder.create<RefDefineOp>(dst, src);
-    } else // Other types, give up and leave a connect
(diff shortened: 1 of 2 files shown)


dtzSiFive (approved):
LGTM.

Should this ever insert whatever the om.any cast thing is? If unsure maybe better to not worry just was comparing with other paths!

mikeurbach:
Good question, and I think we don't need that logic. It seems like emitConnect is already assuming the types are compatible, and leaving it to the respective connection operation's verifier to enforce that. I think we don't need to insert casts to AnyType on demand here; those should already be inserted when necessary earlier (only ever in the parser, at the moment).

dtzSiFive:
> Good question, and I think we don't need that logic. It seems like emitConnect is already assuming the types are compatible, and leaving it to the respective connection operation's verifier to enforce that. I think we don't need to insert casts to AnyType on demand here; those should already be inserted when necessary earlier (only ever in the parser, at the moment).

Perfect, that was my understanding/assessment too, just checking.  Thanks!

PR #6736, opened 2024-02-22
 [CreateSiFiveMetadata] Generate firrtl.class instead of om.class 

prithayan:
Update the object model generated by the CreateSiFiveMetadata pass to generate `firrtl` dialect, instead of `om` dialect.
Ensure all `firrtl.class` associated with the metadata are instantiated inside a top level `firrtl.class` `SiFive_Metadata` , which is instantiated inside the top level `firrtl.module`.  This is required for ensuring that subsequent passes can lower it correctly.
Generate all the hierarchical paths to the memory and associate them with the path operation for the memory metadata.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
@@ -18,6 +18,7 @@
 #include "circt/Dialect/FIRRTL/FIRRTLOps.h"
 #include "circt/Dialect/FIRRTL/FIRRTLTypes.h"
 #include "circt/Dialect/FIRRTL/FIRRTLUtils.h"
+#include "circt/Dialect/FIRRTL/Namespace.h"
 #include "circt/Dialect/FIRRTL/Passes.h"
 #include "circt/Dialect/HW/HWAttributes.h"
 #include "circt/Dialect/HW/InnerSymbolNamespace.h"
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -38,128 +39,281 @@ using circt::igraph::InstancePath;
...
+      builder.create<ObjectOp>(memoryMetadataClass,
+                               builder.getStringAttr("memory_metadata"));
+
+    if (retimeModulesMetadataClass)
+      builder.create<ObjectOp>(
+          retimeModulesMetadataClass,
+          builder.getStringAttr("retime_modules_metadata"));
+
+    builder.setInsertionPointToEnd(topMod.getBodyBlock());
+    builder.create<ObjectOp>(sifiveMetadataClass,
```
mikeurbach, line 297: Should a reference to this object become an output of the topMod? In the current output, I don't think anything references the object.

prithayan: Do we want to update the top module signature to return the metadata object ? You are right, nothing references it now, we can define the ABI for the python script to interpret it correctly.

mikeurbach: If the ABI is "instantiate the SiFive_Metadata class", that would be reasonable to me. And in that case, we don't even need to instantiate the class within the top-module, we can leave the class as the "entrypoint".

mikeurbach (commented):
I took another brief look, and I think this is looking good. Is this ready for a final review?

prithayan:
> I took another brief look, and I think this is looking good. Is this ready for a final review?

Thanks Mike, yes, this is ready for final review.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -35,136 +32,369 @@
...
-      return builderOM.create<om::ConstantOp>(constVal.cast<mlir::TypedAttr>());
+    auto *context = builderOM.getContext();
+    auto createConstField = [&](Attribute constVal) -> Value {
+      if (auto boolConstant = dyn_cast_or_null<mlir::BoolAttr>(constVal))
+        return builderOM.create<BoolConstantOp>(boolConstant);
+      if (auto intConstant = dyn_cast_or_null<mlir::IntegerAttr>(constVal))
+        return builderOM.create<FIntegerConstantOp>(intConstant);
+      if (auto strConstant = dyn_cast_or_null<mlir::StringAttr>(constVal))
+        return builderOM.create<StringConstantOp>(strConstant);
+      // static_assert(false, "Attribute not handled");
```
mikeurbach, line 235: nit: please make an assert or remove the commented out code.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -35,136 +32,369 @@
...
+              .Case("extraPorts", {}));
+      if (!propVal) {
+        if (field.value().equals("hierarchy"))
+          propVal = hierpaths;
+        else
+          propVal = extraPorts;
+      }
+
+      auto inPort =
+          builderOM.create<ObjectSubfieldOp>(object, 2 * field.index());
```
mikeurbach, line 318: Can you add a comment about the 2? I know it's because we're making a "simple class" with input tied to output for each field, but it will be helpful to know what's going on here in the future.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -612,26 +840,34 @@ void CreateSiFiveMetadataPass::runOnOperation() {
...
+    if (auto objectOp = omir.instantiateSifiveMetadata(topMod)) {
+      SmallVector<std::pair<unsigned, PortInfo>> ports = {
+          {topMod.getNumPorts(),
+           PortInfo(StringAttr::get(objectOp->getContext(), "metadataObj"),
+                    objectOp.getType(), Direction::Out)}};
+      topMod.insertPorts(ports);
+    }
 
   // This pass does not modify the hierarchy.
   markAnalysesPreserved<InstanceGraph>();
```
mikeurbach, line 873: Since FIRRTL classes are FModuleLike, I think the new classes and objects added by this path do affect the instance graph. So we should either update the InstanceGraph as we create classes and objects, or not mark that analysis preserved.

mikeurbach (commented):
Some minor comments. The one thing I think we need to consider is whether we're keeping the instance graph up to date or not. I don't think the current PR is valid, since it adds classes and objects but then says we can preserve the instance graph.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
prithayan: Added the comment explaining the logic.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
prithayan: Thanks for spotting this, removed the `markAnalysesPreserved`.

prithayan:
Thanks Mike, yes this pass is updating the hierarchy, the analysis is no longer preserved.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
mikeurbach: Sounds good, I think we can move forward with this now that it's safe. If re-computing the instance graph after this pass becomes painful we know what we have to do to keep it up to date.

mikeurbach (approved):

Issue #6715, opened 2024-02-18
[FIRRTL] Remove "ref foo is <string>" syntax


seldridge:
Now that probes have been operational for some time, we no longer need to accept non-standard formats that allow string paths to be exposed from inside external modules.  This has been fully subsumed with the FIRRTL ABI that produces a ref file of Verilog macros.

E.g., there are tests that accept the following in `parse-basic.fir`:

```
  extmodule RefExtMore :
    input in : UInt<1>
    output r : Probe<UInt<1>>
    output data : UInt<3>
    output r2 : Probe<{a : UInt<3>}[3]>
    ref r2 is "in"
    ref r is "path.to.internal.signal"
```

These tests, parsing, and any handling of "internalpaths" can be removed.

dtzSiFive:
These are and have been in the FIRRTL spec since probes/ref types were
introduced.

You could think of them as ref literals, and pre-date the ABI.

They may have utility for use with non-FIRRTL-defined modules, e.g.
https://github.com/chipsalliance/chisel/pull/3486 .

Having Chisel emit "ABI" files directly itself for this use case is a
pain.  And arguably the wrong level but makes it not part of the
declaration which has strengths.

Perhaps a friction that some tooling could assist or just a burden to take
on anyway (as that linked PR did for an earlier ABI, sorry if that broke
with the ABI change by the way!)?

Anyway they're not currently used but they're not non-standard so uh please
do not remove them until they are.

They're also the only way to XMR into/through an intrinsic, which is ...
fun 🙃..  but was requested when I touched this last, so a consideration
for this direction.

dtzSiFive:
> need to accept non-standard formats

Fix?

---- 

As for the rest, wrote from phone so perhaps some tone was lost -- "here are other things you might also not be aware of".

Having a stable versioned ABI for these things sounds good, but still presently it's a bit painful to roll code to emit the right macros for.  That was not mean to say ref statements are better or even good.  And "wrong layer" I mean that it encodes what's known at a high-level directly into low-level implementation details, making it impossible to understand this information without understanding the verilog macros.  "Wrong" is maybe hasty but hopefully you see what I mean.

This means FIRRTL itself cannot be used to represent a circuit's probes into non-FIRRTL bits, without some sort of Verilog preprocessor + file scan or whatever, and this may be undesirable if wanting the extmodule to be a simulated thing or whatever.

[...]

PR #6738, opened 2024-02-23
[FIRRTL] Reject ref statements in 4.0.0+.

dtzSiFive:
cc #6715 .

Continue to support for versions it's part of the FIRRTL spec.

It may make sense to simplify parser (and IR/passes a smidge) and not support at some point?
The test mentioned in that issue (and corresponding `parse-errors.fir`) as well as the spec-conformance ref test (`ext.fir`) will can be adjusted if this is dropped entirely.

Cannot drop "internalPaths" and pass support until this is removed which also requires annotations that rely on it to be deprecated / removed.


seldridge (approved):
LGTM

I do note that ref statements never got into the wild because they never had a Chisel API. For exactness with the spec it is reasonable to reject them this way. However, I expect that practically, these could be entirely dropped.

seldridge:
If not hard dropping this, moving the tests of the parsing of this into a feature-gated test would be good.

darthscsi (approved):

PR #6739, opened 2024-02-23
[Ibis] Split ContainerOp in two

### LINKED ISSUE #6658: Ibis.class must implement symboltable if ibis.container is to be nested below
LLVM bump fails as IBIS is violating the correctness constraint for symbols.

teqdruid:
In order to not abuse symbol tables, create an OuterContainerOp for top-level containers, and InnerContainerOp for nested containers. The former has a symbol and is an InnerSymbolTable. The latter has an InnerSymbol.

Fixes #6658.

### DIFF
--- a/include/circt/Dialect/Ibis/IbisOps.td
+++ b/include/circt/Dialect/Ibis/IbisOps.td
@@ -568,14 +568,12 @@ def ThisOp : IbisOp<"this", [
 // Low-level Ibis operations
 // ===---------------------------------------------------------------------===//
 
-def ContainerOp : IbisOp<"container", [
-  Symbol,
-  // TODO: @mortbopet: ContainerOp should be an innersymbol table (or equivalent)
-  // whenever we figure out how to support nested symbol tables...
-  // InnerSymbolTable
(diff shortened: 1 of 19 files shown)


blakep-msft (approved):
Thanks for doing this.

include/circt/Dialect/Ibis/IbisOps.td
```diff
@@ -584,36 +582,57 @@ def ContainerOp : IbisOp<"container", [
...
+  code extraContainerClassDeclaration = [{}];
+  let extraClassDeclaration = extraContainerClassDeclaration # [{
+    // Implement RegionKindInterface.
+    static RegionKind getRegionKind(unsigned index) { return RegionKind::Graph; }
+  }];
+
+  let skipDefaultBuilders = 1;
+}
+
+def OuterContainerOp : ContainerOp<"container.outer", [
```
mortbopet, line 596: LGTM - only nit i have is that i think both `OuterContainerOp` should - in their `description` field describe _why_ we have two separate container operations.

PR #6741, opened 2024-02-24
[FIRRTL] Put layer collateral in testbench dir

seldridge:
Change the LowerLayers pass to put generated modules inside the testbench directory if one is specified and a DUT is present.  This is a stopgap measure that keeps layers aligned with existing CIRCT compilation of FIRRTL.  This is intended to be replaced with a better FIRRTL ABI based on filelists in the future.

This is currently missing handling of instances instantiated by the layer. Those need to also get moved to the testbench area. I'll mark this non-draft once it's fully working.

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -247,8 +247,14 @@ struct CircuitLoweringState {
     // Pre-populate the dutModules member with a list of all modules that are
     // determined to be under the DUT.
     auto inDUT = [&](igraph::ModuleOpInterface child) {
+      auto isBind = [](igraph::InstanceRecord *instRec) {
+        auto inst = instRec->getInstance();
+        if (auto *finst = dyn_cast<InstanceOp>(&inst))
+          return finst->getLowerToBind();
+        return false;
+      };
       if (auto parent = dyn_cast<igraph::ModuleOpInterface>(*dut))
-        return getInstanceGraph().isAncestor(child, parent);
+        return getInstanceGraph().isAncestor(child, parent, isBind);
       return dut == child;
     };
     circuitOp->walk([&](FModuleLike moduleOp) {
@@ -292,7 +298,8 @@ struct CircuitLoweringState {
   FModuleLike getTestHarness() { return testHarness; }
 
   // Return true if this module is the DUT or is instantiated by the DUT.
-  // Returns false if the module is not instantiated by the DUT.
+  // Returns false if the module is not instantiated by the DUT or is
+  // instantiated under a bind.
   bool isInDUT(igraph::ModuleOpInterface child) {
     return dutModules.contains(child);
   }
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -1073,9 +1073,11 @@ FIRRTLModuleLowering::lowerModule(FModuleOp oldModule, Block *topLevelModule,
     newModule->setAttr("firrtl.extract.cover.extra", builder.getUnitAttr());
 
   // If this is in the test harness, make sure it goes to the test directory.
+  // Do not update output file information if it is already present.
   if (auto testBenchDir = loweringState.getTestBenchDirectory())
     if (loweringState.isInTestHarness(oldModule)) {
-      newModule->setAttr("output_file", testBenchDir);
+      if (!newModule->hasAttr("output_file"))
+        newModule->setAttr("output_file", testBenchDir);
       newModule->setAttr("firrtl.extract.do_not_extract",
                          builder.getUnitAttr());
       newModule.setCommentAttr(
--- a/lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
@@ -1711,27 +1711,42 @@ void GrandCentralPass::runOnOperation() {
   instancePaths = &instancePathCache;
 
   /// Contains the set of modules which are instantiated by the DUT, but not a
-  /// companion or instantiated by a companion.  If no DUT exists, treat the top
-  /// module as if it were the DUT.  This works by doing a depth-first walk of
-  /// the instance graph, starting from the "effective" DUT and stopping the
-  /// search at any modules which are known companions.
-  DenseSet<igraph::ModuleOpInterface> dutModules;
-  FModuleOp effectiveDUT = dut;
-  if (!effectiveDUT)
-    effectiveDUT = cast<FModuleOp>(
-        *instancePaths->instanceGraph.getTopLevelNode()->getModule());
-  auto dfRange =
-      llvm::depth_first(instancePaths->instanceGraph.lookup(effectiveDUT));
-  for (auto i = dfRange.begin(), e = dfRange.end(); i != e;) {
-    auto module = cast<FModuleLike>(*i->getModule());
-    if (AnnotationSet(module).hasAnnotation(companionAnnoClass)) {
-      i.skipChildren();
-      continue;
+  /// companion, instantiated by a companion, or instantiated under a bind.  If
+  /// no DUT exists, treat the top module as if it were the DUT.  This works by
+  /// doing a depth-first walk of the instance graph, starting from the
+  /// "effective" DUT and stopping the search at any modules which are known
+  /// companions or any instances which are marked "lowerToBind".
+  DenseSet<InstanceGraphNode *> dutModules;
+  InstanceGraphNode *effectiveDUT;
+  if (dut)
+    effectiveDUT = instancePaths->instanceGraph.lookup(dut);
+  else
+    effectiveDUT = instancePaths->instanceGraph.getTopLevelNode();
+  {
+    SmallVector<InstanceGraphNode *> modules({effectiveDUT});
+    while (!modules.empty()) {
+      auto *m = modules.pop_back_val();
+      for (InstanceRecord *a : *m) {
+        auto *mod = a->getTarget();
+        // Skip modules that we've visited, that are are under the companion
+        // module, or are bound/under a layer block.
+        if (auto block = a->getInstance()->getParentOfType<LayerBlockOp>()) {
+          auto diag = a->getInstance().emitOpError()
+                      << "is instantiated under a '" << block.getOperationName()
+                      << "' op which is unexpected by GrandCentral (did you "
+                         "forget to run the LowerLayers pass?)";
+          diag.attachNote(block.getLoc())
+              << "the '" << block.getOperationName() << "' op is here";
+          removalError = true;
+        }
+        if (dutModules.contains(mod) ||
+            AnnotationSet(mod->getModule()).hasAnnotation(companionAnnoClass) ||
+            cast<InstanceOp>(*a->getInstance()).getLowerToBind())
+          continue;
+        modules.push_back(mod);
+        dutModules.insert(mod);
+      }
     }
-    dutModules.insert(i->getModule<igraph::ModuleOpInterface>());
-    // Manually increment the iterator to avoid walking off the end from
-    // skipChildren.
-    ++i;
   }
 
   // Maybe return the lone instance of a module.  Generate errors on the op if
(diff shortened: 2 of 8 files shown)


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -646,27 +695,23 @@ void LowerLayersPass::runOnOperation() {
...
       includes.append("\n");
     }
 
     // Write header to a verbatim.
-    builder
-        .create<sv::VerbatimOp>(layerOp.getLoc(), includes + "`ifndef " +
-                                                      prefix + "\n" +
-                                                      "`define " + prefix)
-        ->setAttr("output_file", outputFileAttr);
+    setOutputFile(builder.create<sv::VerbatimOp>(
```
nandor, line 705: Please consider using the `emit::File` and `emit::Verbatim` operations, they are active & supported.

uenoku:
Does thin mean we cannot use layers for non-testbench parts at least until it's replaced with.a better ABI? Not having any actual use case for non-testbench but just curios.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: I can switch over to these now that they are available. Thanks!

seldridge: I'd like to handle this in a cleanup pass if that's alright.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1073,9 +1080,11 @@ FIRRTLModuleLowering::lowerModule(FModuleOp oldModule, Block *topLevelModule,
     newModule->setAttr("firrtl.extract.cover.extra", builder.getUnitAttr());
 
   // If this is in the test harness, make sure it goes to the test directory.
+  // Do not update output file information if it is already present.
   if (auto testBenchDir = loweringState.getTestBenchDirectory())
     if (loweringState.isInTestHarness(oldModule)) {
-      newModule->setAttr("output_file", testBenchDir);
+      if (!newModule->hasAttr("output_file"))
```
dtzSiFive, line 1086: When might it already be set -- is this something new with this change?

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -1711,27 +1711,36 @@ void GrandCentralPass::runOnOperation() {
...
   /// Contains the set of modules which are instantiated by the DUT, but not a
-  /// companion or instantiated by a companion.  If no DUT exists, treat the top
-  /// module as if it were the DUT.  This works by doing a depth-first walk of
-  /// the instance graph, starting from the "effective" DUT and stopping the
-  /// search at any modules which are known companions.
+  /// companion, instantiated by a companion, or instantiated under a bind.  If
+  /// no DUT exists, treat the top module as if it were the DUT.  This works by
+  /// doing a depth-first walk of the instance graph, starting from the
+  /// "effective" DUT and stopping the search at any modules which are known
+  /// companions.
```
dtzSiFive, line 1718:
```suggestion
  /// companions or marked to be lowered to bind.
```
?
(modulo phrasing, sorry)

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
seldridge: This was trying to future proof the conversion. I wasn't aware of this being a known problem though.

High level: it seems like the existing `output_file` attribute should be respected even if we don't have an example of a pass hitting this edge case. If it's not, then that's likely a bug. The other place that an `output_file` is set is on binds and that will _not_ overwrite a pre-existing `output_file`.

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: Fixed in: https://github.com/llvm/circt/pull/6741/commits/dbef859185bae85ee13b1c90e8c482d98e4a8224

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
dtzSiFive: Sounds perfect, thanks for the consideration and explanation! :rocket: :100:

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -1711,27 +1711,36 @@ void GrandCentralPass::runOnOperation() {
...
-  for (auto i = dfRange.begin(), e = dfRange.end(); i != e;) {
-    auto module = cast<FModuleLike>(*i->getModule());
-    if (AnnotationSet(module).hasAnnotation(companionAnnoClass)) {
-      i.skipChildren();
-      continue;
+  {
+    SmallVector<igraph::ModuleOpInterface> modules({effectiveDUT});
+    while (!modules.empty()) {
+      auto m = modules.pop_back_val();
+      for (InstanceRecord *a : *instancePaths->instanceGraph.lookup(m)) {
```
dtzSiFive, line 1728: Could the worklist be InstanceRecord*'s instead, to avoid looking them up as this walks?

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -1711,27 +1711,36 @@ void GrandCentralPass::runOnOperation() {
...
+        auto mod = a->getTarget()->getModule();
+        // Skip modules that we've visited, that are are under the companion
+        // module, or are bound/under a layer block.
+        assert(
+            !a->getInstance()->getParentOfType<LayerBlockOp>() &&
+            "not expecting to see any layerblocks in Grand Central, this will "
```
dtzSiFive, line 1734: Thanks for the defensive check!!

Your call but if this is reachable and not a logic bug this might be better as a diagnostic and triggering pass failure?

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -597,10 +638,21 @@ void LowerLayersPass::runOnOperation() {
   llvm::sys::SmartMutex<true> mutex;
   circuitMutex = &mutex;
 
+  // Extract information from circuit-level annotations.
+  for (auto anno : AnnotationSet(circuitOp)) {
+    if (!anno.isClass(testBenchDirAnnoClass))
+      continue;
+    auto dir = anno.getMember<StringAttr>("dirname");
+    assert(dir && "invalid test bench annotation");
+    testBenchDir = dir.getValue();
```
dtzSiFive, line 647: Last annotation wins works for me.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -597,10 +638,21 @@ void LowerLayersPass::runOnOperation() {
...
+  }
+
   // Determine names for all modules that will be created.  Do this serially to
-  // avoid non-determinism from creating these in the parallel region.
+  // avoid non-determinism from creating these in the parallel region.  While
+  // walking modules, set the DUT if it is so marked.
   CircuitNamespace ns(circuitOp);
   circuitOp->walk([&](FModuleOp moduleOp) {
+    (void)extractDUT(moduleOp, dut);
```
dtzSiFive, line 655: What's going on here -- why ignoring the failure (which looks like will print a diagnostic)?

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -1711,27 +1711,36 @@ void GrandCentralPass::runOnOperation() {
...
+            "cause incorrect output directories");
+        if (dutModules.contains(mod) ||
+            AnnotationSet(mod).hasAnnotation(companionAnnoClass) ||
+            a->getInstance()->hasAttr("lowerToBind"))
```
dtzSiFive, line 1738: Earlier code uses `getLowerToBind` -- is that not available easily here due to, e.g., having the interface version or something?

dtzSiFive (approved):
Made a pass re:code mostly, didn't look super close the behavior -- if this does what's needed and especially if we're on the path towards changing it, LGTM!

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: Yes, there's a better organization of this to use `InstanceGraphNode *`. Will switch to that. Nice suggestion.

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: It would be reachable from a manually crafted pass pipeline that either doesn't run `LowerLayers` or moves `LowerLayers` after `GrandCentral`. Given that, maybe adding the check and signaling pass failure is better.

seldridge: I'll change this to do the check and signal pass failure.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: There's a bunch of scattered opinion on this type of thing throughout the passes. There is a stalled effort on making the annotations more first-class. The incompletely rolled out stop-gap is to have dedicated annotation "extractors" which get the information out of a might-exist annotation, while also checking that the annotation is correct. Cf. https://github.com/llvm/circt/blob/10562214dcd5a54724e2e5d50c2d980ef98103b3/lib/Dialect/FIRRTL/FIRRTLAnnotations.cpp#L693

There's a missing function/piece of infrastructure that you tell it what you want to pull out of an array of annotations and it does it and the validation.

All that said, moving towards first-class support where the annotations are actually validated in IR and not when used is much better.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: I thought there was a reason that this could be ignored here. However, I don't think that's right. I'll add the check.

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: Will switch to: `cast<InstanceOp>(*a->getInstance()).getLowerToBind())`.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: I think my original reasoning here was that `extractDUT` is used and checked elsewhere in the pipeline, so erroring here, too, was weird. That said, passes should work standalone and not require a pass pipeline ordering. Reiterating that the check is now in there.

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: Fixed in https://github.com/llvm/circt/pull/6741/commits/0a451ea5b9d85972aefee0565d608d70b952e8aa.

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: Fixed in https://github.com/llvm/circt/pull/6741/commits/45b7a307dff21f47bbb23dad224b5e362cafcd82. Test added in https://github.com/llvm/circt/pull/6741/commits/0a451ea5b9d85972aefee0565d608d70b952e8aa.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: Changed to signal pass failure in: https://github.com/llvm/circt/pull/6741/commits/6264b954a75864af329dae84b995c12910664452 A test is included in: https://github.com/llvm/circt/pull/6741/commits/fb80c1b3b6023aa949188f1696712aee77a65e5a.

seldridge:
I'll let CI pass, then rebase and land.

PR #6092, opened 2023-09-08
[ExportVerilog] Add verilog debug locations to output MLIR

prithayan:
Use the `PrettyPrinter` `CallbackToken` API, to record print events and store the verilog output locations for each operation.
This updates the location attribute of the op to include the verilog location range. Each op can correspond to multiple verilog locations.
This feature is disabled by default under an emission flag `emitVerilogLocations`.

PR #6744, opened 2024-02-25
[Emit] Emit SV ops nested in a file

nandor:
This change allows a small subset of the SV ops to be nested in files.


dtzSiFive (approved):
LGTM!

dtzSiFive:
Curious: Does the new Emit stuff work re:the location callback bits (#6092)?

Looks like so far it's all things that don't really get location information anyway but wondering how that might shake out -- anyway hype hype this is really cleaning things up, go go go! :rocket:

darthscsi (approved):
LGTM

PR #6747, opened 2024-02-26
[FIRRTL][FIRParser] Enforce 4.0.0 main module must be public.

dtzSiFive:
The "public" nature of the main module is not optional, require it to be explicitly specified.

cc https://github.com/chipsalliance/firrtl-spec/pull/185 .


darthscsi (approved):
LGTM

PR #6750, opened 2024-02-26
[HGLDD] Fix instance output port emission

### LINKED ISSUE #6735: [HGLDD] Don't access instance ports as hierarchical path
HGLDD emission currently emits instance output ports as `bar.a` hierarchical paths, instead of pointing to the wire that is attached to `bar`'s port `a`. Consider this example:
```mlir
hw.module @Foo(out x : i36, out y : i36) {
  %a, %b = hw.instance "bar" @Bar() -> (a: i36, b: i36)
  hw.output %a, %b : i36, i36
}
hw.module.extern @Bar(out a : i36, out b : i36) attributes {verilogName = "Bar"}
```
Running this through `firtool -g --emit-hgldd reduced15.mlir` produces the following problematic HGLDD JSON snippet:
```
"value": {"field":"a","var_ref":{"sig_name":"bar"}},
"value": {"field":"b","var_ref":{"sig_name":"bar"}},
```
In this case, the `{"sig_name":"bar"}` is trying to refer to the instance of `Bar`, which is invalid. Instead, EmitHGLDD should do the following:
- Also look at the `hw.output` op to see if an instance output is directly used as a module's output, and use that output's name in the HGLDD. We already do this for local wires.
- Never emit the `{"field":"a","var_ref":{"sig_name":"bar"}}` style of hierarchical path access into instances. Instead, just drop the instance port if it cannot be materialized.

fabianschuiki:
The `EmitHGLDD` translation currently emits instance output ports as XMR `<instName>.<portName>`. This is not valid HGLDD. Remove this special handling of instances. Instead if a value cannot be emitted as HGLDD, try to follow the value's use as input to a wire or assignment and use the destination name instead.

Also fix an issue where ops with a single result and `hw.verilogName` attribute would be emitted as signal references with that Verilog name. This code path would trigger for instances with a single output port, causing the instance name to be used for the signal reference. Limit that code path to only trigger on Comb dialect ops.

Fixes #6735.

### DIFF
--- a/lib/Target/DebugInfo/EmitHGLDD.cpp
+++ b/lib/Target/DebugInfo/EmitHGLDD.cpp
@@ -641,43 +641,13 @@ EmittedExpr FileEmitter::emitExpression(Value value) {
   auto result = cast<OpResult>(value);
   auto *op = result.getOwner();
 
-  // If the operation has only this one result and is named in some form, use
-  // that name.
-  if (op->getNumResults() == 1) {
-    // If a `hw.verilogName` is available, emit the value as just a reference to
-    // that name.
-    if (auto name = op->getAttrOfType<StringAttr>("hw.verilogName");
-        name && !name.empty())
+  // If the operation is a named signal in the output Verilog, use that name.
+  if (isa<hw::WireOp, sv::WireOp, sv::RegOp, sv::LogicOp>(op)) {
+    auto name = op->getAttrOfType<StringAttr>("hw.verilogName");
+    if (!name || name.empty())
+      name = op->getAttrOfType<StringAttr>("name");
+    if (name && !name.empty())
       return {hglddSigName(name), result.getType()};
-
-    // Use the "name" attribute of certain Verilog-visible ops directly.
-    if (auto name = op->getAttrOfType<StringAttr>("name");
-        name && !name.empty() &&
-        isa<hw::WireOp, sv::WireOp, sv::RegOp, sv::LogicOp>(op))
-      return {hglddSigName(name), result.getType()};
-  }
-
-  // Emit references to instance ports as `<instName>.<portName>`.
-  if (auto instOp = dyn_cast<hw::InstanceOp>(op)) {
-    auto instName = instOp->getAttrOfType<StringAttr>("hw.verilogName");
-    if (!instName)
-      instName = instOp.getInstanceNameAttr();
-    if (!instName)
-      return {};
-    auto *moduleOp =
-        symbolCache->getDefinition(instOp.getReferencedModuleNameAttr());
-    auto portName =
-        cast<hw::HWModuleLike>(moduleOp)
-            .getPort(instOp.getPortIdForOutputId(result.getResultNumber()))
-            .getVerilogName();
-    if (portName.empty())
-      return {};
-    auto inner = hglddSigName(instName);
-    return {JObject({
-                {"var_ref", std::move(inner)},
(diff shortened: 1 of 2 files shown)


lib/Target/DebugInfo/EmitHGLDD.cpp
```diff
@@ -641,45 +641,22 @@ EmittedExpr FileEmitter::emitExpression(Value value) {
...
-  // that name.
-  if (op->getNumResults() == 1) {
-    // If a `hw.verilogName` is available, emit the value as just a reference to
-    // that name.
+  // If this is a combinational operation, and it produces a single result, and
+  // it has a `hw.verilogName` associated with it, use that name.
+  if (op->getNumResults() == 1 && isa<comb::CombDialect>(op->getDialect()))
     if (auto name = op->getAttrOfType<StringAttr>("hw.verilogName");
         name && !name.empty())
       return {hglddSigName(name), result.getType()};
```
uenoku, line 649: This might be dead code. IIRC comb operations don't get hw.verilogName attributes.

lib/Target/DebugInfo/EmitHGLDD.cpp
```diff
@@ -904,6 +881,30 @@ EmittedExpr FileEmitter::emitExpression(Value value) {
...
+    // Use name of `sv.assign` destination that is assigned this value.
+    if (auto assignOp = dyn_cast<sv::AssignOp>(user))
+      if (assignOp.getSrc() == result)
+        return emitExpression(assignOp.getDest());
+    // Use module output port name that carries this value.
+    if (isa<hw::OutputOp>(user)) {
+      auto mod = cast<hw::HWModuleLike>(user->getParentOp());
+      auto modType = mod.getHWModuleType();
+      auto portName = modType.getOutputName(use.getOperandNumber());
+      return {hglddSigName(portName), result.getType()};
```
uenoku, line 904: I think we need PortInfo::getVerilogName to get a correct port name.

lib/Target/DebugInfo/EmitHGLDD.cpp
```diff
@@ -904,6 +881,30 @@ EmittedExpr FileEmitter::emitExpression(Value value) {
             muxOp.getType()};
   }
 
+  // As a last resort, look for any named wire-like ops this value feeds into.
+  // This is useful for instance output ports for example since we cannot access
+  // instance ports as `<instName>.<portName>` in HGLDD. Instead, we look for a
+  // wire hooked up to the instance output, which is very likely to be present
+  // after Verilog emission.
```
uenoku, line 888: Does this need a condition `hasSingleUse()`? I think it might be unpredictable to rely on the order of uses.

uenoku (approved):
LGTM.

lib/Target/DebugInfo/EmitHGLDD.cpp
fabianschuiki: We do want this for instance ports with multiple uses. Otherwise the instance ports would disappear in the HGLDD output. I think the order of uses is deterministic, but somewhat hard to predict. Since the name we dig up here is not user-facing, we should be fine if the name is a somewhat random but deterministic pick among the available options.

lib/Target/DebugInfo/EmitHGLDD.cpp
fabianschuiki: Great point!

PR #6751, opened 2024-02-26
[MSFT][Emit] Replace `output_file` with an `emit::File`

nandor:
`output_file` is to be replaced with explicit files constructed via the emit dialect. This PR replaces its use in the MSFT dialect.
Behaviour is slighly altered: the verbatim is always emitted to a file as we should not mix in TCL with SV and it is omitted from the default file list as it should always point to SV.

### DIFF
--- a/include/circt/Dialect/MSFT/MSFTPasses.td
+++ b/include/circt/Dialect/MSFT/MSFTPasses.td
@@ -11,7 +11,11 @@ include "mlir/Pass/PassBase.td"
 def ExportTcl: Pass<"msft-export-tcl", "mlir::ModuleOp"> {
   let summary = "Create tcl ops";
   let constructor = "circt::msft::createExportTclPass()";
-  let dependentDialects = ["circt::sv::SVDialect", "circt::hw::HWDialect"];
+  let dependentDialects = [
+    "circt::emit::EmitDialect",
+    "circt::hw::HWDialect",
+    "circt::sv::SVDialect",
+  ];
   let options = [
     ListOption<"tops", "tops", "std::string",
                "List of top modules to export Tcl for",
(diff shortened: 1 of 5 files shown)


include/circt/Dialect/MSFT/MSFTPasses.td
```diff
@@ -11,7 +11,11 @@ include "mlir/Pass/PassBase.td"
 def ExportTcl: Pass<"msft-export-tcl", "mlir::ModuleOp"> {
   let summary = "Create tcl ops";
   let constructor = "circt::msft::createExportTclPass()";
-  let dependentDialects = ["circt::sv::SVDialect", "circt::hw::HWDialect"];
+  let dependentDialects = [
+    "circt::emit::EmitDialect",
+    "circt::hw::HWDialect",
```
dtzSiFive, line 16: Thanks for maintaining this!

lib/Dialect/MSFT/ExportQuartusTcl.cpp
```diff
@@ -332,17 +333,12 @@ LogicalResult TclEmitter::emit(Operation *hwMod, StringRef outputFile) {
...
-
-  // When requested, give the verbatim op an output file.
-  if (!outputFile.empty()) {
-    auto outputFileAttr =
-        OutputFileAttr::getFromFilename(builder.getContext(), outputFile);
-    verbatim->setAttr("output_file", outputFileAttr);
-  }
+  auto builder = ImplicitLocOpBuilder::atBlockEnd(
+      UnknownLoc::get(hwMod->getContext()), hwMod->getBlock());
+  builder.create<emit::FileOp>(outputFile, [&] {
```
dtzSiFive, line 338: Probably this is correct, but does the empty string here do what's intended?

dtzSiFive (approved):
LGTM

lib/Dialect/MSFT/ExportQuartusTcl.cpp
nandor: If it's okay with the MSFT folks, I'll add a check to bail out if the filename is empty.

PR #6753, opened 2024-02-26
[HGLDD] Uniquify object names during emission

### LINKED ISSUE #6749: [HGLDD] Uniquify object names during emission
Make HGLDD emission add some form of uniquification suffix to objects with duplicate names. Right now, if there are two debug variables with the same name, the emitter will happily include both with the same name. This is likely to crash in the tool that ingests the debug information (some may materialize the debug variables as Verilog XMRs or similar kinds of "views").

Also prevent empty names from being printed.

fabianschuiki:
Ensure the object and variable names emitted by the `EmitHGLDD` translation are unique and don't collide with any known Verilog keywords.

Refactor the emission code by pulling most of the shared emission state into a `GlobalState` struct that is created by the `Emitter` and is shared by each `FileEmitter`.

Move the `objectNamespace` that is used to uniquify object names into the new global emitter state, and use `sv::legalizeName` to uniquify module names instead of `Namespace`. This avoids keyword collisions. Also add a per-module `moduleNamespace` that is prepopulated with the global `objectNamespace`, and use it to uniquify variable and instance names.

Fixes #6749.

### DIFF
--- a/lib/Target/DebugInfo/EmitHGLDD.cpp
+++ b/lib/Target/DebugInfo/EmitHGLDD.cpp
@@ -160,11 +160,50 @@ static bool makePathRelative(StringRef path, StringRef relativeTo,
   return outputIt == outputEnd;
 }
 
+/// Legalize the given `name` such that it only consists of valid identifier
+/// characters in Verilog and does not collide with any Verilog keyword, and
+/// uniquify the resulting name such that it does not collide with any of the
+/// names stored in the `nextGeneratedNameIDs` map.
+///
+/// The HGLDD output is likely to be ingested by tools that have been developed
+/// to debug Verilog code. These have the limitation of only supporting valid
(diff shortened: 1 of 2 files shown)


lib/Target/DebugInfo/EmitHGLDD.cpp
```diff
@@ -160,12 +160,41 @@ static bool makePathRelative(StringRef path, StringRef relativeTo,
   return outputIt == outputEnd;
 }
 
+static StringRef legalizeName(StringRef name, llvm::StringMap<size_t> &ns) {
+  return sv::legalizeName(name, ns, true);
```
uenoku, line 164: nit: Can you add a comment why it uses sv::legalizeName?

fabianschuiki: Great point, thanks. I have added a comment.

uenoku (approved):
LGTM!

PR #6754, opened 2024-02-27
[HW to BTOR2] Add support for initial values

Dobios:
Registers like compreg have a `powerOn` value. This PR simply adds support for that at the btor2 level, where  `powerOn` values generate an `init` statement.

This allows us to stop the model checker from arbitrarily picking values for registers at cycle 0, thus making verification possible.


lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -272,6 +272,21 @@ struct ConvertHWToBTOR2Pass
     return constlid;
   }
 
+  // Generates an init statement, which allows for the use of powerOnValue
+  // operands in compreg registers
+  void genInit(Operation *reg, Value initVal, int64_t width) {
+    // Retrieve the various identifies we require for this
```
fzi-hielscher, line 278: Nit:
```suggestion
    // Retrieve the various identifiers we require for this
```

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -474,10 +489,12 @@ struct ConvertHWToBTOR2Pass
       width = hw::getBitWidth(reg.getType());
       next = reg.getInput();
       resetVal = reg.getResetValue();
+
```
fzi-hielscher, line 492: Nit: Blank lines before `else` are rather unusual style, but apparently clang-format does not mind...

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -474,10 +489,12 @@ struct ConvertHWToBTOR2Pass
...
     } else if (auto reg = dyn_cast<seq::FirRegOp>(op)) {
       width = hw::getBitWidth(reg.getType());
       next = reg.getNext();
       resetVal = reg.getResetValue();
+
```
fzi-hielscher, line 497: As above

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -778,8 +800,27 @@ struct ConvertHWToBTOR2Pass
...
 
-    // Generate state instruction (represents the register declaration)
-    genState(reg, w, regName);
+    // Check for initial values which must be emitted before the state in btor2
+    Value pov = reg.getPowerOnValue();
+    if (pov) {
+      // Start by visiting the powerOnValue, we make the assumption that it will
+      // always be a constant and thus no operands need to be added to the
+      // worklist
+      dispatchTypeOpVisitor(pov.getDefiningOp());
```
fzi-hielscher, line 809: Why assume when it can easily be checked? `llvm::isa_and_nonull<hw::ConstantOp>` should do the trick and ideally produce a diagnostic message if it's unsupported.

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -778,8 +800,27 @@ struct ConvertHWToBTOR2Pass
...
+
+      // Add it to the list of visited operations
+      handledOps.insert(pov.getDefiningOp());
+
+      // Generate state instruction (represents the register declaration)
+      genState(reg, w, regName);
+
+      // Finally generate the init statement
+      genInit((Operation *)reg, pov, w);
```
fzi-hielscher, line 818: No need to do an explicit (unsafe) cast here.

fzi-hielscher (commented):
Thank you for moving this to its own PR.

I'm concerned by the liberal use of `getDefiningOp` without checking if the result is null, i.e. if the value is a BlockArgument / module input. Not just in this PR but in this pass in general. E.g., having a `WireOp`directly on an input port will cause trouble, as it will invoke `setOpAlias` with a `nullptr` argument. And there are probably other scenarios which can produce incorrect outputs. But I think this will need to be fixed in a new PR. For now, maybe add assertions to `setOpLid`and `setOpAlias` to make it at least fail loudly?

Dobios:
Thanks for the review! You're right about the wires of block arguments, I didn't consider that case. I added a loud failure for now and will add proper support for it when I find the time.

fzi-hielscher (approved):

PR #6757, opened 2024-02-27
[HW] Remove the file list attribute from HW

nandor:
The file list operations from the `emit` dialect should be used instead


dtzSiFive (approved):
LGTM

fabianschuiki (approved):
Awesome 🥳

PR #6759, opened 2024-02-27
[ESI][Runtime] Building wheels

teqdruid:
Adds the ability to create Python wheels for the ESI runtime.

### DIFF
--- a/lib/Dialect/ESI/runtime/pyproject.toml
+++ b/lib/Dialect/ESI/runtime/pyproject.toml
@@ -26,6 +26,3 @@ description = "ESI accelerators runtime"
 authors = [{ name = "John Demme", email = "John.Demme@microsoft.com" }]
 dependencies = []
 requires-python = ">=3.8"
-
-# [project.urls]
-# "Homepage" = "https://circt.llvm.org/"
--- a/lib/Dialect/ESI/runtime/pyproject.toml
+++ b/lib/Dialect/ESI/runtime/pyproject.toml
@@ -1,0 +1,31 @@
+[build-system]
+requires = [
+  "setuptools>=68",
+  "setuptools_scm>=8.0",
+  "wheel",
+  "cmake>=3.12",
+  "pybind11>=2.9",
+]
+build-backend = "setuptools.build_meta"
+
+# Enable version inference from Git.
+[tool.setuptools_scm]
+root = "../../../.."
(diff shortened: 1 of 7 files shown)


lib/Dialect/ESI/runtime/pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+[project]
+name = "esiaccel"
+dynamic = ["version"]
+description = "ESI accelerators runtime"
+authors = [{ name = "John Demme", email = "John.Demme@microsoft.com" }]
+dependencies = []
+requires-python = ">=3.8"
+
+# [project.urls]
+# "Homepage" = "https://circt.llvm.org/"
```
mikeurbach, line 31: nit: uncomment or remove? or is this placeholder for something else?

lib/Dialect/ESI/runtime/setup.py
```diff
@@ -0,0 +1,124 @@
+# Build/install the pycde python package. Blatantly copied from npcomp.
```
mikeurbach, line 1: nit: this comment should maybe be updated to mention "ESI Runtime python package" or something. npcomp is a real throwback!

lib/Dialect/ESI/runtime/setup.py
```diff
@@ -0,0 +1,124 @@
...
+    if cc is not None:
+      cmake_args.append("-DCMAKE_C_COMPILER={}".format(cc))
+
+    if "CIRCT_EXTRA_CMAKE_ARGS" in os.environ:
+      cmake_args += os.environ["CIRCT_EXTRA_CMAKE_ARGS"].split(" ")
+
+    subprocess.check_call(["cmake", src_dir] + cmake_args, cwd=cmake_build_dir)
+
+    # Run the build.
+    subprocess.check_call([
```
mikeurbach, line 81: with the benefit of hindsight, is there any reason to run the `--build` step? if we just go straight to `--install`, shouldn't it build the relevant targets? I could be wrong, since i don't usually run cmake directly.

mikeurbach (approved):
Some minor comments, but I think the setup looks good. As usual, I'm not a cmake expert...

lib/Dialect/ESI/runtime/setup.py
teqdruid: It does not build the relevant targets. Fuck cmake.

PR #6760, opened 2024-02-28
Support: Edge-Filtered Graph

dtzSiFive:
Alter InstanceGraph to expose the edges and what it means to get the destination of an edge.

Define an edge-filtered graph and GraphTraits that exposes the underlying graph but dropping edges as indicated by the filter predicate.

Reaching through the filtered nodes require either `.node` or using `operator*`.

Drop a use of this filtered graph based on the InstanceGraph into GrandCentral.cpp as a small example, but with no edge filtering enabled.

---

Add helpers for df_iterator usage.

df_iterator's default storage uses SmallPtrSet which requires the
graph being iterator over has NodeRef's that are PointerLike.

This doesn't work for the edge-filtering graph,
so create an appropriate storage class (including dummy "complete" method),
and a convenience method for create the df range using it.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
@@ -24,6 +24,7 @@
 #include "circt/Dialect/HW/InnerSymbolNamespace.h"
 #include "circt/Dialect/SV/SVOps.h"
 #include "circt/Support/Debug.h"
+#include "circt/Support/FilteredGraph.h"
 #include "mlir/IR/ImplicitLocOpBuilder.h"
 #include "llvm/ADT/DepthFirstIterator.h"
 #include "llvm/ADT/TypeSwitch.h"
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -1720,15 +1721,18 @@ void GrandCentralPass::runOnOperation() {
...
+      make_filtered_graph(instancePaths->instanceGraph.lookup(effectiveDUT),
+                          [](auto edge) { return true; });
+  auto dfRange = make_df_range(fGraph);
   for (auto i = dfRange.begin(), e = dfRange.end(); i != e;) {
-    auto module = cast<FModuleLike>(*i->getModule());
+    auto module = cast<FModuleLike>(*(**i)->getModule());
     if (AnnotationSet(module).hasAnnotation(companionAnnoClass)) {
       i.skipChildren();
       continue;
     }
```
seldridge, line 1734: For completeness, does it make sense to also have node more integrated node filtering? What I didn't like a lot about this code was the need for the `i.skipChildren()` here. However, maybe that's fine?

dtzSiFive: Totally, node filtering would be a good thing to support while at it and should be a short hop-skip given the rest -- although how this works re:iterating all nodes `node_begin/node_end` (not defined for the FilteredGraph at all presently) is less clear.
I was thinking if this was useful the NodeRef/EdgeRef maybe should point to the "FilteredGraph" itself which could contain the filters for each or whatever makes sense.  Bit more indirection, but also makes the comparison operators more natural, was the thinking.

Offhand I think a similar effect could be had by filtering out edges if the edge's destination (here, `getTarget()`) has the companion annotation?  WDYT?

PR #6761, opened 2024-02-28
[WireDFT] Remove WireDFT

nandor:
WireDFT is no longer used within SiFive. No issues popped up while the pass could be optionally enabled via a flag. It is reasonable to proceed with its removal, internally we agreed that we can pay off some tech debt at the cost of having to branch off CIRCT main to support internal releases which still depend on WireDFT.

Aiming to merge this PR into the 1.68.0 release.


uenoku (commented):
Please remove dftTestModeEnableAnnoClass and dftClockDividerBypassAnnoClass from AnnotationDetails.h as well.

fabianschuiki (approved):
Awesome work! :partying_face: It's mind-boggling that this accounts for about 1000 lines of code :100:

lib/Dialect/FIRRTL/FIRRTLAnnotations.cpp
```diff
@@ -694,9 +694,8 @@ LogicalResult circt::firrtl::extractDUT(const FModuleOp mod, FModuleOp &dut) {
   if (!AnnotationSet(mod).hasAnnotation(dutAnnoClass))
     return success();
 
-  // TODO: This check is duplicated multiple places, e.g., in
-  // WireDFT.  This should be factored out as part of the annotation
-  // lowering pass.
+  // TODO: This check is duplicated multiple places. This should be factored
+  // out as part of the annotation lowering pass.
```
seldridge, line 698: Nice comment cleanup.

seldridge (approved):
Clean removal. LGTM

PR #6762, opened 2024-02-28
[Emit] Introduce `emit.ref` to pull ops into file bodies

nandor:
This PR introduces a minimal implementation of file-to-operation references. Presently, only modules can be referenced and each module can be referenced by a single file once. As an example, the `HWMemSimImpl` pass is ported over.

Subsequent PRs will rely on this mechanism to phase out the `output_file` attribute.

### DIFF
--- a/include/circt/Dialect/Emit/EmitOpInterfaces.td
+++ b/include/circt/Dialect/Emit/EmitOpInterfaces.td
@@ -1,0 +1,22 @@
+//===- EmitOpInterfaces.td - Operation Interfaces ----------*- tablegen -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This describes the Emit operation interfaces and traits.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_DIALECT_EMIT_EMITOPINTERFACES
+#define CIRCT_DIALECT_EMIT_EMITOPINTERFACES
+
+include "mlir/IR/OpBase.td"
+
+def Emittable : NativeOpTrait<"Emittable"> {
+  let cppNamespace = "::circt::emit";
+}
+
+#endif // CIRCT_DIALECT_EMIT_EMITOPINTERFACES
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -6043,15 +6043,17 @@ public:
   void emit(emit::FileListOp op);
 
 private:
+  void emitOp(emit::RefOp op);
   void emitOp(emit::VerbatimOp op);
 };
 
 void FileEmitter::emit(emit::FileOp op) {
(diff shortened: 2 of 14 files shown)


lib/Dialect/Emit/EmitOps.cpp
```diff
@@ -47,6 +47,18 @@ void FileOp::build(OpBuilder &builder, OperationState &result,
     bodyCtor();
 }
 
+//===----------------------------------------------------------------------===//
+// RefOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult RefOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  auto target = getTargetAttr();
+  auto *op = symbolTable.lookupNearestSymbolFrom(getOperation(), target);
```
uenoku, line 56: Can we reject unsupported op types in the verifier?

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6079,6 +6081,19 @@ void FileEmitter::emit(emit::FileListOp op) {
...
+void FileEmitter::emitOp(emit::RefOp op) {
+  StringAttr target = op.getTargetAttr().getAttr();
+  auto *targetOp = state.symbolCache.getDefinition(target);
+
+  TypeSwitch<Operation *>(targetOp)
+      .Case<hw::HWModuleOp>(
+          [&](auto module) { ModuleEmitter(state).emitHWModule(module); })
+      .Default([&](auto op) {
+        op->emitError("cannot be emitted to a file");
+        state.encounteredError = true;
```
uenoku, line 6093: nit: EmitterBase::emitOpError() would set the error flag automatically

lib/Dialect/Emit/EmitOps.cpp
nandor: The `emit` dialect can't link to the other dialects to reference the ops for `isa` AFAIK. We should introduce a trait that makes ops emittable, but that is a bit overkill at the moment.

nandor: On second thought, I added the interface.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6076,6 +6078,17 @@ void FileEmitter::emit(emit::FileListOp op) {
   ps.eof();
 }
 
+void FileEmitter::emitOp(emit::RefOp op) {
+  StringAttr target = op.getTargetAttr().getAttr();
+  auto *targetOp = state.symbolCache.getDefinition(target);
```
uenoku, line 6083: Please add an assertion that targetOp is emittable.

include/circt/Dialect/Emit/EmitOpInterfaces.td
```diff
@@ -0,0 +1,23 @@
...
+// This describes the Seq operation interfaces.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_DIALECT_EMIT_EMITOPINTERFACES
+#define CIRCT_DIALECT_EMIT_EMITOPINTERFACES
+
+include "mlir/IR/OpBase.td"
+
+def Emittable : OpInterface<"Emittable"> {
```
uenoku, line 18: Are you going to use to create methods to the interface? If not, maybe trait is better fit for this?

uenoku (approved):
One question regarding interface vs trait but otherwise LGTM

PR #6763, opened 2024-02-28
[ESI][Runtime] Build and publish wheels

teqdruid:
Make the publishing pipeline actually work to produce wheels for multiple python versions.


mikeurbach (approved):
The workflow yaml looks good to me.

teqdruid:
Thanks for the help!

PR #6765, opened 2024-02-28
[Calyx] Switch sequential memories to be true single port memories

andrewb1999:
Currently, Calyx seq memories have read enable/done and write enable/done signals. This gives the wrong impression that you can perform simultaneous reads and writes, like how you could with a simple dual port memory, but the intention is for seq memories to be single port (as they have a single set of addresses). This PR changes seq mems to have a single content enable/done interface, with a write enable signal that is not a go signal. The intended way to perform a read is the set content enable high and write enable low (and wait for content done to go high). For a write we do the same thing but set the write enable high. 

This interface more closely matches how real hardware designers implement single port memories.

These changes need to be merged simultaneously with a new Calyx release with matching changes in the native compiler. See PR with native changes here: https://github.com/calyxir/calyx/pull/1610


mikeurbach (approved):
Changing this in lockstep with the native Calyx compiler makes sense to me. I think the implementation looks great.

cgyurgyik (approved):

rachitnigam:
Awesome! I think we should merge this since `main` on Calyx already has the required changes. I think after this release of the Calyx compiler, we should discuss a better approach towards maintaining both of these toolchains together and lock-stepping release cycles.

mikeurbach:
👍 sounds good!

Issue #6767, opened 2024-02-29
[Comb][Canonicalize] Missing best effort dialect attribute propagation


7FM:
Similar to my previous issue https://github.com/llvm/circt/issues/5531, several canonicalizations drop attributes (except `sv.namehint`). While I agree that it is unsafe to propagate discardable attributes freely, I think that the current implementation might be slightly to conservative.
As @teqdruid pointed out in https://github.com/llvm/circt/pull/5532#discussion_r1251266106:
> In general, dialect attributes should be propagated best effort, so I think it's appropriate to propagated them through the canonicalizers where it's obvious what the destination/source op is.

IMHO one such example would be when the canonicalization rewrites the operation without changing the dialect operation class.
Here is an example for the canonicalize pattern [`extract(olo, extract(ilo, x)) = extract(olo + ilo, x)`](https://github.com/llvm/circt/blob/c4038382fbbaea7b7b1abebaf4ba769d51609b11/lib/Dialect/Comb/CombFolds.cpp#L632):
```
module {
  hw.module @attrsGone(in %arg0: i64, out res: i32) {
    %0 = comb.extract %arg0 from 0 : (i64) -> i42
    %1 = comb.extract %0 from 0 {custom_dialect.attr = "test"} : (i42) -> i32
    hw.output %1 : i32
  }
}
```
Current result after `-canonicalize`:
```
module {
  hw.module @attrsGone(in %arg0 : i64, out res : i32) {
    %0 = comb.extract %arg0 from 0 : (i64) -> i32
    hw.output %0 : i32
  }
}
```
With a best effort propagation:
```
module {
  hw.module @attrsGone(in %arg0 : i64, out res : i32) {
    %0 = comb.extract %arg0 from 0 {custom_dialect.attr = "test"} : (i64) -> i32
    hw.output %0 : i32
  }
}
```

darthscsi:
No, MLIR doesn't provide mechanisms to safely propagate unknown attribute.
https://discourse.llvm.org/t/on-querying-an-operations-intrinsic-core-vs-external-user-defined-attributes/4498/5

PR #6768, opened 2024-02-29
[RFC] [Comb][Canonicalize] Best effort dialect attribute propagation

7FM:
Addresses https://github.com/llvm/circt/issues/6767.
This is a first draft how a less conservative dialect attribute propagation could be implemented.

I checked the uses of `replaceOpWithNewOpAndCopyName` and found one candidate where I would personally not propagate the dialect attributes: https://github.com/llvm/circt/blob/c4038382fbbaea7b7b1abebaf4ba769d51609b11/lib/Dialect/Comb/CombFolds.cpp#L1685-L1689
It should be possible to opt-out of the implicit attribute propagation by explicitly using `Operation*` as second template parameter:
```
replaceOpWithNewOpAndCopyName<MulOp,  Operation*>(rewriter, op, op.getType(),
                                                  ArrayRef<Value>(shlOp), false);
```
Alternatively, the implementation may be changed to add a boolean template parameter to explicitly opt-in/out of the attribute propagation.

### DIFF
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -84,15 +84,21 @@ static void replaceOpAndCopyName(PatternRewriter &rewriter, Operation *op,
 /// A wrapper of `PatternRewriter::replaceOpWithNewOp` to propagate
 /// "sv.namehint" attribute. If a replaced op has a "sv.namehint" attribute,
 /// this function propagates the name to the new value.
-template <typename OpTy, typename... Args>
-static OpTy replaceOpWithNewOpAndCopyName(PatternRewriter &rewriter,
-                                          Operation *op, Args &&...args) {
-  auto name = op->getAttrOfType<StringAttr>("sv.namehint");
+template <typename NewOpTy, typename OldOpTy, typename... Args>
+static NewOpTy replaceOpWithNewOpAndCopyName(PatternRewriter &rewriter,
+                                             OldOpTy op, Args &&...args) {
   auto newOp =
-      rewriter.replaceOpWithNewOp<OpTy>(op, std::forward<Args>(args)...);
-  if (name && !newOp->hasAttr("sv.namehint"))
-    rewriter.modifyOpInPlace(newOp,
-                             [&] { newOp->setAttr("sv.namehint", name); });
+      rewriter.replaceOpWithNewOp<NewOpTy>(op, std::forward<Args>(args)...);
+
+  if constexpr (std::is_same<NewOpTy, OldOpTy>::value) {
+    rewriter.modifyOpInPlace(
+        newOp, [&] { newOp->setDialectAttrs(op->getDialectAttrs()); });
+  } else {
+    auto name = op->template getAttrOfType<StringAttr>("sv.namehint");
+    if (name && !newOp->hasAttr("sv.namehint"))
+      rewriter.modifyOpInPlace(newOp,
+                               [&] { newOp->setAttr("sv.namehint", name); });
+  }
 
   return newOp;
 }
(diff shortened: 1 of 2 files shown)


dtzSiFive:
How do we know the dialect attributes' intended semantics are valid across the transformation?

Looks like in this PR the proposed answer is "if it's the same operation name, it's probably fine"?

I appreciate that dropping them may be problematic but I also really don't think we should be preserving attributes we don't know are safe to be preserved.  While we could spin up some sort of, e.g., dialect interface to answer "should this be preserved" offhand I'm not sure that's reasonable to do generically -- "does my transformation preserve your attribute meaning?" and as such only makes sense for the subset of dialect attributes that question can be simply answered for "yes if operations are the same".

If you're folding two operations, do you merge the attributes? Which do you take? So on.

Looking through the linked discussion, I disagree that we're obligated to use heuristics to propagate these -- there's a reason the MLIR core infrastructure doesn't do this.

Is there a solution that helps your use case specifically without encoding this sort of assumption about the nature/purpose/meaning of dialect attributes generally?

uenoku:
I agree with Will's points. This has been discussed on MLIR forums several times and I think basically we cannot safely propagate unknown attributes.  https://discourse.llvm.org/t/canonicalization-passes-dont-keep-attributes/59750/2?u=uenoku

One thing you could try is to attach an operation to retain your custom attributes before canonicalizations and pull back after canonicalizations, e.g:
```
  hw.module @attrsGone(in %arg0: i64, out res: i32) {
    %0 = comb.extract %arg0 from 0 : (i64) -> i42
    %1 = comb.extract %0 from 0 : (i42) -> i32
    some_dialect.keep_attribute %1 {custom_dialect.attr = "test"}
    hw.output %1 : i32
  }
=> 
  hw.module @attrsGone(in %arg0: i64, out res: i32) {
    %0 = comb.extract %arg0 from 0 : (i64) -> i32
    some_dialect.keep_attribute %0 {custom_dialect.attr = "test"}
    hw.output %0 : i32
  }
```

This approach is quite similar to what Debug dialect is preserving original signal names. 

Or another approach might be to introduce a dialect interface to create an allow-list for attributes that can be propagated, and users can dynamically configure them while the dialect registration.

7FM:
> One thing you could try is to attach an operation to retain your custom attributes before canonicalizations and pull back after canonicalizations, e.g:
> 
> ```
>   hw.module @attrsGone(in %arg0: i64, out res: i32) {
>     %0 = comb.extract %arg0 from 0 : (i64) -> i42
>     %1 = comb.extract %0 from 0 : (i42) -> i32
>     some_dialect.keep_attribute %1 {custom_dialect.attr = "test"}
>     hw.output %1 : i32
>   }
> => 
>   hw.module @attrsGone(in %arg0: i64, out res: i32) {
>     %0 = comb.extract %arg0 from 0 : (i64) -> i32
>     some_dialect.keep_attribute %0 {custom_dialect.attr = "test"}
>     hw.output %0 : i32
>   }
> ```

Yes, I have thought about that solution as well and was wondering to which degree it would inhibit optimizations. Maybe this can be mitigated by adding canonicalizations to `some_dialect.keep_attributes` to remove operands when they are only used by `some_dialect.keep_attributes`.
That being said, it kind of feels wrong to define an operation that essentially acts like an attribute, but I haven't found a better solution either.

> This approach is quite similar to what Debug dialect is preserving original signal names.

Thanks for the pointer. I will take a look at it, specifically regarding the optimization problem.

uenoku:
> Thanks for the hint. I will take a look at it, specifically regarding optimizations.

Attaching operations would affect DCE, so you may need to run another DCE(probably you can use `-remove-dead-values` pass in MLIR upstream) after moving attributes back to operations.

darthscsi:
You cannot safely propagate unknown attributes.
https://discourse.llvm.org/t/on-querying-an-operations-intrinsic-core-vs-external-user-defined-attributes/4498/5

7FM:
Thanks for the feedback! I will close this PR and the corresponding issue for now.

PR #6769, opened 2024-02-29
[Calyx] Fix memory import locations

andrewb1999:
Updates memory import locations to be correct for the current calyx native compiler `main`.


cgyurgyik (approved):

mikeurbach (approved):
I haven't been keeping up, but makes sense to me.

rachitnigam:
LGTM!

Issue #6603, opened 2024-01-24
[FIRRTL] Dedup should update class reference ports if classes dedup.


mikeurbach:
In https://github.com/llvm/circt/pull/6582, initial support for classes and objects was added in Dedup.

However, this doesn't yet support updating class ports that are ClassType when the classes being referenced in those ports are deduped.

To see what I mean, in test/firtool/classes-dedupe.fir, you can adjust `class Foo_1` and `class Foo_2` so they dedup (just change their single output to be the same name in both classes). There are places where objects of those classes are used in ports, and after the classes/objects dedup, we try to propassign an object reference of the wrong type to the port.

I think this is kind of a new thing, since references to instances is not possible in FIRRTL, so we need to decide where and how to update Dedup to take this into consideration for classes/objects. At a minimum, this will require updating the block arg and portTypes of ClassLikes.

PR #6582, opened 2024-01-17
[FIRRTL] Initial support for classes and objects in Dedup.

mikeurbach:
This adds initial support for Dedup to handle deduping classes and objects. For the most part, this amounts to ensuring core functionality is implemented in terms of the FModuleLike and FInstanceLike interfaces, which classes and objects implement, respectively.

There are a couple places where some specific logic is added for objects, similar to the specific logic that was already there for instance ops.

Finally, a small change related to paths is needed in LowerClasses. There is some logic there that confirms a hierarchical path's root module is contained within the path's owning module. In the case of deduping hierarchical paths, an extra layer of hierarchy is added to hierpath ops, which breaks this logic. A new condition is added, and if the owning module is anywhere in the hierarchical path, it is used as the root module.

An end-to-end firtool test is added to demonstrate that classes and objects dedup, and the final paths are valid in both the local and hierarchical cases.

This initial support is capable of generating valid paths, but does not actually confirm that paths which dedup point to entities that dedup. Comments have been left, and a ticket opened to address this:
https://github.com/llvm/circt/issues/6583.

PR #6770, opened 2024-03-01
[FIRRTL] Handle reference ports when Classes dedup.

mikeurbach:
In https://github.com/llvm/circt/pull/6582, initial support for classes and objects was added in Dedup. However, since classes and objects are new constructs, not every possibility was handled in the initial implementation.

This specifically handles the case where references to objects are passed through class ports, and the class type of such objects has changed because their classes deduped.

In the final fixup pass through the instance graph, if we find a class, we check if it has any reference ports that need to be updated, and if so update them. When this happens, we also update an objects of the class to reflect the newly updated class type.

Fixes https://github.com/llvm/circt/issues/6603.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -1391,6 +1391,72 @@ private:
 // Fixup
 //===----------------------------------------------------------------------===//
 
+/// This fixes up ClassLikes with ClassType ports, when the classes have
+/// deduped. For each ClassType port, if the object reference being assigned is
+/// a different type, update the port type. Returns true if the ClassOp was
+/// updated and the associated ObjectOps should be updated.
+bool fixupClassOp(ClassOp classOp) {
+  // New port type attributes, if necessary.
+  SmallVector<Attribute> newPortTypes;
+  bool anyDifferences = false;
+
+  // Check each port.
+  for (size_t i = 0, e = classOp.getNumPorts(); i < e; ++i) {
+    // Check if this port is a ClassType. If not, save the original type
+    // attribute in case we need to update port types.
+    auto portClassType = dyn_cast<ClassType>(classOp.getPortType(i));
+    if (!portClassType) {
+      newPortTypes.push_back(classOp.getPortTypeAttr(i));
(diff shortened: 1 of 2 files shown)


mikeurbach:
@uenoku since you pointed this out on the initial PR, do you mind taking a look? Or maybe @youngar as the resident dedup expert. Since the whole notion of references to instances is a new concept, I wasn't totally sure where to implement the change. It seemed appropriate in the final fixupAllModules walk of the instance graph.

uenoku (commented):
Looking great! I think for class op it correctly deduplicates ports. I'm curious if we need something similar for operations in module ops, especially wire ops? For example I expect the test still successfully dedups with the following patch. Probably it's outside of the scope of the PR so it would be fine to fix in a follow-up. 

```diff
diff --git a/test/firtool/classes-dedupe.fir b/test/firtool/classes-dedupe.fir
index 5b0bbe78a..b7fef164c 100644
--- a/test/firtool/classes-dedupe.fir
+++ b/test/firtool/classes-dedupe.fir
@@ -33,7 +33,9 @@ circuit Test : %[[
     output om_out : AnyRef
 
     object om of OM_1
-    propassign om_out, om
+    wire tmp: Inst<OM_1>
+    propassign tmp, om
+    propassign om_out, tmp
 
     ; CHECK: hw.instance "fetch_1" sym [[SYM2]]
     inst fetch_1 of Fetch_1
@@ -49,7 +51,9 @@ circuit Test : %[[
     output om_out : AnyRef
 
     object om of OM_2
-    propassign om_out, om
+    wire tmp: Inst<OM_2>
+    propassign tmp, om
+    propassign om_out, tmp
 
     inst fetch_1 of Fetch_2
     inst fetch_2 of Fetch_2
```

mikeurbach:
Yeah, that is a good point, this isn't handled in this PR. I think we can file a separate issue to handle that. It's legal IR, but we do have a very early pass that removes wires holding properties, since properties require SSA anyway, so by the time we get to dedup we shouldn't see IR like this in practice. It would be nice if this invariant was enforced, so later passes could assume there are no property wires.

uenoku (approved):
That makes sense, thanks! LGTM

PR #6771, opened 2024-03-01
[FIRRTL] Use the class map in ObjectOp parser.

mikeurbach:
We already store a mapping from class name to ClassOp when we parse the ClassOp signatures, in case the class name is referenced in any ClassOp port types. This re-uses the same data structure when parsing an ObjectOp to a) ensure the class exists, and b) construct the ObjectOp using the ClassOp.

The previous implementation would look up the ClassOp symbol in the circuit, but this would simply scan the circuit body again and again. This change makes a significant difference for performance, on the order of 90% overall parsing time reduction for large designs with many objects.


mikeurbach:
@rwy7 any concerns here? I think you've added the support for ObjectOp to the parser, and the classMap.

seldridge (approved):
LGTM

rwy7:
Looks good!

PR #6773, opened 2024-03-01
[ImportVerilog] Add assignment statements

fabianschuiki:
Add continuous, blocking, and non-blocking assignments to the Moore dialect. These represent the corresponding constructs in SystemVerilog.

Add a `ScopedHashTable` to the ImportVerilog conversion context and add variable declarations to it. This allows expression conversion to resolve name expressions to the MLIR value that was generated for the corresponding declaration. The Slang AST node pointer itself is used as a lookup key, since the AST already has all names resolved.

Add a basic expression handling mechanism to the ImportVerilog conversion. A new `convertExpression` entry point calls an `ExprVisitor` to convert expression nodes in the Slang AST to the corresponding MLIR operations. This currently only supports name expressions, which simply looks up the converted MLIR value for whatever the name was resolved to, and blocking and non-blocking assignments.

Extend the module conversion to properly handle continuous assignments at the module level.

Thanks @albertethon, @hailongSun2000, and @angelzzzzz for the help to get this started.


dtzSiFive (approved):
LGTM!! Wow!

hailongSun2000 (approved):

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -0,0 +1,74 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#include "ImportVerilogInternals.h"
+#include "slang/syntax/AllSyntax.h"
+
+using namespace circt;
+using namespace ImportVerilog;
+
+// NOLINTBEGIN(misc-no-recursion)
```
maerhart, line 15: Why is it fine to always do a recursive call to traverse expressions? Isn't this like using recursion to traverse the def-use chain?

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
```diff
@@ -40,6 +41,8 @@ struct Context {
 
   /// Convert a slang `SourceLocation` into an MLIR `Location`.
   Location convertLocation(slang::SourceLocation loc);
+  /// Covnert a slang `SourceRange` into an MLIR `Location`.
```
maerhart, line 44:
```suggestion
  /// Convert a slang `SourceRange` into an MLIR `Location`.
```

maerhart (approved):
Shouldn't the expression converter use a worklist where the visitor can push expressions onto instead of recursive calls? Or is there some reason why this is fine that I'm not aware of?

Otherwise, looks good to me! Nice to see all this progress 💯

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Good point. I'm not sure it actually is fine. Well, probably in the beginning, but this probably has to be come that worklist-based approach you were suggesting. Not sure how exactly that would work -- if you'd push all child expressions of a node onto the worklist and then check some hash map to get the actual child expressions.

If I'm not mistaken, Slang uses a recursive descent parser to construct the AST. That would mean any expression tree being visited here has been visited recursively while parsing already. That might indicate that parsing is going to fail first or at the same time as this lowering code.

I'm going to create an issue to turn this into a worklist once the remaining Slang integration PRs have landed.

PR #6774, opened 2024-03-02
[capi][python] Add Emit Dialect

seldridge:
Add the emit dialect to the C-API and to Python.  This is both missing and is necessary for downstream Python tooling to not break.


mikeurbach:
LGTM!

PR #6777, opened 2024-03-04
[SV] Use a symbol in macro identifiers

nandor:
Macro identifiers now use a proper symbol to handle references. The verifiers are not yet enabled as more intrusive pipeline changes are required, to be done in follow-ups.

### DIFF
--- a/lib/Dialect/SV/SVOps.cpp
+++ b/lib/Dialect/SV/SVOps.cpp
@@ -91,6 +91,18 @@ static Operation *lookupSymbolInNested(Operation *symbolTableOp,
   return nullptr;
 }
 
+/// Verifies symbols referenced by macro identifiers.
+static LogicalResult
+verifyMacroIdentSymbolUses(Operation *op, FlatSymbolRefAttr attr,
+                           SymbolTableCollection &symbolTable) {
+  auto *refOp = symbolTable.lookupNearestSymbolFrom(op, attr);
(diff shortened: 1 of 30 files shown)


lib/Dialect/SV/SVOps.cpp
```diff
@@ -91,6 +91,18 @@ static Operation *lookupSymbolInNested(Operation *symbolTableOp,
...
+verifyMacroIdentSymbolUses(Operation *op, FlatSymbolRefAttr attr,
+                           SymbolTableCollection &symbolTable) {
+  auto *refOp = symbolTable.lookupNearestSymbolFrom(op, attr);
+  if (!refOp)
+    return op->emitError("references an undefined symbol: ") << attr;
+  if (!isa<MacroDeclOp>(refOp))
+    return op->emitError("must reference a macro declaration");
+  return success();
+}
+
```
uenoku, line 105: Please feel free to land this change separately.

uenoku (approved):
LGTM. I think it's reasonable approach to make verifySymbolUses empty and actually implement a follow-up. Once verifySymbolUses is implemented, a serious of the changes will require downstream users to ensure uniqueness of the macro symbols so it would be nice to put a note on weekly meeting doc.

dtzSiFive:
Awesome work, thanks for pushing on this!

PR #6778, opened 2024-03-04
[Sim] Emit a SYNTHESIS macro declaration if needed

nandor:


lib/Conversion/SimToSV/SimToSV.cpp
```diff
@@ -144,20 +163,34 @@ struct SimToSVPass : public LowerSimToSVBase<SimToSVPass> {
...
+      patterns.add<SimulatorStopLowering<sim::FatalOp, sv::FatalOp>>(context,
+                                                                     state);
+      return applyPartialConversion(module, target, std::move(patterns));
+    };
+
+    if (failed(mlir::failableParallelForEach(
+            context, circuit.getOps<hw::HWModuleOp>(), lowerModule)))
+      return signalPassFailure();
+
+    if (state.used_SYNTHESIS && !circuit.lookupSymbol("SYNTHESIS")) {
```
uenoku, line 189: Please use symbol table instead of Operation::lookupSymbol to avoid O(N^2) behaivor.

uenoku: Nvm, it's ran only once so don't need to use symbol table.

uenoku, line 189: Can we error out when circuit.lookupSymbol is not sv::MacroDeclOp?

uenoku (approved):
LGTM

PR #6780, opened 2024-03-04
[SV] Verify macro reference symbols

### LINKED ISSUE #6779: [SV] Emit macro declarations for all referenced macros
Operations which reference macros should verify that a corresponding decl exists. The pipeline must be updated to support this check.

nandor:
Macro references are now fully verified. Passes have been adjusted to emit the correct declarations.

Closes #6779

### DIFF
--- a/lib/Conversion/SeqToSV/SeqToSV.cpp
+++ b/lib/Conversion/SeqToSV/SeqToSV.cpp
@@ -456,48 +456,69 @@ void SeqToSVPass::runOnOperation() {
   if (failed(applyPartialConversion(circuit, target, std::move(patterns))))
     signalPassFailure();
 
+  bool needsMemRandomization = !memsByModule.empty();
+
+  auto loc = UnknownLoc::get(context);
+  auto b = ImplicitLocOpBuilder::atBlockBegin(loc, circuit.getBody());
+  if (needsRegRandomization || needsMemRandomization) {
+    b.create<sv::MacroDeclOp>("ENABLE_INITIAL_REG_");
+    b.create<sv::MacroDeclOp>("ENABLE_INITIAL_MEM_");
+    if (needsRegRandomization) {
+      b.create<sv::MacroDeclOp>("FIRRTL_BEFORE_INITIAL");
+      b.create<sv::MacroDeclOp>("FIRRTL_AFTER_INITIAL");
+    }
+    if (needsMemRandomization)
+      b.create<sv::MacroDeclOp>("RANDOMIZE_MEM_INIT");
+    b.create<sv::MacroDeclOp>("RANDOMIZE_REG_INIT");
+    b.create<sv::MacroDeclOp>("RANDOMIZE");
+    b.create<sv::MacroDeclOp>("RANDOMIZE_DELAY");
+    b.create<sv::MacroDeclOp>("RANDOM");
+    b.create<sv::MacroDeclOp>("INIT_RANDOM");
+    b.create<sv::MacroDeclOp>("INIT_RANDOM_PROLOG_");
+  }
+
   bool hasRegRandomization = needsRegRandomization && !disableRegRandomization;
-  bool hasMemRandomization = !memsByModule.empty() && !disableMemRandomization;
+  bool hasMemRandomization = needsMemRandomization && !disableMemRandomization;
   if (!hasRegRandomization && !hasMemRandomization)
     return;
 
   // Build macros for FIRRTL-style register and memory initialization.
   // Insert them at the start of the module, after any other verbatims.
-  auto loc = UnknownLoc::get(context);
-  auto b = ImplicitLocOpBuilder::atBlockBegin(loc, circuit.getBody());
   for (Operation &op : *circuit.getBody()) {
-    if (!isa<sv::VerbatimOp>(&op)) {
+    if (!isa<sv::VerbatimOp, sv::IfDefOp>(&op)) {
       b.setInsertionPoint(&op);
       break;
     }
   }
 
+  // Create SYNTHESIS/VERILATOR macros if other passes have not done so already.
+  {
+    StringSet<> symbols;
+    for (auto sym : circuit.getOps<sv::MacroDeclOp>())
+      symbols.insert(sym.getName());
+    if (!symbols.count("SYNTHESIS"))
+      b.create<sv::MacroDeclOp>("SYNTHESIS");
+    if (!symbols.count("VERILATOR"))
+      b.create<sv::MacroDeclOp>("VERILATOR");
+  }
+
   // TODO: We could have an operation for macros and uses of them, and
   // even turn them into symbols so we can DCE unused macro definitions.
-  StringSet<> emittedDecls;
-  auto emitDefine = [&](StringRef name, StringRef body, ArrayAttr args = {}) {
-    if (!emittedDecls.count(name)) {
-      emittedDecls.insert(name);
-      OpBuilder::InsertionGuard guard(b);
-      b.setInsertionPointToStart(circuit.getBody());
-      b.create<sv::MacroDeclOp>(name, args, StringAttr());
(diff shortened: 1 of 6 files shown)


lib/Conversion/SeqToSV/SeqToSV.cpp
```diff
@@ -456,48 +456,68 @@ void SeqToSVPass::runOnOperation() {
...
+  if (needsRegRandomization || needsMemRandomization) {
+    b.create<sv::MacroDeclOp>("ENABLE_INITIAL_REG_");
+    b.create<sv::MacroDeclOp>("ENABLE_INITIAL_MEM_");
+    if (needsRegRandomization) {
+      b.create<sv::MacroDeclOp>("FIRRTL_BEFORE_INITIAL");
+      b.create<sv::MacroDeclOp>("FIRRTL_AFTER_INITIAL");
+    }
+    if (needsMemRandomization) {
+      b.create<sv::MacroDeclOp>("RANDOMIZE_MEM_INIT");
+    }
```
uenoku, line 472:
```suggestion
    if (needsMemRandomization)
      b.create<sv::MacroDeclOp>("RANDOMIZE_MEM_INIT");
```

uenoku (approved):
LGTM

test/Dialect/Seq/firreg.mlir
```diff
@@ -1,5 +1,5 @@
 // RUN: circt-opt %s -verify-diagnostics --lower-seq-to-sv | FileCheck %s --check-prefixes=CHECK,COMMON
-// RUN: circt-opt %s -verify-diagnostics --pass-pipeline="builtin.module(lower-seq-to-sv{disable-reg-randomization})" | FileCheck %s --check-prefix COMMON --implicit-check-not RANDOMIZE_REG
+// RUN: circt-opt %s -verify-diagnostics --pass-pipeline="builtin.module(lower-seq-to-sv{disable-reg-randomization})" | FileCheck %s --check-prefix DISABLED
```
dtzSiFive, line 2: Consider:
```suggestion
// RUN: circt-opt %s -verify-diagnostics --pass-pipeline="builtin.module(lower-seq-to-sv{disable-reg-randomization})" | FileCheck %s --check-prefixes COMMON,DISABLED
```

To include the `-LABEL`'s?  This helps delimit the `-NOT`'s.

Nit:
As-is it looks like this has a sequence of (redundant) `-NOT` directives -- there's no positive check to separate them.
This could probably use `--implicit-check-not` to capture those directives but if having the explicit `DISABLED-NOT` inline serves a purpose that works too.

test/Dialect/Seq/hw-memsim.mlir
```diff
@@ -2,14 +2,17 @@
 // RUN: circt-opt -pass-pipeline="builtin.module(hw-memory-sim{read-enable-mode=ignore})" %s | FileCheck %s --check-prefixes=COMMON,IGNORE
 // RUN: circt-opt -pass-pipeline="builtin.module(hw-memory-sim{read-enable-mode=zero})" %s | FileCheck %s --check-prefixes=COMMON,ZERO
 // RUN: circt-opt -pass-pipeline="builtin.module(hw-memory-sim{add-mux-pragmas})" %s | FileCheck %s --check-prefixes=COMMON,PRAMGAS
-// RUN: circt-opt -pass-pipeline="builtin.module(hw-memory-sim{disable-mem-randomization})" %s | FileCheck %s --check-prefix COMMON --implicit-check-not RANDOMIZE_MEM
-// RUN: circt-opt -pass-pipeline="builtin.module(hw-memory-sim{disable-reg-randomization})" %s | FileCheck %s --check-prefix COMMON --implicit-check-not RANDOMIZE_REG
-// RUN: circt-opt -pass-pipeline="builtin.module(hw-memory-sim{disable-mem-randomization disable-reg-randomization})" %s | FileCheck %s --check-prefix COMMON --implicit-check-not RANDOMIZE_REG --implicit-check-not RANDOMIZE_MEM
+// RUN: circt-opt -pass-pipeline="builtin.module(hw-memory-sim{disable-mem-randomization})" %s | FileCheck %s --check-prefix COMMON --check-prefix NO-RAND-MEM
```
dtzSiFive, line 5: The `-NOT`'s here are applied to the entire output (only directive on each run) which makes sinking them to line 80 a bit unexpected -- that seems to suggest the checks start at that point or are relevant there.

What's the thinking here -- these FileCheck lines are doing too much, let's put them into a series of directives under one prefix each instead?

I think this change is functionally equivalent to putting the `sv.ifdef.procedural @RANDOMIZE_REG` in the FileCheck implicit-check-not -- is that right? But that makes an already long line longer.

Anyway other than the new prefixes' file-wide check-not being defined unexpectedly late, this is mostly preference/stylistic.

( Looks like `NO-RAND` could be `--check-prefixes NO-RAND-MEM,NO-RAND-REG` ? )

lib/Conversion/SeqToSV/SeqToSV.cpp
```diff
@@ -456,48 +456,68 @@ void SeqToSVPass::runOnOperation() {
...
   }
 
+  // Create SYNTHESIS/VERILATOR macros if other passes have not done so already.
+  StringSet<> symbols;
+  for (auto sym : circuit.getOps<sv::MacroDeclOp>())
+    symbols.insert(sym.getName());
+  if (!symbols.count("SYNTHESIS"))
+    b.create<sv::MacroDeclOp>("SYNTHESIS");
+  if (!symbols.count("VERILATOR"))
+    b.create<sv::MacroDeclOp>("VERILATOR");
```
dtzSiFive, line 502: Nit: consider inserting these as part of the check (`insert("..").second` -style, which hopefully StringSet supports or at least a bool return on insert) -- otherwise we leave a set around that looks like it has all active macros in it but might be missing two.

lib/Conversion/SeqToSV/SeqToSV.cpp
```diff
@@ -456,48 +456,68 @@ void SeqToSVPass::runOnOperation() {
   if (failed(applyPartialConversion(circuit, target, std::move(patterns))))
     signalPassFailure();
 
+  bool needsMemRandomization = !memsByModule.empty();
+
+  auto loc = UnknownLoc::get(context);
+  auto b = ImplicitLocOpBuilder::atBlockBegin(loc, circuit.getBody());
+  if (needsRegRandomization || needsMemRandomization) {
+    b.create<sv::MacroDeclOp>("ENABLE_INITIAL_REG_");
```
dtzSiFive, line 464: Why are these declared without checking if a previous pass declared them (like below)?  Are these /only/ for code here to rely on?

dtzSiFive: (and an error if any other pass declares them?)

dtzSiFive (approved):
Thanks for this verification and updating everything!

Grep'ing around for these macros I see other users -- e.g., `lib/Dialect/Seq/Transforms/HWMemSimImpl.cpp` -- should that also add declarations as needed?

Added some comments, please take a look :+1: .

lib/Conversion/SeqToSV/SeqToSV.cpp
nandor: Only one can be declared, so if they are redefined, we get a double-declaration error. I don't want to add that many more checks since I'd like to re-do this whole mechanism very soon.

lib/Conversion/SeqToSV/SeqToSV.cpp
nandor: I scoped this better

test/Dialect/Seq/firreg.mlir
nandor: I don't want to pass that much information on the command line, I think it's better to use `DISABLE-NOT` here.

PR #6781, opened 2024-03-04
[Seq] Allow presets for more types on firreg

nandor:


lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -449,8 +449,26 @@ ParseResult FirRegOp::parse(OpAsmParser &parser, OperationState &result) {
   result.addTypes({ty});
 
   if (presetValue) {
-    result.addAttribute("preset",
-                        parser.getBuilder().getIntegerAttr(ty, *presetValue));
+    uint64_t width = 0;
+    if (hw::type_isa<seq::ClockType>(ty)) {
+      width = 1;
+    } else {
+      int64_t maybeWidth = hw::getBitWidth(ty);
```
seldridge, line 456: Does this work for 64-bit unsigned types?

seldridge (commented):
Generally looks fine. Can you add tests that check behavior of registers that have 64-but widths (either scalars or aggregates)?

How should this work for aggregates that are larger than 64-bit?

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -449,8 +449,26 @@ ParseResult FirRegOp::parse(OpAsmParser &parser, OperationState &result) {
...
+      if (maybeWidth < 0)
+        return parser.emitError(loc, "cannot preset register of unknown width");
+      width = maybeWidth;
+    }
+
+    APInt presetResult = presetValue->sextOrTrunc(width);
+    if (presetValue->getBitWidth() > width) {
+      if (presetResult.sext(presetValue->getBitWidth()) != *presetValue)
+        return parser.emitError(loc, "integer value too large");
+    }
```
seldridge, line 466: I think this logic may not be correct given that this is rejecting both of:

``` mlir
  %c = seq.firreg %x clock %clock preset 3 : i2
  %d = seq.firreg %x clock %clock preset 2 : i2
```

With:

```
/Users/schuylereldridge/repos/github.com/llvm/circt/test/Dialect/Seq/round-trip.mlir:83:19: error: custom op 'seq.firreg' integer value too large
  %c = seq.firreg %x clock %clock preset 2 : i2
                  ^
/Users/schuylereldridge/repos/github.com/llvm/circt/test/Dialect/Seq/round-trip.mlir:83:19: error: custom op 'seq.firreg' integer value too large
  %d = seq.firreg %x clock %clock preset 3 : i2
                  ^
```

Can the PR add a test that presets that use all the available bits still work?

Can this test the same thing for a constant that is the full bit width of an aggregate?

Nit: the error message using `loc` is not placing the carat in the correct place. A user might be confused with the error message here thinking that `%x` was wrong.

test/Dialect/Seq/firreg-errors.mlir
```diff
@@ -4,3 +4,11 @@ hw.module @NeedsBothResetAndResetValue(in %input: i1, in %clk: !seq.clock) {
...
   "seq.firreg"(%input, %clk) { name = "reg", isAsync } : (i1, !seq.clock) -> i1
 }
+
+// -----
+
+// CHECK-LABEL: @preset_too_large
+hw.module @preset_too_large(in %clock: !seq.clock, in %reset: i1, in %next4: i4) {
+  // expected-error@below {{custom op 'seq.firreg' integer value too large}}
+  seq.firreg %next4 clock %clock preset 1024 : i4
+}
```
seldridge, line 14: Nit: missing newline.

seldridge (approved):
Generally looks good.

There seem to be some errors in the checking logic and in the location of error messages. Please fix those before landing.

PR #6783, opened 2024-03-05
[arcilator] Introduce integrated JIT for simulation execution

Moxinilian:
This PR adds a JIT runtime for arcilator, backed by MLIR's ExecutionEngine. This JIT allows executing `arc.sim` operations directly from the arcilator binary.


tools/arcilator/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ target_link_libraries(arcilator
   CIRCTSupport
   CIRCTTransforms
   MLIRBuiltinToLLVMIRTranslation
+  MLIRExecutionEngine
```
maerhart, line 18: I know that this PR is still a draft, but just wanted to make you aware that the MLIRExecutionEngine is only conditionally built IIRC when the host machine architecture is supported (and can also be manually disabled) which means we would need to make arcilator conditional on this as well. However, arcilator could also be useful to build the simulation binary on one machine and run it on another one (with potentially different architecture), so maybe we can always build arcilator but use some `#ifdef` in the code to only make the JIT part conditional? Not exactly sure what's the best approach here, maybe just guarding whole arcilator is good enough for a start.

Moxinilian:
Not really sure about what to do with respect to that short integration testing fail. This seems like an issue of its specific configuration but from a first inspection I do not see why the JIT runtime would not be built with it. Is it a cache issue perhaps?

maerhart:
> Not really sure about what to do with respect to that short integration testing fail. This seems like an issue of its specific configuration but from a first inspection I do not see why the JIT runtime would not be built with it. Is it a cache issue perhaps?

Not really sure, but did you try `add_compile_definitions(ARCILATOR_ENABLE_JIT=1)` (adding the `=1`)?

fabianschuiki:
The short integration test passes different options and definitions to its `cmake` call. Is the MLIR execution engine enabled there? You probably have to add a feature flag to `lit.cfg.py` that is set when arcilator's JIT feature is available. Then add `REQUIRES: arcilator-jit` lines to your tests. Check out the Slang tests, they had to do a similar thing since the Slang frontend can be disabled.

Moxinilian:
> The short integration test passes different options and definitions to its `cmake` call. Is the MLIR execution engine enabled there? You probably have to add a feature flag to `lit.cfg.py` that is set when arcilator's JIT feature is available. Then add `REQUIRES: arcilator-jit` lines to your tests. Check out the Slang tests, they had to do a similar thing since the Slang frontend can be disabled.

The odd thing is that I looked up the parameters in the GitHub workflow file and they are a superset of the ones I use locally (notably including the LLVM target for the host), so it should include ExcutionEngine. But maybe I looked in the wrong place.

Moxinilian:
I just realized that LLHD sim, which is enabled by default, unconditionally depends on execution engine being here. So there definitely is a configuration problem somewhere, this should work. I need to investigate.

Moxinilian:
Well, it is a bit annoying but I think the best way is indeed to make the tests not be ran in the short integration tests. The other tests should already run it on every PR so it should be alright.

maerhart:
Making the tests conditional on the arcilator JIT actually being there is great.

It looks like the short integration test CI uses a unified build of CIRCT and LLVM. I'm wondering whether this is just an issue specific to the CI machine or if this configuration does not work with unified builds in general. Did you try to reproduce this on a unified build locally?

Also just noticed that your JIT execution tests are in the `test/` directory. Did you try to move the tests that run the JIT to the 
`integration_test` directory? I haven't checked the test runtimes, but tests running JIT can have quite some overhead (to start the JIT, etc.) and shouldn't be run in the regular regression test suite. This is also a shortcoming of the LLHD Sim tests. I.e., they should only be run in the short integration tests CI on each PR and during the nightly run.

Moxinilian:
@maerhart I moved the JIT tests to the integration tests. They still are built conditionally because that indeed seems useful.

I did not reproduce in a unified build but that was my guess too. It just seemed like a CMake rabbit hole so I did not look into it more. I can if you require it for merging.

CMakeLists.txt
```diff
@@ -587,6 +587,16 @@ if(CIRCT_SLANG_FRONTEND_ENABLED)
   endif()
 endif()
 
+#-------------------------------------------------------------------------------
+# Arcilator JIT
+#-------------------------------------------------------------------------------
+
+if(MLIR_ENABLE_EXECUTION_ENGINE)
```
darthscsi, line 594: Is there a missing check on the JIR architecture support?  Does the existence of the option imply sufficient dependencies to get something working?

tools/arcilator/arcilator.cpp
```diff
@@ -68,89 +70,97 @@ using namespace arc;
 // Command Line Arguments
 //===----------------------------------------------------------------------===//
 
-static cl::OptionCategory mainCategory("arcilator Options");
+static llvm::cl::OptionCategory mainCategory("arcilator Options");
```
darthscsi, line 73: Changing namespaces and using is not that far off from a whitespace change.  If you are going to do it, please do it in a separate PR to avoid noise.

tools/arcilator/arcilator.cpp
```diff
@@ -528,11 +598,26 @@ int main(int argc, char **argv) {
...
-  cl::AddExtraVersionPrinter(
+  llvm::cl::AddExtraVersionPrinter(
       [](raw_ostream &os) { os << getCirctVersion() << '\n'; });
 
   // Parse pass names in main to ensure static initialization completed.
-  cl::ParseCommandLineOptions(argc, argv, "MLIR-based circuit simulator\n");
+  llvm::cl::ParseCommandLineOptions(argc, argv,
+                                    "MLIR-based circuit simulator\n");
+
+  if (!runSimulation.empty()) {
```
darthscsi, line 608: Alternately don't even have the cmdline option if jit is not enabled.

tools/arcilator/arcilator.cpp
Moxinilian: The reason I am doing it here is because the execution engine import introduces a name conflict if both the `mlir` and `llvm` namespaces are used. Hence why I had to drop one, and I picked llvm because it seemed to be the one most unrelated. Is a second PR really necessary? It is very much related to the meaningful content of this PR.

CMakeLists.txt
Moxinilian: If `MLIR_ENABLE_EXECUTION_ENGINE` is enabled, it means MLIR has generated the MLIRExecutionEngine library, so all the needed components must be in place. `MLIR_ENABLE_EXECUTION_ENGINE` is defined by MLIR's main CMake file, which does the necessary target checking.

tools/arcilator/arcilator.cpp
Moxinilian: I am worried that if the command line option disappears, it may create confusion for users that built CIRCT without knowing about the conditional build, and believe that the JIT interface has changed or does not exist.

tools/arcilator/arcilator.cpp
```diff
@@ -162,32 +172,35 @@ enum Until {
...
+    llvm::cl::desc("Specify output format"),
+    llvm::cl::values(clEnumValN(OutputMLIR, "emit-mlir", "Emit MLIR dialects"),
+                     clEnumValN(OutputLLVM, "emit-llvm", "Emit LLVM"),
+                     clEnumValN(OutputDisabled, "disable-output",
+                                "Do not output anything")),
+    llvm::cl::init(OutputLLVM), llvm::cl::cat(mainCategory));
+
+static llvm::cl::opt<std::string> runSimulation(
+    "run",
+    llvm::cl::desc("Name of the function to run containing the simulation"),
```
maerhart, line 202:
```suggestion
    llvm::cl::desc("Name of the function containing the simulation to run"),
```

tools/arcilator/arcilator.cpp
```diff
@@ -333,6 +346,32 @@ static LogicalResult processBuffer(
...
+
+    if (!llvm::isa<LLVM::LLVMFuncOp, func::FuncOp>(toCall)) {
+      llvm::errs() << "entry point '" << runSimulation
+                   << "' was found but on an operation of type '"
+                   << toCall->getName() << "' while a function was expected\n";
+      llvm::errs() << "supported functions: 'func.func', 'llvm.func'\n";
+      return failure();
+    }
+
+    auto toCallFunc = llvm::cast<FunctionOpInterface>(toCall);
```
maerhart, line 366: Nit: could just inline this variable since it's only used once

tools/arcilator/arcilator.cpp
```diff
@@ -333,6 +346,32 @@ static LogicalResult processBuffer(
...
+    if (toCallFunc.getNumArguments() != 0) {
+      llvm::errs() << "entry point '" << runSimulation
+                   << "' must have no arguments\n";
+      return failure();
+    }
+  }
+#endif // ARCILATOR_ENABLE_JIT
```
maerhart, line 373: What's the consideration of checking this before the lowering rather than once everything is lowered to LLVM. At that point you wouldn't have to check for `func.func` and could also support other custom functions, right?

integration_test/arcilator/JIT/err-wrong-func.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: ! (arcilator %s --run=main 2> %t) && (cat %t | FileCheck %s)
+// REQUIRES: arcilator-jit
+
+// CHECK: entry point 'main' must have no arguments
```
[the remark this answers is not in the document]
Moxinilian: I figured the return type can be safely ignored. What would really be a problem would be the dangling references of inputs.

CMakeLists.txt
```diff
@@ -587,6 +587,24 @@ if(CIRCT_SLANG_FRONTEND_ENABLED)
...
 
+#-------------------------------------------------------------------------------
+# Arcilator JIT
+#-------------------------------------------------------------------------------
+
+# If building as part of a unified build, whether or not MLIR's execution engine
+# is enabled must be fetched from its subdirectory scope.
+if(NOT (CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR))
+  get_directory_property(FETCHED_MLIR_ENABLE_EXECUTION_ENGINE
+      DIRECTORY "${CMAKE_SOURCE_DIR}/../mlir"
```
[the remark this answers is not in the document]
Moxinilian: Thank you for spotting this, I hate myself for missing it as I even looked for something like that…

tools/arcilator/arcilator.cpp
Moxinilian: I am sure I had a reason to do that but I cannot remember it… I guess I’ll move it down, perhaps while doing so it will come back to me.

tools/arcilator/arcilator.cpp
```diff
@@ -375,6 +414,36 @@ static LogicalResult processBuffer(
...
+      llvm::Error err = expectedFunc.takeError();
+      llvm::errs() << "failed to run simulation: " << err << "\n";
+      llvm::consumeError(std::move(err));
+      return failure();
+    }
+
+    void (*simulationFunc)(void **) = *expectedFunc;
+    (*simulationFunc)(nullptr);
+
+    return success();
```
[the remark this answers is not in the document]
Moxinilian: That’s a great question. My original approach was to have the JIT be another flavor of the emit flag, but then it was a bit weird to specify an entry point.

But perhaps I could do that and default the entry point to `main`, with the ability to use a second flag to set the entry point! What do you think?

tools/arcilator/arcilator.cpp
```diff
@@ -375,6 +414,36 @@ static LogicalResult processBuffer(
   if (failed(pmLlvm.run(module.get())))
     return failure();
 
+#ifdef ARCILATOR_ENABLE_JIT
+  // Handle JIT execution.
+  if (!runSimulation.empty()) {
+    mlir::ExecutionEngineOptions engineOptions;
```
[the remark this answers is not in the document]
Moxinilian: It’s simply because I don’t really know what that is. I figured the optimization option set to agressive would be enough but maybe I am missing parts of it. I’ll look into this tomorrow.

tools/arcilator/arcilator.cpp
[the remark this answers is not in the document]
maerhart: Sounds good to me. Maybe the default should be `entry` (or something else) though. Because with the JIT you can also emit the object file which could be an option to skip the separate opt and llc calls. But the `main` function needs to have two arguments which the entry points must not. We could then just insert the main function as a wrapper around `entry`

tools/arcilator/arcilator.cpp
Moxinilian: I remember now! It was so the output could contain the name of the operation it was pointing to so that it is consistent with the input. It would be weird for the error message to talk about a lowered state of the op, one could end up being very confused. Do you think this is worth it?

maerhart: It's nice to have good error messages, but I think here it's not worth it.

CMakeLists.txt
```diff
@@ -145,6 +145,12 @@ else()
   include_directories(SYSTEM ${MLIR_INCLUDE_DIR})
   include_directories(SYSTEM ${MLIR_TABLEGEN_OUTPUT_DIR})
 
+  # If building as part of a unified build, whether or not MLIR's execution engine
+  # is enabled must be fetched from its subdirectory scope.
+  get_directory_property(MLIR_ENABLE_EXECUTION_ENGINE
+      DIRECTORY ${MLIR_MAIN_SRC_DIR}
+      DEFINITION MLIR_ENABLE_EXECUTION_ENGINE)
+
```
maerhart, line 153: This looks even cleaner as before, nice! It is also done in the separate build now, right? Did you test that it still works fine with that; because the JIT tests are only run with the unified build in the CI?

maerhart (approved):
Not super familiar with the DLTI and CMake stuff, but this looks great to me now! Mostly just a question about the error reporting remaining.

CMakeLists.txt
Moxinilian: If you look closely, all this code is in a big if statement that is only ran in the unified build. In any case, I tested both unified and separate, and they both get the JIT enabled properly.

tools/arcilator/arcilator.cpp
```diff
@@ -28,11 +28,14 @@
 #include "mlir/Bytecode/BytecodeWriter.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlow.h"
+#include "mlir/Dialect/DLTI/DLTI.h"
```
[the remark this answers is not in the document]
Moxinilian: Ah yes it seems that I forgot to link it explicitly but it got linked anyway by chance. It should be fixed now.

CMakeLists.txt
maerhart: Ah, perfect!

fabianschuiki:
:partying_face: Very nice!

PR #6784, opened 2024-03-05
Bump LLVM to 1e828f838cc0f15074f3dbbb04929c06ef0c9729.

mikeurbach:
This picks up a bugfix in DialectConversion:
https://github.com/llvm/llvm-project/commit/9606655fbb03b1cf1c69d624c7320630e85f33e6


mikeurbach:
Opening this PR to share the status with the latest LLVM.

@seldridge pointed out the failure in nightly CI: https://github.com/llvm/circt/actions/runs/8155859443/job/22292261940#step:6:4600

That seems to be related to an upstream bug in DialectConversion: https://github.com/llvm/llvm-project/commit/9606655fbb03b1cf1c69d624c7320630e85f33e6

However, bumping past that, we are now hitting asserts in DialectConversion, which appear to be added in: https://github.com/llvm/llvm-project/commit/6008cd40b7bbdc66555550c2e38648d5ce99cc78

We are probably use DialectConversion in a way that is illegal, and now the infra is being stricter. The commit message above explains a concrete example, and how to rewrite it. We probably need to track down the failures and do the same.

mikeurbach:
Ibis folks--please see the linked issue. I tried to figure out how to avoid the new assertions in the Ibis transforms, but I'm not familiar enough with these passes to know the right fix. I'm proposing to XFAIL these tests so we can make progress with getting this LLVM bump in and fix the use-after-free we currently enjoy on main.

mikeurbach:
No longer XFAILing the Ibis tests. I'll merge this once CI completes.

PR #6785, opened 2024-03-05
[OM] Add C API and Python bindings for EvaluatorValue::Reference.

mikeurbach:
In some OM dialect constructs, it is possible to receive EvaluatorValue::Reference values. In the Python bindings, where we are converting an EvaluatorValue to a Python value, we need to dereference the Reference, to get at the underlying EvaluatorValue that was set during evaluation.

This adds the necessary C APIs, and updates the Python bindings to use them. If we encounter a Reference, we dereference it and recursively call the converter function.

A Python test was added using an example IR from the Evaluator unit tests, which delays evaluation and introduces references.

### DIFF
--- a/include/circt-c/Dialect/OM.h
+++ b/include/circt-c/Dialect/OM.h
@@ -207,6 +207,15 @@ omEvaluatorValueIsAPath(OMEvaluatorValue evaluatorValue);
 MLIR_CAPI_EXPORTED MlirAttribute
 omEvaluatorPathGetAsString(OMEvaluatorValue evaluatorValue);
 
+/// Query if the EvaluatorValue is a Reference.
+MLIR_CAPI_EXPORTED bool
+omEvaluatorValueIsAReference(OMEvaluatorValue evaluatorValue);
+
+/// Dereference a Reference EvaluatorValue. Emits an error and returns null if
+/// the Reference cannot be dereferenced.
+MLIR_CAPI_EXPORTED OMEvaluatorValue
+omEvaluatorValueGetReference(OMEvaluatorValue evaluatorValue);
+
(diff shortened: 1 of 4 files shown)


include/circt-c/Dialect/OM.h
```diff
@@ -207,6 +207,15 @@ omEvaluatorValueIsAPath(OMEvaluatorValue evaluatorValue);
...
 omEvaluatorPathGetAsString(OMEvaluatorValue evaluatorValue);
 
+/// Query if the EvaluatorValue is a Reference.
+MLIR_CAPI_EXPORTED bool
+omEvaluatorValueIsAReference(OMEvaluatorValue evaluatorValue);
+
+/// Dereference a Reference EvaluatorValue. Emits an error and returns null if
+/// the Reference cannot be dereferenced.
+MLIR_CAPI_EXPORTED OMEvaluatorValue
+omEvaluatorValueGetReference(OMEvaluatorValue evaluatorValue);
```
prithayan, line 217: hyper nit: `omEvaluatorValueGetReferenceValue` ?
The `omEvaluatorValueGetReference` sounds like it returns the reference to a value.

prithayan (approved):

include/circt-c/Dialect/OM.h
mikeurbach: Yeah, good point, I'll update that.

uenoku (commented):
It looks good but I didn't expect reference values are exposed to outside of evaluation... Evaluator::finalize was meant to eliminate intermediate reference values from values. I don't block but I'll take a look what's going there, thank you for catching this!

PR #6786, opened 2024-03-05
[OM] Extend APSInts as necessary before performing arithmetic.

mikeurbach:
Most interesting arithmetic on APSInt asserts that both operands are the same bitwidth, but the IntegerAttrs that we are working with may have used the smallest necessary bitwidth to represent the number they hold, and won't necessarily match.

This extends the smaller operand if necessary. This is safe to do, even for operands like the RHS of a shift right, because it will zero or sign extend without adding "significant bits" in the APSInt representation. So if we had a shift right amount that previously fit in 64 bits, and we had to extend it beyond 64 bits, we are still able to truncate to 64 bits when performing the shift.


prithayan (approved):
Looks Good.

PR #6789, opened 2024-03-06
[Emit] Group file header ops into `emit.fragment`

nandor:
Instead of using the replicated op mechanism, fragments of SV code which must precede certain modules are grouped into fragment ops. Modules which rely on them reference them through a `emit.fragment` attribute. The fragments are printed through ExportVerilog.

Subsequent PRs will fully replace replicated ops using fragments.

### DIFF
--- a/include/circt/Conversion/Passes.td
+++ b/include/circt/Conversion/Passes.td
@@ -418,10 +418,17 @@ def LowerFIRRTLToHW : Pass<"lower-firrtl-to-hw", "mlir::ModuleOp"> {
     Lower a module of FIRRTL dialect to the HW dialect family.
   }];
   let constructor = "circt::createLowerFIRRTLToHWPass()";
-  let dependentDialects = ["comb::CombDialect", "hw::HWDialect",
-                           "seq::SeqDialect", "sv::SVDialect",
-                           "ltl::LTLDialect", "verif::VerifDialect",
-                           "sim::SimDialect"];
+
+  let dependentDialects = [
+    "comb::CombDialect",
+    "emit::EmitDialect",
+    "hw::HWDialect",
+    "ltl::LTLDialect",
+    "seq::SeqDialect",
+    "sim::SimDialect",
+    "sv::SVDialect",
+    "verif::VerifDialect",
+  ];
   let options = [
     Option<"enableAnnotationWarning", "warn-on-unprocessed-annotations",
            "bool", "false",
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -6039,16 +6039,22 @@ class FileEmitter : public EmitterBase {
 public:
   explicit FileEmitter(VerilogEmitterState &state) : EmitterBase(state) {}
 
-  void emit(emit::FileOp op);
+  void emit(emit::FileOp op) {
+    emit(op.getBody());
+    ps.eof();
+  }
+  void emit(emit::FragmentOp op) { emit(op.getBody()); }
   void emit(emit::FileListOp op);
 
 private:
+  void emit(Block *block);
+
   void emitOp(emit::RefOp op);
   void emitOp(emit::VerbatimOp op);
 };
 
-void FileEmitter::emit(emit::FileOp op) {
-  for (Operation &op : *op.getBody()) {
+void FileEmitter::emit(Block *block) {
+  for (Operation &op : *block) {
     TypeSwitch<Operation *>(&op)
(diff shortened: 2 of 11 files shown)


test/Conversion/ExportVerilog/emit-fragment.mlir
```diff
@@ -0,0 +1,66 @@
...
+}
+
+emit.fragment @FragmentC {
+  sv.ifdef @MacroC {
+  } else {
+    sv.macro.def @MacroC "C"
+  }
+}
+
+hw.module @FirstModule(in %in : i32, out out : i32) attributes { "emit.fragments" = [@FragmentA, @FragmentB] } {
```
maerhart, line 60: Isn't adding the required fragments as discardable attributes quite fragile?

I'm wondering if you considered also putting the `hw.module` in its own fragment that references `@FragmentA` and `@FragmentB` as dependencies. Then, a file can be constructed by choosing a set of fragments that should go there (e.g., because they contain modules that should be in that file). The exporter then includes all fragments it depends on transitively in a legal order.

(I don't know a lot about the goals of this, how the compiler decides which module(s) go in which file, and ExportVerilog in general. I'm looking at the emit dialect because in arcilator we'd like to also emit things in separate files for parallel LLVM compilation and using the emit dialect for that could be nice)

test/Conversion/ExportVerilog/emit-fragment.mlir
```diff
@@ -0,0 +1,66 @@
...
+  hw.output %in : i32
+}
+
+hw.module @SecondModule(in %in : i32, out out : i32) attributes { "emit.fragments" = [@FragmentB, @FragmentC] } {
+  hw.output %in : i32
+}
```
maerhart, line 66: Nit: newline missing

maerhart (commented):
Just some drive-by comments

test/Conversion/ExportVerilog/emit-fragment.mlir
nandor: `hw.module` will soon be referenced by `emit.ref` and we'll be able to remove this attribute, using refs instead.

nandor:
Thanks for the comments, addressed.

include/circt/Conversion/Passes.td
```diff
@@ -418,10 +418,17 @@ def LowerFIRRTLToHW : Pass<"lower-firrtl-to-hw", "mlir::ModuleOp"> {
...
-                           "seq::SeqDialect", "sv::SVDialect",
-                           "ltl::LTLDialect", "verif::VerifDialect",
-                           "sim::SimDialect"];
+
+  let dependentDialects = [
+    "comb::CombDialect",
+    "emit::EmitDialect",
+    "hw::HWDialect",
+    "seq::SeqDialect",
+    "sv::SVDialect",
```
darthscsi, line 427: minor nit: alphabetize.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6338,20 +6346,37 @@ void SharedEmitterState::collectOpsForFile(const FileInfo &file,
...
       thingsToEmit.emplace_back(replicatedOps[lastReplicatedOp]);
 
+    // Pull in the fragments that the op references. In one file, each
+    // fragment is emitted only once.
+    if (auto fragments = op->getAttrOfType<ArrayAttr>("emit.fragments")) {
+      for (auto sym : fragments.getAsRange<FlatSymbolRefAttr>()) {
+        auto it = fragmentMapping.find(sym.getAttr());
+        if (it == fragmentMapping.end()) {
+          encounteredError = true;
+          op->emitError("cannot find referenced fragment ") << sym;
```
darthscsi, line 6368: Shouldn't this be caught during validation?

darthscsi: Specifically the generic symbol use verifier from SymbolUserOpInterface

darthscsi (commented):
Is there a reason not to make a dialect attribute declared in the td files for the fragment referencing array?

nandor:
This is a stopgap measure that improves replicated ops a bit to better group them until we sanitize the downstream flows sufficiently to be able to include them from a header, so I wouldn't want to make it a first-class attribute on modules, similarly to the `output_file` attribute which isn't specified either.

seldridge (approved):
Let's get this unblocked. Sorry for the delay.

How is this being tested for the `LowerToHW` changes? Is there enough coverage already with end-to-end tests? I was surprised that this did not need changes to existing `LowerToHW` tests. Can you include some tests of this if it is not already covered.

PR #6790, opened 2024-03-06
[InferReadWrite] Add heuristic to infer unmasked memory

prithayan:
This PR updates the heuristic to infer an unmasked memory.
If all the bits of a masked memory are exactly the same, then it can be replaced with an unmasked memory.
This is an attempt to fix a use case, in which `firtool` introduces masked memory for an aggregate data type when the user expected an unmasked one.


lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
```diff
@@ -335,11 +398,13 @@ struct InferReadWritePass : public InferReadWriteBase<InferReadWritePass> {
...
             // Check what is the mask field directly connected to.
-            // If, a constant 1, then we can replace with unMasked memory.
-            if (auto maskVal = getConnectSrc(sf))
-              if (auto constVal = dyn_cast<ConstantOp>(maskVal.getDefiningOp()))
-                if (constVal.getValue().isAllOnes())
-                  isMasked = false;
+            // If we can infer that all the bits of the mask are always assigned
+            // the same value, then the memory is unmasked.
+            if (auto maskVal = getConnectSrc(sf)) {
+              SmallVector<Value> bits;
```
uenoku, line 404: I'm not sure it's reasonable to use Value to represent bit precise states. I think llvm::KnownBits would be right better fit here. You may want to refer the similar analysis for comb https://github.com/llvm/circt/blob/e1a69e3f3ae4ffcc40ae663817ec315fc9422cd1/lib/Dialect/Comb/CombAnalysis.cpp#L86. Also note that it will cause compile time regression if we don't set a recursion depth limit.

prithayan: Thanks @uenoku , the analysis is very similar, but seems like its only trying to figure out the 1s and 0s in the bits.
But in this PR, the analysis is different.
1. In general the mask value is going to be constructed by concat of few 1 bit values, which may not be constants.
2. If the same Value is used to mask each field of an aggregate memory data, then each bit of the mask can be traced back to that Value.
3. So, this unmasked inference analysis is only trying to follow the dataflow into each of the bits of an Value to a source Value and then check if the source value is same for each bit.
I agree about the recursion depth, I will make this an iterative algorithm.

prithayan:
@seldridge , @debs-sifive This PR updates the heuristic to infer an unmasked memory, ready for a review.

mikeurbach (commented):
I'm not familiar here either but I think this change seems reasonable.

PR #6792, opened 2024-03-06
[FIRRTL][FIRParser] Add deprecation warning about printf/when-encoding.

dtzSiFive:
Only warn once per module to keep it from being entirely overwhelming. (and report how many, in terms of printf's, to convey magnitude)

Split out when-encoding parser behavior to dedicated file, and look for the deprecation diagnostics in tests that use them.


dtzSiFive:
If we know what release we plan to remove it, clarifying that would be an improvement.
I believe Chisel versions themselves no longer using this will make use of firtool with this warning included, but this helps catch anything missed or warn before removing it entirely.

fabianschuiki (approved):
LGTM! Love to see the deprecation warning and that you factored the corresponding tests into a separate file that will be easy to delete once we entirely nuke the printf-encoded verif statements.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5296,16 +5296,35 @@ FIRCircuitParser::parseModuleBody(DeferredModuleToParse &deferredModule) {
...
+        ++numVerifPrintfs;
+        if (!printfLoc)
+          printfLoc = loc;
+      }
+    }
+    if (numVerifPrintfs > 0) {
+      auto diag =
+          mlir::emitWarning(deferredModule.moduleOp.getLoc(),
+                            "module contains ")
+          << numVerifPrintfs
```
uenoku, line 5322: Can we open an issue for deprecation of printf-encoded verification and mention the issue in the warning message so that users can know what to do for migrations?

uenoku (approved):
LGTM!

dtzSiFive:
Marking draft until move off printf-encoded verif is completed.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5296,16 +5296,35 @@ FIRCircuitParser::parseModuleBody(DeferredModuleToParse &deferredModule) {
...
-      [&buffer](PrintFOp printFOp) { buffer.push_back(printFOp); });
-
-  for (auto printFOp : buffer) {
-    auto result = circt::firrtl::foldWhenEncodedVerifOp(printFOp);
-    if (failed(result))
-      return result;
+  {
+    // Gather PrintFOps and then process to avoid walking while mutating.
+    SmallVector<PrintFOp> buffer;
+    deferredModule.moduleOp.walk(
```
darthscsi, line 5302: This walk seems unnecessary.  Should we be collecting printfs during parse in something in the statement parser?  I realize this is a pre-existing condition.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5296,16 +5296,35 @@ FIRCircuitParser::parseModuleBody(DeferredModuleToParse &deferredModule) {
...
+          printfLoc = loc;
+      }
+    }
+    if (numVerifPrintfs > 0) {
+      auto diag =
+          mlir::emitWarning(deferredModule.moduleOp.getLoc(),
+                            "module contains ")
+          << numVerifPrintfs
+          << " printf-encoded verification operation(s), which are deprecated "
+             "and will be removed in the future";
```
darthscsi, line 5324: Once per module might be annoying (but do we care since they are going away?) and could be once per circuit.

We might consider having a generic deprecated feature use collector.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
dtzSiFive: Agreed.  Inclined to leave as-is for now with plan to phase all of this out soon (and so would rather not thread that through just to remove it).

lib/Dialect/FIRRTL/Import/FIRParser.cpp
dtzSiFive: Put up an issue here: https://github.com/llvm/circt/issues/6970 .

dtzSiFive: Link added to diagnostic.

PR #6794, opened 2024-03-06
[FIRRTL] Fix sign bit truncation in constant parser

fabianschuiki:
Fix an issue in the parser of the `firrtl.constant` op, which would truncate the sign bit of negative constants without reporting an error. This would for example accept -2 as a valid 1-bit constant, truncating the `...11111110` bit pattern to `0`.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLOps.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOps.cpp
@@ -3933,8 +3933,10 @@ ParseResult ConstantOp::parse(OpAsmParser &parser, OperationState &result) {
     } else if (width < value.getBitWidth()) {
       // The parser can return an unnecessarily wide result with leading
       // zeros. This isn't a problem, but truncating off bits is bad.
-      if (value.getNumSignBits() < value.getBitWidth() - width)
-        return parser.emitError(loc, "constant too large for result type ")
+      unsigned neededBits = value.isNegative() ? value.getSignificantBits()
+                                               : value.getActiveBits();
+      if (width < neededBits)
+        return parser.emitError(loc, "constant out of range for result type ")
                << resultType;
       value = value.trunc(width);
     }
--- a/test/Dialect/FIRRTL/errors.mlir
+++ b/test/Dialect/FIRRTL/errors.mlir
@@ -139,7 +139,7 @@ firrtl.module @Foo() {
 
 firrtl.circuit "Foo" {
 firrtl.module @Foo() {
-  // expected-error @+1 {{constant too large for result type}}
+  // expected-error @+1 {{constant out of range for result type}}
   firrtl.constant 100 : !firrtl.uint<4>
 }
 }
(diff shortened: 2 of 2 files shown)


test/Dialect/FIRRTL/errors.mlir
```diff
@@ -155,6 +155,15 @@ firrtl.module @Foo() {
 
 // -----
 
+firrtl.circuit "Foo" {
+firrtl.module @Foo() {
+  // expected-error @+1 {{constant too large for result type}}
```
maerhart, line 160: Nit: the error message is a little unfortunate because here it's actually too small to fit

dtzSiFive (approved):
Eep! Thanks for fixing this!

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -3933,7 +3933,9 @@ ParseResult ConstantOp::parse(OpAsmParser &parser, OperationState &result) {
     } else if (width < value.getBitWidth()) {
       // The parser can return an unnecessarily wide result with leading
       // zeros. This isn't a problem, but truncating off bits is bad.
-      if (value.getNumSignBits() < value.getBitWidth() - width)
+      unsigned neededBits = value.isNegative() ? value.getSignificantBits()
+                                               : value.getActiveBits();
+      if (width < neededBits)
```
maerhart, line 3938: Are we actually allowed to remove the leading zero if it is a positive value (which would happen here because `getActiveBits == bitWidth - numLeadingZeros`)? I think the APInt is just used as a signless bitvector and the return type actually determines how this bitvector should be interpreted. If the type is `sint` then it's two's complement. E.g., we have the binary number `01` (1 in decimal) and we truncate it to `1` then it's -1 in decimal. 1 (decimal) cannot be represented in a 1-bit signed integer. If the APInt is interpreted as unsigned, then we can truncate to `1`.

test/Dialect/FIRRTL/errors.mlir
fabianschuiki: Haha yeah good point. Magnitude is too large. "constant is too magnific ..."?

dtzSiFive: It's too large in magnitude? :upside_down_face: 
But yeah, out of range/cannot fit anyway.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
fabianschuiki: I think it makes sense to look at the parsed value and see if that value as it was parsed can be truncated to the bit width. If it's positive it has a leading 0 bit (the binary `0b10` actually parses into APInt `010`), and if it's negative it has a leading 1 bit. That means we can always check if the user typed a negative or positive value. Then this check ensures that by cutting off leading bits we don't change the interpretation of the number, or rather, that there is no overflow.

maerhart:
Btw. please don't wait for me, just dumped my thoughts 😅

PR #6797, opened 2024-03-07
[arcilator] Strip Emit dialect operations

fzi-hielscher:
This PR enables the arcilator tool to run on IR containing Emit dialect nodes. However, they are simply discarded. It is mainly intended to allow lowering of designs containing blackboxed stubs, notably the plusargreader in rocket-chip. Obviously, this is just a stopgap solution until we can feed them to ImportVerilog. :wink:

### DIFF
--- a/lib/Dialect/Arc/Transforms/StripSV.cpp
+++ b/lib/Dialect/Arc/Transforms/StripSV.cpp
@@ -9,6 +9,7 @@
 #include "circt/Dialect/Arc/ArcOps.h"
 #include "circt/Dialect/Arc/ArcPasses.h"
 #include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/Emit/EmitDialect.h"
 #include "circt/Dialect/OM/OMDialect.h"
 #include "circt/Dialect/SV/SVOps.h"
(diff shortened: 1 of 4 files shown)


lib/Dialect/Arc/Transforms/StripSV.cpp
```diff
@@ -9,6 +9,7 @@
 #include "circt/Dialect/Arc/ArcOps.h"
 #include "circt/Dialect/Arc/ArcPasses.h"
 #include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/Emit/EmitDialect.h"
```
maerhart, line 12: Don't we need to link `ArcTransforms` against `CIRCTEmit`?

maerhart (approved):
Thanks!

lib/Dialect/Arc/Transforms/StripSV.cpp
fzi-hielscher: Yes, good catch. It did not fail on my builds, but that's probably coincidental.

nandor (approved):
LGTM, thanks for fixing this!

fabianschuiki (approved):
Thanks for the fix!

PR #6798, opened 2024-03-07
[CMake] Add CMake targets to install all CIRCT libraries.

mikeurbach:
These targets are useful for downstream projects that want to install
all CIRCT libraries. The CMake commands follow how similar targets are
set up in upstream MLIR.


teqdruid (approved):
This seems reasonable to me... with the usual caveat.

mikeurbach:
> with the usual caveat.

😄 yep, I like the ones where it's just copying something that worked for upstream MLIR.

PR #6802, opened 2024-03-08
[SMT] Add Dialect rationale and boilerplate

maerhart:
This PR adds the rationale and boilerplate for the SMT dialect, as discussed in the ODM on January 17th. It also adds a page about formal verification in CIRCT in general.

The goal is to build the infrastructure as outlined in the attached markdown documents and replace the internals of the existing `circt-lec` tool with an implementation based on this new infrastructure. The old code will then be removed. If people agree, we can also remove the link dependency on Z3 again (we only need the dynamic library installed on the CI machine to run the integration tests).

In the ODM, a concern was raised that there is already an SMT dialect in the works that might be upstreamed to MLIR. The work referred to is an [xDSL](https://github.com/xdslproject/xdsl) based [SMT dialect](https://github.com/opencompl/xdsl-smt) by @math-fehr. This implementation of an SMT dialect is very similar, but not exactly the same. We discussed how we want to proceed with this and concluded that the best approach is to add this dialect to CIRCT and only further upstream it to MLIR if there are people outside of CIRCT interested in it.

Another concern was that adding more lowering steps could make it more difficult to back-propagate counter examples given by the solver. I'd argue that it actually gets easier because we can mix in the debug dialect with the SMT dialect and share code for back-propagation. This way, the relevant SSA def-use mappings stay intact and don't have to be stored in separate custom data structures. 

This infrastructure not only replaces the current `circt-lec` implementation but is also relevant to work done by @TaoBi22, @math-fehr, and @luisacicolini. If anyone else has immediate plans to use this dialect, feel free to reach out to discuss and influence design decisions.

If you want to have a more detailed look at how this will look once everything is upstreamed, you can refer to [this branch](https://github.com/maerhart/circt/tree/maerhart-lec-redesign). You can even play around with the new LLVM JIT-based `circt-lec` there (it is already functionally equivalent to the current LEC). Note, however, that it includes many experimental things that won't be upstreamed, and the other parts will get more polished and modified according to some lessons learned during experimentation (it should be seen more as a prototype or proof-of-concept). There already exists an implementation of a CIRCT bounded model checker based on this dialect on that branch as well. This dialect also makes it easier to add SMT encodings for higher-level dialects to the LEC and BMC tools, as some experiments on the handshake dialect have shown.


docs/FormalVerification.md
```diff
@@ -0,0 +1,326 @@
...
+
+Let's suppose we want to implement our own custom LEC tool for our novel HDL
+language which lowers to a CIRCT/MLIR based IR. There are two ways to implement
+this:
+1. Lower this IR to the CIRCT core representation (HW, Comb, Seq) and just use
+the same pipeline as `circt-lec`. This has the advantage that we don't need to
+think about SMT encodings and we might already compile to those dialects anyway
+for Verilog emission.
+2. Implement a pass that sets up the LEC problem (lowering to the `verif.lec`
+operation), and another pass that encodes the IRs operations and types in SMT
```
fabianschuiki, line 80:
```suggestion
operation), and another pass that encodes the IR operations and types in SMT
```

fabianschuiki (approved):
This is really cool and impressive work! I love all the documentation and the added general page on formal verification in CIRCT 🥳.

The documentation now "promises" ops like `verif.lec`, despite them not being upstreamed yet. Would it make sense to land those parts of the documentation together with the ops themselves? This would prevent the docs from talking about features that aren't there yet.

maerhart:
> The documentation now "promises" ops like `verif.lec`, despite them not being upstreamed yet. Would it make sense to land those parts of the documentation together with the ops themselves? This would prevent the docs from talking about features that aren't there yet.

Do you think we should only add the SMT dialect rationale here and add the general formal verification doc once all the features have landed?

fabianschuiki:
That could work! And maybe keep the introductory paragraph of the formal verif page as well, to show the intent of the doc. And then as we upstream, the documentation grows and documents the features that got added.

A very minor thing. I'm also totally fine with landing all this wholesale, because it's great to indicate direction. 😁

math-fehr (approved):
Really nice to see where this is going!
As Martin mentionned, I've been using a very similar design for quite a while, and it works fine!
The only difference is that this is lowering to LLVM, which I think makes a lot of sense for your use cases!
And yes, I think it makes a lot of sense to try to push this in CIRCT now (as some people in CIRCT are interested by it), and then see if we can push it to MLIR later on!

luisacicolini (commented):
Great job!!

darthscsi (approved):
Exciting to see this coming along.

PR #6805, opened 2024-03-10
[SMT] Add arithmetic and bitwise operations

maerhart:
Add most of the basic arithmetic and bitwise operations we need to lower from `comb`.


include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -61,4 +61,50 @@ def BVConstantOp : SMTBVOp<"constant", [
   let hasFolder = true;
 }
 
+class BVArithmeticOrBitwiseOp<string mnemonic, string desc> :
+    SMTBVOp<mnemonic, [Pure, SameOperandsAndResultType]> {
```
math-fehr, line 65: Can you add the `ConditionallySpeculatable` trait? That's just a trait to say that this operation has no undefined behavior, so you can hoist it out of a loop safely (which Pure doesn't let you do).

math-fehr (approved):
Very nice!

include/circt/Dialect/SMT/SMTBitVectorOps.td
maerhart: `Pure` means speculatable and memory-effect-free. See the definitions from the tablegen file:
```
def AlwaysSpeculatable : TraitList<[ConditionallySpeculatable, AlwaysSpeculatableImplTrait]>;
def Pure : TraitList<[AlwaysSpeculatable, NoMemoryEffect]>;
```

math-fehr: Oh, my bad, I thought it was distinct!

fabianschuiki (approved):
LGTM!

PR #6806, opened 2024-03-10
[SMT] Add declare_const, assert, solver, and check operations

maerhart:
The `declare_const` operation could be pure if `fresh` is not specified. If `fresh` maybe an allocate side-effect could work? I think tablegen does not have the traits to specify this so fine-grained, we'd need to fall back to C++, so I'd leave it to another PR. Also to have it as the sole focus. There might be some corner cases to think about, e.g., what if there are `declare_const` with the same identifier, but different sort/type. The lack of such traits and optimizations makes it behave like SMT-LIB, though.

### DIFF
--- a/include/circt/Dialect/SMT/SMTOps.td
+++ b/include/circt/Dialect/SMT/SMTOps.td
@@ -16,8 +16,152 @@ include "mlir/IR/EnumAttr.td"
 include "mlir/IR/OpAsmInterface.td"
 include "mlir/Interfaces/InferTypeOpInterface.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
+include "mlir/Interfaces/ControlFlowInterfaces.td"
 
 class SMTOp<string mnemonic, list<Trait> traits = []> :
   Op<SMTDialect, mnemonic, traits>;
+
+def DeclareConstOp : SMTOp<"declare_const", [
+  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>
+]> {
+  let summary = "declare a symbolic value of a given sort";
+  let description = [{
+    This operation declares a symbolic value just as the `declare-const`
+    statement in SMT-LIB 2.6. The result type determines the SMT sort of the
+    symbolic value. The returned value can then be used to refer to the symbolic
+    value instead of using the identifier like in SMT-LIB.
+
+    The optionally provided string will be used as a prefix for the newly
+    generated identifier (useful for easier readability when exporting to
+    SMT-LIB). Each `declare_const` will always provide a unique new symbolic
+    value even if the identifier strings are the same.
+
+    Note that this operation cannot be marked as Pure since two operations (even
+    with the same identifier string) could then be CSEd, leading to incorrect
+    behavior.
+  }];
+
+  let arguments = (ins OptionalAttr<StrAttr>:$namePrefix);
+  let results = (outs Res<AnySMTType, "a symbolic value", [MemAlloc]>:$result);
+
+  let assemblyFormat = [{
+    ($namePrefix^)? attr-dict `:` qualified(type($result))
+  }];
+}
+
+def SolverOp : SMTOp<"solver", [
+  IsolatedFromAbove,
+  SingleBlockImplicitTerminator<"smt::YieldOp">,
+]> {
+  let summary = "create a solver instance within a lifespan";
+  let description = [{
+    This operation defines an SMT context with a solver instance. SMT operations
+    are only valid when being executed between the start and end of the region
+    of this operation. Any invocation outside is undefined. However, they do not
+    have to be direct children of this operation. For example, it is allowed to
+    have SMT operations in a `func.func` which is only called from within this
+    region. No SMT value may enter or exit the lifespan of this region (such
+    that no value created from another SMT context can be used in this scope and
+    the solver can deallocate all state required to keep track of SMT values at
+    the end).
+
+    As a result, the region is comparable to an entire SMT-LIB script, but
+    allows for concrete operations and control-flow. Concrete values may be
+    passed in and returned to influence the computations after the `smt.solver`
+    operation.
+
+    Example:
+    ```mlir
+    %0:2 = smt.solver (%in) {smt.some_attr} : (i8) -> (i8, i32) {
+    ^bb0(%arg0: i8):
+      %c = smt.declare_const "c" : !smt.bool
+      smt.assert %c
+      %1 = smt.check sat {
+        %c1_i32 = arith.constant 1 : i32
+        smt.yield %c1_i32 : i32
+      } unknown {
+        %c0_i32 = arith.constant 0 : i32
+        smt.yield %c0_i32 : i32
+      } unsat {
+        %c-1_i32 = arith.constant -1 : i32
+        smt.yield %c-1_i32 : i32
+      } -> i32
+      smt.yield %arg0, %1 : i8, i32
+    }
+    ```
+
+    TODO: solver configuration attributes
+  }];
+
+  let arguments = (ins Variadic<AnyNonSMTType>:$inputs);
+  let regions = (region SizedRegion<1>:$bodyRegion);
+  let results = (outs Variadic<AnyNonSMTType>:$results);
+
+  let assemblyFormat = [{
+    `(` $inputs `)` attr-dict `:` functional-type($inputs, $results) $bodyRegion
+  }];
+
+  let hasRegionVerifier = true;
+}
+
+def AssertOp : SMTOp<"assert", []> {
+  let summary = "assert that a boolean expression holds";
+  let arguments = (ins BoolType:$input);
+  let assemblyFormat = "$input attr-dict";
+}
+
+def CheckOp : SMTOp<"check", [
+  NoRegionArguments,
+  SingleBlockImplicitTerminator<"smt::YieldOp">,
+]> {
+  let summary = "check if the current set of assertions is satisfiable";
+  let description = [{
+    This operation checks if all the assertions in the solver defined by the
+    nearest ancestor operation of type `smt.solver` are consistent. The outcome
+    an be 'satisfiable', 'unknown', or 'unsatisfiable' and the corresponding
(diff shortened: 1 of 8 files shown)


include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -16,8 +16,152 @@ include "mlir/IR/EnumAttr.td"
...
+    Unlike the SMT-LIB version, this operation allows for a `fresh` attribute to
+    indicate that a new unique value should be generated. The provided string
+    will then be used as a prefix for the newly generated identifier. If this
+    attribute is not present and there are two such operations
+    `%a = smt.declare_const "a"` and `%b = smt.declare_const "a"` then `%a` and
+    `%b` are equivalent, i.e., they refer to the same symbolic value.
+
+    Note that this operation cannot be unconditionally marked as Pure since two
+    operations with the `fresh` attribute and the same identifier string could
+    then be CSEd, leading to incorrect behavior.
```
fabianschuiki, line 41: Does this need to have two separate SSA values that refer to the same declared constant ever arise? Could it make sense to always create a fresh constant, and then mark this operation pure? I could imagine that the aliasing/reuse of same name constants is more of a SMT-LIB writing convenience, but if you're lowering to that from an SSA representation, you might as well enforce a single definition rule.

maerhart: Good point. I agree that we should always create a fresh symbolic value. This non-fresh behavior was the initial implementation modeled too closely to an AST.
We cannot mark it 'pure' though since two symbolic values with the same prefix string will represent separate symbolic values and them being pure would allow CSE to merge them, i.e., change the behavior.

maerhart: I now declare the `declare_const` operation's result as a memory allocation, so it is DCEd but not CSEd.

test/Dialect/SMT/basic.mlir
```diff
@@ -5,3 +5,51 @@
...
 }
+
+func.func @core(%in: i8) {
+  // CHECK: %a = smt.declare_const "a" {smt.some_attr} : !smt.bool
+  %a = smt.declare_const "a" {smt.some_attr} : !smt.bool
+  // CHECK: smt.declare_const {smt.some_attr} : !smt.bv<32>
+  %b = smt.declare_const {smt.some_attr} : !smt.bv<32>
+
+  // CHECK: smt.assert %a {smt.some_attr}
+  smt.assert %a {smt.some_attr}
```
math-fehr, line 16: Just to be clear, this supposedly has undefined semantics, right?
As there is no toplevel `solver` here.

test/Dialect/SMT/cse-test.mlir
```diff
@@ -0,0 +1,12 @@
...
+func.func @declare_const_cse(%in: i8) -> (!smt.bool, !smt.bool){
+  // CHECK: smt.declare_const "a" : !smt.bool
+  %a = smt.declare_const "a" : !smt.bool
+  // CHECK-NEXT: smt.declare_const "a" : !smt.bool
+  %b = smt.declare_const "a" : !smt.bool
+  // CHECK-NEXT: return
+  %c = smt.declare_const "a" : !smt.bool
+
+  return %a, %b : !smt.bool, !smt.bool
+}
```
math-fehr, line 12: Same here, should this be put in a `smt.solver`?

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -16,8 +16,152 @@ include "mlir/IR/EnumAttr.td"
...
+    the end).
+
+    As a result, the region is comparable to an entire SMT-LIB script, but
+    allows for concrete operations and control-flow. Concrete values may be
+    passed in and returned to influence the computations after the `smt.solver`
+    operation.
+
+    Example:
+    ```mlir
+    %0:2 = smt.solver (%in) {smt.some_attr} : (i8) -> (i8, i32) {
```
math-fehr, line 75: What does the input/output correspond to?

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -16,8 +16,152 @@ include "mlir/IR/EnumAttr.td"
...
+      } -> i32
+      smt.yield %arg0, %1 : i8, i32
+    }
+    ```
+
+    TODO: solver configuration attributes
+  }];
+
+  let arguments = (ins Variadic<AnyNonSMTType>:$inputs);
+  let regions = (region SizedRegion<1>:$bodyRegion);
```
math-fehr, line 97: If you want to allow `scf.if` in your `smt.solver`, then you'll need to relax this and allow arbitrary CFGs, right?.

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -16,8 +16,152 @@ include "mlir/IR/EnumAttr.td"
...
+
+  let assemblyFormat = [{
+    ($namePrefix^)? attr-dict `:` qualified(type($result))
+  }];
+}
+
+def SolverOp : SMTOp<"solver", [
+  IsolatedFromAbove,
+  SingleBlockImplicitTerminator<"smt::YieldOp">,
+]> {
```
math-fehr, line 55: Do you think it would make more sense to have a solver SSA Value instead, that is passed to the other operations (smt.check, or smt.assert for instance)? If I recall, this is the design you used to have in the draft PR?

I feel it makes the restriction "being called inside an `smt.solver`" more easier to check, as this is just an SSA value that is passed.

Or does this design makes it easier to lower to LLVM? Or what else does it simplify?

math-fehr (commented):
Nice! I'm just having a question about `smt.solver`, but otherwise looks good to me!

test/Dialect/SMT/basic.mlir
maerhart: It depends. In this file, no entry point function is defined, so if you insert a `main` function that has an `smt.solver` that has a call to this function in its body, it would be well-defined. If you run this function directly, it's undefined (in practice, it will crash because no SMT context is initialized). I didn't focus on this here because it's just a regression test for the parser and printer, but I can add one if you want.

I was also thinking about requiring all smt operations being (transitive) children of the `smt.solver` operation, but that would make things a lot more rigid (making it more difficult to reuse code, forcing inlining of everything, etc.). On the positive side, we could properly verify a context is there.

include/circt/Dialect/SMT/SMTOps.td
maerhart: The body of the `smt.solver` is isolated from above. The reason is that SMT constants cannot be hoisted out of the solver and we have a bit more leverage to check that no SMT value from another solver is used in this solver. Both of which is a big problem in the LLVM lowering. The earlier because no smt context is initialized outside the solver's body and just attempting to build an expression node would lead to a crash, the latter because the nodes would have been built in different contexts with potentially other configs, i.e., are not compatible.
Still, it's useful to be able to pass in concrete values to influence the construction of the SMT expressions. And the output is useful to report the result of the solver (was is sat or unsat, a counter example, etc.). Since they are just passthrough values, they must match the block arguments and yield operands respectively.

include/circt/Dialect/SMT/SMTOps.td
maerhart: Only if I want to already lower them to the CF dialect before lowering the SMT dialect, right? Not sure if we need that. On the other hand, I find it quite handy when we can just assume there is only one block.

include/circt/Dialect/SMT/SMTOps.td
maerhart: The `smt.solver` does not only instantiate and limit the live-time of a solver that can be passed to the assert and check operations (like in the Z3 python API), but also the overall SMT context that stores all the config flags, nodes, reference counters, etc. Every operation in this dialect lowers to one (or multiple) calls to a Z3 C API function, all of which take this context as an argument. 

Having the region defining the SMT context's lifespan is quite important because when lowering to LLVM, you need to know when to create and initialize a context instance and when to deallocate it again. We could have a solver value provided as a special block argument in this region. But we still need the region checking that no solver value can enter or exit the region, otherwise it would be possible to mix contexts (pass SMT expression nodes built with different contexts to a constructor call for a new node) which would also be UB.

As you saw in the draft branch, we had this solver type (which we can lower to a struct containing a pointer to the context and another one to the solver) that was passed to the assert and check operations, which could be used to support multiple solvers next to each other. However, to solve the issue of guaranteeing that a context is available, we need this pointer to the context present at all operations. So, we could also lower all the SMT values (bool, bitvector, int, etc.) to a pair of pointers (one for the AST node, one for the context) which means we don't need to add the solver value to all operations, but only to the constant operations (bool, bitvector, int constants, declare_const, etc.) which don't get the pointer from an operand in the first place.

The current design basically targets lowering the context and solver pointers to LLVM globals. At the start of the solver region, we initialize them, and at the end, we deallocate them. All operations just access those globals, and if they follow the specification, they get a legal context; otherwise, they get an invalid pointer, thus UB.

Do you think we should add a solver/context type, add an implicit block argument to the solver region with this type, and require it as an operand for all operations that currently don't have an SMT value operand (i.e., constants)? This means we need to funnel the new solver value through all function calls and other control flow ops, i.e., slightly more complicated lowering, and have to add additional operands to the constant operations, but we can get rid of the globals, and get stronger guarantees.
Or leave it as is and have a somewhat easier lowering, require globals, and get fewer guarantees?

include/circt/Dialect/SMT/SMTOps.td
math-fehr: Oh yeah that makes sense, I also got reminded of `scf.execute_region` which can be used in case we really need that.

maerhart:
Is it okay for you if I merge this as is, or do you want me to change something?

math-fehr (approved):

include/circt/Dialect/SMT/SMTOps.td
maerhart: For the record, @math-fehr @fabianschuiki and I discussed this offline and came to the conclusion that the current design and implementation make the most sense.

PR #6807, opened 2024-03-11
[SMT] Add comparison operations

maerhart:
Unfortunately, I had to write a custom parser for the `eq` and `distinct` operations to only specify one of the operands types (`eq %a, %b : !smt.bool` vs. `eq %a, %b : !smt.bool, !smt.bool`). Let me know if there's a feature for that in tablegen in case I missed something.


fabianschuiki (approved):
LGTM! I'm not aware of how you can specify a single type in the assembly format for that variadic thing. Especially since you don't have a result type that matches the operands, which you could use.

I wish TableGen would adopt xDSL's notion of type variables, to have a type variable to talk about instead of `type(<somevalue>)`. Would help a lot.

TaoBi22 (approved):
This all looks good to me!

PR #6811, opened 2024-03-11
[FIRRTL] Don't enforce owning module for local ref in LowerClasses.

mikeurbach:
We have an existing check that paths are in the same owning module as the entity they target. This is absolutely required for hierarchical references to compose with basepaths. However, we have legacy flows that extract entities out of their owning module, and we want to be able to target those things. This works around these conflicting requirements by skipping the owning module check for local targets only. This is safe because we still check such local targets are only instantiated once, so there is no ambiguity even though they are not instantiated within the owning module.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -269,8 +269,17 @@ LogicalResult LowerClassesPass::processPaths(
         if (node->getModule() == owningModule)
           break;
         // If there are no more parents, then the path op lives in a different
-        // hierarchy than the HW object it references, which is an error.
+        // hierarchy than the HW object it references, which is an error if the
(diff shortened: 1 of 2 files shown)


prithayan (approved):
Seems like a reasonable solution to handle local targets that are outside the current hierarchy.

dtzSiFive (approved):
Let's try to not let this live too long, but LGTM.

If filing an issue would help not forget then great but up to you.

Random thought: can the symbol become a hierpath due to dedup or something? If so, is that okay? Maybe it's fine anyway and not likely but just bouncing it off ya re: legality/safety.

mikeurbach:
> If filing an issue would help not forget then great but up to you.

Yeah I will file an issue to revert this once we stop using ExtractInstances

> can the symbol become a hierpath due to dedup or something?

This can happen, and if it does the target will be subject to the existing "owning module" constraint. This only skips the constraint for targets that are still fully local by the time they reach LowerClasses.

PR #6813, opened 2024-03-12
[SMT] Add concat, extract, repeat operations

maerhart:
The last few operations to be able to fully lower comb (except `comb.mux` for which we need more boolean operators to use `ite`, and `comb.parity`).

### DIFF
--- a/include/circt/Dialect/SMT/SMTBitVectorOps.td
+++ b/include/circt/Dialect/SMT/SMTBitVectorOps.td
@@ -146,4 +146,96 @@ def BVCmpOp : SMTBVOp<"cmp", [Pure, SameTypeOperands]> {
   }];
 }
 
+def ConcatOp : SMTBVOp<"concat", [
+  Pure,
+  DeclareOpInterfaceMethods<InferTypeOpInterface, ["inferReturnTypes"]>
+]> {
+  let summary = "bit-vector concatenation";
+  let description = [{
+    This operation concatenates bit-vector values with semantics equivalent to
+    the `concat` operator defined in the SMT-LIB 2.6 standard. More precisely in
+    the [theory of FixedSizeBitVectors](https://smtlib.cs.uiowa.edu/Theories/FixedSizeBitVectors.smt2)
+    and the [QF_BV logic](https://smtlib.cs.uiowa.edu/Logics/QF_BV.smt2)
+    describing closed quantifier-free formulas over the theory of fixed-size
+    bit-vectors.
+
+    Note that the following equivalences hold:
+    * `smt.bv.concat %a, %b : !smt.bv<4>, !smt.bv<4>` is equivalent to
+      `(concat a b)` in SMT-LIB
+    * `(= (concat #xf #x0) #xf0)`
+  }];
+
+  let arguments = (ins BitVectorType:$lhs, BitVectorType:$rhs);
+  let results = (outs BitVectorType:$result);
+
+  let assemblyFormat = "$lhs `,` $rhs attr-dict `:` qualified(type(operands))";
+}
+
+def ExtractOp : SMTBVOp<"extract", [Pure]> {
+  let summary = "bit-vector extraction";
+  let description = [{
+    This operation extracts the range of bits starting at the 'lowBit' index
+    (inclusive) up to the 'lowBit' + result-width index (exclusive). The
+    semantics are equivalent to the `extract` operator defined in the SMT-LIB
+    2.6 standard. More precisely in the
+    [theory of FixedSizeBitVectors](https://smtlib.cs.uiowa.edu/Theories/FixedSizeBitVectors.smt2)
+    and the [QF_BV logic](https://smtlib.cs.uiowa.edu/Logics/QF_BV.smt2)
+    describing closed quantifier-free formulas over the theory of fixed-size
+    bit-vectors.
+
+    Note that `smt.bv.extract %bv from 2 : (!smt.bv<32>) -> !smt.bv<16>` is
+    equivalent to `((_ extract 17 2) bv)`, i.e., the SMT-LIB operator takes the
+    low and high indices where both are inclusive. The following equivalence
+    holds: `(= ((_ extract 3 0) #x0f) #xf)`
+  }];
+
+  let arguments = (ins I32Attr:$lowBit, BitVectorType:$input);
+  let results = (outs BitVectorType:$result);
+
+  let assemblyFormat = [{
+    $input `from` $lowBit attr-dict `:` functional-type($input, $result)
+  }];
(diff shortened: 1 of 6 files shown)


include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -146,4 +146,91 @@ def BVCmpOp : SMTBVOp<"cmp", [Pure, SameTypeOperands]> {
...
+    the `concat` operator defined in the SMT-LIB 2.6 standard. More precisely in
+    the [theory of FixedSizeBitVectors](https://smtlib.cs.uiowa.edu/Theories/FixedSizeBitVectors.smt2)
+    and the [QF_BV logic](https://smtlib.cs.uiowa.edu/Logics/QF_BV.smt2)
+    describing closed quantifier-free formulas over the theory of fixed-size
+    bit-vectors.
+
+    Note that the following equivalences hold:
+    * `smt.bv.concat %a, %b : !smt.bv<4>, !smt.bv<4>` is equivalent to
+      `(concat a b)` in SMT-LIB
+    * `(= (concat #xf #x0) #xf0)`
```
fabianschuiki, line 165: Nice clarification of the arrangement of bits after concatenation.

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -146,4 +146,91 @@ def BVCmpOp : SMTBVOp<"cmp", [Pure, SameTypeOperands]> {
...
+  let results = (outs BitVectorType:$result);
+
+  let assemblyFormat = "$lhs `,` $rhs attr-dict `:` qualified(type(operands))";
+}
+
+def ExtractOp : SMTBVOp<"extract", [Pure]> {
+  let summary = "bit-vector extraction";
+  let description = [{
+    This operation extracts the range of bits starting at the 'lowBit' index
+    (including) up to the 'lowBit' + result-width index (excluding). The
```
fabianschuiki, line 178:
```suggestion
    (inclusive) up to the 'lowBit' + result-width index (exclusive). The
```
Maybe?

fabianschuiki (approved):
LGTM!

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -146,4 +146,91 @@ def BVCmpOp : SMTBVOp<"cmp", [Pure, SameTypeOperands]> {
...
+    This operation is a shorthand for repeated concatenation of the same
+    bit-vector value, i.e.,
+    ```mlir
+    smt.bv.repeat 5 times %a
+    // is the same as
+    %0 = smt.bv.repeat 4 times %a
+    smt.bv.concat %a, %0
+    // or also 
+    %0 = smt.bv.repeat 4 times %a
+    smt.bv.concat %0, %a
```
math-fehr, line 217: The types are missing here

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -146,4 +146,91 @@ def BVCmpOp : SMTBVOp<"cmp", [Pure, SameTypeOperands]> {
...
+    
+    The semantics are equivalent to the `repeat` operator defined in the SMT-LIB
+    2.6 standard. More precisely in the
+    [theory of FixedSizeBitVectors](https://smtlib.cs.uiowa.edu/Theories/FixedSizeBitVectors.smt2)
+    and the [QF_BV logic](https://smtlib.cs.uiowa.edu/Logics/QF_BV.smt2)
+    describing closed quantifier-free formulas over the theory of fixed-size
+    bit-vectors.
+  }];
+
+  let arguments = (ins I32Attr:$count, BitVectorType:$input);
```
math-fehr, line 228: Note, you don't actually need `count`, as you can derive it from the input/output type directly.
But this would not allow you to use the declarative format, so both should be fine!
But I guess you would want to have a builder that doesn't need this integer at least!

math-fehr (approved):

include/circt/Dialect/SMT/SMTBitVectorOps.td
maerhart: Good point! We could do it in the declarative format if we only print the input and result types, but that makes it more difficult to tell how often it is repeated, which is probably the primary information you want to get looking at the op, so I just wrote a custom parser.

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -146,4 +146,96 @@ def BVCmpOp : SMTBVOp<"cmp", [Pure, SameTypeOperands]> {
...
+    semantics are equivalent to the `extract` operator defined in the SMT-LIB
+    2.6 standard. More precisely in the
+    [theory of FixedSizeBitVectors](https://smtlib.cs.uiowa.edu/Theories/FixedSizeBitVectors.smt2)
+    and the [QF_BV logic](https://smtlib.cs.uiowa.edu/Logics/QF_BV.smt2)
+    describing closed quantifier-free formulas over the theory of fixed-size
+    bit-vectors.
+
+    Note that `smt.bv.extract %bv from 2 : (!smt.bv<32>) -> !smt.bv<16>` is
+    equivalent to `((_ extract 17 2) bv)`, i.e., the SMT-LIB operator takes the
+    low and high indicies where both are inclusive. The following equivalence
```
TaoBi22, line 188: Nit: indicies -> indices

TaoBi22, line 188: Is the difference in operand meaning here just to stay more in line with existing extract ops?

test/Dialect/SMT/bitvector-errors.mlir
```diff
@@ -34,3 +34,63 @@ func.func @invalid_bitvector_attr() {
...
+  smt.bv.repeat -2 times %arg0 : !smt.bv<32>
+  return
+}
+
+// -----
+
+// The parser has to extract the bit-width of the input and thus we need to
+// test that this is handled correctly in the parser, we cannot just rely on the
+// verifier.
+func.func @repeat_wrong_input_type(%arg0: !smt.bool) {
```
TaoBi22, line 76: Nice that there's a test for this :)

TaoBi22 (approved):
LGTM!

include/circt/Dialect/SMT/SMTBitVectorOps.td
maerhart: We don't need an attribute for both the high and low index because that would be redundant information together with the result type. I chose the low bit because that's what the comb dialect does, so the lowering is more straight forward.

PR #6814, opened 2024-03-12
[SMT] Add boolean logic operations

maerhart:


include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -222,4 +222,64 @@ def DistinctOp : SMTOp<"distinct", [Pure, SameTypeOperands]> {
   let hasVerifier = true;
 }
 
+def IteOp : SMTOp<"ite", [
+  Pure,
+  AllTypesMatch<["thenValue", "elseValue", "result"]>
+]> {
+  let summary = "a if-then-else function";
```
fabianschuiki, line 229:
```suggestion
  let summary = "an if-then-else function";
```

fabianschuiki (approved):
Neat!

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -222,4 +222,64 @@ def DistinctOp : SMTOp<"distinct", [Pure, SameTypeOperands]> {
...
+    of the SMT-LIB Standard 2.6.
+    }] # explanation;
+
+  let arguments = (ins Variadic<BoolType>:$inputs);
+  let results = (outs BoolType:$result);
+
+  let assemblyFormat = "$inputs attr-dict";
+}
+
+def NotOp : BoolOp<"not", "a boolean negation"> {
```
math-fehr, line 267: To me, this feels weird that this inherit `BoolOp`.
At least, it doesn't feel obvious what aspects of `BoolOp` is actually inherited, since you override the arguments and assemblyFormat.
So I would just inherit from SMTOp directly!

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -222,4 +222,64 @@ def DistinctOp : SMTOp<"distinct", [Pure, SameTypeOperands]> {
...
+def OrOp : BoolOp<"or", "a boolean disjunction">;
+def XOrOp : BoolOp<"xor", "a boolean exclusive OR">;
+
+def ImpliesOp : BoolOp<"implies", "boolean implication"> {
+  let explanation = [{
+    It supports a variadic number of operands, but requires at least two.
+    This is because the operator is annotated with the `:right-assoc` attribute
+    which means that `op a b c` is equivalent to `(op a (op b c))`.
+  }];
+}
```
math-fehr, line 283: For this operation, I'm actually thinking if we should encode it as a variadic or not.
While for `and`, `or`, and `xor` it makes sense as the order does not matter, here the variadicity is just syntactic sugar for a right associative operation.
So to me, it would be better if this was only a binary operation, which is easier to reason about.

math-fehr (approved):
Nice!
I'm just disagreeing a bit on how implies should be encoded, but otherwise it makes sense!

PR #6817, opened 2024-03-13
[FIRRTL] Support alternative base paths in LowerClasses.

mikeurbach:
We previously had a constraint that paths are targeting entities in the same owning module as the path. This allowed us to assume that a single base path can be created for each owning module, and used in all paths. However, we have transforms that extract entities targeted by paths outside the paths' owning module.

To support this, we can instead plumb through base paths created higher up in the instance graph down to the paths that need them. When a path is detected outside the owning module, we find the module in which the entity is instantiated and pass through its base path to the paths that reference the entity.

Most of the work here is just plumbing, and the PathInfoTable has been extended with new data structures and functions to support this.


mikeurbach:
Hmm seems like some minor issue that I didn't see while working on this with an older LLVM version, I'll fix that.

mikeurbach:
I don't know what's going on... I didn't see these errors locally 🤦

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -51,12 +54,84 @@ struct PathInfo {
...
+  llvm::iterator_range<const StringAttr *>
+  getRootsForPassthrough(StringAttr passthroughModuleName) const {
+    auto it = altBasePathsPassthroughs.find(passthroughModuleName);
+    assert(it != altBasePathsPassthroughs.end() &&
+           "expected passthrough module to already exist");
+    return llvm::make_range(it->second.begin(), it->second.end());
+  }
+
+  // Collect alternative base paths passing through `instance`, by looking up
+  // its associated `moduleNameAttr`. The results are colelcted in `result`.
```
rwy7, line 102:
```suggestion
  // its associated `moduleNameAttr`. The results are collected in `result`.
```

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -51,12 +54,84 @@ struct PathInfo {
...
+
+  /// The module name of the root module from which we take an alternative base
+  /// path.
+  StringAttr altBasePathModule = nullptr;
 };
 
 /// Maps a FIRRTL path id to the lowered PathInfo.
 struct PathInfoTable {
+  // Add an alternative base path root module. The default base path from this
+  // module with be passed through to where it is needed.
```
prithayan, line 66:
```suggestion
  // module will be passed through to where it is needed.
```

prithayan (approved):
Looks good.

PR #6818, opened 2024-03-13
[InferRW] Remove dependence of write-mode on enable for memory

prithayan:
The read-write port of a memory has an input enable `en` signal and a write-mode `wmode` signal. The `en` signal enables the port for either read/write mode. The `wmode` is enabled to set the memory in the write mode and disabled to read from the memory.
This PR removes the dependency of `wmode` on the `en` signal of a read-write port of a memory.
The `wmode` signal matters only if the `en` is enabled, hence traverse the expression tree for the `wmode` and replace `en` with a constant `1`.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
@@ -363,9 +363,12 @@ private:
       return defOp;
     };
     SmallVector<Value> stack;
+    llvm::SmallDenseSet<Value> visited;
     stack.push_back(wmodeDriver);
     while (!stack.empty()) {
       auto driver = stack.pop_back_val();
+      if (!visited.insert(driver).second)
+        continue;
       auto *defOp = getDriverOp(driver);
       if (!defOp)
         continue;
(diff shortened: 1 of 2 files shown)


prithayan:
@seldridge , @debs-sifive , I checked that this PR fixes the issue with the redundant `AND` for write-mode with the enable. Its ready for review.

lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
```diff
@@ -312,6 +314,70 @@ struct InferReadWritePass : public InferReadWriteBase<InferReadWritePass> {
...
+    while (!stack.empty()) {
+      auto driver = stack.pop_back_val();
+      auto *defOp = getDriverOp(driver);
+      if (!defOp)
+        continue;
+      for (auto operand : llvm::enumerate(defOp->getOperands())) {
+        if (operand.value() == enableDriver)
+          defOp->setOperand(operand.index(), constOne);
+        else
+          stack.push_back(operand.value());
```
mikeurbach, line 376: This is safe because we can have no cycles in FIRRTL? This is just doing a backward dataflow slice right?

mikeurbach (commented):
I don't know much about this pass, or if there is a better way to solve the problem at a high level. The implementation seems fine to me.

lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
prithayan: Right, we don't expect any cycles here. So I avoided a check for visited set, but I think for clarity I will add a visited set check here.

prithayan:
Merged this now for the next release, post merge comments are welcome.

PR #4656, opened 2023-02-14
[LowerAnnotations] Allow wiring type-equivalent types.

dtzSiFive:
Loosen the wiring check: types must be FIRRTL types, and either identical or type-equivalent.

Fixes #4651.

PR #6822, opened 2024-03-13
[FIRRTL][LowerAnnotations] Fix non-probe type compat check.

### LINKED ISSUE #6819: Wiring problem error with const source and non-const sink w/no-ref-type option
Consider:
```
FIRRTL version 4.0.0

circuit ZW: %[[
  {
    "class": "sifive.enterprise.grandcentral.DataTapsAnnotation",
    "keys": [
      {
        "class": "sifive.enterprise.grandcentral.ReferenceDataTapKey",
        "source": "~ZW|ZW>in",
        "sink": "~ZW|ZW/c:Child>sink"
      }
    ]
  }
]]
  module Child:
    wire sink : UInt<0>
    invalidate sink
 

  public module ZW:
    input in : const UInt<0>

    inst c of Child
```

Built with `firtool  --lower-annotations-no-ref-type-ports` this produces an error:

```
tap_zw_const.fir:21:11: error: Wiring Problem source type '!firrtl.const.uint<0>' does not match sink type '!firrtl.uint<0>'
    input in : const UInt<0>
          ^
tap_zw_const.fir:16:5: note: The sink is here.
    wire sink : UInt<0>
    ^
```

cc #4651.

dtzSiFive:
Add test as "legacy wiring" as non-ref-type-port path is a hidden option.

Compatibility dest/source operands were backwards, introduced in #4656 .

Fixes #6819.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
@@ -858,7 +858,7 @@ LogicalResult LowerAnnotationsPass::solveWiringProblems(ApplyState &state) {
       // Otherwise they must be identical or FIRRTL type-equivalent
       // (connectable).
       if (sourceFType != sinkFType &&
-          !areTypesEquivalent(sourceFType, sinkFType)) {
+          !areTypesEquivalent(sinkFType, sourceFType)) {
         auto diag = mlir::emitError(source.getLoc())
                     << "Wiring Problem source type " << sourceType
                     << " does not match sink type " << sinkType;
--- a/test/Dialect/FIRRTL/legacy-wiring.mlir
+++ b/test/Dialect/FIRRTL/legacy-wiring.mlir
@@ -206,3 +206,34 @@ firrtl.circuit "IntWidths" attributes {
     firrtl.connect %x, %invalid_ui1 : !firrtl.uint, !firrtl.uint
   }
 }
+
+// -----
+
+// Check direction of compatibility using const/non-const issue encountered (#6819).
+firrtl.circuit "Issue6819" attributes {
+ rawAnnotations = [
+  {
+    class = "firrtl.passes.wiring.SourceAnnotation",
+    target = "~Issue6819|Bar>y",
+    pin = "xyz"
+  },
+  {
+    class = "firrtl.passes.wiring.SinkAnnotation",
+    target = "~Issue6819|Issue6819>x",
+    pin = "xyz"
+  }
+  ]} {
+  firrtl.module private @Bar() {
+    %y = firrtl.wire interesting_name : !firrtl.const.uint<4>
+    %invalid = firrtl.invalidvalue : !firrtl.const.uint<4>
+    firrtl.strictconnect %y, %invalid : !firrtl.const.uint<4>
+  }
+  // CHECK-LABEL: module @Issue6819
+  firrtl.module @Issue6819() {
+    // CHECK: firrtl.connect %x, %{{[^ ]*}} : !firrtl.uint, !firrtl.const.uint<4>
+    firrtl.instance bar interesting_name @Bar()
+    %x = firrtl.wire interesting_name : !firrtl.uint
+    %invalid_ui1 = firrtl.invalidvalue : !firrtl.uint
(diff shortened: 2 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
```diff
@@ -858,7 +858,7 @@ LogicalResult LowerAnnotationsPass::solveWiringProblems(ApplyState &state) {
       // Otherwise they must be identical or FIRRTL type-equivalent
       // (connectable).
       if (sourceFType != sinkFType &&
-          !areTypesEquivalent(sourceFType, sinkFType)) {
+          !areTypesEquivalent(sinkFType, sourceFType)) {
```
seldridge, line 861: Type equivalence is not commutative? This seems like an issue with the type equivalence check?

seldridge (commented):
Confused by the fix here. However, the test looks great.

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
dtzSiFive: LOL yeah it's not symmetric and therefore also not an equivalence relation, possibly for other reasons I'm forgetting.  This is @mmaloney-sf's favorite part about it :upside_down_face: :wink: .

dtzSiFive: To be clear this is the FIRRTL spec's notion of "type equivalence", I don't think it's bugged (not for lack of being an equivalence anway) but at a very-quick glance it doesn't mention const there presently, although it's still not symmetric (Reset can connect to UInt, can't connect UInt to Reset).

seldridge: Good point about resets which are obviously not commutative.

I think the problem here is more that the spec is presenting this as an "equivalence" when it's really "verification of connection source and destination". It's not inherently wrong that equality isn't commutative, it's just weird.

dtzSiFive: Right, shouldn't be calling its connectivity rules (or whatever) equivalence or equality probably as that's misleading if nothing else.  I haven't checked but can't help but wonder if it was originally a proper equivalence and then drifted or something.  Welp.

Anyway, mind approving the PR? :+1:

seldridge: I think it was before abstract reset. The spec was always written as something like "the source and destination must be equivalent types (see type equivalence section for a definition)." Notably, I don't think that "type equivalence" actually mattered except for connects. I.e., this was always that aforementioned connect verification. We should probably just roll it into the connect section.

seldridge (approved):

test/Dialect/FIRRTL/legacy-wiring.mlir
```diff
@@ -206,3 +206,34 @@ firrtl.circuit "IntWidths" attributes {
...
+    firrtl.strictconnect %y, %invalid : !firrtl.const.uint<4>
+  }
+  // CHECK-LABEL: module @Issue6819
+  firrtl.module @Issue6819() {
+    // CHECK: firrtl.connect %x, %{{[^ ]*}} : !firrtl.uint, !firrtl.const.uint<4>
+    firrtl.instance bar interesting_name @Bar()
+    %x = firrtl.wire interesting_name : !firrtl.uint
+    %invalid_ui1 = firrtl.invalidvalue : !firrtl.uint
+    firrtl.connect %x, %invalid_ui1 : !firrtl.uint, !firrtl.uint
+  }
```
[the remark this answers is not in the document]
dtzSiFive: Went to do this but it triggers the special wire2node behavior added for GCView pipeline assumption reasons that I think makes it less obvious what the test is about (what happened to the wire? why? Is that relevant to whats being tested here?) and couples the test to that behavior unnecessarily.

Once found myself crafting a comment explaining why it becomes a node (and possibly what to expect in practice/otherwise) seems like maybe keeping it as-is is better as it's less surprising and doesn't need explanation.

I'll drop the other first "invalidate" to trim it a bit as you suggest.

PR #6823, opened 2024-03-13
[FIRRTL][CheckCombLoops] Add "xfail" tests for simple rwprobeop loops.

dtzSiFive:
Add some cycles involving rwprobeop as tests that check we don't emit any diagnostics.

(as suggested here: https://github.com/llvm/circt/pull/6821#discussion_r1523766869)


prithayan:
Thanks for tests, Included them, https://github.com/llvm/circt/pull/6824

dtzSiFive:
Awesome, thanks! Closing.

Issue #6820, opened 2024-03-13
[FIRRTL][CheckCombLoops] Missing RWProbeOp support, crashes


dtzSiFive:
Example FIRRTL:
```
FIRRTL version 3.3.0
circuit Bar :
  module Foo :
    input clock : Clock
    output clockProbe_bore : RWProbe<Clock>

    define clockProbe_bore = rwprobe(clock)

  module Bar :
    input clock : Clock
    output clockProbe : RWProbe<Clock>

    inst foo of Foo
    connect foo.clock, clock
    define clockProbe = foo.clockProbe_bore

    force(clock, UInt<1>(1), clockProbe, asClock(UInt<1>(0)))
```

`--parse-only`:
```
firrtl.circuit "Bar" {
  firrtl.module private @Foo(in %clock: !firrtl.clock sym @sym, out %clockProbe_bore: !firrtl.rwprobe<clock>) {
    %0 = firrtl.ref.rwprobe <@Foo::@sym> : !firrtl.rwprobe<clock>
    firrtl.ref.define %clockProbe_bore, %0 : !firrtl.rwprobe<clock>
  }
  firrtl.module @Bar(in %clock: !firrtl.clock, out %clockProbe: !firrtl.rwprobe<clock>) attributes {convention = #firrtl<convention scalarized>} {
    %foo_clock, %foo_clockProbe_bore = firrtl.instance foo @Foo(in clock: !firrtl.clock, out clockProbe_bore: !firrtl.rwprobe<clock>)
    firrtl.strictconnect %foo_clock, %clock : !firrtl.clock
    firrtl.ref.define %clockProbe, %foo_clockProbe_bore : !firrtl.rwprobe<clock>
    %c1_ui1 = firrtl.constant 1 : !firrtl.uint<1>
    %c0_ui1 = firrtl.constant 0 : !firrtl.uint<1>
    %0 = firrtl.asClock %c0_ui1 : (!firrtl.uint<1>) -> !firrtl.clock
    firrtl.ref.force %clock, %c1_ui1, %clockProbe, %0 : !firrtl.clock, !firrtl.uint<1>, !firrtl.clock
  }
}
```

Crashes (or assert failure) when trying to resolve the force.

Flagged in review: https://github.com/llvm/circt/pull/5647#discussion_r1362126296 .

dtzSiFive:
[...]

PR #6824, opened 2024-03-13
[CheckCombLoops] Handle RWProbeOp

prithayan:
Handle the `RWProbeOp` in `CheckCombLoop`. 
This fixes https://github.com/llvm/circt/issues/6820

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/CheckCombLoops.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CheckCombLoops.cpp
@@ -37,6 +37,7 @@
 #include "circt/Dialect/FIRRTL/FIRRTLUtils.h"
 #include "circt/Dialect/FIRRTL/FIRRTLVisitors.h"
 #include "circt/Dialect/FIRRTL/Passes.h"
+#include "circt/Dialect/HW/InnerSymbolTable.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/EquivalenceClasses.h"
 #include "llvm/ADT/PostOrderIterator.h"
@@ -63,9 +64,11 @@ public:
   DiscoverLoops(
       FModuleOp module, InstanceGraph &instanceGraph,
       const DenseMap<FModuleLike, DrivenBysMapType> &otherModulePortPaths,
-      DrivenBysMapType &thisModulePortPaths)
+      DrivenBysMapType &thisModulePortPaths, SymbolTable &symtbl,
+      hw::InnerSymbolTableCollection &istc)
       : module(module), instanceGraph(instanceGraph),
-        modulePortPaths(otherModulePortPaths), portPaths(thisModulePortPaths) {}
+        modulePortPaths(otherModulePortPaths),
+        portPaths(thisModulePortPaths), irn{symtbl, istc} {}
 
   LogicalResult processModule() {
     LLVM_DEBUG(llvm::dbgs() << "\n processing module :" << module.getName());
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/CheckCombLoops.cpp
```diff
@@ -103,6 +106,19 @@ class DiscoverLoops {
...
+            hw::InnerSymTarget refTarget = irn.lookup(rwProbe.getTarget());
+            if (!refTarget) {
+              rwProbe->emitWarning("cannot resolve target innerRef");
+              return;
+            }
+            FieldRef ref = getRefForIST(refTarget);
+            if (!ref)
+              return;
+            auto data = rwProbe.getResult();
+            addDrivenBy({data, 0}, ref);
```
dtzSiFive, line 119: Does this work if the rwprobe target isn't a leaf (multiple FieldRef's)?

(not sure how this works, but do we create edge for all fields connected?)

lib/Dialect/FIRRTL/Transforms/CheckCombLoops.cpp
```diff
@@ -421,6 +436,9 @@ class DiscoverLoops {
               continue;
             }
         }
+        if (!type_isa<RefType>(srcArgType) && sinkPortIsForceable)
+          recordProbe(srcPort, sinkPort);
```
dtzSiFive, line 440: I believe you but if you could spare a word for what this does/why that'd be swell.  Thank you!

If there's a path from a non-probe to a rwprobe, record a probe edge sink -> source? I'm not sure that's right, consider:

```
FIRRTL version 3.3.0
circuit Bar :
  module Foo :
    input clock : Clock
    output out : Clock
    output clockProbe_bore : RWProbe<Clock>

    node n = clock

    define clockProbe_bore = rwprobe(n)
    connect out, clock

  module Bar :
    input clock : Clock
    output clockProbe : RWProbe<Clock>

    inst foo of Foo
    connect foo.clock, clock
    define clockProbe = foo.clockProbe_bore

    force(clock, UInt<1>(1), clockProbe, foo.out)
```

Which reports a cycle but there isn't one, forcing `n` is unused.

lib/Dialect/FIRRTL/Transforms/CheckCombLoops.cpp
dtzSiFive: Consider:
```
FIRRTL version 3.3.0
circuit Bar :
  module Foo :
    input agg : { x: Clock, y : Clock }
    output ref : RWProbe<{ x: Clock, y : Clock }>
    output refx : RWProbe<Clock>

    define ref = rwprobe(agg)
    define refx = rwprobe(agg.x)

  module Bar :
    input clock : Clock

    inst foo of Foo
    connect foo.agg.x, clock
    connect foo.agg.y, clock

    force(clock, UInt<1>(1), foo.ref.x, read(foo.refx))
```

Which should report a cycle but doesn't.

dtzSiFive (commented):
Thanks!! Glad this generally doesn't look like a big change.

Unsure of some things, gotta run but will circle back.

Please take a look at the examples posted.

lib/Dialect/FIRRTL/Transforms/CheckCombLoops.cpp
prithayan: Thanks for spotting this, I need to fix the logic to track which ports can be updated by a probe. Working on it.

PR #6826, opened 2024-03-14
[Emit][Seq] Emit random init headers using fragments

nandor:
The verbatims and SV macros are grouped into fragments now. As a beneficial side effect, a module is prefixed with fragments in its file only when it is actually using the fragments, reducing output size.

### DIFF
--- a/lib/Conversion/SeqToSV/SeqToSV.cpp
+++ b/lib/Conversion/SeqToSV/SeqToSV.cpp
@@ -15,6 +15,7 @@
 #include "FirMemLowering.h"
 #include "FirRegLowering.h"
 #include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/Emit/EmitOps.h"
 #include "circt/Dialect/HW/ConversionPatterns.h"
 #include "circt/Dialect/HW/HWAttributes.h"
 #include "circt/Dialect/HW/HWOps.h"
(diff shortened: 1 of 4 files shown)


lib/Conversion/SeqToSV/SeqToSV.cpp
```diff
@@ -528,60 +561,72 @@ void SeqToSVPass::runOnOperation() {
...
+    emitGuard("INIT_RANDOM_PROLOG_", [&]() {
+      b.create<sv::IfDefOp>(
+          "RANDOMIZE",
+          [&]() {
+            emitGuardedDefine("VERILATOR", "INIT_RANDOM_PROLOG_",
+                              "`INIT_RANDOM",
+                              "`INIT_RANDOM #`RANDOMIZE_DELAY begin end");
+          },
+          [&]() { b.create<sv::MacroDefOp>("INIT_RANDOM_PROLOG_", ""); });
+    });
```
fabianschuiki, line 598: Should this check if the fragments already exist, and only add them if they don't? Otherwise you can't pass MLIR that already has run through SeqToSV through firtool.

fabianschuiki (approved):
LGTM! This is a nice piece of cleanup.

lib/Conversion/SeqToSV/SeqToSV.cpp
```diff
@@ -404,6 +405,32 @@ void SeqToSVPass::runOnOperation() {
...
+  auto randomInitMemFragmentName =
+      FlatSymbolRefAttr::get(context, "RANDOM_INIT_MEM_FRAGMENT");
+
+  auto regFragments = ArrayAttr::get(
+      context, {randomInitFragmentName, randomInitRegFragmentName});
+  auto memFragments = ArrayAttr::get(
+      context, {randomInitFragmentName, randomInitMemFragmentName});
+
+  auto addFragments = [&](HWModuleOp module, ArrayAttr fragments) {
+    constexpr StringRef attrName = "emit.fragments";
```
prithayan, line 421: Nit: How about adding it to the headers and using something like, `getFragmentsAttrName()` , instead of hardcoding the attrname in different files.

prithayan (approved):
Looks good.

lib/Conversion/SeqToSV/SeqToSV.cpp
nandor: That was broken, will stay broken.

PR #6827, opened 2024-03-14
[SMT] Add basic array operations

maerhart:


TaoBi22 (approved):
LGTM! Would it maybe be worth having verifiers to make sure operands match the Array domain/range types where appropriate (e.g. index is of domain type in a store/select op)?

fabianschuiki (approved):
LGTM! Whoop whoop array theory!

maerhart:
> Would it maybe be worth having verifiers to make sure operands match the Array domain/range types where appropriate (e.g. index is of domain type in a store/select op)?

That's already done by the `TypesMatchWith` traits added in the ODS file.

PR #6828, opened 2024-03-14
[FSM][Emit] Convert the FSMToSV pass to use `emit` ops

nandor:
Instead of relying on replicated ops, the pass now directly sets up its file structure using `emit` ops. As a side effect, the type scope file is no longer included in the default file list as that one is only going to include SV files generated from `hw.module`.


fabianschuiki (approved):
I'm not too familiar with the FSM dialect, but the cleanup and emit dialect use look good!

PR #6832, opened 2024-03-14
[Verif] Add LEC operation

maerhart:
Add a first implementation of a LEC operation. This has to be refined as we go forward, e.g., to support reporting of counter-examples. An operation to express a LEC statement explicitly in the IR at the CIRCT Core level allows us to state the general intent to perform LEC without committing to one specific approach/backend yet.

This PR also already adds one such lowering (constructing a simple miter circuit encoded in SMT). A followup PR will introduce a separate pass that takes two `hw.module`s and transforms them into one `verif.lec` operation.


math-fehr (approved):
I'm definitely not familiar with miter circuits, so I'm approving it without knowing much what it does ;)

TaoBi22 (approved):
This all looks great to me!

PR #6833, opened 2024-03-15
[circt-lec] Add ConstructLEC pass

maerhart:
Once this is merged and the lowering to LLVM is added, we can port the `circt-lec` tool over to a pass pipeline + JIT and remove the existing internals + the link dependency on Z3.


lib/Tools/circt-lec/ConstructLEC.cpp
```diff
@@ -0,0 +1,140 @@
...
+  // Lookup the modules.
+  auto moduleA = lookupModule(firstModule);
+  if (!moduleA)
+    return signalPassFailure();
+  auto moduleB = lookupModule(secondModule);
+  if (!moduleB)
+    return signalPassFailure();
+
+  if (moduleA.getModuleType() != moduleB.getModuleType()) {
+    moduleA.emitError("module's IO types don't match second module's: ")
```
TaoBi22, line 87: Nit: ->`module's IO types don't match second modules:`

lib/Tools/circt-lec/ConstructLEC.cpp
```diff
@@ -0,0 +1,140 @@
...
+    builder.createBlock(&mainFunc.getBody(), {}, {i32Ty, ptrTy}, {loc, loc});
+    builder.create<func::CallOp>(loc, entryFunc, ValueRange{});
+    // TODO: don't use LLVM here
+    Value constZero = builder.create<LLVM::ConstantOp>(loc, i32Ty, 0);
+    builder.create<func::ReturnOp>(loc, constZero);
+  }
+
+  builder.createBlock(&entryFunc.getBody());
+  auto lecOp = builder.create<verif::LogicEquivalenceCheckingOp>(loc);
+  auto *yieldA = moduleA.getBodyBlock()->getTerminator();
```
TaoBi22, line 112: Nit: Happy to be convinced otherwise but I think yieldA could be a slightly confusing name here, when you're creating yield ops to yield A and B further down - maybe something that makes it clearer that they're the original output ops?

include/circt/Tools/circt-lec/Passes.td
```diff
@@ -0,0 +1,53 @@
...
+
+include "mlir/Pass/PassBase.td"
+include "mlir/Rewrite/PassUtil.td"
+
+def ConstructLEC : Pass<"construct-lec", "::mlir::ModuleOp"> {
+  let summary = "Lower CIRCTs core dialects to a LEC problem statement";
+  let description = [{
+    Lowers HW and Comb operations to the SMT and Func dialect. HW modules become
+    functions that take SMT bit-vectors as input and return SMT bit-vectors
+    according to the original ports of the module. The Comb operations in the
```
TaoBi22, line 24: It doesn't look from the tests like the Comb ops are converted to SMT by this pass?

TaoBi22 (approved):
A couple of nits but looks good to me!

include/circt/Tools/circt-lec/Passes.td
TaoBi22: Put this in an unclear place - it's addressing the sentence that runs off the end of the block GitHub previews!

maerhart: Oh, I should have noticed that. It's just the version from the experimental branch without taking the redesign into consideration. Thanks!

PR #6834, opened 2024-03-15
[FIRParser] Do not swallow stop ops when parsing FIR

nandor:


dtzSiFive (approved):
LGTM!

Does the test fail without this change? (it may)

If I'm reading this right, we just dropped stop conditions following printf's?! Wow.

nandor:
Yes. The test failed before the change.

PR #6835, opened 2024-03-16
[Ibis] Introduce 'ibis.design'

teqdruid:
Encapsulate everything into a new op. It's an InnerSymbolTable so that
we can identify _everything_ with inner symbols.

Changes required:
- Put all ibis operations in the new `ibis.design` op.
- Replace all symbols with `InnerSymAttr`s. Make sure they are unique within an `ibis.design`.
- Replace all symbol refs with `InnerRefAttr`s.


blakep-msft (approved):
LGTM

PR #6839, opened 2024-03-17
[Arc] Add basic support for assert operation.

elhewaty:
- Adds basic Assertion support to ARC
- Fixed: https://github.com/llvm/circt/issues/6810﻿

### DIFF
--- a/include/circt/Dialect/Arc/ArcOps.td
+++ b/include/circt/Dialect/Arc/ArcOps.td
@@ -800,4 +800,21 @@ def VectorizeReturnOp : ArcOp<"vectorize.return", [
   let assemblyFormat = "operands attr-dict `:` qualified(type(operands))";
 }
 
+def AssertOp : ArcOp<"assert"> {
+  let summary = "assert that a properity holds.";
+  let description = [{
+    Asserts that an operation holds at runtime, it takes a single boolean
+    as an input, and an error message.
+    If the property holds (= `true`) the operation has no effect. Otherwise, the
+    execution aborts and provide an error message to the user.
+    ```mlir
+    assert %proberty, "Some error message in case of failure!"
+    ```
+  }];
+  // I thought of making the string message optional.
+  // Also should it take clock as parameters?
+  let arguments = (ins I1: $property, StrAttr:$msg);
+  let assemblyFormat = "$property `,` $msg attr-dict";
+}
+
 #endif // CIRCT_DIALECT_ARC_ARCOPS_TD
--- a/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
+++ b/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
@@ -19,6 +19,7 @@
 #include "mlir/Conversion/FuncToLLVM/ConvertFuncToLLVM.h"
 #include "mlir/Conversion/LLVMCommon/ConversionTarget.h"
 #include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+ #include "mlir/Conversion/LLVMCommon/PrintCallHelper.h"
 #include "mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlow.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
@@ -161,6 +162,34 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
         op.getLoc(), adaptor.getStorage().getType(), rewriter.getI8Type(),
         adaptor.getStorage(), offset);
     rewriter.replaceOp(op, ptr);
+    return success();
+  }
+};
+
+struct AssertOpLowering : public OpConversionPattern<arc::AssertOp> {
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(arc::AssertOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto condition = adaptor.getProperty();
+    auto module = op->getParentOfType<ModuleOp>();
+    rewriter.replaceOpWithNewOp<scf::IfOp>(
+        op, condition, [&](auto &builder, auto loc) {
+          // If condition holds then print the message and abort
+
+          // Any hint about how to implement this??!!
+          // LLVM::createPrintStrCall(rewriter, loc, module, "assert_msg", op.getMsg(),
+          //                    *getTypeConverter(), /*addNewLine=*/false,
+          //                    /*runtimeFunctionName=*/"puts");
+          auto abortFunc = module.lookupSymbol<LLVM::LLVMFuncOp>("abort");
+          // If abort is not exist create one
+          if(!abortFunc) {
+            OpBuilder::InsertionGuard guard(rewriter);
+            rewriter.setInsertionPointToStart(module.getBody());
+            builder.template create<LLVM::LLVMFuncOp>(loc, "abort", builder.getNoneType());
+          }
+          builder.template create<LLVM::CallOp>(loc, abortFunc, std::nullopt);
+        });
     return success();
   }
 };
--- a/test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
+++ b/test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
@@ -222,3 +222,8 @@ func.func @WriteArray(%arg0: !arc.state<!hw.array<4xi1>>, %arg1: !hw.array<4xi1>
   arc.state_write %arg0 = %arg1 : <!hw.array<4xi1>>
   return
 }
+
+func.func @assert(%arg: i1) {
+  arc.assert %arg, "Oops, failed"
+  return
+}
--- a/test/Dialect/Arc/basic-errors.mlir
+++ b/test/Dialect/Arc/basic-errors.mlir
@@ -548,3 +548,10 @@ hw.module @vectorize(in %in0: i4, in %in1: i4, out out0: i4) {
 
 // expected-error @below {{state type must have a known bit width}}
 func.func @InvalidStateType(%arg0: !arc.state<index>)
+
+// -----
+
+hw.module @assert(in %in : i4) {
+  arc.assert %in, "Oops, failed"
+  hw.output
+}
--- a/test/Dialect/Arc/basic.mlir
+++ b/test/Dialect/Arc/basic.mlir
@@ -347,3 +347,10 @@ hw.module @vectorize_both_sides_lowered(in %in0: i1, in %in1: i1, in %in2: i1, i
 //       CHECK: [[V8:%.+]] = vector.extract [[V7]][0]
 //       CHECK: [[V9:%.+]] = vector.extract [[V7]][1]
 //       CHECK: hw.output [[V3]], [[V4]], [[V8]], [[V9]] :
+
+// -----
+
+hw.module @assert(in %in0: i1) {
+  arc.assert %in0, "Oops, failed"
+  hw.output
+}


elhewaty:
Aside from the conflicts, I will  resolve them later:
Can you share your opinions?
how to generate the `CHECK` tests?
for basic-error.mlir : this is the error `unexpected error: use of value '%in' expects different type than prior uses: 'i1' vs 'i4'`
if I write `expected-error @below {{use of value '%in' expects different type than prior uses: 'i1' vs 'i4'}}` why doesn't it work?

for the print message before the abort, I want to make something like `call i32 @puts(ptr @.str)`
can you give some hints about converting the message to `Value`?
I tried `LLVM::createPrintStrCall`, but it didn't work here.

test/Dialect/Arc/basic-errors.mlir
```diff
@@ -548,3 +548,10 @@ hw.module @vectorize(in %in0: i4, in %in1: i4, out out0: i4) {
 
 // expected-error @below {{state type must have a known bit width}}
 func.func @InvalidStateType(%arg0: !arc.state<index>)
+
+// -----
+
+hw.module @assert(in %in : i4) {
+  arc.assert %in, "Oops, failed"
+  hw.output
+}
```
maerhart, line 557: Please always end the file with a newline before EOF.

include/circt/Dialect/Arc/ArcOps.td
```diff
@@ -800,4 +800,21 @@ def VectorizeReturnOp : ArcOp<"vectorize.return", [
   let assemblyFormat = "operands attr-dict `:` qualified(type(operands))";
 }
 
+def AssertOp : ArcOp<"assert"> {
+  let summary = "assert that a properity holds.";
```
maerhart, line 804:
```suggestion
  let summary = "assert that a property holds.";
```

include/circt/Dialect/Arc/ArcOps.td
```diff
@@ -800,4 +800,21 @@ def VectorizeReturnOp : ArcOp<"vectorize.return", [
...
+  let description = [{
+    Asserts that an operation holds at runtime, it takes a single boolean
+    as an input, and an error message.
+    If the property holds (= `true`) the operation has no effect. Otherwise, the
+    execution aborts and provide an error message to the user.
+    ```mlir
+    assert %proberty, "Some error message in case of failure!"
+    ```
+  }];
+  // I thought of making the string message optional.
```
maerhart, line 814: Sounds reasonable since it is also optional in `verif.assert`. Would you otherwise lower it to the empty string if it is not present in `verif.assert`?

include/circt/Dialect/Arc/ArcOps.td
```diff
@@ -800,4 +800,21 @@ def VectorizeReturnOp : ArcOp<"vectorize.return", [
...
+  // Also should it take clock as parameters?
```
maerhart, line 815: In the long run, a single clock is not enough since the asserted property could reason across multiple clock domains. But for a start it's probably easiest to add a single clock operand (also easier than having no clock).

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -165,6 +166,34 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
 
+struct AssertOpLowering : public OpConversionPattern<arc::AssertOp> {
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(arc::AssertOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto condition = adaptor.getProperty();
+    auto module = op->getParentOfType<ModuleOp>();
+    rewriter.replaceOpWithNewOp<scf::IfOp>(
+        op, condition, [&](auto &builder, auto loc) {
```
maerhart, line 177: You probably intended to invert `condition` before passing it to the `scf.if` because the message should be printed if the assertion does not hold.

maerhart:
> how to generate the `CHECK` tests?

You typically write them by hand. You could, however, run the `RUN` command at the start of the file in the CLI, copy-paste it to the test file, and insert the `CHECK` prefixes, etc to save some typing. Don't forget to double check if the output actually matches what you expect.

> for basic-error.mlir : this is the error `unexpected error: use of value '%in' expects different type than prior uses: 'i1' vs 'i4'` if I write `expected-error @below {{use of value '%in' expects different type than prior uses: 'i1' vs 'i4'}}` why doesn't it work?

Did you check the location of the error and insert the `expected-error` directly at the line before?

> for the print message before the abort, I want to make something like `call i32 @puts(ptr @.str)` can you give some hints about converting the message to `Value`? I tried `LLVM::createPrintStrCall`, but it didn't work here.

Can you give more detail about why `createPrintStrCall` didn't work?
The rough procedure is: create a global constant for the string, insert the `puts` function declaration, insert a 'addressof' to get a pointer to the global, pass its result to a call to `puts`.
You could take a look at the lowering pattern for `arc.sim.emit` for an example (it uses `printf` instead though).

seldridge:
> how to generate the CHECK tests?

There's a bit of an art to this. You're trying to balance:

1. Exactness/terseness of the test. Try to keep the test simple and only check what the test needs to test not unrelated features or those tested elsewhere.
2. Robustness of the test. Ideally, the test should only fail for patches which actually cause the test to fail, not due to name changes, SSA value changes, or later IR additions. This is not possible to guarantee always. That's fine.
3. Readability of the test. Hopefully, someone reading the test (or your future self) will understand what it is trying to test. Striving for (1) helps here. Comments may be necessary to explain what is going on.

Generally, your tests look good!

Usually the only thing to avoid is exact copy-pasting of output.  As Martin suggests, this can be a great starting point. Sometimes it is a useful exercise to intentionally write the test from scratch to make sure you aren't inadvertently relying on incorrect output.

elhewaty:
I think we need to implement like llvm `update_test_checks.py` for circt-opt.

elhewaty:
@fabianschuiki, @maerhart ping :bell: !!

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,73 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
 
+struct AssertOpLowering : public OpConversionPattern<arc::AssertOp> {
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(arc::AssertOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+
+    rewriter.replaceOpWithNewOp<scf::IfOp>(
+        op, adaptor.getProperty(),
+        [&](auto &builder, auto loc) {
```
maerhart, line 184: I'd use the concrete types here and below rather than `auto`. MSVC sometimes has issues with that. Not sure if that's also the case here. But it also improves readability.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,73 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+          // if property is true then assert will not execute
+          builder.template create<scf::YieldOp>(loc);
```
maerhart, line 186: Why not negate the property and only have a then-branch?

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,73 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+        },
+        [&](auto &builder, auto loc) {
+          // If the condition doesn't hold then print the message and abort
+          auto message = op.getMsg().has_value() ? op.getMsg().value() : "";
```
maerhart, line 190: I think there exists a member function of `optional` called `value_or` which would simplify this.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,73 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+          auto module = op->getParentOfType<ModuleOp>();
+
+          LLVM::LLVMFuncOp putsFunc;
+          LLVM::GlobalOp globalOp;
+
+          if (!message.empty()) {
```
maerhart, line 196: I think the `AssertOp` should have a verifier checking that if a message is given, it is not the empty string, otherwise we could just make it a default-valued attribute. This could then slightly simplify this lowering.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,73 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+            Type int8Type = IntegerType::get(op.getContext(), 8);
+
+            // Create a global constant with the message.
+            globalOp = rewriter.create<LLVM::GlobalOp>(
+                loc, LLVM::LLVMArrayType::get(int8Type, message.size()),
+                /*isConstant=*/true, LLVM::Linkage::Internal, "message",
+                rewriter.getStringAttr(msgRef));
+          }
+
+          if (!message.empty()) {
```
maerhart, line 216: Instead of having another if-statement with the same condition, you could just put the insertion guard in a nested scope.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,73 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+          if (!abortFunc) {
+            // if there's no abort create it
+            OpBuilder::InsertionGuard guard(rewriter);
+            rewriter.setInsertionPointToStart(module.getBody());
+            auto abortFuncTy = LLVM::LLVMFunctionType::get(
+                LLVM::LLVMVoidType::get(getContext()), {});
+            abortFunc =
+                rewriter.create<LLVM::LLVMFuncOp>(loc, "abort", abortFuncTy);
+          }
+          builder.template create<LLVM::CallOp>(loc, abortFunc, std::nullopt);
```
maerhart, line 235: There exists a utility header (`FunctionCallUtils`) in the LLVM dialect with a function `lookupOrCreateFn`. I think it does exactly that, or am I missing something?

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
```diff
@@ -222,3 +222,56 @@ func.func @WriteArray(%arg0: !arc.state<!hw.array<4xi1>>, %arg1: !hw.array<4xi1>
   arc.state_write %arg0 = %arg1 : <!hw.array<4xi1>>
   return
 }
+
+func.func @assert_with_cond() {
+  // CHECK-LABEL: @assert_with_cond(
+  // CHECK-NEXT:    [[COND:%.+]] = llvm.mlir.constant(false) : i1
+  // CHECK-NEXT:    llvm.cond_br [[COND]], ^[[TRUE_BRANCH:[[:alnum:]]+]], ^[[FALSE_BRANCH:[[:alnum:]]+]]
```
maerhart, line 229: Never seen `[[:alnum:]]` before. What does it do? Is it the same as `[0-9]`?

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
```diff
@@ -222,3 +222,56 @@ func.func @WriteArray(%arg0: !arc.state<!hw.array<4xi1>>, %arg1: !hw.array<4xi1>
...
+  // CHECK-NEXT:      llvm.br ^[[END:[[:alnum:]]+]]
+  // CHECK-NEXT:    ^[[FALSE_BRANCH]]:
+  // CHECK-NEXT:      llvm.call @abort() : () -> ()
+  // CHECK-NEXT:      llvm.br ^[[END:[[:alnum:]]+]]
+  // CHECK-NEXT:    ^[[END]]:
+  // CHECK-NEXT:      llvm.return
+  // CHECK-NEXT:  }
+  %cond = hw.constant 0 : i1
+  %clock = hw.constant 1: i1
+  arc.assert %cond, %clock
```
maerhart, line 258: I think at this stage of the Arc lowering pipeline, the `clock` operand should not be present anymore. Either it is folded into the `cond` operand or it is gone because the assert is in a function that is called only at the posedge of this clock. So the lowering pass to LLVM would just error out. But the clock operand will be useful in the lowering from `verif` and be dealt with in the passes between those two lowerings.
Also, shouldn't the clock operand be of `seq.clock` type?
@fabianschuiki What do you think?

test/Dialect/Arc/basic-errors.mlir
```diff
@@ -548,3 +548,12 @@ hw.module @vectorize(in %in0: i4, in %in1: i4, out out0: i4) {
 
 // expected-error @below {{state type must have a known bit width}}
 func.func @InvalidStateType(%arg0: !arc.state<index>)
+
+// -----
+
+// expected-note @below {{prior use here}}
+hw.module @assert(in %in : i4) {
+  // expected-error @below {{use of value '%in' expects different type than prior uses: 'i1' vs 'i4'}}
+  arc.assert %in, hw.constant true
```
maerhart, line 557: What does the `hw.constant true` do here? Can we remove it?

include/circt/Dialect/Arc/ArcOps.td
```diff
@@ -880,4 +880,22 @@ def VectorizeReturnOp : ArcOp<"vectorize.return", [
...
+  let description = [{
+    Asserts that an operation holds at runtime, it takes a single boolean
+    as an input, and an error message.
+    If the property holds (= `true`) the operation has no effect. Otherwise, the
+    execution aborts and provide an error message to the user.
+    ```mlir
+    assert %proberty, %clock, "Some error message in case of failure!"
+    ```
+  }];
+  let arguments = (ins I1:$property, Optional<I1>:$clock,
```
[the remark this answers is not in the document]
elhewaty: `verif.assert` doesn't have a clock operand. But I think I will need to give it a default value. `verif.assert 0`, has no clock operand. what do you think?

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
elhewaty: Nope, you're right. I did it just for experimental reasons and I forgot to change it. Sorry.

test/Dialect/Arc/basic-errors.mlir
elhewaty: Yes. We can. I just provided a constant `i1` parameter.

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
elhewaty: It means alpha-numeric. May be I am missing something but you will find it a lot in mlir tests https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Arith/canonicalize.mlir#L2306

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
elhewaty: I put a clock operand so we can make use of it for further steps, I am not sure why we need it in `assert` which checks the condition.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
elhewaty: I implemented it like this at first.
to negate the condition I need to `xor` it with `true`. 
It generates more operations, but maybe in some following step, the `then` branch will be removed, by some `CFG` simplifications. 

```
bool cond = 0;
assert(0);
--> 
bool cond = 0;
bool trueVal = 1;
bool invertedCond = cond ^ trueVal;
if(invertedCond) {}
--------------
bool cond = 0;
if(cond) {// do nothing}
else {}
```
I think I should negate it as you suggested, the code looks naive, and If it doesn't the xor is faster than the branch predictor anyway. I just wanted to share my thoughts.

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
maerhart: Ah, that's neat. Thanks!

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
maerhart: Yes, having the optional clock operand makes sense. But the clock operand just provides the raw value of the clock signal, but the assert is typically only checked at one or both of the clock edges (see `ltl.clock` operation). This behavior needs to be lowered and will probably happen before the ArcToLLVM pass. During this process this edge-checking logic could be folded into the condition.

elhewaty: >the clock operand should not be present anymore. Either it is folded into the cond operand or it is gone because the assert is in a function that is called only at the posedge of this clock.

@maerhart, can you explain this a little further?

maerhart:
This operation looks a lot like the one proposed in #6982
We should discuss with @fabianschuiki if the TableGen declaration should be moved to the Verif dialect, or if it makes sense to have two almost identical operations in both dialects.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,68 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
 
+struct AssertOpLowering : public OpConversionPattern<arc::AssertOp> {
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(arc::AssertOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+
+    Value negateCondition = rewriter.create<arith::XOrIOp>(
+        op.getLoc(), adaptor.getProperty(),
+        rewriter.create<arith::ConstantIntOp>(op.getLoc(), 1, 1));
```
maerhart, line 184:
```suggestion
        rewriter.create<arith::ConstantOp>(op.getLoc(), rewriter.getBoolAttr(true)));
```

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,68 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+          {
+            OpBuilder::InsertionGuard guard(rewriter);
+            rewriter.setInsertionPointToStart(module.getBody());
+            putsFunc = lookupOrCreateFn(
+                module, "puts", LLVM::LLVMPointerType::get(getContext()),
+                LLVM::LLVMVoidType::get(getContext()), false);
+            llvm::StringRef msgRef(message);
+
+            // Get the integer type with the correct bit width.
+            Type int8Type = IntegerType::get(op.getContext(), 8);
```
maerhart, line 205: Nit:
```suggestion
            Type int8Type = rewriter.getI8Type()
```

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,68 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+          LLVM::LLVMFuncOp putsFunc;
+          LLVM::GlobalOp globalOp;
+
+          {
+            OpBuilder::InsertionGuard guard(rewriter);
+            rewriter.setInsertionPointToStart(module.getBody());
+            putsFunc = lookupOrCreateFn(
+                module, "puts", LLVM::LLVMPointerType::get(getContext()),
+                LLVM::LLVMVoidType::get(getContext()), false);
+            llvm::StringRef msgRef(message);
```
maerhart, line 202: Unnecessary?

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,68 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+
+    Value negateCondition = rewriter.create<arith::XOrIOp>(
+        op.getLoc(), adaptor.getProperty(),
+        rewriter.create<arith::ConstantIntOp>(op.getLoc(), 1, 1));
+
+    rewriter.replaceOpWithNewOp<scf::IfOp>(
+        op, negateCondition,
+        [&](OpBuilder &builder, Location loc) {
+          // If the condition doesn't hold then print the message and abort
+          auto message = op.getMsg().value_or("None");
```
maerhart, line 190: If no message is given, no additional message should be appended.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,68 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+                LLVM::LLVMVoidType::get(getContext()), false);
+            llvm::StringRef msgRef(message);
+
+            // Get the integer type with the correct bit width.
+            Type int8Type = IntegerType::get(op.getContext(), 8);
+
+            // Generate new names for global strings.
+            // Messgaes with the same contents are mapped to the same names
+            std::string msgGlobalName =
+                std::string("Arc_Assert_Op_Msg: ") + message.str();
```
maerhart, line 210: Maybe adding the source location to the message could be useful.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,68 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+            if (!globalOp)
+              globalOp = rewriter.create<LLVM::GlobalOp>(
+                  loc, LLVM::LLVMArrayType::get(int8Type, message.size()),
+                  /*isConstant=*/true, LLVM::Linkage::Internal, msgGlobalName,
+                  rewriter.getStringAttr(msgRef));
+          }
+
+          Value globalString =
+                rewriter.create<LLVM::AddressOfOp>(loc, globalOp);
+          builder.template create<LLVM::CallOp>(loc, putsFunc,
```
maerhart, line 223: Is the `.template` really necessary?

lib/Dialect/Arc/ArcOps.cpp
```diff
@@ -618,6 +618,17 @@ LogicalResult SimStepOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// AssertOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult AssertOp::verify() {
+  if (getMsg().has_value() && getMsg().value().empty())
+    emitOpError("assertion message can not be empty");
```
maerhart, line 627:
```suggestion
    return emitOpError("assertion message cannot be empty");
```

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
elhewaty: I can't think of a way to get the location of the here are two cases:
- If the message is provided directly to the operation (`arc.assert bool, clock, "some message"`), so the location of it is the location of the operation (aka `arc.assert`), here we map the same message contents to the same global string(address) which is very useful.
- If the message is provided as a variable (`arc.assert bool, clock, StrOp`), I need to find a way to get the location of the operand and then map all users to the same message, as the location will be different for different messages, and maybe they have the same contents, but in different locations.

Is there any way to convert `Location` to `StringRef`?

maerhart: Yes, I mean the location that you get from `assertOp.getLoc()`. The `Location` class has a `print` function that you can use. Something like this:
```
std::string str;
llvm::raw_string_ostream(str);
loc.print(stream);
// use str here
```
Because the message is a string attribute, you can just concatenate them at compile-time and still merge globals referring to the same string.

This is not super important, so if you're having issues with it, we can also just leave it away.

elhewaty: That's great, but I tried it now it generates a location like this:
```
/Path/To/circt/test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir:276:3

```
I can use some `regex` matching and the output will be `lower-arc-to-llvm.mlir:276:3`, but If someone tries to add new tests at the start of the file this will change the line of the assertion op, the test will fail.
can we add it without test coverage?

elhewaty: That's great, but I tried it now it generates a location like this:
```
/Path/To/circt/test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir:276:3

```
I can use some `regex` matching and the output will be `lower-arc-to-llvm.mlir:276:3`,  but I think it will differ from path to path, I tried but I couldn't match it so I ignored this part!
If you have further suggestions, I will address them.

elhewaty:
@maerhart I don't know what's wrong with the checker! I run clang-format on my local repo and I don't get any errors!!

maerhart:
In the github actions summary you can download a patch it produced.

Here it is:
```
diff --git a/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp b/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
index 120a140..7e06ed8 100644
--- a/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
+++ b/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
@@ -185,8 +185,7 @@ struct AssertOpLowering : public OpConversionPattern<arc::AssertOp> {
                                            rewriter.getBoolAttr(true)));
 
     rewriter.replaceOpWithNewOp<scf::IfOp>(
-        op, negateCondition,
-        [&](OpBuilder &builder, Location loc) {
+        op, negateCondition, [&](OpBuilder &builder, Location loc) {
           // If the condition doesn't hold then print the message and abort
           auto message = op.getMsg().value_or("");
           auto module = op->getParentOfType<ModuleOp>();
```

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,72 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+
+          LLVM::LLVMFuncOp putsFunc;
+          LLVM::GlobalOp globalOp;
+
+          // If there's no message just abort
+          if (!message.empty()) {
+            {
+              OpBuilder::InsertionGuard guard(rewriter);
+              rewriter.setInsertionPointToStart(module.getBody());
+              putsFunc = lookupOrCreateFn(
```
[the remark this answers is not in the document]
dtzSiFive: While looking into what to suggest, looks like there are various useful helpers for basically exactly this flow.

Additionally, `cf.assert` has an LLVM lowering -- was wondering if makes sense to lower to/use `cf.assert` since this seems rather similar?

Regardless, the lowering there uses the rewriter-aware and useful helpers for grabbing a print-like method as well as global strings:

https://github.com/llvm/llvm-project/blob/9d4f7f44b64d87d1068859906f43b7ce03a7388b/mlir/lib/Conversion/ControlFlowToLLVM/ControlFlowToLLVM.cpp#L52

It doesn't re-use the global string definition, FWIW, but anyway interesting related bits if not already on your radar.

And actually, looks like it ultimately just calls `lookupOrCreateFn` (or wrappers) anyway: https://github.com/llvm/llvm-project/blob/cc54129b983799e1aaea77aa0ff3040dc30cbc8c/mlir/lib/Conversion/LLVMCommon/PrintCallHelper.cpp#L63 .  Not sure how creating a global is safe?

Seems like could do what this is doing using rewriter (instead of the builder it creates)?
https://github.com/llvm/llvm-project/blob/cc54129b983799e1aaea77aa0ff3040dc30cbc8c/mlir/lib/Dialect/LLVMIR/IR/FunctionCallUtils.cpp#L51
(possible upstream patch, but if nothing else locally if looking to address that...)

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -173,6 +173,72 @@ struct StorageGetOpLowering : public OpConversionPattern<arc::StorageGetOp> {
...
+                  LLVM::LLVMVoidType::get(getContext()), false);
+
+              // Get the integer type with the correct bit width.
+              Type int8Type = rewriter.getI8Type();
+
+              // Generate new names for global strings.
+              // Messgaes with the same contents, and the same locations
+              // are mapped to the same names
+              std::string msgGlobalName =
+                  std::string("Arc_Assert_Op_Msg: ") + message.str();
```
dtzSiFive, line 212: Especially with the colon and space it seems unlikely, but can we rely on that symbol not being defined?

Symbol name having the entire message seems like a lot? 🤔

PR #6841, opened 2024-03-18
[FIRRTL] Add `chisel_assert` intrinsic

uenoku:
This implements `chisel_assert` intrinsic that is similar to `chisel_assert_assume` except that it doesn't create a companion assume. Internally `with_companion_assume` attribute is attached to assert op for assertions with companion assume (currently all assertions are expected to create companion assertions, unfortunately).

### DIFF
--- a/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
+++ b/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
@@ -95,6 +95,28 @@ The enable input is sampled at the rising edge of the input clock; any changes o
 | en   | input     | UInt<1>  | enable for the output clock |
 | out  | output    | Clock    | gated output clock          |
 
+### circt.chisel_assert
+
+Generate a clocked SV assertion, with optional formatted error message.
+
+| Parameter | Type   | Description                                                                         |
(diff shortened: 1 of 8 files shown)


dtzSiFive (commented):
Awesome, thanks!

Few quick notes:
* Consider adding this to ODS, with expectation we'll remove it "soon".
* When using the option to convert to SVA's, all assert varities -- ifElseFatal via intrinsic or printf, even FIRRTL "assert" (!) become "assert assume".  This seems like this breaks that behavior, I think in practice this may be a problem for the ifElseFatal paths.

One "solution" could be setting that attribute on all asserts except this one?  With idea that if "ifElseFatal" it's ignored, I suppose.

May need to push on sorting out the lowering option story in order to land/move forward with this.

Anyway the intrinsic looks good and serves our needs until the FIRRTL op can be used for exactly this instead :+1:.

fabianschuiki (approved):
LGTM. I agree with adding the `with_companion_assume` property/attribute to ODS for as long as it has significant effect on the output.

uenoku:
Thank you for the reviews! This PR will be subsumed by https://github.com/llvm/circt/pull/6878 in a better way.

PR #6842, opened 2024-03-18
[SMT] Add quantifier operations

maerhart:
Add operations to represent `forall` and `exists`.

### DIFF
--- a/include/circt/Dialect/SMT/SMTOps.td
+++ b/include/circt/Dialect/SMT/SMTOps.td
@@ -179,7 +179,8 @@ def YieldOp : SMTOp<"yield", [
   Pure,
   Terminator,
   ReturnLike,
-  ParentOneOf<["smt::SolverOp", "smt::CheckOp"]>,
+  ParentOneOf<["smt::SolverOp", "smt::CheckOp",
+               "smt::ForallOp", "smt::ExistsOp"]>,
 ]> {
   let summary = "terminator operation for various regions of SMT operations";
   let arguments = (ins Variadic<AnyType>:$values);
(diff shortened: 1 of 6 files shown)


include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -337,4 +338,64 @@ def ImpliesOp : SMTOp<"implies", [Pure]> {
...
+    the block arguments must match the ones of the 'body' region and (other
+    than there) must be used at least once in the 'patterns' region. It may also
+    not contain any operations that bind variables, such as quantifiers. While
+    the 'body' region must always yield exactly one `!smt.bool`-typed value, the
+    'patterns' region can yield an arbitrary number (but at least one) of SMT
+    values.
+
+    The 'no_patterns' attribute is only allowed when no 'patterns' region is
+    specified and forbids the solver to generate and use patterns for this
+    quantifier.
```
math-fehr, line 363: Shouldn't this be inferred from the number of blocks in the `patterns` region?

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -337,4 +338,64 @@ def ImpliesOp : SMTOp<"implies", [Pure]> {
...
+
+    The 'weight' attribute indicates the importance of this quantifier being
+    instantiated compared to other quantifiers that may be present. The default
+    value is zero.
```
math-fehr, line 367: I couldn't find information about that attribute on the SMT-LIB documentation, do you have a link to it?

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -337,4 +338,64 @@ def ImpliesOp : SMTOp<"implies", [Pure]> {
...
+  }];
+
+  let arguments = (ins DefaultValuedAttr<I32Attr, "0">:$weight,
+                       UnitAttr:$noPattern,
+                       StrArrayAttr:$boundVarNames);
+  let regions = (region SizedRegion<1>:$body,
+                        MaxSizedRegion<1>:$patterns);
```
math-fehr, line 374: My understanding is that `patterns` should be a variadic region, right?
For instance: 
image

include/circt/Dialect/SMT/SMTOps.td
```diff
@@ -337,4 +338,64 @@ def ImpliesOp : SMTOp<"implies", [Pure]> {
...
+    quantifier.
+
+    The 'weight' attribute indicates the importance of this quantifier being
+    instantiated compared to other quantifiers that may be present. The default
+    value is zero.
+  }];
+
+  let arguments = (ins DefaultValuedAttr<I32Attr, "0">:$weight,
+                       UnitAttr:$noPattern,
+                       StrArrayAttr:$boundVarNames);
```
math-fehr, line 372: Can we make this optional somehow?
I'm not sure we want to force users to give names?

math-fehr (commented):
Nice! I just have questions about the `weights` attribute and mostly the `patterns` regions, but otherwise it looks good!

include/circt/Dialect/SMT/SMTOps.td
maerhart: No block in the patterns region and the `no_pattern` attribute are two separate things:
* `no_pattern`: doesn't allow the solver to auto-generate patterns
* no block in patterns region: I don't give the solver any patterns, but it is allowed to use heuristics (or whatever) to generate patterns and use the overall solving technique that requires patterns.

More info here: https://microsoft.github.io/z3guide/docs/logic/Quantifiers/#no-patterns

include/circt/Dialect/SMT/SMTOps.td
maerhart: SMT-LIB allows to annotate most (all?) terms with attributes (see SMT-LIB 2.6, section 3.4 and 3.6 (specifically the grammar figure in the beginning and 3.6.5)). Those annotations are not allowed to change the semantics of the language but allow adding metadata for solvers, such as this `weight` attribute which is supported, e.g., by Z3: https://stackoverflow.com/questions/62512506/which-smt-lib-attributes-does-z3-support-and-why

The `no-pattern` attribute is another one. Apparently, only a few attributes are important enough to be mentioned in the standard (such as `:pattern`).

My plan is to support (at least some of) the attributes supported by solvers but not mentioned in the standard such that the passes that generate SMT IR can have more control over the solver.

I can modify the documentation here to make this a bit clearer.

include/circt/Dialect/SMT/SMTOps.td
maerhart: Yes, that makes sense.

include/circt/Dialect/SMT/SMTOps.td
maerhart: Makes sense. I didn't read this part carefully enough.
But isn't it required that in each (multi-)pattern all bound variables have to occur at least once? Why does the second `:pattern` only cover `x0`?

maerhart: Giving this example in the standard to Z3 yields a warning: `WARNING: (14,1): pattern does not contain all quantified variables.`
```
(declare-sort A 0)
(declare-fun r (A A) Bool)
(declare-fun p (A A) Bool)
(declare-const a A)

(assert (forall ((x0 A ) (x1 A ) (x2 A ))
  (! (=> (and (r x0 x1 ) (r x1 x2 )) (r x0 x2 ))
  :pattern ((r x0 x1 ) (r x1 x2 ))
  :pattern ((p x0 a ))
)))

(check-sat)
```

So, I guess, there is either a problem with the example in the standard or Z3 does not strictly adhere to the standard.
I think I'll drop the verifier that checks that all bound variables are used at least once to be consistent with the standard, and because it's not a hard error in Z3.

include/circt/Dialect/SMT/SMTOps.td
math-fehr: Oh gotcha, that makes sense!

include/circt/Dialect/SMT/SMTOps.td
math-fehr: That's perfect thanks!

math-fehr (approved):
Perfect, thanks a lot!

PR #6843, opened 2024-03-18
[Emit] Copy fragments into modules created by ExtractTestCode

nandor:
This ensures that the extracted modules have the same fragments included as the original modules.


dtzSiFive (approved):
LGTM, thanks!

Are all fragments something that makes sense to copy and emit along with the extracted modules (idempotent maybe, benign if not needed/relevant?)?

In practice I think this is fine (set macro to default if not set), but just wondering what the contract of "emit.fragments" is and that relationship to how ETC works.

If this pass is only meant to be used with our pipeline (probably?) maybe that's not a big deal (re:relying on IR only using them in the way that makes this safe).

Maybe this is part of emit.fragments intended/required usage anyway?

nandor:
Copying everything is the safe bet here. Normally, if we move all asserts into the companion module, we can move the fragment there too, but here we don't really have the necessary information to do that nicely. We can safely over-approximate and emit fragments since they are idempotent.

PR #6844, opened 2024-03-18
[NFC] Move 'using namespace' out of headers.

fzi-hielscher:
`using namespace` generally should not be in headers as it can lead to _interesting_ aliasing problems, especially w.r.t. `namespace mlir` and `namespace llvm`.


maerhart (approved):
Thanks for investigating and fixing this @fzi-hielscher!

dtzSiFive:
Thanks for fixing this!

fabianschuiki:
Thanks for the fix :partying_face:!

PR #6845, opened 2024-03-18
[arcilator] Disable JIT on Windows

Moxinilian:
Unfortunately it seems like MSVC cannot build the Arcilator JIT. This PR thus disables the JIT infrastructure on Windows for the time being.


Moxinilian:
This confirms this commit fixes the Windows build issue: https://github.com/Moxinilian/circt/actions/runs/8328898298

maerhart:
Should we guard the includes nonetheless?

PR #6846, opened 2024-03-18
[arcilator] Add support for MemRef in the lowering pipeline

Moxinilian:
MemRef is a useful dialect in the context of writing simulations. This PR adds support for it in the lowering pipeline.

I had to remove the rewriting pattern for the FunctionOp interface. I am not sure why it was here in the first place and it was incorrectly lowering memref arguments. I do not know if the pattern is misused or generally incorrect, though.


maerhart:
Could you describe in more detail what you plan to use memref for?
Memref is quite a big thing to pull in and has some weird parts, so I just want to make sure it's actually the best approach to achieving our goals.

Moxinilian:
I use memref for memory allocation, basically.

Moxinilian:
In the end this is probably not necessary. Future readers, feel free to comment if you disagree.

PR #6848, opened 2024-03-19
[Sim] Introduce an op to represent prints

nandor:
This PR introduces a `sim.print` op that mainly acts as an intermediary between FIRRTL and SV to carry print statements.
Its output is hard-wired to `stderr`, similarly to the FIRRTL version. This could be relaxed in the future if the need arises.
Presenly, the goal of this op is to provide a compact representation that makes prints easier to analyze, without having to trace the parent op to find their clocking. Additionally, since FIRRTL stop lowers to a print + stop operation combination, this op makes it easier for some flows to perform fusion.

### DIFF
--- a/include/circt/Dialect/Sim/SimOps.td
+++ b/include/circt/Dialect/Sim/SimOps.td
@@ -52,6 +52,29 @@ def FatalOp : SimOp<"fatal"> {
   let results = (outs);
 
   let assemblyFormat = "$clk `,` $cond attr-dict";
+
+}
+
+def PrintOp : SimOp<"print"> {
+  let summary = "Prints a formatted string to `stderr`";
+
+  let description = [{
+    Prints a formatted string, using standard SystemVerilog placeholders,
+    to `stderr` on the rising edge of `clk` when `cond` is high.
+  }];
+
+  let arguments = (ins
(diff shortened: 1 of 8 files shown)


include/circt/Dialect/Sim/SimOps.td
```diff
@@ -52,6 +52,29 @@ def FatalOp : SimOp<"fatal"> {
...
+    I1:$cond,
+    StrAttr:$format_string,
+    Variadic<AnyType>:$substitutions
+  );
+  let results = (outs);
+
+  let assemblyFormat = [{
+    $clk `,` $cond `,` $format_string attr-dict (`(` $substitutions^ `)`
+      `:` qualified(type($substitutions)))?
+  }];
```
fabianschuiki, line 77: Could this use the Verif dialect's [FormatVerilogStringOp](https://github.com/llvm/circt/blob/main/include/circt/Dialect/Verif/VerifOps.td#L46-L78) instead of doing formatting? I think the consensus there was to try factor formatting out into one op, instead of having many ops deal with the weird SV formatting.

I guess using Verif's `PrintOp` is not really posisble here, since you want an op with a clock that you can put in a graph region, right?

PR #6850, opened 2024-03-19
[HW][IST] Verify simple inner-ref-user ops sequentially, perf fix.

dtzSiFive:
When an IRN has a huge number of top-level operations that need inner ref verification, the current code scaled poorly as the amount of per-op verification work is tiny (handful of hashtable lookups) compared to the overhead and atomic access of failableParallelForEach.

Scan through the operations sequentially as a pre-pass and directly verify operations with no regions, only using parallelization for verifying recursive walks of modules.


youngar (approved):
Nice work!

KidiIT (commented):
All reviewed and approved🫶❤️

KidiIT:
All done here🫶❤️

PR #6851, opened 2024-03-19
StripDebugInfoWithPred: Fix parallelization perf issue.

dtzSiFive:
Handle operations with no regions sequentially, parallelize walking the rest.  Avoid cost of parallelForEach being more than processing the op directly.

Similar fix as with #6850 .

### DIFF
--- a/lib/Transforms/StripDebugInfoWithPred.cpp
+++ b/lib/Transforms/StripDebugInfoWithPred.cpp
@@ -83,20 +83,39 @@ void StripDebugInfoWithPred::runOnOperation() {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


dtzSiFive:
This too showed up on same timing report on large design and has same basic issue as #6850, but has not been specifically measured in practice to address the issue.

Really wanted to give this to pass manager instead (and schedule as nested pass on all operations), but that would have same problem unfortunately.

mikeurbach (approved):
Nice, thanks for tracking these down.

uenoku (approved):
LGTM

PR #6856, opened 2024-03-20
[Seq] Enforce non-negative preset values in FirRegOp printer/parser

fzi-hielscher:
Parsing a negative preset value for the `FirRegOp` will currently produce the slightly inappropriate message "preset value too large". On the other hand, the printer does output negative integers if the value's MSB is set.

This PR adds a more specific error message to the parser and ensures the printed value is always non-negative.

### DIFF
--- a/lib/Dialect/Seq/SeqOps.cpp
+++ b/lib/Dialect/Seq/SeqOps.cpp
@@ -441,7 +441,10 @@ ParseResult FirRegOp::parse(OpAsmParser &parser, OperationState &result) {
     OptionalParseResult presetIntResult =
         parser.parseOptionalInteger(presetValue.emplace());
     if (!presetIntResult.has_value() || failed(*presetIntResult))
-      return parser.emitError(loc, "expected integer value");
+      return parser.emitError(presetValueLoc, "expected integer value");
+    if (presetValue->isNegative())
+      return parser.emitError(presetValueLoc,
+                              "preset value must not be negative");
   }
 
   Type ty;
@@ -464,7 +467,7 @@ ParseResult FirRegOp::parse(OpAsmParser &parser, OperationState &result) {
 
     APInt presetResult = presetValue->sextOrTrunc(width);
     if (presetResult.zextOrTrunc(presetValue->getBitWidth()) != *presetValue)
-      return parser.emitError(loc, "preset value too large");
+      return parser.emitError(presetValueLoc, "preset value too large");
 
     auto builder = parser.getBuilder();
     auto presetTy = builder.getIntegerType(width);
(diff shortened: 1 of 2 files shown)


lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -441,7 +441,10 @@ ParseResult FirRegOp::parse(OpAsmParser &parser, OperationState &result) {
     OptionalParseResult presetIntResult =
         parser.parseOptionalInteger(presetValue.emplace());
     if (!presetIntResult.has_value() || failed(*presetIntResult))
-      return parser.emitError(loc, "expected integer value");
+      return parser.emitError(presetValueLoc, "expected integer value");
+    if (presetValue->isNegative())
+      return parser.emitError(presetValueLoc,
+                              "preset value must not be negative");
```
seldridge, line 447: Thanks for correcting the location. I assume this will get the carat indicator in the correct place now?

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -508,7 +511,12 @@ void FirRegOp::print(::mlir::OpAsmPrinter &p) {
   }
 
   if (auto preset = getPresetAttr()) {
-    p << " preset " << preset.getValue();
+    p << " preset ";
+    const auto &presetVal = preset.getValue();
+    if (presetVal.isNonNegative())
+      p << presetVal;
+    else
+      p << presetVal.zext(presetVal.getBitWidth() + 1);
```
seldridge, line 519: This is supposed to change the round-tripping so that negative preset values will no longer be emitted, correct?

Is there a test that could be added for this?

lib/Dialect/Seq/SeqOps.cpp
fzi-hielscher:
```
error: custom op 'seq.firreg' expected integer value
    %preset_0 = seq.firreg %next32 clock %clock preset NaN : i32
                                                       ^
```

lib/Dialect/Seq/SeqOps.cpp
fzi-hielscher: Correct. Sorry for the hasty PR, it is really just a hotfix to make round-tripping work again. I did not even notice that there already is a `seq/round-trip.mlir` test.
To be honest, I had to resist the urge of rewriting the preset attribute entirely. Something about bitcasting a raw integer to whatever type the register has just rubs me the wrong way. I would prefer it to be a `TypedAttrInterface` instead of an `APIntAttr`. But I can see that this will create new parsing headaches regarding aggregate values.

fzi-hielscher:
I'm pulling this out of oblivion, as I ran into the problem of not being able to round-trip the textual IR again.

PR #6857, opened 2024-03-20
[FIRRTL] Remove support for circt.Intrinsic annotation.

dtzSiFive:
Documentation indicates this was intended to be removed once FIRRTL language supports intrinsics, and we have intmodule's now.

### DIFF
--- a/docs/Dialects/FIRRTL/FIRRTLAnnotations.md
+++ b/docs/Dialects/FIRRTL/FIRRTLAnnotations.md
@@ -858,19 +858,6 @@ Example:
 }
 ```
 
-### circt.Intrinsic
-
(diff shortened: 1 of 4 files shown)


fabianschuiki (approved):
LGTM! I haven't seen any uses of `circt.Intrinsic` out in the wild anymore. Everything seems to go through `intmodule`. Otherwise we can always revert the PR.

dtzSiFive:
Looks like it's unused but technically not removed from Chisel presently.
I had checked for the string but right it's the class name.  It's dead regardless.

PR removing it here:
https://github.com/chipsalliance/chisel/pull/3945 .

PR #6859, opened 2024-03-20
[ImportVerilog] Add assign and pre/post increment/decrement expressions

fabianschuiki:
Add support for pre and post increment and decrement expressions, like `x++` and `--x`, as well as assign expressions, like `a += 5`. Slang represents these assignments as `Assign(a, Add(LValueRef, 5))` in the AST. The `LValueRef` node contextually refers to the parent assignment's left-hand side. To deal with this, also add a corresponding lvalue stack to the conversion context. Assignments push and pop their lvalues onto and off of this stack.

These expressions require a mechanism in the IR to express _when_ a variable is read. To capture this, add a new `moore.read_lvalue` op. It currently looks like an identity operation with a `MemRead` side effect. Further down the road, we may want to introduce a proper reference type for variables, ports, nets, and other things, and have `read_lvalue` and the various assigns operate on that type instead. This sets the foundation for that.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -144,10 +144,24 @@ def VariableOp : MooreOp<"variable", [
     See IEEE 1800-2017 § 6.8 "Variable declarations".
   }];
   let arguments = (ins StrAttr:$name, Optional<UnpackedType>:$initial);
-  let results = (outs UnpackedType:$result);
+  let results = (outs Res<UnpackedType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
     `` custom<ImplicitSSAName>($name) ($initial^)? attr-dict
     `:` type($result)
+  }];
+}
+
+def ReadLValueOp : MooreOp<"read_lvalue", [SameOperandsAndResultType]> {
+  let summary = "Read the current value of a declaration";
+  let description = [{
+    Samples the current value of a declaration. This is a helper to capture the
+    exact point at which declarations that can be targeted by assignments are
+    read.
+  }];
+  let arguments = (ins Arg<UnpackedType, "", [MemRead]>:$input);
+  let results = (outs UnpackedType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($result)
   }];
 }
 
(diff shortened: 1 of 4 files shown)


hailongSun2000 (approved):
LGTM!

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -53,6 +53,13 @@ struct ExprVisitor {
     return {};
   }
 
+  // Handle references to the left-hand side of a parent assignment.
+  Value visit(const slang::ast::LValueReferenceExpression &expr) {
```
maerhart, line 57: Do continuous assignments use a different system for lvalues or can they not be nested?

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -77,7 +84,9 @@ struct ExprVisitor {
   // Handle blocking and non-blocking assignments.
   Value visit(const slang::ast::AssignmentExpression &expr) {
     auto lhs = context.convertExpression(expr.left());
+    context.lvalueStack.push_back(lhs);
     auto rhs = context.convertExpression(expr.right());
+    context.lvalueStack.pop_back();
```
maerhart, line 89: There is no test that really exercises this stack, i.e., a case where more than one value is on the stack.

maerhart (approved):
LGTM, just some minor comments.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -135,6 +135,7 @@ def ProcedureOp : MooreOp<"procedure", [
 //===----------------------------------------------------------------------===//
 
 def VariableOp : MooreOp<"variable", [
+  MemoryEffects<[MemAlloc]>,
```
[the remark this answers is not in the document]
fabianschuiki: Cool, I didn't know that that was a thing. Let me give it a try!

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: They don't allow for the `+=` syntax. `+=` is always a blocking assignment within a procedural block.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Great point. I'll add one.

PR #6862, opened 2024-03-21
[InferReadWrite] Update the heuristic to infer the enable signal

prithayan:
This commit updates the heuristic to infer the enable signal drivers. Since temporary wires can be introduced by circt passes, the immediate `getConnectSrc` is not enough to infer the enable signal that might also drive the `wmode`. 
Traverse the assignment chain to get all the temporary wires that can potentially drive the enable. Use the set to simplify the write mode expression.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
@@ -411,6 +411,16 @@ private:
     return llvm::all_equal(valueBitsSrc[val]);
   }
 
+  void getEnableDrivers(Value enable, llvm::SmallDenseSet<Value> &enableSet) {
+    while (auto src = getConnectSrc(enable)) {
+      if (!enableSet.insert(src).second)
+        return;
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/InferReadWrite.cpp
```diff
@@ -411,6 +411,26 @@ struct InferReadWritePass : public InferReadWriteBase<InferReadWritePass> {
...
+  // Note: circt passes can introduce such temporary wires.
+  void getEnableDrivers(Value enable, llvm::SmallDenseSet<Value> &enableSet) {
+    while (auto src = getConnectSrc(enable)) {
+      if (!enableSet.insert(src).second)
+        return;
+      if (!isa_and_nonnull<WireOp>(src.getDefiningOp()))
+        return;
+      enable = src;
+    }
+  }
```
uenoku, line 432: I think this is almost identical to what firrtl::getModuleScopedDriver provides. Can we use/extend that helper function?

PR #6863, opened 2024-03-22
[FIRRTL] Add CreateCompanionAssume pass; Decouple UNROnlyAssume generation from AssertOp lowering

uenoku:
This PR separates UNROnlyAssume generation from AssertOp lowering into a dedicate op.

This commit adds:
* LowerToHW support for `UnclockedPredicateIntrinsicOp` (added in https://github.com/llvm/circt/pull/6867).
* CreateCompanionAssume pass which explicitly introduces companion assumes to the IR. Normal assume and UnclockedPredicateIntrinsicOp are used based on the guard's content.

The commit is intended not to change semantics of generated verilog. This commit should only introduces cosmetic changes regarding guards(because now they are reused).

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/CreateCompanionAssume.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CreateCompanionAssume.cpp
@@ -38,8 +38,8 @@ struct CreateCompanionAssumePass
         });
       }
 
-      // TODO: Currently messages are dropped to preserve the old behaviour. Copy
-      // messages once we confirmed that it works well with UNR tools.
+      // TODO: Currently messages are dropped to preserve the old behaviour.
+      // Copy messages once we confirmed that it works well with UNR tools.
       if (isUnrOnlyAssert)
         // If UNROnly, use UnclockedAssumeIntrinsicOp.
         assume = builder.create<firrtl::UnclockedAssumeIntrinsicOp>(
--- a/lib/Dialect/FIRRTL/Transforms/CreateCompanionAssume.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CreateCompanionAssume.cpp
@@ -38,8 +38,8 @@ struct CreateCompanionAssumePass
         });
       }
 
-      // TODO: Currently messages are dropped to preserve the old behaivor. Copy
-      // messages once it works well with UNR tools.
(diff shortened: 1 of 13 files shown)


lib/Dialect/FIRRTL/Transforms/CreateCompanionAssume.cpp
```diff
@@ -0,0 +1,66 @@
...
+        isUnrOnlyAssert = llvm::any_of(guards, [](Attribute attr) {
+          StringAttr strAttr = dyn_cast<StringAttr>(attr);
+          return strAttr && strAttr.getValue() == "USE_UNR_ONLY_CONSTRAINTS";
+        });
+      }
+
+      // If so, use AssumeEdgedPredicateIntrinsicOp.
+      if (isUnrOnlyAssert)
+        assume = builder.create<firrtl::AssumeEdgedPredicateIntrinsicOp>(
+            assertOp.getLoc(), assertOp.getPredicate(), assertOp.getEnable(),
```
dtzSiFive, line 43: Companion assume's -- both kinds -- don't have the message or substitution arguments today, see:

https://github.com/dtzSiFive/firrtl-verif-example/blob/master/assert/conc_true-guard_unr-fmt_%22sva%22-msg_%22message%22-name_%22label%22/output.sv
https://github.com/dtzSiFive/firrtl-verif-example/blob/master/assert/conc_true-guard_no-fmt_%22sva%22-msg_%22message%22-name_%22label%22/output.sv

Does seem nice to have them, just a detail to keep in mind.

dtzSiFive (commented):
Generally LGTM, awesome thanks!

lib/Dialect/FIRRTL/Transforms/CreateCompanionAssume.cpp
uenoku: Yeah, thank you for pointing out! For normal assume I think it's totally Ok to have substitutions. For UNROnly assume, probably they are not enabled in the normal simulation but I agree that the behavior is uncertain.

fabianschuiki (approved):
Great idea to factor out the companion assume business into its own pass, and that there is a dedicated intrinsic to capture the very peculiar flavor of assumes that some formal tools require. Pushes the ugly parts to the boundary and makes for a cleaner flow. 🥳

PR #6865, opened 2024-03-22
[SV][Verif] Extract verif ops in SVExtractTestCode

### LINKED ISSUE #6864: [SV] SVExtractTestCode not Extracting Assert Properties
The following code example results in an unextracted assertion:

```mlir
hw.module @Foo(in %a : i1, in %b : i1) {
  %true = hw.constant true
  %0 = comb.xor bin %b, %true : i1
  %1 = ltl.delay %true, 1, 0 : i1
  %2 = ltl.concat %a, %1 : i1, !ltl.sequence
  %3 = ltl.implication %2, %0 : !ltl.sequence, i1
  verif.assert %3 label "foo" : !ltl.property
  hw.output
}
```

This produces with `circt-opt -sv-extract-test-code`:

```mlir
hw.module @Foo(in %a : i1, in %b : i1) {
  %true = hw.constant true
  %0 = comb.xor bin %b, %true : i1
  %1 = ltl.delay %true, 1, 0 : i1
  %2 = ltl.concat %a, %1 : i1, !ltl.sequence
  %3 = ltl.implication %2, %0 : !ltl.sequence, i1
  verif.assert %3 label "foo" : !ltl.property
  hw.output
}
```

seldridge:
Fix a bug where SVExtractTestCode was not extracting asserts, assumes, and covers from the verif dialect.  This resulted in unexpected end-to-end "failures" from Chisel where Chisel asserts, assumes, and covers were not extracted and left in the design.

Fixes #6864.

### DIFF
--- a/lib/Dialect/SV/Transforms/SVExtractTestCode.cpp
+++ b/lib/Dialect/SV/Transforms/SVExtractTestCode.cpp
@@ -23,6 +23,7 @@
 #include "circt/Dialect/SV/SVPasses.h"
 #include "circt/Dialect/Seq/SeqDialect.h"
 #include "circt/Dialect/Seq/SeqOps.h"
+#include "circt/Dialect/Verif/VerifOps.h"
 #include "circt/Support/Namespace.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/IRMapping.h"
@@ -552,7 +553,8 @@ static bool isAssertOp(hw::HWSymbolCache &symCache, Operation *op) {
     return false;
   }
 
-  return isa<AssertOp, FinishOp, FWriteOp, AssertConcurrentOp, FatalOp>(op);
+  return isa<AssertOp, FinishOp, FWriteOp, AssertConcurrentOp, FatalOp,
+             verif::AssertOp>(op);
 }
 
 static bool isCoverOp(hw::HWSymbolCache &symCache, Operation *op) {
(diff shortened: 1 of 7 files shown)


uenoku (approved):
LGTM

lib/Dialect/SV/Transforms/SVExtractTestCode.cpp
```diff
@@ -552,7 +553,8 @@ static bool isAssertOp(hw::HWSymbolCache &symCache, Operation *op) {
     return false;
   }
 
-  return isa<AssertOp, FinishOp, FWriteOp, AssertConcurrentOp, FatalOp>(op);
+  return isa<AssertOp, FinishOp, FWriteOp, AssertConcurrentOp, FatalOp,
+             verif::AssertOp>(op);
```
dtzSiFive, line 557: Can't wait for layers to replace the need for this sort of thing!  And push what-goes-in-what-optional-layer to the user, instead of "fwrite is an assert" or having this error-string-prefix matching in an "SV" pass.

dtzSiFive (approved):
LGTM.

I think yes but am not 100% -- should "Verif" now be added to the dependent dialects for this pass in its TD declaration?

lib/Dialect/SV/Transforms/SVExtractTestCode.cpp
seldridge: I agree entirely.

The more serious issue with this PR is that this isn't just extracting the assert, but the LTL-based SVAs likely involve expressions which _really need to be extracted_. E.g., this was first reported due to a `===` showing up in the output and _not_ because of the assert.

Having advanced layer sink online would help with this, but it's not critical now that I think about it.

seldridge:
@dtzSiFive wrote:
> I think yes but am not 100% -- should "Verif" now be added to the dependent dialects for this pass in its TD declaration?

Yes, I think you're right. I added this in: https://github.com/llvm/circt/pull/6865/commits/959a12391430c58abeb1ca2a104a95a55f430116.

PR #6867, opened 2024-03-25
[FIRRTL] Add intrinsic for UNR only assume

uenoku:
This adds UnclockedAssumeIntrinsicOp which generates a SV assume statement whose predicate is used in a sensitivity list of the enclosing always block.


docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
```diff
@@ -199,3 +199,31 @@ Example SV output:
 ```systemverilog
 cover__label: cover property (@(posedge clock) enable & cond);
 ```
+
+### circt.assume_edged_predicate
+
+Generate a SV assume statement whose predicate is used in a sensitivity list of the enclosing always block.
```
dtzSiFive, line 205: No clock, that's right. Interesting!

docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
```diff
@@ -199,3 +199,31 @@ Example SV output:
...
+| enable    | input     | UInt<1>  | enable signal              |
+| ...       | input     | Signals  | arguments to format string |
+
+Example SV output:
+```systemverilog
+ifdef USE_FORMAL_ONLY_CONSTRAINTS
+ `ifdef USE_UNR_ONLY_CONSTRAINTS
+   wire _GEN = ~enable | pred;
+   always @(edge _GEN)
+     assume_label: assume(_GEN) else $error("Conditional compilation example for UNR-only and formal-only assert");
```
dtzSiFive, line 226: Companion assume's don't have error messages today, but seems reasonable to support on the intrinsic (can use or not when migrating).

Was thinking about implications of not being under a clock in the case when the error has substitution arguments.

dtzSiFive (commented):
Generally LGTM!

Consider having this work the same way AssumeOp does re:Fold (?) and InferWidths (so doesn't hit "Default" case and error out? untested).

uenoku:
> Consider having this work the same way AssumeOp does re:Fold (?) and InferWidths (so doesn't hit "Default" case and error out? untested).

Added, thanks!

docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
uenoku: Yeah thank you for pointing out. We add `%sampled` for concurrent assertions but not for immediate assertions but I'm uncertain we need $sampled for this. I'm fine to drop messages for migration.

fabianschuiki:
Maybe `UnclockedAssumeIntrinsicOp` could work? Because the fact that it uses this edge syntax is probably more of an accident than a core feature. `always @* assume ...` might work just as well 😃

uenoku:
> UnclockedAssumeIntrinsicOp

Yeah that sounds great!

PR #6869, opened 2024-03-25
[FIRRTL] Generic intrinsic op, rewrite intmodules to ops, LowerIntrinsics as conversion

dtzSiFive:
Add generic intrinsic op to FIRRTL IR.  This captures FIRRTL intrinsics generically in an expression-like format that can be directly parsed into, and is intended to replace what today is accomplished via instances of intmodule's.
This is internal-only to start with, no parser (or emitter) support yet.

Add compatibility pass that converts instances of intmodule's to generic intrinsic operations (half of what LowerIntrinsics does today).  Intmodule's with multiple outputs have a bundle result type.  Also implement current special handling of defname=EICG_wrapper modules via option.
This separates out intmodule/instance handling and allows for simpler LowerIntrinsics implementation that only needs to worry about the op form of intrinsics.

Rework LowerIntrinsics to rewrite generic intrinsic ops to their specific implementations.  Make this a module pass and use the rewriter conversion infrastructure.


dtzSiFive:
Draft for feedback and because PR will be broken up at least into "add op" "add new pass and tests for it" and finally "rewrite LowerIntrinsics and modify pipeline".

include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -671,19 +671,30 @@ def LowerXMR : Pass<"firrtl-lower-xmr", "firrtl::CircuitOp"> {
...
   }];
-  let constructor = "circt::firrtl::createLowerIntrinsicsPass()";
+  let constructor = "circt::firrtl::createIntrinsicInstancesToOpsPass()";
   let options = [
     Option<"fixupEICGWrapper", "fixup-eicg-wrapper", "bool", "false",
       "Lower `EICG_wrapper` modules into clock gate intrinsics">,
   ];
 }
 
+def LowerIntrinsics : Pass<"firrtl-lower-intrinsics", "firrtl::FModuleOp"> {
```
fabianschuiki, line 690: It's very nice that this gets to be a FIRRTL module pass now, instead of a circuit pass. :partying_face:

include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -671,19 +671,30 @@ def LowerXMR : Pass<"firrtl-lower-xmr", "firrtl::CircuitOp"> {
   ];
 }
 
-def LowerIntrinsics : Pass<"firrtl-lower-intrinsics", "firrtl::CircuitOp"> {
-  let summary = "Lower intrinsics";
+def IntrinsicInstancesToOps : Pass<"firrtl-intrinsic-instances-to-ops", "firrtl::CircuitOp"> {
```
fabianschuiki, line 674: Mega-bikeshed: how about `LowerIntmodules`?

include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.h
```diff
@@ -13,105 +13,228 @@
...
   void add(StringRef name) {
-    addConverter<T>(intmods, name);
-  }
-
-  /// Registers a converter to an extmodule name.
-  template <typename T>
-  void addExtmod(StringRef name) {
-    addConverter<T>(extmods, name);
+    addConverter<T>(name);
   }
```
fabianschuiki, line 238: Could we make `add` do the replacement of `.` to `_` in the intrinsic name and call `addConverter` again with it? That way you'd only have to register an intrinsic lowering once with its proper `circt.foo.bar_baz` name, and you'd get the `circt_foo_bar_baz` by convention.

include/circt/Dialect/FIRRTL/Passes.td
dtzSiFive: I actually meant to explicitly invite suggestions for a better name, thanks!

include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.h
dtzSiFive: Good point, thanks!

I wonder if we want to support both anymore anyway?  Being slightly unsure, left this as-is for now.

The new syntax could support periods in the intrinsic name but it's simpler and maybe better to have that just be a single (identifier) token -- in practice we only use the non-underscore version in documentation (see below PR) and in some internal tests.  This isn't currently reachable, all paths must use the underscores presently.

But yeah this does seem like a silly thing to push onto users of this class, as well as causes typo-prone name duplications.

cc https://github.com/llvm/circt/pull/6860 .

include/circt/Dialect/FIRRTL/Passes.td
dtzSiFive: Going with the suggestion, thanks!
Wasn't sure since it's not obviously "lower" to go to the generic op format, and also operates mostly on instances OF intmodule's but I think this makes sense and is a lot more ergonomic.

lib/Dialect/FIRRTL/Transforms/LowerIntmodules.cpp
```diff
@@ -0,0 +1,188 @@
...
+      use->erase();
+      inst.erase();
+    }
+    // Remove intmodule from IR and instance graph.
+    ig.erase(node);
+    op.erase();
+  }
+
+  // Special handling for magic EICG wrapper extmodule.  Deprecate and remove.
+  if (fixupEICGWrapper) {
```
uenoku, line 139: nit: It's unfortunate that we have to duplicate similar code just for single use case. Do you happen to know it's already safe to remove the flag? Or should we just define the EICG wrapper as an intrinsic?

uenoku (commented):
Awesome, thank you for pushing on this! :100: Using converter/rewriter looks really great!

include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.h
fabianschuiki: Yeah good point. I guess this is more or less inspired by LLVM's use of `.` as grouping/namespacing character in its intrinsics, but then needing a workaround since Chisel/FIRRTL can't generate that 😁

lib/Dialect/FIRRTL/Transforms/LowerIntmodules.cpp
fabianschuiki: We still haven't flipped the switch on this. @mwachs5 has been working on a final missing piece in `ExtractInstances`, where we're not yet properly handling the intrinsic. Once that's in, we can enable this option by default for a few weeks, and then declare it canon and rip out the old `EICG_wrapper` handling everywhere. But I guess this will stay for a few more months until all Chisel code bases are migrated over to using clock gating intrinsics.

uenoku: That makes sense. Thank you for elaborating this!

dtzSiFive: Awesome thanks for the summary and explanation!

FWIW I intentionally didn't maintain this as some sort of generic extmodule support (or try to combine for code re-use, as not-awesome as that is) specifically because I want this to be super easy to delete when the time comes (and not easily become more load-bearing, standing out as a weird thing we wish we didn't do) 👍 .

Special weird magic handling does not sully the heart of the pass and must sit in the corner with a dunce cap on it 😉 .  Anyway should be able to drop this "soon"!

dtzSiFive:
Closing, as bits of this are landing / split into their own PR's.

Thank you very much for review + discussion! 🙇

PR #6870, opened 2024-03-25
[SMT] Add SMT-LIB export translation

maerhart:
Quantifier patterns and other attributes are not yet supported, but will be added in a future PR since this one is already quite big. Feel free to let me know if I should split it up

### DIFF
--- a/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
+++ b/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
@@ -1,0 +1,646 @@
+//===- ExportSMTLIB.cpp - SMT-LIB Emitter -----=---------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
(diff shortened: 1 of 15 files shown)


lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -0,0 +1,646 @@
...
+  TypeVisitor typeVisitor;
+  Namespace &names;
+  ExpressionVisitor exprVisitor;
+};
+
+} // namespace
+
+//===----------------------------------------------------------------------===//
+// Unified and Split Emitter implementation
+//===----------------------------------------------------------------------===//
```
fabianschuiki, line 537: Does this really do unified or split emission? Looks like just unified emission at first glance.

fabianschuiki (approved):
I'm no expert on the SMT-LIB format, but from what I've seen of it, this looks good. I like the use of a visitor to produce the output, and the proper organization into a `Target`. And `ExportSMTLIB` itself is nicely factored into emission per supported theory. Very neat.

LGTM 🥳

math-fehr (approved):
Really really nice, I don't have anything to say there!

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
maerhart: Good catch. Copy-pasted from another pass and forgot to remove this.

PR #6871, opened 2024-03-25
[FIRRTL] Cache a symbol table instead of doing linear lookups every instance.

darthscsi:
roughly triples modules parsing performance on large designs.  This is about a 2x speedup for overall parsing stage.

### DIFF
--- a/lib/Dialect/FIRRTL/Import/FIRParser.cpp
+++ b/lib/Dialect/FIRRTL/Import/FIRParser.cpp
@@ -1541,12 +1541,14 @@ struct FIRStmtParser : public FIRParser {
   explicit FIRStmtParser(Block &blockToInsertInto,
                          FIRModuleContext &moduleContext,
                          hw::InnerSymbolNamespace &modNameSpace,
-                         FIRVersion version, SymbolRefAttr layerSym = {})
+                         SymbolTable &circuitSymTbl, FIRVersion version,
+                         SymbolRefAttr layerSym = {})
       : FIRParser(moduleContext.getConstants(), moduleContext.getLexer(),
                   version),
         builder(UnknownLoc::get(getContext()), getContext()),
         locationProcessor(this->builder), moduleContext(moduleContext),
-        modNameSpace(modNameSpace), layerSym(layerSym) {
+        modNameSpace(modNameSpace), layerSym(layerSym),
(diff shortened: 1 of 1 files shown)


seldridge (approved):

mikeurbach:
I just linked an issue I opened about this. I believe there are a couple other places we could do the lookup besides the one you've updated here. The issue had this search query: https://github.com/search?q=repo%3Allvm%2Fcirct%20path%3Alib%2FDialect%2FFIRRTL%2FImport%2FFIRParser.cpp%20%22lookupSymbol(%22&type=code

mikeurbach:
I guess just one other, we look up optionGroup from the circuit's symbol table here: https://github.com/llvm/circt/blob/43e6f074db0de934f5411ace120080ca962c27fe/lib/Dialect/FIRRTL/Import/FIRParser.cpp#L3882-L3883

That should be able to use the cached symbol table.

The final use of lookupSymbol is within `optionGroup`, which is itself a symbol table, so we couldn't use the cache there, but it should also be a much smaller scan.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5453,10 +5459,17 @@ ParseResult FIRCircuitParser::parseCircuit(
   // proactively touch it to make sure that it is always already created.
   (void)getLexer().translateLocation(info.getFIRLoc());
 
+  // Pre-verify symbol table, so we can construct it next.  Ideally, we would do
+  // this verification through the trait.
+  if (failed(mlir::detail::verifySymbolTable(circuit)))
```
dtzSiFive, line 5464: Little nervous about verifying symbol /users/ generically at this point, but probably it's fine.

It appears SymbolTable::lookup is actually "const", can we make the plumbed reference const as well? Might help ensure accidental insertion or mutation, hopefully that "just works"?

dtzSiFive (commented):
LGTM, thanks for addressing this.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
darthscsi: users shouldn't even exist, but it is an extra scan.

PR #6872, opened 2024-03-26
[LowerToHW] Set fragments outside the parallel region

nandor:


lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -401,6 +401,17 @@ struct CircuitLoweringState {
     macroDeclNames.insert(name);
   }
 
+  /// The list of fragments on which the modules rely. Must be set outside the
+  /// parallelized module lowering since module type reads access it.
+  DenseMap<hw::HWModuleOp, SetVector<Attribute>> fragments;
```
seldridge, line 406: Good use of `SetVector` so this will be stable.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -401,6 +401,17 @@ struct CircuitLoweringState {
...
+  std::mutex fragmentsMutex;
```
seldridge, line 407: `llvm::sys:SmartMutex<true>` may be preferable over `std`.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -401,6 +401,17 @@ struct CircuitLoweringState {
...
+
+  void addFragment(hw::HWModuleOp module, StringRef fragment) {
+    std::unique_lock<std::mutex> lock(fragmentsMutex);
+    fragments[module].insert(FlatSymbolRefAttr::get(
+        StringAttr::get(circuitOp.getContext(), fragment)));
```
seldridge, line 412: Nit: just using the Context and StringRef `get` method is likely cleaner than wrapping these in a `StringAttr` given that they use the same arguments:
```suggestion
    fragments[module].insert(FlatSymbolRefAttr::get(
        circuitOp.getContext(), fragment));
```

seldridge (approved):

Issue #4343, opened 2022-11-21
[docs] FIRRTLAnnotation images broken on website


dtzSiFive:
https://circt.llvm.org/docs/Dialects/FIRRTL/FIRRTLAnnotations/#targets

Links to https://circt.llvm.org/docs/Dialects/FIRRTL/FIRRTLAnnotations/includes/img/firrtl-folded-module.png .
Image is currently placed here: https://circt.llvm.org/includes/img/firrtl-folded-module.png

Related: https://github.com/llvm/circt/pull/3338

teqdruid:
same think occurs on getting started page. the fix in #3338 isn't ideal.

matthew-humphrey:
There is a [stack overflow article](https://stackoverflow.com/questions/13808020/include-an-svg-hosted-on-github-in-markdown) on how to link to images within the markdown. The two techniques they mention are to use relative URLs or to utilize raw.github.com for the URL. It seems like for the Getting Started page, using the relative link would be a simple and effective way to fix this.

matthew-humphrey:
> @matthew-humphrey I think this issue is about what appears on circt.llvm.org, not what GitHub renders when viewing the markdown in the docs directory on GitHub. We use this markdown to generate the website via https://github.com/llvm/circt-www. Ideally, we'd have no broken links on the website (i.e. this issue should be resolved), and it would also be great for browsing the docs on GitHub to also work. I think using relative links works on GitHub, but caused issues on the website.

Thanks, @mikeurbach  - I missed that subtlety.

PR #6873, opened 2024-03-27
[Docs] GettingStarted: Fix images and LLVM/MLIR contributing guide

ubfx:


mikeurbach (commented):
😞 the issue with the links and images is known. It has been brought up before, for example here: https://github.com/llvm/circt/issues/4343#issuecomment-1583330754

Basically, if we make the change in this PR, they'll work when viewed on GitHub, but will be broken on the actual website circt.llvm.org. No one has taken the time to figure out an approach that works in both scenarios.

The part about the LLVM/MLIR contribution process is great though, thanks for updating that. If you back out the changes to the links, we can merge that part.

If you're interested in figuring out how to specify the links in a way that works on both GitHub and the website, that would be great too, but the current change would break the website, which we want to keep working.

ubfx:
Ah, thanks for the explanation. I was wondering if I was missing something. I removed the image related changes from this PR but will take a look if I can figure out how to make this work with the website

mikeurbach (approved):

mikeurbach:
Awesome, that sounds good to me.

PR #6874, opened 2024-03-27
[FIRRTL] Add generic intrinsic op.

dtzSiFive:
Add generic intrinsic operation to FIRRTL.  Not used anywhere, just adding to the IR to build on.
Re-use module parameter printing/parsing for use as custom printer on the op.

Bit of context:

This is intended to have a FIRRTL textual equivalent that allows intrinsics to be used as expressions or statements, as a replacement for instances of intmodule's in the future, for ergonomic but primarily practical reasons that limit intrinsic use today.

The op is generic so that intrinsics may be parsed without the parser knowing details of supported intrinsics while still having sufficient type information to produce IR and good diagnostics during parsing (return value in particular).

Operands and results are required to be passive (and base) as that's all that is supported today (crashes at various points in the pipeline if not), and because passive specifically fits best with the model of intrinsics as having inputs it reads from and outputs distinctly.  This can be loosened in the future should this be needed (and supported).
There is a single (optional) return value, for use as a FIRRTL expression in a single simple format.  Multiple outputs become a single result bundle.

cc #6869 which shows some of what's planned next and demonstrates this in practice (internal to CIRCT anyway):
a pass converting instances+intmodules to these operations, and reworking LowerIntrinsics to expect these ops as the canonical form, making it possible to be a module pass and implement lowering as patterns for safety and simpler lowering code.

For now, start by adding the generic intrinsic operation.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
@@ -25,11 +25,11 @@ def GenericIntrinsicOp : FIRRTLOp<"int.generic",
 
   let arguments = (
     ins StrAttr:$intrinsic,
-    Variadic<AnyType>:$operands,
+    Variadic<PassiveType>:$operands,
     DefaultValuedAttr<ParamDeclArrayAttr, "{}">:$parameters
   );
 
-  let results = (outs Optional<AnyType>:$result);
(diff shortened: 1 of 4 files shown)


uenoku (approved):
LGTM

dtzSiFive:
Simple op, landing.  Post-merge review welcome and of course may change in the future.

PR #6875, opened 2024-03-27
[FIRRTL] Change Port Direction attribute from an APInt to a DenseArray.

darthscsi:
APInt requires an allocation to start large values.  Using apint for port directions requires copying the apint when inspecting port directions, thus introducing huge amount of copies throughout the code.  On large designs just this allocation overhead amounted to over 25% of the time spent (plus copy times).  DenseArrays are less space efficient (by 8x as used), but can be accessed by reference, removing these allocations.

Overall improvement is about 35% - 40%.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
@@ -49,13 +49,13 @@ static inline StringRef toString(bool direction) {
   return toString(get(direction));
 }
 
-/// Return a \p DenseBoolArrayAttr containing the packed representation of an array
-/// of directions.
+/// Return a \p DenseBoolArrayAttr containing the packed representation of an
+/// array of directions.
 mlir::DenseBoolArrayAttr packAttribute(MLIRContext *context,
                                        ArrayRef<Direction> directions);
 
-/// Return a \p DenseBoolArrayAttr containing the packed representation of an array
-/// of directions.
+/// Return a \p DenseBoolArrayAttr containing the packed representation of an
+/// array of directions.
 mlir::DenseBoolArrayAttr packAttribute(MLIRContext *context,
                                        ArrayRef<bool> directions);
 
--- a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
@@ -33,22 +33,33 @@ namespace direction {
 
 /// Return an output direction if \p isOutput is true, otherwise return an
 /// input direction.
-inline Direction get(bool isOutput) { return (Direction)isOutput; }
+static inline Direction get(bool isOutput) { return (Direction)isOutput; }
+
+/// Convert from Direction to bool.  The opposite of get;
+static inline bool unGet(Direction dir) { return (bool)dir; }
 
 /// Flip a port direction.
 Direction flip(Direction direction);
 
-inline StringRef toString(Direction direction) {
+static inline StringRef toString(Direction direction) {
   return direction == Direction::In ? "in" : "out";
+}
+
+static inline StringRef toString(bool direction) {
+  return toString(get(direction));
 }
 
 /// Return a \p IntegerAttr containing the packed representation of an array
 /// of directions.
-IntegerAttr packAttribute(MLIRContext *context, ArrayRef<Direction> directions);
+mlir::DenseBoolArrayAttr packAttribute(MLIRContext *context, ArrayRef<Direction> directions);
+
+/// Return a \p IntegerAttr containing the packed representation of an array
+/// of directions.
+mlir::DenseBoolArrayAttr packAttribute(MLIRContext *context, ArrayRef<bool> directions);
 
 /// Turn a packed representation of port attributes into a vector that can
 /// be worked with.
-SmallVector<Direction> unpackAttribute(IntegerAttr directions);
+SmallVector<Direction> unpackAttribute(mlir::DenseBoolArrayAttr directions);
 
 } // namespace direction
 } // namespace firrtl
--- a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
@@ -54,7 +54,7 @@ static inline StringRef toString(bool direction) {
 mlir::DenseBoolArrayAttr packAttribute(MLIRContext *context,
                                        ArrayRef<Direction> directions);
 
-/// Return a \p IntegerAttr containing the packed representation of an array
+/// Return a \p DenseBoolArrayAttr containing the packed representation of an array
 /// of directions.
 mlir::DenseBoolArrayAttr packAttribute(MLIRContext *context,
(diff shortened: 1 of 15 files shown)


include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
```diff
@@ -33,22 +33,35 @@ namespace direction {
...
+static inline StringRef toString(Direction direction) {
   return direction == Direction::In ? "in" : "out";
 }
 
+static inline StringRef toString(bool direction) {
+  return toString(get(direction));
+}
+
+/// Return a \p IntegerAttr containing the packed representation of an array
+/// of directions.
```
seldridge, line 53:
```suggestion
/// Return a \p DenseBoolArrayAttr containing the packed representation of an array
/// of directions.
```

Ditto for other comments which are now wrong below.

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -130,6 +130,7 @@ def FModuleOp : FIRRTLModuleLike<"module", [SingleBlock, NoTerminator]> {
   }];
   let arguments =
       (ins ConventionAttr:$convention,
+           DenseBoolArrayAttr:$portDirections,
```
seldridge, line 133: It's weird that this _wasn't_ in the argument list before.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -221,7 +221,8 @@ static void emitZeroWidthIndexingValue(PPS &os) {
 
 /// Return the verilog name of the port for the module.
 static StringRef getPortVerilogName(Operation *module, size_t portArgNum) {
-  return cast<HWModuleLike>(module).getPort(portArgNum).getVerilogName();
+  auto hml = cast<HWModuleLike>(module);
+  return hml.getPort(portArgNum).getVerilogName();
```
seldridge, line 225: Nit: unrelated change? Also, IMO inlining single-use seems better.

lib/Conversion/HandshakeToHW/HandshakeToHW.cpp
```diff
@@ -290,7 +290,7 @@ portToFieldInfo(llvm::ArrayRef<hw::PortInfo> portInfo) {
 // Convert any handshake.extmemory operations and the top-level I/O
 // associated with these.
 static LogicalResult convertExtMemoryOps(HWModuleOp mod) {
-  auto ports = mod.getPortList();
+  //  auto ports = mod.getPortList();
```
seldridge, line 293: Delete if not used.

lib/Dialect/FIRRTL/FIRRTLAttributes.cpp
```diff
@@ -34,26 +34,27 @@ Direction direction::flip(Direction direction) {
...
+mlir::DenseBoolArrayAttr
+direction::packAttribute(MLIRContext *context, ArrayRef<Direction> directions) {
   // Pack the array of directions into an APInt.  Input is zero, output is one.
-  auto size = directions.size();
-  APInt portDirections(size, 0);
-  for (size_t i = 0; i != size; ++i)
-    if (directions[i] == Direction::Out)
-      portDirections.setBit(i);
-  return IntegerAttr::get(IntegerType::get(context, size), portDirections);
+  SmallVector<bool> dirs;
```
seldridge, line 40: Doing a resize (as is done in a later method) would avoid repeated multiple allocations.

lib/Dialect/FIRRTL/FIRRTLAttributes.cpp
```diff
@@ -34,26 +34,27 @@ Direction direction::flip(Direction direction) {
...
+
+SmallVector<Direction>
+direction::unpackAttribute(mlir::DenseBoolArrayAttr directions) {
   SmallVector<Direction> result;
-  result.reserve(size);
-  for (size_t i = 0; i != size; ++i)
-    result.push_back(direction::get(value[i]));
+  result.reserve(directions.size());
+  for (size_t i = 0ULL, e = directions.size(); i != e; ++i)
+    result.push_back(direction::get(directions[i]));
```
seldridge, line 57: The `size_t` index is likely unnecessary now that this isn't an `IntegerAttr`. This is cleaner as:

```
for (auto d : directions) ...
```

Or similar.

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -857,7 +857,6 @@ void InferResetsPass::traceResets(InstanceOp inst) {
   LLVM_DEBUG(llvm::dbgs() << "Visiting instance " << inst.getName() << "\n");
 
   // Establish a connection between the instance ports and module ports.
-  auto dirs = module.getPortDirections();
```
seldridge, line 860: How is there just dead code laying around? 😖

test/Dialect/FIRRTL/errors.mlir
```diff
@@ -35,7 +35,7 @@ firrtl.circuit "" {
 
 firrtl.circuit "foo" {
 // expected-error @+1 {{requires 1 port directions}}
-firrtl.module @foo(in %a : !firrtl.uint<1>) attributes {portDirections = 3 : i2} {}
+firrtl.module @foo(in %a : !firrtl.uint<1>) attributes {portDirections = array<i1: true, true>} {}
```
seldridge, line 38: FWIW, this is easier to read now. 🙃

seldridge (approved):
Nits throughout. Looks good otherwise.

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
darthscsi: Nothing was, basically.  Added most of the things last week, but apparently missed this one.

lib/Dialect/FIRRTL/FIRRTLAttributes.cpp
darthscsi: no, because DenseArrayRef don't have begin and end!

include/circt/Dialect/FIRRTL/FIRRTLAttributes.h
```diff
@@ -33,22 +33,35 @@ namespace direction {
...
+static inline StringRef toString(bool direction) {
+  return toString(get(direction));
+}
+
+/// Return a \p DenseBoolArrayAttr containing the packed representation of an array
+/// of directions.
+mlir::DenseBoolArrayAttr packAttribute(MLIRContext *context,
+                                       ArrayRef<Direction> directions);
+
 /// Return a \p IntegerAttr containing the packed representation of an array
```
SpriteOvO, line 57:
```suggestion
/// Return a \p DenseBoolArrayAttr containing the packed representation of an array
```
The comment should be updated.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
darthscsi: Unrelated, but matches other code adjacent.  Will be rewritten soon.

PR #6876, opened 2024-03-27
[FIRRTL] Add LowerIntmodules pass.

dtzSiFive:
Add pass that converts instances of intmodules to the generic intrinsic op (firrtl.int.generic).

Also add optional special-case handling for recognizing extmodule with defname "EICG_wrapper" as encoding the clock gate intrinsic.

The pass is not included in the pipeline.

cc #6874 , #6869 .


fabianschuiki (approved):
LGTM! I like the idea of lowering IntModules into a more expression-like op, and then later just handle that simplified op in `LowerIntrinsics`. :+1:

PR #6878, opened 2024-03-28
[FIRRTL] Deprecate AssertAssume intrinsic and rename it to Assert

uenoku:
As we moved companion assume generation to a pass, having `AssertAssume` would not be necessary anymore. We still need assert instrinsic so this PR just renames the intrinsic to `assert`


fabianschuiki (approved):
LGTM!

PR #6881, opened 2024-03-29
[FIRRTL] Treat blackboxes in layers as "testbench"

### LINKED ISSUE #6880: [firrtl] extmodules with HasBlackBoxInline instantiated only in a layer not emitted into layers location
What I expect to happen:
------------------------

When I create an extmodule and generate it using HasBlackBoxInline annotation and it is instantiated only in a Layer, I expect that module to be emitted into the place where that layer belongs (e.g. in the TestBenchDirAnnotation-indicated directory when --extract-test-code is run).

What actually happens:
----------------------

Note that this works correctly for FIRRTL modules instantiated only in layers. It doesnt' work for those generated with HasBlackBoxInline.

For HasBlackBoxInline, they are generated alongside the non-layer code.

Steps to reproduce:
-------------------

```
bash-4.4$ firtool --version
LLVM (http://llvm.org/):
  LLVM version 19.0.0git
  Optimized build.
CIRCT firtool-1.70.0
```

in example.fir:
```
FIRRTL version 4.0.0

circuit FooTestbench: %[[
 {    "class":"sifive.enterprise.firrtl.MarkDUTAnnotation",
    "target":"FooTestbench.Foo"
 },
 {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~FooTestbench|Foo>in"
  },
   {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~FooTestbench|Foo>out"
  },
   {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~FooTestbench|FirrtlLogger>in_snoop"
  },
   {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~FooTestbench|FirrtlLogger>out_snoop"
  },
  {
    "class":"firrtl.transforms.BlackBoxInlineAnno",
    "target":"FooTestbench.BBLogger",
    "name":"BBLogger.sv",
    "text":"Black Box Logger verilog goes here!\n"
  },
  {
    "class": "sifive.enterprise.firrtl.TestBenchDirAnnotation",
    "dirname": "testbench"
  }
]]  
  
  layer A, bind :

  extmodule BBLogger:
    input in_snoop: UInt<1>
    input out_snoop: UInt<1>

  module FirrtlLogger:
    input in_snoop: UInt<1>
    input out_snoop: UInt<1>
    
  module Foo:
    input in: UInt<1>
    output out: UInt<1>
    connect out, in

    layerblock A:
      inst bb_logger of BBLogger
      connect bb_logger.in_snoop, in
      connect bb_logger.out_snoop, out

      inst firrtl_logger of FirrtlLogger
      connect firrtl_logger.in_snoop, in
      connect firrtl_logger.out_snoop, out

  public module FooTestbench enablelayer A:
    inst foo of Foo
    foo.in is invalid
```

running:

```
 firtool --split-verilog --o=firrtool_outputs3 --extract-test-code  example.fir
```

gives:

```
bash-4.4$ tree
.
├── BBLogger.sv
├── extern_modules.sv
├── filelist.f
├── firrtl_black_box_resource_files.f
├── Foo.sv
└── testbench
    ├── FirrtlLogger.sv
    ├── Foo_A.sv
    ├── FooTestbench.sv
    └── layers_FooTestbench_A.sv
```

I would have expected `BBLogger.sv` to be a sibling to `FirrtlLogger.sv`.

Note that there is no issue in the actual code, BBLogger is correctly only bound-in in `layers_FooTestbench_A.sv`, this is just a file location issue.

seldridge:
When determining which directory to place a blackbox into, make two changes to get this working correctly with layers:

  1. Never change a directory if one is already specified.
  2. Treat any bound instance as not under the DUT.

Taken together, this preserves both the original behavior of grand central blackbox extraction and causes blackboxes instantiated under layers to be placed in the testbench directory.

Fixes #6880.

Note: that this intentionally does _not_ test what happens when something is inside a Grand Central View. I'm going to go out on a limb and not specify this and state that we probably don't want to support it. Grand Central Views are planned to be fully replaced with layers making this question irrelevant.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
@@ -398,8 +398,6 @@ StringAttr BlackBoxReaderPass::loadFile(Operation *op, StringRef inputPath,
 OutputFileInfo BlackBoxReaderPass::getOutputFile(Operation *origOp,
                                                  StringAttr fileNameAttr,
                                                  bool isCover) {
-  // If the original operation has a specified output file that is not a
-  // directory, then just use that.
   auto outputFile = origOp->getAttrOfType<hw::OutputFileAttr>("output_file");
   if (outputFile && !outputFile.isDirectory()) {
     return {outputFile.getFilename(), Priority::TargetDir,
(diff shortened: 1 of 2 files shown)


mwachs5 (commented):
Tests look good!

PR #6683, opened 2024-02-09
[LowerToHW] Fix shr(0-bit, n) lowering

seldridge:
Fix an error if a 0-bit was shifted right by any amount and this made it all the way to LowerToHW (and wasn't canonicalized away earlier).  FIRRTL semantics interpret this case as a 1-bit zero.  Insert the 1-bit zero if we ever see this.

Fixes #6652.

PR #6882, opened 2024-03-30
[SeqToSV] Prepend RANDOMIZE-defining snippets

seldridge:
Fix an issue where the RANDOMIZE Verilog macro was used before it was
defined.

### DIFF
--- a/lib/Conversion/SeqToSV/SeqToSV.cpp
+++ b/lib/Conversion/SeqToSV/SeqToSV.cpp
@@ -413,9 +413,9 @@ void SeqToSVPass::runOnOperation() {
       FlatSymbolRefAttr::get(context, "RANDOM_INIT_MEM_FRAGMENT");
 
   auto regFragments = ArrayAttr::get(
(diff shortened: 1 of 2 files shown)


nandor:
Wouldn't it be simpler to change the order at https://github.com/llvm/circt/blob/aca154a607191430f544d211cfcf7cb02cb464e3/lib/Conversion/SeqToSV/SeqToSV.cpp#L416?

nandor:
Actually, I don't think that prepend/append is the problem here, it's the fragments themselves that have to be shuffled around

seldridge:
You are entirely correct. Swapping the order gets things working and this fixes the observed internal bug.

nandor (approved):
LGTM

nandor (changes requested):
Still not correct, behaviour is iffy because fragments can get mixed as reg init -> init -> mem init -> init. I'm working on addressing that.

nandor:
Possible fix here: https://github.com/llvm/circt/pull/6883

seldridge:
Closing in favor of #6683.

PR #6883, opened 2024-03-30
[SeqToSV] Fix the ordering of the memory/register random init fragments

nandor:


lib/Conversion/SeqToSV/SeqToSV.cpp
```diff
@@ -448,26 +423,58 @@ void SeqToSVPass::runOnOperation() {
   // Lower memories and registers in modules in parallel.
   bool needsRegRandomization = false;
   bool needsMemRandomization = false;
+  DenseMap<HWModuleOp, std::pair<bool, bool>> moduleFragments;
```
seldridge, line 426: The `std::pair<bool, bool>` makes it difficult to know what means what. A `struct` would be better.

lib/Conversion/SeqToSV/SeqToSV.cpp
```diff
@@ -448,26 +423,58 @@ void SeqToSVPass::runOnOperation() {
...
       needsMemRandomization = true;
     }
   });
 
+  auto randomInitFragmentName =
+      FlatSymbolRefAttr::get(context, "RANDOM_INIT_FRAGMENT");
+  auto randomInitRegFragmentName =
+      FlatSymbolRefAttr::get(context, "RANDOM_INIT_REG_FRAGMENT");
+  auto randomInitMemFragmentName =
+      FlatSymbolRefAttr::get(context, "RANDOM_INIT_MEM_FRAGMENT");
```
seldridge, line 457: Really nitty, however, these identifiers are all very similar and it makes it slightly difficult to check their usage later. Shorter names would make it more obvious that things like the following are correct without having to scan into the middle of the identifier:

``` c++
if (needsReg)
  fragmentAttrs.push_back(regFragment);
```

test/Conversion/SeqToSV/header.mlir
```diff
@@ -0,0 +1,79 @@
+// RUN: circt-opt %s -verify-diagnostics --lower-seq-to-sv | FileCheck %s
+
+
```
seldridge, line 3: Nit: extra newline.

seldridge (approved):
LGTM

This is passing the internal test that was failing.

PR #6884, opened 2024-04-01
[Moore] Introduce a new operation - netOp for net declaration

cepheus69:
According to the IEEE 1800-2017 std, we design netOp for net declaration. Various NetKind are supported to generate in the net declaration:
- `Supply0` `Supply1`
- `Wire` `Uwire`
- `WAnd` `WOr`
- `Tri0` `Tri1`
- `Tri` `TriAnd` `TriOr` `TriReg`
The interconnect net declaration and user-defined net types are not supported yet.


cepheus69:
The netOp is mainly used for declaring wire elements in the module.

hailongSun2000:
Thanks for helping code review in advance :smiley: !

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
```diff
@@ -83,6 +83,10 @@ struct Context {
   /// not been converted yet.
   std::queue<const slang::ast::InstanceBodySymbol *> moduleWorklist;
 
+  /// A list of ports for recording locations of ports to bypass extra
+  /// declaration of net or variable operation.
+  DenseMap<const slang::SourceLocation *, moore::Direction> portMap;
```
uenoku, line 88: Using SourceLocation as a key looks slightly counterintuitive to me. Is SourceLocation stable API that we can use in such a way? (I'm not much familiar with slang so please correct me if I'm wrong :) Can we use `const slang::Symbol*` or maybe `StringAttr` instead?

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -151,6 +177,28 @@ def VariableOp : MooreOp<"variable", [
...
+    See IEEE 1800-2017 § 6.7 "Net declarations".
+  }];
+  let arguments = (ins
+    StrAttr:$name,
+    StrAttr:$kind,
+    Optional<UnpackedType>:$assignment
+  );
+  let results = (outs UnpackedType:$result);
+  let assemblyFormat = [{
+    `` custom<ImplicitSSAName>($name) $kind ($assignment^)? attr-dict
```
uenoku, line 197: nit: Do we need `? 
```suggestion
    custom<ImplicitSSAName>($name) $kind ($assignment^)? attr-dict
```

uenoku (commented):
I have limited knowledge for Moore/Slang but this looks good to me :+1:

include/circt/Dialect/Moore/MooreOps.td
cepheus69: I did not notice, Thank you to point it out !

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
cepheus69: Thanks for your review! 🥰 We have used `stringAttr` to construct the container in the dev branch which has the same purpose. It was directly inserted into a scopedHashTable of type `stringAttr` as the primary key. After some adjustments, the type of the original scopedHashtable had been changed due to inappropriate type design, the type of primary key was changed to `slang::ast::ValueSymbol`. It has nothing to do with PortSymbol, MultiPortSymbol, and other symbols. The sourceLocation class represents a location in the source code (or within a macro expression). Every element in AST has a unique location including ports. We even could decode it back to its original locations (line, column, file information in a file location or a macro location)......
The name and location could be both ways to help find when and where we can create the port operation in a Moore module. `slang::ast::Symbol` is the parent of all subdivided symbols defined in slang. Defining the type of the container as `slang::ast::symbol` is sufficient for this purpose. The direction information is still important. It is not accepted because we suspect that the conversions applicable in this particular case should not use the top-level class.

maerhart (commented):
I'm not familiar with all the subtleties of SystemVerilog, but here are some of my thoughts/observations.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -134,6 +134,32 @@ def ProcedureOp : MooreOp<"procedure", [
...
+def InOut: I32EnumAttrCase<"InOut", 2>;
+def Ref: I32EnumAttrCase<"Ref", 3>;
+
+def DirectionAttr: I32EnumAttr<"Direction", "Port direction",
+                                  [In, Out, InOut, Ref]>{
+  let cppNamespace = "circt::moore";
+}
+
+def PortOp : MooreOp<"port", [
+  MemoryEffects<[MemAlloc]>,
```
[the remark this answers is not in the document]
hailongSun2000: Refer to #6859

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
fabianschuiki: @cepheus69 @hailongSun2000  One thing I'm wondering is whether you actually need `moore.port` at the moment for your Moore-to-Core lowering. As far as I understand, Slang always creates nets and variables corresponding to the ports. And the ports feel like they are more about describing the outside-facing interface that you interact with when you instantiate the module, not the nets inside the module. Do you already support module instances in your Moore-to-Core lowering? If not, maybe we can skip `moore.port` for now and figure it out together with the ports/operands on a `moore.instance` operation.

hailongSun2000: > @cepheus69 @hailongSun2000 One thing I'm wondering is whether you actually need `moore.port` at the moment for your Moore-to-Core lowering. As far as I understand, Slang always creates nets and variables corresponding to the ports. And the ports feel like they are more about describing the outside-facing interface that you interact with when you instantiate the module, not the nets inside the module. Do you already support module instances in your Moore-to-Core lowering? If not, maybe we can skip `moore.port` for now and figure it out together with the ports/operands on a `moore.instance` operation.

@cepheus69 expends the original definition of `moore. instance`, we can handle the port connections. IR looks like `moore. instance "bar" @Bar(%a) -> (%b) : (!moore.logic) -> (!moore.logic)`. We also support `moore.module` to `hw.module` and `moore.instance` to `hw.instance`.
The idea is to extend `moore.port` support without changing the code, so we create a new container rather than tweak the `ValueSymbols`.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -151,6 +177,28 @@ def VariableOp : MooreOp<"variable", [
...
+    "result", "assignment", "$_self">,
+]> {
+  let summary = "A net declaration";
+  let description = [{
+    See IEEE 1800-2017 § 6.7 "Net declarations".
+  }];
+  let arguments = (ins
+    StrAttr:$name,
+    StrAttr:$kind,
+    Optional<UnpackedType>:$assignment
```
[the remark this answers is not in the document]
hailongSun2000: > Maybe you could add a short paragraph to the description about what `kind` is (e.g., can it be any string or only a few selected ones?) and why an optional `assignment` operand is supported rather than using `NetOp + AssignOp` (maybe that's obvious to people more familiar with SystemVerilog, but could still help newcomers). In the tests, only "wire" is used as a kind. Are there only a limited number of kinds, and could an enum be used, or does it actually have to be a string? Is this necessary because of how user defined nettypes are modeled in slang?

Sorry for the late reply.  Using optional `assignment` is for showing the difference between `net declaration assignment` and `continuous assignment`.
Please see IEEE standard 1800-2017 § 10.3.1 "The net declaration assignment"
image

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
fabianschuiki: In that case, could we just skip the port AST nodes (not create a `moore.port` op for them), but instead populate the input and output argument list of the `moore.module`, like `moore.module @Bar (%in: type, ...) -> (%out: type, ...)` to match the syntax of the instance?

Later on, as we expand our support for the weird non-ANSI/ANSI ports that Verilog allows, we can lower more complicated port constructs to these simple lists of input and output arguments of the module. We could also think about having only a single list of arguments, and using a special `!moore.out<T>` and `!moore.inout<T>` wrapper type to capture port directions. This would align fairly well with Verilog's model of being able to refer to a port by name and then assigning to it, and instances just being weird ways of translating from instance-site SSA values to module SSA values.

hailongSun2000: If I remember correctly, we can directly get port lists when we handle the module even though skipping the port AST nodes. But we have to expand the definition of the moore::SVModuleOp. Maybe this will take some time to produce a module with a single list of arguments and a special wrapper type to capture the port directions.

fabianschuiki: We could also limit ourselves to simple input and output ports, and make modules use the `(%in, ...) -> (%out, ...)` pattern that you already use for instances. When we see anything that isn't trivially this in the AST, we just error out and inform the user that their port pattern is not supported. We can then expand on this later on.

cepheus69:
The `portOp` will be removed. New module Op would handle the original function of `portOp`

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -146,7 +146,40 @@ def VariableOp : MooreOp<"variable", [
...
+  OptionalTypesMatchWith<"assigned value and variable types match",
+    "result", "assignment", "$_self">,
+]> {
+  let summary = "A net declaration";
+  let description = [{
+    The `moore.net' operation is a net declaration. Net types defines different types
+    of net connection in SV. There are twelve built-in net types defined in the official
+    standard construct of the operation:
+    `supply0`, `supply1`, `tri`, `triand`, `trior`, `trireg`, `tri0`, `tri1`, `uwire`,
+    `wire`, `wand`, `wor`.
```
fabianschuiki, line 165: Since these twelve net types are part of the SV spec, could we represent them as an enum attribute instead of a string, for a less stringly-typed IR. Something like `I32EnumAttr` or, if it exists, even `I8EnumAttr` could do the trick. I think we had something like that floating around for port directions in the original PR.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -146,7 +146,40 @@ def VariableOp : MooreOp<"variable", [
   let arguments = (ins StrAttr:$name, Optional<UnpackedType>:$initial);
   let results = (outs Res<UnpackedType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
-    `` custom<ImplicitSSAName>($name) ($initial^)? attr-dict
+    custom<ImplicitSSAName>($name) ($initial^)? attr-dict
```
fabianschuiki, line 149: I think the `` is necessary here as a hint to the MLIR printer that it should not insert a space here. (ImplicitSSAName already adds a space.) Otherwise you get a double-space after the operation name, like `moore.var  name "foo" ...`. Same below for the NetOp.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -146,7 +146,40 @@ def VariableOp : MooreOp<"variable", [
...
+    `:` type($result)
+  }];
+}
+
+def NetOp : MooreOp<"net", [
```
fabianschuiki, line 154: Since you are adding a new dialect operation, please also add a parsing/printing test to `test/Dialect/Moore/basic.mlir`, ideally one test for each possible form of the declaration. With and without name. With and without the optional assignment. And then the twelve net types. `VarOp` should already have a similar test, but without the net type.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -45,6 +45,20 @@ module Basic;
   var v0;
   int v1;
   int v2 = v1;
+  
+  // CHECK: [[TMP0:%.+]] = moore.constant 0 : !moore.int
+  // CHECK: [[TMP1:%.+]] = moore.conversion [[TMP0]] : !moore.int -> !moore.packed<range<logic<signed>, 31:0>>
+  // CHECK: [[TMP2:%.+]] = moore.conversion [[TMP1]] : !moore.packed<range<logic<signed>, 31:0>> -> !moore.logic
+  // CHECK: %i = moore.net  "wire" [[TMP2]] : !moore.logic
+  wire i = 0;
```
fabianschuiki, line 53: Could you turn this into a
```verilog
wire w0;
wire w1 = w0;
```
to avoid any of the other IR fluff (constant and conversion)? That should give you just two net ops.

Can you also add a section that explicitly checks the twelve net types? Something like
```verilog
supply0 w2;
supply1 w3;
tri w4;
...
```

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -45,6 +45,20 @@ module Basic;
...
+
+  // CHECK: %j = moore.net "wire" : !moore.packed<range<logic, 15:0>>
+  wire [15:0] j;
+
+  // CHECK: [[TMP3:%.+]] = moore.constant 12 : !moore.packed<range<bit, 15:0>>
+  // CHECK: [[TMP4:%.+]] = moore.conversion [[TMP3]] : !moore.packed<range<bit, 15:0>> -> !moore.packed<range<logic, 15:0>>
+  // CHECK: moore.assign %j, [[TMP4]] : !moore.packed<range<logic, 15:0>>
+  assign j = 16'd12;
```
fabianschuiki, line 61: I don't think this tests anything new once you have a `wire w1 = w0;`, so this test can probably be deleted. The `w1 = w0` will check that wires partake in name resolution. And assignments are already checked in a separate part of the file.

fabianschuiki (commented):
Cool! Thanks for pushing on this :+1: Added a few comments.

cepheus69:
> The netOp is mainly used for declaring NetKind elements in the module. The NetKind has: `Unknown`,`Wire`,`WAnd`,`WOr`,`Tri`,`TriAnd`,`TriOr`,`Tri0`,`Tri1`,`TriReg`,`Supply0`, `Supply1`,`UWire`,~`Interconnect`~,~`UserDefined`~

Mark `Unknow` `Interconnect` `Userdefined` as not support!

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -35,6 +35,40 @@ convertProcedureKind(slang::ast::ProceduralBlockKind kind) {
...
+  case slang::ast::NetType::WAnd:
+    return moore::NetKind::WAnd;
+  case slang::ast::NetType::WOr:
+    return moore::NetKind::WOr;
+  case slang::ast::NetType::Interconnect:
+  case slang::ast::NetType::Unknown:
+  case slang::ast::NetType::UserDefined:
+    break;
+  }
+  llvm_unreachable("unsupported net kind");
```
fabianschuiki, line 69: `llvm_unreachable` is weird in that it doesn't really print an error message in non-debug/non-assertion builds. It will just hint the compiler at this branch being unreachable in release builds. However, you do want to report an error to the user here. Something like the following:
```
static FailureOr<moore::NetKind> convertNetKind(slang::ast::NetType::NetKind kind, Location loc) {
  switch (kind) {
  // ...
  case slang::ast::NetType::Interconnect:
  case slang::ast::NetType::Unknown:
  case slang::ast::NetType::UserDefined:
    return mlir::emitError(loc, "unsupported net type `") << slang::ast::toString(kind) << "`";
  }
}
```

test/Dialect/Moore/basic.mlir
```diff
@@ -25,6 +52,10 @@ moore.module @Foo {
 
   // CHECK: moore.assign %v1, %v2 : !moore.bit
   moore.assign %v1, %v2 : !moore.bit
+  // CHECK: moore.assign %w0, %w11 : !moore.logic
+  moore.assign %w0, %w11 : !moore.logic
+  // CHECK: moore.assign %w0, %w12 : !moore.logic
+  moore.assign %w0, %w12 : !moore.logic
```
fabianschuiki, line 58:
```suggestion
```
I don't think you need these, because the parser for `moore.assign` doesn't really know where `%v1` and `%v2` come from, and so the fact that `%v1` comes from `moore.variable` and `%w0` comes from `moore.net` doesn't really affect it, or cover any new code paths.

cepheus69:
> LGTM! Thanks 🥳

Thanks a lot !!! ❤️ The improved module & instance will be ready soon.

PR #6886, opened 2024-04-01
[NFC] Massive Export Verilog Speedup

darthscsi:
Getting all the ports to lookup one attribute is very expensive.  Mainly from GetAllPortLocs requiring several lookups and allocations.  Just avoid doing this by not using the summary functions and looking up individual ports directly (as should generally be done to avoid the overly broad expensive functions).  Granted that getPort goes through the summary functions isn't obvious or ideal.  It should be fixed.

### DIFF
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -230,8 +230,8 @@ static StringRef getInputPortVerilogName(Operation *module, size_t portArgNum) {
   auto hml = cast<HWModuleLike>(module);
   auto pId = hml.getHWModuleType().getPortIdForInputId(portArgNum);
   if (auto attrs = dyn_cast_or_null<DictionaryAttr>(hml.getPortAttrs(pId)))
-    if (auto updatedName = attrs.get("hw.verilogName"))
-      return updatedName.cast<StringAttr>().getValue();
+    if (auto updatedName = attrs.getAs<StringAttr>("hw.verilogName"))
+      return updatedName.getValue();
   return hml.getHWModuleType().getPortName(pId);
 }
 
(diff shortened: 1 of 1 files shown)


lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -228,8 +228,11 @@ static StringRef getPortVerilogName(Operation *module, size_t portArgNum) {
 /// Return the verilog name of the port for the module.
 static StringRef getInputPortVerilogName(Operation *module, size_t portArgNum) {
   auto hml = cast<HWModuleLike>(module);
-  return hml.getPort(hml.getHWModuleType().getPortIdForInputId(portArgNum))
-      .getVerilogName();
+  auto pId = hml.getHWModuleType().getPortIdForInputId(portArgNum);
+  if (auto attrs = dyn_cast_or_null<DictionaryAttr>(hml.getPortAttrs(pId)))
+    if (auto updatedName = attrs.get("hw.verilogName"))
+      return updatedName.cast<StringAttr>().getValue();
```
seldridge, line 234: This may be slightly more compact as:

```suggestion
    if (auto updatedName = attrs.getAs<StringAttr>("hw.verilogName"))
      return updatedName.getValue();
```

This trades a `cast` for a `dyn_cast_or_null` which may be slightly slower due to the extra non-assert checks.

seldridge (approved):

dtzSiFive (approved):
Change LGTM but just seconding that yes this points out there's definitely room for improvement so we don't have to hand-roll this sort of thing to avoid performance bugs lurking in the friendlier accessors.
This and the internal use of deprecated accessors, let's see if we can't chart a path out and see this through 👍 .

Thanks for work on all this, and of course specifically for tracking this down and fixing it!

PR #6887, opened 2024-04-02
[LowerToHW] Wrap signed operands of `PrintFOp` in sv `$signed()`

ubfx:
When we have a firrtl printf operation with `SInt` operands, it gets lowered something like this:

```verilog
//FIRRTL version 3.3.0
//circuit Test :
// module Test :
//    input clock : Clock
//    input reset : UInt<1>
//    input a : SInt<8>
//    node int = SInt<8>(-0h10)
//
//    printf(clock, UInt<1>(0h1), "int : %d %d\n", int, add(a, int)) : printf_1

[...]
  input [7:0] a
[...]
$fwrite(32'h80000002, "int : %d %d\n", 8'hF0, {a[7], a} - 9'h10);
[...]
```

As far as verilog is concerned, the operands to `$fwrite` are unsigned, which means that we get inconvenient printing for negative decimals. This patch wraps signed operands of `firrtl.printf` in `$signed()` during lowering. This is realized as `sv.system "signed" ...` for now. In the next step, we could introduce a separate `sv.signed()` Op and insert this in places where we need this conversion.

With this patch, the above firrtl gets lowered like this:
```verilog
$fwrite(32'h80000002, "int : %d %d\n", $signed(8'hF0),
                $signed({a[7], a} - 9'h10));
```

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -1416,6 +1416,7 @@ struct FIRRTLLowering : public FIRRTLVisitor<FIRRTLLowering, LogicalResult> {
   Value getLoweredNonClockValue(Value value);
   Value getLoweredAndExtendedValue(Value value, Type destType);
   Value getLoweredAndExtOrTruncValue(Value value, Type destType);
+  Value getLoweredFmtOperand(Value operand);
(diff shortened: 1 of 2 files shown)


uenoku (commented):
This looks reasonable to me. Format string is not limited to only PrintF but also assert/assume so could you please update there as well?

uenoku (approved):
LGTM

PR #6888, opened 2024-04-02
[Pipeline] Remove `Pure` trait from Pipeline operations

mortbopet:
Pipelines may be side-effectful - remove `Pure` trait to prevent such pipelines from being DCE'd during canonicalization.


blakep-msft (approved):
LGTM, thanks for doing this.

PR #6890, opened 2024-04-02
[FIRRTL] Expose clock dividers as a FIRRTL intrinsic

nandor:
This PR adds an intrinsic to expose `seq.clock_div` to FIRRTL.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerIntrinsics.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerIntrinsics.cpp
@@ -30,6 +30,14 @@
 using namespace circt;
 using namespace firrtl;
 
+// Get parameter by the given name.  Null if not found.
+static ParamDeclAttr getNamedParam(ArrayAttr params, StringRef name) {
+  for (auto param : params.getAsRange<ParamDeclAttr>())
+    if (param.getName().getValue().equals(name))
+      return param;
+  return {};
+}
+
 namespace {
 
 class CirctSizeofConverter : public IntrinsicConverter {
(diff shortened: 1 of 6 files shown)


lib/Dialect/FIRRTL/Transforms/LowerIntrinsics.cpp
```diff
@@ -161,6 +169,32 @@ class CirctClockInverterConverter : public IntrinsicConverter {
...
+  LogicalResult convert(InstanceOp inst) override {
+    auto pow2 = cast<IntegerAttr>(
+        getNamedParam(mod.getParameters(), "POW_2").getValue());
+
+    ImplicitLocOpBuilder builder(inst.getLoc(), inst);
+    auto in = builder.create<WireOp>(inst.getResult(0).getType()).getResult();
+    inst.getResult(0).replaceAllUsesWith(in);
+    auto out = builder.create<ClockDividerIntrinsicOp>(in, pow2);
+    auto name = inst.getInstanceName();
+    Value outWire = builder.create<WireOp>(out.getType(), name).getResult();
```
dtzSiFive, line 190: This extra wire only exists to carry the name given to the instance, is that right?

What sorts of names end up here, could they be handled by Chisel inserting a node with the desired name (or not, if naming it doesn't matter)?

Asking because:

I'm working to move to intrinsic expression/statements vs instance-of-intmodule -- where instead of lowering to a wire the user would generate something like:
```
node myname = intrinsic clock_divider.... 
```
(syntax details pending)

Does that generally seem like it works for the use case motivating these wires being inserted (here and clock inverter IIRC)?

dtzSiFive (approved):
LGTM!

Small comment about the wires being inserted here, since they're only done for two intrinsics and because I'm mid re-working how a bunch of this works but nothing blocking 👍 .

lib/Dialect/FIRRTL/Transforms/LowerIntrinsics.cpp
nandor: That would be great, what he have here are pretty much hacks to preserve a `sv.namehint` and push it into the op later.

dtzSiFive: Great, thanks for the explanation/context and discussion offline 👍 .

PR #6891, opened 2024-04-02
[FIRRTL] Split intrinsic op into expr vs stmt.

dtzSiFive:
Expression form is Pure and must have a result.
Optionally carries a name (like other expressions).

FIRRTL operations are expected to be one of expression or statement (see Visitor, firrtl::isExpression, so on) so split the combined operation into these two kinds.

Statement form is not pure and does not have a result.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
@@ -20,7 +20,7 @@ include "FIRRTLDialect.td"
 //===----------------------------------------------------------------------===//
 
 def GenericIntrinsicExprOp : FIRRTLOp<"int.generic.expr",
-    [HasCustomSSAName]> {
+    [HasCustomSSAName, Pure]> {
   let summary = "Generic intrinsic operation for FIRRTL intrinsic expressions";
 
   let arguments = (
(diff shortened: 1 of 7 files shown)


include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
```diff
@@ -16,23 +16,35 @@
 include "FIRRTLDialect.td"
 
 //===----------------------------------------------------------------------===//
-// Generic intrinsic operation for parsing into before lowering.
+// Generic intrinsic operations for parsing into before lowering.
 //===----------------------------------------------------------------------===//
 
-def GenericIntrinsicOp : FIRRTLOp<"int.generic",
+def GenericIntrinsicExprOp : FIRRTLOp<"int.generic.expr",
     [HasCustomSSAName]> {
```
uenoku, line 23: Is it necessary to mark this pure if we use it as an expression?

dtzSiFive: I must've lost that in my shuffling/rebasing, good catch thank you.

dtzSiFive:
Thanks for review, looks like keeping single op for now.  Closing.

PR #6893, opened 2024-04-03
[FIRRTL][CAPI] Allow constructing integers larger than 64 bits

SpriteOvO:
We are building `firrtl.constant` IRs from Chisel through MLIR, `mlirIntegerAttrGet` is used to construct constants, however it only accepts a `int64_t` value as an argument, so constants larger than 64 bits are not possible to be constructed at the moment.

Upstream report: https://github.com/llvm/llvm-project/issues/84190#issuecomment-2035552035

Refer to @mikeurbach's PR #6787, I propose to likewise add a string hack to workaround this problem. Using this new added C-API `firrtlAttrGetIntegerFromString`, users (e.g. Chisel) could construct a `MlirIntegerAttr` from a big int string.

### DIFF
--- a/include/circt-c/Dialect/FIRRTL.h
+++ b/include/circt-c/Dialect/FIRRTL.h
@@ -146,6 +146,11 @@ MLIR_CAPI_EXPORTED MlirAttribute firrtlAttrGetMemDir(MlirContext ctx,
 MLIR_CAPI_EXPORTED MlirAttribute
 firrtlAttrGetEventControl(MlirContext ctx, FIRRTLEventControl eventControl);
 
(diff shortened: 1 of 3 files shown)


mikeurbach (commented):
I think this is reasonable given the current limitations. Can you please add some tests of positive and negative numbers both larger and smaller than what fits in 64 bits?

mikeurbach (approved):
Thanks for adding the tests. This looks good to me to get something going.

SpriteOvO:
Tested working fine in Chisel binder, thanks to Mike's review.

PR #6897, opened 2024-04-04
[FIRRTL] Generic intrinsic parsing/emitter support

dtzSiFive:
Add parsing and emitter support for FIRRTL intrinsic.

Refactor parameter parsing on (ext/int)modules as these use same syntax.

Snippet from parse-basic.fir without FileCheck directives:

```firrtl
circuit GenericIntrinsics:
  public module GenericIntrinsics:
    input clock : Clock
    input data : UInt<32>
    input c : UInt<1>

    ; Statements
    intrinsic(circt_verif_assert, c)
    intrinsic(circt_fpga_probe, data, clock)

    ; Expressions
    node n = intrinsic(circt_plusargs_value<FORMAT = "foo"> : { found : UInt<1>, result : UInt<5> })
    node n2 = intrinsic(circt_plusargs_test<FORMAT = "bar"> : UInt<3>)

    ; Statement with unused return value.
    intrinsic(circt_clock_gate : Clock, clock, c)

    ; Nesting, statement and expression forms.
    intrinsic(circt_verif_assert, intrinsic(circt_isX: UInt<1>, data))
```

And somewhat similarly from the emit-basic test:
```
  public module GenericIntrinsic :
    input clk : Clock
    input c : UInt<1>
    output s : UInt<32>
    output io1 : UInt<1>
    output io2 : UInt<1>
    output io3 : UInt<1>
    output io4 : UInt<5>

    connect s, intrinsic(circt_sizeof : UInt<32>, clk)
    connect io1, intrinsic(circt_isX : UInt<1>, clk)
    connect io2, intrinsic(circt_plusargs_test<FORMAT = "foo"> : UInt<1>)
    intrinsic(circt_clock_gate : Clock, clk, c)
    node _gen_int
      = intrinsic(circt_plusargs_value<FORMAT = "foo">
                    : { found : UInt<1>, result : UInt<5> })
    connect io3, _gen_int.found
    connect io4, _gen_int.result
    intrinsic(circt_verif_assert, intrinsic(circt_isX : UInt<1>, c))
    node _gen_int_0 = intrinsic(circt_isX : UInt<1>, c)
    intrinsic(circt_verif_assert, intrinsic(circt_isX : UInt<1>, _gen_int_0))
 ```

### DIFF
--- a/lib/Dialect/FIRRTL/Export/FIREmitter.cpp
+++ b/lib/Dialect/FIRRTL/Export/FIREmitter.cpp
@@ -62,6 +62,10 @@ struct Emitter {
   void emitDeclaration(OptionOp op);
   void emitEnabledLayers(ArrayRef<Attribute> layers);
 
+  void emitParamAssign(ParamDeclAttr param, Operation *op,
+                       std::optional<PPExtString> wordBeforeLHS = std::nullopt);
+  void emitGenericIntrinsic(GenericIntrinsicOp op);
+
   // Statement emission
   void emitStatementsInBlock(Block &block);
   void emitStatement(WhenOp op);
@@ -91,6 +95,7 @@ struct Emitter {
   void emitStatement(RefReleaseOp op);
   void emitStatement(RefReleaseInitialOp op);
   void emitStatement(LayerBlockOp op);
+  void emitStatement(GenericIntrinsicOp op);
 
   template <class T>
   void emitVerifStatement(T op, StringRef mnemonic);
@@ -120,6 +125,7 @@ struct Emitter {
   void emitExpression(DoubleConstantOp op);
   void emitExpression(ListCreateOp op);
   void emitExpression(UnresolvedPathOp op);
+  void emitExpression(GenericIntrinsicOp op);
 
   void emitPrimExpr(StringRef mnemonic, Operation *op,
                     ArrayRef<uint32_t> attrs = {});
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: lib/Dialect/FIRRTL/Export/FIREmitter.cpp [hunk 1]
```diff
@@ -1087,6 +1134,20 @@ void Emitter::emitStatement(InvalidValueOp op) {
   emitLocationAndNewLine(op);
 }
 
+void Emitter::emitStatement(GenericIntrinsicOp op) {
+  startStatement();
+  if (op.use_empty())
+    emitGenericIntrinsic(op);
+  else {
```
fabianschuiki, line 1141: Could we also emit the intrinsic inline if it has a single use?

fabianschuiki (approved):
Nice refactoring of the parameter parsing! Looks good to me. Gonna let others comment on the syntax.

lib/Dialect/FIRRTL/Export/FIREmitter.cpp
dtzSiFive: See `isEmittedInline` -- and the answer is "nearly" (I had it implemented that way at first) -- but only if we know the use will be emitted (exactly) once.

To address this, the idea is to only emit it inline if it's used exactly once and that use itself is not emitted inline.

Probably this should be pulled into a pre-pass that inserts declarations as needed and generates names for declarations that don't have any, so on.  That's left as future work and this just tries to get good output using a simple but correct heuristic.  But please look carefully :+1:.

dtzSiFive: On the subject, this "inline" suitability is arguably (?) not quite right re:re-ordering operations that are side-effecting if the appear in one order but are used in a different order.  I don't think it's worth worrying about that as what they "effect" is outside the design's dataflow and state and is unreachable without some effort, but just a thought.  I think all our intrinsics are pure when expression-like (have a result, and possibly a use) presently.

fabianschuiki: Oh yeah I hadn't considered that this might be emitted as part of a larger expression that potentially gets inlined multiple times. Better be conservative. And your point about inlining affecting the order of execution is also a good one. I know we don't really give any guarantees in FIRRTL, but no need to artificially move stuff around for no reason. Emission into a node looks great!

mikeurbach (commented):
I took a quick look through it, and the implementation looks good to me. I'm happy to approve as-is, not sure if anyone else wants to weigh in or have a deeper design review. To me, the fact that the parsing logic is relatively straightforward and not going through crazy gyrations means the syntax is effective.

dtzSiFive:
Added commit moving to the left-paren keyword specifically so that this is safe w.r.t. backwards-compatibility with designs that parse today as it only changes what happens when someone writes `intrinsic(` which is never legal presently.

rwy7 (approved):

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -3307,6 +3396,74 @@ ParseResult FIRStmtParser::parseRWProbeStaticRefExp(FieldRef &refResult,
...
+  locationProcessor.setLoc(loc);
+
+  auto op = builder.create<GenericIntrinsicOp>(
+      type, builder.getStringAttr(intrinsic), operands, parameters);
+  if (type)
+    result = op.getResult();
+  return success();
+}
+
+/// params ::= '<' param param* '>'
```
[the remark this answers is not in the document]
dtzSiFive: Oh, you're right.  I wrote that with expectation that at least one was required if you had the `<`/`>` but that's not what's implemented and not sure that's useful to bother to reject.  Thanks for catching this, will apply your suggestion 👍 .

mikeurbach (approved):

dtzSiFive:
Landing, thanks for the reviews/feedback all!

Syntax is not official, but let's keep this train moving and modify as needed 👍

PR #6900, opened 2024-04-05
[SMT] Minor width related fixes for BitVectorAttr

fzi-hielscher:
Some minor changes to the construction of BitVector attributes in the SMT dialect:

- Fix parsing of `smt.bv.constant #smt.bv<-1> : !smt.bv<1>` which currently trips the width check due to `odsParser.parseInteger` creating an unnecessarily wide `APInt`. The new logic is copy-pasted from the FIRRTL ConstantOp parser. See #6794.  
- Change the type of the attribute builder's `value` argument from `unsigned` to `uint64_t` matching the signature of the `APInt` constructor, to allow values up to 64 bits and avoid architecture dependent behavior.
- Prevent left-shifts wider than (or equal to) the shifted operand's number of bits in the width checking logic to avoid undefined behavior.

### DIFF
--- a/lib/Dialect/SMT/SMTAttributes.cpp
+++ b/lib/Dialect/SMT/SMTAttributes.cpp
@@ -86,16 +86,16 @@ BitVectorAttr::getChecked(function_ref<InFlightDiagnostic()> emitError,
   return Base::getChecked(emitError, context, *maybeValue);
 }
 
-BitVectorAttr BitVectorAttr::get(MLIRContext *context, unsigned value,
+BitVectorAttr BitVectorAttr::get(MLIRContext *context, uint64_t value,
                                  unsigned width) {
   return Base::get(context, APInt(width, value));
 }
 
 BitVectorAttr
 BitVectorAttr::getChecked(function_ref<InFlightDiagnostic()> emitError,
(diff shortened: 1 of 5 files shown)


lib/Dialect/SMT/SMTAttributes.cpp
```diff
@@ -117,14 +117,20 @@ Attribute BitVectorAttr::parse(AsmParser &odsParser, Type odsType) {
...
+    val = val.sext(width);
+  } else if (width < val.getBitWidth()) {
+    // The parser can return an unnecessarily wide result.
+    // This isn't a problem, but truncating off bits is bad.
+    unsigned neededBits =
+        val.isNegative() ? val.getSignificantBits() : val.getActiveBits();
+    if (width < neededBits) {
       odsParser.emitError(loc)
-          << "integer value out of range for given bit-vector type";
+          << "integer value out of range for given bit-vector type " << odsType;
```
maerhart, line 133: Expect of the added comments and the slightly changed error message, nothing changed here, or am I overlooking something?

maerhart (approved):
LGTM. Thanks for fixing this!

PR #6901, opened 2024-04-05
[FIRRTL] Use a flag to implement scalarization of internal modules.  …

darthscsi:
…This is precursor to making lower-types a module-pass and optional.  This will deprecate the flags to lower-types and make aggregate preservation a function of NOT running lower-types, rather than having to control it to ensure ports are correct.


prithayan (approved):
Looks good.

PR #6902, opened 2024-04-08
[SMT] Add lowering to LLVM IR

maerhart:
* Add the pass boilerplate, helper functionality, and lowerings for the first few operations.
* Disallow functions as operands of `smt.eq` and `smt.distinct` because that's not supported by Z3. In Z3Py and the SMT-LIB format it is possible to compare them, but not in a symbolic manner. Those comparisons are just lowered to a structural (syntactic) expression comparison (`Z3_is_eq_ast` or `Z3_is_eq_func_decl` instead of `Z3_mk_eq` in the Z3 C API). If needed it is probably better to have a separate operation for that.


lib/Conversion/SMTToZ3LLVM/LowerSMTToZ3LLVM.cpp
```diff
@@ -0,0 +1,925 @@
...
+  // Z3_ast Z3_API Z3_mk_and(Z3_context c, unsigned num_args, Z3_ast const
+  // args[])
+  // ```
+  // The API function requires num_args > 0. This is not guaranteed by the
+  // `smt.and` operation and thus the pattern will not apply when no operand is
+  // present. The constant folder of the operation is assumed to fold this to
+  // a constant 'true' (neutral element of AND).
+  ADD_VARIADIC_PATTERN(AndOp, "Z3_mk_and", 1);
+
+#undef ADD_VARIADIC_PATTERN
```
fabianschuiki, line 849: Not sure what the CIRCT/MLIR/LLVM policy is for using macros, but this could possibly be just a templated lambda, or, if there are only two uses of `ADD_VARIADIC_PATTERN`, it might be more readable to just write out the `patterns.add ...` calls.

fabianschuiki (approved):
I have only limited experience with the Z3 API, but the conversion looks good to me. It's great to see this come to life and for us to be able to express SMT interactions in CIRCT. And not just simple checks, but actual executable programs that interact with the solver. I can't wait to start making use of this :smiley:

lib/Conversion/SMTToZ3LLVM/LowerSMTToZ3LLVM.cpp
maerhart: I would opt for just leaving the macros for now until we update to C++20 and can then replace them with templated lambdas for the following reasons:
* I couldn't find anything in the LLVM coding standards and programmers manual that forbids macros.
* All the visitors in CIRCT use macros in this exact same way. They are very local and self-contained, which shouldn't lead to debugging or readability issues.
* AFAIK templated lambdas are a C++20 feature, but CIRCT uses C++17. 
* As you have already seen in the follow-up PR, quite a few lowering patterns use those macro shortcuts. Inlining all of them would lead to decreased readability and make it more cumbersome to, i.e., add a constructor argument in the future.

fabianschuiki: Yeah makes total sense to me! And you're right, this looks very much like our established uses in the visitors :sweat_smile:

Issue #6840, opened 2024-03-18
[Comb] Missed optimization opportunity for multiply used boolean operations 


uenoku:
```mlir
hw.module @Baz(in %cond : i2, in %value : i2, out o1: i2, out o2: i2) {
  %0 = comb.and bin %cond, %value : i2
  %1 = comb.and bin %0, %value : i2
  hw.output %0, %1: i2, i2
}
```

Current output (canonicalization + export-verilog):
```verilog
// Generated by CIRCT unknown git version
module Baz(     // nya.mlir:1:1
  input  [1:0] cond,    // nya.mlir:1:19
               value,   // nya.mlir:1:34
  output [1:0] o1,      // nya.mlir:1:51
               o2       // nya.mlir:1:63
);

  wire [1:0] _GEN = cond & value;       // nya.mlir:2:8
  assign o1 = _GEN;     // nya.mlir:2:8, :4:3
  assign o2 = _GEN & value;     // nya.mlir:2:8, :3:8, :4:3
endmodule
```

`o2` could be just `_GEN` because `value` is a duplication. This is currently not optimized because `_GEN` is not flattened into `o2` since `_GEN` is multiply used.

dtzSiFive:
This looks to be fixed by 6903 -- are there other opportunities to address, or okay to close?

uenoku:
Sorry for delayed response, yes let me close this. https://github.com/llvm/circt/pull/6903

PR #6903, opened 2024-04-08
[Comb] Remove more idempotent operands

hovind:
I'm unsure if this is a good way to go about this, but I wanted to try my hand at #6840. I'm unsure if this is a good way to go about the problem, any feedback would be greatly appreciated.

The example provided by @uenoku in #6840 is
```
hw.module @Baz(in %cond : i2, in %value : i2, out o1: i2, out o2: i2) {
  %0 = comb.and bin %cond, %value : i2
  %1 = comb.and bin %0, %value : i2
  hw.output %0, %1: i2, i2
}
```
which canonicalizes to
```
$ ./bin/circt-opt --canonicalize bin/baz.mlir
module {
  hw.module @Baz(in %cond : i2, in %value : i2, out o1 : i2, out o2 : i2) {
    %0 = comb.and bin %cond, %value : i2
    %1 = comb.and bin %0, %value : i2
    hw.output %0, %1 : i2, i2
  }
}
```
instead of
```
module {
  hw.module @Baz(in %cond : i2, in %value : i2, out o1 : i2, out o2 : i2) {
    %0 = comb.and bin %cond, %value : i2
    hw.output %0, %0 : i2, i2
  }
}
```
Currently `template <typename Op> bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter)` makes sure that a value can appear _maximum one time_ in `op`'s operands. This patch makes `template <typename Op> bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter)` consider whether any of `op`s operands are themselves defined by an `Op`, whose operands we can omit from `op`'s operands.

In the example above, that means

1. `%1` is a `comb.and` whose operands are `%0`, `%value`
2. `%0` is an `comb.and`, matching `%1`s operation
3. `%value` and be removed from `%1`, since input `0` of `%1` is a `comb.and` with a `%value` operand

### DIFF
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -936,8 +936,8 @@ template <typename Op>
 static bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter) {
   auto inputs = op.getInputs();
 
-  llvm::SmallSetVector<Value, 8> alreadyCovered;
-  llvm::SmallSetVector<Value, 8> checked;
+  llvm::SmallSetVector<Value, 8> uniqueInputs(inputs.begin(), inputs.end());
+  llvm::SmallDenseSet<Value, 8> checked;
   checked.insert(op);
 
   llvm::SmallVector<Value, 8> worklist;
@@ -948,20 +948,17 @@ static bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter) {
 
   while (!worklist.empty()) {
     auto element = worklist.pop_back_val();
-    checked.insert(element);
 
     if (auto idempotentOp = element.getDefiningOp<Op>()) {
       for (auto input : idempotentOp.getInputs()) {
-        alreadyCovered.insert(input);
+        uniqueInputs.remove(input);
 
-        if (!checked.contains(input))
+        if (checked.insert(input).second)
           worklist.push_back(input);
       }
     }
   }
 
-  llvm::SmallSetVector<Value, 8> uniqueInputs(inputs.begin(), inputs.end());
-  uniqueInputs.set_subtract(alreadyCovered);
   if (uniqueInputs.size() < inputs.size()) {
     replaceOpWithNewOpAndCopyName<Op>(rewriter, op, op.getType(),
                                       uniqueInputs.getArrayRef());
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -937,15 +937,26 @@ static bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter) {
   auto inputs = op.getInputs();
 
   llvm::SmallSetVector<Value, 8> alreadyCovered;
-  llvm::SmallVector<Value, 8> worklist(inputs.begin(), inputs.end());
+  llvm::SmallSetVector<Value, 8> checked;
+  checked.insert(op);
+
+  llvm::SmallVector<Value, 8> worklist;
+  for (auto input : inputs) {
+    if (input != op)
+      worklist.push_back(input);
+  }
+
   while (!worklist.empty()) {
     auto element = worklist.pop_back_val();
+    checked.insert(element);
 
     if (auto idempotentOp = element.getDefiningOp<Op>()) {
-      alreadyCovered.insert(idempotentOp.getInputs().begin(),
-                            idempotentOp.getInputs().end());
-      worklist.append(idempotentOp.getInputs().begin(),
-                      idempotentOp.getInputs().end());
+      for (auto input : idempotentOp.getInputs()) {
+        alreadyCovered.insert(input);
+
+        if (!checked.contains(input))
+          worklist.push_back(input);
+      }
     }
   }
 
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -935,11 +935,22 @@ static Value getCommonOperand(Op op) {
 template <typename Op>
 static bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter) {
   auto inputs = op.getInputs();
-  llvm::SmallSetVector<Value, 8> uniqueInputs;
 
-  for (const auto input : inputs)
-    uniqueInputs.insert(input);
+  llvm::SmallSetVector<Value, 8> alreadyCovered;
+  llvm::SmallVector<Value, 8> worklist(inputs.begin(), inputs.end());
+  while (!worklist.empty()) {
+    auto element = worklist.pop_back_val();
 
+    if (auto idempotentOp = element.getDefiningOp<Op>()) {
+      alreadyCovered.insert(idempotentOp.getInputs().begin(),
+                            idempotentOp.getInputs().end());
+      worklist.append(idempotentOp.getInputs().begin(),
+                      idempotentOp.getInputs().end());
+    }
+  }
+
+  llvm::SmallSetVector<Value, 8> uniqueInputs(inputs.begin(), inputs.end());
+  uniqueInputs.set_subtract(alreadyCovered);
   if (uniqueInputs.size() < inputs.size()) {
     replaceOpWithNewOpAndCopyName<Op>(rewriter, op, op.getType(),
                                       uniqueInputs.getArrayRef());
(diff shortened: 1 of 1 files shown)


lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -935,11 +935,22 @@ static Value getCommonOperand(Op op) {
...
+      worklist.append(idempotentOp.getInputs().begin(),
+                      idempotentOp.getInputs().end());
+    }
+  }
 
+  llvm::SmallSetVector<Value, 8> uniqueInputs(inputs.begin(), inputs.end());
+  uniqueInputs.set_subtract(alreadyCovered);
   if (uniqueInputs.size() < inputs.size()) {
     replaceOpWithNewOpAndCopyName<Op>(rewriter, op, op.getType(),
                                       uniqueInputs.getArrayRef());
```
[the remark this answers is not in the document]
hovind: Thank you so much for taking the time to review and leave thoughtful comments!

I may be misunderstanding here, but did you see lines 952, 953?

To my understanding, `set_subtract(alreadyCovered)` should ensure that the proposed `uniqueInputs` is _smaller than_ or _as small as_ the previous `uniqueInputs`.

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -962,8 +973,9 @@ LogicalResult AndOp::canonicalize(AndOp op, PatternRewriter &rewriter) {
     return success();
 
   // and(..., x, ..., x) -> and(..., x, ...) -- idempotent
+  // and(..., x, and(..., x, ...)) -> and(..., and(..., x, ...)) -- idempotent
   // Trivial and(x), and(x, x) cases are handled by [AndOp::fold] above.
-  if (size > 2 && canonicalizeIdempotentInputs(op, rewriter))
+  if (size > 1 && canonicalizeIdempotentInputs(op, rewriter))
     return success();
```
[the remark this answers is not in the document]
hovind: > Could it make sense to tweak your implementation to not inline nested and ops, but to filter out any and operands that were already covered by a nested one? For example, by doing a pass over all nested and ops, adding their operands into a set, and then doing a second pass to remove all operands from the main and op that are already in that set.

I think that makes a whole lot of sense, and that is what I attempted to do in this PR. I had some difficulties wording the pull request text, and the attempted explanation turned out very poor.

> You could also use that set of already covered operands to prevent re-visiting operations through the worklist if you've already seen them, which would prevent the canonicalizer from looping infinitely if the IR has a combinational loop in it.

I had not thought about combinational loops, oops!

My initial thought was to do

```diff
diff --git a/lib/Dialect/Comb/CombFolds.cpp b/lib/Dialect/Comb/CombFolds.cpp
index 0cb0464ad..cda924a0b 100644
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -936,10 +936,14 @@ template <typename Op>
 static bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter) {
   auto inputs = op.getInputs();
 
-  llvm::SmallSetVector<Value, 8> alreadyCovered;
   llvm::SmallVector<Value, 8> worklist(inputs.begin(), inputs.end());
+  llvm::SmallSetVector<Value, 8> alreadyCovered;
+  alreadyCovered.insert(op);
+
   while (!worklist.empty()) {
     auto element = worklist.pop_back_val();
+    if (alreadyCovered.contains(element))
+      continue;
 
     if (auto idempotentOp = element.getDefiningOp<Op>()) {
       alreadyCovered.insert(idempotentOp.getInputs().begin(),
```

but that prevents us from recursing to `%0` in when canonicalizing `%2` in
```
hw.module @Baz(in %a : i2, in %b : i2, in %c : i2, out o1: i2, out o2: i2, out o3: i2) {
  %0 = comb.and bin %a, %b : i2
  %1 = comb.and bin %0, %c : i2
  %2 = comb.and bin %1, %a : i2
  hw.output %0, %1, %2: i2, i2, i2
}
```
resulting in
```
module {
  hw.module @Baz(in %a : i2, in %b : i2, in %c : i2, out o1 : i2, out o2 : i2, out o3 : i2) {
    %0 = comb.and bin %a, %b : i2
    %1 = comb.and bin %0, %c : i2
    %2 = comb.and bin %1, %a : i2
    hw.output %0, %1, %2 : i2, i2, i2
  }
}
```
instead of
```
module {
  hw.module @Baz(in %a : i2, in %b : i2, in %c : i2, out o1 : i2, out o2 : i2, out o3 : i2) {
    %0 = comb.and bin %a, %b : i2
    %1 = comb.and bin %0, %c : i2
    hw.output %0, %1, %1 : i2, i2, i2
  }
}
```

I've pushed a proposal to 29b793b0eeda292f4f2ac3bb79d4dfdd236a47da, but it didn't turn out as pretty as I would like, and some advice on cleaning it up would be greatly appreciated.

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -935,11 +935,33 @@ static Value getCommonOperand(Op op) {
...
-  llvm::SmallSetVector<Value, 8> uniqueInputs;
 
-  for (const auto input : inputs)
-    uniqueInputs.insert(input);
+  llvm::SmallSetVector<Value, 8> alreadyCovered;
+  llvm::SmallSetVector<Value, 8> checked;
+  checked.insert(op);
 
+  llvm::SmallVector<Value, 8> worklist;
+  for (auto input : inputs) {
```
hovind, line 944: Make sure `op` doesn't make it into the worklist.

fabianschuiki: I think this makes sense! Having a set of checked nodes around to prevent recursion is a good pattern :smiley: Let me add a few comments.

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -935,11 +935,33 @@ static Value getCommonOperand(Op op) {
 template <typename Op>
 static bool canonicalizeIdempotentInputs(Op op, PatternRewriter &rewriter) {
   auto inputs = op.getInputs();
-  llvm::SmallSetVector<Value, 8> uniqueInputs;
 
-  for (const auto input : inputs)
-    uniqueInputs.insert(input);
+  llvm::SmallSetVector<Value, 8> alreadyCovered;
+  llvm::SmallSetVector<Value, 8> checked;
```
fabianschuiki, line 940: Since you only need `checked` to be a set, but you don't care about preserving the order in which elements were added:
```suggestion
  llvm::SmallDenseSet<Value, 8> checked;
```

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -935,11 +935,33 @@ static Value getCommonOperand(Op op) {
...
+
+    if (auto idempotentOp = element.getDefiningOp<Op>()) {
+      for (auto input : idempotentOp.getInputs()) {
+        alreadyCovered.insert(input);
+
+        if (!checked.contains(input))
+          worklist.push_back(input);
+      }
+    }
+  }
```
fabianschuiki, line 961: The `insert` method on sets and maps returns a boolean indicating whether the element was inserted (true), or whether there was already an element in the set and nothing was added (false):
```suggestion
  while (!worklist.empty()) {
    auto element = worklist.pop_back_val();

    if (auto idempotentOp = element.getDefiningOp<Op>()) {
      for (auto input : idempotentOp.getInputs()) {
        alreadyCovered.insert(input);

        if (checked.insert(input))
          worklist.push_back(input);
      }
    }
  }
```

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -935,11 +935,33 @@ static Value getCommonOperand(Op op) {
...
+      worklist.push_back(input);
+  }
+
+  while (!worklist.empty()) {
+    auto element = worklist.pop_back_val();
+    checked.insert(element);
+
+    if (auto idempotentOp = element.getDefiningOp<Op>()) {
+      for (auto input : idempotentOp.getInputs()) {
+        alreadyCovered.insert(input);
```
fabianschuiki, line 955: The following should do the trick without requiring you to build up the `alreadyCovered` set:
```suggestion
        uniqueInputs.remove(input);
```

lib/Dialect/Comb/CombFolds.cpp
hovind: Thanks! :raised_hands:

lib/Dialect/Comb/CombFolds.cpp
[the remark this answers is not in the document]
hovind: Good point, thanks!

lib/Dialect/Comb/CombFolds.cpp
hovind: Much better :smiley:

hovind:
Thanks for the very instructive pointers, @fabianschuiki. I pushed your suggestions to fd972d2dc1d95bfded157443d61135e8c17b8293.

fabianschuiki (approved):
Really cool! LGTM

hovind:
> Really cool! LGTM

Thanks for helping out, makes it much easier to attempt to contribute as a beginner! Feel free to merge at your convenience, or wait for others to chime in if you think that would be helpful. :)

PR #6904, opened 2024-04-08
[FIRRTL]  Use lower-sigs to handle port scalarization of internal modules.  This removes Lower Types from dealing with it in anticipation of Lower Types being simplified.

darthscsi:


dtzSiFive (commented):
Context/plan is appreciated (thanks!) but this diff looks like a PR that would be titled "Use scalarized convention on internal modules by default"?

Which is first picked up by LowerSignatures, but it already is handling that and this PR does not change that as much as by setting the convention it changes behavior when used with non-default `-preserve-aggregate` options and modulo whether LowerSignatures and LowerTypes do quite the same thing for signatures (IIRC not precisely).

The lack of test churn is surprising -- apparently nothing tests or depends on the default behavior? (was that included in previous PR?)

Anyway change looks fine, modulo being clear about the change in behavior that might exist between LowerSignatures and LowerTypes.  I thought there was some but LMK if that's not the case.

FWIW, depending on a list of things -- mixing agg-preservation of signatures and bodies may change or break what works/doesn't re:probes-of-aggregate usage (esp rwprobe's).

I'm curious what the end-goal for the options provided and functionality is here (and perhaps motivation for this overall direction?) -- is the plan to continue the decoupling of "aggregate preservation" of bodies from signatures (already done at "boundaries")?   Let compiler pick what (non-ABI-wise) is preserved as it sees fit?

PR #6906, opened 2024-04-09
[ImportVerilog] Add replicate and extract operations.

hailongSun2000:
`moore.replicate` is the extended op for `moore.concate`. 
`moore.extract` represents a single bit or range bits selection, and it can receive a variable as its operand. And for the range bits selection, it has three `RangeSelectionKind`, `Simple`, `IndexedUp`, and `IndexedDown`. Respective example, 
simple_vec[7:4], up_vec[3+:1]/up_vec[3*variable+:1], down_vec[3-:1]/down_vec[3*variable-:1]. 

For the simple_vec, the variable won't as its operands, and for the up_vec and down_vec, the LHS could be non-constant(like 3*variable), but RHS must be a constant, and it represents the width. 
These limitations are defined by SV IEEE 1800-2017 § 11.5.1 and § 11.5.2.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -361,10 +361,58 @@ struct ExprVisitor {
     for (auto *operand : expr.operands()) {
       auto value = context.convertExpression(*operand);
       if (!value)
-        return {};
+        continue;
+      value = convertToSimpleBitVector(value);
       operands.push_back(value);
     }
     return builder.create<moore::ConcatOp>(loc, operands);
+  }
+
+  // Handle replications.
+  Value visit(const slang::ast::ReplicationExpression &expr) {
+    auto type = context.convertType(*expr.type);
+    if (isa<moore::VoidType>(type))
+      return {};
+
+    auto value = context.convertExpression(expr.concat());
+    if (!value)
+      return {};
+    return builder.create<moore::ReplicateOp>(loc, type, value);
+  }
+
+  // Handle single bit selections.
+  Value visit(const slang::ast::ElementSelectExpression &expr) {
+    auto type = context.convertType(*expr.type);
+    auto value = context.convertExpression(expr.value());
+    auto lowBit = context.convertExpression(expr.selector());
+
+    if (!value || !lowBit)
+      return {};
+    return builder.create<moore::ExtractOp>(loc, type, value, lowBit);
+  }
+
+  // Handle range bits selections.
+  Value visit(const slang::ast::RangeSelectExpression &expr) {
+    auto type = context.convertType(*expr.type);
+    auto value = context.convertExpression(expr.value());
+    Value lowBit;
+    if (expr.getSelectionKind() == slang::ast::RangeSelectionKind::Simple) {
+      if (expr.left().constant && expr.right().constant) {
+        auto lhs = expr.left().constant->integer().as<uint64_t>().value();
+        auto rhs = expr.right().constant->integer().as<uint64_t>().value();
+        lowBit = lhs < rhs ? context.convertExpression(expr.left())
+                           : context.convertExpression(expr.right());
+      } else {
+        mlir::emitError(loc, "unsupported a variable as the index in the")
+            << slang::ast::toString(expr.getSelectionKind()) << "kind";
+        return {};
+      }
+    } else
+      lowBit = context.convertExpression(expr.left());
+
+    if (!value || !lowBit)
+      return {};
+    return builder.create<moore::ExtractOp>(loc, type, value, lowBit);
   }
 
   /// Emit an error for all other expressions.
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -656,4 +656,65 @@ def ConcatOp : MooreOp<"concat", [
   }];
 }
 
+def ReplicateOp : MooreOp<"replicate", [
+  Pure
+]> {
+  let summary = "Multiple concatenation of expressions";
+  let description = [{
+    This operation indicates a joining together of that many copies of the 
+    concatenation `{constant{w}}`. Which enclosed together within brace.
+    The 'constant' must a non-negative, non-x, and non-z constant expression.
+    The 'constant' may be a value of zero, but it only exists in parameterized
+    code, and it will be ignored(type is changed to the void).
+
+    Example:
+    ```
+      {0{w}}   // empty! ignore it.
+      {4{w}}   // the same as {w, w, w, w}
+    ```
+    See IEEE 1800-2017 §11.4.12 "Concatenation operators".
+  }];
+  let arguments = (ins SimpleBitVectorType:$value);
+  let results = (outs SimpleBitVectorType:$result);
+  let assemblyFormat = [{
+    $value attr-dict `:` functional-type($value, $result)
+  }];
+}
+
+def ExtractOp : MooreOp<"extract"> {
+  let summary = "Addressing operation";
+  let description = [{
+    This operation includes the vector bit/part-select, array, and memory 
+    addressing.If the address is invalid--out of bounds or has x or z bit--
+    then it will produce x for 4-state or 0 for 2-state.
+    Bit-select results are unsigned, regardless of the operands.
+    Part-select results are unsigned, regardless of the operands even if 
+    the part-select specifies the entire vector.
+    See IEEE 1800-2017 § 11.8.1 "Rules for expression types"
+
+    Example:
+    ```
+    logic v [7:0];
+    v[1];                      // the bit-select addressing
+    v[3:0];                    // the part-select addressing
+    v[3-:4];  v[0+:4];         // They are equivalent to v[3:0]
+    ```
+    See IEEE 1800-2017 § 11.5.1 "Vector bit-select and part-select addressing".
+
+    Example:
+    ```
+    // an array of 256-by-256 8-bit elements
+    logic [7:0] twod_array [0:255][0:255];
+    logic [7:0] mem_name [0:1023];      // a memory of 1024 8-bit words
+    ```
+    See IEEE 1800-2017 § 11.5.2 "Array and memory addressing".
+  }];
+  let arguments = (ins UnpackedType:$input, UnpackedType:$lowBit);
+  let results = (outs UnpackedType:$result);
+  let assemblyFormat = [{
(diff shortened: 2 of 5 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -361,12 +361,56 @@ struct ExprVisitor {
     for (auto *operand : expr.operands()) {
       auto value = context.convertExpression(*operand);
       if (!value)
-        return {};
+        continue;
```
hailongSun2000, line 364: Example:
```
  int a;
  assign a = {{0{32'b1}, 32'b0};
```
We just ingore `{0{32'b1}}`, then continue to handle `32'b0`, rather than directly `return{}`.
Please see IEEE 1800-2017§11.4.12.1 "Replication operator"
image

But there is no test case, I'll add it when supporting `void` type.

hailongSun2000:
Thanks for the code review in advance! :smiley:

lib/Conversion/ImportVerilog/Expressions.cpp
uenoku: Can't we just use i0 constant for a zero replication constant?

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -361,12 +361,56 @@ struct ExprVisitor {
...
+  // Handle single bit selections.
+  Value visit(const slang::ast::ElementSelectExpression &expr) {
+    auto type = context.convertType(*expr.type);
+    auto value = context.convertExpression(expr.value());
+    auto lowBit = context.convertExpression(expr.selector());
+
+    if (!value || !lowBit)
+      return {};
+    return builder.create<moore::ExtractOp>(loc, type, value, lowBit);
+  }
```
fabianschuiki, line 392: What happens if the expression is an array, not an integer? I think it would make sense to have different operations for the different types that the bit select expressions can operate on. For example, accessing a single bit in an integer is likely going to be quite a bit different than accessing a single element in an array.

It would be good if this code looked at the type of the value being bit-selected into, handled the specific types that it understands, and errors out on all other cases. (You want the IR to have very clear semantics and reject any input that you can't really handle early on.)

How about adding a `moore.array_get` operation, similar to `hw.array_get`, which can operate on packed/unpacked arrays? And leave `moore.extract` to operate on integers. Then this code would check the type of `expr.value()` and generate `extract` for integers, `array_get` for arrays, and an error otherwise.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -361,12 +361,56 @@ struct ExprVisitor {
...
+    auto type = context.convertType(*expr.type);
+    auto value = context.convertExpression(expr.value());
+    Value lowBit;
+    if (expr.getSelectionKind() == slang::ast::RangeSelectionKind::Simple) {
+      if (expr.left().constant && expr.right().constant) {
+        auto lhs = expr.left().constant->integer().as<uint64_t>().value();
+        auto rhs = expr.right().constant->integer().as<uint64_t>().value();
+        lowBit = lhs < rhs ? context.convertExpression(expr.left())
+                           : context.convertExpression(expr.right());
+      }
```
fabianschuiki, line 405: You probably need an else branch here in case left and/or right is non-constant. If this is a `RangeSelectionKind::Simple`, but one of the operands is non-constant, `lowBit` remains null and the code fails silently.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -361,12 +361,56 @@ struct ExprVisitor {
...
+    } else
+      lowBit = context.convertExpression(expr.left());
+
+    if (!value || !lowBit)
+      return {};
+    return builder.create<moore::ExtractOp>(loc, type, value, lowBit);
```
fabianschuiki, line 411: Does this do the right thing for integers or arrays that don't start at `0`? For example in `bit [7:4] x; x[5:4]`, the range select `[5:4]` should select the lowest and second-lowest bit in `x`. Is this shifting part of `ExtractOp`? Or does this have to be handled here?

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -361,12 +361,56 @@ struct ExprVisitor {
...
+    Value lowBit;
+    if (expr.getSelectionKind() == slang::ast::RangeSelectionKind::Simple) {
+      if (expr.left().constant && expr.right().constant) {
+        auto lhs = expr.left().constant->integer().as<uint64_t>().value();
+        auto rhs = expr.right().constant->integer().as<uint64_t>().value();
+        lowBit = lhs < rhs ? context.convertExpression(expr.left())
+                           : context.convertExpression(expr.right());
+      }
+    } else
+      lowBit = context.convertExpression(expr.left());
```
fabianschuiki, line 407: What happens with the right-hand side expression? If I write `x[a:b]`, this just stores `x` and `a` into the op, but `b` gets lost. Shouldn't that go somewhere in the op? I think it would be good to just reject all the cases here that we can't really fully handle. It's totally fine to say that we only support constant left/right expressions for now. Or that we only support `[a+:k]` or `[a-:k]` if it has a constant k expression. We then probably want different ops for the different cases (extract with two constants, extract with constant width but dynamic offset, extract where both sides are dynamic -- probably very last).

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: `i0` is permitted in FIRRTL, but an error will be thrown when existing in SV. I'll add the related error test cases for it.
Like this `logic l1 = {{0{1'b1}}};` Slang will throw an error: "concatenation cannot be empty".

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: Yes, it can also work. We can select a range of bits arbitrarily only if we follow big endian and little endian. For example, if we select bits like `[4:5]` from`[7:4]`, it will trigger an error: endianness of selection must match declared range (type is 'logic[3:0]')

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: In Slang AST, the RHS like `k` just represents the width, we don't need to capture it(must be a positive constant, and x/z/non-constant will trigger errors).
In your case x[a:b](belong to `RangeSelectionKind::Simple`), firstly, `a` and `b` must be constant, then they are meaningful to capture to distinguish big and little endian like `[0:3]` and `[3:0]`. When we select a range of bits like `[1:2]` from `[0:3]` and `[3:2]` from `[3:0]`, the `lowBit` is`1` and `2` separately.
`else` statement serves at `RangeSelectionKind::IndexedUp` and `RangeSelectionKind::IndexedDown`.

hailongSun2000: I'll modify the git message for `moore.extract`, it's not accurate right now.

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: Please let me know if I misunderstood. 
We can support selection from `int a;`(like `a[1]` or `a[var]`), `arr[3:1];`(like `arr[1]` or `arr[var]`), and the type of `arr` can be `predefinedWidthType`(like int, byte, and so on), `moore.packedArray` and `moore.unpackedArray`. But we need to handle the special situation when a variable as `lowBit` like `logic [3:0] l1;    logic [1:0] l2;   logic l3 = l2[l1];`

lib/Conversion/ImportVerilog/Expressions.cpp
uenoku: System Verilog doesn't support but SV dialect support i0. CIRCT removes i0 while ExportVerilog so I think it's OK to use i0.

fabianschuiki: I don't think we can really treat the `void` type as a simple bit vector type. As far as I know, it's a pretty isolated type in the SV spec. We might have to add special treatment for replication operators nested in concatenations, to support zero-width repetitions. They can't exist as isolated expressions, so we probably shouldn't try to turn them into an expression at all. Maybe in the lowering for concatenations we can check if a concatenated value is a repetition, and then add special handling for the zero-width case.

If we switched the Moore dialect over to actually have a simple bit vector type, like `!moore.logic<16>`, we _could_ just lower zero repetitions to `!moore.logic<0>`. Maybe that's worth a shot.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: What happens if you do something like `a[b:c]`, where `b` and `c` are both variables?

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: That sounds great, so the user gets a good error message. The code should still have an else branch, in case Slang changes behaviour or we see unforeseen input, and just error out with something like `error: unsupported indexing` or similar.

hailongSun2000: You're right! I'll append an else branch.

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: `a[b:c]` will trigger an error: reference to non-constant variable 'c' is not allowed in a constant expression. This error will be generated by either variable `b` or `c`.

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: The related IR:
```
%x = moore.variable : !moore.packed<range<bit, 7:4>>
%0 = moore.constant 4 : !moore.int
%1 = moore.extract %x from %0 : !moore.packed<range<bit, 7:4>>, !moore.int -> !moore.packed<range<bit, 5:4>>
```
And when we lower `moore.extract` into `comb.extract`, we will use `comb.shr` to guarantee to extract the correct bits.

hailongSun2000:
Hey, @fabianschuiki. Are there any questions? If not, can I merge this PR after updating the commit history?

fabianschuiki (approved):
LGTM! Thanks for adding this :partying_face:

PR #6907, opened 2024-04-09
[ImportVerilog] Fix the types mismatch for variable declarations.

hailongSun2000:
When declaring a variable with an initial value, like `bit b = 1'b1;`, or `logic [3:0] l1; logic [1:0] l2 = l1[3:2];`. These all will trigger an error: 'moore.variable' op failed to verify that initial value and variable types match.

### DIFF
--- a/test/Conversion/ImportVerilog/basic.sv
+++ b/test/Conversion/ImportVerilog/basic.sv
@@ -46,6 +46,12 @@ module Basic;
   int v1;
   int v2 = v1;
 
+  // CHECK: %b1 = moore.variable : !moore.packed<range<bit, 0:0>>
+  // CHECK: [[TMP:%.+]] = moore.conversion %b1 : !moore.packed<range<bit, 0:0>> -> !moore.bit
+  // CHECK: %b2 = moore.variable [[TMP]] : !moore.bit
+  bit [0:0] b1;
+  bit b2 = b1;
+
   // CHECK: moore.procedure initial {
   // CHECK: }
   initial;
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -82,6 +82,9 @@ struct MemberVisitor {
       initial = context.convertExpression(*init);
       if (!initial)
         return failure();
+
+      if (initial.getType() != type)
+        initial = builder.create<moore::ConversionOp>(loc, type, initial);
     }
 
     auto varOp = builder.create<moore::VariableOp>(


hailongSun2000:
Thanks for the code review in advance.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -46,6 +46,11 @@ module Basic;
   int v1;
   int v2 = v1;
 
+  // CHECK: [[TMP1:%.+]] = moore.constant true : !moore.packed<range<bit, 0:0>>
+  // CHECK: [[TMP2:%.+]] = moore.conversion [[TMP1]] : !moore.packed<range<bit, 0:0>> -> !moore.bit
+  // CHECK: %b1 = moore.variable [[TMP2]] : !moore.bit
+  bit b1 = 1'b1;
```
fabianschuiki, line 52: Could you turn this into
```
logic b1;
bit b2 = b1;
```
such that we have a very explicit type written out, and don't rely on the implicit type of the int literal `1'b1`?

hailongSun2000: Please let me know if I misunderstood anything. 
This error isn't triggered if we have a very explicit type like your example. But `3:2`-vs-`1:0` also exists. Like this
```
  logic [3:0] l1;
  logic [1:0] l2 = l1[3:2];
```
```json
{
            "name": "l1",
            "kind": "Variable",
            "addr": 2199025897768,
            "type": "logic[3:0]",
            "lifetime": "Static"
          },
          {
            "name": "l2",
            "kind": "Variable",
            "addr": 2199025898144,
            "type": "logic[1:0]",
            "initializer": {
              "kind": "RangeSelect",
              "type": "logic[3:2]",
              "selectionKind": "Simple",
              "value": {
                "kind": "NamedValue",
                "type": "logic[3:0]",
                "symbol": "2199025897768 l1"
              },
              "left": {
                "kind": "IntegerLiteral",
                "type": "int",
                "value": "3",
                "constant": "3"
              },
              "right": {
                "kind": "IntegerLiteral",
                "type": "int",
                "value": "2",
                "constant": "2"
              }
            },
```
Note: This case relies on range bits selection.

fabianschuiki: Oh I see, in my `b2 = b1` example Slang directly inserts a conversion in the AST, such that your code doesn't run. Good point. Does the following work?
```
bit [0:0] b1;
bit b2 = b1;
```

hailongSun2000: The same as above. There is no type conversion in the Slang AST.
image

hailongSun2000: Update the test case using your example.

fabianschuiki (approved):
LGTM! Thanks :slightly_smiling_face:

hailongSun2000:
> LGTM! Thanks 🙂

It's my honor! :wink:

PR #6908, opened 2024-04-09
[circt-lec] Port to SMT dialect based compiler pipeline

maerhart:


TaoBi22 (approved):
Looks great to me, very exciting!!

PR #6909, opened 2024-04-10
[Ibis] Divorce port name (hints) from port symbol names

mortbopet:
Port symbols now have no effect on the eventual port name of a module. This also implies that multiple ports can have the same namehint, and will be uniqued when lowering `ibis.container` to `hw.module` (see new test in `containers_to_hw.mlir`).

### DIFF
--- a/include/circt/Dialect/Ibis/IbisOps.td
+++ b/include/circt/Dialect/Ibis/IbisOps.td
@@ -749,10 +749,10 @@ class PortLikeOp<string mnemonic, list<Trait> traits = []> :
     related, and all references to a port is done through the port symbol.
   }];
 
-  let arguments = (ins StrAttr:$name, InnerSymAttr:$inner_sym, TypeAttrOf<AnyType>:$type);
+  let arguments = (ins OptionalAttr<StrAttr>:$name, InnerSymAttr:$inner_sym, TypeAttrOf<AnyType>:$type);
   let results = (outs PortRefType:$port);
   let assemblyFormat = [{
-    $name `sym` $inner_sym `:` $type attr-dict
+    ($name^)? `sym` $inner_sym `:` $type attr-dict
   }];
 
   let extraClassDeclaration = [{
@@ -785,9 +785,6 @@ class PortLikeOp<string mnemonic, list<Trait> traits = []> :
   code extraPortClassDefinition = ?;
   let extraClassDefinition =  extraPortClassDefinition # [{
     LogicalResult $cppClass::verify() {
-      if(getName().empty())
-        return emitOpError("port name cannot be empty");
-
       if(getType().isa<ScopeRefType>())
         return emitOpError("port type cannot be a scope reference");
 
--- a/include/circt/Dialect/Ibis/IbisOps.td
+++ b/include/circt/Dialect/Ibis/IbisOps.td
@@ -742,10 +742,17 @@ class PortLikeOp<string mnemonic, list<Trait> traits = []> :
       HasCustomSSAName,
       DeclareOpInterfaceMethods<InnerSymbol,["getTargetResultIndex"]>
   ])> {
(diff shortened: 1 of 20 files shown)


lib/Dialect/Ibis/Transforms/IbisTunneling.cpp
```diff
@@ -296,7 +296,8 @@ LogicalResult Tunneler::tunnelDown(InstanceGraphNode *currentContainer,
   llvm::DenseMap<StringAttr, OutputPortOp> outputPortOps;
   for (PortInfo &pi : portInfos) {
     outputPortOps[pi.portName] = rewriter.create<OutputPortOp>(
-        op.getLoc(), circt::hw::InnerSymAttr::get(pi.portName), pi.getType());
+        op.getLoc(), pi.portName, circt::hw::InnerSymAttr::get(pi.portName),
```
teqdruid, line 299: Unrelated to this PR, but will `circt::hw::InnerSymAttr::get(pi.portName)` create a unique symbol?

lib/Dialect/Ibis/Transforms/IbisContainersToHW.cpp
```diff
@@ -34,41 +35,54 @@ struct ContainerPortInfo {
...
     };
 
-    // Gather in and output port ops.
+    // Gather in and output port ops to define the hw.module interface. Here, we
+    // also perform uniqueing of the port names.
+    Namespace portNs;
     for (auto input : container.getBodyBlock()->getOps<InputPortOp>()) {
-      opInputs[input.getInnerSym().getSymName()] = input;
-
+      auto uniquePortName =
```
teqdruid, line 66: I may be wrong about this, but I don't think port names are required to be unique in the `hw.module` IR. IIRC ExportVerilog will take care of uniquing port names. In the IR, they are pretty much only name hints -- only order matters.

lib/Dialect/Ibis/Transforms/IbisContainersToHW.cpp
```diff
@@ -14,6 +14,7 @@
 #include "circt/Dialect/Ibis/IbisPasses.h"
 #include "circt/Dialect/Ibis/IbisTypes.h"
 
+#include "circt/Support/Namespace.h"
```
teqdruid, line 17: I don't see where this is used.

include/circt/Dialect/Ibis/IbisOps.td
```diff
@@ -778,6 +785,9 @@ class PortLikeOp<string mnemonic, list<Trait> traits = []> :
   code extraPortClassDefinition = ?;
   let extraClassDefinition =  extraPortClassDefinition # [{
     LogicalResult $cppClass::verify() {
+      if(getName().empty())
+        return emitOpError("port name cannot be empty");
```
teqdruid, line 789: Should we default to the symbol name?

teqdruid (approved):
Do you plan to do the same for containers in a subsequent PR?

lib/Dialect/Ibis/Transforms/IbisContainersToHW.cpp
mortbopet: l. 64 in this file

include/circt/Dialect/Ibis/IbisOps.td
mortbopet: In that case, i think the assembly format should be `ibis.port.input sym @foo` - that is, making the namehint optional. We can do that.

teqdruid: I don't feel strongly about it either way.

mortbopet:
> Do you plan to do the same for containers in a subsequent PR?
Yep. I think ports are highest priority for getting fixed, and then containers.

PR #6911, opened 2024-04-10
[FIRRTL][LowerIntrinsics] Add stat and preserve if no changes.

dtzSiFive:


uenoku (approved):
This looks safe to me. As far as I see applyPartialConversion doesn't start a thread though they are implemented as therad-safe.

dtzSiFive:
> This looks safe to me. As far as I see applyPartialConversion doesn't start a thread though they are implemented as therad-safe.

Thanks, @uenoku .  That's my concern -- they sure look like something that's meant to be used concurrently in theory if nothing else.

My other doubt was that patterns themselves don't know if a thing was converted -- only that they reported at this point it could be (and used the rewriter to express what that means).  So the number possibly could be off depending on the rewrite strategy / exploration, or more simply if ultimately conversion fails.

If this was functionally relevant that would be more concerning, but this should fail conservatively reliably for the latter and cause analyses to not be preserved which is what happens without this PR anyway.

And if this is a problem in the future or we regret it it's not a big deal to back out or just drop entirely.

The conversion framework DOES have the ability to let you track the set of operations converted but that seems kinda heavyweight for just counting.  Maybe it is preferable though? 🤔 .

PR #6912, opened 2024-04-11
Update annotation handling to fix partial field reset behavior

adkian-sifive:
If you have an initial value for a Register that includes `DontCare` for some of the fields, those fields will be excluded from the Full Async Reset Transform. While `InferResets` and `SFCCompat` separately do the right thing, `InferResets` runs before `SFCCompat` and removes the FART annotation. 

This PR adds additional handling for the FART annotation by letting `InferResets` pass through annotation

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
@@ -47,12 +47,13 @@ void SFCCompatPass::runOnOperation() {
   SmallVector<InvalidValueOp> invalidOps;
 
   bool fullAsyncResetExists = false;
-  AnnotationSet::removePortAnnotations(getOperation(), [&](unsigned argNum, Annotation anno) {
-    if (!anno.isClass(fullAsyncResetAnnoClass))
-        return false;
-    fullAsyncResetExists = true;
-    return true;
-  });
+  AnnotationSet::removePortAnnotations(
+      getOperation(), [&](unsigned argNum, Annotation anno) {
+        if (!anno.isClass(fullAsyncResetAnnoClass))
+          return false;
+        fullAsyncResetExists = true;
+        return true;
+      });
 
   auto result = getOperation()->walk([&](Operation *op) {
     // Populate invalidOps for later handling.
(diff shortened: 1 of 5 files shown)


test/firtool/async-reset-anno.fir
```diff
@@ -0,0 +1,19 @@
+; RUN: firtool %s -mlir-print-ir-after=firrtl-infer-resets 2>&1 | FileCheck %s --check-prefix POST-INFER-RESETS
+; RUN: firtool %s -mlir-print-ir-after=firrtl-sfc-compat 2>&1 | FileCheck %s --implicit-check-not POST-SFC-COMPAT-NOT
```
seldridge, line 2: It's a bit janky to do this with `firtool -mlir-print-ir-after`. How about this:

```suggestion
; RUN: firtool %s -parse-only | circt-opt -firrtl-infer-resets | FileCheck %s --check-prefix POST-INFER-RESETS
; RUN: firtool %s -parse-only | circt-opt -firrtl-infer-resets -firrtl-sfc-compat | FileCheck %s --implicit-check-not POST-SFC-COMPAT-NOT
```

seldridge: As a further alternative, this could be two checks on the passes individually:

1. An MLIR test that infer resets doesn't delete the annotation. (The inversion of the test that was removed.)
2. A test that SFC Compat does remove the annotation.

I'm fine with either the more end-to-end approach here or the very directed MLIR tests in (1) and (2).

lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
```diff
@@ -44,6 +45,16 @@ void SFCCompatPass::runOnOperation() {
...
   bool madeModifications = false;
   SmallVector<InvalidValueOp> invalidOps;
+
+  bool fullAsyncResetExists = false;
+  AnnotationSet::removePortAnnotations(
+      getOperation(), [&](unsigned argNum, Annotation anno) {
+        if (!anno.isClass(fullAsyncResetAnnoClass))
+          return false;
+        fullAsyncResetExists = true;
+        return true;
```
seldridge, line 55: Possibly less readable, but I think you can do:

```suggestion
        return fullAsyncResetExists = true;
```

seldridge (approved):
LGTM

A few ideas throughout.

@jackkoenig: Can you review the test case and see if there are any test cases that you have in mind that could be added?

test/firtool/async-reset-anno.fir
adkian-sifive: i might keep these tests, i think the end-to-end approach is slicker

PR #6914, opened 2024-04-11
 [LowerFirReg] Reimplement the mux reachability analysis

prithayan:
Reopened the PR  #6709 , after fixing the issue that required the merged
commit to be reverted.

This implements a new heuristic to determine if a Mux is reachable from a
 FirReg. In general an operation is reachable from a register if its in the
 fanout of the register. For FirReg lowering, an if/else structure is required
 for proper enable inference, if a mux is within the fanout from the register.
The fanout path can only consist of MuxOp, ArrayGetOp or ArrayCreateOp.
Thus any ops other than MuxOp, ArrayGetOp or ArrayCreateOp block the
 reachability. The analysis is built lazily when its queried and the result is
 cached to avoid redundant traversal of the IR.


dtzSiFive:
> What was the issue and fix here?

Windows builder specifically errored because the SmallVector's element size crossed a threshold.
(maybe `std::function` size? not sure)

CI build failure (Windows doesn't run on PR's but was promptly reverted until fixed (thanks!)):
https://github.com/llvm/circt/actions/runs/8622542911/job/23633892020

Looks like fix was doing what was asked by that error and explicitly specifying number of elements to allocate, here:

https://github.com/llvm/circt/pull/6914/commits/8350b0d7901791455ae1f81e041e81b33285776d

prithayan:
Thanks @dtzSiFive , that's right. The issue was default vector size, which was fixed in the commit.

uenoku:
Thanks, just wanted to make sure what was going on :) Thank you for the details.

PR #6916, opened 2024-04-12
[FIRRTL] Add --log-firrtl-dedup option

jackkoenig:
This option enables logging of all deduplicated modules.

Still needs a test but wanted to get some eyes on this.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -35,6 +35,8 @@
 #include "llvm/Support/Format.h"
 #include "llvm/Support/SHA256.h"
 
+#define DEBUG_TYPE "firrtl-dedup"
+
 using namespace circt;
 using namespace firrtl;
 using hw::InnerRefAttr;
@@ -1575,6 +1577,10 @@ struct DenseMapInfo<ModuleInfo> {
 
 namespace {
 class DedupPass : public DedupBase<DedupPass> {
+public:
+  using DedupBase::logDedup;
+
+private:
   void runOnOperation() override {
     auto *context = &getContext();
     auto circuit = getOperation();
@@ -1662,6 +1668,10 @@ class DedupPass : public DedupBase<DedupPass> {
     if (result.failed())
       return signalPassFailure();
 
+    if (logDedup) {
+      llvm::errs() << "FIRRTL Deduplication:\n";
+    }
+
     for (auto [i, module] : llvm::enumerate(modules)) {
       auto moduleName = module.getModuleNameAttr();
       auto &hashAndModuleNamesOpt = hashesAndModuleNames[i];
@@ -1695,6 +1705,10 @@ class DedupPass : public DedupBase<DedupPass> {
         deduper.dedup(original, module);
         ++erasedModules;
         anythingChanged = true;
+        if (logDedup) {
+          llvm::errs() << "- dedup " << module.getModuleNameAttr() << " into "
+                       << original.getModuleNameAttr() << "\n";
+        }
         continue;
       }
       // Any module not deduplicated must be recorded.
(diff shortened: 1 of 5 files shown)


jackkoenig:
Does this need a test? Do we usually test debug logging?

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -1662,6 +1664,8 @@ class DedupPass : public DedupBase<DedupPass> {
     if (result.failed())
       return signalPassFailure();
 
+    LLVM_DEBUG(llvm::dbgs() << "FIRRTL Deduplication:\n");
```
uenoku, line 1667: nit: Can you bring this to the start of `runOnOperation` and use `debugPassHeader(this)` that prints a nice header?

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -1695,6 +1699,10 @@ class DedupPass : public DedupBase<DedupPass> {
         deduper.dedup(original, module);
         ++erasedModules;
         anythingChanged = true;
+        LLVM_DEBUG({
+          llvm::dbgs() << "- dedup " << module.getModuleNameAttr() << " into "
+                       << original.getModuleNameAttr() << "\n";
```
uenoku, line 1704: `module` is already erased within `deduper.dedup` so it's necessary to emit log before line 1699.

uenoku:
> Does this need a test? Do we usually test debug logging?

We don't usually add a test for logging so I think it's ok.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
mikeurbach: You may need to build with ASAN/UBSAN or run with valgrind to notice issues here. Unfortunately we've not enabled that in CI, so if this leads to use after free, it could fly under the radar of your local testing / CI.

jackkoenig: Right that makes sense. Thanks for the catch @uenoku and the explanation @mikeurbach!

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
jackkoenig: Done, also added the corresponding footer at the end.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
jackkoenig: I went a slightly different way, using the `moduleName` variable and adding a new one for `originalName`, WDYT?

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -1691,10 +1696,15 @@ class DedupPass : public DedupBase<DedupPass> {
...
         // Record the group ID of the other module.
-        dedupMap[moduleName] = original.getModuleNameAttr();
+        auto originalName = original.getModuleNameAttr();
+        dedupMap[moduleName] = originalName;
         deduper.dedup(original, module);
         ++erasedModules;
         anythingChanged = true;
+        LLVM_DEBUG({
+          llvm::dbgs() << "- dedup " << moduleName << " into " << originalName
+                       << "\n";
```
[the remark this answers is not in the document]
jackkoenig: I will switch it to something like this. Should modules that don't dedup into anything or have anything dedup into them be included in the results?

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
uenoku: That looks great! :100:

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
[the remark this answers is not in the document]
seldridge: You're the primary consumer of this debugging information right now... so, I'll redirect the question to you: would it be useful to include information about modules which don't dedup with anything in your quest to roll out `Invalidate`? (Probably, yes.)

Issue #6913, opened 2024-04-11
[SMT] Add quantifier attribute and pattern support to ExportSMTLIB


maerhart:
The SMT operations `smt.forall` and `smt.exists` support attributes called `weight` and `no_pattern` to give hints to the solver backend (e.g., Z3). We should add support for them in the `ExportSMTLIB` pass. Currently, it just errors out when it encounters those attributes:

https://github.com/llvm/circt/blob/6d4939d89d55d0c3ef4001fe8afafa24997d3e9d/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp#L295-L301

Furthermore, the quantifier operations allow a variadic number of regions representing a multi-pattern each, i.e., each region should be emitted as one `:pattern` attribute such that each operand of the `smt.yield` operation in the region represents one pattern in the multi-pattern.

Information on how such attributes have to be annotated in SMT-LIB can be found in the [SMT-LIB standard version 2.6](https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf) sections 3.6.5 and 3.6.6

Z3-specific documentation of those features can be found here:
https://microsoft.github.io/z3guide/docs/logic/Quantifiers/#patterns
https://microsoft.github.io/z3guide/docs/logic/Quantifiers/#multi-patterns
https://microsoft.github.io/z3guide/docs/logic/Quantifiers/#no-patterns

luisacicolini:
Happy to work on this!

PR #6920, opened 2024-04-15
[SMT] Added weight attribute support for ExportSMTLIB

luisacicolini:
Answers [this issue](https://github.com/llvm/circt/issues/6913) by adding support for the `weight` attribute in ExportSMTLIB. Includes also a new test.

### DIFF
--- a/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
+++ b/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
@@ -292,10 +292,8 @@ struct ExpressionVisitor
   template <typename OpTy>
   LogicalResult quantifierHelper(OpTy op, StringRef operatorString,
                                  VisitorInfo &info) {
-    // TODO: add support
     auto weight = op.getWeight();
-    // if (op.getWeight() != 0)
-    //   return op.emitError() << "non-zero weights not supported yet";
+    // TODO: add support
     if (op.getNoPattern())
       return op.emitError() << "no-pattern attribute not supported yet";
     if (!op.getPatterns().empty())
@@ -339,7 +337,7 @@ struct ExpressionVisitor
     newInfo.stream.indent(newInfo.indentLevel);
     if (failed(printExpression(worklist, newInfo)))
       return failure();
-    
+
     info.stream << info.valueMap.lookup(yieldedValue);
 
     if(weight != 0 && newInfo.openParens > 0){
--- a/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
+++ b/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
@@ -293,8 +293,9 @@ struct ExpressionVisitor
   LogicalResult quantifierHelper(OpTy op, StringRef operatorString,
                                  VisitorInfo &info) {
     // TODO: add support
-    if (op.getWeight() != 0)
-      return op.emitError() << "non-zero weights not supported yet";
+    auto weight = op.getWeight();
+    // if (op.getWeight() != 0)
+    //   return op.emitError() << "non-zero weights not supported yet";
     if (op.getNoPattern())
       return op.emitError() << "no-pattern attribute not supported yet";
     if (!op.getPatterns().empty())
@@ -321,7 +322,10 @@ struct ExpressionVisitor
       delimiter = " ";
     }
 
+
     info.stream << ")\n";
+
+
 
     // Print the quantifier body. This assumes that quantifiers are not deeply
     // nested (at least not enough that recursive calls could become a problem).
@@ -334,9 +338,17 @@ struct ExpressionVisitor
     newInfo.stream.indent(newInfo.indentLevel);
     if (failed(printExpression(worklist, newInfo)))
       return failure();
+    if (weight != 0)
+      info.stream << "( ! ";
+
     info.stream << info.valueMap.lookup(yieldedValue);
+    if(weight != 0)
+      info.stream <<" "<< weight << ")";
+
     for (unsigned k = 0; k < newInfo.openParens + 1; ++k)
       info.stream << ")";
+
+
 
     return success();
   }
--- a/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
+++ b/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
@@ -320,7 +320,6 @@ struct ExpressionVisitor
       delimiter = " ";
     }
 
-
     info.stream << ")\n";
 
     if (weight != 0)
@@ -340,14 +339,13 @@ struct ExpressionVisitor
 
     info.stream << info.valueMap.lookup(yieldedValue);
 
-    for(int j=0;j<newInfo.openParens;++j)
-      info.stream <<")";
-
-    if(weight != 0){
-      info.stream <<" :weight "<< weight << "))";
-    } else{
+    for (int j = 0; j < newInfo.openParens; ++j)
       info.stream << ")";
-    }
+
+    if (weight != 0)
+      info.stream << " :weight " << weight << ")";
+
+    info.stream << ")";
 
     return success();
   }
--- a/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
+++ b/lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
@@ -348,18 +348,27 @@ struct ExpressionVisitor
     info.stream << info.valueMap.lookup(yieldedValue);
 
     // llvm::outs()<<"\n\n\nOpen Parens is "<<newInfo.openParens;
+    // llvm::outs()<<"\n\n\nIndent Level is "<<newInfo.indentLevel;
+    // llvm::outs()<<"\n\n\nValue Map is "<<newInfo.valueMap;
 
 
 
-    for (unsigned k = 0; k < newInfo.openParens; ++k)
-      info.stream << ")";
 
-    if(weight != 0 && newInfo.openParens > 0)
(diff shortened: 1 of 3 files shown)


maerhart (commented):
Thanks for working on this!
That's a really good start already. I think we should add one integration test to check that Z3 does not complain about the emitted expression structure, etc. There are already a few in the `integration_test/Target/ExportSMTLIB` directory that you could use as a template.

luisacicolini:
Thanks a lot @maerhart for the feedback! Will work on the integration tests today :)

luisacicolini:
Happy to add more tests if necessary! I also changed the management of parentheses depending on the nesting of `let`

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -334,10 +339,28 @@ struct ExpressionVisitor
...
+
+    // printExpression(worklist, newInfo);
+
     info.stream << info.valueMap.lookup(yieldedValue);
-    for (unsigned k = 0; k < newInfo.openParens + 1; ++k)
+
+    // llvm::outs()<<"\n\n\nOpen Parens is "<<newInfo.openParens;
+
+    if(weight != 0 && newInfo.openParens > 0)
+      info.stream <<") :weight "<< weight << ")";
```
maerhart, line 353: What if there is more than one `let` expression inside the quantifier? Don't we need to close all of them before the `:weight` attribute then?

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -334,10 +339,28 @@ struct ExpressionVisitor
...
+    else if (weight != 0)
+      info.stream <<" :weight "<< weight << "))";
+    else
+      info.stream << ")";
+
+    for (unsigned k = 0; k < newInfo.openParens; ++k)
```
maerhart, line 359: There might be a bug in my code above (the initialization of `newInfo`) where we inherit the number of open parens from outside the quantifier, but here we don't subtract that. Maybe we should initialize `openParens` for `newInfo` to 0. But this probably only affects your change if the number of let expressions outside the quantifier is more than 1.
(not expecting you to fix my bugs; just mentioning it such that you are aware of it)

test/Target/ExportSMTLIB/attributes.mlir
```diff
@@ -0,0 +1,37 @@
...
+  // CHECK:         [[V10]]))
+
+  // CHECK-INLINED: (assert (forall (([[A:.+]] Int) ([[B:.+]] Int))
+  // CHECK-INLINED:                 (= [[A]] [[B]])))
+  %1 = smt.forall ["a", "b"] {
+  ^bb0(%arg2: !smt.int, %arg3: !smt.int):
+    %2 = smt.eq %arg2, %arg3 : !smt.int
+    smt.yield %2 : !smt.bool
+  }
+  smt.assert %1
```
maerhart, line 19: This test already exists in another file. Instead, we could also add a weight attribute here.

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
luisacicolini: as far as I understood from the docs yes: `:weight` should be after the `let ... expr` where `expr` is the body of the quantifier, no matter how many other expressions/`let`s are nested in there

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
luisacicolini: I'm not sure I understood what you mean, why should we subtract the number of parens from outside?

maerhart: Such that we don't close `let` scopes from outside the quantifier too early. The `assert` should close them later on instead.
You can try this example to trigger the issue:
```
  %2 = smt.exists {
  ^bb0(%arg2: !smt.int, %arg3: !smt.int):
    %3 = smt.eq %arg2, %arg3 : !smt.int
    smt.yield %3 : !smt.bool
  }
  %3 = smt.exists {
  ^bb0(%arg2: !smt.int, %arg3: !smt.int):
    %5 = smt.eq %arg2, %arg3 : !smt.int
    smt.yield %5 : !smt.bool
  }
  %5 = smt.and %2, %3
  smt.assert %5
```

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
maerhart: Yes, that's correct! But what happens if there is one more `let` inside the `!`?
I.e., like this: ` (assert (let (( tmp1 ( exists ((x Int) (y Int)) ( ! ( let ((tmp2 (exists ((z Int)) (= z y)))) (let ((tmp3 (and tmp2 tmp2))) tmp3)) :weight 2)))) tmp1))`
Now there need to be two closing parens before `:weight`, but if I understand the code correctly, only one would be emitted currently.

luisacicolini: Oh I see what you mean now! I'll try to generate one such example and see what happens

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
luisacicolini: Clear, thanks a lot!

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -334,9 +337,21 @@ struct ExpressionVisitor
...
-      info.stream << ")";
+
+    if(weight != 0 && newInfo.openParens > 0){
+      for(int j=0;j<newInfo.openParens;j++)
+        info.stream <<")";
+      info.stream <<" :weight "<< weight << "))";
+    } else if (weight != 0){
+      info.stream <<" :weight "<< weight << "))";
+      for(int j=0;j<newInfo.openParens;j++)
+        info.stream <<")";
```
maerhart, line 350: Isn't this dead code because `newInfo.openParens` will always be `<= 0` here?

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -334,9 +337,21 @@ struct ExpressionVisitor
...
+    } else{
+      for (unsigned k = 0; k < newInfo.openParens+1; ++k)
+        info.stream << ")";
```
maerhart, line 353: Nit: This and the loop in the first branch could be factored out and be before the if statement chain with `newInfo.openParens` as the upper bound. Then we only have to print one closing parenthesis here. Also, the first two branches would be the same then and could be merged.

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -334,9 +337,21 @@ struct ExpressionVisitor
     newInfo.stream.indent(newInfo.indentLevel);
     if (failed(printExpression(worklist, newInfo)))
       return failure();
+
     info.stream << info.valueMap.lookup(yieldedValue);
-    for (unsigned k = 0; k < newInfo.openParens + 1; ++k)
-      info.stream << ")";
+
+    if(weight != 0 && newInfo.openParens > 0){
+      for(int j=0;j<newInfo.openParens;j++)
```
maerhart, line 344: Super-nit: the LLVM codebase typically prefers `++j` instead of `j++`

maerhart (approved):
Very nice! Just some nitpicky comments about the if-statement which can probably be written slightly differently to be easier to read.

Let me know if I should merge it for you once you are happy with it to be merged.

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
luisacicolini: oh cool! I did not know :)

luisacicolini:
@maerhart thanks a lot for your reviews, I cleared the code and everything should be good now, ready for merging from my side :)

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -334,9 +337,17 @@ struct ExpressionVisitor
...
-    for (unsigned k = 0; k < newInfo.openParens + 1; ++k)
+
+    for(int j=0;j<newInfo.openParens;++j)
+      info.stream <<")";
+
+    if(weight != 0){
+      info.stream <<" :weight "<< weight << "))";
+    } else{
       info.stream << ")";
+    }
```
maerhart, line 350: Nit: If statements where the body of all branches is just one line should omit the curly braces according to the LLVM coding style guide

maerhart:
Thanks! Could you run clang-format over this file to make the CI pass?

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
maerhart: We could also remove one of the two closing parenthesis at the end of the statement in the then-branch and move the statement in the else-branch out such that the if-statement only has a then-statement.

luisacicolini:
> Thanks! Could you run clang-format over this file to make the CI pass?

Sorry for the delay, took me a while to build it!

maerhart:
Thanks! It looks like you accidentally also ran clang-format over the `.mlir` files, at least they are formatted strangely now. Maybe you could revert the formatting for the `.mlir` files?

maerhart:
This looks great now! Thanks a lot and congrats to your first CIRCT contribution! :tada:

PR #6921, opened 2024-04-16
[FIRRTL] Make input probes illegal

dtzSiFive:
Disallow input probes.
cc https://github.com/chipsalliance/firrtl-spec/pull/183 .

Adds verifier for this, and mostly updates tests.

Delete ProbeDCE.

Many passes can (and should) be simplified if they do not need to consider this.

Incomplete list:

* Inliner
* LowerXMR (!)
* Dedup
* HoistPassthrough


dtzSiFive:
Not sure about this direction (thoroughly ripping this out from all the things), but leaving here for now.

test/Dialect/FIRRTL/infer-widths-errors.mlir
```diff
@@ -159,24 +159,26 @@ firrtl.circuit "Foo" {
...
+//     firrtl.connect %w2, %y : !firrtl.uint, !firrtl.uint<2>
+// 
+//     %inst1_ref = firrtl.instance inst1 @InRef(in ref: !firrtl.rwprobe<uint>)
+//     %inst2_ref = firrtl.instance inst2 @InRef(in ref: !firrtl.rwprobe<uint>)
+//     firrtl.ref.define %inst1_ref, %w1_ref : !firrtl.rwprobe<uint>
+//     // expected-note-DISABLED @below {{width is constrained to be at most 1 here:}}
+//     // expected-note-DISABLED @below {{width is constrained to be at least 2 here:}}
+//     firrtl.ref.define %inst2_ref, %w2_ref : !firrtl.rwprobe<uint>
+//   }
+// }
```
seldridge, line 181: Deleting is preferable to commenting out.

seldridge (commented):
I'm generally good with the approach.

test/Dialect/FIRRTL/infer-widths-errors.mlir
dtzSiFive: Yes, just left it here as a sore thumb so I wouldn't forget to either rework or delete it.   But 100% don't plan to commit it this way :+1:.

dtzSiFive: Will file an issue about it.

dtzSiFive: Dropping.  Was going to file an issue -- I like having the "don't infer probe chain to 'max' width ever" under test but not sure how to make that reachable now, depending on instance choice rules this may not be possible (?).

test/Dialect/FIRRTL/errors.mlir
```diff
@@ -2276,21 +2281,21 @@ firrtl.circuit "InstanceOfClass" {
 // -----
 
 firrtl.circuit "InputProbePublic" {
-  // expected-error @below {{input probe not allowed on public module}}
+  // expected-error @below {{input probe not allowed}}
```
seldridge, line 2284: This could add a test of input probe on private module being disallowed, too.

seldridge (approved):
🎯

LGTM

test/Dialect/FIRRTL/infer-widths-errors.mlir
```diff
@@ -157,27 +157,6 @@ firrtl.circuit "Foo" {
...
-    firrtl.connect %w2, %y : !firrtl.uint, !firrtl.uint<2>
-
-    %inst1_ref = firrtl.instance inst1 @InRef(in ref: !firrtl.rwprobe<uint>)
-    %inst2_ref = firrtl.instance inst2 @InRef(in ref: !firrtl.rwprobe<uint>)
-    firrtl.ref.define %inst1_ref, %w1_ref : !firrtl.rwprobe<uint>
-    // expected-note @below {{width is constrained to be at most 1 here:}}
-    // expected-note @below {{width is constrained to be at least 2 here:}}
-    firrtl.ref.define %inst2_ref, %w2_ref : !firrtl.rwprobe<uint>
-  }
-}
```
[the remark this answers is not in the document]
dtzSiFive: We need multiple drivers for the same static (in the IR, like a specific wire or port).  Since probe's must be driven exactly once this is a bit tricky to test.

An input probe driven in different instantiation contexts with different widths is clearest way I know to do this (this is test I had commented out earlier for same "seems like a good thing to have under test still" thinking).
( https://github.com/llvm/circt/pull/6921#discussion_r1594774391 )

I had made a comment previously about instance-choice (if has an output probe of unknown width, in theory the two choices might try to infer to different widths based on their contents), but I'm not sure story there re:unknown widths and interaction with inferences.

So, not sure how to keep this as a test but that would be nice, agreed, thanks!

test/Dialect/FIRRTL/errors.mlir
dtzSiFive: Done!

PR #6923, opened 2024-04-16
[FIRRTL] Gather intrinsic lowering via dialect interface.

dtzSiFive:
Make it easier to extend the set of intrinsic lowering patterns by implementing the dialect interface for your dialect.

Current lowering patterns are implemented as part of the FIRRTL implementation of this dialect interface.

This doesn't facilitate adding intrinsic implementations (lowerings) unrelated to a different dialect but leave that sort of extensibility as future work.


prithayan (approved):
This is great, thanks for adding the interfaces support.

PR #6928, opened 2024-04-17
[Arc] Fix segfault in SplitLoops

Moxinilian:
There was an interesting segfault where unrelated calls to functions would confuse the SplitLoops pass into thinking it is a call to an Arc.


fabianschuiki (approved):
LGTM! Thanks for fixing this :smiley:

maerhart (approved):
Thanks!

PR #6929, opened 2024-04-18
[ImportVerilog] Add case statement.

angelzzzzz:
Use `scf.if` and `moore.or` to implement the selection logic for case statements. For example, if a default case exists, it will perform an OR operation on all the results above, to determine whether the default statement needs to be executed. If the result of the operation is zero, the default statement will be executed.
However, this still leaves some issues, according to the "IEEE Standard 1800-2017 for SystemVerilog", which mentions that "During the linear search, if one of the Case_item_expressions matches the case_expression, then the statement associated with that case_item should be executed, and the linear search shall terminate." I want to know how we can implement this short-circuit logic?

### DIFF
--- a/lib/Conversion/ImportVerilog/Statements.cpp
+++ b/lib/Conversion/ImportVerilog/Statements.cpp
@@ -110,6 +110,59 @@ struct StmtVisitor {
     return success();
   }
 
+  // Handle case statements.
+  LogicalResult visit(const slang::ast::CaseStatement &caseStmt) {
+    auto caseExpr = context.convertExpression(caseStmt.expr);
+    auto items = caseStmt.items;
+    // Used to generate the condition of the default case statement.
+    SmallVector<Value> defaultConds;
+    // Traverse the case items.
+    for (auto item : items) {
+      // One statement will be matched with multi-conditions.
+      // Like case(cond) 0, 1 : y = x; endcase.
+      SmallVector<Value> allConds;
+      for (const auto *expr : item.expressions) {
+        auto itemExpr = context.convertExpression(*expr);
+        auto newEqOp = builder.create<moore::EqOp>(loc, caseExpr, itemExpr);
+        allConds.push_back(newEqOp);
+      }
+      // Bound all conditions of an item into one.
+      auto cond = allConds.back();
+      allConds.pop_back();
+      while (!allConds.empty()) {
+        cond = builder.create<moore::OrOp>(loc, allConds.back(), cond);
+        allConds.pop_back();
+      }
+      // Gather all items' conditions.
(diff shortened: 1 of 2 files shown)


uenoku:
Lowering case to if-statement sounds great. I'm curious we could use CF switch op `https://github.com/llvm/llvm-project/blob/fd98f80f602e11d523d252feef301634c2c689a7/mlir/test/Dialect/ControlFlow/ops.mlir#L13` for this though we'll need to lower switch to mux/if anyway since there is no corresponding construct in core dialects.

angelzzzzz:
> Lowering case to if-statement sounds great. I'm curious we could use CF switch op `https://github.com/llvm/llvm-project/blob/fd98f80f602e11d523d252feef301634c2c689a7/mlir/test/Dialect/ControlFlow/ops.mlir#L13` for this though we'll need to lower switch to mux/if anyway since there is no corresponding construct in core dialects.

That's a good idea! But I have some doubts that there are a few blocks in the test of `cf.switch` while `moore.module` has `SingleBlock` trait, which means it only allows one block in the module. In this case, it seems that it is not appropriate for `case`, and how can I solve this problem?

hailongSun2000:
LGTM! For the `moore` IR, it doesn't showcase short-circuit logic exactly. Maybe I think the nested `if-else` can solve this problem, or add an extra identifier to estimate whether the prior `if` has executed? If that(prior is true), the second `if` doesn't execute(extra workload). I'm not sure.

uenoku:
> But I have some doubts that there are a few blocks in the test of cf.switch while moore.module has SingleBlock trait

Good point! Thank you for pointing it out. Seems like cf.switch is more like a branch op. Maybe? we can use scf.index_switch(https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfindex_switch-scfindexswitchop) but converting conditions into `index` seems like unnecessary complicated. So I think the current approach looks great to me :)

fabianschuiki (approved):
This looks good to me, and thanks for adding the nice tests! :partying_face:

Short circuiting is usually done by pushing the OR-ing or AND-ing of conditions into control flow. In this example, you'd turn `case a, b, c: ...` into:
```
matched = if a {
  true
} else {
  if b {
    true
  } else {
    if c {
      true
    } else {
      false
    }
  }
}
```
And then you'd rely on later compiler optimizations to turn this into `matched = a | b | c` in case `a`, `b`, and `c` are side-effect free.

I think the PR you have right now is fine to land as-is, because we have other places in the lowering to Moore where we're not supporting short-circuiting right now. We can do a dedicated short-circuiting rework in a separate later PR.

PR #6932, opened 2024-04-18
Bump llvm

azidar:
Housekeeping bump of LLVM.


dtzSiFive:
Link error looks like upstream issue, sent PR to fix it: https://github.com/llvm/llvm-project/pull/89318 .

It does not appear to be caught by/problematic with static builds (the default), the failure above is in a shared library build which may be why it failed but not when tested locally.

Hopefully will get this fixed soon and can use a commit with the fix, alternatively can walk the update back to before the breaking commit.

dtzSiFive:
I plan to merge this later today.

PR #6933, opened 2024-04-19
[HW] Rework InnerSym infra to support nested symbol tables

mortbopet:
This PR aims to rework the existing `InnerSym` infrastructure to properly support nested symbol tables. This is a feature that we'll be needing over on the `Ibis` dialect side, and which probably will be useful in the future if/when other dialects also need >1 levels of symbol table nesting, with `InnerSym`-like lookup semantics.

The change is backwards-compatible with the existing InnerSym infrastructure; hence, no changes in any `.mlir` FIRRTL or HW tests, and only minimal changes in some FIRRTL and HW-related files.

Changes:
1. InnerSymRef can now specify a path of symbol references (`<@A::@B::@C`)
2. The `InnerSymbolTable` trait is now an interface that inherits from `InnerSymbol`; i.e. `InnerSymbolTable`s must now themselves define a symbol (to be able to be referenced in a path).
3. `InnerSymbolTable`s must now be nested directly within either an `InnerRefNamespaceLike` or `InnerSymbolTableOpInterface` parent operation.
4. `InnerSymbolTable` operations are now required to implement **either** `SymbolOpInterface` or `InnerSymbolOpInterface`; previously, only `SymbolOpInterface` was allowed/required (disallowing nesting).
5. Somewhat in extension of the above, `InnerRefNamespace` now no longer requires the operation that it's attached to to be a `SymbolTable` - since relying on this would imply that `InnerSymbolTable`-defining operations always are `mlir::SymbolOp`... which they are not, due to the above point. AFAICT, the `SymbolTable` constraint was mainly used in `InnerRefNamespace` to scan top-level operations. Now, we "just" do a manual scan for `InnerSymbolTable`-defining operations within the `InnerRefNamespace`. To be seen whether the performance impact of this is significant.
6. Partly motivated by 5., `InnerRefNamespace` is now a stand-alone analysis. This implies that it no longer takes `SymbolTable&` and `InnerSymbolTableCollection&` references (a design choice which i don't fully understand, since it allows users to create completely unrelated symbol table to InnerSymbolTable collection relations). The refactor for doing this was fairly small, and I didn't find any usecases that indicated that the prior separation was absolutely necessary.
7. ... which implies that `InnerSymbolTableCollection` is now private to the `InnerRefNamespace` implementation, and all prior uses of `InnerSymbolTableCollection` now uses an `InnerRefNamespace` - analysis (execution) times should (hopefully) be identical, it's just a matter of how the symbol table collection is wrapped.
8. `static InnerSymbolTable::walkSymbols` now takes an additinal (optional) callback that is called for each `InnerSymbolTable` encountered; currently, this is used to build an `InnerSymbolTable`.
   - Doesn't use the built-in `Operation::walk` function; instead, implements it's own walk s.t. it can track the currently active `NestedSymbolTable`.

Because I like graphs, i decided to maintain the nested structure as a graph of `InnerSymbolTable`s. This implies that each `InnerSymbolTable` maintains the symbols defined directly within their scope, as well as any nested symbol tables. Lookups will therefore be performed recursively into the nested symbol tables. An alternative designpoint would have been to only maintain a single, top-level `InnerSymbolTable` which maintains its symbol table as a `DenseMap<hw::InnerRefAttr, InnerSymTarget>`; a design which implies that all symbols are recorded in a single place with their fully qualified name (the `InnerRefAttr`). Given the single namespace, it would most likely also imply a slightly faster lookup time for (deeply) nested symbols. However, since we don't currently have any significant performance requirements for deeply nested symbol tables (all existing firrtl/hw usecases uses a single level of nesting, so lookup times should be identical to prior to this PR), I'd personally err on the side of readability and extendability, and go with the graph-like structure - fight me in the comment section :).

Finally, the graph-like structure is also implemented using double linking; that is, child `InnerSymbolTable`s have a reference to their parent symbol table. This is somewhat leftover of my initial design which supported relative references. I personally think it's a nice thing to have, but since it isn't strictly necessary it could be removed if someone deems it unacceptable.

### DIFF
--- a/docs/RationaleSymbols.md
+++ b/docs/RationaleSymbols.md
@@ -71,7 +71,9 @@ in languages are usually a human convenience, and not necessarily a good IR
 construct. In general, front-ends will generate fully qualified names early in
 the compilation process suitable for use in the IR. And if references truly are
 relative, then using a dataflow (SSA) representation is more appropriate.
-Specifically, references which are relative to an instance hierarchy are dynamically scoped, which is a different implementation than symbols, which are lexically scoped.
+Specifically, references which are relative to an instance hierarchy are
+dynamically scoped, which is a different implementation than symbols, which are
+lexically scoped.
 
 ## Inner Symbol Reference Attribute
 
@@ -128,7 +130,6 @@ utility for `InnerRefNamespace`.
 This is also both a class and a trait.
 
 #### Class
-
 
 Combines `InnerSymbolTableCollection` with a `SymbolTable` for resolution of
 `InnerRefAttr`s, primarily used during verification as argument to the
--- a/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
@@ -64,7 +64,8 @@ class FIRRTLModuleLike<string mnemonic, list<Trait> traits = []> :
     DeclareOpInterfaceMethods<PortList>,
     DeclareOpInterfaceMethods<InstanceGraphModuleOpInterface>,
     DeclareOpInterfaceMethods<SymbolUserOpInterface>,
-    OpAsmOpInterface, InnerSymbolTable
+    OpAsmOpInterface,
+    InnerSymbolTable
   ]> {
 
   /// Additional class definitions inside the module op.


blakep-msft (approved):
LGTM

lib/Dialect/HW/HWAttributes.cpp
```diff
@@ -210,14 +210,30 @@ EnumFieldAttr EnumFieldAttr::get(Location loc, StringAttr value,
 // InnerRefAttr
 //===----------------------------------------------------------------------===//
 
-Attribute InnerRefAttr::parse(AsmParser &p, Type type) {
+Attribute circt::hw::parseInnerRefAttr(AsmParser &p, bool parseLess) {
```
darthscsi, line 213: The attribute should have a unique parse.  Is there a reason to have `<` and `>` at all?

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -64,7 +64,9 @@ class FIRRTLModuleLike<string mnemonic, list<Trait> traits = []> :
     DeclareOpInterfaceMethods<PortList>,
     DeclareOpInterfaceMethods<InstanceGraphModuleOpInterface>,
     DeclareOpInterfaceMethods<SymbolUserOpInterface>,
-    OpAsmOpInterface, InnerSymbolTable]> {
+    OpAsmOpInterface,
```
darthscsi, line 67: Please commit whitespace changes separately (Directly as NFC is fine).

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,71 @@ include "circt/Dialect/HW/HWDialect.td"
...
-  let summary = "Refer to a name inside a module";
+  let summary = "Refer to a name inside a list of scopes";
   let description = [{
-    This works like a symbol reference, but to a name inside a module.
+    This works like a symbol reference, but to a name inside a list of `InnerSymbolTable`'s.
   }];
   let mnemonic = "innerNameRef";
-  let parameters = (ins "::mlir::FlatSymbolRefAttr":$moduleRef,
-                        "::mlir::StringAttr":$name);
+  let parameters = (ins ArrayRefParameter<"::mlir::StringAttr">:$path);
```
darthscsi, line 25: Are paths rooted at the (lexical) top?  If so, is the first level required to be a real symbol?  If so, why drop the FlatSymbolRefAttr?
Wouldn't this be something like `optional(?)<FlatSymbolRefAttr>, ArrayRefParameter<FlatInnerSymbolRefAttr>`?

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,71 @@ include "circt/Dialect/HW/HWDialect.td"
...
   let hasCustomAssemblyFormat = 1;
+  let genVerifyDecl = 1;
 
   let extraClassDeclaration = [{
-    /// Get the InnerRefAttr for an operation and add the sym on it.
-    static InnerRefAttr getFromOperation(mlir::Operation *op,
-                                         mlir::StringAttr symName,
-                                         mlir::StringAttr moduleName);
+    ::mlir::StringAttr getRootReference() const {
+      assert(!getPath().empty() && "Empty path");
```
darthscsi, line 67: Should the verifier have ensured this?

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,71 @@ include "circt/Dialect/HW/HWDialect.td"
...
+      return getPath().front();
+    }
+    ::mlir::StringAttr getLeafReference() const {
+      assert(!getPath().empty() && "Empty path");
```
darthscsi, line 71: again, shouldn't the verifier have checked?

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,71 @@ include "circt/Dialect/HW/HWDialect.td"
...
+      return getPath().back();
+    }
+    ::mlir::StringAttr getName() const {
+      return getLeafReference();
+    }
 
-    /// Return the name of the referenced module.
-    mlir::StringAttr getModule() const { return getModuleRef().getAttr(); }
+    // Old InnerRefAttr compatability accessors
```
darthscsi, line 78: Why have these?

include/circt/Dialect/HW/HWOpInterfaces.h
```diff
@@ -77,25 +84,38 @@ class InnerRefNamespace : public TraitBase<ConcreteType, InnerRefNamespace> {
...
+  // symbol, this is essential to how InnerRef's work.
+  static_assert(
+      ConcreteType::template hasTrait<
+          circt::hw::InnerSymbolOpInterface::Trait>() ||
+          ConcreteType::template hasTrait<mlir::SymbolOpInterface::Trait>(),
+      "expected operation to define either an InnerSymbol or Symbol");
+
+  // InnerSymbolTable's must be directly nested under either another
+  // InnerSymbolTable operation or an InnerRefNamespace.
+  auto *parent = op->getParentOp();
```
darthscsi, line 108: I don't believe region verifiers are allowed to access their parents.  I believe this is why there is both innersymboltable and innrefnamespace (@dtzSiFive ?).

include/circt/Dialect/HW/HWStructure.td
```diff
@@ -106,7 +106,6 @@ class HWModuleOpBase<string mnemonic, list<Trait> traits = []> :
     ::circt::hw::PortInfo $cppClass::getPort(size_t idx) {
       return ::getPort(*this, idx);
     }
-
```
darthscsi, line 109: NIT: whitespace changes as separate commit.

include/circt/Dialect/HW/InnerSymbolTable.h
```diff
@@ -33,12 +37,16 @@ class InnerSymTarget {
   explicit InnerSymTarget(Operation *op) : InnerSymTarget(op, 0) {}
 
   /// Target an operation and a field (=0 means the op itself).
-  InnerSymTarget(Operation *op, size_t fieldID)
-      : op(op), portIdx(invalidPort), fieldID(fieldID) {}
+  InnerSymTarget(Operation *op, size_t fieldID,
+                 Operation *symbolTableOp = nullptr)
```
darthscsi, line 41: This seems unsafe.  The op with a symboltable on it may not be legal to access by users of an InnerSymTarget in the context in which they are running.

include/circt/Dialect/HW/InnerSymbolTable.h
```diff
@@ -136,102 +161,152 @@ class InnerSymbolTable {
...
+      Operation *op, FuncTy &&symCallback,
+      std::optional<InnerSymTableCallbackFn> symTblCallback = std::nullopt) {
     if constexpr (std::is_void_v<RetTy>)
       return (void)walkSymbols(
-          op, InnerSymCallbackFn(
-                  [&](StringAttr name, const InnerSymTarget &target) {
-                    std::invoke(std::forward<FuncTy>(callback), name, target);
-                    return success();
-                  }));
+          op,
```
darthscsi, line 214: It deviates from mlir's walk, but what about using `FailureOr`?

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -749,9 +749,18 @@ struct Equivalence {
 
   // NOLINTNEXTLINE(misc-no-recursion)
   void check(InFlightDiagnostic &diag, Operation *a, Operation *b) {
-    hw::InnerSymbolTable aTable(a);
-    hw::InnerSymbolTable bTable(b);
-    ModuleData data(aTable, bTable);
+    auto aTable = hw::InnerSymbolTable::get(a);
+    if (failed(aTable)) {
+      diag.attachNote(a->getLoc()) << "failed to create symbol table";
```
darthscsi, line 754: Seems like these should be standard asserts.  @youngar ?

darthscsi (commented):
I didn't get through the InnerSymbolTable and InnerRefNamespace implementations yet.  @dtzSiFive should definately take a look.

There must be an update of the symbol rational document.

include/circt/Dialect/HW/HWAttributesNaming.td
mortbopet: Yes, they're always rooted at the top. By "real symbol" do you mean an `mlir::SymbolOp`? If so, then no (as described by point 5). We could still have a `FlatSymbolRefAttr` and interpret that as "just" a name (which may refer to a top-level operation, that may itself implement either an `mlir::SymbolOp` or `hw::InnerSymbolOp`.

include/circt/Dialect/HW/HWAttributesNaming.td
mortbopet: Mostly to reduce the footprint of this PR. Can remove them as part of this PR if reviewers are willing to accept the increase in PR scope.

include/circt/Dialect/HW/InnerSymbolTable.h
teqdruid: By context, do you mean in a parallelization context?

I think this is the doublely-linked stuff @mortbopet was referring to in the PR comments, yes? If so, I think it makes sense to just ditch it.

include/circt/Dialect/HW/InnerSymbolTable.h
```diff
@@ -101,11 +115,17 @@ class InnerSymTarget {
 };
 
 /// A table of inner symbols and their resolutions.
-class InnerSymbolTable {
+/// enable_shared_from_this is used to enable cross-linking between symbol
+/// tables during construction.
+/// Given this, InnerSymbolTable is only legal to construct as a shared_ptr.
```
teqdruid, line 120: This complexity is a result of the doubly-linked stuff? If so, I don't think it's worth the added complexity.

teqdruid: Plus, shared_ptr isn't effective in the presence of pointer cycles.

include/circt/Dialect/Ibis/IbisOps.td
```diff
@@ -632,11 +627,8 @@ def ContainerOp : IbisOp<"container", [
     llvm::StringRef getModuleName() {
       return getModuleNameAttr();
     }
-  }];
-
-  let extraClassDefinition = [{
     // For hw:InnerSymbol - the symbol targets the operation and not any result.
-    std::optional<size_t> $cppClass::getTargetResultIndex() {
+    static std::optional<size_t> getTargetResultIndex() {
```
teqdruid, line 631: I don't understand how this works without the `$cppClass`.

lib/Dialect/ESI/AppID.cpp
```diff
@@ -201,6 +201,7 @@ FailureOr<ArrayAttr> AppIDIndex::getAppIDPathAttr(hw::HWModuleLike fromMod,
     FailureOr<hw::InnerSymbolOpInterface> op = modIDs->lookup(appid, loc);
     if (failed(op))
       return failure();
+    op->dump();
```
teqdruid, line 204: Debug?

lib/Dialect/HW/InnerSymbolTable.cpp
```diff
@@ -13,57 +13,129 @@
...
 
+#define DEBUG_TYPE "innersymboltable"
+
 using namespace circt;
 using namespace hw;
 
 namespace circt {
 namespace hw {
 
+StringAttr InnerSymbolTable::getName(Operation *innerSymbolOp) {
```
teqdruid, line 27: nit: this should be below the comment below.

teqdruid (commented):
Did not review any of the cpp definitions. Also, I'm not familiar enough with this subsystem to be a qualified reviewer. I really don't understand the complexity of the existing InnerSymbol stuff...

What I've seen in the headers and tblgen files is more expansive than I expected, but does seem to be reasonable.

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,71 @@ include "circt/Dialect/HW/HWDialect.td"
...
-                                         mlir::StringAttr moduleName);
+    ::mlir::StringAttr getRootReference() const {
+      assert(!getPath().empty() && "Empty path");
+      return getPath().front();
+    }
+    ::mlir::StringAttr getLeafReference() const {
+      assert(!getPath().empty() && "Empty path");
+      return getPath().back();
+    }
+    ::mlir::StringAttr getName() const {
```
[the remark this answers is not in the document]
mortbopet: Agreed; I'll in general refactor usage of "leaf" to "target" - seems more apt.

include/circt/Dialect/Ibis/IbisOps.td
mortbopet: This is inside `extraClassDeclaration` so it'll get emitted under the class definition, in the generated `.h` file.

include/circt/Dialect/HW/InnerSymbolTable.h
mortbopet: Parent linkage has now been removed.

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,71 @@ include "circt/Dialect/HW/HWDialect.td"
...
+  
+    AttrBuilderWithInferredContext<(ins "::mlir::ArrayRef<::mlir::StringAttr>":$pathAttr,
+                                        "::mlir::StringAttr":$leaf), [{
+      llvm::SmallVector<::mlir::StringAttr> fullPath(pathAttr.begin(), pathAttr.end());
+      fullPath.push_back(leaf);
+      return InnerRefAttr::get(leaf.getContext(), fullPath);
+    }]>,
+
+    // @mortbopet: Old InnerRef single-module compatability builders.
+    AttrBuilderWithInferredContext<(ins "::mlir::StringAttr":$leaf), [{
```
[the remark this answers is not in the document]
mortbopet: Changed to `target`

include/circt/Dialect/HW/InnerSymbolTable.h
mortbopet: Ditched!

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,71 @@ include "circt/Dialect/HW/HWDialect.td"
...
+      return getLeafReference();
+    }
 
-    /// Return the name of the referenced module.
-    mlir::StringAttr getModule() const { return getModuleRef().getAttr(); }
+    // Old InnerRefAttr compatability accessors
+    ::mlir::StringAttr getModule() const {
+      return getRootReference();
+    }
+    ::mlir::FlatSymbolRefAttr getModuleRef() const {
```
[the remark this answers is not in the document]
mortbopet: As safe as it was in the existing implementation :).

lib/Dialect/HW/HWAttributes.cpp
mortbopet: > Is there a reason to have < and > at all?

Just following the semantics of the existing `InnerRefAttr` parsing logic. This has been extracted out purely to avoid code duplication in `ibis.scoperef` **type** which wants to parse a `hw.innerref` as `!ibis.scoperef<@a::@b>`.

mortbopet:
Alrightey, dove in and
* changed `hw::InnerRef` leaf/getName() to `getTarget`
* changed `hw::InnerRef::getModule` to `getRoot`
* Removed doubly-linking of innersymboltables
* Some prose updated in the rationale doc

@teqdruid @darthscsi @dtzWill feel free to re-review.

include/circt/Dialect/HW/HWAttributesNaming.td
darthscsi: FlatSymbolRefAttr should always refer to a symbol.  We should coopt an existing typed attribute to refer to a new kind of thing.

include/circt/Dialect/HW/HWAttributesNaming.td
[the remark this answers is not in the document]
darthscsi: The question was really whether the root was a FlatSymbolRefAttr which refered to a symbol.  Which is to say is this always rooted at a symbol table (which I think it should be).

include/circt/Dialect/HW/HWAttributesNaming.td
darthscsi: Things which are correct by construction do have a cost in having extra assertions.  Both in debug time and human readability.  The constructor for the attribute should be where the assertion is.

docs/RationaleSymbols.md
```diff
@@ -35,32 +35,49 @@ symbol-like data to allow forming non-SSA linkage between disparate elements.
...
+symbols attached to the operation and its fields. Operations containing an inner
+symbol must implement the `InnerSymbol` interface.
 
 Inner Symbols are different from normal symbols due to MLIR symbol table 
 resolution rules.  Specifically, normal symbols are resolved by first going up 
 to the closest parent symbol table and resolving down from there (recursing 
-back down for nested symbol paths).  In FIRRTL and HW, modules define a symbol in a 
-`firrtl.circuit` or `builtin.module` symbol table.  For instances to be able to resolve the 
+back down for nested symbol paths). Conversely, Inner symbol symbol resolution
+is rooted at some encapsulating `InnerRefNamespace`, which defines the top-level
```
darthscsi, line 48: InnerRefNamespace is an artifact of implementation to allow inner symbol verifiers to be written.  An op it is on is required to also be a SymbolTable.  From a description standpoint, saying resolution starts from the parent SymbolTable seems more clear.

docs/RationaleSymbols.md
```diff
@@ -35,32 +35,49 @@ symbol-like data to allow forming non-SSA linkage between disparate elements.
...
+symbol table hierarchy. In FIRRTL and HW, modules define a symbol in a
+`firrtl.circuit` or `builtin.module` symbol table, which is then interpreted as an
+`InnerRefNamespace`.  For instances to be able to resolve the 
 modules they instantiate, the symbol use in an instance must resolve in the 
-top-level symbol table.  If a module were a symbol table, instances resolving a 
+top-level `InnerRefNamespace`.  If a module were a symbol table, instances resolving a
```
darthscsi, line 53: They are, right now, real symbols and resolve using symbol resolution in symbol tables.

docs/RationaleSymbols.md
```diff
@@ -35,32 +35,49 @@ symbol-like data to allow forming non-SSA linkage between disparate elements.
...
 symbol would start from their own module, never seeing other modules (since 
 resolution would start in the parent module of the instance and be unable to go 
-to the global scope).  The second problem arises from nesting.  Symbol 
+to the global scope).  The second problem arises from nesting. MLIR Symbol 
 defining operations must be immediate children of a symbol table.  FIRRTL and HW/SV 
 operations which define an `inner_sym` are grandchildren, at least, of a symbol 
 table and may be much further nested.  Lastly, ports need to define `inner_sym`, 
 something not allowed by normal symbols.
 
+Given that all `InnerSym` references are rooted at some top-level `InnerRefNamespace`,
```
darthscsi, line 62: I would stick with rooting at a symboltable.

docs/RationaleSymbols.md
```diff
@@ -35,32 +35,49 @@ symbol-like data to allow forming non-SSA linkage between disparate elements.
...
+inner symbols are not able to be resolved as relative references (i.e. if a
+reference occurs somewhere within a set of nested `InnerSymbolTable` operations,
+and a relative reference is used to reference e.g. a cousin symbol).  
+This design has been chosen to make inner symbol resolution both more robust
+as well as simpler to implement, based on an observation that relative references
+in languages are usually a human convenience, and not necessarily a good IR
```
darthscsi, line 68: The argument is really that symbols and innersymbols are lexically scoped.

docs/RationaleSymbols.md
```diff
@@ -35,32 +35,49 @@ symbol-like data to allow forming non-SSA linkage between disparate elements.
...
+construct. In general, front-ends will generate fully qualified names early in
+the compilation process suitable for use in the IR. And if references truly are
```
darthscsi, line 70: Specifically, references which are relative to the instance hierarchy are dynamically scoped, which is a different implementation than symbols, which are lexically scoped.

docs/RationaleSymbols.md
```diff
@@ -35,32 +35,49 @@ symbol-like data to allow forming non-SSA linkage between disparate elements.
...
+relative, then using a dataflow (SSA) representation is more appropriate.
+
 ## Inner Symbol Reference Attribute
 
 An attribute `InnerRefAttr` is provided to encapsulate references to inner 
-symbols.  This attribute stores the parent symbol and the inner symbol.  This 
-provides a uniform type for storing and manipulating references to inner 
-symbols.  An `InnerRefAttr` resolves in an `InnerRefNamespace`.
+symbols.  This attribute stores the full path to a symbol, through the
```
darthscsi, line 76: to an inner symbol?  or to a symbol or inner symbol?

darthscsi, line 76: What does full path mean?

include/circt/Dialect/HW/HWAttributesNaming.td
darthscsi: The root of my original comment is the loss of type information in the path, even if just on the first element.

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,65 @@ include "circt/Dialect/HW/HWDialect.td"
...
   let builders = [
-    AttrBuilderWithInferredContext<(ins "::mlir::StringAttr":$module,
-                                        "::mlir::StringAttr":$name), [{
-      return $_get(
-          module.getContext(), mlir::FlatSymbolRefAttr::get(module), name);
-    }]>
+
+    AttrBuilderWithInferredContext<(ins "::mlir::ArrayRef<::mlir::StringAttr>":$pathAttr), [{
+      assert(!pathAttr.empty() && "Empty path");
+      assert(pathAttr.front() && "Null path element");
```
darthscsi, line 31: nit: Probably should check something like:
`llvm::all_of(pathAttr, [](auto& e) { assert(e); });`
any null element is malformed.

include/circt/Dialect/HW/HWAttributesNaming.td
```diff
@@ -17,32 +17,65 @@ include "circt/Dialect/HW/HWDialect.td"
...
-    /// Get the InnerRefAttr for an operation and add the sym on it.
-    static InnerRefAttr getFromOperation(mlir::Operation *op,
-                                         mlir::StringAttr symName,
-                                         mlir::StringAttr moduleName);
+    ::mlir::StringAttr getRoot() const {
+      assert(!getPath().empty() && "Empty path");
+      return getPath().front();
+    }
+
+    ::mlir::FlatSymbolRefAttr getRootRef() const {
```
darthscsi, line 71: Although nothing complains, per-se, FlatSymbolRefAttr should only be used to refer to a real symbol, we shouldn't just shove any old string in one.  A string array as a path with this explicit semi-cast is fine, if the verifier checks that the first element of the path resolves to a symbol.

include/circt/Dialect/HW/HWOpInterfaces.h
darthscsi: I see that you are just copying this over.  I agree it is not likely to cause a problem, but my note that it might not be allowed stands.

darthscsi (approved):
Generally LGTM.  I'd like the rational to be more clear on the conceptual nesting v.s. the implementation helpers.  By this I mean references being <Symbol, InnerSymbol...>, thus the traits being <SymbolTable, InnerSymbolTable...>  The namespace I believe are just an artifact to achieve efficient verification.

I'll implore @dtzSiFive or @youngar to go over the namespace and collection code.

docs/RationaleSymbols.md
mortbopet: > op it is on is required to also be a SymbolTable

I'm hoping i made it clear in the rationale changes that `InnerRefNamespace` **no longer is required to be a SymbolTable**... It **can be**, it's just not going to be used by any of the lookup code (InnerRefNamespace now implements it's own scan of its inner region for `InnerSymbolTable`-defining ops).

mortbopet:
@darthscsi I'm hoping i made it clear in the rationale changes that InnerRefNamespace **no longer is required to be a SymbolTable**... It **can** be, it's just not going to be used by any of the lookup code (InnerRefNamespace now implements it's own scan of its inner region for InnerSymbolTable-defining ops).

As i see it, InnerRefNamespace cannot anymore rely on the "top-level" `InnerSymbolTable`s being discoverable through a `SymbolTable` since this would imply that the top-level operations **always** implement an MLIR symbol. This assumption works for HW and FIRRTL, but not for Ibis, where, for instance, we are using `ibis.container` as the nest-able thing, that is, our `InnerSymbolTable` operation. If `ibis.container` were to implement an MLIR symbol, this would imply that an `ibis.container` itself would have to be a `SymbolTable` which would defeat the whole purpose of this.

Instead, the provided change to `InnerRefNamespace`/`InnerSymbolTable` requires that an `InnerSymbolTable`-defining operation is **either** an MLIR symbol or an InnerSymbol.  

InnerRefNamespace will therefore scan the top-level "manually", i.e. not using a `SymbolTable` analysis.

include/circt/Dialect/HW/HWAttributesNaming.td
mortbopet: Will remove and change at current `getRootRef` usage sites - given that the first element of the path isn't guaranteed to resolve to a symbol.

docs/RationaleSymbols.md
```diff
@@ -35,32 +35,49 @@ symbol-like data to allow forming non-SSA linkage between disparate elements.
 To accomplish this, an attribute named `inner_sym` is attached, providing a 
 scoped symbol-like name to the element.  An operation with an `inner_sym`
 resides in arbitrarily-nested regions of a region that defines an
-`InnerSymbolTable` and a `Symbol` .  `InnerSymbolTable` operations must reside
-within an `InnerRefNamespace`.  The `inner_sym` attribute must be an `InnerSymAttr`
-which defines the inner symbols attached to the operation and its fields.
-Operations containing an inner symbol must implement the `InnerSymbol` interface.
+`InnerSymbolTable`.  `InnerSymbolTable` operations can themselves be arbitrarily
```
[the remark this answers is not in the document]
mortbopet: This is also listed later on in the `InnerSymbolTable##Trait`. Added your suggestion to increase clarity.

docs/RationaleSymbols.md
mortbopet: Elaborated as:
> This attribute stores the path to a symbol, through the
`InnerRefNamespace` symbol table hierarchy, this path the combination of
[`InnerSymbolTable`-operation symbols..., target symbol]. The path must be non-empty
and may as minimum refer to a target symbol in the top-level `InnerRefNamespace`.

mortbopet: Refactored to:
> An attribute `InnerRefAttr` is provided to encapsulate references to inner 
symbols.  This attribute stores the path to an inner symbol through the
`InnerRefNamespace` symbol table hierarchy - this path being a combination of
[`InnerSymbolTable`-operation symbols..., target inner symbol]. The path must be non-empty.  
`InnerRefAttr` may also refer to top-level symbols (i.e. a single-element path containing only the target).
In this case, it is legal to refer to MLIR symbols residing at the top-level in
the `InnerRefNamespace`.

docs/RationaleSymbols.md
mortbopet: Added

mortbopet:
@dtzSiFive, @youngar, @teqdruid I'd really appreciate some eyes on this if you have time. I'm leaving for some time off come next week, and we have some internal work requiring this change, so I'm hoping to get this change in before then! 🙏

teqdruid (approved):
I'm not qualified to review the details of the core changes, but this all seems reasonable to me, modulo some nits.

include/circt/Dialect/Ibis/IbisOps.td
```diff
@@ -138,7 +139,6 @@ def InstanceOp : InstanceOpBase<"instance"> {
   let extraInstanceClassDefinition = [{
     ClassOp InstanceOp::getClass(const hw::InnerRefNamespace &ns) {
       return ns.lookupOp<ClassOp>(getTargetNameAttr());
-      // return dyn_cast_or_null<ClassOp>(getReferencedModule(&ns));
```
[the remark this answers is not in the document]
mortbopet: which has been removed 👀...

darthscsi (changes requested):
Changing to a block until at least field-sensitive inner-symbols being inner symbol tables is figured out and resolving the root of resolution for all things.

youngar:
Hey just want to let you know that I and other are looking at this PR and have been thinking about it at a high level, especially how should regular Symbols and InnerSymbols interact.  One thing that would be helpful is if the PR description was updated to cover how this works, or if the rationale captured these details.  I think a change that you will be making is that an InnerRef should encode if each ref in the array is a SymbolRef or an InnerSymbolRef.

I think one of the use cases needed is that you want an `ibis.container` op that is an InnerSymbol and InnerSymbolTable and can live directly under the `mlir.module` op, and you don't want to have to create a "`ibis.top_container`" op that defines a symbol just so it can live under the `mlir.module`, is that right?
```mlir
circt.container [IST] {
  ibis.container @Outer [IS, IST] {
    ibis.container @Inner [IS, IST] { }
  }
}
```

Can a SymbolTable op live under an InnerSymbolTable, and how do reference through those tables work?  I think its important that we make sure we don't break something like SymbolDCE by resolving references in a different way.
```mlir
  ; This example has a ref that wants to resolve a SymbolRef through an InnerSymbolTable.
  ; The SR will be resolved according to the parent SymbolTable (`circt.container`), not `@Bar` like intended. 
circt.container IRN, IST {
  hw.sym_table_op @Bar ST, IS {
    hw.module @Baz S { }
  }
  hw.ref InnerRef[InnerSymbolRef<"Bar">, SymbolRef<"Baz">]
}
```
It seems to me that it only makes sense for SymbolRefs to appear at the root of an InnerRefs, otherwise we are inevitably changing the resolution rules for SymbolRefs.

InnerSymbols support per-field symbols on aggregate objects, and I was wondering how that works with InnerSymbolTables.   Does it make sense to have per-field SymbolTables, or maybe SymbolTables can only be used when the op's InnerSymbol is not associated with a result and does not support per-field symbols (see `getTargetResultIndex` and `supportsPerFieldSymbols`).  When these aggregate ops are split into multiple ops (by something like FlattenIO/FIRRTLLowerTypes), where does that symbol table go?

lib/Dialect/HW/InnerSymbolTable.cpp
```diff
@@ -173,15 +382,32 @@ InnerSymbolTableCollection::populateAndVerifyTables(Operation *innerRefNSOp) {
         return op->hasTrait<OpTrait::InnerSymbolTable>();
       }));
 
-  // Ensure entries exist for each operation.
-  llvm::for_each(innerSymTableOps,
-                 [&](auto *op) { symbolTables.try_emplace(op, nullptr); });
+  // Prime the top-level symbol tables.
+  bool scanFailed = false;
+  llvm::for_each(innerSymTableOps, [&](auto *op) {
```
youngar, line 387: cc @dtzSiFive What if the InnerRefNamespace is also an InnerSymbolTable, you won't find the InnerSymbol defined under it, e.g.
```mlir
circt.container [IST, IRN] {
  ibis.container [IS, IST] { }
  }
```

We also allow InnerSymbols to be defined in nested regions:
```mlir
circt.container [IST, IRN] {
  circt.region_op {
    ibis.container [IS, IST] { }
  }
}
```

include/circt/Dialect/HW/HWOpInterfaces.h
```diff
@@ -77,25 +84,38 @@ class InnerRefNamespace : public TraitBase<ConcreteType, InnerRefNamespace> {
...
 };
 } // namespace OpTrait
 } // namespace mlir
 
 #include "circt/Dialect/HW/HWOpInterfaces.h.inc"
 
+namespace mlir {
+namespace OpTrait {
+template <typename ConcreteType>
+LogicalResult InnerSymbolTable<ConcreteType>::verifyRegionTrait(Operation *op) {
```
youngar, line 97: I don't think this code belongs under `verifyRegion`s.  This is usually used when the parent op needs to verify something about op's in its region.  I think this can be moved to the regular trait verifier.

youngar:
What happens when an op is both a Symbol and InnerSymbol, are they exclusive?

youngar:
Depending on how complex this is, it might make sense to create some test ops just for writing tests for this code.

youngar:
Symbol visibility usually determines if SymbolTables expose internal symbols, does this work for InnerSymbols and InnerSymbolTables? (And Symbols+InnerSymbolTables)
https://mlir.llvm.org/docs/SymbolsAndSymbolTables/#symbol-visibility

I think that InnerSymbols are currently missing the split between Nested and Private visibility right now.

lib/Dialect/HW/InnerSymbolTable.cpp
```diff
@@ -78,33 +147,124 @@ LogicalResult InnerSymbolTable::walkSymbols(Operation *op,
...
+  // target should be a nested symbol table.
+  auto nestedSymTbl = nestedSymbolTables.find(path.front());
+  if (nestedSymTbl == nestedSymbolTables.end())
+    return {};
+
+  // Perform the internal lookup into the nested symbol table.
+  return nestedSymTbl->second->lookup(path.drop_front());
+}
+
+void InnerSymbolTable::dump(llvm::raw_ostream &os, int indent) const {
```
dtzSiFive, line 237: `dump()` usually doesn't take an ostream/indent, so you can just `dump()`, say, from a debugger.

Usually it uses the `dbgs()` and calls some other method, e.g., `print`.

There are also macros for these so they play well with the larger LLVM project (drop in full release build, optionally enable them regardless, always have in debug builds), for example:
https://github.com/llvm/llvm-project/blob/26060de06378375f784b35df80ec6d43425e5908/llvm/lib/MC/MCInst.cpp#L65

include/circt/Dialect/HW/HWOpInterfaces.td
```diff
@@ -478,14 +478,22 @@ def InnerSymbol : OpInterface<"InnerSymbolOpInterface"> {
           this->getOperation()->removeAttr(InnerSymbolTable::getInnerSymbolAttrName());
       }]
     >,
-    InterfaceMethod<"Returns an InnerRef to this operation's top-level inner symbol, which must be present.",
+    InterfaceMethod<"Returns an InnerRef to this operation through the inner symbol table hierarchy of which it resides.",
```
dtzSiFive, line 481: "top-level" here wasn't re:root of the namespace, but clarifying this is for `fieldID=0`.

mortbopet:
@youngar, @dtzSiFive thank you for the reviews. Based on your feedback, and the fact that I'll be OOF for a few weeks come Wednesday, I think the scope of the changes/thinking required is larger than what I'll be able to do before then. Based on how critical this becomes internally, @teqdruid might chime in here in one of the following weeks. If not, I'll get back to working on this come June.

darthscsi:
> for instance, we are using `ibis.container` as the nest-able thing, that is, our `InnerSymbolTable` operation. If `ibis.container` were to implement an MLIR symbol, this would imply that an `ibis.container` itself would have to be a `SymbolTable` which would defeat the whole purpose of this.

For the nesting case, no.  `ibis.container` can have an optional symbol and an optional inner symbol, mutually exclusive.  Top level containers would define a symbol, nested ones would define an inner symbol.

One thing I'm wondering the more I think about this is why you need nested inner symbol tables.  If we assume the general practice in circt that (inner) symbol name doesn't imply user-visible name, and given that inner-symbols can be defined arbitrarily deep in region nesting, then the only thing nesting really gives you is aliasing in inner symbol values.  I can already have:
` ST{ op[S=foo] { op[IS=bar] { op[IS=baz] } }` and refer to `baz` uniquely and canonically with `hierpaths [foo, baz]`.  I do have to ensure `baz` isn't defined in other nestings, but since the name itself shouldn't have meaning, this shouldn't be a big deal.

teqdruid:
> One thing I'm wondering the more I think about this is why you need nested inner symbol tables. If we assume the general practice in circt that (inner) symbol name doesn't imply user-visible name, and given that inner-symbols can be defined arbitrarily deep in region nesting, then the only thing nesting really gives you is aliasing in inner symbol values. I can already have:
> ` ST{ op[S=foo] { op[IS=bar] { op[IS=baz] } }` and refer to `baz` uniquely and canonically with `hierpaths [foo, baz]`. I do have to ensure `baz` isn't defined in other nestings, but since the name itself shouldn't have meaning, this shouldn't be a big deal.

That's what we said in the meeting last Nov (IIRC). It just doesn't sit right with me (and even moreso with Morten). Why do you need nested symbols at all? Why not just have a flat symbol namespace and make all CIRCT symbols (to differentiate from upstream symbols) unique? This is essentially what we've been talking about with using opaque attributes, yes? (Frankly, I think this would save a lot of complexity so I'm beginning to get on board with it.) There's something arbitrary about two-level symbols which just doesn't sit right with us.

PR #6934, opened 2024-04-19
[Handshake] Add merge decomposition pattern

mortbopet:
This pass decomposes the (rather complex) semantics of a >2 input cmerge operation into a series of 2-input cmerge operations + supporting logic. This simpler structure is better suited for `dc` lowering, which only supports lowering 2-input control merge operations.  

Also decomposes >2 input `handshake.merge` operations into 2-input merges.

In pictures:
image

### DIFF
--- a/include/circt/Dialect/Handshake/HandshakeOps.td
+++ b/include/circt/Dialect/Handshake/HandshakeOps.td
@@ -447,12 +447,14 @@ def ControlMergeOp : Handshake_Op<"control_merge", [
   let results = (outs AnyType : $result, AnyType : $index);
   
   let builders = [OpBuilder<
-    (ins "ValueRange":$operands), [{
+    (ins "ValueRange":$operands, CArg<"Type", "{}">:$indexType), [{
       assert(!operands.empty() && "cmerge needs at least one operand");
       $_state.addOperands(operands);
+      if(!indexType)
+        indexType = $_builder.getIndexType();
       // By default, the index result has an Index type
       $_state.addTypes(ArrayRef<Type>{operands[0].getType(),
-                                      $_builder.getIndexType()});
+                                      indexType});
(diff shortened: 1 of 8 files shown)


mortbopet:
@Dinistro feel free to do a drive-by logic check here.

lib/Dialect/Handshake/Transforms/SplitCMerges.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+struct HandshakeSplitControlMerges
+    : public HandshakeSplitControlMergesBase<HandshakeSplitControlMerges> {
+  void runOnOperation() override {
+    ConversionTarget target(getContext());
+    target.addLegalDialect<handshake::HandshakeDialect, arith::ArithDialect>();
+    target.addDynamicallyLegalOp<handshake::ControlMergeOp>(
+        [&](handshake::ControlMergeOp op) { return op.getNumOperands() <= 2; });
+    RewritePatternSet patterns(&getContext());
+    patterns.insert<DeconstructCMergePattern>(&getContext());
```
Dinistro, line 96: The conversion frameworks seems quite overkill for this case, no? I suggest to use a normal greedy pattern rewrite to achieve the same thing.

Dinistro (commented):
This makes sense to me, but I'm currently on the move and could only give you an in-depth logical review next week.

lib/Dialect/Handshake/Transforms/SplitCMerges.cpp
mortbopet: Agree - changed to a greedy rewrite pattern.

Dinistro (approved):
LGTM, but maybe wait for someone else who is a bit fresher in this field than I am 🙂

PR #6935, opened 2024-04-19
[FIRRTL,PASS] Simple module summary looking for likely to dedup larger or numerous modules.  This is to guide developers in using chisel better.

darthscsi:


lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+  };
+
+  size_t countOps(FModuleOp mod) {
+    size_t retval = 0;
+    mod.walk([&](Operation *op) { retval += 1; });
+    return retval;
+  }
+
+  SmallVector<size_t> protSig(FModuleOp mod) {
```
mikeurbach, line 32: nit: `portSig`?

mikeurbach (approved):
Nice, looks good to me. We can tweak the stats this computes, but after seeing some results from this, the current stats are useful.

lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
```diff
@@ -0,0 +1,99 @@
...
+    SmallVector<size_t> ports;
+    for (auto p : mod.getPortTypes())
+      ports.push_back(cast<FIRRTLBaseType>(cast<TypeAttr>(p).getValue())
```
uenoku, line 35: Ports could have PropertyType or RefType s probably `cast<FIRRTLBaseType>` is not safe ?

lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
```diff
@@ -0,0 +1,99 @@
+//===- ModuleSummaryt.cpp ---------------------------------------*- C++ -*-===//
```
dtzSiFive, line 1:
```suggestion
//===- ModuleSummary.cpp ----------------------------------------*- C++ -*-===//
```
(not 100% on dash count, just added one because removed a char)

lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
```diff
@@ -0,0 +1,99 @@
...
+void ModuleSummaryPass::runOnOperation() {
+  auto circuit = getOperation();
+
+  using MapTy = DenseMap<KeyTy, SmallVector<FModuleOp>>;
+  MapTy data;
+
+  for (auto mod : circuit.getOps<FModuleOp>()) {
+    auto p = protSig(mod);
+    auto n = countOps(mod);
+    data[{p, n}].push_back(mod);
```
dtzSiFive, line 75: Nit:
Seems a bit heavy to have SmallVector's as keys in a DenseMap, especially since it's only used to group by hash not actually important to have the vector data itself.  Could have keys be port-size-hash plus op count or something.

Since a debugging pass maybe not worth worrying about but just a thought..

lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
```diff
@@ -0,0 +1,99 @@
...
+              return std::get<0>(lhs).opcount * std::get<1>(lhs).size() *
+                         std::get<1>(lhs).size() >
+                     std::get<0>(rhs).opcount * std::get<1>(rhs).size() *
+                         std::get<1>(rhs).size();
+            });
+  llvm::errs() << "cost, opcount, ports, modcount, examplename\n";
+  for (auto &p : sortedData)
+    if (p.second.size() > 1) {
+      llvm::errs() << (p.first.opcount * p.second.size() * p.second.size())
+                   << "," << p.first.opcount << ", " << p.first.portSizes.size()
```
dtzSiFive, line 90:
```suggestion
                   << ", " << p.first.opcount << ", " << p.first.portSizes.size()
```
Consistent space after the comma?

Consider one of the interleave methods.

Would be nice to have passes like this abstracted over the raw_ostream to use so it's easy to use this in a tool and have it go to a file or something.  Not sure about answers there but seems likely to have prior art.

Just curious, what's the expected story for a user to invoke this?

lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
```diff
@@ -0,0 +1,99 @@
+//===- ModuleSummaryt.cpp ---------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "PassDetails.h"
+#include "mlir/IR/Threading.h"
```
dtzSiFive, line 10: I don't think threading is needed? But didn't check.

dtzSiFive (approved):
LGTM.  I left some comments but it's simple in a good way so works for me.

Useful indeed.

lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
darthscsi: The key is the vector of the bitwidth of each port.  so differences in port widths or differences in number of ports should both cause the hash to be different.

dtzSiFive: Right! That's why it'd be port-size-hash and not just port-size as part of the key.  We only need the actual vector to get enough to know which summary entry to bin this with, don't really need the vector itself if that makes sense.

However having it in the key helps with equality-checking beyond hash-equivalence and welp anyway not sure how to slot this in neatly and if it matters we'll know and probably other things will fall over first anyway.

Marking resolved, thanks for the response.

lib/Dialect/FIRRTL/Transforms/ModuleSummary.cpp
darthscsi: Not when I was running it, but where I want to, yes.

PR #6936, opened 2024-04-19
[FIRRTL] Error if asked to add a port to a public module.

dtzSiFive:
Add option to allow this for compatibility.

Add statistic to count ports added to public modules, as an interesting bit of information but particularly to track how many are still being added when updating a code base to avoid this.

### DIFF
--- a/include/circt/Dialect/FIRRTL/Passes.td
+++ b/include/circt/Dialect/FIRRTL/Passes.td
@@ -41,6 +41,8 @@ def LowerFIRRTLAnnotations : Pass<"firrtl-lower-annotations", "firrtl::CircuitOp
       "Ignore unknown annotations.">,
     Option<"noRefTypePorts", "no-ref-type-ports", "bool", "false",
       "Create normal ports, not ref type ports.">,
+    Option<"allowAddingPortsOnPublic", "allow-adding-ports-on-public", "bool", "false",
+      "Allow public modules to gain additional ports as a result of wiring.">,
   ];
   let dependentDialects = ["hw::HWDialect"];
   let statistics = [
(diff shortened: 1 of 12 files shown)


lib/Firtool/Firtool.cpp
```diff
@@ -415,6 +416,11 @@ struct FirtoolCmdOptions {
           "wiring problems inside the LowerAnnotations pass"),
       llvm::cl::init(false), llvm::cl::Hidden};
 
+  llvm::cl::opt<bool> allowAddingPortsOnPublic{
+      "allow-adding-ports-on-public",
```
seldridge, line 420: Nit: 
```suggestion
      "allow-adding-ports-on-public-modules",
```

include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -41,6 +41,8 @@ def LowerFIRRTLAnnotations : Pass<"firrtl-lower-annotations", "firrtl::CircuitOp
       "Ignore unknown annotations.">,
     Option<"noRefTypePorts", "no-ref-type-ports", "bool", "false",
       "Create normal ports, not ref type ports.">,
+    Option<"allowAddingPortsOnPublic", "allow-adding-ports-on-public", "bool", "false",
```
seldridge, line 44: Nit:
```suggestion
    Option<"allowAddingPortsOnPublic", "allow-adding-ports-on-public-modules", "bool", "false",
```

seldridge (approved):
LGTM

This needs to update the C-API for `Firtool.h`. After that is fixed, this is good to land.

CC: @sequencer as this likely affects bindings.

rwy7 (approved):

PR #6937, opened 2024-04-19
[FIRRTL] Disambiguate paths when possible in ResolvePaths.

mikeurbach:
We would previously error if paths are contained in multiply instantiated modules, because there is no unique path in this case. However, we could expand out the path into multiple unique paths in this case, but this is only safe to do if the path is already used in a list where multiple paths are acceptable.

This relaxes the requirement and implements the path expansion in this scenario. When we hit a multiply instantiated module, we expand out all the instance paths to that module from the owning module, and create multiple annotations and resolved paths.


mikeurbach:
It helps slightly to review this with whitespace-only diffs removed.

prithayan (approved):
Looks good.

mikeurbach:
Thanks @prithayan!

PR #6940, opened 2024-04-22
[MooreToCore] Lower moore operators into comb or hw.

hailongSun2000:
Aim to lower `moore` elementary operators into `comb` or `hw`, excluding `assignments` and `variable declarations`. The latter needs to analyze the entire IR to map variables and their values. And `moore` without `mux(?  :)`, which is substituted by `scf.if`.
The elementary operators include `unary, binary, relational, shift, constant, concatenation, extract, and conversion` operations.
The `moore.extract` operation receives a variable as its operand, but `comb.extract` does not. I asked @fabianschuiki for help. He taught me maybe I could use `comb.shr` to bring a variable down to bit position 0, and then capture the correct number of bits from 0 with `comb.extract`.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -56,6 +56,51 @@ static Value adjustIntegerWidth(OpBuilder &builder, Value value,
   return builder.create<comb::MuxOp>(loc, isZero, lo, max, false);
 }
 
+/// Due to the result type of the `lt`, or `le`, or `gt`, or `ge` ops are
+/// always unsigned, estimating their operands type.
+static bool isSignedType(Operation *op) {
+  return TypeSwitch<Operation *, bool>(op)
+      .template Case<LtOp, LeOp, GtOp, GeOp>([&](auto op) -> bool {
+        return cast<UnpackedType>(op->getOperand(0).getType())
+                   .castToSimpleBitVector()
+                   .isSigned() &&
+               cast<UnpackedType>(op->getOperand(1).getType())
+                   .castToSimpleBitVector()
+                   .isSigned();
+      })
+      .Default([&](auto op) -> bool {
+        return cast<UnpackedType>(op->getResult(0).getType())
+            .castToSimpleBitVector()
+            .isSigned();
+      });
+}
+
+/// Not define the predicate for `relation` and `equality` operations in the
+/// MooreDialect, but comb needs it. Return a correct `comb::ICmpPredicate`
+/// corresponding to different moore `relation` and `equality` operations.
+static comb::ICmpPredicate getCombPredicate(Operation *op) {
+  using comb::ICmpPredicate;
+  return TypeSwitch<Operation *, ICmpPredicate>(op)
+      .Case<LtOp>([&](auto op) {
+        return isSignedType(op) ? ICmpPredicate::slt : ICmpPredicate::ult;
+      })
+      .Case<LeOp>([&](auto op) {
+        return isSignedType(op) ? ICmpPredicate::sle : ICmpPredicate::ule;
+      })
+      .Case<GtOp>([&](auto op) {
+        return isSignedType(op) ? ICmpPredicate::sgt : ICmpPredicate::ugt;
+      })
+      .Case<GeOp>([&](auto op) {
+        return isSignedType(op) ? ICmpPredicate::sge : ICmpPredicate::uge;
+      })
+      .Case<EqOp>([&](auto op) { return ICmpPredicate::eq; })
+      .Case<NeOp>([&](auto op) { return ICmpPredicate::ne; })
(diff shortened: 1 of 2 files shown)


hailongSun2000:
Thanks for the code review in advance! :heart:

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -283,16 +449,39 @@ static void populateOpConversion(RewritePatternSet &patterns,
...
-    ShrOpConversion,
-    AShrOpConversion,
-    UnrealizedConversionCastConversion
+    // Patterns of miscellaneous operations.
+    ConstantOpConv, ConcatOpConversion, ReplicateOpConversion,
+    ExtractOpConversion, ConversionOpConversion,
+
+    // Patterns of unary operations.
+    UnaryOpConversion<BoolCastOp>, UnaryOpConversion<ReduceAndOp>,
+    UnaryOpConversion<ReduceOrOp>, UnaryOpConversion<ReduceXorOp>,
```
fabianschuiki, line 458: I think these operations might still be missing. `UnaryOpConversion` above creates a `comb::ParityOp` for all of these, which feels wrong except for `ReduceXorOp`. I think you'll want something like the following:

- `BoolCastOp`: carefully convert the operand to its "truthy" equivalent. If the operand can be converted to a simple bit vector, this should convert to that simple bit vector and then check if it is `!= 0`. All other cases we could just reject for now, since they will be more complicated (stuff like objects, classes, virtual interfaces, etc.).
- `ReduceAndOp`: should probably be something like `== <all-ones-value>`.
- `ReduceOrOp`: should probably be something like `!= 0`.
- `ReduceXorOp`: works as `ParityOp`.

hailongSun2000: Hey, @fabianschuiki. I reimplement these conversions separately. Do I correctly understand the concept of these `UnaryOp` this time?

hailongSun2000: So `comb.parity` just with correspondence `moore.reduceXor`. Other unary ops above are not applicable with even/odd checks. Right?

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -82,6 +127,169 @@ struct ConcatOpConversion : public OpConversionPattern<ConcatOp> {
...
+    if (isa<DivOp>(op) && isSignedType(op)) {
+      rewriter.replaceOpWithNewOp<comb::DivSOp>(op, adaptor.getLhs(),
+                                                adaptor.getRhs(), false);
+      return success();
+    }
+    if (isa<ModOp>(op) && isSignedType(op)) {
+      rewriter.replaceOpWithNewOp<comb::ModSOp>(op, adaptor.getLhs(),
+                                                adaptor.getRhs(), false);
+      return success();
+    }
```
maerhart, line 251: This special case isn't particularly elegant. Maybe we should create another pattern with 3 template arguments so that the signed version of an op can be given as an argument.

hailongSun2000: Get it. I'll modify it.

hailongSun2000: > This special case isn't particularly elegant. Maybe we should create another pattern with 3 template arguments so that the signed version of an op can be given as an argument.

How about this?

maerhart: Very nice!

hailongSun2000: > Very nice!

Thanks so much for pointing out the code inadequacies!!! :heart:

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -82,6 +127,163 @@ struct ConcatOpConversion : public OpConversionPattern<ConcatOp> {
...
+            .castToSimpleBitVectorOrNull()) {
+      Type resultType = typeConverter->convertType(op.getInput().getType());
+      Value zero = rewriter.create<hw::ConstantOp>(op->getLoc(), resultType, 0);
+
+      rewriter.replaceOpWithNewOp<comb::ICmpOp>(op, comb::ICmpPredicate::ne,
+                                                adaptor.getInput(), zero);
+      return success();
+    }
+
+    return success();
```
fabianschuiki, line 216:
```suggestion
    return failure();
```

fabianschuiki (approved):
Looks good to me once `BoolCastOpConversion` returns `failure()` in case the operand isn't a simple bit vector 😃. Thanks for adding these conversions! Really cool to see SV get a path into the core dialects.

hailongSun2000:
> Looks good to me once `BoolCastOpConversion` returns `failure()` in case the operand isn't a simple bit vector 😃. Thanks for adding these conversions! Really cool to see SV get a path into the core dialects.

It's an honor for me to participate in this! :smile:

PR #6941, opened 2024-04-22
[HandshakeToDC] Add pack/unpack lowering patterns

mortbopet:
Adds lowering patterns for `handshake.pack/unpack` to DC. For the actual (data-side) tuple operations, `hw.struct`s are used. This is purely for pragmatic reasons; Firstly, `hw.struct` is for all intents and purposes a generic struct type. Secondly, we don't want data-only operations to leak into `DC`.

### DIFF
--- a/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
+++ b/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
@@ -65,12 +65,12 @@ static Value pack(OpBuilder &b, Value token, Value data = {}) {
 }
 
 // NOLINTNEXTLINE(misc-no-recursion)
-static Type tupleToStruct(TupleType tuple) {
+static StructType tupleToStruct(TupleType tuple) {
   auto *ctx = tuple.getContext();
   mlir::SmallVector<hw::StructType::FieldInfo, 8> hwfields;
   for (auto [i, innerType] : llvm::enumerate(tuple)) {
     Type convertedInnerType = innerType;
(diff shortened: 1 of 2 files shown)


teqdruid:
hw.struct is the struct type in CIRCT and both Handshake and DC are in CIRCT. It's the practical choice. And it's not tied to hardware in any particular semantic -- just the name.  We wouldn't have it there if there was a struct type upstream we could use!

I think keeping the argument names around is more important than dialect name purity. And I know you share my aversion to duplicating stuff in each dialect.

Plus, there's an argument to be made that there's nothing in the hw or comb dialects that are specifically tied to hardware.

mortbopet:
> Plus, there's an argument to be made that there's nothing in the hw or comb dialects that are specifically tied to hardware.

... except for - for the former - the dialect name itself 🙂.

You have a good point, hw.struct is - if we ignore the dialect prefix - an overall more well suited type for this than `tuple`.

Dinistro:
Side note: There was some discussion at EuroLLVM about introducing a general dialect to model struct types in MLIR. There are multiple places where this would be helpful, i.e., Flang, LLVM dialect, CIR, and CIRCT.
I'm not sure what the status of this is, but it might make sense to try and raise this properly on discourse.

mortbopet:
@teqdruid updated the PR s.t. this is now strictly a `HandshakeToDC` PR, which uses `hw.struct`.

teqdruid (approved):

PR #6942, opened 2024-04-22
[HandshakeToDC] Fix constant unit-rate op lowering

mortbopet:
Constant-like unit-rate operations didn't have any output tokens assigned. Fixed s.t. constant-like operations (with no inputs) are emitted using a `dc.source`.

### DIFF
--- a/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
+++ b/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
@@ -308,15 +308,27 @@ public:
   matchAndRewrite(Operation *op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const override {
     llvm::SmallVector<Value, 4> inputData;
-    llvm::SmallVector<Value, 4> inputTokens;
-    for (auto input : operands) {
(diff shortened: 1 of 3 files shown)


lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
```diff
@@ -308,16 +308,28 @@ struct UnitRateConversionPattern : public ConversionPattern {
...
+      if (!op->hasTrait<OpTrait::ConstantLike>())
+        return op->emitOpError(
+            "no-operand operation which isn't constant-like. Too dangerous "
+            "to assume semantics - won't convert");
+
+      // Constant-like operation; assume the token can be represented as a
+      // constant `dc.source`.
+      outToken = rewriter.create<dc::SourceOp>(op->getLoc());
+    } else {
+      llvm::SmallVector<Value, 4> inputTokens;
```
Dinistro, line 322:
```suggestion
     SmallVector<Value> inputTokens;
```
Nit: I would not specify a size if there is no particular reason. I rather suggest to call reserve for such cases, as this ensures that there will be at most one heap allocation.

Dinistro (approved):
LGTM % nit comment.

PR #6943, opened 2024-04-22
[DC] Add merge lowering

mortbopet:
Modelled after the Handshake lowering, however, becomes trivial with a 2-input merge.

### DIFF
--- a/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
+++ b/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
@@ -198,6 +198,50 @@ public:
   }
 };
 
+class MergeOpConversion : public DCOpConversionPattern<handshake::MergeOp> {
+public:
+  using DCOpConversionPattern<handshake::MergeOp>::DCOpConversionPattern;
+  using OpAdaptor = typename handshake::MergeOp::Adaptor;
+
+  LogicalResult
+  matchAndRewrite(handshake::MergeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (op.getNumOperands() > 2)
+      return op.emitOpError("only two inputs supported");
+
+    llvm::SmallVector<Value, 4> tokens, data;
+
+    for (auto input : adaptor.getDataOperands()) {
+      auto up = unpack(rewriter, input);
+      tokens.push_back(up.token);
+      if (up.data)
+        data.push_back(up.data);
+    }
+
+    // Control side
+    Value selectedIndex = rewriter.create<dc::MergeOp>(op.getLoc(), tokens);
+    auto selectedIndexUnpacked = unpack(rewriter, selectedIndex);
+    Value mergeOutput;
+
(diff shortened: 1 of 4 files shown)


lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
```diff
@@ -198,6 +198,50 @@ class JoinOpConversion : public DCOpConversionPattern<handshake::JoinOp> {
...
+    if (!data.empty()) {
+      // Data-merge; mux the selected input.
+      auto dataMux = rewriter.create<arith::SelectOp>(
+          op.getLoc(), selectedIndexUnpacked.data, data[0], data[1]);
+      convertedOps->insert(dataMux);
+
+      // Pack the data mux with the control token.
+      mergeOutput = pack(rewriter, selectedIndexUnpacked.token, dataMux);
+    } else {
+      // Control-only merge; throw away the index value of the dc.merge
```
teqdruid, line 235: - Unlike `dc.merge`, `handshake.merge` doesn't output the idx of the value which got selected. This is why you had alarm bells going off.
- When would a `handshake.merge` have no data? Since handshake is all about dynamic data flow, I don't think it ever would. In the context of a lowering, even `none` would be considered data, yes?
- If `none` is implicitly converted to `dc.token`, then this makes sense.

teqdruid (approved):
This looks about right to me.

lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
```diff
@@ -198,6 +198,50 @@ class JoinOpConversion : public DCOpConversionPattern<handshake::JoinOp> {
...
+    }
+
+    // Control side
+    Value selectedIndex = rewriter.create<dc::MergeOp>(op.getLoc(), tokens);
+    auto selectedIndexUnpacked = unpack(rewriter, selectedIndex);
+    Value mergeOutput;
+
+    if (!data.empty()) {
+      // Data-merge; mux the selected input.
+      auto dataMux = rewriter.create<arith::SelectOp>(
```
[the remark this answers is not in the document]
mortbopet: `map-arith-to-comb` will lower this to a `comb.mux`.

lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
mortbopet: > If none is implicitly converted to dc.token, then this makes sense.
it is - see https://github.com/llvm/circt/blob/b9f8bc610346a2efa6de0cdc65f893e4046f8293/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp#L66

> When would a handshake.merge have no data? Since handshake is all about dynamic data flow, I don't think it ever would. In the context of a lowering, even none would be considered data, yes?

True, in practice, i've never seen this used. The DHLS lowerings used data `merge` ops but these could always be boiled down to single-input merges (i.e. a no-op). So your observation is correct, and there is a good argument for removing `handshake.merge` from the dialect.

PR #6944, opened 2024-04-22
[FIRRTL][CAPI] Add function for getting mask type

SpriteOvO:
Used for building `firrtl.mem` from external programs.

https://github.com/llvm/circt/blob/db973f13193728d082cb8bd8873ae524575291e0/lib/Dialect/FIRRTL/FIRRTLOps.cpp#L2982


mikeurbach (approved):
Makes sense, thanks for adding nice unit tests.

PR #6945, opened 2024-04-23
[CombToArith] Fix coarsening of division by zero UB

### LINKED ISSUE #6938: [Arc] Generates signalling code
`comb.divs` (and probably `comb.divu`) generate code that causes a `SIGFPE` signal on my system, due to division by zero.

Below is a small sample that should be equivalent to `b == 0 ? 0 : a / b`, which results in
```
$ ./build/bin/arcilator --jit-entry=main --run build/bin/div.mlir 
output = 7ffeb8e6e2a0
```
when run with `--jit`.

```
// RUN: arcilator %s --run --jit-entry=main | FileCheck %s
// REQUIRES: arcilator-jit
// CHECK: output = 0

hw.module @Baz(in %a: i8, in %b: i8, out c: i8) {
  %zero = hw.constant 0 : i8
  %0 = comb.divs %a, %b : i8
  %2 = comb.icmp bin eq %zero, %b : i8
  %3 = comb.mux bin %2, %zero, %0 : i8
  hw.output %3: i8
}

func.func @main() {
  %eight = arith.constant 8 : i8
  %zero = arith.constant 0 : i8

  arc.sim.instantiate @Baz as %model {
    arc.sim.set_input %model, "a" = %eight : i8, !arc.sim.instance<@Baz>
    arc.sim.set_input %model, "b" = %zero : i8, !arc.sim.instance<@Baz>
    
    arc.sim.step %model : !arc.sim.instance<@Baz>
  
    %res = arc.sim.get_port %model, "c" : i8, !arc.sim.instance<@Baz>
    arc.sim.emit "output", %res : i8
  }

  return
}
```

maerhart:
Fixes #6938

### DIFF
--- a/integration_test/arcilator/JIT/div-by-zero.mlir
+++ b/integration_test/arcilator/JIT/div-by-zero.mlir
@@ -1,0 +1,49 @@
+// RUN: arcilator %s --run --jit-entry=main | FileCheck %s
+// REQUIRES: arcilator-jit
+// CHECK: divu = 0{{$}}
+// CHECK: divs = 0{{$}}
+// CHECK: modu = 0{{$}}
+// CHECK: mods = 0{{$}}
+
+hw.module @Baz(in %a: i8, in %b: i8, out divu: i8, out divs: i8, out modu: i8, out mods: i8) {
+  %zero = hw.constant 0 : i8
+  %is_zero = comb.icmp eq %zero, %b : i8
+
+  %0 = comb.divu %a, %b : i8
+  %1 = comb.mux %is_zero, %zero, %0 : i8
+
+  %2 = comb.divs %a, %b : i8
+  %3 = comb.mux %is_zero, %zero, %2 : i8
+
+  %4 = comb.modu %a, %b : i8
+  %5 = comb.mux %is_zero, %zero, %4 : i8
+
+  %6 = comb.mods %a, %b : i8
+  %7 = comb.mux %is_zero, %zero, %6 : i8
+
+  hw.output %1, %3, %5, %7 : i8, i8, i8, i8
+}
+
+func.func @main() {
+  %eight = arith.constant 8 : i8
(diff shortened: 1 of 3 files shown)


fzi-hielscher (commented):
Thanks! This will probably do for most practical cases. But I am unsure if we may not still be in trouble if we try to do `0 / 0`. Going by the specification of the arith dialect, any division by zero is (immediate) UB. [Alive2 also confirms this](https://alive2.llvm.org/ce/z/NNmpbk) if we compare a clang compilation of the `b == 0 ? 0 : a/b` reference against a slightly massaged output of the integration test:
```
Transformation doesn't verify!

ERROR: Source is more defined than target

Example:
i8 %#0 = #x00 (0)
i8 %#1 = #x00 (0)

Source:
i1 %#5 = #x1 (1)
  >> Jump to %#4
i8 %#7 = #x00 (0)

Target:
i1 %#3 = #x1 (1)
i8 %#4 = #x00 (0)
i8 %#5 = UB triggered!
```

For practical purposes, I tested it on both x86_64 and ARM64 and the test produced 0 even for zero divded by zero. This is not so surprising for ARM64. I would have expected a trap on x86_64 though, see #6726 . This could be a subtlety of the JIT engine and the way it handles UB. But maybe I've just messed something up, which wouldn't be surprising either. 😖 

On a more general note, I do think we need to tinker with the HW/Comb specification to get a clean separation between undefined behavior, undefined values (undef, poison, frozen poison, X, you name it...) and the conditions which must/may/mustn't terminate a simulation.

fabianschuiki (approved):
LGTM!

darthscsi (approved):
LGTM

maerhart:
@fzi-hielscher Thanks for the detailed review! I've changed it to divide by 1 if the divisor is 0 originally.

fzi-hielscher (approved):
Nice, thank you.

PR #6946, opened 2024-04-23
[ESI] Add cmake switch to enable ESI_COSIM

mortbopet:
It is mildly annoying to have cosim-related capabilities silently failing late in the ESI process (e.g. funky import errors due to linking issues:
```
ImportError: /home/mpetersen/Work/FPGAToolsAndCompilers/CIRCT/build-Debug/tools/circt/lib/Dialect/ESI/runtime/python/esi/esiCppAccel.cpython-310-x86_64-linux-gnu.so: undefined symbol: 
_ZTIN3esi8backends5cosim16CosimAcceleratorE
```
only to find out that cosimulation was not enabled in the build. Unless there's a good reason for ESI to implicitly decide when or when not to include cosim support, i think ESI_COSIM should be opt-in. If not, at least until the ESI cosim path is more fleshed out and robust s.t. the above errors don't occur at a late stage.

### DIFF
--- a/lib/Dialect/ESI/runtime/CMakeLists.txt
+++ b/lib/Dialect/ESI/runtime/CMakeLists.txt
@@ -77,38 +77,46 @@ IF(MSVC)
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS 1)
 ENDIF(MSVC)
 
-# If Cap'nProto hasn't been explicitly disabled, find it.
-option(CAPNP_DISABLE "Disable Cap'nProto (needed for cosimulation).")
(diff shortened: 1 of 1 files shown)


teqdruid:
I don't think this solves that particular problem. I would imagine it still occurs when `ESI_COSIM` is disabled. And I _though_ I had solved this problem. In the C++ code, it is sufficient to just not compile or link in the backend. There is a hack in the Python bindings, however, that requires it. The real (short term) fix is to conditionally disable that hack.

teqdruid:
Fixed in the right short-term way 20ef1a2cb. I'm gonna close this PR.

PR #6950, opened 2024-04-25
[ImportVerilog] Add conditional operator.

angelzzzzz:
Handle conditional operator `?:`.
Use `scf.if` to control the selection of expressions, and return the selected value through `scf.yield`.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -223,17 +223,18 @@ def NetOp : MooreOp<"net", [
 ]> {
   let summary = "A net declaration";
   let description = [{
-    The `moore.net' operation is a net declaration. Net types defines different types
-    of net connection in SV. There are twelve built-in net types defined in the official
-    standard construct of the operation:
-    `supply0`, `supply1`, `tri`, `triand`, `trior`, `trireg`, `tri0`, `tri1`, `uwire`,
-    `wire`, `wand`, `wor`.
-    Optional assignment argument allows net operation to be initialized with specific
-    values as soon as it is created. Only one net declaration assignment can be made for
-    a particular net. See IEEE 1800-2017 § 10.3.1 "The net declaration assignment" for
-    the differences between net declaration assignments and continuous assign statements.
-    It has some features that are not supported: declaring an interconnect net and using
-    user-defined types in the net operation.
+    The `moore.net' operation is a net declaration. Net types defines different
+    types of net connection in SV. There are twelve built-in net types defined
+    in the official standard construct of the operation:
+    `supply0`, `supply1`, `tri`, `triand`, `trior`, `trireg`, `tri0`, `tri1`,
+    `uwire`, `wire`, `wand`, `wor`.
+    Optional assignment argument allows net operation to be initialized with
+    specific values as soon as it is created. Only one net declaration
+    assignment can be made for a particular net. See IEEE 1800-2017 § 10.3.1
+    "The net declaration assignment" for the differences between net declaration
+    assignments and continuous assign statements. It has some features that are
+    not supported: declaring an interconnect net and using user-defined types in
+    the net operation.
     
     See IEEE 1800-2017 § 6.7 "Net declarations".
   }];
@@ -961,27 +962,27 @@ def StructInjectOp : MooreOp<"struct_inject"> {
 def UnionCreateOp : MooreOp<"union_create"> {
   let summary = "Union Create operation";
   let description = [{
-  A union is a data type that represents a single piece
-  of storage that can be accessed using one of
-  the named member data types. Only one of the
-  data types in the union can be used at a time.
-  By default, a union is unpacked, meaning there
-  is no required representation for how members
-  of the union are stored. Dynamic types and chandle
-  types can only be used in tagged unions.
-  See IEEE 1800-2017 § 7.3 "Unions"
+    A union is a data type that represents a single piece
+    of storage that can be accessed using one of
+    the named member data types. Only one of the
+    data types in the union can be used at a time.
+    By default, a union is unpacked, meaning there
+    is no required representation for how members
+    of the union are stored. Dynamic types and chandle
+    types can only be used in tagged unions.
+    See IEEE 1800-2017 § 7.3 "Unions"
 
-  Example:
-  ```
-  typedef union { int i; shortreal f; } num; // named union type
-  num n;
-  n.f = 0.0; // set n in floating point format
-  typedef struct {
-  bit isfloat;
-  union { int i; shortreal f; } n;           // anonymous union type
-  } tagged_st;                               // named structure
-  ```
-  See IEEE 1800-2017 § 7.3 "Unions"
+    Example:
+    ```
+    typedef union { int i; shortreal f; } num; // named union type
+    num n;
+    n.f = 0.0; // set n in floating point format
+    typedef struct {
+    bit isfloat;
+    union { int i; shortreal f; } n;           // anonymous union type
+    } tagged_st;                               // named structure
+    ```
+    See IEEE 1800-2017 § 7.3 "Unions"
   }];
   let arguments = (ins StrAttr:$unionName, UnpackedType:$input);
   let results = (outs UnpackedType:$result);
@@ -989,30 +990,30 @@ def UnionCreateOp : MooreOp<"union_create"> {
   $unionName `,` $input attr-dict `:`
   type($input) `->` type($result)
   }];
- }
+}
 
- def UnionExtractOp : MooreOp<"union_extract"> {
+def UnionExtractOp : MooreOp<"union_extract"> {
   let summary = "Union Extract operation";
   let description = [{
-  With packed unions, writing one member and reading another is
-  independent of the byte ordering of the machine,
-  unlike an unpacked union of unpacked structures,
-  which are C-compatible and have members in ascending address order.
-  See IEEE 1800-2017 § 7.3.1 "Packed unions"
+    With packed unions, writing one member and reading another is
+    independent of the byte ordering of the machine,
+    unlike an unpacked union of unpacked structures,
+    which are C-compatible and have members in ascending address order.
+    See IEEE 1800-2017 § 7.3.1 "Packed unions"
 
-  Example:
-  ```
-  typedef union packed { // default unsigned
-  s_atmcell acell;
-  bit [423:0] bit_slice;
-  bit [52:0][7:0] byte_slice;
-  } u_atmcell;
-  u_atmcell u1;
-  byte b; bit [3:0] nib;
-  b = u1.bit_slice[415:408]; // same as b = u1.byte_slice[51];
-  nib = u1.bit_slice [423:420];
-  ```
-  See IEEE 1800-2017 § 7.3.1 "Packed unions"
+    Example:
+    ```
+    typedef union packed { // default unsigned
+    s_atmcell acell;
+    bit [423:0] bit_slice;
+    bit [52:0][7:0] byte_slice;
+    } u_atmcell;
+    u_atmcell u1;
+    byte b; bit [3:0] nib;
+    b = u1.bit_slice[415:408]; // same as b = u1.byte_slice[51];
+    nib = u1.bit_slice [423:420];
+    ```
+    See IEEE 1800-2017 § 7.3.1 "Packed unions"
   }];
   let arguments = (ins StrAttr:$memberName, UnpackedType:$input);
   let results = (outs UnpackedType:$result);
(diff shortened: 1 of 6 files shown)


hailongSun2000 (approved):
LGTM!

maerhart:
The lowering does not match the standard for 4-valued conditions with X or Z. Apparently it should then evaluate both the 2nd and 3rd operand, compare them for equivalence, etc.
The tests only seem to consider 2-valued conditions. Should we display an error for 4-valued conditions for now?

angelzzzzz:
> The lowering does not match the standard for 4-valued conditions with X or Z. Apparently it should then evaluate both the 2nd and 3rd operand, compare them for equivalence, etc. The tests only seem to consider 2-valued conditions. Should we display an error for 4-valued conditions for now?

Good point! I have added some error tests for 4-valued conditions and look forward to your suggestions.

maerhart:
> Good point! I have added some error tests for 4-valued conditions and look forward to your suggestions.

Thanks! This only concerns literals now, right? What if the condition expression is not a literal but, e.g., some input port that is 4-valued and may contain 'x'? Can't we check the type of the expression for 4-valuedness instead of the expression itself, or would that be too restrictive for our use-cases?

angelzzzzz:
> Thanks! This only concerns literals now, right? What if the condition expression is not a literal but, e.g., some input port that is 4-valued and may contain 'x'? Can't we check the type of the expression for 4-valuedness instead of the expression itself, or would that be too restrictive for our use-cases?

Thanks for your thoughtful consideration! Do you mean we need to check the runtime values? If so, I guess we can only analyze the type of expression at compile time rather than at runtime. Additionally, we cannot determine whether it is x or z based on the type of expression, e.g., x and z may be `logic` or `reg` type.

fabianschuiki:
I think what @maerhart is getting at is the complicated handling that is necessary when the condition is X or Z (*"11.4.11 Conditional operator"* in IEEE 1800-2017). When the condition is of a two-valued type (derived from `bit`), your `scf.if` solution is perfect. You either evaluate the left or right expression, depending on whether the condition is 1 or 0. If the condition is of a four-valued type however (derived from `logic`), and it is X or Z, you have to evaluate *both* the left *and* right expression, and merge them. The merging is very annoying -- there's an entire list of rules mentioned in 11.4.11. It goes something like the following:

1. If the condition is 1, evaluate and return the left expression. If the condition is 0, evaluate and return the right expression. Otherwise:
2. Evaluate the left and right expression. Check them for equality (`==` or `===`, not sure which). If they are equal, return one of them. Otherwise:
3. If the left and right expression are of integral data type (can be cast to a simple bit vector), apply a special bitwise truth table mentioned in 11.4.11 between the left and right expression. Otherwise:
4. Follow the remaining rules in 11.4.11.
5. If no rule was able to handle the ambiguity, return the default value for the type of the condition (0 for `bit` and X for `logic`).

So I think one thing you could do is explicitly check if the condition is 0 or 1, and otherwise print an error, raise an exception, or do something else. For example, pseudocode:
```
scf.if (cond === 0) {
  %0 = eval leftExpr
  scf.yield %0
} else {
  scf.if (cond === 1) {
    %1 = eval rightExpr
    scf.yield %1
  } else {
    verif.assert %false
  }
}
```
That would cause the simulation to fail properly if we encounter an unsupported condition in the mux, and we know where to start once we improve the lowering in the future. We'll probably want to have a dedicated op in the Moore dialect that performs the mux/conditional merging of the left and right expression. And we'll want to eventually implement all the rules in 11.4.11 and apply the correct one depending on the data type of what we're muxing.

You can also kick the can down the road and introduce a new op that is similar to `scf.if` but encodes the special behaviour of the expression:
```
moore.conditional %cond {
  moore.yield leftExpr
} {
  moore.yield rightExpr
}
```
That would allow you to have a later lowering pass that takes `moore.conditional` and replaces it with the correct behaviour. And encapsulating the left and right expression in a region captures the fact that you may only be evaluating one of the expressions depending on the situation.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -958,4 +958,69 @@ def UnionCreateOp : MooreOp<"union_create"> {
...
+  let summary = "Conditional operation";
+  let description = [{
+    If cond_predicate is true, the operator returns the value of the first 
+    expression without evaluating the second expression; if false, it returns 
+    the value of the second expression without evaluating the first expression.
+    If cond_predicate evaluates to an ambiguous value (x or z), then both the
+    first expression and the second expression shall be evaluated, and compared
+    for logical equivalence. If that comparison is true (1), the operator shall
+    return either the first or second expression. Otherwise the operator returns
+    a result based on the data types of the expressions.
```
hailongSun2000, line 976: Can I consider logical equivalence as `<->`?

angelzzzzz:
@fabianschuiki Thank you very much for your suggestion :smiley: ! I chose the second method and defined two new operations for the conditional operator: `moore.conditional` and `moore.yield`.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -514,6 +514,47 @@ struct ExprVisitor {
...
+    // Handle left expression.
+    builder.setInsertionPointToStart(conditionalOp.getBody(0));
+    auto trueValue = context.convertExpression(expr.left());
+    if (!trueValue)
+      // When `trueValue` is empty (e.g. expression is X or Z for now) , use
+      // builder without value to avoid errors.
+      builder.create<moore::YieldOp>(loc);
+    else
+      builder.create<moore::YieldOp>(loc, convertToSimpleBitVector(trueValue));
+
```
hailongSun2000, line 544: `convertToSimpleBitVector()` may return `{}` if the type of trueValue isn't SBVT(Like `logic [0:5][2:0] p3`).

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -514,6 +514,47 @@ struct ExprVisitor {
...
+    // Handle right expression.
+    builder.setInsertionPointToStart(conditionalOp.getBody(1));
+    auto falseValue = context.convertExpression(expr.right());
+    if (!falseValue)
+      // When `trueValue` is empty (e.g. expression is X or Z for now) , use
+      // builder without value to avoid errors.
+      builder.create<moore::YieldOp>(loc);
+    else
+      builder.create<moore::YieldOp>(loc, convertToSimpleBitVector(falseValue));
+
```
hailongSun2000, line 554: As above.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -958,4 +958,69 @@ def UnionCreateOp : MooreOp<"union_create"> {
...
+  Terminator, 
+  HasParent<"ConditionalOp">
+]> {
+  let summary = "conditional yield and termination operation";
+  let description = [{
+    "moore.yield" yields an SSA value from the Moore dialect op region and
+    terminates the regions. The semantics of how the values are yielded is
+    defined by the parent operation.
+    If "moore.yield" has any operands, the operands must match the parent
+    operation's results.
```
hailongSun2000, line 1014: How do you ensure the types of operands of `moore.yield` match the result types of `moore.conditional`? Maybe you should add the related method to check this.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -581,6 +581,37 @@ module Expressions;
     // CHECK: moore.or [[TMP1]], [[TMP6]] : i1
     c = a inside { a, b, [a:b] };
 
+    //===------------------------------------------------------------------===//
+    // Conditional operator
+
```
hailongSun2000, line 586: nit: It would be better if you could add some non-SBVT tests.

include/circt/Dialect/Moore/MooreOps.td
angelzzzzz: Nice advice! However, Slang already ensures a match between result type and expressions type, so additional checks seem unnecessary.

fabianschuiki: I think you still need a check to make sure that the IR is not malformed. Most of the time the IR will come from import through Slang, but in many other cases the input IR will be handwritten in case of unit tests, or it will be generated from transformation passes. It is important that operations check that the IR is correct to catch bugs in those other cases as soon as possible.

The `hw::OutputOp`, `arc::OutputOp`, `scf::YieldOp`, or `func::ReturnOp` have to do a similar kind of check -- you might find some inspiration on how to do it there. I'm not 100% if the output/yield op verifies that it matches the parent condition/module/arc/func/if op, or if the parent op checks that it has the right terminator.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -514,6 +514,47 @@ struct ExprVisitor {
...
+    // Create blocks for true region and false region.
+    conditionalOp.getTrueRegion().emplaceBlock();
+    conditionalOp.getFalseRegion().emplaceBlock();
+
+    OpBuilder::InsertionGuard g(builder);
+
+    // Handle left expression.
+    builder.setInsertionPointToStart(conditionalOp.getBody(0));
+    auto trueValue = context.convertExpression(expr.left());
+    if (!trueValue)
```
fabianschuiki, line 538: In this case you should just `return {}`, since `convertExpression` returning a null value indicates that an error has happened that has already been reported in a diagnostic by the expression conversion code. Similar to how you further above check `if (!cond) return {};`.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -514,6 +514,47 @@ struct ExprVisitor {
...
+      // When `trueValue` is empty (e.g. expression is X or Z for now) , use
+      // builder without value to avoid errors.
+      builder.create<moore::YieldOp>(loc);
+    else
+      builder.create<moore::YieldOp>(loc, trueValue);
+
+    // Handle right expression.
+    builder.setInsertionPointToStart(conditionalOp.getBody(1));
+    auto falseValue = context.convertExpression(expr.right());
+    if (!falseValue)
```
fabianschuiki, line 548: Same as above: `return {};`

lib/Conversion/ImportVerilog/Expressions.cpp
angelzzzzz: If just `return {}`, it will return an unexpected error that "error: 'moore.conditional' op expects a non-empty block".

fabianschuiki: That is very strange! This means that even though you returned `{}` here, the Slang-to-MLIR conversion continued and at the end the MLIR verifier was called to make sure that the IR is correct. We might be missing additional checks for `if (!<expr>) return {};` elsewhere in the converter.

In theory what should happen is: when you encounter an error and the `convert*` functions start returning `{}`, this abort and return with `{}` should propagate all the way to the top, and in ImportVerilog, conversion should be aborted without the MLIR verifier running at all.

Do you have the input Verilog that causes this issue? I can try to help you find why this happens :smiley:

angelzzzzz: Sure! Here is the input:
```
module top();
  int a,b;
  initial begin
    a = b ? 'x : b;
  end
endmodule
```

fabianschuiki: Great catch! It looks like there was a very subtle bug in `convertStatement` where a failure from `convertExpression` would not be propagated properly. That was also the reason why we were able to check for multiple errors in `errors.sv` in parallel, which was suspicious. (Import should have aborted after first error.)

I have pushed a fix straight to the main branch: https://github.com/llvm/circt/commit/dd68e3c6f4dd1b503d3dc15a68df1d74d0b6d374

angelzzzzz: Nice! Fixing this bug has been of great help to me. 🥰

angelzzzzz:
@fabianschuiki @hailongSun2000 I sincerely appreciate your suggestions! I have added verify to `YieldOp`, and fixed some issues. I'm glad to hear more suggestions. :smile:

fabianschuiki:
This looks very cool! Thanks for adding all the fixes and the new ops :heart_eyes:.

Can you add a test to `test/Dialect/Moore/basic.mlir` for `moore.conditional` and `moore.yield`? That file tests whether the ops in the dialect parse and print properly. So basically you'd add an example of your new ops there and then use `CHECK:` to ensure that the ops get printed out in the exact same way :slightly_smiling_face: 

Similarly, since you've added new error diagnostics to the dialect in `moore.yield`, can you add checks for each of the diagnostics to `test/Dialect/Moore/errors.mlir`?

angelzzzzz:
> This looks very cool! Thanks for adding all the fixes and the new ops 😍.
> 
> Can you add a test to `test/Dialect/Moore/basic.mlir` for `moore.conditional` and `moore.yield`? That file tests whether the ops in the dialect parse and print properly. So basically you'd add an example of your new ops there and then use `CHECK:` to ensure that the ops get printed out in the exact same way 🙂
> 
> Similarly, since you've added new error diagnostics to the dialect in `moore.yield`, can you add checks for each of the diagnostics to `test/Dialect/Moore/errors.mlir`?

That's a good idea! I have added several tests for `moore.conditional` and `moore.yield` and I will add more if we need more tests. :smile:

fabianschuiki (approved):
Cool, looks really nice! Thanks a lot for all your work 🥳

PR #6951, opened 2024-04-25
[Scheduling] computeStartTimesInCycle: clear before recomputation

7FM:
Currently, `computeStartTimesInCycle()` potentially computes wrong start times given the following pseudo code:

```C++
ChainingProblem prob;
scheduleAlgorithm1(prob);
computeStartTimesInCycle(prob);
// do something with the first solution
schedlueAlgorithm2(prob); // Compute a different solution based on the same problem
computeStartTimesInCycle(prob); // <- computes wrong results
```

The issue here is that `computeStartTimesInCycle()` checks whether the ChainingProblem already contains a value for `getStartTimeInCycle` for every predecessor of an operation and if so computes the *startTimeInCycle* for this operation.
Otherwise, the *startTimeInCycle* of the predecessors must be computed first.
After the first call to `computeStartTimesInCycle`, all operations and thereby their predecessors have a *startTimeInCycle*.
Since the *startTimeInCycle*s were previously not cleared in `computeStartTimesInCycle()`, results from the first scheduling solution are potentially used for the second solution and might cause invalid *startTimesInCycle*.

### DIFF
--- a/include/circt/Scheduling/Problems.h
+++ b/include/circt/Scheduling/Problems.h
@@ -373,6 +373,7 @@ public:
   void setStartTimeInCycle(Operation *op, float time) {
     startTimeInCycle[op] = time;
   }
+  void clearStartTimeInCycle() { startTimeInCycle.clear(); }
 
   virtual PropertyStringVector getProperties(Operation *op) override;
   virtual PropertyStringVector getProperties(OperatorType opr) override;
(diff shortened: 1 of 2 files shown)


jopperm (approved):
Thanks, good idea to add this! (Could also be done for other solution properties, when the need arises.)

PR #6952, opened 2024-04-25
Pipeline For Lowering PyTorch Tensor Add To FPGA

jiahanxie353:
## Overview
This patch is the first one of a sequence of future PRs that will try to support a complete pipeline from PyTorch Add operation to FPGA.

The lowering process consists of the following steps:
1. Lowering from a PyTorch source program written in Python to MLIR;
2. Lowering from the emitted MLIR program to CIRCT;
3. Export System Verilog programs and try to simulate on FPGAs.

### Step 1: From PyTorch to MLIR
The source PyTorch program is a simple `torch.add`:
```python
class Model(nn.Module):
    def __init__(self):
        super(Model, self).__init__()

    def forward(self, x, y):
        return x + y
```
This is done through [Allo](https://github.com/cornell-zhang/allo). Using example:
```python
model = Model()
model.eval()
example_inputs = [torch.rand(1, 3, 10, 10), torch.rand(1, 3, 10, 10)]
mlir_mod = allo.frontend.from_pytorch(
    model, example_inputs=example_inputs, verbose=False, enable_tensor=True
)
print(mlir_mod)
```
we can emit MLIR:
```mlir
module {
  func.func @forward(%arg0: tensor<1x3x10x10xf32>, %arg1: tensor<1x3x10x10xf32>) -> tensor<1x3x10x10xf32> attributes {itypes = "__", otypes = "_"} {
    %0 = tensor.empty() : tensor<1x3x10x10xf32>
    %1 = linalg.add {name = "add", op_name = "add_0"} ins(%arg0, %arg1 : tensor<1x3x10x10xf32>, tensor<1x3x10x10xf32>) outs(%0 : tensor<1x3x10x10xf32>) -> tensor<1x3x10x10xf32>
    return %1 : tensor<1x3x10x10xf32>
  }
}
```
### Step 2: Conversions in CIRCT
Before using the conversions in CIRCT, we need to first get rid of some high-level dialect operations, such as `tensor.empty()`, and `linalg.add`. 

To this end, we can run `mlir-opt torch-add.mlir --empty-tensor-to-alloc-tensor --one-shot-bufferize="allow-return-allocs-from-loops bufferize-function-boundaries" --buffer-results-to-out-params --convert-linalg-to-loops --canonicalize` to emit the following code:
```mlir
module {
  func.func @forward(%arg0: memref<1x3x10x10xf32, strided<[?, ?, ?, ?], offset: ?>>, %arg1: memref<1x3x10x10xf32, strided<[?, ?, ?, ?], offset: ?>>, %arg2: memref<1x3x10x10xf32>) attributes {itypes = "__", otypes = "_"} {
    %c10 = arith.constant 10 : index
    %c3 = arith.constant 3 : index
    %c1 = arith.constant 1 : index
    %alloc = memref.alloc() {alignment = 64 : i64} : memref<1x3x10x10xf32>
    scf.for %arg3 = %c0 to %c3 step %c1 {
      scf.for %arg4 = %c0 to %c10 step %c1 {
        scf.for %arg5 = %c0 to %c10 step %c1 {
          %0 = memref.load %arg0[%c0, %arg3, %arg4, %arg5] : memref<1x3x10x10xf32, strided<[?, ?, ?, ?], offset: ?>>
          %1 = memref.load %arg1[%c0, %arg3, %arg4, %arg5] : memref<1x3x10x10xf32, strided<[?, ?, ?, ?], offset: ?>>
          %2 = arith.addf %0, %1 : f32
          memref.store %2, %alloc[%c0, %arg3, %arg4, %arg5] : memref<1x3x10x10xf32>
        }
      }
    }
    memref.copy %alloc, %arg2 : memref<1x3x10x10xf32> to memref<1x3x10x10xf32>
    return
  }
}
```
Now, we can use the tools in CIRCT, which consists of the following steps:
1. Lower SCF to Calyx;
2. Lower Calyx to `hw`, `seq`, `comb`, and `sv` to get ready to generate System Verilog code.

## Patch Specific Works
During lowering SCF to Calyx, we don't have support for `memref::CopyOp`. Thanksfully, `handshake` gives us a pass named `--handshake-legalize-memrefs` to convert `memref::CopyOp` to `memref::LoadOp` and `memref::StoreOp`.

This patch makes a placeholder for `memref::CopyOp` in SCF-to-Calyx pass to remind future users.

### DIFF
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -212,7 +212,7 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
                              scf::YieldOp, scf::WhileOp, scf::ForOp,
                              /// memref
                              memref::AllocOp, memref::AllocaOp, memref::LoadOp,
-                             memref::StoreOp,
+                             memref::StoreOp, memref::CopyOp,
                              /// standard arithmetic
                              AddIOp, SubIOp, CmpIOp, ShLIOp, ShRUIOp, ShRSIOp,
                              AndIOp, XOrIOp, OrIOp, ExtUIOp, ExtSIOp, TruncIOp,
@@ -266,6 +266,7 @@ private:
   LogicalResult buildOp(PatternRewriter &rewriter, memref::AllocaOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, memref::LoadOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, memref::StoreOp op) const;
+  LogicalResult buildOp(PatternRewriter &rewriter, memref::CopyOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, scf::WhileOp whileOp) const;
   LogicalResult buildOp(PatternRewriter &rewriter, scf::ForOp forOp) const;
   LogicalResult buildOp(PatternRewriter &rewriter, CallOp callOp) const;
@@ -505,6 +506,25 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
   getState<ComponentLoweringState>().addBlockScheduleable(loadOp->getBlock(),
                                                           group);
   return success();
+}
+
+LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
+                                     memref::CopyOp copyOp) const {
+  auto source = copyOp.getSource();
+  auto target = copyOp.getTarget();
+
+  auto srcType = source.getType().cast<MemRefType>();
+  auto tgtType = target.getType().cast<MemRefType>();
+
+  assert(srcType.getElementType() == tgtType.getElementType() &&
+         srcType.getShape() == tgtType.getShape() &&
+         "Source and target are expected to have the same element type and "
+         "shape.");
+
+  copyOp.emitError("Please run --handshake-legalize-memrefs to replace "
+                   "memre.copy with memref.load and memref.store");
+
+  return failure();
 }
 
 LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -507,6 +508,25 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+
+  auto srcType = source.getType().cast<MemRefType>();
+  auto tgtType = target.getType().cast<MemRefType>();
+
+  assert(srcType.getElementType() == tgtType.getElementType() &&
+         srcType.getShape() == tgtType.getShape() &&
+         "Source and target are expected to have the same element type and "
+         "shape.");
+
+  copyOp.emitError("Please run --handshake-legalize-memrefs to replace "
```
mikeurbach, line 524: I'm all for good error messages, but it seems a little weird to handle `memref::CopyOp` only to have `buildOp` unconditionally throw an error.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -507,6 +508,25 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
 
+LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
+                                     memref::CopyOp copyOp) const {
+  auto source = copyOp.getSource();
+  auto target = copyOp.getTarget();
+
+  auto srcType = source.getType().cast<MemRefType>();
+  auto tgtType = target.getType().cast<MemRefType>();
+
+  assert(srcType.getElementType() == tgtType.getElementType() &&
```
mikeurbach, line 519: I think these invariants should be enforced upstream by the `memref::CopyOp` verifier, I don't think they need to be asserted here.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -212,7 +212,7 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
                              scf::YieldOp, scf::WhileOp, scf::ForOp,
                              /// memref
                              memref::AllocOp, memref::AllocaOp, memref::LoadOp,
-                             memref::StoreOp,
+                             memref::StoreOp, memref::CopyOp,
```
mikeurbach, line 215: If you don't make this change, does the pass fail with the below error of `"Unhandled operation during BuildOpGroups()"`? I think that is reasonable behavior--the pass doesn't handle this op, and expects an earlier pass in the compiler pipeline to handle it.

mikeurbach: _Maybe_ in the `Default` case, you could add some conditional logic like

```
if (isa<memref::CopyOp>(op))
  error.addNote("Please run --handshake-legalize-memrefs")
```

mikeurbach (commented):
Thanks for all the context in the PR description. Exciting to see more interest in the Calyx HLS flow for ML.

I'm not sure about this specific change though. It seems weird to handle an op just to throw an error asking the user to run a pass. I think the default error message for unhandled ops is reasonable.

It's really up to the compiler developer to set up the correct pass pipeline, and provide a tool like `hlstool` to end users.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
mikeurbach: But even here, I don't like the idea of naming a specific pass--that will inevitably get out of date. And, we should be providing a correct pass pipeline in a tool like `hlstool` regardless.

jiahanxie353: > If you don't make this change, does the pass fail with the below error of "Unhandled operation during BuildOpGroups()"?

Yes, that what will happen.

> But even here, I don't like the idea of naming a specific pass--that will inevitably get out of date. And, we should be providing a correct pass pipeline in a tool like hlstool regardless.

I agree. My thought was just: since we want to lower from Calyx to dialects like `hw`, `seq`,`comb`, and eventually `sv` to have System Verilog, it's unnatural to think of a pass provided by `handshake`. So I wanted to remind the future developers that there's actually an available conversion by emitting this error message.

mikeurbach: I think the current error message is good enough for future compiler developers. In general, error messages are to help end users diagnose problems in their input: https://llvm.org/docs/CodingStandards.html#error-and-warning-messages. I don't think we need a custom error message for the situation where a compiler developer constructed the wrong compiler pipeline.

PR #6953, opened 2024-04-25
SCF To Calyx Support Float Add and Float SeqMemory Read/Write

jiahanxie353:
This is the second patch stacking on #6952 

This patch tries to legalize `AddFOp` and support floating point `SeqMemory` read/write in SCFToCalyx.

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -114,7 +114,7 @@ def MemoryOp : CalyxPrimitive<"memory", []> {
   );
 
   let results = (outs
-    Variadic<AnySignlessInteger>:$results
+    Variadic<AnyTypeOf<[AnySignlessInteger, AnyFloat]>>:$results
   );
 
   let assemblyFormat = [{
@@ -184,7 +184,7 @@ def SeqMemoryOp : CalyxPrimitive<"seq_mem", []> {
   );
 
   let results = (outs
-    Variadic<AnySignlessInteger>:$results
+    Variadic<AnyTypeOf<[AnySignlessInteger, AnyFloat]>>:$results
   );
 
   let assemblyFormat = [{
@@ -197,6 +197,12 @@ def SeqMemoryOp : CalyxPrimitive<"seq_mem", []> {
     OpBuilder<(ins
       "StringRef":$sym_name,
       "int64_t":$width,
+      "ArrayRef<int64_t>":$sizes,
+      "ArrayRef<int64_t>":$addrSizes
+    )>,
+    OpBuilder<(ins 
+      "StringRef":$sym_name,
+      "Type":$type,
       "ArrayRef<int64_t>":$sizes,
       "ArrayRef<int64_t>":$addrSizes
     )>
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -331,6 +331,15 @@ class CombinationalArithBinaryLibraryOp<string mnemonic> :
   let results = (outs AnyType:$left, AnyType:$right, AnyType:$out);
 }
 
+class ArithBinaryFloatingPointLibraryOp<string mnemonic> : ArithBinaryLibraryOp<mnemonic, [
+  SameTypeConstraint<"left", "out">
+  ]> {
+    let results = (outs I1:$clk, I1:$reset, I1:$go, I1:$control, I1:$subOp,
+      AnyFloat:$left, AnyFloat:$right, AnySignlessInteger:$roundingMode, AnyFloat:$out,
+      AnySignlessInteger:$execptionalFlags, I1:$done);
+}
+
+def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {}
 def AddLibOp  : CombinationalArithBinaryLibraryOp<"add"> {}
 def SubLibOp  : CombinationalArithBinaryLibraryOp<"sub"> {}
 def ShruLibOp : CombinationalArithBinaryLibraryOp<"shru"> {}
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -10,12 +10,48 @@
 //
 //===----------------------------------------------------------------------===//
 
+include "mlir/IR/BuiltinAttributeInterfaces.td"
+
 /// Base class for Calyx primitives.
 class CalyxPrimitive<string mnemonic, list<Trait> traits = []> :
   CalyxCell<mnemonic, traits> {
     let assemblyFormat = "$sym_name attr-dict `:` qualified(type(results))";
     let arguments = (ins SymbolNameAttr:$sym_name);
     let skipDefaultBuilders = 1;
+}
+
+def ConstantOp: CalyxOp<"constant",
+    [Pure, ConstantLike, FirstAttrDerivedResultType,
+    DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>,
+    AllTypesMatch<["value", "result"]>
+    ]> {
+  let summary = "integer or floating point constant";
+  let description = [{
+    The `constant` operation produces an SSA value equal to some integer or
+    floating-point constant specified by an attribute.
+
+    Example:
+
+    ```
+    // Integer constant
+    %1 = calyx.constant 42 : i32
+
+    // Equivalent generic form
+    %1 = "calyx.constant"() {value = 42 : i32} : () -> i32
+    ```
+  }];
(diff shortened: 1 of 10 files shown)


mikeurbach:
I'd like to hear more about what the story for floating point types is. There's a reason nothing in CIRCT handles them--hardware IRs work very naturally for bitvectors modeled as (signless) integer types. But floating point is a whole beast with very different hardware circuits. How does Calyx handle floats?

jiahanxie353:
> I'd like to hear more about what the story for floating point types is. There's a reason nothing in CIRCT handles them--hardware IRs work very naturally for bitvectors modeled as (signless) integer types. But floating point is a whole beast with very different hardware circuits. How does Calyx handle floats?

Sure! Calyx plans to [introduce floating point adders and multipliers as primitives using HardFloat](https://github.com/calyxir/calyx/pull/1928) 

> hardware IRs work very naturally for bitvectors modeled as (signless) integer types
Suppose Calyx can do floating point manipulations, then the crux lies in lowering Calyx to HW dialect. Can we convert the floating point number to the corresponding bitvector based on IEEE754 and the problem is then solved? Am I oversimplifying the scenario?

mikeurbach:
Thanks for the pointer. Yes, I think the crux of the problem is how Calyx floating points convert to bitvectors in the HW dialect. I don't think anyone in CIRCT has modeled IEEE754 floats, but if you are fine modeling them as a bitvector of the appropriate width, the core CIRCT dialects should support that.

jiahanxie353:
> Thanks for the pointer. Yes, I think the crux of the problem is how Calyx floating points convert to bitvectors in the HW dialect. I don't think anyone in CIRCT has modeled IEEE754 floats, but if you are fine modeling them as a bitvector of the appropriate width, the core CIRCT dialects should support that.

Sounds good! I'm super excited to try this out!

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -2911,7 +2989,53 @@ LogicalResult SliceLibOp::verify() {
             DictionaryAttr::get(getContext())};                                \
   }
 
+#define ImplBinFloatingPointOpCellInterface(OpType)                            \
```
mikeurbach, line 2992: We generally discourage macros. Could this be templated on `OpType` instead?

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -289,6 +331,15 @@ class CombinationalArithBinaryLibraryOp<string mnemonic> :
...
 
+class ArithBinaryFloatingPointLibraryOp<string mnemonic> : ArithBinaryLibraryOp<mnemonic, [
+  SameTypeConstraint<"left", "out">
+  ]> {
+    let results = (outs I1:$clk, I1:$reset, I1:$go, I1:$control, I1:$subOp,
+      AnyFloat:$left, AnyFloat:$right, AnySignlessInteger:$roundingMode, AnyFloat:$out,
+      AnySignlessInteger:$execptionalFlags, I1:$done);
+}
+
+def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {}
```
mikeurbach, line 342: What is the `addFN` name? "Add floating number" or something like that? I don't usually nitpick on the mnemonics, but it is worth thinking through. I get the impression this aligns with Calyx native IR, so if that's the case, `addFN` sounds good to me.

mikeurbach (commented):
This is looking good.

I would suggest to break it up into smaller changes if possible. It seems there are a few things that can be added in separate PRs:

* Addition of the constant op
* Addition of the floating point add op
* Addition of the memory handling

Also, please include both happy path and failure tests, as applicable.

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -18,6 +20,40 @@ class CalyxPrimitive<string mnemonic, list<Trait> traits = []> :
...
+    floating-point constant specified by an attribute.
+
+    Example:
+
+    ```
+    // Integer constant
+    %1 = calyx.constant 42 : i32
+
+    // Equivalent generic form
+    %1 = "calyx.constant"() {value = 42 : i32} : () -> i32
```
cgyurgyik, line 40: Nit: I don't this is necessary; it is just a detail of LLVM.

lib/Dialect/Calyx/CalyxOps.cpp
cgyurgyik: +1, I see `ImplBinPipeOpCellInterface` is doing something similar and I'm not sure why 
 upon a quick glance. It might be OK to refactor both in a follow-up PR, or open an issue to do so and add a TODO.

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -936,6 +957,33 @@ void Emitter::emitLibraryPrimTypedByFirstOutputPort(
            << LParen() << bitWidth << RParen() << semicolonEndL();
 }
 
+void Emitter::emitLibraryFloatingPoint(Operation *op) {
+  auto cell = cast<CellInterface>(op);
+  unsigned bitWidth =
+      cell.getOutputPorts()[0].getType().getIntOrFloatBitWidth();
+  unsigned expWidth, sigWidth;
+  if (bitWidth == 16) {
```
cgyurgyik, line 965: Shouldn't this be expWidth, sigWidth, and signBit?

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -936,6 +957,33 @@ void Emitter::emitLibraryPrimTypedByFirstOutputPort(
            << LParen() << bitWidth << RParen() << semicolonEndL();
 }
 
+void Emitter::emitLibraryFloatingPoint(Operation *op) {
+  auto cell = cast<CellInterface>(op);
+  unsigned bitWidth =
+      cell.getOutputPorts()[0].getType().getIntOrFloatBitWidth();
```
cgyurgyik, line 963: Can we assert this has length > 0?

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
cgyurgyik, line 965: I know this is the default size for Half, but there are other 16-bit values. Maybe comment saying this is for Half-float?

lib/Dialect/Calyx/Transforms/CalyxLoweringUtils.cpp
```diff
@@ -169,6 +169,13 @@ MemoryInterface::MemoryInterface(const MemoryPortsImpl &ports) : impl(ports) {
 MemoryInterface::MemoryInterface(calyx::MemoryOp memOp) : impl(memOp) {}
 MemoryInterface::MemoryInterface(calyx::SeqMemoryOp memOp) : impl(memOp) {}
 
+bool MemoryInterface::isSeqMem() {
+  if (auto *memOp = std::get_if<calyx::SeqMemoryOp>(&impl); memOp) {
+    return true;
+  }
+  return false;
+}
+
```
cgyurgyik, line 178:
```suggestion
bool MemoryInterface::isSeqMem() {
  return std::holds_alternative<calyx::SeqMemoryOp>(&impl);
}
```

lib/Dialect/Calyx/Transforms/CalyxLoweringUtils.cpp
```diff
@@ -730,7 +737,8 @@ BuildBasicBlockRegs::partiallyLowerFuncToComp(mlir::func::FuncOp funcOp,
 
     for (auto arg : enumerate(block->getArguments())) {
       Type argType = arg.value().getType();
-      assert(isa<IntegerType>(argType) && "unsupported block argument type");
+      assert((isa<IntegerType>(argType) || isa<FloatType>(argType)) &&
```
cgyurgyik, line 740:
```suggestion
      assert(isa<IntegerType, FloatType>(argType) &&
```

(I think this should work)

lib/Dialect/Calyx/Transforms/CalyxLoweringUtils.cpp
```diff
@@ -753,11 +761,11 @@ BuildReturnRegs::partiallyLowerFuncToComp(mlir::func::FuncOp funcOp,
 
   for (auto argType : enumerate(funcOp.getResultTypes())) {
     auto convArgType = calyx::convIndexType(rewriter, argType.value());
-    assert(isa<IntegerType>(convArgType) && "unsupported return type");
-    unsigned width = convArgType.getIntOrFloatBitWidth();
+    assert((isa<IntegerType>(convArgType) || isa<FloatType>(convArgType)) &&
```
cgyurgyik, line 764: (Similar case for here)

cgyurgyik:
(Hi sorry, I'm just seeing this is in draft mode. Feel free to ignore my comments.)

PR #6955, opened 2024-04-26
[FIRRTL] Don't fail compilation when Dedup group annotation is on EICG_wrapper

uenoku:
Currently `DedupGroup` is added to almost every module including `EICG_wrapper`. However EICG_wrapper later becomes intrinsic when a compiler option `fixup-eicg-wrapper` is given. This causes a compilation error since annotation is added to an intrinsic. The right fix is to use to EICG intrinsic in the first place but there are still a few things to migrate ( ExtractInstances).  Since dropping DedupGroup annotation on EICG_wrapper is practically safe this PR removes the annotation with a warning.


dtzSiFive (approved):
LGTM.

uenoku:
Thank you for the review!

PR #6958, opened 2024-04-28
[LLHD] Replace entity with module

maerhart:
* use `!hw.inout` instead of `!llhd.sig` to make things work with `hw.module` (since inout types don't allow nested signals)
* remove the `llhd.entity` operation and use the `hw.module` instead. This has two disadvantages: (1) all ports are now inout instead of separated in and out ports, however this was never verified in LLHD anyways and out ports were always more like inout, (2) instance and signal names cannot be verified to be unique within one module anymore
* the HWToLLHD pass is removed because it only converted `hw.module` to `llhd.entity` and `hw.instance` to `llhd.inst`. We can potentially implement the ModuleLike interface for ProcOp and get rid of `llhd.inst` as well.
* the function inlining pass does not work yet


fabianschuiki:
This is awesome :heart_eyes: !

fabianschuiki (approved):
:partying_face: LGTM!

This makes me think that we could make `llhd.proc` an op that can be placed inline in a `hw.module` in the future, without the need for an `llhd.inst`.

PR #6959, opened 2024-04-28
[Comb] Officialize support for zero-width integers

Moxinilian:
This PR makes the undocumented support of i0 in comb official by removing the documentation stating it is not supported and adding a few tests to the canonicalizer (and marginally to the lowering to SV which has already been testing it for the most part).

[Associated RFC](https://discourse.llvm.org/t/rfc-support-zero-width-integers-in-the-comb-dialect-circt/78492)

### DIFF
--- a/docs/Dialects/Comb/RationaleComb.md
+++ b/docs/Dialects/Comb/RationaleComb.md
@@ -9,7 +9,6 @@ other [MLIR Rationale docs](https://mlir.llvm.org/docs/Rationale/).
 - [`comb` Dialect Rationale](#comb-dialect-rationale)
   - [Introduction to the `comb` Dialect](#introduction-to-the-comb-dialect)
   - [Type System for `comb` Dialect](#type-system-for-comb-dialect)
-    - [Zero-bit integer width is not supported](#zero-bit-integer-width-is-not-supported)
   - [Comb Operations](#comb-operations)
     - [Fully associative operations are variadic](#fully-associative-operations-are-variadic)
     - [Operators carry signs instead of types](#operators-carry-signs-instead-of-types)
@@ -32,30 +31,6 @@ substrate that may be extended with higher level dialects mixed into it.
 
 TODO: Simple integer types, eventually parametrically wide integer type
 `hw.int<width>`.  Supports type aliases.  See HW rationale for more info.
-
-### Zero-bit integer width is not supported
-
-Combinational operations like add and multiply work on values of signless
-standard integer types, e.g. `i42`, but they do not allow zero bit inputs.  This
-design point is motivated by a couple of reasons:
-
-1) The semantics of some operations (e.g. `comb.sext`) do not have an obvious
-   definition with a zero bit input.
-
-1) Zero bit operations are useless for operations that are definable, and their
-   presence makes the compiler more complicated.
-
-On the second point, consider an example like `comb.mux` which could allow zero
-bit inputs and therefore produce zero bit results.  Allowing that as a design
-point would require us to special case this in our cost models, and we would
-have that optimizes it away.
-
-By rejecting zero bit operations, we choose to put the complexity into the
-lowering passes that generate the HW dialect (e.g. LowerToHW from FIRRTL).
-
-Note that this decision only affects the core operations in the `comb` dialect
-itself - it is perfectly reasonable to define your operations and mix them into
-other `comb` constructs. 
 
 ## Comb Operations
 
--- a/test/Conversion/ExportVerilog/pruning.mlir
+++ b/test/Conversion/ExportVerilog/pruning.mlir
@@ -11,6 +11,7 @@ hw.module @zeroWidthPAssign(in %arg0 : i0, in %clk: i1, out out: i0) {
   %1 = sv.read_inout %0 : !hw.inout<i0>
   hw.output %1 : i0
 }
+
 // CHECK-LABEL: module zeroWidthLogic(
 // CHECK-NOT: reg
 hw.module @zeroWidthLogic(in %arg0 : i0, in %sel : i1, in %clk : i1, out out : i0) {
@@ -18,6 +19,14 @@ hw.module @zeroWidthLogic(in %arg0 : i0, in %sel : i1, in %clk : i1, out out : i
   %rr = sv.read_inout %r : !hw.inout<i0>
   %2 = comb.mux %sel, %rr, %arg0 : i0
   hw.output %2 : i0
+}
+
+// CHECK-LABEL: module zeroWidthArith(
+hw.module @zeroWidthArith(in %arg0 : i0, in %arg1 : i0, out out : i0) {
+  %add = comb.add %arg0, %arg1 : i0
+  %sub = comb.sub %arg0, %arg1 : i0
+  %and = comb.and %add, %sub : i0
+  hw.output %and : i0
 }
 
 // CHECK-LABEL: module Concat(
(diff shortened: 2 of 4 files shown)


test/Conversion/ExportVerilog/pruning.mlir
```diff
@@ -20,6 +21,14 @@ hw.module @zeroWidthLogic(in %arg0 : i0, in %sel : i1, in %clk : i1, out out : i
   hw.output %2 : i0
 }
 
+// CHECK-LABEL: module zeroWidthArith(
```
uenoku, line 24: Could you check outputs? Probably nothing is emitted except for a comment?

docs/Dialects/Comb/RationaleComb.md
```diff
@@ -33,30 +32,6 @@ substrate that may be extended with higher level dialects mixed into it.
 TODO: Simple integer types, eventually parametrically wide integer type
 `hw.int<width>`.  Supports type aliases.  See HW rationale for more info.
 
-### Zero-bit integer width is not supported
```
darthscsi, line 36: Given the change, I would make this explicitly positive.  "Zero-bit integer width is supported" with notes on how i0 is handled and how ops which break on 0 are handled.  Although not a comb issuer per-se, noting what is expected in export verilog is helpful too.  Also perhaps notes that zero in types is structurally significant.

Moxinilian:
I also adjusted the verifier for the replicate op so it is more consistent with the mathematical definition of a multiple.

test/Dialect/Comb/errors.mlir
```diff
@@ -11,3 +11,17 @@ hw.module @err(in %a: i1, in %b: i1) {
...
+  // expected-error @+1 {{op replicate must produce integer multiple of operand}}
+  %0 = comb.replicate %a : (i0) -> i16
+}
+
+// -----
+
+hw.module @err(in %a: i0) {
+  // expected-error @+1 {{from bit too large for input}}
+  %0 = comb.extract %a from 0 : (i0) -> i0
+}
```
fabianschuiki, line 27: I'm wondering if this is the behavior you'd expect. Intuitively you can replace `comb.extract %a from 0 : (T) -> T` with just `%a`, since the extract is a no-op. That would suggest you can do the same for `T = i0`, which makes this error somewhat unintuitive. I think `comb.extract` should be valid as you've written it. Maybe we should instead check if `lowBit >= 0` and `lowBit + resultWidth <= inputWidth`, which would allow this case, and also `comb.extract %a from 42 : (i42) -> i0`.

Moxinilian: Do we want to allow `comb.extract %a from 100000 : (i16)-> i0` then? Because conceptually it is the same kind of extraction outside the bounds of the integer. Intuitively it is taking bits where the original value has none, which is only fine if you are taking zero bits.

maerhart: Let me try to approach this from a more mathematical perspective: Let's consider the operation `comb.extract %a from x : (iy) -> iz` where x, y, z are integers >= 0. Let's decompose the input type to the interval of bit indices `I = [0,y)`, and the output type to bit indices `O = [x,x+z)`. Then, the minimal constraint we need for the op to work is `O \subset I` (otherwise, we'd need to define what should be returned for non `i0` OOB accesses). If `z = 0`, then O is the empty set/interval, which is a subset of any other set; thus, it wouldn't matter what x and y are for the operation to be valid (this is what @Moxinilian points out). In the current implementation, we have the additional constraint that `x \element I` which cannot be satisfied if `y = 0` since I is the empty set. @fabianschuiki is suggesting that this constraint should be changed to `x \element (I \union [y, y])`.

I think we shouldn't drop the second constraint entirely because I can't see a case where `comb.extract %a from 100000 : (i16)-> i0` would ever be useful but is more likely a symptom of a bug somewhere that can be caught by the second constraint. In the end, it's a question of definition, and which one we choose is a question of what is more practical to work with in compiler passes. I suppose the extra flexibility @fabianschuiki proposes can be useful when working with module parameters (e.g., if a parameter N is used to extract the N LSB/MSB from an integer).

Moxinilian:
@maerhart @fabianschuiki I have also documented it so people can come and remove the constraint if they can provide a situation where they need it.

test/Dialect/Comb/canonicalization.mlir
```diff
@@ -1662,3 +1662,60 @@ hw.module @cantCombineOppositeNonBinCmpIntoConstant(in %tag_0: i4, in %tag_1: i4
...
+
+// Ensure canonicalization is not confused by i0.
+// CHECK-LABEL: hw.module @i0checks
+// CHECK-SAME: in %[[I32_VAL:[^ ]*]] : i32, in %[[I1_VAL:[^ ]*]] : i1, in %[[I0_VAL:[^ ]*]] : i0
+// CHECK-DAG: %[[FALSE:.*]] = hw.constant false
+// CHECK-DAG: %[[ZERO:.*]] = hw.constant 0 : i0
+// CHECK-DAG: %[[CONCAT:.*]] = comb.concat %[[I0_VAL]], %[[I32_VAL]] : i0, i32
+// CHECK-DAG: %[[CONCAT_REVERSE:.*]] = comb.concat %[[I32_VAL]], %[[I0_VAL]] : i32, i0
+// CHECK-DAG: %[[CONCAT_REMAINS:.*]] = comb.concat %[[I32_VAL]], %[[I0_VAL]], %i1_val : i32, i0, i1
+// CHECK-DAG: %[[CONCAT_ITSELF:.*]] = comb.concat %[[I0_VAL]], %[[ZERO]] : i0, i0
```
maerhart, line 1674: Is there a particular reason the `i0` values are not canonicalized away, or is it just not implemented yet?

Moxinilian: It is not implemented yet. I feel like it would be good to separate fixes from new features.

test/Dialect/Comb/errors.mlir
darthscsi: `comb.extract %a from 100000 : (i16)-> i0` should be illegal.  It is very reasonable that in the IR, indexes are legal.  If a front end language wants to have operations which allow "over-the-limit" extractions (like firrtl does), that's fine, but that complexity shouldn't impact comb.

The alternate argument is that parameters will make this hard to enforce.

darthscsi: Or put alternately, ops should not encode an implicit extension.  So if `comb.extract %a from 100000 : (i16)-> i0` is legal, that implies the op always implicitly extends the argument sufficiently to perform the extraction.  This makes analysis much harder.  Lots of bugs in FIRRTL come from places where people forgot that the ops have implicit extension and we are slowly reworking the dialect to rectify that.

Moxinilian: Extracting an i0 from bit 10000 is not an implicit extension though. The property "all extracted bits are defined" is satisfied even when no extension is carried out. But in any case, the verifier I implemented will not accept it, at least for now.

docs/Dialects/Comb/RationaleComb.md
```diff
@@ -33,29 +32,11 @@ substrate that may be extended with higher level dialects mixed into it.
...
-By rejecting zero bit operations, we choose to put the complexity into the
-lowering passes that generate the HW dialect (e.g. LowerToHW from FIRRTL).
-
-Note that this decision only affects the core operations in the `comb` dialect
-itself - it is perfectly reasonable to define your operations and mix them into
-other `comb` constructs. 
+Operations in the `comb` dialect support zero-width integers.  Any well-defined
+value of type `i0` is assumed to be of zero value, from which sign semantics
+can be infered for signed operations.  In particular, division on `i0` is
+undefined, as it would only yield division by zero.  Similarly, replication of
```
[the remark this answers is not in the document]
Moxinilian: Let me add a link to that, but I'm not sure I understand what additional documentation you would want here?

Moxinilian:
Sorry everyone to have forgotten this PR! Hopefully now comments are addressed, let me know if you approve!

uenoku:
@Moxinilian thank you for merging this!

PR #6960, opened 2024-04-29
Fix invalid rewriter API usages

7FM:
According to https://github.com/llvm/llvm-project/commit/6008cd40b7bbdc66555550c2e38648d5ce99cc78:
> This is invalid rewriter API usage: ops that were replaced/erased should not be accessed

I probably did not catch all invalid rewriter API usages since I have only checked `replaceOpWithNewOp` usages.
In general, one can create similar invalid API usages with `eraseOp` and `replaceOp`.

### DIFF
--- a/lib/Conversion/HWArithToHW/HWArithToHW.cpp
+++ b/lib/Conversion/HWArithToHW/HWArithToHW.cpp
@@ -208,7 +208,9 @@ struct DivOpLowering : public OpConversionPattern<DivOp> {
                       ->getOpResult(0);
 
     // Carry over any attributes from the original div op.
-    divResult.getDefiningOp()->setDialectAttrs(op->getDialectAttrs());
+    auto *divOp = divResult.getDefiningOp();
+    rewriter.modifyOpInPlace(
+        divOp, [&]() { divOp->setDialectAttrs(op->getDialectAttrs()); });
 
     // finally truncate back to the expected result size!
     Value truncateResult = extractBits(rewriter, loc, divResult, /*startBit=*/0,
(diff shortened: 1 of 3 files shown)


lib/Conversion/HWArithToHW/HWArithToHW.cpp
```diff
@@ -302,9 +302,10 @@ struct ICmpOpLowering : public OpConversionPattern<ICmpOp> {
     Value rhsValue = extendTypeWidth(rewriter, loc, adaptor.getRhs(), cmpWidth,
                                      rhsType.isSigned());
 
-    auto newOp = rewriter.replaceOpWithNewOp<comb::ICmpOp>(
-        op, combPred, lhsValue, rhsValue, false);
+    auto newOp = rewriter.create<comb::ICmpOp>(op->getLoc(), combPred, lhsValue,
+                                               rhsValue, false);
     newOp->setDialectAttrs(op->getDialectAttrs());
```
7FM, line 307: Do we have to wrap this with `rewriter.modifyOpInPlace`?

uenoku (approved):
LGTM, thank you for fixing these bugs!

lib/Conversion/HWArithToHW/HWArithToHW.cpp
uenoku: Yes, it would be necessary.

7FM: Done in https://github.com/llvm/circt/pull/6960/commits/097d0c75e586abeb15252e3ca120fdd71f2e1429

mikeurbach:
Thanks @7FM , looks good to me!

PR #405, opened 2021-01-05
[FIRRTL] Add division folds

seldridge:
This adds folding to the FIRRTL dialect division operand. Two folds
are implemented:

  - div(a, a) -> 1
  - div(a, 1) -> a

The former exploits the fact that the FIRRTL specification allows
division by zero to be undefined. This is then used to treat division
by zero as 1 (which is the same as any division by self) and produce a
constant. The latter is straightforward.

Tests and an updated description are included.

See discussion on #363 and chipsalliance/firrtl#2029.

PR #6962, opened 2024-04-29
[NFCI] Document division and the rational for the handling of divide by zero

darthscsi:
Be explicit that divide by zero is undefined and show examples of how different semantics for source or target languages can be achieved.

### DIFF
--- a/docs/Dialects/Comb/RationaleComb.md
+++ b/docs/Dialects/Comb/RationaleComb.md
@@ -212,7 +212,7 @@ between compilations, is opaque, and return any value in the target's type syste
 is guaranteed that repeated evaluation of the same operation with the same operands will
 return the same result.
 
-A dividision by zero, for example, could return any constant, either of its input, 
+A division by zero, for example, could return any constant, either of its input, 
 `x` or `z` (in SV or VHDL), the sum of its input, or the result of any other 
 combinatorial function.
 
--- a/docs/Dialects/Comb/RationaleComb.md
+++ b/docs/Dialects/Comb/RationaleComb.md
@@ -68,13 +68,13 @@ TODO: describe why add/xor/or are variadic
 
 ### Operators carry signs instead of types
 
-Operators, in LLVM-2.0 style, which have consistent behavior in module 
+Operators, in [LLVM-2.0 style](https://mlir.llvm.org/docs/Rationale/Rationale/#integer-signedness-semantics), which have consistent behavior in module 
 arithmetic with respect to signedness are not modeled with sign.  `comb` 
 operates on signless types with signless operations.  This is in accordance
 with LLVM's approach.
 
-Some operations, such as `divu` and `divs`, have different behaviors for signed
-v.s. unsigned types, thus they are modeled with different ops.
+Some operations, such as division, have different behaviors for signed v.s. 
+unsigned types, thus they are modeled with different ops (`divu` and `divs`).
 
 ### Selectable truth-table
 
@@ -192,9 +192,9 @@ semantics required for that language.  For example, system verilog returns an
 `x` on divide by zero, thus its representation may be
 `mux(denominator == 0, sv.constantx, divu(numerator,denominator))` whereas VHDL
 has a runtime-trap in simulation, thus it may require 
-`if(denominator==0) { assert() } else { divu(numerator,denominator)} ` and
-FIRRTL requires division by zero to return 0, so it may require
-`mux(denominator==0, 0, divu(numerator,denominator))`.
+`if(denominator==0) { assert() } else { divu(numerator,denominator)} `.
+Some vendor division blocks produce 0 in this case and
+that could be modeled as `mux(denominator==0, 0, divu(numerator,denominator))`.
 
 Since division in general is very rare in real synthesizable HW, circt doesn't 
 make much effort to optimize divide by zero (nor even division in general, as 
--- a/docs/Dialects/Comb/RationaleComb.md
+++ b/docs/Dialects/Comb/RationaleComb.md
@@ -68,7 +68,13 @@ TODO: describe why add/xor/or are variadic
 
 ### Operators carry signs instead of types
 
-TODO: describe why we have divu/divs but not addu/adds, and not sint vs uint.
+Operators, in LLVM-2.0 style, which have consistent behavior in module 
+arithmetic with respect to signedness are not modeled with sign.  `comb` 
+operates on signless types with signless operations.  This is in accordance
+with LLVM's approach.
+
+Some operations, such as `divu` and `divs`, have different behaviors for signed
+v.s. unsigned types, thus they are modeled with different ops.
 
 ### Selectable truth-table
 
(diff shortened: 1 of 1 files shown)


docs/Dialects/Comb/RationaleComb.md
```diff
@@ -68,7 +68,13 @@ TODO: describe why add/xor/or are variadic
 
 ### Operators carry signs instead of types
 
-TODO: describe why we have divu/divs but not addu/adds, and not sint vs uint.
+Operators, in LLVM-2.0 style, which have consistent behavior in module 
+arithmetic with respect to signedness are not modeled with sign.  `comb` 
+operates on signless types with signless operations.  This is in accordance
+with LLVM's approach.
```
seldridge, line 74: While this is obvious to those who know the history, is there either a short explanation or a link to MLIR's section on signless types (https://mlir.llvm.org/docs/Rationale/Rationale/#integer-signedness-semantics) which explains some of the backstory. I didn't audit this rationale to see if this was already mentioned.

docs/Dialects/Comb/RationaleComb.md
```diff
@@ -68,7 +68,13 @@ TODO: describe why add/xor/or are variadic
...
+
+Some operations, such as `divu` and `divs`, have different behaviors for signed
+v.s. unsigned types, thus they are modeled with different ops.
```
seldridge, line 77: Since these operations only  take signless types, this sentence is incorrect.

docs/Dialects/Comb/RationaleComb.md
```diff
@@ -177,6 +183,24 @@ We agreed that we'd revisit in the future if there were a specific reason to
...
+
+`divu` and `divs` have undefined behavior when the numerator is 0.  It is 
+expected that a frontend will use additional operations to implement the 
+semantics required for that language.  For example, system verilog returns an 
+`x` on divide by zero, thus its representation may be
+`mux(denominator == 0, sv.constantx, divu(numerator,denominator))` whereas VHDL
+has a runtime-trap in simulation, thus it may require 
+`if(denominator==0) { assert() } else { divu(numerator,denominator)} ` and
+FIRRTL requires division by zero to return 0, so it may require
+`mux(denominator==0, 0, divu(numerator,denominator))`.
```
seldridge, line 197: FIRRTL (the spec) doesn't require this. FIRRTL compilers have historically been forced to implement exactly two division optimizations: `div(a, a) -> 1` and `div(a, 1) -> a`. This was due to the SFC implementing these optimizations and then MFC needing this to match during the transition. See discussion: https://github.com/llvm/circt/pull/405

Today, CIRCT will emit a bare divide by zero if you provide it:

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  public module Foo:
    input a: UInt<2>
    output b: UInt<2>

    connect b, div(a, UInt<2>(0))
```

Produces:

``` verilog
// Generated by CIRCT firtool-1.73.0-50-gef689399c
module Foo(
  input  [1:0] a,
  output [1:0] b
);

  assign b = a / 2'h0;
endmodule
```

seldridge (approved):
LGTM

Thanks for keeping the rationales up to date. 💯

docs/Dialects/Comb/RationaleComb.md
```diff
@@ -177,6 +183,24 @@ We agreed that we'd revisit in the future if there were a specific reason to
 add it.  Until then we represent the `array_create`/`array_get` pattern for
 frontends that want to generate this.
 
+
+### Undefined behavior for division
+
+`divu` and `divs` have undefined behavior when the numerator is 0.  It is
```
maerhart, line 189: It should be denominator or divisor here, right?

darthscsi:
> Is there a reason to go straight for undefined behavior on division by zero, instead of just stating that the result is undefined

The main motivation is so that targets which produce runtime-traps in these ops don't need code to be generated to avoid them.  VHDL I believe traps on divide by zero, so to implement comb semantics, you would have to codegenerate control flow (not data flow) which avoided the trap.

darthscsi:
Take a look again.  I did like what I did for FIRRTL and took @fzi-hielscher 's suggestion to reuse that.  The main thing is this precludes simulation-time traps and makes mapping to vhdl potentially more expensive.

docs/Dialects/Comb/RationaleComb.md
```diff
@@ -177,6 +183,39 @@ We agreed that we'd revisit in the future if there were a specific reason to
...
+
+An operation which produces an undefined value (as produced by `divu`, for example) 
+under some conditions is considered to have an instance-specific, static, pure function
+which takes as arguments the operands of the operation and produces a result.  This
+function is potentially unique to each instance of the operation, may be different
+between compilations, is opaque, and return any value in the target's type system.  It
+is guaranteed that repeated evaluation of the same operation with the same operands will
+return the same result.
+
+A dividision by zero, for example, could return any constant, either of its input,
```
fzi-hielscher, line 215: Typo: dividision

fzi-hielscher (approved):
Thank you for the update. I think it makes perfect sense to define division by zero as an arbitrary combinatorial function on the numerator. I guess this also means that division on `i0` would produce a well-defined zero after all, since this would be the only possible function (ping #6959 ).

The situation on the VHDL front is actually a bit more nuanced. Assuming we are using the standard `numeric_std` library, a division by zero should [cause an assertion violation of severity 'error'](https://github.com/ghdl/ghdl/blob/master/libraries/ieee/numeric_std-body.vhdl#L173). Going by the 2008 version of the standard, execution is expected to _continue_ after such a violation, but implementations may still chose to terminate. 
So basically, we have no clue what is actually going to happen and IMHO we are better off not even trying to reason about non-dataflow semantics in comb. Otherwise, we"ll paint ourselves in a corner regarding the legality of transformations that end up changing the sequence of evaluation.  Perhaps the 'sim' dialect could/should handle this?

darthscsi:
> . I guess this also means that division on `i0` would produce a well-defined zero after all, since this would be the only possible function

there are several: F(i0) = x, f(i0) = z, f(i0) = 0

Moxinilian:
> > . I guess this also means that division on `i0` would produce a well-defined zero after all, since this would be the only possible function
> 
> there are several: F(i0) = x, f(i0) = z, f(i0) = 0

I'm really not sure about giving x and z semantics to i0. The x and z abstraction works well to say that a given bit is z or x, not really on a whole value. In particular i0 has no bits, so saying its value is x or z seems a bit weird. What meaning would you give to an i0 in being x when i0 carries no information?

maerhart:
This is also a bit unclear to me. The doc says: "return any value in the target's type system". If we assume that `iN` represents 4-valuedness, then for `i1`, the only valid return values would be `0`, `1`, `x`, and `z`. Combinatorially, for an `N` digit value, we have `4^N` possible return values, i.e., for `i0` it'd be `4^0=1`. On discourse, we defined this singleton to be `0`.

PR #6964, opened 2024-04-29
[HW] Moved and renamed arc/inlineModules to hw/flattenModules

dobios:
This PR moves the `arc::inlineModulesPass` to the HW dialect, all while renaming it to something more suitable for what it actually does, as per @fabianschuiki 's recommendation.

### DIFF
--- a/include/circt/Dialect/HW/Passes.td
+++ b/include/circt/Dialect/HW/Passes.td
@@ -44,6 +44,20 @@ def FlattenIO : Pass<"hw-flatten-io", "mlir::ModuleOp"> {
   ];
 }
 
+def FlattenModules : Pass<"hw-flatten-modules", "mlir::ModuleOp"> {
+  let summary = "Eagerly inline private modules";
+  let description = [{
+    This pass eagerly inlines private HW modules into their instantiation sites.
+    After outlining combinational logic and registers into arcs, module bodies
+    become fairly lightweight. Since arc definitions now fulfill the purpose of
+    code reuse by allowing a single definition to be called multiple times, the
+    module hierarchy degenerates into a purely cosmetic construct. At that point
+    it is beneficial to fully flatten the module hierarchy to simplify further
+    analysis and optimization of state transfer arcs.
+  }];
+  let constructor = "circt::hw::createFlattenModulesPass()";
+}
+
 def HWSpecialize : Pass<"hw-specialize", "mlir::ModuleOp"> {
   let summary = "Specializes instances of parametric hw.modules";
   let constructor = "circt::hw::createHWSpecializePass()";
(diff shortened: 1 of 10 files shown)


tools/arcilator/arcilator.cpp
```diff
@@ -66,6 +67,7 @@
 
 using namespace mlir;
 using namespace circt;
+using namespace hw;
```
maerhart, line 70: Unused?

include/circt/Dialect/HW/Passes.td
```diff
@@ -44,6 +44,20 @@ def FlattenIO : Pass<"hw-flatten-io", "mlir::ModuleOp"> {
...
+def FlattenModules : Pass<"hw-flatten-modules", "mlir::ModuleOp"> {
+  let summary = "Eagerly inline private modules";
+  let description = [{
+    This pass eagerly inlines private HW modules into their instantiation sites.
+    After outlining combinational logic and registers into arcs, module bodies
+    become fairly lightweight. Since arc definitions now fulfill the purpose of
+    code reuse by allowing a single definition to be called multiple times, the
+    module hierarchy degenerates into a purely cosmetic construct. At that point
+    it is beneficial to fully flatten the module hierarchy to simplify further
+    analysis and optimization of state transfer arcs.
```
uenoku, line 56: Could you rephrase these sentences a bit since it's HW pass now?

uenoku (approved):
Refactoring looks great to me! InlineModules pass currently doesn't handle inner symbols so can we open a tracking issue?  

```bash
$ cat test.mlir
  hw.hierpath @hier [@Foo::@bar, @Bar::@a]
  hw.module private @Bar(in %a : i1 {hw.exportPort = #hw<innerSym@a>}, out b : i1) {
    hw.output %a : i1
  }
  hw.module @Foo(in %a : i1, out b : i1) {
    %bar.b = hw.instance "bar" sym @bar @Bar(a: %a: i1) -> (b: i1)
    hw.output %bar.b : i1
  }

$ circt-opt test.mlir -hw-flatten-modules -hw-verify-irn
test.mlir:1:3: error: 'hw.hierpath' op  module: "Foo" does not contain any instance with symbol: "bar"
  hw.hierpath @hier [@Foo::@bar, @Bar::@a]  
  ^
test.mlir:1:3: note: see current operation: "hw.hierpath"() {namepath = [#hw.innerNameRef<@Foo::@bar>, #hw.innerNameRef<@Bar::@a>], sym_name = "hier"} : () -> ()

include/circt/Dialect/HW/Passes.td
dobios: yeah ofc!

maerhart (approved):

PR #6965, opened 2024-04-29
[HW to BTOR2] Add support for encoded assertions

dobios:
Some assertions lowered directly from Chisel are encoded using `sv.error`, `sv.fatal` or `sv.exit` for historical reasons. These particular encodings are not currently supported by the btor2 backend. 

This PR adds support for those weirdly encoded assertions and ignores a lot of the sv macros generated by direct lowerings from FIRRTL because they do not impact the resulting design.

### DIFF
--- a/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
+++ b/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
@@ -52,7 +52,7 @@ private:
   raw_ostream &os;
 
   // Create a counter that attributes a unique id to each generated btor2 line
-  size_t lid = 1;          // btor2 line identifiers usually start at 1
+  size_t lid = 1; // btor2 line identifiers usually start at 1
   size_t nclocks = 0;
 
   // Create maps to keep track of lid associations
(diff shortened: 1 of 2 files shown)


lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -770,13 +773,63 @@ struct ConvertHWToBTOR2Pass
...
+    if (auto ifop = dyn_cast<sv::IfOp>(((Operation *)op)->getParentOp())) {
+      Value cond = ifop.getOperand();
+
+      // Check that the operand is a condition and not an sv macro, otherwise
+      // look for another parent. Some functions like printf will be converted
+      // using additional macros as first level conditions to allow for more
+      // error-reporting control later down the road. To do this we simply make
+      // sure that it's not part of the sv dialect, given that at this point a
+      // condition should not be expressed using operations from that dialect.
+      if (cond.getDefiningOp()->getDialect() == op->getDialect()) {
```
uenoku, line 802: It crashes when cond is a BlockArgument.  Also even when the dialect is same(SV), we still need to check the op is MacroRef.

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -770,13 +773,63 @@ struct ConvertHWToBTOR2Pass
...
+  // Assertions are sometimes encoded using a combination of fatal and error, we
+  // thus need to find the parenting condition and generate a bad statement
+  // using that.
+  void visitEncodedAssert(Operation *op) {
+    genSort("bitvec", 1);
+
+    // Fatal simply fails, so we need to find in which case we are failing
+    // For some reason Chisel assertions are sometimes encoded using a fatal and
+    // an error.
+    if (auto ifop = dyn_cast<sv::IfOp>(((Operation *)op)->getParentOp())) {
```
uenoku, line 793: Strictly speaking I guess we need to check that its parent always statement has a same clock seen in the walk?

uenoku (commented):
Personally I feel it's better not to implement a pattern matching to recognize if-else-fatal style assertion since there are a fair amount of corner cases we need to consider. Instead of pattern matching SV, there is a compile option to convert all assertions to SVA (`--verification-flavor=sva`) so I think it's reasonable restriction to require users to use SVA option for btor emission .

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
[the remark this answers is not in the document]
dobios: My IDE was giving me issues without it, but you're right this shouldn't be necessary.

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
dobios: Is this really the only situation we can run into ? Can't it also be defined using other sv constructs when lowering from Chisel?

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
dobios: There is an assumption of a single clock for now, that's specified in the clock lowering.

dobios:
Generally speaking, I agree with you @uenoku . However this workaround is currently needed to get this to work directly out of chisel without any additional cleaning up of the MLIR. Would you suggest I just integrate the `--verification-flavor=sva` in the btor2 pipeline in  firtool instead? In general this encoding made sense to me because we're simply looking for states that make the design crash with BMC, so an `sv.fatal` does just that no? Let me know what you think would be the best solution to handle the historical lowering of assertions from FIRRTL.

PR #6966, opened 2024-04-30
[ExportVerilog] Ensure DivS/ModS are signed regardless of context.

### LINKED ISSUE #6961: [FIRRTL][ExportVerilog] Signed Division Bug
If an expression is inlined into another expression which contains a FIRRTL signed division, then this can cause the entire division to be treated as unsigned.

Consider the following:

``` firrtl
FIRRTL version 3.3.0
circuit Test2 :
  module Test2 :
    input a: UInt<3>
    input b: UInt<3>
    input c: UInt<3>
    output d: UInt<5>

    connect d, add(c, asUInt(div(asSInt(a), asSInt(b))))
```

This produces the following Verilog:

``` verilog
// Generated by CIRCT firtool-1.73.0-7-g21b9395e2
module Test2(
  input  [2:0] a,
               b,
               c,
  output [4:0] d
);

  assign d = {2'h0, c} + {1'h0, $signed({a[2], a}) / $signed({b[2], b})};
endmodule
```

In the above the signed division is not treated as signed due to context sensitivity.

This is reduced from the original report of this in upstream Chisel [[1]] and discussed on Verilator [[2]]. I haven't validated the reduced example for correctness, though! This should be tested on EDA playground.

[1]: https://github.com/chipsalliance/chisel/issues/3740
[2]: https://github.com/verilator/verilator/issues/4814

dtzSiFive:
Fixes #6961.

### DIFF
--- a/test/Conversion/ExportVerilog/hw-dialect.mlir
+++ b/test/Conversion/ExportVerilog/hw-dialect.mlir
@@ -171,9 +171,9 @@ hw.module @TESTSIMPLE(in %a: i4, in %b: i4, in %c: i2, in %cond: i1,
 // CHECK-NEXT:      assign r2 = a - b;
 // CHECK-NEXT:      assign r4 = a * b;
 // CHECK-NEXT:      assign r6 = a / b;
-// CHECK-NEXT:      assign r7 = $signed(a) / $signed(b);
+// CHECK-NEXT:      assign r7 = $signed($signed(a) / $signed(b));
 // CHECK-NEXT:      assign r8 = a % b;
-// CHECK-NEXT:      assign r9 = $signed(a) % $signed(b);
+// CHECK-NEXT:      assign r9 = $signed($signed(a) % $signed(b));
 // CHECK-NEXT:      assign r10 = a << b;
 // CHECK-NEXT:      assign r11 = a >> b;
 // CHECK-NEXT:      assign r12 = $signed($signed(a) >>> b);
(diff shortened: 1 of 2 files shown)


dtzSiFive:
Have not tested this in EDA playground, and would be good to confirm w/ModS (and put under test).

dtzSiFive:
Poked a bit, this maybe shows this: https://www.edaplayground.com/x/CEdZ maybe -- dropping the `{1'h0, ....}` part else they're the same.

test/Conversion/ExportVerilog/hw-dialect.mlir
```diff
@@ -457,7 +458,7 @@ hw.module @signs(in %in1: i4, in %in2: i4, in %in3: i4, in %in4: i4)  {
   sv.assign %awire, %b4: i4
 
   // https://github.com/llvm/circt/issues/369
-  // CHECK: assign awire = 4'sh5 / -4'sh3;
+  // CHECK: assign awire = $signed(4'sh5 / -4'sh3);
```
seldridge, line 461: 🥲 It's unfortunate that this shows up. However, to know that this can be elided would mean teaching `ExportVerilog` about context sensitive widths. This would be a huge pain and it's better to just do the safe, cheap thing. (If non-obvious, this is commentary and non-blocking.)

seldridge (approved):
This looks like the right fix. 👍

seldridge:
Agree about a `ModS` test. This would be good to have under an integration test. That can be handled on the Chisel side.

uenoku (approved):
LGTM, thank you for fixing this. Glad that the simple change fixes the issue :)

dtzSiFive:
Mod acting similarly: https://www.edaplayground.com/x/Qzbs . Integration test Chisel-side works for me.

Change to output for ModS is included in test.

PR #6971, opened 2024-04-30
[FIRRTL] Output directory control for layers and modules

rwy7:
Add output directory control for layers and firrtl.

For specifying the output directories:
- Add a new annotation called "circt.OutputDirAnnotation", for specifying the output directory of modules
- For layers, add syntax in the firrtl surface language for optionally specifying the output directory. We can't use an annotation here because, annotations can't target layers.

In the Lower-Layers Pass:
- Place layer collateral (the bindfile, any layerblocks->modules) under the output directory of the layer, if there is any.
- Stop outputting layer collateral under the testbench/views 

Add a new pass, AssignOutputDirs, which will sink modules into the output directories they are instantiated from. This pass runs after lower-layers. In conjunction with the changes to the lower-layers pass, this means that modules which are only used under a particular layer will be sunk into that layer's output directory.


uenoku:
What is an expected behavior when modules with different output dir annotations are deduplicated?

rwy7:
>  What is an expected behavior when modules with different output dir annotations are deduplicated?

Hey, thanks for taking a look! Two modules will dedup iff they have the same output directory annotation (modulo dirname canonicalization when constructing the the hw::OutputFileAttr). I've added a test for this [here](https://github.com/llvm/circt/pull/6971/files#diff-b1b7ec9d98af2915b95c892b26303a825f48e9c1478ffdcf1d3e95d88d2714b1).

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -155,33 +168,37 @@ class LowerLayersPass : public LowerLayersBase<LowerLayersPass> {
...
+  SymbolTable *symbolTable;
+
   /// A map of layer blocks to module name that should be created for it.
   DenseMap<LayerBlockOp, StringRef> moduleNames;
 
-  /// The design-under-test (DUT) as indicated by the presence of a
-  /// "sifive.enterprise.firrtl.MarkDUTAnnotation".  This will be null if no
-  /// annotation is present.
-  FModuleOp dut;
+  DenseMap<SymbolRefAttr, hw::OutputFileAttr> layerOutputDirTable;
```
seldridge, line 176: Please add some documentation on what the new members are. 😉

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -155,33 +168,37 @@ class LowerLayersPass : public LowerLayersBase<LowerLayersPass> {
...
 
-  /// The directory for verification collateral as indicated by the presence of
-  /// a "sifive.enterprise.firrtl.TestBenchDirAnnotation".  Empty if not
-  /// specified.
-  StringRef testBenchDir;
+  hw::OutputFileAttr getLayerOutputDir(SymbolRefAttr ref) {
+    return layerOutputDirTable[ref];
+  }
```
seldridge, line 180: This is a private member function wrapping access to a private member. This may be cleaner as just the latter.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -604,26 +621,73 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+  for (auto &op : *circuit.getBodyBlock()) {
+    if (auto module = dyn_cast<FModuleOp>(op)) {
+      module->walk([&](LayerBlockOp layerBlock) {
+        auto name = moduleNameForLayer(module.getModuleName(),
+                                       layerBlock.getLayerName());
+        moduleNames.insert({layerBlock, ns.newName(name)});
+      });
+    }
+  }
+}
```
seldridge, line 649: This function appears unused.

test/Dialect/FIRRTL/parse-basic.fir
```diff
@@ -1525,6 +1525,10 @@ circuit Layers:
     ; CHECK-NEXT:   }
     ; CHECK-NEXT: }
 
+  ; CHECK: firrtl.layer @WithDirectory1 bind attributes {output_file = #hw.output_file<"foo/">}
+  layer WithDirectory1, bind, "foo":
+  ; CHECK: firrtl.layer @WithDirectory2 bind attributes {output_file = #hw.output_file<"foo/">}
+  layer WithDirectory2, bind, "foo/":
```
seldridge, line 1531: It just occurred to me that the directory is really only there for the "bind" convention. This may be cleaner as something like the following:

```
layer WithDirectory1, bind<"foo">:
layer WithDirectory1, bind("foo"):
layer WithDirectory1, bind{"foo"}:
```

This is related to this _not_ making sense for the inline convention.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -604,26 +621,73 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+        stack.pop_back();
+        path.pop_back();
+      }
+
+      if (stack.empty())
+        break;
+
+      assert(idx(stack.back()) != end(stack.back()));
+      layer = *idx(stack.back());
+      ++idx(stack.back());
```
seldridge, line 688: What is this logic doing?

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -748,27 +806,43 @@ void LowerLayersPass::runOnOperation() {
...
+    if (auto outputFile =
+            layerOp->getAttrOfType<hw::OutputFileAttr>("output_file")) {
+      auto dir = outputFile.getDirectoryAttr().getValue();
+      bindFile = hw::OutputFileAttr::getFromDirectoryAndFilename(
+          &getContext(), dir, prefix + ".sv",
+          /*excludeFromFileList=*/true);
+    } else {
+      bindFile = hw::OutputFileAttr::getFromFilename(
+          &getContext(), prefix + ".sv", /*excludeFromFileList=*/true);
+    }
```
seldridge, line 819: Having an empty `auto dir` might make this cleaner as both code paths can use `getFromDirectoryAndFilename`.

test/Dialect/FIRRTL/annotations.mlir
```diff
@@ -1905,3 +1905,19 @@ firrtl.circuit "Top" attributes {
...
 }
+
+// -----
+
+// CHECK-LABEL: firrtl.circuit "Top"
+firrtl.circuit "Top" attributes {
+  rawAnnotations = [
+    {
+      class = "circt.OutputDirAnnotation",
+      dirname = "foobarbaz",
```
seldridge, line 1916: Two more cases may be interesting to test:

1. `foobarbaz/`
2. `foobarbaz/qux`

test/Dialect/FIRRTL/assign-output-dirs.mlir
```diff
@@ -0,0 +1,89 @@
...
+      //    A       B
+      //  C   D
+      annotations = [
+        {class = "circt.DeclareOutputDirAnnotation", name = "B", parent = ""},
+        {class = "circt.DeclareOutputDirAnnotation", name = "C", parent = "A"},
+        {class = "circt.DeclareOutputDirAnnotation", name = "D", parent = "A"}
+      ]
+  } {
+
+  firrtl.module private @AssignOutputDirs() {}
```
seldridge, line 16: I'm surprised CIRCT accepts this as I thought the main module had to be public.

test/Dialect/FIRRTL/assign-output-dirs.mlir
```diff
@@ -0,0 +1,89 @@
...
+
+// when a module doesnt have an output directory, it's dragged
+
+// when a module is used in both a child and parent, it is dragged
+
+// when a module is used by two children, it is dragged to parent.
+
+// circular parent/child relationship
+// parent is null / root
+// child name is empty???
```
seldridge, line 89: Nit: strip comments and add file trailing newline.

test/firtool/dedup-modules-with-output-dirs.fir
```diff
@@ -0,0 +1,179 @@
+; RUN: firtool --split-input-file %s | FileCheck %s
```
seldridge, line 1: This test would be tighter as MLIR and running only on dedup. This allows for violations of initialization checking, i.e., the test can have modules that are basically empty. Alternatively, `firtool Foo.fir -parse-only | circt-opt -firrtl-dedup` gives you FIRRTL syntax for the test.

seldridge (commented):
The LCA algorithm comes across as more complicated than I think it should be. It may be easier to adopt a naive prefix LCA (as that is very understandable) or to go the direction of naive Euler Tour + RMQ (which should be recognizable and has room for performance improvements later if needed). This does appear to be doing _a lot of LCA queries_ hence the RMQ formulation may make sense

Comments throughout.

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
```diff
@@ -387,6 +387,42 @@ static LogicalResult applyLoadMemoryAnno(const AnnoPathValue &target,
...
+  if (dirname.empty())
+    return error() << "dirname must not be empty";
+
+  auto outputFile =
+      hw::OutputFileAttr::getAsDirectory(context, dirname.getValue());
+
+  if (auto moduleOp = dyn_cast<FModuleOp>(op)) {
+    moduleOp->setAttr("output_file", outputFile);
+    return success();
+  }
```
uenoku, line 421: This silently drops attributes if (mistakenly) multiple output dir annotations are attached to the module. Can we check and emit an error?

test/firtool/dedup-modules-with-output-dirs.fir
dtzSiFive: I think the idea is this is testing that firtool -- whatever its pipeline is -- has the expected (or at least consistent, as it's under test) behavior from a user's experience regarding interaction of dedup + output_file (in terms exposed to user, not when/how/where we set what attributes internally).

For example, as-is, moving the output-dir-assignment pass before dedup -- while not changing dedup's behavior -- would cause test failures (that a test narrowed to `dedup` would not catch).  Specifically regarding the last test in this file (and lower-layers.fir test too, apparently).

And is this indeed the intended behavior / semantics for "output dir" information end-to-end?

Would be good to pin down what is really expected/promised by using these knobs/annotations -- to inform the behavior of dedup or anything else that alters instance hierarchy in a way that changes what-goes-where (inlining, for example -- what if you inline a module with an output_file attribute on it (especially modules that it was the only parent of)?).

How'd y'all frame the intent here -- best-effort output directives, providing users ability to get their desired output even if not specifically promised in a particular way re:pipeline happenstance (but they could use more annotations / etc/ to change what happens if they don't like it?).

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
```diff
@@ -1455,6 +1455,62 @@ Example:
...
+Example:
+```json
+{
+  "class": "circt.OutputDirAnnotation",
+  "dirname": "verification",
+  "target": "~Foo|Bar"
+}
+```
+
+### DeclareOutputDirAnnotation
```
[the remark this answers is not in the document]
rwy7: I really don't like the name! Please, suggest something :) 

`OutputDirPrecedenceAnnotation`?

Although just as a note, you actually can declare a directory who's parent is "no output directory", by using an empty string as the parent name. It's just that there is no need.

include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -911,4 +911,24 @@ def SpecializeOption :
...
+    possible. For example, if a module is only instantiated under the
+    verification layer, then it might be assigned to the verification output
+    directory. If a module is instantiated under multiple directories, then it
+    is assigned to the directory that is the "least common ancestor" of all
+    directories it is instantiated under.
+
+    The least common ancestor is chosen according to a precedence graph provided
+    by the user through a circuit level annotation called
+    `DeclareOutputDirAnnotation`. The root design directory is implicitly the
+    greatest ancestor of all output directories.
```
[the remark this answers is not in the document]
rwy7: Yeah I may be making up terminology here 😆 , I'll nix it.

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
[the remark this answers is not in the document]
dtzSiFive: Interpreting empty string as default output directory (and a valid entry) is unexpected to me (if this is important, consider mentioning it? Maybe omit the parent entry instead? 🤷 )! But that makes sense.

Names ARE hard, haha, I'll circle back :+1:.

Your suggestion is an improvement IMO 😄 .

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
rwy7: If dir is empty, I believe getFromDirectoryAndFilename will create a path "/<filename>", (a file in the root), rather than a relative directory name, which is what we want in this case.

test/Dialect/FIRRTL/parse-basic.fir
rwy7: I see what you're saying, I can do this if you'd like.

My plan was to "not parse" or "not accept" the output directory after parsing an inline convention.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
rwy7: The logic here is meant to update the current layer, move the iterator forwards, and restart the loop. If I did it right...

rwy7: The logic here is meant to update the current layer, move the iterator forwards, and restart the loop. If I did it right...

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+namespace {
+
+struct OutputDirInfo {
+  OutputDirInfo(StringAttr name, size_t depth = SIZE_MAX,
+                size_t parent = SIZE_MAX)
+      : name(name), depth(depth), parent(parent) {}
+  StringAttr name;
+  size_t depth;
+  size_t parent;
```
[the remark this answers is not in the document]
rwy7: I'm using the constructor for emplacing into a vector, and the StringAttr is used for fast equality :+1:

seldridge (approved):
I think this looks fine.

I do realize that we made a mistake in the representation of the output directory structure and that this should be both: (1) in the FIRRTL spec and (2) sunk into the IR directly. This avoids the entirety of the `OutputDirInfo` which is having to be created to track information that would be better represented in a symbol table.

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
```diff
@@ -1458,6 +1458,66 @@ Example:
...
+| class      | string  | `circt.OutputDirPrecedenceAnnotation`   |
+| name       | string  | The output directory                    |
+| parent     | string  | The parent output directory             |
+
+Specify the "parent" of an output directory.
+
+When Verilog is output, some modules will have user-specified output
+directories. When a module `M` is only instantiated by modules which have a
+common output directory `D`, we can also output module `M` in that same
+directory `D`.
```
darthscsi, line 1494: Why are we not just putting it where the OutputDirAnnotation says?

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
```diff
@@ -1458,6 +1458,66 @@ Example:
...
+
+If the module M is instantiated under a set of directories `DS`, then the module
+`M` is placed in the output directory that is the least-common-ancestor (LCA) of
+the directories `DS`. The LCA is identified according to this output directory
+declaration annotation.
```
darthscsi, line 1499: This seems to be trying to tie directory output as specified to the use to instantiation structure.  Why?  If the user is specifying the directory, why not just use it?

Second, this is implicitly assuming that the compiler is responsible for the DUT/TB distinction.  Why are we interested in preserving this legacy SFC feature?

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,263 @@
...
+      }
+      current->depth = parent->depth + 1;
+      if (stack.empty())
+        break;
+      i = stack.back();
+      stack.pop_back();
+    }
+  }
+  return success();
+}
```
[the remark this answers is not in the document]
darthscsi: Needing symbols for this really seems like it's jumping the shark.

darthscsi (commented):
This seems like we are adding lots of complexity and I'm not sure we really need it.
These seem obvious and would be a good stand-alone PR:
a) all relative paths should be relative to the output directory (-o flag)
b) Modules without outputs set go to the output directory
c) Layers without outputs set go to the output directory
d) Layers extracted to a module inherit the output directory of the layer.
e) Anything for which the user explicitly specifies a directory has that directory respected.

The key 2 question are:
a) whether modules are placed in directories based on the elaboration (instantiation) tree.  If so, it seems the Least common ancestor is the natural place for that.
b) were public modules go.  Do they move based on instantiation graph or always go in the output dir.  What if they specify a location?

For a, saying LCA seems fine.
For b, either seems fine.
For both, I think anything which explicitly specifies a directory must have that respected.

What I think should be a non-goal is making directory nesting encode layer dependency despite what the user specifies for output directory of layers.

seldridge:
☝️ After discussing with @darthscsi more about this. This comment is saying that we can remove the `OutputDirPrecedenceAnnotation` and any logic associated with it and instead compute the LCA based on the LCA of the directories themselves. E.g., `SimpleLCA("verilog/design/", "verilog/testbench/") == "verilog/"`. This cannot express certain things like we do internally where `AnnoyingLCA("verilog/design/", "verilog/testbench/") == "verilog/design/")`. However, we can change this internally so that we can use the `SimpleLCA` computation.

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
```diff
@@ -1458,6 +1458,66 @@ Example:
...
+The intuition behind this annotation is that modules output in the declared
+directory should only depend on modules which are output in an ancestor
+directory. Then, the LCA can be thought of as the "most specific" output
+directory that still makes a module available at all its instantiation
+sites.
+
+When an output directory isn't explicitly declared, then its parent directory is
+implicitly the default output directory. To explicitly declare that a
+directory's parent is the default output directory, use an empty string as the
+parent.
```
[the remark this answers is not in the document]
rwy7: I'm not 100% sure I am interpreting this comment right

> Is there any problem that this allows the description of arbitrary directed, cyclic graphs as opposed to only trees?

Regarding cycles, and multiple parents, there are checks in the AssignOutputDirs pass that ensure the precedence "graph/tree" is acyclic and that every directory has a single parent declaration.

Regarding multiple roots in the precedence graph/tree, that's not possible either, the graph/tree has a single root, the "default output directory".

Are you suggesting I rework the precedence annotations, so that it can only express a tree by construction? I am open to suggestions, but I chose this representation so that there wouldn't have to be a single monolithic annotation declaring all output directories in a single shot.

I will s/graph/tree/ in the PR for clarity.

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,263 @@
...
+  while (a.depth > b.depth)
+    a = infoTable[a.parent];
+  while (b.depth > a.depth)
+    b = infoTable[b.parent];
+  while (a.name != b.name) {
+    a = infoTable[a.parent];
+    b = infoTable[b.parent];
+  }
+  return a.name;
+}
```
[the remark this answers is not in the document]
rwy7: We report an error while constructing the table, if a directory has multiple parents, or if there is a cycle. Is there any other way that there might be multiple LCAs?

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,263 @@
...
+  if (failed(outDirTable.initialize(circuit)))
+    return signalPassFailure();
+
+  DenseSet<InstanceGraphNode *> visited;
+  for (auto *root : getAnalysis<InstanceGraph>()) {
+    for (auto *node : llvm::inverse_post_order_ext(root, visited)) {
+      auto module = dyn_cast<FModuleOp>(node->getModule().getOperation());
+      if (!module || module->getAttrOfType<hw::OutputFileAttr>("output_file") ||
+          module.isPublic())
+        continue;
```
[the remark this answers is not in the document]
rwy7: I don't have a strong opinion here. My thought was, a public module should go "where the user said". If the user doesn't explicitly specify an output directory, then a public module should go into the "default output directory".

An an example, this check would prevent a DUT from getting pulled into the testbench directory.

I can change this behaviour, if wanted.

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
rwy7: If there is an OutputDirAnnotation, we will use that :+1:. We only assign output directories for modules that are:
- private
- don't already have user-specified output directories

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+using hw::OutputFileAttr;
+
+static StringRef lca(StringRef a, OutputFileAttr file) {
+  if (!file)
+    return StringRef();
+
+  auto b = file.getDirectory();
+  if (llvm::sys::path::is_absolute(b))
+    return StringRef();
```
seldridge, line 34: So, the logic here will write to the default directory if there is a parent with an absolute path?

This seems wrong. I think this whole thing wants to just be a greatest common prefix, even if that means that it writes outside the output directory.

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+  for (auto i = path::begin(b), e = path::end(b); i != e; ++i)
+    if (*i == "..")
+      return StringRef();
```
seldridge, line 38: Is this possible to hit given the canonicalization of the output file attributes?

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,108 @@
...
+  for (; i < e; ++i)
+    if (a[i] != b[i])
+      break;
+
+  auto dir = a.substr(0, i);
+  if (dir.ends_with(llvm::sys::path::get_separator()))
+    return dir;
+
+  return llvm::sys::path::parent_path(dir);
+}
```
seldridge, line 51: There are a bunch of logic there that I don't think it hit in the tests. E.g., `..` and absolute path logic. I don't think we actually want this. However, everything here needs to be tested.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: This looks like overkill. Can't this just lookup the layer by symbol and see if it has an output file attribute when it needs it? Alternatively, this could be cached, though isn't that just a walk over `LayerOp`?

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,145 @@
...
+  void runOnOperation() override;
+};
+} // namespace
+
+void AssignOutputDirsPass::runOnOperation() {
+  SmallString<64> outputDir(outputDirOption);
+  if (fs::make_absolute(outputDir)) {
+    signalPassFailure();
+    return;
+  }
```
seldridge, line 88: Some `llvm::errs` here would be useful. I'm not exactly sure what kind of error can show up here, though.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -119,10 +119,29 @@ static SmallString<32> fileNameForLayer(StringRef circuitName,
...
+      return nullptr;
+    return layer->getAttrOfType<hw::OutputFileAttr>("output_file");
+  }
+
+  hw::OutputFileAttr outputFileForLayer(StringRef circuitName,
+                                        SymbolRefAttr layerName) {
+    if (auto file = getOutputFile(layerName))
+      return hw::OutputFileAttr::getFromDirectoryAndFilename(
+          &getContext(), file.getDirectory(),
+          fileNameForLayer(circuitName, layerName));
```
seldridge, line 135: Is this missing the `excludeFromFileList` boolean option?

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -155,33 +170,32 @@ class LowerLayersPass : public LowerLayersBase<LowerLayersPass> {
...
-  /// annotation is present.
-  FModuleOp dut;
-
-  /// The directory for verification collateral as indicated by the presence of
-  /// a "sifive.enterprise.firrtl.TestBenchDirAnnotation".  Empty if not
-  /// specified.
-  StringRef testBenchDir;
 };
 
+static Operation *setOutputFile(Operation *op, hw::OutputFileAttr file);
```
seldridge, line 179: This appears to never have its result used. Consider changing to `void` return.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -605,26 +619,15 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
-/// non-empty test bench directory.
-void LowerLayersPass::maybeSetOutputDir(Operation *op) {
-  if (!testBenchDir.empty() && dut)
-    op->setAttr("output_file", hw::OutputFileAttr::getAsDirectory(
-                                   op->getContext(), testBenchDir,
-                                   /*excludeFromFileList=*/true));
+static Operation *setOutputFile(Operation *op, hw::OutputFileAttr file) {
+  op->setAttr("output_file", file);
+  return op;
 }
```
seldridge, line 625: In addition to the unused return, this is almost so simple now that it could be inlined. That said, the raw string in `setAttr` isn't ideal for inlining. Do we already have a function that does this?

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -727,10 +719,12 @@ void LowerLayersPass::runOnOperation() {
   OpBuilder builder(circuitOp);
   SmallVector<std::pair<LayerOp, StringAttr>> layers;
   StringRef circuitName = circuitOp.getName();
+
   circuitOp.walk<mlir::WalkOrder::PreOrder>([&](LayerOp layerOp) {
     auto parentOp = layerOp->getParentOfType<LayerOp>();
     while (parentOp && parentOp != layers.back().first)
       layers.pop_back();
+
```
seldridge, line 727: Mega-nits: unrelated whitespace changes. No problem landing anything like this separately on `main`, though!

test/Dialect/FIRRTL/annotations-errors.mlir
```diff
@@ -417,3 +417,99 @@ firrtl.circuit "Issue5947"
...
+        {
+      class = "circt.OutputDirAnnotation",
+      dirname = "foo",
+      target = "~Top|Top"
+    }
+  ]
+} {
+  // expected-error @below {{circt.OutputDirAnnotation must target a public module}}
+  firrtl.module private @Top() {}
+}
```
seldridge, line 515: Mega-nit: it would be better to have the order of the test match the order of the checks. I.e., this would be better earlier in the test.

That said, thank you for thoroughly testing all error cases!

test/firtool/lower-layers-directories.fir
```diff
@@ -118,10 +118,10 @@ circuit Testbench: %[[
...
 ; The following modules are all in the Testbench directory.
 ;
-; CHECK-DAG: FILE "testbench{{[/\]}}Bar.sv"
-; CHECK-DAG: FILE "testbench{{[/\]}}DUT_A.sv"
-; CHECK-DAG: FILE "testbench{{[/\]}}Testbench.sv"
-; CHECK-DAG: FILE "testbench{{[/\]}}layers_Testbench_A.sv"
+; CHECK-DAG: FILE "verification{{[/\]}}testbench{{[/\]}}layers_Testbench_A.sv"
+; CHECK-DAG: FILE "verification{{[/\]}}testbench{{[/\]}}Bar.sv"
+; CHECK-DAG: FILE "verification{{[/\]}}testbench{{[/\]}}DUT_A.sv"
+; CHECK-DAG: FILE "verification{{[/\]}}testbench{{[/\]}}Testbench.sv"
```
seldridge, line 124: This nicely matches the spirit of the original test.

test/Dialect/FIRRTL/assign-output-dirs.mlir
```diff
@@ -0,0 +1,95 @@
...
+    firrtl.instance ra @ByRA()
+    firrtl.instance ab @ByAB()
+    firrtl.instance a  @ByA()
+    firrtl.instance ac @ByAC()
+  }
+
+  firrtl.module @InB() attributes {output_file = #hw.output_file<"B/foo">} {
+    firrtl.instance ab @ByAB()
+    firrtl.instance bc @ByBC()
+  }
```
seldridge, line 59: This is a great edge case to test!

seldridge (approved):
LGTM

Thanks for the iteration on this, Rob. I think you've arrived at an excellent, minimal, and understandable design point. 💯

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,158 @@
...
+          } else {
+            moduleOutputDir = outputDir;
+          }
+          ++i;
+          break;
+        }
+      }
+      for (; i != e; ++i) {
+        if (auto parent =
+                dyn_cast<FModuleOp>((*i)->getParent()->getModule<FModuleOp>()))
```
dtzSiFive, line 133: Probably the dyn_cast AND the `getModule<FModuleOp>` are redundant? Whichever style works for me, but use same above too please :+1:.

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  void runOnOperation() override;
+};
+} // namespace
+
+void AssignOutputDirsPass::runOnOperation() {
+  SmallString<64> outputDir(outputDirOption);
+  if (fs::make_absolute(outputDir)) {
+    llvm::errs()
```
dtzSiFive, line 92: Passes should prefer to emit diagnostics rather than printing to `llvm::errs()` themselves.

This is a bit unusual as its a pass option so doesn't obviously have a location to report on-- maybe report on the current operation or if that's strange, maybe `UnknownLoc`? WDYT?

dtzSiFive (approved):
Thanks for pushing on this!!! Small feedback, generally LGTM! 🎉

rwy7:
Thanks everyone!

PR #6972, opened 2024-04-30
[Arc] Hoist reset value in CompReg when lowering for simulation

Moxinilian:
Arcilator currently does not support non-zero reset values for compreg. This change canonicalizes reset values to a mux on the input, as it currently makes no difference for simulation purposes, and sidesteps the reset value limitations.

### DIFF
--- a/lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
+++ b/lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
@@ -560,6 +560,12 @@ CompRegCanonicalizer::matchAndRewrite(seq::CompRegOp op,
   if (!op.getReset())
     return failure();
 
+  // Because Arcilator supports constant zero reset values, skip them.
+  if (auto *resetValueDefiningOp = op.getResetValue().getDefiningOp())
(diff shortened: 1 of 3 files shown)


maerhart:
As you pointed out, the ConvertToArcs pass supports zero-resets by adding them as operands to the `arc.state` operations. This patch removes zero-resets as well, thus relying on InferStateProperties to detect and add it again.

Should we instead only apply this pattern if the reset value is non-zero?

Moxinilian:
I updated it accordingly. I used a pattern match to check for constant zero instead of the naive `hw.constant` check in Arcilator in case canonicalization patterns next to the CompRegCanonicalizer would still need to be ran to apply simplification.

maerhart (approved):
LGTM. Thanks for adding support!

If you want to be very thorough you could add a very small integration test with just a compreg that has a non-zero reset value since this patch assumes that the arc canonicalizer will always be run before ConvertToArcs. So, if someone changes the Arcilator pipeline, the loss of support could go unnoticed.

Moxinilian:
I wanted to do that and forgot about it! Good idea.

maerhart (approved):
Nice, thanks!

fabianschuiki:
Very cool, thanks for adding this! :partying_face:

PR #6975, opened 2024-04-30
[firtool] Integrate AssertProperty lowering into BTOR2 pipeline

dobios:
This PR depends on #6974 and simply adds the LTL lowering to the btor2 pipeline in firtool. This should enable the btor2 pass to be exposed in Chisel.


lib/Firtool/Firtool.cpp
```diff
@@ -386,6 +386,7 @@ LogicalResult firtool::populateFinalizeIR(mlir::PassManager &pm,
 LogicalResult firtool::populateHWToBTOR2(mlir::PassManager &pm,
                                          const FirtoolOptions &opt,
                                          llvm::raw_ostream &os) {
+  pm.addNestedPass<hw::HWModuleOp>(circt::createLowerLTLToCorePass());
```
dtzSiFive, line 389: Register this pass over in `tools/firtool/firtool.cpp` next to the `registerConvertHWToBTOR2Pass();` line

test/firtool/btor2-assertproperty.fir
```diff
@@ -0,0 +1,80 @@
+; RUN: firtool --btor2 %s | FileCheck %s
+
+FIRRTL version 3.2.0
+circuit Counter :
+  intmodule HasBeenResetIntrinsic :
```
dtzSiFive, line 5: Consider intrinsic expressions instead of intmodule's, but this is fine as-is (even if in the future intmodule support is dropped, NBD).

dtzSiFive (approved):
LGTM, although please register in `firtool` so the `-mlir-print-*` options know about it / work for it.

PR #6980, opened 2024-05-01
[FIRRTL] Add inline convention to layers

seldridge:
Add an "inline" convention to layers by adding this to the existing enumeration.  By the nature of how the parser works, this automatically adds support to parse this attribute without modifications to the parser.

This does not add any support for this in the FIRRTL pipeline.


dtzSiFive:
Super neat all it took was just extending the enum!!!

Consider checking this emitter-side but looks like that just works similarly (untested)! So up to you.

dtzSiFive (approved):

seldridge:
> Consider checking this emitter-side but looks like that just works similarly (untested)! So up to you.

Yes, the emitter happily works, too. I modified an emitter test to check this here: 5063cc018

PR #6981, opened 2024-05-01
[FIRRTL] Drop dead ScalaClassAnnotation.

dtzSiFive:
We've rejected this during LowerAnnotations since we've had LowerAnnotations, so it's safe to say this is thoroughly dead.

The internal source of this was removed in January.


youngar (approved):
🥳

PR #6984, opened 2024-05-03
[FIRRTL][SFCCompat] Fix tests and handling of fullasyncreset on non-port.

dtzSiFive:
Fix test and extend to cover non-port case:
* Fix async-reset-anno test to fail with its backing change reverted, as-is the passes weren't run and the FileCheck command was wrong for the "did it remove the annotation" check.
* Extend async-reset-anno.fir and async-reset.fir tests to cover use of the fullasyncreset annotation on non-ports, which presently fails.

Fix handling of non-port fullasyncreset in SFCCompat (including annotation removal)
* Modify SFCCompat to look for (and remove) fullasyncreset annotations anywhere, not limited to ports.

cc #6912 .

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
@@ -46,13 +46,22 @@ void SFCCompatPass::runOnOperation() {
   bool madeModifications = false;
   SmallVector<InvalidValueOp> invalidOps;
 
-  bool fullAsyncResetExists = false;
-  AnnotationSet::removePortAnnotations(
+  auto fullAsyncResetAttr =
+      StringAttr::get(&getContext(), fullAsyncResetAnnoClass);
+  auto isFullAsyncResetAnno = [fullAsyncResetAttr](Annotation anno) {
+    return anno.getClassAttr() == fullAsyncResetAttr;
+  };
+  bool fullAsyncResetExists = AnnotationSet::removePortAnnotations(
       getOperation(), [&](unsigned argNum, Annotation anno) {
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
```diff
@@ -46,13 +46,22 @@ void SFCCompatPass::runOnOperation() {
...
+  };
+  bool fullAsyncResetExists = AnnotationSet::removePortAnnotations(
       getOperation(), [&](unsigned argNum, Annotation anno) {
-        if (!anno.isClass(fullAsyncResetAnnoClass))
-          return false;
-        return fullAsyncResetExists = true;
+        return isFullAsyncResetAnno(anno);
       });
+  if (!fullAsyncResetExists) {
+    getOperation()->walk([&](Operation *op) {
```
dtzSiFive, line 59: This walk is unfortunate.

It's also good to be sure to fully remove these annotations.

uenoku: Is it guaranteed that there is no isFullAsyncResetAnno when isFullAsyncResetAnno is on ports?

uenoku (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
dtzSiFive: It's checked by a previous pass but that's not great, good question.  Might be better to just walk it all here anyway.

dtzSiFive: Fixed to always walk, thanks!

PR #2968, opened 2022-04-25
[FIRRTL] Do not dedup testharness memories

prithayan:
This commit adds a unique groupID to each Testharness memory to prevent their
 Dedup. This commit is a temporary fix to match SFC behavior. 
Currently, the memory dedup happens during the metadata emission, because that
 is the first pass that requires the name of the final memory module. The plan
 is to create a new Memory lowering pass, and move all the memory transformations to it.

PR #6985, opened 2024-05-03
[FIRRTL][Dedup] Alter dedup group handling, avoid exponential behavior.

### LINKED ISSUE #6979: [FIRRTL] Performance Degradation in Dedup Runtime between Chisel3.6 and Chisel6
When running de-duplication on a .fir file with several ChipTops, the dedup process increases by 6-10x. 
For chisel3.6 dedup time is `9.2838 (  3.0%)    9.2838 (  9.6%)    Dedup`.
For chisel6 dedup time is `63.3732 ( 17.5%)   63.3732 ( 41.8%)    Dedup`

You can recreate this,at least in chipyard, by building the following config on different versions of chisel with 
`make -C sims/vcs CONFIG=RocketMultiChipConfig`

Example Multi ChipTop Config
---------------------------------
```scala
class RocketMultiChipConfig
    extends Config(
      new chipyard.harness.WithHomogeneousMultiChip(
        n = 32,
        new Config(
          new RocketConfig
        )
      ) ++
        new chipyard.harness.WithAbsoluteFreqHarnessClockInstantiator
    )
```

While doing a lot of printf debugging, I found that the increased time is spent in this section of the code, which seems to be just processing the annotation: https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/Transforms/Dedup.cpp#L1074-L1118

Some observations I also encountered were that deleting all of the DedupGroupAnnotations in the .fir dropped the dedup time substantially (which I suppose would be an obvious solution since previous versions of chisel didn't use these annotations). 

In a custom Config that uses `WithHomogeneousMultiChip` , I have built 4 ChipTops and also seen the total memory balloon to 100GB of RAM during the entire dedup process. 

Workaround in the meantime is to just add the --no-dedup flag.

dtzSiFive:
Transmute these annotations to simple (temporary) attributes on modules.
This causes the desired behavior for hashing and since this must match, when deduplicating no work is needed.

Presently, as annotations, each module dedup'd into the group walks all annotations and "adds context" (makes them non-local) and interns a new annotation array with those annotations. At the end of the pass, all dedupGroup annotations are removed.

This causes a lot of unnecessary symbols, hierpaths, annotations, and arrays of annotations as well as some quadratic behavior.

DedupGroup annotations are commonly on every single module, so for those designs large dedup groups scaled poorly.

Instead, drop these annotations immediately and add as a simple named attribute indicating the group.

Drop the (temporary) dedup group attributes at end of pass, as they are just an in-IR way to track this per-module during the pass.

Fixes #6979.

----------

Benchmarking (n=10, median) on some quick synthetic examples put together, FWIW (top has instances of the many identical empty modules):

|Name     |Modules|Instances|real before (s)|real after (s)|Before/After|maxrss before (kb)|maxrss after (kb)|Before/After|
|---------|-------|---------|---------------|--------------|------------|------------------|-----------------|------------|
|Groups   |3001   |3000     |1.035          |0.311         |3.33        |124,096           |72,944           |1.70        |
|Groups_3x|3001   |9000     |2.237          |0.361         |6.20        |216,696           |77,456           |2.80        |
|Many     |10001  |30000    |33.349         |3.461         |9.64        |1,802,784         |475,366          |3.79        |

Significantly improves performance where there are many instances of deduplicated modules and there are dedup group annotations.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -117,10 +117,8 @@ struct StructuralHasher {
       : constants(constants){};
 
   std::pair<std::array<uint8_t, 32>, SmallVector<StringAttr>>
-  getHashAndModuleNames(FModuleLike module, StringAttr group) {
+  getHashAndModuleNames(FModuleLike module) {
     update(&(*module));
-    if (group)
-      sha.update(group.str());
     auto hash = sha.final();
     return {hash, referredModuleNames};
   }
@@ -337,7 +335,7 @@ struct Equivalence {
   Equivalence(MLIRContext *context, InstanceGraph &instanceGraph)
       : instanceGraph(instanceGraph) {
     noDedupClass = StringAttr::get(context, noDedupAnnoClass);
-    dedupGroupClass = StringAttr::get(context, dedupGroupAnnoClass);
+    dedupGroupAttrName = StringAttr::get(context, "firrtl.dedup_group");
     portDirectionsAttr = StringAttr::get(context, "portDirections");
     nonessentialAttributes.insert(StringAttr::get(context, "annotations"));
     nonessentialAttributes.insert(StringAttr::get(context, "name"));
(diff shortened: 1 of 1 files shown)


youngar:
Just to connect the dots, this is similar to what we do to prevent memories in the testharness and (i think) prefix-module-groups from deduping with each other: https://github.com/llvm/circt/pull/2968

youngar (approved):
As we talked about offline, one idea for the future is to change the annotation handler to attach this attribute (in LowerAnnotations) instead of doing it here.  Change looks great to me, thanks!

PR #6991, opened 2024-05-06
[FSM]New builders for StateOp and TransitionOp.

HahaLan97:
1. Now there is a new builder for `StateOp`, which also takes a ValueRange and uses it to create the `OutputOp` with it.
2. Change the first builder of `TransitionOp` to be able to take two `std::function<void>` as arguments, which are similar to the `sv::IfOp` and are by default null. Which means, this change doesn't affect anything already existed.

These are simple changes, but would be very helpful when creating these two ops. One doesn't have to always use `ensureGuard` and `ensureAction` to create the blocks and set the InsertionPoint. Don't mention that `ensureGuard` creates an empty `ReturnOp`, which in most cases need to be erased every time. The empty `OutputOp` created in StateOp is in most cases useless as well.


HahaLan97:
OK... In my own repo the build test doesn't show detailed information. The test failed because of the FileCheck of `test/Conversion/FSMToSv/test_basic.mlir`, because I change the behavior of empty block printing. Then I comment out the change and test the pass again, it worked without error! The error I encountered last year vanished :rofl: Now Let's try the GitHub test again.

lib/Dialect/FSM/FSMOps.cpp
```diff
@@ -401,19 +413,40 @@ LogicalResult OutputOp::verify() {
...
 //===----------------------------------------------------------------------===//
 
-void TransitionOp::build(OpBuilder &builder, OperationState &state,
-                         StringRef nextState) {
-  state.addRegion(); // guard
-  state.addRegion(); // action
-  state.addAttribute("nextState",
-                     FlatSymbolRefAttr::get(builder.getStringAttr(nextState)));
-}
+// void TransitionOp::build(OpBuilder &builder, OperationState &state,
```
mortbopet, line 416: Remove commented out code

include/circt/Dialect/FSM/FSMOps.td
```diff
@@ -301,8 +302,11 @@ def TransitionOp : FSMOp<"transition", [HasParent<"StateOp">, NoTerminator]> {
   }];
 
   let builders = [
-    OpBuilder<(ins "StringRef":$nextState)>,
-    OpBuilder<(ins "fsm::StateOp":$nextState)>
+    // OpBuilder<(ins "StringRef":$nextState)>,
+    OpBuilder<(ins "fsm::StateOp":$nextState)>,
+    OpBuilder<(ins "StringRef":$nextState,
+                   CArg<"const std::function<void()>&", "{}">:$guardCtor,
```
mortbopet, line 308: use `llvm::function_ref` instead of `const std::function<void()>&`

lib/Dialect/FSM/FSMOps.cpp
```diff
@@ -401,19 +413,40 @@ LogicalResult OutputOp::verify() {
...
+                     FlatSymbolRefAttr::get(builder.getStringAttr(nextState)));
+  OpBuilder::InsertionGuard guard(builder);
+
+  Region *guardRegion = state.addRegion(); // guard
+  if (guardCtor) {
+    builder.createBlock(guardRegion);
+    guardCtor();
+  }
+
+  Region *actionRegion = state.addRegion(); // guard
```
mortbopet, line 443: `// guard` -> `// action`?

HahaLan97:
Thanks @mortbopet, the changes are pushed ;)

mortbopet (approved):

mortbopet:
@HahaLan97 your changes already LGTM - did you request a re-review because you want me to merge the changes?

HahaLan97:
> @HahaLan97 your changes already LGTM - did you request a re-review because you want me to merge the changes?

Actually, I was trying to notify another reviewer and accidentally clicked request review button :rofl: But it would be so nice if you merge into the main branch ;)

PR #6995, opened 2024-05-07
[ImportVerilog] Fix unknown name caused by local variables.

hailongSun2000:
If we don't collect the local temporary variables, it triggers the following error:
For example:
```
module Temporary();
  int x, y, z;
  always_comb begin
    automatic int a;
    a = x + 1;
    y = a;
    a = a + 1;
    z = a;
  end
endmodule
```
```cpp
Temporary.sv:5:5: error: unknown name `a`
    a = x + 1;
```

### DIFF
--- a/lib/Conversion/ImportVerilog/Statements.cpp
+++ b/lib/Conversion/ImportVerilog/Statements.cpp
@@ -63,8 +63,11 @@ struct StmtVisitor {
         return failure();
     }
 
-    builder.create<moore::VariableOp>(loc, type,
-                                      builder.getStringAttr(var.name), initial);
+    // Collect local temporary variables.
+    auto varOp = builder.create<moore::VariableOp>(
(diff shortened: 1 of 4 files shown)


uenoku (approved):
LGTM! Could you add a test?

hailongSun2000:
> LGTM! Could you add a test?

No problem! But I'll wait to add it tomorrow. And the IR is the same as non-local variables.

fabianschuiki (approved):
Thanks for the fix! I agree with @uenoku: once there's a test this is good to go 😃

hailongSun2000:
> Thanks for the fix! I agree with @uenoku: once there's a test this is good to go 😃

I have added the test cases.

hailongSun2000:
> LGTM! Could you add a test?

I have added the test cases.

PR #6999, opened 2024-05-07
[firtool] Add FlattenModulesPass to the btor2 emission pipeline

dobios:
In order to support multi-module designs in the formal backend, we need to inline the modules before converting them to btor2. 

This is a naive approach but will work just fine for now.

### DIFF
--- a/tools/firtool/firtool.cpp
+++ b/tools/firtool/firtool.cpp
@@ -710,7 +710,7 @@ int main(int argc, char **argv) {
     firrtl::registerPasses();
     om::registerPasses();
     sv::registerPasses();
-    hw::registerPasses();
+    hw::registerFlattenModulesPass();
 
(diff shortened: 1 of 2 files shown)


tools/firtool/firtool.cpp
```diff
@@ -710,6 +710,7 @@ int main(int argc, char **argv) {
     firrtl::registerPasses();
     om::registerPasses();
     sv::registerPasses();
+    hw::registerPasses();
```
dtzSiFive, line 713: This pulls in a few passes more than strictly needed, including some passes unreachable from `firtool`.

I'm okay with this but as HW grows more passes we might not want to pull them all in uncritically.

Looks like we're already pulling in all SV passes which is more problematic, so something to cleanup ....

PR #6993, opened 2024-05-06
LLVM Bump

dobios:

PR #7002, opened 2024-05-08
[FIRRTL] AnnoTarget: use LLVM style casts

youngar:
This changes AnnoTargets to use LLVM style casts instead of "trailing" member functions. The LLVM style casts are now usable with AnnoTargets with improvements to the upstream framework.  The current style of cast used with AnnoTargets will no longer be supported by TypeSwitch, so this PR switches to the recommended idiom.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
@@ -415,25 +415,6 @@ protected:
 struct AnnoTarget {
   AnnoTarget(detail::AnnoTargetImpl impl = nullptr) : impl(impl){};
 
-  template <typename U>
-  bool isa() const { // NOLINT(readability-identifier-naming)
-    assert(*this && "isa<> used on a null type.");
-    return U::classof(*this);
-  }
-  template <typename U>
-  U dyn_cast() const { // NOLINT(readability-identifier-naming)
-    return isa<U>() ? U(impl) : U(nullptr);
-  }
-  template <typename U>
-  U dyn_cast_or_null() const { // NOLINT(readability-identifier-naming)
-    return (*this && isa<U>()) ? U(impl) : U(nullptr);
-  }
-  template <typename U>
-  U cast() const {
-    assert(isa<U>());
-    return U(impl);
-  }
-
   operator bool() const { return impl; }
   bool operator==(const AnnoTarget &other) const { return impl == other.impl; }
   bool operator!=(const AnnoTarget &other) const { return !(*this == other); }
@@ -521,6 +502,29 @@ LogicalResult extractDUT(FModuleOp mod, FModuleOp &dut);
 //===----------------------------------------------------------------------===//
 
 namespace llvm {
+
+/// Add support for llvm style casts to AnnoTarget.
+template <typename To, typename From>
+struct CastInfo<
+    To, From,
+    std::enable_if_t<
+        std::is_same_v<circt::firrtl::AnnoTarget, std::remove_const_t<From>> ||
+        std::is_base_of_v<circt::firrtl::AnnoTarget, From>>>
+    : NullableValueCastFailed<To>,
+      DefaultDoCastIfPossible<To, From, CastInfo<To, From>> {
+  static inline bool isPossible(From target) {
+    /// Return a constant true instead of a dynamic true when casting to self or
+    /// up the hierarchy.
+    if constexpr (std::is_base_of_v<To, From>) {
+      return true;
+    } else {
+      return To::classof(target);
+    }
+  }
+  static inline To doCast(circt::firrtl::AnnoTarget target) {
+    return To(target.getImpl());
+  }
+};
 
 /// Make `Annotation` behave like a `Attribute` in terms of pointer-likeness.
 template <>
--- a/include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
@@ -507,23 +507,18 @@ namespace llvm {
 template <typename To, typename From>
 struct CastInfo<
     To, From,
-    std::enable_if_t<
-        std::is_same_v<circt::firrtl::AnnoTarget, std::remove_const_t<From>> ||
-        std::is_base_of_v<circt::firrtl::AnnoTarget, From>>>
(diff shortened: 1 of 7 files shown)


include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
```diff
@@ -522,6 +503,29 @@ LogicalResult extractDUT(FModuleOp mod, FModuleOp &dut);
 
 namespace llvm {
 
+/// Add support for llvm style casts to AnnoTarget.
+template <typename To, typename From>
+struct CastInfo<
+    To, From,
+    std::enable_if_t<
+        std::is_same_v<circt::firrtl::AnnoTarget, std::remove_const_t<From>> ||
```
dtzSiFive, line 511: What does the `is_same_v` part of this do that isn't covered by `is_base_of_v`? Maybe something with incomplete types?
Looking into this a bit, `is_base_of` here returns false if the types are incomplete (apparently it's undefined behavior for second type to be incomplete) (so really should avoid making it possible to use this with an incomplete derived type, which is easy by just putting it at the end like you've done).

Anyway, is the `is_same_v` doing anything?  (I don't think so?)
I see this combination in some MLIR code but nowhere else, hmm.  Not blocking, just curious.

https://github.com/llvm/llvm-project/blob/50b45b24220ead33cf5cedc49c13e0336297e4eb/llvm/include/llvm/Support/Casting.h#L275-L280 for example seems just fine using only `std::is_base_of_v`.
And I was wondering why we need to do the special handling for self/upcasts ourselves, but probably by defining this we override that behavior?

FWIW this is defining cast support /from/ AnnoTarget to anything else.  This appears to be done in some MLIR bits too.  Might be good idea to limit this for `To` that are also AnnoTarget-or-derived, so that (in theory) casting for other types could be specialized/defined (perhaps with different `doCast` than feeding the impl pointer to the constructor of `To`)?

dtzSiFive (approved):
Awesome, thanks for handling this!  Was curious about C++ bits a bit but LGTM :+1:.

dobios:
Any chance we could merge this so that I can finish up the LLVM bump?

seldridge:
> Any chance we could merge this so that I can finish up the LLVM bump?

Drive by comment: one way to handle this is to create a "stacked" PR. Unfortunately, GitHub doesn't have great support for this. However, you can fake it by rebasing the LLVM bump on top of this branch and then changing the base branch of the LLVM bump PR to point at this branch. When this PR lands, GitHub will automatically change the base branch of the LLVM bump PR to point at `main`.

There are a number of edge cases that can cause this to break, but it kind of works and I tend to do it frequently. 🤓

include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
```diff
@@ -507,23 +507,18 @@ namespace llvm {
...
-        std::is_base_of_v<circt::firrtl::AnnoTarget, From>>>
+    std::enable_if_t<std::is_base_of_v<::circt::firrtl::AnnoTarget, From>>>
     : NullableValueCastFailed<To>,
       DefaultDoCastIfPossible<To, From, CastInfo<To, From>> {
   static inline bool isPossible(From target) {
-    /// Return a constant true instead of a dynamic true when casting to self or
-    /// up the hierarchy.
-    if constexpr (std::is_base_of_v<To, From>) {
+    // Allow constant upcasting.  This also gets around the fact that AnnoTarget
+    // does not implement classof.
```
dtzSiFive, line 515: 👍 

Leaves `To` open to non-AnnoTarget's but I think it's good as-is.
What's annoying, kinda, is that isPossible I think by default is these things (base_of_v for statically resolving when self/upcast, `T::classof` otherwise) but we need to tell it how casts for these work since value-value.

Anyway! Looks great, thanks!

fabianschuiki (approved):

include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
youngar: > Anyway, is the is_same_v doing anything?

AFAICT it isn't doing anything, it might just be a mistake upstream because `is_base_of` is a little unintuitive. Quote from [cppreference](https://en.cppreference.com/w/cpp/types/is_base_of) "Although no class is its own base, std::is_base_of<T, T>::value is true because the intent of the trait is to model the "is-a" relationship, and T is a T. "

> FWIW this is defining cast support /from/ AnnoTarget to anything else. This appears to be done in some MLIR bits too. Might be good idea to limit this for To that are also AnnoTarget-or-derived, so that (in theory) casting for other types could be specialized/defined (perhaps with different doCast than feeding the impl pointer to the constructor of To)?

I tried to limit the scope of this specialization to only apply when `To` is a subclass of `AnnoTarget`, but since `is_base_of` does not support incomplete types, I first need to check that the `To` type is complete.  Without this check, this specialization is considered for all sorts of random things in LLVM and causes an explosion of errors.  It seems like [it might be possible](https://stackoverflow.com/questions/1625105/how-to-write-is-complete-template) to create a template which checks this, but [seems like a bad idea](https://lists.isocpp.org/std-proposals/2021/11/3309.php).

dtzSiFive: > I tried to limit the scope of this specialization to only apply when `To` is a subclass of `AnnoTarget`, but since `is_base_of` does not support incomplete types

Ahhh, right.  Okay.

Weird because `From` could be incomplete too? But okay, and thanks for the links/summary/explanation!

include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
dtzSiFive: Whatever you think is best, thank you very much.

As-is or two ones if you'd prefer :+1:. 🙏

dobios:
Caught a whole bunch of deprecated casts all over CIRCT. Also TypeSwitch doesn't support null arguments anymore, so that had to be handled as well, it's all in the [LLVM Bump PR](https://github.com/llvm/circt/pull/6993)

include/circt/Dialect/FIRRTL/FIRRTLAnnotations.h
youngar: I tried a couple things but couldn't get it to work, so I just left it as is.  We can always make this better in a follow up PR.

PR #7003, opened 2024-05-08
[SV] Add sv.func, sv.return and sv.func.call operations

uenoku:
This PR adds function call constructs (sv.func, sv.return and sv.func.call) to SV dialect. 

* `sv.func` -- this op is between hw.module and func.func. This op implement interfaces `PortList`, `HWModuleLike` and `FunctionOpInterface`. Similar to hw.module, we can preserve argument names and mix the order of input and output arguments. Following func.func empty body indicates that the op is a declaration. 
* `sv.return` -- terminator for sv.func. We cannot reuse neither hw.output nor func.func since they have constraints on parent op(and I think it's not reasonable to loose them). 
* `sv.func.call` and `sv.func.call.procedural`. Two operations are added because output arguments cannot be used non-procedural regions.

### DIFF
--- a/include/circt/Dialect/SV/SVStatements.td
+++ b/include/circt/Dialect/SV/SVStatements.td
@@ -12,6 +12,9 @@
 
 include "mlir/IR/EnumAttr.td"
 include "mlir/IR/OpAsmInterface.td"
+include "mlir/Interfaces/FunctionInterfaces.td"
+include "mlir/Interfaces/ControlFlowInterfaces.td"
+include "circt/Dialect/Emit/EmitOpInterfaces.td"
 
 //===----------------------------------------------------------------------===//
 // Control flow like-operations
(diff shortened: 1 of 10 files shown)


include/circt/Dialect/SV/SVStatements.td
```diff
@@ -879,3 +882,167 @@ def MacroDefOp : SVOp<"macro.def",
...
+//===----------------------------------------------------------------------===//
+
+def FuncOp : SVOp<"func",
+     [IsolatedFromAbove, Symbol, OpAsmOpInterface, ProceduralRegion,
+      DeclareOpInterfaceMethods<HWModuleLike>,
+      DeclareOpInterfaceMethods<PortList>,
+      FunctionOpInterface, HasParent<"mlir::ModuleOp">]> {
+  let summary = "A System Verilog function";
+  let description = [{
+    `sv.func` reresents System Verilog function in SV spec 13.4.
```
fabianschuiki, line 898:
```suggestion
    `sv.func` reresents SystemVerilog function in SV spec 13.4.
```

Can you also mention which exact version of the spec this is? They tend to insert paragraphs, which makes the numbering unstable. Something like `see IEEE 1800-2017 section 13.4 "section title"` maybe?

lib/Dialect/SV/SVOps.cpp
```diff
@@ -2106,6 +2108,313 @@ ModportStructAttr ModportStructAttr::get(MLIRContext *context,
...
+
+  // Non-procedural call cannot have output arguments.
+  if (referencedOp.getNumOutputs() != 1 ||
+      !referencedOp.getExplicitlyReturnedType()) {
+    auto diag = emitError()
+                << "function called in a non-procedural region must "
+                   "return a single result";
+    diag.attachNote(referencedOp.getLoc()) << "doesn't satisfy the constraint";
+    return failure();
+  }
```
fabianschuiki, line 2414: Nice!

fabianschuiki (approved):
Very cool :partying_face:

uenoku:
I'm going to merge the PR. Post-merge review is welcomed. The current implementation around a return value is slightly hacky especially the use of "sv.func.explicitly_returned". Probably a better way would be to introduce HWFunctionType with an explicit return type but I'll consider it as a follow-up.

PR #6977, opened 2024-05-01
[SV][ExportVerilog] Add sv.func, sv.func.call{.procedural}, sv.func.dpi.import

uenoku:
This PR adds operations sv.func, sv.func.call{.procedural} and sv.func.dpi.import to SV. 

sv.func op is model as a mixture of func.func and hw.module. Similar to hw.module, sv.func can mix the order of input and output arguments. sv.func implements FunctionOpInterface and can declare a function by an empty body region like func.func

Code around PrepareForEmission is ported from https://github.com/llvm/circt/pull/6559.

```mlir
sv.func @fib(in %n : i32, out out : i32 {sv.func.explicitly_returned}) {
  %one = hw.constant 1 : i32
  %two = hw.constant 2 : i32
  %cond = comb.icmp bin ule %n, %one: i32
  %result = sv.logic : !hw.inout<i32>
  sv.if %cond {
    sv.bpassign %result, %n: i32 
  } else {
    %n1 = comb.sub %n, %one: i32
    %result1 = sv.func.call.procedural @fib(%n1) : (i32) -> i32
    %n2 = comb.sub %n, %two: i32
    %result2 = sv.func.call.procedural @fib(%n2) : (i32) -> i32
    %add = comb.add %n1, %n2: i32
    sv.bpassign %result, %add: i32 
  }
  %read = sv.read_inout %result: !hw.inout<i32>
  sv.return %read : i32
}
```

 
```verilog
function automatic logic [31:0] fib(    // test.mlir:1:1
  input [31:0] n        // test.mlir:1:17
);

  logic           [31:0] _fib_0;        // test.mlir:12:16
  logic           [31:0] _fib_0_0;      // test.mlir:10:16
  logic           [31:0] result;        // test.mlir:5:13
  if (n <= 32'h1)       // test.mlir:2:10, :4:11, :6:3
    result = n; // test.mlir:7:5
  else begin    // test.mlir:6:3
    logic           [31:0] _GEN = n - 32'h1;    // test.mlir:2:10, :9:11
    logic           [31:0] _GEN_0 = n - 32'h2;  // test.mlir:3:10, :11:11
    _fib_0_0 = fib(_GEN);       // test.mlir:9:11, :10:16
    _fib_0 = fib(_GEN_0);       // test.mlir:11:11, :12:16
    result = _GEN + _GEN_0;     // test.mlir:9:11, :11:11, :13:12, :14:5
  end
  fib = result; // test.mlir:16:11, :17:3
endfunction
```

PR #7004, opened 2024-05-08
[ExportVerilog][HW] Introduce HWEmittableModuleLike interface and use it for Prepare, NFC

uenoku:
This is separated from https://github.com/llvm/circt/pull/6977. This changes pre-passes for ExportVerilog to run on HWEmittableModuleLike instead of HWModuleOp. https://github.com/llvm/circt/pull/6977 is going to add a support for SV func op and legalization needs to be ran on SV func as well. 

HWEmittableModuleLike is a new interface that inherits Emittable+HWModuleLike. I considered to use a trait but we cannot use a trait for pass scheduling so an interface is used.

### DIFF
--- a/include/circt/Dialect/HW/HWOpInterfaces.td
+++ b/include/circt/Dialect/HW/HWOpInterfaces.td
@@ -15,6 +15,7 @@
 
 include "mlir/IR/SymbolInterfaces.td"
 include "mlir/IR/OpBase.td"
+include "circt/Dialect/Emit/EmitOpInterfaces.td"
 include "circt/Support/InstanceGraphInterface.td"
 
 def PortList : OpInterface<"PortList", []> {
@@ -400,6 +401,15 @@ def HWMutableModuleLike : OpInterface<"HWMutableModuleLike", [HWModuleLike]> {
     "void", "appendOutputs", (ins
       "ArrayRef<std::pair<StringAttr, Value>>":$outputs)>
   ];
+}
+
+def HWEmittableModuleLike : OpInterface<"HWEmittableModuleLike", [HWModuleLike,
+                                        Emittable]> {
+  let cppNamespace = "::circt::hw";
+  let description = [{
+    This interface indicates that the module like op is emitatble in SV and
+    requires to be ran SV legalization on its body.
+  }];
 }
 
 
--- a/include/circt/Dialect/HW/HWOpInterfaces.td
+++ b/include/circt/Dialect/HW/HWOpInterfaces.td
@@ -407,8 +407,8 @@ def HWEmittableModuleLike : OpInterface<"HWEmittableModuleLike", [HWModuleLike,
                                         Emittable]> {
   let cppNamespace = "::circt::hw";
   let description = [{
-    This interface indicates that the module like op is emitatble in SV and
-    requires to be ran SV legalization on its body.
+    This interface indicates that the module like op is emittable in SV and
+    requires SV legalization on its body.
   }];
 }
 
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -1416,7 +1416,8 @@ StringAttr ExportVerilog::inferStructuralNameForTemporary(Value expr) {
 
   // Module ports carry names!
   if (auto blockArg = dyn_cast<BlockArgument>(expr)) {
-    auto moduleOp = cast<HWModuleOp>(blockArg.getOwner()->getParentOp());
+    auto moduleOp =
+        cast<HWEmittableModuleLike>(blockArg.getOwner()->getParentOp());
     StringRef name = getPortVerilogName(moduleOp, blockArg.getArgNumber());
     result = StringAttr::get(expr.getContext(), name);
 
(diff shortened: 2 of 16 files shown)


teqdruid:
I'm not sure this is a good idea. Ops that implement hwmodulelike don't necessarily have semantics which directly map to systemverilog semantics. Have you searched for ops that implement it to ensure that we want them to be directly emitted to verilog?

uenoku:
Thank you for the feedback and that's good point. I thought HWModuleOps feed into ExportVerilog are assumed to be emittable as SV but I agree that it would be nicer to explicitly check that. 

I'd like to propose to use Emittable trait https://github.com/llvm/circt/blob/7073e2cfdcd1c76d4aec47be9138663412a570e6/include/circt/Dialect/Emit/EmitOpInterfaces.td#L18-L20 and run PrepareForEmission on HWModuleLike with Emittable trait (or extend Emittable trait to op interface).

teqdruid:
> Thank you for the feedback and that's good point. I thought HWModuleOps feed into ExportVerilog are assumed to be emittable as SV but I agree that it would be nicer to explicitly check that.
> 
> I'd like to propose to use Emittable trait
> 
> https://github.com/llvm/circt/blob/7073e2cfdcd1c76d4aec47be9138663412a570e6/include/circt/Dialect/Emit/EmitOpInterfaces.td#L18-L20
> 
> and run PrepareForEmission on HWModuleLike with Emittable trait (or extend Emittable trait to op interface).

Does ExportVerilog / PrepareForEmission already depend on (know about) the Emission dialect?

I'm more comfortable with HWModuleLike ops to have to opt-in to Emittable. I'd be even more comfortable if Emittable were an OpInterface with a `bool emittedToVerilog()` method.

This raises the question: why do we want all HWModuleLike ops to be capable of being emitted to Verilog? Why not lower them into HWModuleOps? Is there a specific use-case?

teqdruid:
Or is this intended for preparation only? In that case, I think that any op which implements a future method like `Emittable.shouldBePreparedForSVEmission()` (or whatever name) should get legalized -- regardless of whether or not it implements HWModuleLike. Given that you probably need things like ports when preparing, though, this can be considered future work so long as it's documented as a restriction.

uenoku:
> Does ExportVerilog / PrepareForEmission already depend on (know about) the Emission dialect?

Yes. 

> This raises the question: why do we want all HWModuleLike ops to be capable of being emitted to Verilog? Why not lower them into HWModuleOps? Is there a specific use-case?

sv.func is the first use case that is not HWModuleop and also needs PrepareForeEmission. HWModuleLike provides sufficient interface for PrepareForEmisson especially regarding port name legalization. 

> Or is this intended for preparation only? 

Yes. PrepareForEmission and probably PrettifyVerilog.

teqdruid:
ok. I have no problem with this assuming that you do something with the `Emittable` trait.

uenoku:
@teqdruid Sorry for the late update. I iterated a bit more and I'm inclined to introduce a new interface `HWEmittableModuleLike` which inherits `HWModuleLike` and `Emittable` to run SV specific legalization.  
 PrepareForEmission, PrettifyVerilog and HWLegalizeModule should be converted into interface pass (in the current PR only PrepareForEmission is modified). 

I thought it might make sense to put this interface to SV but I end up adding the interface to HW due to the dependency around HWModuleLike.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6785,8 +6786,9 @@ LogicalResult circt::exportVerilog(ModuleOp module, llvm::raw_ostream &os) {
   LoweringOptions options(module);
   if (failed(lowerHWInstanceChoices(module)))
     return failure();
-  SmallVector<HWModuleOp> modulesToPrepare;
-  module.walk([&](HWModuleOp op) { modulesToPrepare.push_back(op); });
+  SmallVector<HWEmittableModuleLike> modulesToPrepare;
+  module.walk(
+      [&](HWEmittableModuleLike op) { modulesToPrepare.push_back(op); });
```
uenoku, line 6791: It doesn't need to be walk so will change it.

include/circt/Dialect/HW/HWOpInterfaces.td
```diff
@@ -402,6 +403,15 @@ def HWMutableModuleLike : OpInterface<"HWMutableModuleLike", [HWModuleLike]> {
   ];
 }
 
+def HWEmittableModuleLike : OpInterface<"HWEmittableModuleLike", [HWModuleLike,
+                                        Emittable]> {
+  let cppNamespace = "::circt::hw";
+  let description = [{
+    This interface indicates that the module like op is emitatble in SV and
```
dtzSiFive, line 410:
```suggestion
    This interface indicates that the module like op is emittable in SV and
```

include/circt/Dialect/HW/HWOpInterfaces.td
```diff
@@ -402,6 +403,15 @@ def HWMutableModuleLike : OpInterface<"HWMutableModuleLike", [HWModuleLike]> {
...
+    requires to be ran SV legalization on its body.
```
dtzSiFive, line 411:
```suggestion
    requires SV legalization on its body.
```
(or other phrasing)

dtzSiFive (approved):
LGTM! Thanks for iterating on this.

fabianschuiki (approved):
LGTM! :party

teqdruid:
This approach sounds reasonable to me. Had a quick look through the diff and it looks about right to me.

PR #7005, opened 2024-05-08
[SV] Add DPI import op

uenoku:
This PR adds an op `sv.func.dpi.import` for "emitting" DPI import statement. `sv.func.dpi.import` doesn't declare function but it takes a function symbol to emit import statement. This design is similar to what `sv.macro.ref/def` does and it enables us to decouple function declaration and emission. 

```
// CHECK-LABEL: import "DPI-C" linkage_name = function void function_declare1(
// CHECK-NEXT:    input [1:0] in_0,
// CHECK-NEXT:                out_0,
// CHECK-NEXT:                in_1,
// CHECK-NEXT:    output out_1
// CHECK-NEXT: );
sv.func.dpi.import linkage "linkage_name" @function_declare1
```

### DIFF
--- a/include/circt/Dialect/SV/SVStatements.td
+++ b/include/circt/Dialect/SV/SVStatements.td
@@ -888,6 +888,17 @@ def MacroDefOp : SVOp<"macro.def",
 // Function/Call
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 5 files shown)


fabianschuiki (approved):
This looks reasonable to me!

PR #7006, opened 2024-05-08
[HWToSMT][circt-lec] Resolve transitive !smt.bool -> i1 -> !smt.bv<1> casts.

fzi-hielscher:
For single bit module outputs the circt-lec  pipeline initially inserts an unrealized conversion cast to `i1`. At a later stage the `i1` value gets converted to `!smt.bv<1>`. This works fine if the original value also was of type  `!smt.bv<1>`, since the casting chain of   `!smt.bv<1>` -> `i1` ->  `!smt.bv<1>` can be reconciled.
However, some SMT operations (e.g. `EqOp`) produce results of type `!smt.bool`. In this case the indirection via `i1` prevents the required conversion to `!smt.bv<1>`  from being materialized.

This PR adds a dedicated target materialization for conversions to  `!smt.bv<1>`  from `i1`  values that are the result of an unrealized conversion cast from `!smt.bool`.

### DIFF
--- a/integration_test/circt-lec/comb.mlir
+++ b/integration_test/circt-lec/comb.mlir
@@ -68,7 +68,21 @@ hw.module @decomposedAnd(in %in1: i1, in %in2: i1, out out: i1) {
 // TODO
 
 // comb.icmp
-// TODO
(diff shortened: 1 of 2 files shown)


maerhart (approved):
Thanks for fixing this!

PR #7009, opened 2024-05-08
[FIRRTL] Reject intrinsic modules >= 4.0.0

dtzSiFive:
https://github.com/chipsalliance/firrtl-spec/pull/210

Should not merge until/if they're removed from spec, but it's ready when we are.

Follow-on "eventually" would be dropping them from our IR / rejecting on all versions.


fabianschuiki (approved):
:partying_face:

PR #7011, opened 2024-05-09
[Moore] Make simple bit vectors a proper MLIR type

fabianschuiki:
The core type most SystemVerilog expressions are interested in and operate on is a "simple bit vector type". These are individual integer atoms like `bit` or `logic`, integral types like `int`, or packed arrays with a single dimension and an integer atom inner type, like `bit [42:0]`. So in a nutshell, simple bit vector types are MLIR's `i42` in the two-valued (`bit`) case, or the four-valued equivalent (`logic`).

Up until this point, the Moore dialect reflected this pattern by providing an `IntType` for the integer atoms like `bit` and `int`, and using the `PackedRangeDim` for single dimension vectors of `bit`. A `SimpleBitVectorType` helper struct would then summarize the _actual_ bit vector that was expressed by the packed range and integer atom. This makes working with the types in TableGen files very annoying, since the thing you are actually interested in -- the simple bit vector -- is not a proper MLIR type, but more like a helper struct on the side.

This commit rips out the existing `IntType` and its composition with a packed array dimension, and replaces it with a proper simple bit vector type that is actually an MLIR type. As a result, SystemVerilog types like `int unsigned`, `bit [42:0]`, `reg`, `logic signed [31:0]`, or `integer` are all translated into the same MLIR type. This new simple bit vector MLIR type retains the `IntType` name, and prints as `!moore.i42` or `!moore.l42`, depending on whether it is a two-valued or four-valued integer. Single `bit` and `logic` atoms become `i1` and `l1` respectively.

This makes the Moore type system a lot easier to work with and removes a lot of unnecessary noise. Operations can now simply use `llvm::isa<IntType>` to check if a value is a simple bit vector.

### DIFF
--- a/include/circt/Dialect/Moore/MooreTypes.td
+++ b/include/circt/Dialect/Moore/MooreTypes.td
@@ -61,24 +61,23 @@ def UnpackedType : MooreType<CPred<"llvm::isa<moore::UnpackedType>($_self)">,
 
 /// A simple bit vector type.
 def SimpleBitVectorType : MooreType<CPred<[{
-    llvm::isa<moore::UnpackedType>($_self) &&
-    llvm::cast<moore::UnpackedType>($_self).isSimpleBitVector()
-  }]>, "simple bit vector type", "moore::UnpackedType">;
+    llvm::isa<moore::IntType>($_self)
+  }]>, "simple bit vector type", "moore::IntType">;
 
-/// A single bit type (`bit`, `logic`, `reg`).
+/// A single two or four-valued bit.
 def AnySingleBitType : MooreType<CPred<[{
     llvm::isa<moore::IntType>($_self) &&
-    llvm::cast<moore::IntType>($_self).getBitSize() == 1
+    llvm::cast<moore::IntType>($_self).getWidth() == 1
   }]>, "single bit type", "moore::IntType">;
 
-/// A `bit` type.
+/// A single two-valued bit.
 def BitType : MooreType<CPred<[{
     llvm::isa<moore::IntType>($_self) &&
-    llvm::cast<moore::IntType>($_self).getBitSize() == 1 &&
+    llvm::cast<moore::IntType>($_self).getWidth() == 1 &&
     llvm::cast<moore::IntType>($_self).getDomain() == moore::Domain::TwoValued
   }]>, "`bit` type", "moore::IntType"> {
   let builderCall = [{
-    $_builder.getType<moore::IntType>(IntType::Kind::Bit)
+    IntType::getInt($_builder.getContext(), 1)
   }];
 }
 
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -27,13 +27,8 @@ struct ExprVisitor {
   Value convertToSimpleBitVector(Value value) {
     if (!value)
       return {};
-    if (auto type = dyn_cast_or_null<moore::UnpackedType>(value.getType())) {
-      if (type.isSimpleBitVector())
-        return value;
-      if (auto sbvt = type.castToSimpleBitVectorOrNull())
-        return builder.create<moore::ConversionOp>(
-            loc, sbvt.getType(builder.getContext()), value);
-    }
+    if (isa<moore::IntType>(value.getType()))
+      return value;
     mlir::emitError(loc, "expression of type ")
         << value.getType() << " cannot be cast to a simple bit vector";
     return {};
@@ -125,7 +120,8 @@ struct ExprVisitor {
     if (!preValue)
       return {};
     preValue = builder.create<moore::ReadLValueOp>(loc, preValue);
-    auto one = builder.create<moore::ConstantOp>(loc, preValue.getType(), 1);
+    auto one = builder.create<moore::ConstantOp>(
+        loc, cast<moore::IntType>(preValue.getType()), 1);
     auto postValue =
         isInc ? builder.create<moore::AddOp>(loc, preValue, one).getResult()
               : builder.create<moore::SubOp>(loc, preValue, one).getResult();
@@ -352,7 +348,8 @@ struct ExprVisitor {
       return {};
     }
     auto truncValue = value.as<uint64_t>().value();
-    return builder.create<moore::ConstantOp>(loc, type, truncValue);
+    return builder.create<moore::ConstantOp>(loc, cast<moore::IntType>(type),
+                                             truncValue);
   }
 
   // Handle `'0`, `'1`, `'x`, and `'z` literals.
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -160,16 +160,13 @@ struct BoolCastOpConversion : public OpConversionPattern<BoolCastOp> {
   LogicalResult
   matchAndRewrite(BoolCastOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    if (cast<UnpackedType>(op.getInput().getType())
-            .castToSimpleBitVectorOrNull()) {
-      Type resultType = typeConverter->convertType(op.getInput().getType());
+    Type resultType = typeConverter->convertType(op.getInput().getType());
+    if (isa_and_nonnull<IntegerType>(resultType)) {
       Value zero = rewriter.create<hw::ConstantOp>(op->getLoc(), resultType, 0);
-
       rewriter.replaceOpWithNewOp<comb::ICmpOp>(op, comb::ICmpPredicate::ne,
                                                 adaptor.getInput(), zero);
       return success();
     }
-
     return failure();
   }
 };
@@ -415,15 +412,13 @@ static void populateLegality(ConversionTarget &target) {
 
 static void populateTypeConversion(TypeConverter &typeConverter) {
   typeConverter.addConversion([&](IntType type) {
-    return mlir::IntegerType::get(type.getContext(), type.getBitSize());
+    return mlir::IntegerType::get(type.getContext(), type.getWidth());
   });
 
   // Directly map simple bit vector types to a compact integer type. This needs
   // to be added after all of the other conversions above, such that SBVs
   // conversion gets tried first before any of the others.
   typeConverter.addConversion([&](UnpackedType type) -> std::optional<Type> {
-    if (auto sbv = type.getSimpleBitVectorOrNull())
-      return mlir::IntegerType::get(type.getContext(), sbv.size);
     if (isa<UnpackedRangeDim, PackedRangeDim>(type))
       return mlir::IntegerType::get(type.getContext(),
                                     type.getBitSize().value());
--- a/lib/Dialect/Moore/MooreTypes.cpp
+++ b/lib/Dialect/Moore/MooreTypes.cpp
@@ -60,25 +60,6 @@ StringRef moore::getKeywordFromSign(const Sign &sign) {
 }
 
 //===----------------------------------------------------------------------===//
-// Simple Bit Vector Type
-//===----------------------------------------------------------------------===//
-
-PackedType SimpleBitVectorType::getType(MLIRContext *context) const {
-  if (!*this)
-    return {};
-
-  // Build the core integer bit type.
-  auto kind = domain == Domain::TwoValued ? IntType::Bit : IntType::Logic;
-  auto intType = IntType::get(context, kind);
-
-  // If the vector is wider than a single bit, or the dimension was explicit in
-  // the original type, add a dimension around the bit type.
-  if (size > 1)
-    return PackedRangeDim::get(intType, size);
-  return intType;
-}
-
-//===----------------------------------------------------------------------===//
 // Unpacked Type
 //===----------------------------------------------------------------------===//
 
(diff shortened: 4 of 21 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -125,7 +120,8 @@ struct ExprVisitor {
     if (!preValue)
       return {};
     preValue = builder.create<moore::ReadLValueOp>(loc, preValue);
-    auto one = builder.create<moore::ConstantOp>(loc, preValue.getType(), 1);
+    auto one = builder.create<moore::ConstantOp>(
+        loc, preValue.getType().cast<moore::IntType>(), 1);
```
hailongSun2000, line 124: I learned that `x.cast()/x.dyn_cast()/x.isa()` is being deprecated by llvm. Could we tweak it into `cast<moore::IntType>(preValue.getType())`?

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -305,226 +313,226 @@ module Expressions;
...
+    // CHECK: moore.constant -1 : !moore.i32
     c = '1;
-    // CHECK: moore.constant 42 : !moore.int
+    // CHECK: moore.constant 42 : !moore.i32
     c = 42;
-    // CHECK: moore.constant 42 : !moore.packed<range<bit, 18:0>>
+    // CHECK: moore.constant 42 : !moore.i19
     c = 19'd42;
-    // CHECK: moore.constant 42 : !moore.packed<range<bit, 18:0>>
+    // CHECK: moore.constant 42 : !moore.i19
```
hailongSun2000, line 324: I think the type of `c = 19'sd42` is `!moore.l19`.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -352,7 +348,8 @@ struct ExprVisitor {
       return {};
     }
     auto truncValue = value.as<uint64_t>().value();
-    return builder.create<moore::ConstantOp>(loc, type, truncValue);
+    return builder.create<moore::ConstantOp>(loc, cast<moore::IntType>(type),
```
hailongSun2000, line 351: I have a question about why do we need to `cast<moore::IntType>(type)`? If I don't misunderstand, the `context.convertType()` will convert slang types like `ScalarType::Bit` into `!moore.i1`?

My working computer is encountering CPU over temperature error. So I can't immediately test these.

hailongSun2000:
There are some outdated C API of `moore` remaining in the `lib/CAPI/Dialect/Moore.cpp`. Like `IntType` and `SBVT`。

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -160,17 +160,12 @@ struct BoolCastOpConversion : public OpConversionPattern<BoolCastOp> {
...
-    if (cast<UnpackedType>(op.getInput().getType())
-            .castToSimpleBitVectorOrNull()) {
-      Type resultType = typeConverter->convertType(op.getInput().getType());
-      Value zero = rewriter.create<hw::ConstantOp>(op->getLoc(), resultType, 0);
-
-      rewriter.replaceOpWithNewOp<comb::ICmpOp>(op, comb::ICmpPredicate::ne,
-                                                adaptor.getInput(), zero);
-      return success();
-    }
+    Type resultType = typeConverter->convertType(op.getInput().getType());
```
maerhart, line 163: BoolCastOp does not guarantee that the input type is a simple bitvector, right? Thus `resultType` could be null and should be checked.

lib/Dialect/Moore/MooreTypes.cpp
```diff
@@ -197,143 +127,31 @@ namespace moore {
...
 struct IntTypeStorage : TypeStorage {
   using KeyTy = unsigned;
-  using Kind = IntType::Kind;
 
-  IntTypeStorage(KeyTy key) : kind(static_cast<Kind>(key)) {}
-  static KeyTy pack(Kind kind) { return static_cast<unsigned>(kind); }
-  bool operator==(const KeyTy &key) const { return pack(kind) == key; }
+  IntTypeStorage(KeyTy key)
+      : width(key >> 1), domain(static_cast<Domain>((key >> 0) & 1)) {}
+  static KeyTy pack(unsigned width, Domain domain) {
```
maerhart, line 133: What if `width = UINT_MAX`?
Should we at least assert this here to not silently drop a bit?

lib/Dialect/Moore/MooreTypes.cpp
```diff
@@ -197,143 +127,31 @@ namespace moore {
 namespace detail {
 struct IntTypeStorage : TypeStorage {
   using KeyTy = unsigned;
-  using Kind = IntType::Kind;
 
-  IntTypeStorage(KeyTy key) : kind(static_cast<Kind>(key)) {}
-  static KeyTy pack(Kind kind) { return static_cast<unsigned>(kind); }
-  bool operator==(const KeyTy &key) const { return pack(kind) == key; }
+  IntTypeStorage(KeyTy key)
+      : width(key >> 1), domain(static_cast<Domain>((key >> 0) & 1)) {}
```
hailongSun2000, line 132: May I ask what the role played by the `>>` and `<<`?   Just my personal question, I searched on Groq and ChatGPT, but I still don't understand it. Thanks! :smile:

test/Dialect/Moore/types.mlir
```diff
@@ -46,34 +32,34 @@ func.func @RealTypes(
...
+  // CHECK-SAME: %arg6: !moore.unpacked<assoc<i1, string>>,
+  // CHECK-SAME: %arg7: !moore.unpacked<queue<i1>>,
+  // CHECK-SAME: %arg8: !moore.unpacked<queue<i1, 9001>>
+  %arg2: !moore.unpacked<unsized<i1>>,
+  %arg3: !moore.unpacked<array<i1, 42>>,
+  %arg4: !moore.unpacked<range<i1, 4:-5>>,
+  %arg5: !moore.unpacked<assoc<i1>>,
+  %arg6: !moore.unpacked<assoc<i1, string>>,
+  %arg7: !moore.unpacked<queue<i1>>,
+  %arg8: !moore.unpacked<queue<i1, 9001>>
```
maerhart, line 52: Nit: the mnemonic `i1` could be a bit confusing since it could be the built-in integer type or the Moore type here if you don't know what the parser does.

unittests/Dialect/Moore/TypesTest.cpp
```diff
@@ -68,23 +68,14 @@ TEST(TypesTest, PackedInt) {
...
-      {IntType::Time, "time", Domain::TwoValued, Sign::Unsigned},
-  };
-
-  for (auto pair : pairs) {
-    auto kind = std::get<0>(pair);
-    auto type = IntType::get(&context, kind);
-    ASSERT_EQ(type.getDomain(), std::get<2>(pair));
-  }
+  auto ui42 = IntType::getInt(&context, 42);
+  auto ul42 = IntType::getLogic(&context, 42);
```
maerhart, line 72: Nit: Maybe calling it `SBVType` or similar might be less confusing since there is an `int` type in SystemVerilog, and this type is not a 1-1 correspondence to that?

maerhart (commented):
Let's say I have a `!moore.packed<range<i1, 4:0>>`. Would that be equivalent to `!moore.i5`?
Both `bit` and `bit [0:0]` correspond to `!moore.i1`, right? But is `bit [0:0][1:0]` considered a simple bitvector? Is there some canonicalization necessary or does Slang already canonicalize this?

lib/Dialect/Moore/MooreTypes.cpp
fabianschuiki: The width and domain are packed into a single unsigned int: "domain" is in the lowest bit, and "width" in the upper 31 bits (the `pack` function does that). This unpacks the unsigned int again: width is set to bits 31:1 of the key, and domain is set to bit 0 of the key.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Oh yeah great catch, thanks!

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Yes you're absolutely right, but `context.convertType()` also converts types like `string` into `!moore.string`. The constant op requires the type to be an `IntType` though (a simple bit vector). Slang guarantees that the type of the int literal is always an `IntType`, but `convertType()` doesn't know that.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -61,25 +62,31 @@ def UnpackedType : MooreType<CPred<"llvm::isa<moore::UnpackedType>($_self)">,
...
+    IntType::getInt($_builder.getContext(), 1)
   }];
 }
 
+class ResultIsSingleBitMatchingInputDomain<string result, string input> :
+  TypesMatchWith<"result is single bit matching input domain",
+    input, result, [{
+    IntType::get($_self.getContext(), 1,
+      llvm::cast<UnpackedType>($_self).getDomain())
+  }]>;
```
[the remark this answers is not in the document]
fabianschuiki: Good point. Probably should :grimacing:

lib/Dialect/Moore/MooreTypes.cpp
hailongSun2000: I understand it. Because of `pack` function. Thanks!!! 😃

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: I get it.

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Yes you're totally right. This should keep the old check.

lib/Dialect/Moore/MooreTypes.cpp
fabianschuiki: Good point

test/Dialect/Moore/types.mlir
fabianschuiki: Yeah that is very true. I was thinking of using `b1` and `l1` to distinguish this from the builtin integers. On the other hand, we might want to eventually replace `IntType` with a `LogicType` for `l42` and use the builtin integers for `i42`, so the `i1` might be easier to update in that case.

unittests/Dialect/Moore/TypesTest.cpp
fabianschuiki: Hmm yeah. SystemVerilog is very inconsistent. It calls simple bit vectors "integral" data types, or sometimes also "integers" exchangeably. And then there's the weird integer atoms which can be the `int` or `integer` you mentioned. My thinking here was that these are actually all integer data types, and the standard is just very clunky and beats around the bush with the simple bit vector types. Inside the dialect and the IR, it would be great to never have to think about the weird stuff that is in the standard, and just think of these values as integers.

fabianschuiki:
> Let's say I have a `!moore.packed<range<i1, 4:0>>`. Would that be equivalent to `!moore.i5`? Both `bit` and `bit [0:0]` correspond to `!moore.i1`, right? But is `bit [0:0][1:0]` considered a simple bitvector? Is there some canonicalization necessary or does Slang already canonicalize this?

With this PR, ImportVerilog maps things that the standard describes as "simple bit vectors" to the Moore integer data type. `int -> i32`, `integer -> l32`, `bit [41:0] -> i42`, `logic [41:0] -> l42`. It also canonicalizes `bit` to `i1` and `bit [0:0]` to `i1`. This is only done for a single packed dimension though, so `bit [3:0][4:0]` becomes `range<i5, 3:0>`, and `bit [1:0][0:0]` becomes `range<i1, 1:0>`. It also maps bit vectors that don't strictly start at 0 to the integer data type, so `bit [4:3]` maps to `i2` even though it's not strictly a simple bit vector type. Canonicalizing these range offsets away during the lowering is going to make working with the IR a lot easier.

test/Dialect/Moore/types.mlir
fabianschuiki: Any preferences @maerhart :grin: ?

maerhart: If the plan is to replace the two-valued Moore IntType with the built-in, I think keeping it as `i1` for now makes a lot of sense. 🙂

unittests/Dialect/Moore/TypesTest.cpp
maerhart: I see. In that case, the name is perfectly fine. I wish the SV standard was written less confusingly 😆

maerhart (approved):
Thanks for the clarifications, makes sense to me now!

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -16,6 +16,7 @@
 include "circt/Dialect/Moore/MooreDialect.td"
 include "mlir/IR/AttrTypeBase.td"
 include "mlir/IR/EnumAttr.td"
+include "mlir/IR/OpBase.td"
```
maerhart, line 19: This can probably be removed now?

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -160,16 +160,13 @@ struct BoolCastOpConversion : public OpConversionPattern<BoolCastOp> {
   LogicalResult
   matchAndRewrite(BoolCastOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    if (cast<UnpackedType>(op.getInput().getType())
-            .castToSimpleBitVectorOrNull()) {
-      Type resultType = typeConverter->convertType(op.getInput().getType());
+    Type resultType = typeConverter->convertType(op.getInput().getType());
+    if (isa<IntegerType>(resultType)) {
```
maerhart, line 164: I think this can fail if `resultType` is the null Type, should be `isa_or_null`, right?

unittests/Dialect/Moore/TypesTest.cpp
fabianschuiki: Ugh yeah :cry:

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Oh yeah good point. Let me add that.

PR #7012, opened 2024-05-09
[docs] Add basic pass tutorial

dobios:
Getting Started contains a decent amount of info but lacks a good tutorial for people who want to quickly learn how to hack together a working CIRCT pass. This PR adds a really basic tutorial that hopefully does just that.

I also added the `-DLLVM_ENABLE_LLD=true` flag to the cmake tutorial, because everyone should be using lld, it reduces the memory usage during linking by orders of magnitude. 

Let me know what you all think!


teqdruid (approved):
Nice! Thanks for doing this. You can separate out the LLD bits and just push them to main.

You should also point out that a more complete tutorial about all aspects of MLIR design (CIRCT just being a set of MLIR dialects and passes) is available in the MLIR repo/documentation. In general, we need to plaster these types of comments all over the documentation.

There's nothing worse than an outdated tutorial, so as a future improvement we should separate out the pass, put the tutorial code in its own directory, and write tests for it.

docs/GettingStarted.md
```diff
@@ -269,4 +276,81 @@ Patches are submitted to LLVM/MLIR via GitHub pull-requests, the basic flow is a
...
+  ```cpp
+  #include "circt/Dialect/HW/HWVisitors.h" // defines dispatchTypeOpVisitor which calls visit(op)
+
+  void visit(hw::WireOp op) { /*...*/ }
+  void visit(/*other hw operations*/) { /*...*/ }
+
+  void FooWires::runOnOperation() {
+    module.walk([&](Operation* op) { dispatchTypeOpVisitor(op); });
+  }
+  ```
```
youngar, line 335: Example here seems incomplete, need to define a struct that CRTP's the visitor.  Should probably also include the invalid operation handler. Also I don't think `hw::WireOp` is actually handled by any visitor AFAICT, although it probably should be handled by this visitor.

docs/GettingStarted.md
```diff
@@ -269,4 +276,81 @@ Patches are submitted to LLVM/MLIR via GitHub pull-requests, the basic flow is a
...
+  }];
+  let constructor = "circt::firrtl::createFooWiresPass()";
+  let statistics = [
+    Statistic<"numChangedWires", "num-changed-wires", "Number of wires changed">
+  ];
+}
+```    
+Once this is added, compile CIRCT. This will generate a base class for your pass following the naming
+defined in the tablegen description. This base class will contain some methods that need to be implemented.
+Your goal is to now implement those.
```
youngar, line 301: maybe add a link to https://mlir.llvm.org/docs/PassManagement/#tablegen-specificationd

docs/GettingStarted.md
```diff
@@ -269,4 +276,81 @@ Patches are submitted to LLVM/MLIR via GitHub pull-requests, the basic flow is a
...
+  ```cpp
+  void FooWires::runOnOperation() {
+    size_t nWires = 0; // Counts the number of wires modified
+    module.walk([&](hw::WireOp wire) { // Walk over every wire in the module
+      wire.setName("foo_" + std::to_string(nWires++)); // Rename said wire
+    });
+  }
+  ``` 
+  Here `module.walk([&](WireOp wire) { ... });` is used to traverse every wire in the design, you can also
+  use it to generically walk over all operations and then visit them individually using a [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) by overloading type visitors defined by each dialect, e.g.
```
youngar, line 325: Not sure the audience of the tutorial, but they might find this link useful: https://mlir.llvm.org/docs/Tutorials/UnderstandingTheIRStructure/

docs/GettingStarted.md
```diff
@@ -269,4 +276,81 @@ Patches are submitted to LLVM/MLIR via GitHub pull-requests, the basic flow is a
...
+3) Define the pass constructor in the dialect's `Passes.h` file, e.g. in `include/circt/Dialect/FIRRTL/Passes.h` add:  
+```cpp
+std::unique_ptr<mlir::Pass> createFooWiresPass();
+```
+
+Now re-run cmake and compile CIRCT and you should be able to run your new pass! 
+By default, this will be accessible using `circt-opt` through the name that was defined in the tablegen description, 
+e.g. in our case the pass can be called by running `circt-opt --firrtl-foo-wires <file_name>.mlir`. 
+These passes can also be included in certain compilation pipelines that well packaged in tools like firtool.  
+We recommend looking at recently closed Pull-Requests that add passes to see more examples of how this can be done beyond our basic pass.
```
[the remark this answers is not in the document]
youngar: Should include how to print the pass statistics, since the example has one.

docs/GettingStarted.md
```diff
@@ -66,7 +69,8 @@ $ cmake -G Ninja ../llvm \
     -DLLVM_ENABLE_PROJECTS="mlir" \
     -DLLVM_TARGETS_TO_BUILD="X86;RISCV" \
     -DLLVM_ENABLE_ASSERTIONS=ON \
-    -DCMAKE_BUILD_TYPE=DEBUG
+    -DCMAKE_BUILD_TYPE=DEBUG \
```
uenoku, line 72: I'm curios if we can also recommend RelWithDebInfo by default. I believe many of us use RelWithDebInfo normally. It would make linking so much easier (probably it's feasible to compile CIRCT on laptop even without lld).

docs/GettingStarted.md
```diff
@@ -81,11 +85,14 @@ $ cmake -G Ninja .. \
     -DMLIR_DIR=$PWD/../llvm/build/lib/cmake/mlir \
     -DLLVM_DIR=$PWD/../llvm/build/lib/cmake/llvm \
     -DLLVM_ENABLE_ASSERTIONS=ON \
-    -DCMAKE_BUILD_TYPE=DEBUG
+    -DCMAKE_BUILD_TYPE=DEBUG \
+    -DLLVM_ENABLE_LLD=true
```
uenoku, line 89: nit: I think `-DLLVM_ENABLE_LLD=ON` is standard.

docs/GettingStarted.md
seldridge: My experience is that I have to keep both builds around. It really depends on if you plan/need to use a lldb/gdb. I find I can't get much done with `RelWithDebInfo` in such an environment. However, `RelWithDebInfo` is almost always sufficient. I think suggesting `RelWithDebInfo` is probably fine.

docs/GettingStarted.md
```diff
@@ -269,4 +276,81 @@ Patches are submitted to LLVM/MLIR via GitHub pull-requests, the basic flow is a
...
+
+Now re-run cmake and compile CIRCT and you should be able to run your new pass! 
+By default, this will be accessible using `circt-opt` through the name that was defined in the tablegen description, 
+e.g. in our case the pass can be called by running `circt-opt --firrtl-foo-wires <file_name>.mlir`. 
+These passes can also be included in certain compilation pipelines that well packaged in tools like firtool.  
+We recommend looking at recently closed Pull-Requests that add passes to see more examples of how this can be done beyond our basic pass.  
+
+
+
+
```
seldridge, line 356: Mega-nit: trailing newlines.

docs/GettingStarted.md
```diff
@@ -66,7 +69,8 @@ $ cmake -G Ninja ../llvm \
     -DLLVM_ENABLE_PROJECTS="mlir" \
     -DLLVM_TARGETS_TO_BUILD="X86;RISCV" \
```
seldridge, line 70: Nit: this should probably just be "host".

docs/GettingStarted.md
dobios: yes you're right, I'll extract this all into a pass file that actually works and compiles with tests and then reference that here to avoid not specifying details like that.

docs/GettingStarted.md
dobios: Compiler engineers that don't know MLIR but want to get a quick and dirty start in CIRCT so that they can easily test their ideas out (i.e. me about a year ago lol). I'll add that link plus a few others scattered around so that people also have access to more detailed tutorials if they want them.

docs/GettingStarted.md
dobios: not sure what you mean by this

docs/GettingStarted.md
```diff
@@ -269,4 +276,81 @@ Patches are submitted to LLVM/MLIR via GitHub pull-requests, the basic flow is a
...
-If you have commit access, you can do this yourself, otherwise a reviewer can do it for you.
+If you have commit access, you can do this yourself, otherwise a reviewer can do it for you.  
+  
+## Writing a basic CIRCT Pass  
+Passes can be added at several levels in CIRCT. Here we illustrate this with a simple pass, added 
+to the `firrtl` dialect, that replaces all of the wire names with `foo`: 
+1) Update your dialect’s Passes.td file to define what your pass will do in a high-level, well documented way, 
+e.g. in `include/circt/Dialect/FIRRTL/Passes.td`:  
+```
+def FooWires : Pass<"firrtl-foo-wires", "firrtl::CircuitOp"> {
```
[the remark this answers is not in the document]
dobios: My bad, I think I originally wrote this for hw and then switched to firrtl for some reason. I'll clean everything up in a external file that actually runs to avoid brain farts like this lol

dobios:
https://github.com/llvm/circt/pull/7018 Let me know if you all agree with the flags I added.

docs/GettingStarted.md
teqdruid: `s/X86;RISCV/host/`

dobios:
Better?

docs/GettingStarted.md
```diff
@@ -67,7 +67,7 @@ $ mkdir llvm/build
 $ cd llvm/build
 $ cmake -G Ninja ../llvm \
     -DLLVM_ENABLE_PROJECTS="mlir" \
-    -DLLVM_TARGETS_TO_BUILD="X86;RISCV" \
+    -DLLVM_TARGETS_TO_BUILD="s/X86;RISCV/host/" \
```
uenoku, line 70: maybe mistake?

uenoku (approved):

docs/GettingStarted.md
dobios: yeah substitution error on my end, should be just host

PR #7013, opened 2024-05-09
[HW][LegalizeModules] Avoid segmentation fault

hovind:
Separated from https://github.com/llvm/circt/pull/6996 as proposed by @uenoku, please merge at your convenience.


uenoku (approved):
Thanks!

PR #7014, opened 2024-05-09
 [AddSeqMemPorts] Add hierpathop to verbatim, instead of raw instance path 

prithayan:
The annotation `AddSeqMemPortsFileAnnotation` causes the pass
 `AddSeqMemPort` to add verbatim ops for the SRAM metadata file.
The file lists each SRAM and provides the mapping to
where it is in the hierarchy, and gives its IO prefix at the DUT top level.

This PR updates the pass to use `HierPathOp` to print the SRAM instance path,
 instead of the raw list of symbol references. 
This is required to ensure that the instance path is valid, even if the hierarchy
is updated by following passes.

This fixes a `firtool` crash that is exposed when using `AddSeqMem` along with
 `ExtractSeqMem`. Which is due to invalid symbol references in the `varbatim`, after
 the hierarchy was updated. Even though, this particular use case is invalid, as these
 two SRAM annotations cannot be used together, it can still be triggered by other
transformations.

Note: This metadata will soon be moved to OM, in a followup PR.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
@@ -16,6 +16,7 @@
 #include "circt/Dialect/FIRRTL/FIRRTLInstanceGraph.h"
 #include "circt/Dialect/FIRRTL/FIRRTLOps.h"
 #include "circt/Dialect/FIRRTL/FIRRTLUtils.h"
+#include "circt/Dialect/FIRRTL/Namespace.h"
 #include "circt/Dialect/FIRRTL/Passes.h"
 #include "circt/Dialect/HW/InnerSymbolNamespace.h"
 #include "circt/Dialect/SV/SVOps.h"
@@ -66,8 +67,11 @@ struct AddSeqMemPortsPass : public AddSeqMemPortsBase<AddSeqMemPortsPass> {
     /// which is for some reason the format of the metadata file.
     std::vector<SmallVector<Attribute>> instancePaths;
   };
+
+  CircuitNamespace circtNamespace;
   /// This maps a module to information about the memories.
   DenseMap<Operation *, MemoryInfo> memInfoMap;
+  DenseMap<Attribute, Operation *> innerRefToInstanceMap;
 
   InstanceGraph *instanceGraph;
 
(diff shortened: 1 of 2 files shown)


test/Dialect/FIRRTL/add-seqmem-ports.mlir
```diff
@@ -176,8 +186,8 @@ firrtl.circuit "Complex" attributes {annotations = [
   // CHECK:               emit.file "metadata{{/|\\\\}}sram.txt" {
   // CHECK-NEXT:            sv.verbatim
   // CHECK-SAME{LITERAL}:     0 -> {{0}}.{{1}}
-  // CHECK-SAME{LITERAL}:     1 -> {{0}}.{{2}}.{{3}}
-  // CHECK-SAME{LITERAL}:     2 -> {{0}}.{{4}}
-  // CHECK-SAME:              {symbols = [@DUT, #hw.innerNameRef<@DUT::@[[MWRITE_EXT]]>, #hw.innerNameRef<@DUT::@[[CHILD]]>, #hw.innerNameRef<@Child::@[[CHILD_MWRITE_EXT]]>, #hw.innerNameRef<@DUT::@[[MWRITE_EXT_0]]>]}
+  // CHECK-SAME{LITERAL}:     1 -> {{0}}.{{2}}
+  // CHECK-SAME{LITERAL}:     2 -> {{0}}.{{3}}
+  // CHECK-SAME:              {symbols = [@DUT, @memNLA, @memNLA_0, @memNLA_1]}
```
fabianschuiki, line 191: Nice, a lot cleaner :100:

fabianschuiki (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
```diff
@@ -308,22 +313,43 @@ void AddSeqMemPortsPass::createOutputFile(igraph::ModuleOpInterface module) {
...
+      // There is no current client for the distinct attr, but it will be used
+      // by OM::path once the metadata is moved to OM, instead of the verbatim.
+      auto id = DistinctAttr::create(UnitAttr::get(builder.getContext()));
+      fields.append("id", id);
+      fields.append("class", builder.getStringAttr("circt.tracker"));
+      fields.append("circt.nonlocal", nlaSymbol);
+      // Now add the nonlocal annotation to the leaf instance.
+      auto *leafInstance = innerRefToInstanceMap[instancePath.front()];
+      // TODO: Check if we can reuse HierPathOp, if it already exists!
+      AnnotationSet annos(leafInstance);
```
dtzSiFive, line 347: Checkout `HierPathCache` from `FIRRTLAnnotationHelper.h`! 😁

prithayan: Awesome, thanks I missed `HierPathCache`. Updated the PR to use the cache.

PR #4856, opened 2023-03-20
[SV] Added DPI import and call ops

nandor:
This PR introduces DPI import and call operations.

To import a function, the following can be declared at the module level:

```sv.dpi.import @test(%arg0: i32) -> (res0: i5, res1: i6)```

This is hard-wire to contextual functions.

To call a function, an op can be added to procedural scopes.

```%res0, %res1 = sv.dpi.call @test(%arg0) : (i32) -> (i5, i6)```

The emission is similar to instances. Temporary registers will be added to capture the results of DPI calls.

In the future, constraints should be relaxed to allow for return types and to place calls in arbitrary contexts.

PR #7015, opened 2024-05-09
[ExportVerilog] Support sv.func.* op emission

uenoku:
This PR (separated from https://github.com/llvm/circt/pull/6977)  implements ExportVerilog support of sv.func, sv.func.dpi.import, sv.func.call and sv.func.call.procedural. 

sv.func emission is similar to hw.module but one difference is a return value.  Surprisingly in SV a function name is used as a placeholder for a return value so name legalization properly sets `hw.verilogName` of a returned value to a function name. 

Result names of sv.func.call.procedural are attached to temporary registers created in PrepareForEmission. This is similar implementation approach as hw.instance. 

A part of PrepareForEmission and ExportVerilog is ported from https://github.com/llvm/circt/pull/4856


lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -3920,9 +3939,9 @@ StmtEmitter::emitAssignLike(Op op, PPExtString syntax,
 }
 
 LogicalResult StmtEmitter::visitSV(AssignOp op) {
-  // prepare assigns wires to instance outputs, but these are logically handled
-  // in the port binding list when outputing an instance.
-  if (dyn_cast_or_null<HWInstanceLike>(op.getSrc().getDefiningOp()))
+  // prepare assigns wires to instance outputs and function results, but these
+  // are logically handled in the port binding list when outputing an instance.
+  if (isa_and_nonnull<HWInstanceLike, FuncCallOp>(op.getSrc().getDefiningOp()))
```
dobios, line 3944: nit: Is the second type argument really necessary?

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4129,6 +4158,109 @@ LogicalResult StmtEmitter::visitStmt(TypedeclOp op) {
...
+
+  ps << PPExtString(getSymOpName(callee)) << "(";
+
+  bool needsComma = false;
+  auto printArg = [&](Value value) {
+    if (needsComma)
+      ps << "," << PP::space;
+    emitExpression(value, ops);
+    needsComma = true;
+  };
```
dobios, line 4195: won't this lead to a trailing comma? Could you maybe do something using the argument index instead to figure out if you need a comma or not?

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4129,6 +4158,109 @@ LogicalResult StmtEmitter::visitStmt(TypedeclOp op) {
...
+  startStatement();
+
+  ps << "import" << PP::nbsp << "\"DPI-C\"" << PP::nbsp;
+
+  // Emit a linkage name if provided.
+  if (auto linkageName = importOp.getLinkageName())
+    ps << *linkageName << PP::nbsp << "=" << PP::nbsp;
+  auto op =
+      cast<FuncOp>(state.symbolCache.getDefinition(importOp.getCalleeAttr()));
+  assert(op.isDeclaration() && "function must be a declaration");
```
dobios, line 4256: I don't think I've seen this way of adding assert messages before lol

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6059,6 +6191,24 @@ void ModuleEmitter::emitHWModule(HWModuleOp module) {
...
+void ModuleEmitter::emitFunc(FuncOp func) {
+  // Nothing to emit for a declaration.
+  if (func.isDeclaration())
+    return;
+
+  currentModuleOp = func;
+  startStatement();
+  ps.addCallback({func, true});
+  // A function is moduled as an automatic function.
+  emitFunctionSignature(*this, ps, func, /*isAutomatic=*/true);
```
dobios, line 6203: nit: commented code

dobios (approved):
Not really an expert on this, but apart from a few nits, the fact that you should update this wrt. the llvm bump fixes and llvm itself to make sure that it won't break on merge, and clang-tidy, LGTM.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
uenoku: Yes, this is per LLVM coding style: https://llvm.org/docs/CodingStandards.html#assert-liberally

lib/Conversion/ExportVerilog/ExportVerilog.cpp
uenoku: Adding an argument name as a comment is encouraged for boolean parameter: https://llvm.org/docs/CodingStandards.html#comment-formatting

lib/Conversion/ExportVerilog/ExportVerilog.cpp
uenoku: Yes, it's necessary. FuncCall op emissions follows a similar emission implementation to HWInstanceOp. The assignments are emitted in FuncCallOp emission so we need to skip here.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
uenoku: Good question! I think it will not produce training comma since a comma is always emitted before an operand.

lib/Conversion/ExportVerilog/PrepareForEmission.cpp
```diff
@@ -1269,6 +1336,7 @@ struct PrepareForEmissionPass
   void runOnOperation() override {
     auto module = getOperation();
     LoweringOptions options(cast<mlir::ModuleOp>(module->getParentOp()));
+    module.dump();
```
dtzSiFive, line 1339:
```suggestion
```

uenoku: Thanks! :bow:

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4262,6 +4291,109 @@ LogicalResult StmtEmitter::visitStmt(TypedeclOp op) {
...
+    assert(explicitReturn.hasOneUse());
+    if (op->getParentOp()->template hasTrait<ProceduralRegion>()) {
+      auto bpassignOp = cast<sv::BPAssignOp>(*explicitReturn.user_begin());
+      emitExpression(bpassignOp.getDest(), ops);
+    } else {
+      auto assignOp = cast<sv::AssignOp>(*explicitReturn.user_begin());
+      emitExpression(assignOp.getDest(), ops);
+    }
+    ps << PP::nbsp << "=" << PP::nbsp;
+  }
```
fabianschuiki, line 4316: This feels like it would produce `x = foo()` in a module body instead of `assign x = foo()` :thinking:

test/Conversion/ExportVerilog/sv-dialect.mlir
```diff
@@ -1670,6 +1670,98 @@ hw.module @IndexPartSelect(out a : i3) {
...
+
+// CHECK-LABEL: func_call
+hw.module @func_call(in %in_0 : i2, in %in_1 : i2, in %in_2: i1, out out: i1) {
+  %fd = hw.constant 0x80000002 : i32
+  // CHECK:      wire        _function_declare2_0;
+  // CHECK-NEXT: logic [1:0] _function_declare3_0;
+  // CHECK-NEXT: logic       _function_declare3_1;
+  // CHECK-NEXT: logic [1:0] _function_declare1_0;
+  // CHECK-NEXT: logic       _function_declare1_1;
+  // CHECK-NEXT: _function_declare2_0 = function_declare2(in_0, in_1);
```
fabianschuiki, line 1686: I think this has to be
```
assign _function_declare2_0 = function_declare2(in_0, in_1);
```

lib/Conversion/ExportVerilog/ExportVerilog.cpp
uenoku: Good point, fixed!

test/Conversion/ExportVerilog/sv-dialect.mlir
uenoku: Fixed by https://github.com/llvm/circt/pull/7015/commits/68a2872706733514725beefdcd172300a53a1559. Thank you for finding this!

uenoku:
I'd like to merge. Post-merge review is also appreciated!

PR #7016, opened 2024-05-09
[FIRRTL] Don't prefix an empty label for unclocked assume.

dtzSiFive:
Add FIR -> SV test for "unclocked_assert" since we care about ensuring it has a specific shape re:output.

Test extracted from #7010.


uenoku (approved):
LGTM

PR #7020, opened 2024-05-10
[OM] Add IsolatedFromAbove to OMClass

uenoku:
This adds `IsolatedFromAbove` to OMClassOp and OMClassExternOp. It should make local verification run parallelly.


dtzSiFive (approved):
LGTM, thanks!

(test fix looks like just need to update source location expected)

uenoku:
@fabianschuiki thank you for fixing typos ;)

PR #7021, opened 2024-05-10
[FIRRTL] Add condition expansion for ExpandWhens on Property intrinsics

dobios:
This PR corrects when expansions for `[Assert, Assume, Cover]Poperty` by creating an `implies` statement between the condition of the when and the property itself.
Basically in the case of the following:
```
when(cond): 
    AssertProperty(prop)
```
is currently being expanded to simply  
```
AssertProperty(prop)
```
which is wrong, so this PR makes so that it gets expanded to
```
AssertProperty(cond implies prop)
```

### DIFF
--- a/test/Dialect/FIRRTL/expand-whens.mlir
+++ b/test/Dialect/FIRRTL/expand-whens.mlir
@@ -636,4 +636,40 @@ firrtl.module @ModuleWithObjectWire(in %in: !firrtl.class<@ClassWithInput(in in:
   firrtl.propassign %0, %in : !firrtl.class<@ClassWithInput(in in: !firrtl.string)>
 }
 
+// Test all verif assert/assume/cover constructs
+// CHECK:   firrtl.module @verif(in %clock: !firrtl.clock, in %cond: !firrtl.uint<1>, in %prop: !firrtl.uint<1>, in %enable: !firrtl.uint<1>, in %reset: !firrtl.uint<1>) {
+firrtl.module @verif(
+  in %clock : !firrtl.clock, in %cond : !firrtl.uint<1>, in %prop : !firrtl.uint<1>,
+  in %enable : !firrtl.uint<1>, in %reset : !firrtl.uint<1>
+) {
+  firrtl.when %cond : !firrtl.uint<1> {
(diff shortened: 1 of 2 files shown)


test/Dialect/FIRRTL/expand-whens.mlir
```diff
@@ -636,4 +636,40 @@ firrtl.module @ModuleWithObjectWire(in %in: !firrtl.class<@ClassWithInput(in in:
...
+// Test all verif assert/assume/cover constructs
+// CHECK:   firrtl.module @verif(in %clock: !firrtl.clock, in %cond: !firrtl.uint<1>, in %prop: !firrtl.uint<1>, in %enable: !firrtl.uint<1>, in %reset: !firrtl.uint<1>) {
+firrtl.module @verif(
+  in %clock : !firrtl.clock, in %cond : !firrtl.uint<1>, in %prop : !firrtl.uint<1>,
+  in %enable : !firrtl.uint<1>, in %reset : !firrtl.uint<1>
+) {
+  firrtl.when %cond : !firrtl.uint<1> {
+    // CHECK:   %0 = firrtl.int.ltl.not %cond : (!firrtl.uint<1>) -> !firrtl.uint<1>
+    // CHECK:   %1 = firrtl.int.ltl.or %0, %prop : (!firrtl.uint<1>, !firrtl.uint<1>) -> !firrtl.uint<1>
+    // CHECK:   firrtl.int.verif.assert %1 : !firrtl.uint<1>
```
fabianschuiki, line 648: It's usually a good idea to use placeholders for SSA value names, as that makes the tests more robust to dumb things like the MLIR printer suddenly picking different names for values. For example:
```suggestion
    // CHECK:   [[TMP0:%.+]] = firrtl.int.ltl.not %cond : (!firrtl.uint<1>) -> !firrtl.uint<1>
    // CHECK:   [[TMP1:%.+]] = firrtl.int.ltl.or [[TMP0]], %prop : (!firrtl.uint<1>, !firrtl.uint<1>) -> !firrtl.uint<1>
    // CHECK:   firrtl.int.verif.assert [[TMP1]] : !firrtl.uint<1>
```

fabianschuiki (approved):
LGTM! Thanks for fixing this :partying_face:

Issue #6982, opened 2024-05-02
[Verif] Introduce a clocked assertion `verif.clocked_assert`


dobios:
The current structure of the `verif` dialect requires the use of `ltl.clock` to associate a clock to an assertion (which is required). This means that a use-def analysis is required everytime we want to work with `verif.assert`, which leads to the `sv.assert` operation being used in most lowerings to Core dialects.

As suggested by @fabianschuiki , I propose that we add a `verif.clocked_assert(clock, prop)` that would be the canonicalization of `verif.assert(ltl.clock(clock, prop))`. This would make it easier to use move away from `sv` dialect ops in the core representations.

fabianschuiki:
You'd probably want to flip the canonicalization around: canonicalize `verif.assert(ltl.clock(clock, prop))` to `verif.clocked_assert(clock, prop)`, in order to allow passes to only work with `verif.clocked_assert`.

maerhart:
I assume the `verif.clocked_assert` will also verify that no (other) `ltl.clock` operation is present in the use-def chain of the `prop` operand?

fabianschuiki:
That's an excellent point. I like the idea of having `clocked_assert` enforce that the operand itself contains no clocking or enable/disable information. The regular `assert` could then have a canonicalizer which pulls single unambiguous clocking and enable/disable signals up into a `clocked_assert` where possible.

PR #7022, opened 2024-05-10
[Verif]  Add clocked Assert Assume Cover ops

dobios:
This PR is in response to #6982 and adds clocked versions of assert, assume, cover. 
What still needs to be done:  
- [x] Update ExportVerilog  
- [x] Add Verifier that checks that we have exactly one disable and clock.
- [x] Add tests


include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -43,6 +47,43 @@ def CoverOp : AssertLikeOp<"cover"> {
...
+def AtPosEdge: I32EnumAttrCase<"Pos", 0, "posedge">;
+/// AtNegEdge triggers on a drop from 1 to 0/X/Z, or X/Z to 0.
+def AtNegEdge: I32EnumAttrCase<"Neg", 1, "negedge">;
+/// AtEdge is syntactic sugar for AtPosEdge or AtNegEdge.
+def AtEdge   : I32EnumAttrCase<"Both", 2, "edge">;
+
+def ClockEdgeAttr : I32EnumAttr<"ClockEdge", "clock edge",
+                                [AtPosEdge, AtNegEdge, AtEdge]> {
+  let cppNamespace = "circt::verif";
+}
```
maerhart, line 62: AFAIK the HW dialect has the same enum (as do LTL and FIRRTL). Is there anything hindering us from just having such an attribute defined once in the HW dialect and using it for all other dialects? That way we wouldn't need to always have those helper functions just converting one enum to the other 1-1.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -43,6 +47,43 @@ def CoverOp : AssertLikeOp<"cover"> {
...
+                      I1:$disable,
+                      ClockEdgeAttr:$edge, I1:$clock,
+                      OptionalAttr<StrAttr>:$label);
+  let assemblyFormat = [{
+    $property `disable` $disable `clock` $edge $clock (`label` $label^)? attr-dict `:` type($property)
+  }];
+}
+
+def ClockedAssertOp : ClockedAssertLikeOp<"clocked_assert"> {
+  let summary = "Assert that a property holds, checked on a given clock's ticks and disabled if a given condition holds.";
```
maerhart, line 76: Nit: 80 cols pls

include/circt/Dialect/Verif/VerifOps.td
dobios: @fabianschuiki recommended redefining the enum to avoid having a dialect dependency (the current pattern being used)

fabianschuiki: Yeah I think the last time we tried to unify this, things didn't go quite as planned because we ended up pulling in other stuff from the dialect and regenerating implementations for other types. I don't know if there's a way to reuse these enum attributes from a different dialect without also generating conflicting declarations/definitions for them.

maerhart: I see. I don't remember when the last time was that we tried to unify this, but there were some quite significant changes related to this in MLIR quite a while ago, so maybe it works now. I'd be quite surprised if sharing attributes isn't possible tbh. But just copy-pasting this single attribute is also a good approach for now, given the uncertainty of the unification approach.

fabianschuiki: I created a "good first issue" for this: https://github.com/llvm/circt/issues/7034

lib/Dialect/Verif/VerifOps.cpp
```diff
@@ -56,6 +60,80 @@ LogicalResult LogicEquivalenceCheckingOp::verifyRegions() {
...
+    // Keeps track of operations that have been visited
+    llvm::DenseSet<Operation *> handledOps;
+
+    Operation *property = clockedAssertLikeOp->getOperand(0).getDefiningOp();
+
+    // Fill in our worklist
+    worklist.insert({property, property->operand_begin()});
+
+    // Process the elements in our worklist
+    while (!worklist.empty()) {
```
uenoku, line 87: Verifier shouldn't follow use-def chain(please see https://mlir.llvm.org/getting_started/DeveloperGuide/#ir-verifier). It will eventually blow up compilation time. I don't have a good alternative solution other than separating the verification into a pass. I think we can make the verification local if types contain clock domain information but probably it's not worth doing that.

lib/Dialect/Verif/VerifOps.cpp
```diff
@@ -16,6 +17,9 @@
 #include "mlir/IR/SymbolTable.h"
 #include "mlir/Interfaces/FunctionImplementation.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
+#include "llvm/ADT/MapVector.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/raw_ostream.h"
```
uenoku, line 22: nit: sees unnecessary?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -43,6 +47,57 @@ def CoverOp : AssertLikeOp<"cover"> {
...
+def AtNegEdge: I32EnumAttrCase<"Neg", 1, "negedge">;
+/// AtEdge is syntactic sugar for AtPosEdge or AtNegEdge.
+def AtEdge   : I32EnumAttrCase<"Both", 2, "edge">;
+
+def ClockEdgeAttr : I32EnumAttr<"ClockEdge", "clock edge",
+                                [AtPosEdge, AtNegEdge, AtEdge]> {
+  let cppNamespace = "circt::verif";
+}
+
+class ClockedAssertLikeOp<string mnemonic, list<Trait> traits = []> :
```
uenoku, line 64: Do we want to allow clocked assert in procedural regions? I think an assumption that clocked_assert op never be in always statement or if-statement would make the semantics a lot clear? I don't have a strong opinion tough :)

fabianschuiki: Yeah great point. Since these ops define their own clock and enable, they are intended to be used in a non-procedural region.

dobios:
@uenoku @fabianschuiki I moved the verification into a dedicated pass, now the question is: when should this pass be called? Just before emission? At the end of the hw pass pipeline?

include/circt/Dialect/Verif/Passes.td
```diff
@@ -0,0 +1,32 @@
...
+ */
+def VerifyClockedAssertLike : Pass<"verif-verify-clocked-assert-like"> {
+  let summary = "Check that clocked-assert-like are valid.";
+  let description = [{
+    Checks that every `clocked_assert`, `clocked_assume` or `clocked_cover` op does not contain 
+    a nested `ltl.clock` or `ltl.disable` operation in its operand tree.
+  }];
+  let constructor = "circt::verif::createVerifyClockedAssertLikePass()";
+
+  let dependentDialects = ["hw::HWDialect"];
```
uenoku, line 29: nit: Probably unnecessary

include/circt/Dialect/Verif/Passes.td
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef CIRCT_DIALECT_VERIF_PASSES_TD
+#define CIRCT_DIALECT_VERIF_PASSES_TD
+
+include "mlir/Pass/PassBase.td"
+
+/**
+ * Checks the validity of clocked assertions, assumptions, and cover ops.
+ */
+def VerifyClockedAssertLike : Pass<"verif-verify-clocked-assert-like"> {
```
uenoku, line 21: nit: Can we make this HWModuleOp pass ?

fabianschuiki:
> when should this pass be called? Just before emission? At the end of the hw pass pipeline?

I would run this pass at least just before `ExportVerilog`, or before any pass that wants to rely on the invariant that clocked asserts contain no `ltl.clock` ops in the property/sequence that they're checking.

uenoku (commented):
Thank you for iterating on this! Separating verification into a pass makes sense to me. 

> @uenoku @fabianschuiki I moved the verification into a dedicated pass, now the question is: when should this pass be called? Just before emission? At the end of the hw pass pipeline?

I think we want to verify after IR constructed so probably after LowerToHW would be good place. It also makes sense to verify before ExportVerilog.

include/circt/Dialect/Verif/Passes.td
dobios: I'll try that out yes

include/circt/Dialect/Verif/Passes.td
```diff
@@ -0,0 +1,32 @@
+//===-- Passes.td - Verif pass definition file ----------------*- tablegen -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file defines the passes that work on the Verfi dialect.
```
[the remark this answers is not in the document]
dobios: me and my typos lol

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4616,17 +4672,15 @@ LogicalResult StmtEmitter::visitSV(CoverConcurrentOp op) {
...
   // an `assert` or `assert property`. If we are in a non-procedural region,
   // e.g., the body of a module, we have to use the concurrent form `assert
   // property` (which also supports plain booleans).
   //
-  // See IEEE 1800-2017 section 16.14.5 "Using concurrent assertion statements
-  // outside procedural code" and 16.14.6 "Embedding concurrent assertions in
-  // procedural code".
+  // See IEEE 1800-2017 section 16.14.5 "Using concurrent assertion
+  // statements outside procedural code" and 16.14.6 "Embedding concurrent
+  // assertions in procedural code".
```
fabianschuiki, line 4683: Stray changes?

lib/Dialect/Verif/Transforms/VerifyClockedAssertLike.cpp
```diff
@@ -0,0 +1,101 @@
...
+  void runOnOperation() override;
+
+private:
+  static void verify(Operation *clockedAssertLikeOp) {
+    // Used to perform a DFS search through the module to visit all operands
+    // before they are used
+    llvm::SmallMapVector<Operation *, OperandRange::iterator, 16> worklist;
+
+    // Keeps track of operations that have been visited
+    llvm::DenseSet<Operation *> handledOps;
```
fabianschuiki, line 45: You may want to pull this `handledOps` set out into the pass itself (and clear it at the beginning of the pass). If I understand correctly, this is the set of operations that contain no nested `ltl.clock` and `ltl.disable` ops. If a subsequence is used in multiple asserts or assumes, which may be a common thing if CSE does its job well, remembering which ops you've already deemed safe avoids re-visiting those common ops over and over again.

For example:
```
%0 = ltl.delay ...
verif.clocked_assert %0
verif.clocked_assert %0
verif.clocked_assert %0
```
would cause you to traverse the entire expression tree three times if you keep `handledOps` scoped locally in `verify`; but if you preserve it throughout the run of the pass, you'll only check the `%0` expression tree once.

test/Conversion/ExportVerilog/verif.mlir
```diff
@@ -298,3 +298,19 @@ hw.module @Issue5763(in %a: i3) {
   %2 = comb.and bin %1, %0 : i1
   verif.assert %2 : i1
 }
+
+
+// CHECK-LABEL: module ClokedAsserts
+hw.module @ClokedAsserts(in %clk: i1, in %a: i1, in %b: i1) {
```
fabianschuiki, line 304:
```suggestion
// CHECK-LABEL: module ClockedAsserts
hw.module @ClockedAsserts(in %clk: i1, in %a: i1, in %b: i1) {
```

lib/Conversion/ExportVerilog/ExportVerilog.cpp
dobios: probably clang-tidy

fabianschuiki: Probably worthwhile reverting these changes. Especially the check for `hasSVAttributes` should probably stay 😄

include/circt/Dialect/Verif/VerifVisitors.h
```diff
@@ -21,9 +21,11 @@ class Visitor {
...
     auto *thisCast = static_cast<ConcreteType *>(this);
     return TypeSwitch<Operation *, ResultType>(op)
-        .template Case<AssertOp, AssumeOp, CoverOp>([&](auto op) -> ResultType {
-          return thisCast->visitVerif(op, args...);
-        })
+        .template Case<AssertOp, AssumeOp, CoverOp, ClockedAssertOp,
+                       ClockedAssumeOp, ClockedCoverOp>(
+            [&](auto op) -> ResultType {
+              return thisCast->visitVerif(op, args...);
+            })
```
fabianschuiki, line 28: I think you also need corresponding `HANDLE(ClockedAssertOp, Unhandled);` lines a bit further down in this file.

test/Dialect/Verif/errors.mlir
```diff
@@ -19,3 +19,5 @@ verif.lec first {
 ^bb0(%arg0: i32):
   verif.yield %arg0 : i32
 }
+
+
```
fabianschuiki, line 23: Stray newlines

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4618,15 +4674,14 @@ LogicalResult StmtEmitter::emitVerifAssertLike(Operation *op, Value property,
...
```
fabianschuiki, line 4684: Stray changes? The newline after emitError seems useful, and the original wrapping of "statements" also seems sensible. Probably just clang-format that ran on an indented/commented version of the code. Might be worthwhile to just restore to the original state with `git restore -p`.

fabianschuiki (approved):
LGTM! Just a few nits on stray changes that you may want to revert/restore. Love the tests!

PR #7024, opened 2024-05-11
[SV] Add `sv.reserve_names` op to disallow names

teqdruid:
Introduce an op to specific a list of names which Export Verilog should never use anywhere. Applies to module names, port names, reg/wire/logic names, etc.

### DIFF
--- a/include/circt/Dialect/SV/SVStatements.td
+++ b/include/circt/Dialect/SV/SVStatements.td
@@ -879,3 +879,17 @@ def MacroDefOp : SVOp<"macro.def",
     MacroDeclOp getReferencedMacro(const hw::HWSymbolCache *cache);
   }];
 }
+
+
+//===----------------------------------------------------------------------===//
+// SV output control.
+//===----------------------------------------------------------------------===//
+
+def ReserveNamesOp : SVOp<"reserve_names", []> {
+  let summary = "Disallow a set of names to be used during emission";
+
+  let arguments = (ins StrArrayAttr:$reservedNames);
+  let assemblyFormat = [{
+    $reservedNames attr-dict
+  }];
+}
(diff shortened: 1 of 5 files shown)


lib/Conversion/ExportVerilog/ExportVerilogInternals.h
```diff
@@ -76,6 +80,9 @@ struct GlobalNameTable {
   // This contains prefixes for any typedecl'd enum types. Keys are type-aliases
   // of enum types.
   DenseMap<Type, StringAttr> enumPrefixes;
+
+  /// List of names which are marked as reserved for any name.
+  StringSet<> reservedNames;
```
uenoku, line 85: nit: Can we make this `DenseSet<StringAttr>`? It should avoid unnecessary string allocation.

include/circt/Dialect/SV/SVStatements.td
```diff
@@ -879,3 +879,17 @@ def MacroDefOp : SVOp<"macro.def",
     MacroDeclOp getReferencedMacro(const hw::HWSymbolCache *cache);
   }];
 }
+
+
+//===----------------------------------------------------------------------===//
+// SV output control.
+//===----------------------------------------------------------------------===//
+
+def ReserveNamesOp : SVOp<"reserve_names", []> {
```
uenoku, line 888: I think the current implementation requires op to be in a top-level, correct? If so could you add `hasParent` here? I guess we could put the op in a HWModule op which affects only its local namespace but not sure it's worth allowing that.

uenoku (commented):
It makes sense to me to define it as an op. Alternatively we could implement it as a lowering option to disallow specific names but it would be more flexible to import names from op.

teqdruid:
@uenoku at the last ODM we agreed it should be an op which lives at the top level.

include/circt/Dialect/SV/SVStatements.td
teqdruid: This is not something we need right now.

uenoku (approved):
LGTM

PR #7026, opened 2024-05-13
[OM] Separates OM object fields verifier to a dedicated pass

uenoku:
Verification for OM object fields performs global analysis on a symbol table op. This cannot be done efficiently today within symbol table verification without nested symbol tables. So this PR separates the verifier into a dedicated pass and run the verification at several times in the pipeline.

### DIFF
--- a/include/circt/Dialect/OM/OMPasses.td
+++ b/include/circt/Dialect/OM/OMPasses.td
@@ -25,3 +25,11 @@ def LinkModules: Pass<"om-link-modules", "mlir::ModuleOp"> {
   }];
   let constructor = "circt::om::createOMLinkModulesPass()";
 }
+
+def VerifyObjectFields: Pass<"om-verify-object-fields"> {
+  let summary = "Verify fields of ObjectOp are valid";
+  let description = [{
+    Verify object fields are valid.
+  }];
+  let constructor = "circt::om::createVerifyObjectFieldsPass()";
+}
--- a/lib/Dialect/OM/OMOps.cpp
+++ b/lib/Dialect/OM/OMOps.cpp
@@ -331,16 +331,6 @@ circt::om::ObjectOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
 }
 
 //===----------------------------------------------------------------------===//
-// ObjectFieldOp
-//===----------------------------------------------------------------------===//
-
-LogicalResult
-circt::om::ObjectFieldOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
-  // This is verified by VerifyObjectFields pass.
-  return success();
-}
-
-//===----------------------------------------------------------------------===//
 // ConstantOp
 //===----------------------------------------------------------------------===//
 
--- a/lib/Dialect/OM/OMOps.cpp
+++ b/lib/Dialect/OM/OMOps.cpp
@@ -336,60 +336,7 @@ circt::om::ObjectOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
 
 LogicalResult
 circt::om::ObjectFieldOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
-  // Get the ObjectInstOp and the ClassLike it is an instance of.
-  ObjectOp objectInst = getObject().getDefiningOp<ObjectOp>();
-  ClassLike classDef = cast<ClassLike>(symbolTable.lookupNearestSymbolFrom(
-      *this, objectInst.getClassNameAttr()));
-
-  // Traverse the field path, verifying each field exists.
-  ClassFieldLike finalField;
-  auto fields = SmallVector<FlatSymbolRefAttr>(
-      getFieldPath().getAsRange<FlatSymbolRefAttr>());
-  for (size_t i = 0, e = fields.size(); i < e; ++i) {
-    // Verify the field exists on the ClassOp.
-    auto field = fields[i];
-    ClassFieldLike fieldDef;
-    classDef.walk([&](ClassFieldLike fieldLike) {
-      if (fieldLike.getNameAttr() == field.getAttr()) {
-        fieldDef = fieldLike;
-        return WalkResult::interrupt();
-      }
-      return WalkResult::advance();
-    });
-    if (!fieldDef) {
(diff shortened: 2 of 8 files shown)


include/circt/Dialect/OM/OMPasses.td
```diff
@@ -25,3 +25,11 @@ def LinkModules: Pass<"om-link-modules", "mlir::ModuleOp"> {
   }];
   let constructor = "circt::om::createOMLinkModulesPass()";
 }
+
+def VerifyObjectFields: Pass<"om-verify-object-fields"> {
```
uenoku, line 29: The pass is intentionally unrestricted so that we can schedule the pass on build.module and firrtl.circuit.

mikeurbach (approved):
Thanks, this looks great to me.

lib/Dialect/OM/Transforms/VerifyObjectFields.cpp
```diff
@@ -0,0 +1,133 @@
...
+                  if (!classType) {
+                    objectField.emitOpError("nested field access into ")
+                        << field << " requires a ClassType, but found "
+                        << fieldDef.getType();
+                    return WalkResult::interrupt();
+                  }
+
+                  // The nested ClassOp must exist, since a field with ClassType
+                  // must be an ObjectInstOp, which already verifies the class
+                  // exists.
```
dtzSiFive, line 100: "me be an ObjectInstOp" -> that is, if this is valid and not an error, right?

If so, might be good to handle this either not finding a class or that the symbol resolves to a non-classlike.  I think `lookupNearestSymbolFrom<ClassLike>(..)` might help with that.

include/circt/Dialect/OM/OMPasses.td
dtzSiFive: Instead of the assert, could use `canScheduleOn` to only run on what's supported? Just a thought.

lib/Dialect/OM/OMOps.cpp
```diff
@@ -336,60 +336,7 @@ circt::om::ObjectOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
-    finalField = fieldDef;
-  }
-
-  // Verify the accessed field type matches the result type.
-  if (finalField.getType() != getResult().getType())
-    return emitOpError("expected type ")
-           << getResult().getType() << ", but accessed field has type "
-           << finalField.getType();
-
+  // This is verified by VerifyObjectFields pass.
```
dtzSiFive, line 339: Is there utility in implementing this / SymbolUserOpInterface if it is only ever success (and our symbol users probably don't work right anyway -- they aren't underneath the SymbolTable defining the symbols referenced AFAICT).

Anyway maybe fine for now, pending a rework esp now we're managing our own verification too.

lib/Dialect/OM/Transforms/VerifyObjectFields.cpp
```diff
@@ -0,0 +1,133 @@
...
+  llvm::MapVector<ClassLike, llvm::DenseMap<StringAttr, ClassFieldLike>> tables;
+  for (auto op : module->getRegion(0).getOps<om::ClassLike>())
+    tables.insert({op, llvm::DenseMap<StringAttr, ClassFieldLike>()});
+
+  // Peel tables parallelly.
+  mlir::parallelForEach(&getContext(), tables, [](auto &entry) {
+    ClassLike classLike = entry.first;
+    auto &table = entry.second;
+    classLike.walk([&](ClassFieldLike fieldLike) {
+      table.insert({fieldLike.getNameAttr(), fieldLike});
```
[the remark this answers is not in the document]
dtzSiFive: (these aren't checked via SymbolTable I think (since ClassLike aren't), but maybe ClassLike's verify this already and it runs before this?)

lib/Dialect/OM/Transforms/VerifyObjectFields.cpp
dtzSiFive: This might not be needed, just checking 👍 .

dtzSiFive (approved):
LGTM, should embrace these as really not being Symbols and symbol users in the MLIR sense and rework into whatever they are for us.

Thanks for chasing this down and working up a solution here!

include/circt/Dialect/OM/OMPasses.td
mikeurbach: Oh nice, I didn't know about `canScheduleOn`.

lib/Dialect/OM/OMOps.cpp
mikeurbach: I think when we finally fix this for real, this will probably be where we re-land the verifier, since `om.object.field` will still reference the `om.class` symbol. The difference is we will re-work the IR to not have to then scan the `om.class` body to perform the verification. So I guess this is fine to rip out now, and we can add it back then, or we can leave the placeholder.

lib/Dialect/OM/Transforms/VerifyObjectFields.cpp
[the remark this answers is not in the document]
uenoku: Added a check, thanks!

include/circt/Dialect/OM/OMPasses.td
uenoku: Yeah, thanks. It looks working well!

lib/Dialect/OM/OMOps.cpp
dtzSiFive: I was thinking that since these aren't not compatible with MLIR symbols the rework wouldn't use this since these aren't symbol users in the normal sense?

But okay, something to chat about later...

lib/Dialect/OM/Transforms/VerifyObjectFields.cpp
uenoku: Yes I think it could crash when object is not created by ObjectInstOp. I added a check and test. Thanks!

lib/Dialect/OM/OMOps.cpp
uenoku: Yeah I think we can drop the interface. ObjectFieldOp would not be a MLIR symbol user.

Issue #6298, opened 2023-10-17
[Arc] Split large LLVM functions


fabianschuiki:
Arcilator has a tendency to produce very large LLVM functions in its output. For example, pretty much an entire clock domain will be grouped into a single function, which for large designs can be millions of LLVM IR ops in size. These cause massive compilation slowdowns (some parts of LLVM scale fairly poorly for excessively large functions -- e.g. O(n³) in regalloc, IIRC).

To fix this, add a pass that splits large functions up into smaller ones. This could be as simple as putting the ops in large functions into groups, and then outlining these groups into separate functions.

PR #7027, opened 2024-05-13
[Arc] Add SplitFuncsPass

TaoBi22:
As proposed in #6298, this PR adds a pass that splits up large functions into smaller functions to (eventually - I haven't included it in the Arcilator pipeline in this PR, as we'll want to actually split up the compilation units when using the pass, which is still on the todo list) work around parts of the pipeline that scale very badly with large funcs. This uses some liveness analysis bits so that (only) the values that are needed further up the call stack are passed up.


include/circt/Dialect/Arc/ArcPasses.h
```diff
@@ -46,6 +46,7 @@ std::unique_ptr<mlir::Pass> createLowerVectorizationsPass(
 std::unique_ptr<mlir::Pass> createMakeTablesPass();
 std::unique_ptr<mlir::Pass> createMuxToControlFlowPass();
 std::unique_ptr<mlir::Pass> createSimplifyVariadicOpsPass();
+std::unique_ptr<mlir::Pass> createSplitFuncsPass(unsigned splitBound = 20000);
```
fabianschuiki, line 49: I think you should be able to use the options struct that table-gen produces for you. That way you don't have to redefine the default split bound 🙂 
```suggestion
std::unique_ptr<mlir::Pass> createSplitFuncsPass(const SplitFuncsOptions &options = {});
```

fabianschuiki (approved):
This is very cool! Might we worth to add an option to arcilator that adds this pass to the pipeline 😃

include/circt/Dialect/Arc/ArcPasses.h
TaoBi22: Very nifty, the redefinition was bugging me! Thanks!

TaoBi22:
@fabianschuiki added an Arcilator option to run the pass - occurred to me while checking it works correctly that since the pass currently operates on `func.func`s the `llvm.funcs` that eventually get produced could potentially be a fair bit longer than the threshold given since some MLIR ops will be lowered to multiple LLVM ops. Guessing this is still probably fine since we can still use this to divide compilation units, but can always make the pass operate on `llvm.func`s instead if we'd rather have more precise control!

fabianschuiki:
I think running on `func.func` as you're doing right now is great! LLVM funcs feels like they're just the very last step before we kick stuff over to LLVM, kind of like an export dialect, but we don't necessarily want to perform optimizations at that point. The size of the splits is probably going to be just a good guess anyway, so if the final LLVM functions end up being 50% larger we probably won't mind 😃.

lib/Dialect/Arc/Transforms/SplitFuncs.cpp
```diff
@@ -0,0 +1,181 @@
...
+    if (failed(lowerFunc(op)))
+      return signalPassFailure();
+}
+
+LogicalResult SplitFuncsPass::lowerFunc(FuncOp funcOp) {
+  assert(splitBound != 0 && "Cannot split functions into functions of size 0");
+  int numOps = funcOp.front().getOperations().size();
+  if (numOps < splitBound)
+    return success();
+  int numBlocks = ceil((float)numOps / splitBound);
```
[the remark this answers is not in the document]
TaoBi22: Oh perfect!!

lib/Dialect/Arc/Transforms/SplitFuncs.cpp
```diff
@@ -0,0 +1,182 @@
...
+  OpBuilder opBuilder(funcOp->getContext());
+  std::vector<Block *> blocks;
+  assert(funcOp->getNumRegions() == 1);
+  Block *frontBlock = &(funcOp.getBody().front());
+  blocks.push_back(frontBlock);
+  for (int i = 0; i < numBlocks - 1; i++) {
+    std::vector<Location> locs;
+    for (unsigned long j = 0; j < frontBlock->getArgumentTypes().size(); j++) {
+      locs.push_back(funcOp.getLoc());
+    }
```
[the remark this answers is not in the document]
TaoBi22: Ah great, I figured there must be a better way to do this!

TaoBi22:
@maerhart thanks for the great review! Think that should be everything addressed, let me know if I missed anything before I merge!

lib/Dialect/Arc/Transforms/SplitFuncs.cpp
```diff
@@ -0,0 +1,164 @@
...
+    SmallVector<Value> outValues;
+    llvm::for_each(liveOut, [&outValues](auto el) {
+      if (!isa<BlockArgument>(el))
+        outValues.push_back(el);
+    });
+    opBuilder.setInsertionPointToEnd(currentBlock);
+    opBuilder.create<ReturnOp>(funcOp->getLoc(), outValues);
+  }
+  // Create and populate new FuncOps
+  for (long unsigned i = 0; i < blocks.size() - 1; i++) {
```
maerhart, line 119: Nit: inconsistent post and pre-increments, LLVM generally prefers pre-increment.

lib/Dialect/Arc/Transforms/SplitFuncs.cpp
```diff
@@ -0,0 +1,164 @@
...
+    auto *funcBlock = newFunc.addEntryBlock();
+    for (auto &op : make_early_inc_range(currentBlock->getOperations())) {
+      op.remove();
+      funcBlock->push_back(&op);
+    }
+    currentBlock->erase();
+
+    opBuilder.setInsertionPointToEnd(funcBlock);
+    for (auto [j, el] : llvm::enumerate(args))
+      replaceAllUsesInRegionWith(el, newFunc.getArgument(j++),
```
maerhart, line 149:
```suggestion
      replaceAllUsesInRegionWith(el, newFunc.getArgument(j),
```

maerhart (approved):
Thanks!

lib/Dialect/Arc/Transforms/SplitFuncs.cpp
```diff
@@ -0,0 +1,164 @@
...
+    if (failed(lowerFunc(op)))
+      return signalPassFailure();
+}
+
+LogicalResult SplitFuncsPass::lowerFunc(FuncOp funcOp) {
+  if (splitBound == 0)
+    return funcOp.emitError("Cannot split functions into functions of size 0.");
+  if (funcOp.getBody().getBlocks().size() > 1)
+    return funcOp.emitError("Regions with multiple blocks are not supported.");
+  assert(splitBound != 0 && "Cannot split functions into functions of size 0");
```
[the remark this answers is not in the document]
TaoBi22: Oops, thought I'd dropped that, thanks for the catch!

TaoBi22:
Happy for me to merge @maerhart?

maerhart:
> Happy for me to merge @maerhart?

Of course, thanks for all the improvements!

PR #7028, opened 2024-05-13
cmake: circt install directory for CAPI

dtzSiFive:
Use our install macro, in same way that `add_circt_library` works.

Ensures CAPI bits have same installation logic as other libraries.


fabianschuiki (approved):
:heart_eyes:

PR #7030, opened 2024-05-13
[FIRRTL][Import] Remove support for printf-encoded verif.

dtzSiFive:
Retain logic to recognize (but not parse or diagnose) printf's of the various "flavors" previously supported so that we can reject designs that rely on this removed support (emit error).

Similar to https://github.com/llvm/circt/pull/7010, but retains logic for recognizing these printf's so we can reject designs with their use instead of silently handling them "incorrectly".

Good parts of this inspired by suggestion on that PR: https://github.com/llvm/circt/pull/7010#discussion_r1595421600 .

This still breaks usage with even recent Chisel, but compared to 7010 informs the user they've used something that doesn't work anymore.

### DIFF
--- a/lib/Dialect/FIRRTL/Import/FIRParser.cpp
+++ b/lib/Dialect/FIRRTL/Import/FIRParser.cpp
@@ -5390,36 +5390,27 @@ FIRCircuitParser::parseModuleBody(const SymbolTable &circuitSymTbl,
   if (failed(result))
     return result;
 
-  // Convert print-encoded verifications after parsing.
-
+  // Scan for printf-encoded verif's to error on their use, no longer supported.
   {
-    // Gather PrintFOps and then process to avoid walking while mutating.
-    SmallVector<PrintFOp> buffer;
-    deferredModule.moduleOp.walk(
-        [&buffer](PrintFOp printFOp) { buffer.push_back(printFOp); });
-
     size_t numVerifPrintfs = 0;
     std::optional<Location> printfLoc;
-    for (auto printFOp : buffer) {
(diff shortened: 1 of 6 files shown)


dtzSiFive:
This breaks compatibility with FIRRTL generated by not-latest Chisel, although this version will detect this and produce an error.  Generally we've mostly kept backwards compatibility with both old FIRRTL and old Chisel-emitted constructs (annotations, etc), and anyway so should only land this if we're ready to change that :+1:.

(thoughts/input welcome, LMK :+1:)

uenoku (approved):
LGTM!

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5390,36 +5390,27 @@ FIRCircuitParser::parseModuleBody(const SymbolTable &circuitSymTbl,
...
           << numVerifPrintfs
-          << " printf-encoded verification operation(s), which are deprecated "
-             "and will be removed in the future";
+          << " printf-encoded verification operation(s), which are no longer "
+             "supported.";
       diag.attachNote(*printfLoc)
-          << "example printf here, will just be a printf in the future";
+          << "example printf here, this is now just a printf and nothing more";
       diag.attachNote() << "For more information, see "
                            "https://github.com/llvm/circt/issues/6970";
```
uenoku, line 5415: Is returning a failure missing?

dtzSiFive: Whoops, changed from warning to error and missed this.  Will fix, thanks!

PR #7032, opened 2024-05-14
[ESI][Runtime] Remove C++ includes from wheels

teqdruid:
There is no good way to use the binaries included with the wheel. Essentially, they are built with the old `std::basic_string` but modern software uses `std::__cxx11::basic_string`. There is now way to change this behavior (AFAICT) in manylinux2014. Newer manylinux images have a way, but we need compatiblity back to glibc 2.17, which none of them provide.


teqdruid:
*sigh*

PR #7035, opened 2024-05-14
[Moore] Move IntType definition into ODS

fabianschuiki:
Replace the C++ definition of `IntType` with a corresponding table-gen definition in `MooreTypes.td`.

### DIFF
--- a/include/circt/Dialect/Moore/MooreTypes.td
+++ b/include/circt/Dialect/Moore/MooreTypes.td
@@ -53,6 +53,46 @@ def EventType : MooreTypeDef<"Event", [], "moore::UnpackedType"> {
 
 
 //===----------------------------------------------------------------------===//
+// Packed Types
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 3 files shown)


include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -52,6 +52,46 @@ def EventType : MooreTypeDef<"Event", [], "moore::UnpackedType"> {
 }
 
 
+//===----------------------------------------------------------------------===//
+// Packed Types
+//===----------------------------------------------------------------------===//
+
+def IntType : MooreTypeDef<"Int", [], "moore::PackedType"> {
+  let typeName = "moore.int";
```
maerhart, line 60: I think one should not use this directly, but the mnemonic instead. However, since you have custom parsers and printers, you shouldn't have to set this at all.

maerhart (approved):
LGTM

uenoku (approved):
Nice!

PR #7036, opened 2024-05-14
[Moore] Simplify and move RealType definition into ODS

fabianschuiki:
Coalesce SystemVerilog's `shortreal`, `real`, and `realtime` types into a single `!moore.real` type understood to be 64 bits wide, and move the definition from C++ land into `MooreTypes.td`. Once ImportVerilog actually supports expressions on real values we may want to revisit this type and either replace it with the standard `f32` and `f64` types, or create dedicated types for `shortreal`, `real`, and `realtime`.


uenoku (approved):
Nice clean up!

PR #7042, opened 2024-05-15
[Sim] Add DPI func/call and lowering

uenoku:
This PR adds DPI func/call op and SimToSV lowering. 

`sim.dpi.func` is a just bridge to `sv.func`. This op can be lowered into `func.func` in the future for Arc integration. 

`sim.func.dpi.call` represents SystemVerilog DPI function call. There are two optional operands `clock` and `enable`.

If `clock` is not provided, the callee is invoked when input values are changed. If provided, the dpi function is called at clock's posedge. The result values behave like registers and the DPI function is used as a state transfer function of them. `enable` operand is used to conditionally call the DPI since DPI call could be quite more expensive than native constructs. When `enable` is low, results of unclocked calls are undefined and in SV results they are lowered into `X`. Users are expected to gate result values by another `enable` to model a default value of results. For clocked calls, a low enable means that its register state transfer function is not called. Hence their values will not be modify in that clock.

A function that returns an explicit return is not supported yet. I'll follow-up.


test/Conversion/SimToSV/dpi.mlir
```diff
@@ -0,0 +1,94 @@
...
+// CHECK:       sv.func private @dpi(out arg0 : i1, in %arg1 : i1, out arg2 : i1)
+// CHECK-NEXT:  emit.fragment @dpi_dpi_import_fragument {
+// CHECK-NEXT:    sv.func.dpi.import @dpi
+// CHECK-NEXT:  }
+
+// VERILOG:      import "DPI-C" function void dpi( 
+// VERILOG-NEXT:   output arg0,
+// VERILOG-NEXT:   input  arg1,
+// VERILOG-NEXT:   output arg2
+// VERILOG-NEXT: );
```
fabianschuiki, line 14: I'm curious what would happen if `%arg1` were an `i32`. Would you get `(output arg0, input int arg1, output arg2)`, or `... output bit arg2)`? I think function arguments in SV inherit the previous argument's type if omitted, similar to types on ports. It might be worthwhile to explicitly emit `bit` for `i1` arguments.

fabianschuiki (approved):
This looks very nice! Minor thought on omitting `bit` types, but otherwise this is fantastic!

uenoku:
Thank you for review!

PR #7043, opened 2024-05-15
[FIRRTL][Intrinsics] source materialization for inferred types.

dtzSiFive:
If the new op's result doesn't match, insert a wire + connect (if valid) to handle the mismatch.

For intrinsics that lower to ops with return type inference, the inferred type may be different than the declared type (e.g., `UInt` vs `UInt<1>`).

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
@@ -7,6 +7,8 @@
 //===----------------------------------------------------------------------===//
 
 #include "circt/Dialect/FIRRTL/FIRRTLIntrinsics.h"
+#include "circt/Dialect/FIRRTL/FIRRTLTypes.h"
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
```diff
@@ -146,10 +146,36 @@ FailureOr<size_t> IntrinsicLowerings::lower(FModuleOp mod,
...
+      return {};
+
+    auto w = builder.create<WireOp>(loc, resultType).getResult();
+    emitConnect(builder, loc, w, inputs.front());
+    return w;
+  };
+  // New result doesn't match? Add wire + connect.
+  typeConverter.addSourceMaterialization(firrtlBaseTypeMaterialization);
+  // New operand doesn't match? Add wire + connect.
+  typeConverter.addTargetMaterialization(firrtlBaseTypeMaterialization);
```
dtzSiFive, line 173: This isn't needed yet but don't see why to not include it.  Just a note 👍 .

PR #7044, opened 2024-05-15
[LTL] Add clocked and disabled property types

dobios:
#7022 highlighted a difficulty in reasoning about certain characteristics of ltl properties, e.g. finding out how many clocks or disables are associated to a given property without resorting to a use-def analysis.
This PR attempts to solve that by encoded those concepts directly in ltl's type-system. 

Let me know what you think!

### DIFF
--- a/lib/Dialect/LTL/LTLTypes.cpp
+++ b/lib/Dialect/LTL/LTLTypes.cpp
@@ -1,0 +1,24 @@
+#include "circt/Dialect/LTL/LTLTypes.h"
+#include "circt/Dialect/LTL/LTLOps.h"
+
+using namespace circt;
+using namespace ltl;
+
+bool circt::ltl::isClocked(mlir::Type type) {
+  return isa<ClockedPropertyType>(type) || isa<ClockedSequenceType>(type) ||
+         isa<ClockedDisabledPropertyType>(type);
+}
+
+bool circt::ltl::isDisabled(mlir::Type type) {
+  return isa<DisabledPropertyType>(type) ||
+         isa<ClockedDisabledPropertyType>(type);
+}
+
+bool circt::ltl::isProperty(mlir::Type type) {
+  return isa<ClockedPropertyType>(type) || isa<DisabledPropertyType>(type) ||
+         isa<ClockedDisabledPropertyType>(type) || isa<PropertyType>(type);
+}
+
+bool circt::ltl::isSequence(mlir::Type type) {
+  return isa<SequenceType>(type) || isa<ClockedSequenceType>(type);
+}
--- a/lib/Dialect/LTL/LTLTypes.cpp
+++ b/lib/Dialect/LTL/LTLTypes.cpp
@@ -5,20 +5,20 @@ using namespace circt;
 using namespace ltl;
 
 bool circt::ltl::isClocked(mlir::Type type) {
-  return isa<ClockedPropertyType>(type) || isa<ClockedSequenceType>(type) ||
-         isa<ClockedDisabledPropertyType>(type);
+  return isa<ClockedPropertyType, ClockedSequenceType,
+    ClockedDisabledPropertyType>(type);
 }
 
 bool circt::ltl::isDisabled(mlir::Type type) {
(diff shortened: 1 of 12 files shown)


lib/Dialect/LTL/LTLTypes.cpp
```diff
@@ -0,0 +1,24 @@
+#include "circt/Dialect/LTL/LTLTypes.h"
+#include "circt/Dialect/LTL/LTLOps.h"
+
+using namespace circt;
+using namespace ltl;
+
+bool circt::ltl::isClocked(mlir::Type type) {
+  return isa<ClockedPropertyType>(type) || isa<ClockedSequenceType>(type) ||
+         isa<ClockedDisabledPropertyType>(type);
```
uenoku, line 9: FYI you can pass multiple classes to `isa<>`, e.g:
```suggestion
  return isa<ClockedPropertyType, ClockedSequenceType, ClockedDisabledPropertyType>(type);
```

ekiwi:
There is a 4th way of fixing this issue: Remove the `ltl.clock` operation and make Chisel instead emit _clocked_ `assert`, `assume` and `cover` statements. This way you do not have to change the type system, make a checker or even a translation pass. Everything would already by in its most useful form when it comes from Chisel. Are there any downsides to removing `ltl.clock`?

dobios:
> There is a 4th way of fixing this issue: Remove the `ltl.clock` operation and make Chisel instead emit _clocked_ `assert`, `assume` and `cover` statements. This way you do not have to change the type system, make a checker or even a translation pass. Everything would already by in its most useful form when it comes from Chisel. Are there any downsides to removing `ltl.clock`?

Hey @ekiwi! Thanks for the comment! The one reason we would want to keep `ltl.clock` is to be able to express properties with multiple clocks, i.e. (from my limited knowledge of SV) in SV you can write something like
```sv
assert property (@(posedge clk) disable iff (cond) (a |-> (@(posedge fclk) b ##1 c)))
```
and we would not be able to express that without having `ltl.clock` or some other way to associate a clock to an arbitrary subproperty without having it be part of the assertion,

That being said, AFAIK we can't actually express that in Chisel (correct me if I'm wrong) but so far the only way I was able to find to connect a clock to a property was via `AssertProperty` which itself only accepts one clock input. So for the Chisel case, we can probably offload this to the front-end entirely and have it only emit `clocked_assert` like operations. I'll look into this a bit more and see if this might be an issue for any other front-end (other than SV which isn't really supported yet).

ekiwi:
> Hey @ekiwi! Thanks for the comment! The one reason we would want to keep `ltl.clock` is to be able to express properties with multiple clocks, i.e. (from my limited knowledge of SV) in SV you can write something like

I never considered this case before, thanks for pointing it out. Do you know what the semantics of this are? Is the property essentially active on either clock edge and then certain automaton transitions are only available depending on which event triggered it?

Anyways, I do feel like once you are in a position where this feature is required, you might be able to come up with a better solution than the very generic `ltl.clock` operation.

dobios:
> > Hey @ekiwi! Thanks for the comment! The one reason we would want to keep `ltl.clock` is to be able to express properties with multiple clocks, i.e. (from my limited knowledge of SV) in SV you can write something like
> 
> I never considered this case before, thanks for pointing it out. Do you know what the semantics of this are? Is the property essentially active on either clock edge and then certain automaton transitions are only available depending on which event triggered it?
> 
> Anyways, I do feel like once you are in a position where this feature is required, you might be able to come up with a better solution than the very generic `ltl.clock` operation.

SV basically has a clock resolution process, where it will check the sequence using the closest parenting clock. So in the case of the example I showed, the consequent sequence in the overlapping implication is describing ``fclk` clock ticks with `##1` rather than `clk` clock ticks. I think this process is described in [IEEE 1800 16.16](https://fpga.mit.edu/6205/_static/F23/documentation/1800-2017.pdf#G21.1030892), and there are some examples of multiple clocks being used later on.

dobios:
After further reasoning, I don't think that this is a good approach, so I'm closing the PR.

Issue #7047, opened 2024-05-16
`-DMLIR_ENABLE_EXPENSIVE_PATTERN_API_CHECKS=ON` might have found some invalid API usages


7FM:
While the [documentation](https://mlir.llvm.org/getting_started/Debugging/#detecting-invalid-api-usage) states that there might be false positives, I think it might be worth a look into the individual test cases and patterns.
If there are no false positives, adding `-DMLIR_ENABLE_EXPENSIVE_PATTERN_API_CHECKS=ON` to the CI could be beneficial.
```
Failed Tests (32):
  CIRCT :: Conversion/LoopScheduleToCalyx/convert_pipeline.mlir
  CIRCT :: Conversion/SCFToCalyx/cider_source_location.mlir
  CIRCT :: Conversion/SCFToCalyx/convert_controlflow.mlir
  CIRCT :: Conversion/SCFToCalyx/convert_func.mlir
  CIRCT :: Conversion/SCFToCalyx/convert_memory.mlir
  CIRCT :: Conversion/SCFToCalyx/convert_simple.mlir
  CIRCT :: Conversion/SCFToCalyx/errors.mlir
  CIRCT :: Dialect/Arc/arc-canonicalizer.mlir
  CIRCT :: Dialect/Arc/canonicalizers.mlir
  CIRCT :: Dialect/Calyx/canonicalization.mlir
  CIRCT :: Dialect/Comb/canonicalization.mlir
  CIRCT :: Dialect/FIRRTL/SFCTests/GrandCentralInterfaces/Wire.fir
  CIRCT :: Dialect/FIRRTL/SFCTests/data-taps.fir
  CIRCT :: Dialect/FIRRTL/SFCTests/invalid-interpretations.fir
  CIRCT :: Dialect/FIRRTL/SFCTests/invalid-reg-pass.fir
  CIRCT :: Dialect/FIRRTL/SFCTests/mem-taps.fir
  CIRCT :: Dialect/FIRRTL/canonicalization.mlir
  CIRCT :: Dialect/FIRRTL/ref.mlir
  CIRCT :: Dialect/FIRRTL/simplify-mems.mlir
  CIRCT :: Dialect/HW/canonicalization.mlir
  CIRCT :: Dialect/HW/inline.mlir
  CIRCT :: Dialect/Handshake/canonicalization.mlir
  CIRCT :: Dialect/LLHD/Canonicalization/extract.mlir
  CIRCT :: firtool/async-reset.fir
[...]

PR #7049, opened 2024-05-16
[FIRRTL][Folds] Fix patterns to use rewriter for RAUW

dtzSiFive:
Two of these caught by -DMLIR_ENABLE_EXPENSIVE_PATTERN_API_CHECKS=ON.

Others by inspection.

cc #7047.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -2644,7 +2644,7 @@ struct FoldReadWritePorts : public mlir::RewritePattern {
           rewriter.replaceOpWithNewOp<WireOp>(wmodeField, wmodeField.getType());
         }
       } else {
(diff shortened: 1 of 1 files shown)


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1960,9 +1960,7 @@ struct NodeBypass : public mlir::RewritePattern {
     if (node.getInnerSym() || !AnnotationSet(node).canBeDeleted() ||
         node.use_empty() || node.isForceable())
       return failure();
-    rewriter.startOpModification(node);
-    node.getResult().replaceAllUsesWith(node.getInput());
```
dtzSiFive, line 1964: This modifies operations that aren't `node` (unless input is its own result), change to use RAUW via rewriter.

darthscsi (approved):
LGTM

PR #7052, opened 2024-05-16
[Handshake] Fix canonicalizer not going through rewriter for RAUW.

dtzSiFive:

cc #7047.

### DIFF
--- a/lib/Dialect/Handshake/HandshakeOps.cpp
+++ b/lib/Dialect/Handshake/HandshakeOps.cpp
@@ -207,12 +207,10 @@ struct EliminateUnusedForkResultsPattern : mlir::OpRewritePattern<ForkOp> {
     auto operand = op.getOperand();
     auto newFork = rewriter.create<ForkOp>(
         op.getLoc(), operand, op.getNumResults() - unusedIndexes.size());
-    rewriter.modifyOpInPlace(op, [&] {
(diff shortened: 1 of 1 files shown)


teqdruid (commented):
Is `modifyOpInPlace` not a valid thing to use anymore? It's generally used for good reason. There may be some subtlety here that necessitates it.

dtzSiFive:
> Is `modifyOpInPlace` not a valid thing to use anymore? It's generally used for good reason. There may be some subtlety here that necessitates it.

Thanks for taking a critical look here!

The RAUW's are updating ops using `parentForkOp` results, not that op itself, which is what I think was going wrong and addressed here.  Possibly could keep it and also use rewriter for the RAUW from within? 🤔

teqdruid:
I'm not sure. Did Morten write this code? If so, the conservative thing to do is to wait for him to come back from vacation and get his review. Should be back in two or three weeks, but this doesn't seem urgent... yes?

dtzSiFive:
> I'm not sure. Did Morten write this code? If so, the conservative thing to do is to wait for him to come back from vacation and get his review. Should be back in two or three weeks, but this doesn't seem urgent... yes?

Not urgent AFAIK.  Sounds good to me, thanks!

mortbopet (approved):

PR #7054, opened 2024-05-16
[Comb] Fix MuxRewriter to eraseOperand in rewriter-aware way.

dtzSiFive:

cc #7047.

### DIFF
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -2211,7 +2211,8 @@ static Value extractOperandFromFullyAssociative(Operation *fullyAssoc,
 
   // If the operation has a single use, mutate it in place.
   if (fullyAssoc->hasOneUse()) {
-    fullyAssoc->eraseOperand(operandNo);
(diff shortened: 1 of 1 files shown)


lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -2211,7 +2211,8 @@ static Value extractOperandFromFullyAssociative(Operation *fullyAssoc,
 
   // If the operation has a single use, mutate it in place.
   if (fullyAssoc->hasOneUse()) {
-    fullyAssoc->eraseOperand(operandNo);
+    rewriter.modifyOpInPlace(fullyAssoc,
```
teqdruid, line 2214: Is there no `rewriter.eraseOperand()`?

teqdruid (approved):
LGTM modulo my question.

lib/Dialect/Comb/CombFolds.cpp
dtzSiFive: Not AFAICT, no.

dtzSiFive:
Thanks!

PR #7055, opened 2024-05-16
[LLHD] Fix operand assignments not going through rewriter.

dtzSiFive:

cc #7047.


fabianschuiki (approved):
Thanks a lot for fixing this!

I'm wondering whether we should have a run with these checks enabled in CI somewhere :thinking:

dtzSiFive:
> I'm wondering whether we should have a run with these checks enabled in CI somewhere 🤔

I think so! Very much so.  Cronjob on some machine if nothing else 😄 .

I'd say also regularly checking ASAN+UBSAN is a good idea :+1:.
And filecheck_lint.py .

Especially once we're clean w.r.t. these checks.

PR #7060, opened 2024-05-17
[LowerClasses]  Improve performance, NFC

uenoku:
This PR fixes a regression caused by PortAnnotation mutation. We have to batch-update the port annotations to avoid compile time regression. Several other improves are added along with the fix:

* `shouldCreateClass` is slightly expensive since it walks entire IR to check instances. `shouldCreateClass` is already lazily evaluated but however `shouldCreateClass` is immediately called for every module. So this PR changes to compute the all results parallelly beforehand instead of lazy evaluation. 
* `std::distance(a, b) == 0` is replaced with `a == b` to avoid potential iterator traversals (maybe it is optimized by clang/gcc though). 
*`processPathTrackers` is refactored into a helper struct `PathTracker` so that we can process paths/annotations parallely.  
We update `pathInfoTable` sequentially afterwards.


lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -36,6 +36,33 @@ using namespace circt::om;
...
+    return isa<PropertyType>(port.type);
+  });
+
+  if (hasClassPorts)
+    return true;
+
+  // Create a class for modules that instantiate classes or modules with
+  // property ports.
+  for (auto op :
+       moduleLike.getOperation()->getRegion(0).getOps<FInstanceLike>())
```
dtzSiFive, line 58: Looks like the InstanceGraph is available at this point -- if so, instead of walking to gather instances, ask the instance graph?

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -193,8 +220,317 @@ struct LowerClassesPass : public LowerClassesBase<LowerClassesPass> {
...
+              const DenseMap<DistinctAttr, FModuleOp> &owningModules)
+      : module(module), moduleNamespace(moduleNamespace),
+        instanceGraph(instanceGraph), symbolTable(symbolTable),
+        owningModules(owningModules) {}
+
+  struct PathInfoTableEntry {
+    Operation *op;
+    DistinctAttr id;
+    StringAttr altBasePathModule;
+    // This is null if the path has no owening module.
```
dtzSiFive, line 244:
```suggestion
    // This is null if the path has no owning module.
```

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -447,7 +602,7 @@ LogicalResult LowerClassesPass::processPaths(
       }
 
       // If we are at a leaf, nothing to do.
-      if (std::distance(it->begin(), it->end()) == 0) {
+      if (it->begin() == it->end()) {
```
dtzSiFive, line 605: `llvm::size(A, B)` exists specifically to confidently do `std::distance(..)` on ranges, just related tip.

For checking distance of zero, comparison seems good to me! 👍

dtzSiFive (approved):
LGTM, nice work thanks!

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -574,44 +741,8 @@ std::unique_ptr<mlir::Pass> circt::firrtl::createLowerClassesPass() {
...
-      if (type_isa<PropertyType>(result.getType())) {
-        shouldCreateClassMemo.insert({moduleLike, true});
-        return true;
-      }
-    }
-  }
-
-  shouldCreateClassMemo.insert({moduleLike, false});
-  return false;
+  assert(it != shouldCreateClassMemo.end() && "it must be computed beforehand");
```
dtzSiFive, line 744: FWIW this is what `at(moduleLike)` does (find, assert, return it->second).

uenoku: Thanks!

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
uenoku: Yeah, thanks!

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -477,6 +635,14 @@ void LowerClassesPass::runOnOperation() {
   hw::InnerSymbolNamespaceCollection namespaces;
   HierPathCache cache(circuit, symbolTable);
 
+  // Fill `shouldCreateClassMemo`.
+  for (auto moduleLike : circuit.getOps<FModuleLike>())
+    shouldCreateClassMemo.insert({moduleLike, false});
+  parallelForEach(
```
darthscsi, line 641: Please also check the single-threaded performance.  Parallelization only improves walltime.  We need to improve absolute time also.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -514,7 +684,7 @@ void LowerClassesPass::runOnOperation() {
           auto hierPath = cache.getOpFor(pathAttr);
           classLoweringState.paths.push_back(hierPath);
         }
-      });
+      }
```
youngar, line 688: Can you take this change further? Something like this:
```
  for (auto *node : instanceGraph) {
    auto moduleLike = node->getModule();
    if (shouldCreateClass(moduleLike)) {
      auto omClass = createClass(moduleLike, pathInfoTable);
      auto &classLoweringState = loweringState.classLoweringStateTable[omClass];
      classLoweringState.moduleLike = moduleLike;
      // Find the module instances under the current module with metadata. These
      // ops will be converted to om objects by this pass. Create a hierarchical
      // path for each of these instances, which will be used to rebase path
      // operations. Hierarchical paths must be created serially to ensure their
      // order in the circuit is deterministc.
      //auto *node = instanceGraph.lookup(moduleLike);
      for (auto *instance : *node) {
        auto inst = instance->getInstance<firrtl::InstanceOp>();
        if (!inst)
          continue;
        // Get the referenced module.
        //auto module =
        //    symbolTable.lookup<FModuleLike>(inst.getReferencedModuleNameAttr());
        auto module = instance->getTarget();
        if (shouldCreateClass(module)) {
          auto targetSym = getInnerRefTo(
              {inst, 0}, [&](FModuleLike module) -> hw::InnerSymbolNamespace & {
                return namespaces[module];
              });
          SmallVector<Attribute> path = {targetSym};
          auto pathAttr = ArrayAttr::get(ctx, path);
          auto hierPath = cache.getOpFor(pathAttr);
          classLoweringState.paths.push_back(hierPath);
        }
      }

File: lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp [hunk 7]
```diff
@@ -193,8 +223,317 @@ struct LowerClassesPass : public LowerClassesBase<LowerClassesPass> {
...
+LogicalResult PathTracker::runOnModule() {
+  // Process the module annotations.
+  if (failed(processPathTrackers(OpAnnoTarget(module))))
+    return failure();
+  // Process module port annotations.
+  for (unsigned i = 0, e = module.getNumPorts(); i < e; ++i)
+    if (failed(processPathTrackers(PortAnnoTarget(module, i))))
+      return failure();
+  module->setAttr("portAnnotations",
+                  ArrayAttr::get(module.getContext(), portAnnotations));
```
youngar, line 314: is there a `setPortAnnotationsAttr` that we can use?  Otherwise maybe `PortAnnoTarget::setAnnotations(..)`?

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -193,8 +223,317 @@ struct LowerClassesPass : public LowerClassesBase<LowerClassesPass> {
...
+  });
+  if (error)
+    return failure();
+
+  if (isa<PortAnnoTarget>(target))
+    // Port annotations are batch-updated afterwards.
+    portAnnotations.push_back(annotations.getArrayAttr());
+  else
+    // Otherwise it's fine to update immediately.
+    target.setAnnotations(annotations);
```
youngar, line 508: ` target.setAnnotations(annotations);` could be made slightly faster with ` cast<OpAnnoTarget>(target)setAnnotations(annotations);`

I don't love something about the way this sometimes mutates an internal structure and other times just sets the annotations. I think I would be happier if this function always appended to an annotation array, and op modifications happened outside this function, and then `portAnnotations` doesn't need to be a member anymore.  It would be even nicer if this whole function didn't care about the target type at all - but we need to know when we create the `targetSym`. WDYT?

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -477,6 +635,14 @@ void LowerClassesPass::runOnOperation() {
...
   HierPathCache cache(circuit, symbolTable);
 
+  // Fill `shouldCreateClassMemo`.
+  for (auto moduleLike : circuit.getOps<FModuleLike>())
+    shouldCreateClassMemo.insert({moduleLike, false});
+  parallelForEach(
+      circuit.getContext(), circuit.getOps<FModuleLike>(), [&](FModuleLike op) {
+        shouldCreateClassMemo[op] = shouldCreateClassImpl(op, instanceGraph);
+      });
+
```
youngar, line 645: Might be better to just iterate the nodes in the instance graph, 
```c++
for (auto *node : instanceGraph)
    shouldCreateClassMemo.insert({node->getModule(), false});
  parallelForEach(
      circuit.getContext(), instanceGraph, [&](InstanceNode *node) {
        // And here you don't have to pass in (op, instanceGraph), it just needs the node.
        shouldCreateClassMemo[node->getModule()] = shouldCreateClassImpl(node);
      });
```

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -193,8 +223,317 @@ struct LowerClassesPass : public LowerClassesBase<LowerClassesPass> {
...
+PathTracker::run(CircuitOp circuit, InstanceGraph &instanceGraph,
+                 hw::InnerSymbolNamespaceCollection &namespaces,
+                 HierPathCache &cache, PathInfoTable &pathInfoTable,
+                 const SymbolTable &symbolTable,
+                 const DenseMap<DistinctAttr, FModuleOp> &owningModules) {
+  SmallVector<PathTracker> trackers;
+  // Prepare workers.
+  for (auto module : circuit.getOps<FModuleLike>())
+    trackers.push_back(PathTracker(module, namespaces[module], instanceGraph,
+                                   symbolTable, owningModules));
```
youngar, line 290: Can you use `emplace_back` here?

Might be slightly faster to iterate the instance graph here, to avoid scanning the hundreds of HierPathOps that are in the root of the circuit, but that is not an "optimization" that we have worried about before, and I'm not totally sure would help.

uenoku: Changed to used emplace_back and instance graph.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
uenoku: Seems like missing method for FModuleLike. Added setPortAnnotationsAttr.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
uenoku: Done, thanks!

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
uenoku: I changed processPathTrackers to return `FailureOr<AnnotationSet>` and mutate annotations from callers. Does that look good to you?

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -36,6 +36,38 @@ using namespace circt::om;
...
+
+  if (isa<firrtl::ClassLike>(moduleLike.getOperation()))
+    return true;
+
+  // Always create a class for public modules.
+  if (moduleLike.isPublic())
+    return true;
+
+  // Create a class for modules with property ports.
+  bool hasClassPorts = llvm::any_of(moduleLike.getPorts(), [](PortInfo port) {
```
dtzSiFive, line 52: Just thinking out loud -- doing this in parallel when it walked the module body made sense, but maybe less so to walk the ports + instance nodes for each?

Something like:
"has property ports" can be computed once per module (and looked up instance-side, instead of re-computed, so tracked in a (temporary?) data structure), and "shouldCreateClassImpl" can then be determined in a single (serial) bottom-up walk of instance graph, computing "hasPropertyPorts" for module definitions along the way.

WDYT?

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
mikeurbach: I checked the single threaded performance on a large internal design, and this change leads to a >90% reduction in runtime in a single threaded context.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
uenoku: I think that's great idea. We have to fix InstanceGraph for bottom-up walk so let me first merge the PR. I'll follow-up to use a bottom-up walk.

PR #7061, opened 2024-05-17
[Arc] Initial SLP support

elhewaty:
co-authors: @fabianschuiki

### DIFF
--- a/include/circt/Dialect/Arc/ArcPasses.td
+++ b/include/circt/Dialect/Arc/ArcPasses.td
@@ -61,8 +61,7 @@ def Dedup : Pass<"arc-dedup", "mlir::ModuleOp"> {
 def FindInitialVectors : Pass<"find-initial-vectors", "mlir::ModuleOp"> {
   let summary = "Finds the ops that can be grouped together into a vector";
   let constructor = "circt::arc::createFindInitialVectorsPass()";
-  let dependentDialects = ["arc::ArcDialect", "comb::CombDialect",
-                           "hw::HWDialect"];
+  let dependentDialects = ["arc::ArcDialect"];
 }
 
 def GroupResetsAndEnables : Pass<"arc-group-resets-and-enables",
--- a/include/circt/Dialect/Arc/ArcPasses.td
+++ b/include/circt/Dialect/Arc/ArcPasses.td
@@ -58,6 +58,13 @@ def Dedup : Pass<"arc-dedup", "mlir::ModuleOp"> {
   ];
 }
 
+def FindInitialVectors : Pass<"find-initial-vectors", "mlir::ModuleOp"> {
+  let summary = "Finds the ops that can be grouped together into a vector";
+  let constructor = "circt::arc::createFindInitialVectorsPass()";
+  let dependentDialects = ["arc::ArcDialect", "comb::CombDialect",
+                           "hw::HWDialect"];
+}
+
 def GroupResetsAndEnables : Pass<"arc-group-resets-and-enables",
                                  "mlir::ModuleOp"> {
   let summary = "Group reset and enable conditions of lowered states";
--- a/lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
+++ b/lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
@@ -43,8 +43,6 @@ namespace arc {
 using namespace circt;
 using namespace arc;
 using llvm::SmallMapVector;
-using llvm::SmallVector;
-using mlir::Operation;
 
 namespace {
 struct TopologicalOrder {
@@ -102,7 +100,7 @@ LogicalResult TopologicalOrder::compute(Block *block) {
 
 namespace {
 using Key = std::tuple<unsigned, StringRef, SmallVector<Type>,
-                       SmallVector<Type>, ArrayRef<NamedAttribute>>;
+                       SmallVector<Type>, DictionaryAttr>;
 
 Key computeKey(Operation *op, unsigned rank) {
   // The key = concat(op_rank, op_name, op_operands_types, op_result_types,
@@ -111,7 +109,7 @@ Key computeKey(Operation *op, unsigned rank) {
       rank, op->getName().getStringRef(),
       SmallVector<Type>(op->operand_type_begin(), op->operand_type_end()),
       SmallVector<Type>(op->result_type_begin(), op->result_type_end()),
-      op->getAttrs());
+      op->getAttrDictionary());
 }
 
 struct Vectorizer {
@@ -139,20 +137,20 @@ template <>
 struct DenseMapInfo<Key> {
   static inline Key getEmptyKey() {
     return Key(0, StringRef(), SmallVector<Type>(), SmallVector<Type>(),
-               ArrayRef<NamedAttribute>());
+               DictionaryAttr());
   }
 
   static inline Key getTombstoneKey() {
     static StringRef tombStoneKeyOpName =
         DenseMapInfo<StringRef>::getTombstoneKey();
     return Key(1, tombStoneKeyOpName, SmallVector<Type>(), SmallVector<Type>(),
-               ArrayRef<NamedAttribute>());
+               DictionaryAttr());
   }
 
   static unsigned getHashValue(const Key &key) {
     return hash_value(std::get<0>(key)) ^ hash_value(std::get<1>(key)) ^
            hash_value(std::get<2>(key)) ^ hash_value(std::get<3>(key)) ^
-           hash_value(ArrayRef(std::get<4>(key)));
+           hash_value(std::get<4>(key));
   }
 
   static bool isEqual(const Key &lhs, const Key &rhs) { return lhs == rhs; }
--- a/lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
+++ b/lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
@@ -1,0 +1,254 @@
+//===- FindInitialVectors.cpp ---------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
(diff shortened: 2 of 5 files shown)


elhewaty:
It's not finished yet, but I need to know whether I am on the right track or not.

@fabianschuiki @maerhart

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
```diff
@@ -0,0 +1,222 @@
...
+std::string computeKey(Operation *op) {
+  // The key = concat(op_rank, op_name, op_operands_types, op_result_types, op_attrs)
+  std::string key = op->getName().getStringRef().str();
+  if (failed(modifyKeyIfPossible(op->getOperandTypes(), key)) ||
+      failed(modifyKeyIfPossible(op->getResultTypes(), key)) || 
+      failed(modifyKeyIfPossible(op->getRegisteredInfo()->getAttributeNames(), key)))
+    assert(false && "Cannot compute the op key");
+
+  return key;
+}
```
fabianschuiki, line 142: I would avoid using a string as key here. String representations can be pretty expensive to compute and are mostly intended for humans to read. You should be able to use an `std::tuple` instead:

```c++
using Key = std::tuple<...>;

Key computeKey(Operation *op, unsigned rank) {
  return std::make_tuple(
    rank,
    op->getName(),
    op->getOperandTypes(), // maybe SmallVector<2>(op->getOperandTypes())
    op->getResultTypes(), // maybe also in a SmallVector
    op->getAttributes()
  );
}
```

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    for(auto &[op, rank] : order.opRanks)
+      candidates[std::to_string(rank) + computeKey(op)].push_back(op);
+
+    return success();
+  }
+
+  LogicalResult vectorize();
+  // Store Isomorphic ops together
+  std::unordered_map<std::string, SmallVector<Operation *>> candidates;
```
fabianschuiki, line 157: You might be able to turn this into a `DenseMap`, which is common in LLVM code bases.

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
```diff
@@ -0,0 +1,222 @@
...
+LogicalResult Vectorizer::vectorize() {
+  // Unachievable?! just in case!
+  if (candidates.empty())
+    return success();
+
+  // Iterate over every group of isomorphic ops  
+  for (const auto &[key, ops] : candidates) {
+    // If the group has only one scalar then it doesn't worth vectorizing
+    if (ops.size() == 1)
+      continue;
```
fabianschuiki, line 173: I think `arc.vectorize` currently only supports a single result in its body region. (MLIR doesn't support VariadicOfVariadic results yet.) It would make sense to skip any ops that have more than one result here, until we have extended `arc.vectorize` to support multi-result ops.

fabianschuiki (commented):
This already looks pretty nice! Some comments inline below.

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    // Add the block arguments 
+    // comb.and %x, %y
+    // comb.and %u, %v
+    // then the operands' vector will be {{x, y}, {u, v}}
+    // we need to create an th block args, so we need the type and the location
+    // the type is a vectpr type
+    for (auto operands : operandsVector) {
+      mlir::Type vectorType = mlir::VectorType::get({-1}, operands.front().getType());
+      auto arg = vectorizeBlock.addArgument(vectorType, ops[0]->getLoc());
```
elhewaty, line 204: @fabianschuiki After this line I think I have the block arguments. but how to bind the block args with the outer variables(ops).

```
%0 = comb.add %x, %y : i4
%1 = comb.add %a, %b : i4
```
after this line I think I will have two vectors {x, y} and {a, b} and the argument type for the `arc.vectorize` block.
```
%2 = arc.vectorize(%x, %y), (%a, %b) : (i4, i4, i4, i4) -> (i4, i4) {
^bb0(%arg0: vector<2xi4>, %arg1: vector<2xi4>) {
   %3 = comb.add %arg0, %arg1 // here, how to bind the vectors to arg0, and arg1??
   arc.output %3: vector<2xi4> 
}
}
```
Is there anything wrong with this?

fabianschuiki: Are you sure you have vectors `{x, y}` and `{a, b}`? Looking at your code snippet, `x` and `a` appear on the LHS of the `comb.add`, and `y` and `b` appear on the RHS. If you collapse the two adds into a vectorized version, you'd want to have vector `{x, a}` on the LHS and `{y, b}` on the RHS.

In the `arc.vectorize` op, your block arguments don't need to have a vector type. Instead, the body of `arc.vectorize` is a scalar description of what should happen on _each_ vector lane. Something like this:
```mlir
%0, %1 = arc.vectorize (%x, %a), (%y, %b) : (i4, i4, i4, i4) -> (i4, i4) {
^bb0(%arg0: i4, %arg1: i4):
   %2 = comb.add %arg0, %arg1 : i4
   arc.vectorize.return %2 : i4
}
```

The vectorize op performs a "zip" over the operand groups and feeds each tuple of operands, `(x, y)` and `(a, b)` in your case, to its body. The body performs its scalar operation and returns the scalar result. The vectorize op then returns all these results. It's more or less like a `map` operation, where the scalar body is applied to each of the vector lanes.

This has the advantage of allowing us to vectorize the operands flowing into and results flowing out of the vectorize op at a later point. And, independently, to vectorize the body at a later point. That decouples forming the vectors from performing any kind of lowering or vectorization of the actual ops.

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
elhewaty:
```
using Key = std::tuple<unsigned, StringRef, SmallVector<Type, 4>,
                       SmallVector<Type, 4>, ArrayRef<NamedAttribute>>;

Key computeKey(Operation *op, unsigned rank) {
  SmallVector<Type, 4> operandTypes(op->operand_type_begin(), op->operand_type_end());
  SmallVector<Type, 4> resultTypes(op->result_type_begin(), op->result_type_end());
  // The key = concat(op_rank, op_name, op_operands_types, op_result_types, op_attrs)
  return std::make_tuple(rank, op->getName(), operandTypes, resultTypes, op->getAttrs());
}
```
```
error: no viable conversion from returned value of type 'tuple<[...], typename __decay_and_strip<OperationName>::__type, [3 * ...]>' to function return type 'tuple<[...], StringRef, [3 * ...]>'
  141 |   return std::make_tuple(rank, op->getName(), operandTypes, resultTypes, op->getAttrs());
```

any hint?

include/circt/Dialect/Arc/ArcPasses.td
```diff
@@ -58,6 +58,13 @@ def Dedup : Pass<"arc-dedup", "mlir::ModuleOp"> {
   ];
 }
 
+def FindInitialVectors : Pass<"find-initial-vectors", "mlir::ModuleOp"> {
+  let summary = "Finds the ops that can be grouped together into a vector";
+  let constructor = "circt::arc::createFindInitialVectorsPass()";
+  let dependentDialects = ["arc::ArcDialect", "comb::CombDialect",
+                           "hw::HWDialect"];
```
fabianschuiki, line 65: The dependent dialects field lists the dialects of operations that your pass creates in some `builder.create<...>` call. If I haven't overlooked anything, your pass just creates `arc.*` ops, so you should be fine with just the Arc dialect here:
```suggestion
  let dependentDialects = ["arc::ArcDialect"];
```

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
```diff
@@ -0,0 +1,254 @@
...
+#define GEN_PASS_DEF_FINDINITIALVECTORS
+#include "circt/Dialect/Arc/ArcPasses.h.inc"
+} // namespace arc
+} // namespace circt
+
+using namespace circt;
+using namespace arc;
+using llvm::SmallMapVector;
+using llvm::SmallVector;
+using mlir::Operation;
```
fabianschuiki, line 47: You may be able to delete these. I think we have re-exports for these in the `circt` namespace.
```suggestion
```

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+Key computeKey(Operation *op, unsigned rank) {
+  // The key = concat(op_rank, op_name, op_operands_types, op_result_types,
+  //                  op_attrs)
+  return std::make_tuple(
+      rank, op->getName().getStringRef(),
+      SmallVector<Type>(op->operand_type_begin(), op->operand_type_end()),
+      SmallVector<Type>(op->result_type_begin(), op->result_type_end()),
+      op->getAttrs());
+}
```
fabianschuiki, line 115: I think you can directly use the attribute dictionary instead of an array of named attributes. That makes the key cheaper to compare: the dictionary is just a pointer (interned in the MLIRContext), so it's just a pointer comparison:
```suggestion
using Key = std::tuple<unsigned, StringRef, SmallVector<Type>,
                       SmallVector<Type>, DictionaryAttr>;

Key computeKey(Operation *op, unsigned rank) {
  // The key = concat(op_rank, op_name, op_operands_types, op_result_types,
  //                  op_attrs)
  return std::make_tuple(
      rank, op->getName().getStringRef(),
      SmallVector<Type>(op->operand_type_begin(), op->operand_type_end()),
      SmallVector<Type>(op->result_type_begin(), op->result_type_end()),
      op->getAttrDictionary());
}
```

fabianschuiki (approved):
Nice! This looks very cool :sunglasses: :partying_face:

elhewaty:
I am working on adding the statistics too.

PR #7066, opened 2024-05-21
[ImportVerilog] Support set membership operator.

angelzzzzz:
Use `==` to compare the left-hand expression with expressions and use `<=`, `>=` to compare the left-hand expression with ranges.
Then, according to the IEEE 1800-2017 § 11.4.13 "Set membership operator", "The return value is effectively the or reduction of all the comparisons in the set with the expression on the left-hand side.", It will perform `or` operation on all comparison results above, and the result of `or` is the result of set membership operator.

### DIFF
--- a/test/Conversion/ImportVerilog/errors.sv
+++ b/test/Conversion/ImportVerilog/errors.sv
@@ -177,3 +177,22 @@ module Foo;
     automatic int b;
   end
 endmodule
+
+// -----
+
+module Foo;
+  int a, b, c;
+  bit d [3:0];
+  initial begin
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = 'x inside { a };
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = a inside { 'z, a };
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = a inside { a, ['x:b] };
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = a inside { a, [b:'z] };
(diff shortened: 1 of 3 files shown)


fabianschuiki (commented):
Very cool! Thanks for tackling this 🥳

angelzzzzz:
@fabianschuiki Thank you for being so patient and supportive! :rose: I have followed your suggestion and added some error tests. But I can't find an example to print "only simple bit vectors supported in 'inside' expressions" for now, since `string`, or `class`, or `pointer` have not been supported yet. So I didn't add this error case.
Looking forward to your suggestions. :smiley:

test/Conversion/ImportVerilog/errors.sv
```diff
@@ -177,3 +177,22 @@ module Foo;
...
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = 'x inside { a };
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = a inside { 'z, a };
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = a inside { a, ['x:b] };
+    // expected-error @below {{literals with X or Z bits not supported}}
+    c = a inside { a, [b:'z] };
+    // expected-error @below {{unpacked arrays in 'inside' expressions not supported}}
+    c = a inside { d };
```
fabianschuiki, line 196: Nice! :slightly_smiling_face:

fabianschuiki (approved):
This looks really great! Thanks for adding this :partying_face:

PR #7067, opened 2024-05-21
[FIRRTL] Convert Wires into Nodes

darthscsi:
This converts wires into nodes when there is one write to the wire and it dominates the reads.  By converting to nodes, this pass does not have to worry about symbols, references, or annotations.  Those are just copied to the node.


darthscsi:
On some of our largest designs (22GB+), this takes less than 16 seconds of user time.

lib/Dialect/FIRRTL/Transforms/EliminateWires.cpp
```diff
@@ -0,0 +1,88 @@
...
+};
+} // end anonymous namespace
+
+// This is the main entrypoint for the lowering pass.
+void EliminateWiresPass::runOnOperation() {
+  LLVM_DEBUG(debugPassHeader(this) << "\n";);
+  auto module = getOperation();
+  auto &dominance = getAnalysis<mlir::DominanceInfo>();
+
+  std::deque<std::pair<WireOp, StrictConnectOp>> worklist;
```
uenoku, line 47: nit: Can we use SmallVector if deque is unnecessary?

lib/Dialect/FIRRTL/Transforms/EliminateWires.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  for (auto wire : module.getOps<WireOp>()) {
```
uenoku, line 49: Is it intentional to walk only wires in the top-level?

uenoku (approved):
Looks great to me!

lib/Dialect/FIRRTL/Transforms/EliminateWires.cpp
darthscsi: Yes, other levels should be dead, but even if not, avoiding them simplifies dominance issues.

lib/Dialect/FIRRTL/Transforms/EliminateWires.cpp
darthscsi: Namely, you can push to it without ever having to move existing elements, while still having decent cache performance from array-like layout.  This makes it quite useful for worklists.

lib/Dialect/FIRRTL/Transforms/EliminateWires.cpp
```diff
@@ -0,0 +1,88 @@
...
+// Pass Infrastructure
+//===----------------------------------------------------------------------===//
+
+namespace {
+struct EliminateWiresPass : public EliminateWiresBase<EliminateWiresPass> {
+  void runOnOperation() override;
+};
+} // end anonymous namespace
+
+// This is the main entrypoint for the lowering pass.
```
dtzSiFive, line 41:
```suggestion
// This is the main entrypoint for the pass.
```

test/Dialect/FIRRTL/eliminate-wires.mlir
```diff
@@ -0,0 +1,30 @@
+// RUN: circt-opt -pass-pipeline='builtin.module(firrtl.circuit(firrtl.module(firrtl-eliminate-wires)))' --allow-unregistered-dialect %s | FileCheck --check-prefixes=CHECK %s
```
dtzSiFive, line 1:
```suggestion
// RUN: circt-opt -pass-pipeline='builtin.module(firrtl.circuit(firrtl.module(firrtl-eliminate-wires)))' --allow-unregistered-dialect %s | FileCheck %s
```

test/Dialect/FIRRTL/eliminate-wires.mlir
```diff
@@ -0,0 +1,30 @@
...
+    %b = firrtl.wire : !firrtl.uint<3>
+    %invalid_ui3 = firrtl.invalidvalue : !firrtl.uint<3>
+    firrtl.strictconnect %b, %invalid_ui3 : !firrtl.uint<3>
+    firrtl.strictconnect %a, %b : !firrtl.uint<3>
+    // CHECK: %[[inv:.*]] = firrtl.invalidvalue : !firrtl.uint<3>
+    // CHECK-NEXT:  %b = firrtl.node %[[inv]] : !firrtl.uint<3>
+    // CHECK-NEXT:  %a = firrtl.node %b : !firrtl.uint<3>
+  }
+
+}
```
dtzSiFive, line 30:
```suggestion
}

```

lib/Dialect/FIRRTL/Transforms/EliminateWires.cpp
dtzSiFive: `getSingleConnectUserOf` also only returns connect's from same block, which is important to avoid considering a connect out from a `when` (even if the connect dominates its users) as unconditional.

dtzSiFive (approved):
LGTM, thanks!

Please split out unrelated changes like those to DropName and the FIRRTLOpInterface.

PR #7068, opened 2024-05-21
[Support] Print timestamp for logging

uenoku:
This PR changes VerbosePassInstrumentation to print timestamp along with logging message.
```
[2024-05-21 07:19:23:firtool]   Running "firrtl-lower-open-aggs"
[2024-05-21 07:19:23:firtool]   -- Done in 0.000 sec
[2024-05-21 07:19:23:firtool]   Running "firrtl-resolve-paths"
[2024-05-21 07:19:23:firtool]   -- Done in 0.000 sec
[2024-05-21 07:19:23:firtool]   Running "firrtl-lower-annotations{allow-adding-ports-on-public-modules=false disable-annotation-classless=false disable-annotation-unknown=false no-ref-type-ports=false}"
[2024-05-21 07:19:23:firtool]   -- Done in 0.000 sec
[2024-05-21 07:19:23:firtool]   Running "firrtl-lower-intmodules{fixup-eicg-wrapper=false}"
```

### DIFF
--- a/include/circt/Support/Passes.h
+++ b/include/circt/Support/Passes.h
@@ -14,6 +14,10 @@
 #include "mlir/Pass/PassInstrumentation.h"
 #include "llvm/Support/Chrono.h"
 #include "llvm/Support/Format.h"
+#include "llvm/Support/Timer.h"
+
+#include <iomanip>
+#include <sstream>
 
 namespace circt {
 // This class prints logs before and after of pass executions when its pass
@@ -28,13 +32,22 @@ class VerbosePassInstrumentation : public mlir::PassInstrumentation {
   int level = 0;
   const char *toolName;
 
+  void emitToolNameAndTimestamp(llvm::raw_fd_ostream &os, TimePoint &now) {
+    auto time = std::chrono::system_clock::to_time_t(now);
+    std::stringstream ss;
+    ss << std::put_time(std::localtime(&time), "%Y-%m-%d %X");
+
+    os << "[" << ss.str() << ":" << toolName << "] ";
+  }
+
 public:
   VerbosePassInstrumentation(const char *toolName) : toolName(toolName){};
   void runBeforePass(Pass *pass, Operation *op) override {
     if (isa<LoggedOpTypes...>(op)) {
-      timePoints.push_back(TimePoint::clock::now());
+      auto now = TimePoint::clock::now();
+      timePoints.push_back(now);
       auto &os = llvm::errs();
-      os << llvm::format("[%s] ", toolName);
+      emitToolNameAndTimestamp(os, now);
       os.indent(2 * level++);
       os << "Running \"";
       pass->printAsTextualPipeline(llvm::errs());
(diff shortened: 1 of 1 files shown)


include/circt/Support/Passes.h
```diff
@@ -28,13 +32,22 @@ class VerbosePassInstrumentation : public mlir::PassInstrumentation {
   int level = 0;
   const char *toolName;
 
+  void emitToolNameAndTimestamp(llvm::raw_fd_ostream &os, TimePoint &now) {
+    auto time = std::chrono::system_clock::to_time_t(now);
+    std::stringstream ss;
+    ss << std::put_time(std::localtime(&time), "%Y-%m-%d %X");
```
seldridge, line 38: Can this include `%Z` to show the time zone, too? These logs _may_ be confusing or require users to guess at when something happened if they are inspecting a log from somebody else that is using a different locale.

include/circt/Support/Passes.h
```diff
@@ -28,13 +32,22 @@ class VerbosePassInstrumentation : public mlir::PassInstrumentation {
...
+
+    os << "[" << ss.str() << ":" << toolName << "] ";
```
seldridge, line 40: Mega-nit: how does just a space look? The ":" looked crowded to me and other logging systems, e.g., journalctl use a space here.

seldridge (approved):
This will help a lot with debugging!

One alternative that might make sense for our use case is to show the start time and then only show offsets from it. `dmesg` does this for `-H`, though it will give you a new time when the time lines up with a minute.

For us, it may be useful to show something like:

```
[May 21, 11:05:42] Parsing Modules
[        +0:05:13] Parsing Module Bodies
```

This is not a request to change this. Only a dump of ideas on how to present the information. What you have in the PR looks great.

prithayan (approved):
Looks good!

fabianschuiki (approved):
Very cool!

dtzSiFive (approved):
LGTM.

Often this is done by part of a logging subsystem or a different tool does it, to separate out this functionality as well as systematically ensure timestamps appear on all lines (or per logical "message"), FWIW.

This seems reasonable to me.

include/circt/Support/Passes.h
dtzSiFive: Interesting.  Maybe just print UTC?  I don't wanna wrangle timezones.

Especially when run in containers, namespaces, so on, these probably should not be expected to reflect a particular locale or even comparable absolute time -- basically don't trust system clocks (might not even be monotonic!) or use them for "happened-before" relations across different clocks (different machines / environments).

That said, not 100% what the goal is here re:printing the system clock occasionally, especially as part of our tool?
Would just printing the system clock at the start communicate the same information?

uenoku: The request was to print timestamps along with messages so users can know "when the last step started" easily. I don't have a strong opinion about the format or timezone. Maybe it makes sense to make the format itself CLI option?

PR #7069, opened 2024-05-21
[FIRRTL] Register reset elimination based on invalid can look through nodes.

darthscsi:
It should be time to make this change.

Builds on #7067


test/Dialect/FIRRTL/sfc-compat.mlir
```diff
@@ -93,18 +93,6 @@ firrtl.circuit "SFCCompatTests" {
...
-    %inv = firrtl.wire  : !firrtl.uint<8>
-    %invalid_ui8 = firrtl.invalidvalue : !firrtl.uint<8>
-    firrtl.connect %inv, %invalid_ui8 : !firrtl.uint<8>, !firrtl.uint<8>
-    %_T = firrtl.node %inv  : !firrtl.uint<8>
-    // CHECK: firrtl.regreset %clock
-    %r = firrtl.regreset %clock, %reset, %_T  : !firrtl.clock, !firrtl.uint<1>, !firrtl.uint<8>, !firrtl.uint<8>
-    firrtl.connect %r, %d : !firrtl.uint<8>, !firrtl.uint<8>
-    firrtl.connect %q, %r : !firrtl.uint<8>, !firrtl.uint<8>
-  }
-
```
seldridge, line 107: This should probably now be a positive test.

lib/Dialect/FIRRTL/Transforms/SFCCompat.cpp
```diff
@@ -75,7 +75,7 @@ void SFCCompatPass::runOnOperation() {
     // If the `RegResetOp` has an invalidated initialization and we
     // are not running FART, then replace it with a `RegOp`.
     if (!fullAsyncResetExists &&
-        walkDrivers(reg.getResetValue(), true, false, false,
+        walkDrivers(reg.getResetValue(), true, true, false,
```
seldridge, line 78: This is my fault originally... this would be better with the `/*name=*/` pseudo-named arguments so it's clear what these are.

seldridge (approved):
I can't think of a reason _not_ to try this. This was always a questionable choice with the SFC definitively in the rear view mirror at this point.

Let's try it. Just please run some designs and see what the changes are!

dtzSiFive (commented):
LGTM but seems to pull in a bunch of unrelated changes, and modifying the pipeline is not part of changing how SFCCompat is able to look through things.

There does not appear to be any reason this change should build on that other PR / new pass.

Makes sense that as part of adding the pass experimentally to the pipeline this was encountered as an issue, but please change the behavior and split out the other changes.

darthscsi:
On a very large design, appears to not affect the output.

PR #7072, opened 2024-05-21
[LowerClasses] Ensure classes are instantiated by an object.

prithayan:
`LowerClass` creates an object for a corresponding instance only if the instantiated module has property ports. 
But a class can be created for a corresponding module based on other conditions like, if the module is  public, or instantiates other classes.
This results in un-instantiated classes that donot correspond to the module hierarchy.
This change ensures that if a class is created for a module, the object is also created from the corresponding instance.
Thus the module hierarchy is also preserved in the `om` IR.
Downstream tools parsing the  IR can assume a single top level class which is required for object model evaluation.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -198,6 +198,9 @@ private:
                              SymbolTable &symbolTable);
 
   // Predicate to check if a module-like needs a Class to be created.
+  bool shouldCreateClass(StringAttr modName);
+
+  // Predicate to check if a module-like needs a Class to be created.
   bool shouldCreateClass(FModuleLike moduleLike);
 
   // Create an OM Class op from a FIRRTL Class op.
@@ -222,7 +225,7 @@ private:
       const DenseMap<StringAttr, firrtl::ClassType> &classTypeTable);
 
   // State to memoize repeated calls to shouldCreateClass.
-  DenseMap<FModuleLike, bool> shouldCreateClassMemo;
+  DenseMap<StringAttr, bool> shouldCreateClassMemo;
 };
 
 struct PathTracker {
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -764,10 +767,16 @@ std::unique_ptr<mlir::Pass> circt::firrtl::createLowerClassesPass() {
...
 }
 
+// Predicate to check if a module-like needs a Class to be created.
+bool LowerClassesPass::shouldCreateClass(StringAttr modName) {
+  // Return a memoized result.
+  return shouldCreateClassMemo.at(modName);
+}
+
 // Predicate to check if a module-like needs a Class to be created.
 bool LowerClassesPass::shouldCreateClass(FModuleLike moduleLike) {
```
mikeurbach, line 777: Any specific reason not to change this predicate, rather than defining a new predicate with a different signature?

mikeurbach (commented):
This makes sense to me.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -882,13 +891,16 @@ void LowerClassesPass::lowerClass(om::ClassOp classOp, FModuleLike moduleLike,
     auto propertyOperands = llvm::any_of(op.getOperandTypes(), [](Type type) {
       return isa<PropertyType>(type);
     });
+    bool needsClone = false;
+    if (auto instance = dyn_cast<InstanceOp>(op))
+      needsClone = shouldCreateClass(instance.getReferencedModuleNameAttr());
```
uenoku, line 896: I think you can lookup FModuleLike through InstanceGraph and probably we don't have to change the type

prithayan: This can avoid avoid an unnecessary instance graph lookup, as the module op is not required here. Querying just using the name seems better here.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
prithayan: This was an oversight, updated the signature .

mikeurbach (approved):
I think this is good to go.

PR #7074, opened 2024-05-22
[FIRRTL] Limited invalid propagation.

darthscsi:
Now that invalid is better defined, we can start propagating it when we can preserver semantics.  This means not breaking the one-op-is-one-consistent-value (e.g. not duplicating).  The only bit mutating op which is touched by this change is not, which is a pure all-bits inversion, which should be safe (tm).

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -3184,10 +3184,17 @@ LogicalResult InvalidValueOp::canonicalize(InvalidValueOp op,
   }
   // Propagate invalids through a single use which is a unary op.  You cannot
   // propagate through multiple uses as that breaks invalid semantics.  Nor
-  // can you propagate through binary ops.
+  // can you propagate through binary ops or generally any op which computes.
+  // Not is an exception as it is a pure, all-bits inverse.
   if (op->hasOneUse() &&
-      isa<BitsPrimOp, HeadPrimOp, ShrPrimOp, TailPrimOp, SubfieldOp, SubindexOp,
-          AsSIntPrimOp, AsUIntPrimOp, NotPrimOp>(*op->user_begin())) {
+      (isa<BitsPrimOp, HeadPrimOp, ShrPrimOp, TailPrimOp, SubfieldOp,
+           SubindexOp, AsSIntPrimOp, AsUIntPrimOp, NotPrimOp, BitCastOp>(
+           *op->user_begin()) ||
+       (isa<CvtPrimOp>(*op->user_begin()) &&
+        type_isa<SIntType>(op->user_begin()->getOperand(0).getType())) ||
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -3182,6 +3182,20 @@ LogicalResult InvalidValueOp::canonicalize(InvalidValueOp op,
     rewriter.eraseOp(op);
     return success();
   }
+  // Propagate invalids through a single use which is a unary op.  You cannot
+  // propagate through multiple uses as that breaks invalid semantics.  Nor
+  // can you propagate through binary ops.
+  if (op->hasOneUse() &&
+      isa<BitsPrimOp, HeadPrimOp, ShrPrimOp, TailPrimOp, SubfieldOp, SubindexOp,
+          AsSIntPrimOp, AsUIntPrimOp, NotPrimOp>(*op->user_begin())) {
```
seldridge, line 3190: There are a few more unwary ops that can be included: cvt, pad, asClock, and the reduction ops.

darthscsi: Pad adds bits, so it isn't safe, cvt interprets the bits and can add bits, so likewise.  asClock is probably safe.

darthscsi: But reductions are a good call.

darthscsi: oddly, a reduction of a zero-width value is NOT invalid, even though all other widths are.

test/Dialect/FIRRTL/canonicalization.mlir
```diff
@@ -3097,6 +3097,106 @@ firrtl.module @RemoveUnusedInvalid() {
...
+
+  // CHECK-NOT firrtl.xorr
+  %invc = firrtl.invalidvalue : !firrtl.uint<4>
+  %xorr = firrtl.xorr %invc : (!firrtl.uint<4>) -> !firrtl.uint<1>
+  firrtl.strictconnect %out1, %xorr : !firrtl.uint<1>
+
+  // CHECK: firrtl.invalidvalue
+  // CHECK-NEXT: firrtl.strictconnect
+
+}
```
seldridge, line 3206: Missing positive/negative tests of `cvt`.

seldridge (approved):
This is legal per the spec. Can you check the impact on a real design? 🧐 

I'm also starting to question if CSE involving invalid value ops is going to make much of this rare to canonicalize.

Re: rarity: do we propagate invalid through single connects with your new wire to node conversion? It's pretty difficult to create an invalid in Chisel as it has to be of a component. Practically, this usually shows up on only ports or wires and both of these are funneling invalid through a connect.

darthscsi:
Doesn't seem to have any effect on at least one large design.

PR #7076, opened 2024-05-22
[FIRRTL] Remove unused and expensive API, NFC

uenoku:
InstanceOp::getPortList is a foot-gun that walks entire circuits. We could rewrite it with instance graph version but apparently there is no user so just delete it.


seldridge (approved):

darthscsi (approved):
So long as we aren't inheriting the PortList trait (indirectly), this is fine.  The portlist trait needs to drop this method too.  I think there is a note to that effect, but some of the smaller dialects were using it.

uenoku:
InstanceOp doesn't inherit PortList trait (otherwise I think the build will fail).

PR #7077, opened 2024-05-22
[FIRRTL] Preserve all analysis if nothing happens

uenoku:
LowerMatchesPass and MemOfVec should preserve analysis if nothing happens.


dtzSiFive (approved):
LGTM, thanks!

LowerMatch could grow a statistic to count how many it lowers in the future, if we wanted.  Not for this PR.

PR #7082, opened 2024-05-23
[Moore] Introduce Mem2Reg to eliminate local variables

hailongSun2000:
### First
```
module Foo;
  int x, y;
  always_comb begin
    int a;
    a = x + 1;
    y = a;
  end
endmodule
```
```cpp
/home/phoenix/work/HDLBits/test01.sv:2:7: error: all variables: %0 = "moore.variable"() <{name = "x"}> : () -> !moore.i32
  int x, y;
      ^
/home/phoenix/work/HDLBits/test01.sv:2:10: error: all variables: %1 = "moore.variable"() <{name = "y"}> : () -> !moore.i32
  int x, y;
         ^
/home/phoenix/work/HDLBits/test01.sv:4:9: error: all variables: %2 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    int a;
        ^
/home/phoenix/work/HDLBits/test01.sv:4:9: error: local varialbes: %2 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    int a;
        ^
/home/phoenix/work/HDLBits/test01.sv:5:5: error: blocking assignments: %2 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    a = x + 1;
    ^
/home/phoenix/work/HDLBits/test01.sv:6:5: error: blocking assignments: %2 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    y = a;
    ^
module {
  moore.module @Foo {
    %x = moore.variable  : !moore.i32
    %y = moore.variable  : !moore.i32
    moore.procedure always_comb {
      %a = moore.variable  : !moore.i32
      %0 = moore.constant 1 : !moore.i32
      %1 = moore.add %x, %0 : !moore.i32
      moore.blocking_assign %a, %1 : !moore.i32
      moore.blocking_assign %y, %a : !moore.i32
    }
  }
}
```
----
# Second
```
module Foo;
  int x, y;
  always_comb begin
    int a;
    a = x + 1;
   // y = a;
  end
endmodule
```
``` cpp
/home/phoenix/work/HDLBits/test01.sv:2:7: error: all variables: %0 = "moore.variable"() <{name = "x"}> : () -> !moore.i32
  int x, y;
      ^
/home/phoenix/work/HDLBits/test01.sv:2:10: error: all variables: %1 = "moore.variable"() <{name = "y"}> : () -> !moore.i32
  int x, y;
         ^
/home/phoenix/work/HDLBits/test01.sv:4:9: error: all variables: %2 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    int a;
        ^
/home/phoenix/work/HDLBits/test01.sv:4:9: error: local varialbes: %2 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    int a;
        ^
/home/phoenix/work/HDLBits/test01.sv:5:5: error: blocking assignments: %2 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    a = x + 1;
    ^
/home/phoenix/work/HDLBits/test01.sv:4:9: error: slot: %3 = "moore.variable"() <{name = "a"}> : () -> !moore.i32
    int a;
        ^
/home/phoenix/work/HDLBits/test01.sv:2:7: error: all variables: %0 = "moore.variable"() <{name = "x"}> : () -> !moore.i32
  int x, y;
      ^
/home/phoenix/work/HDLBits/test01.sv:2:10: error: all variables: %1 = "moore.variable"() <{name = "y"}> : () -> !moore.i32
  int x, y;
         ^
module {
  moore.module @Foo {
    %x = moore.variable  : !moore.i32
    %y = moore.variable  : !moore.i32
    moore.procedure always_comb {
      %0 = moore.constant 1 : !moore.i32
      %1 = moore.add %x, %0 : !moore.i32
    }
  }
}
```

### DIFF
--- a/lib/Dialect/Moore/MooreOps.cpp
+++ b/lib/Dialect/Moore/MooreOps.cpp
@@ -19,6 +19,7 @@
 
 using namespace circt;
 using namespace circt::moore;
+using namespace mlir;
 
 //===----------------------------------------------------------------------===//
 // SVModuleOp
@@ -237,6 +238,33 @@ void VariableOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
   setNameFn(getResult(), getName());
 }
 
+llvm::SmallVector<MemorySlot> VariableOp::getPromotableSlots() {
+  if (isa<SVModuleOp>(getOperation()->getParentOp()))
+    return {};
+  return {MemorySlot{getResult(), getType()}};
+}
+
+Value VariableOp::getDefaultValue(const MemorySlot &slot, OpBuilder &builder) {
+  if (auto value = getInitial())
+    return value;
+  return builder.create<ConstantOp>(
+      getLoc(),
+      cast<moore::IntType>(cast<RefType>(slot.elemType).getNestedType()), 0);
+}
+
+void VariableOp::handleBlockArgument(const MemorySlot &slot,
+                                     BlockArgument argument,
+                                     OpBuilder &builder) {}
+
+::std::optional<::mlir::PromotableAllocationOpInterface>
+VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
+                                    OpBuilder &builder) {
+  if (defaultValue.use_empty())
+    defaultValue.getDefiningOp()->erase();
+  this->erase();
+  return std::nullopt;
+}
+
 //===----------------------------------------------------------------------===//
 // NetOp
 //===----------------------------------------------------------------------===//
@@ -417,6 +445,78 @@ OpFoldResult BoolCastOp::fold(FoldAdaptor adaptor) {
   if (getInput().getType() == getResult().getType())
     return getInput();
   return {};
+}
+
+//===----------------------------------------------------------------------===//
+// BlockingAssignOp
+//===----------------------------------------------------------------------===//
+
+bool BlockingAssignOp::loadsFrom(const MemorySlot &slot) { return false; }
+
+bool BlockingAssignOp::storesTo(const MemorySlot &slot) {
+  return getDst() == slot.ptr;
+}
+
+Value BlockingAssignOp::getStored(const MemorySlot &slot, OpBuilder &builder,
+                                  Value reachingDef,
+                                  const DataLayout &dataLayout) {
+  return getSrc();
+}
+
+bool BlockingAssignOp::canUsesBeRemoved(
+    const MemorySlot &slot, const SmallPtrSetImpl<OpOperand *> &blockingUses,
+    SmallVectorImpl<OpOperand *> &newBlockingUses,
+    const DataLayout &dataLayout) {
+
+  if (blockingUses.size() != 1)
+    return false;
+  Value blockingUse = (*blockingUses.begin())->get();
+  return blockingUse == slot.ptr && getDst() == slot.ptr &&
+         getSrc() != slot.ptr &&
+         getSrc().getType() == cast<RefType>(slot.elemType).getNestedType();
+}
+
+DeletionKind BlockingAssignOp::removeBlockingUses(
+    const MemorySlot &slot, const SmallPtrSetImpl<OpOperand *> &blockingUses,
+    OpBuilder &builder, Value reachingDefinition,
+    const DataLayout &dataLayout) {
+  return DeletionKind::Delete;
+}
+
+//===----------------------------------------------------------------------===//
+// ReadOp
+//===----------------------------------------------------------------------===//
+
+bool ReadOp::loadsFrom(const MemorySlot &slot) {
+  return getOperand() == slot.ptr;
+}
+
+bool ReadOp::storesTo(const MemorySlot &slot) { return false; }
+
+Value ReadOp::getStored(const MemorySlot &slot, OpBuilder &builder,
+                        Value reachingDef, const DataLayout &dataLayout) {
+  llvm_unreachable("getStored should not be called on ReadOp");
+}
+
+bool ReadOp::canUsesBeRemoved(const MemorySlot &slot,
+                              const SmallPtrSetImpl<OpOperand *> &blockingUses,
+                              SmallVectorImpl<OpOperand *> &newBlockingUses,
+                              const DataLayout &dataLayout) {
+
+  if (blockingUses.size() != 1)
+    return false;
+  Value blockingUse = (*blockingUses.begin())->get();
+  return blockingUse == slot.ptr && getOperand() == slot.ptr &&
+         getResult().getType() == cast<RefType>(slot.elemType).getNestedType();
+}
+
+DeletionKind
+ReadOp::removeBlockingUses(const MemorySlot &slot,
+                           const SmallPtrSetImpl<OpOperand *> &blockingUses,
+                           OpBuilder &builder, Value reachingDefinition,
+                           const DataLayout &dataLayout) {
+  getResult().replaceAllUsesWith(reachingDefinition);
+  return DeletionKind::Delete;
 }
 
 //===----------------------------------------------------------------------===//
--- a/tools/circt-verilog/circt-verilog.cpp
+++ b/tools/circt-verilog/circt-verilog.cpp
@@ -13,10 +13,12 @@
 //===----------------------------------------------------------------------===//
 
 #include "circt/Conversion/ImportVerilog.h"
+#include "circt/Conversion/MooreToCore.h"
 #include "circt/Support/Version.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Support/FileUtilities.h"
+#include "mlir/Transforms/Passes.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/InitLLVM.h"
 #include "llvm/Support/SourceMgr.h"
@@ -31,7 +33,14 @@ using namespace circt;
 //===----------------------------------------------------------------------===//
 
 namespace {
-enum class LoweringMode { OnlyPreprocess, OnlyLint, OnlyParse, Full };
+enum class LoweringMode {
+  OnlyPreprocess,
+  OnlyLint,
+  OnlyParse,
+  OutputIRMoore,
+  OutputIRHW,
+  Full
+};
 
 struct CLOptions {
   cl::OptionCategory cat{"Verilog Frontend Options"};
@@ -60,7 +69,13 @@ struct CLOptions {
                      "CIRCT IR"),
           clEnumValN(LoweringMode::OnlyParse, "parse-only",
                      "Only parse and elaborate the input, without mapping to "
-                     "CIRCT IR")),
+                     "CIRCT IR"),
+          clEnumValN(LoweringMode::OutputIRMoore, "ir-moore",
+                     "Run the entire pass manager to just before MooreToCore "
+                     "conversion, and emit the resulting Moore dialect IR"),
+          clEnumValN(LoweringMode::OutputIRHW, "ir-hw",
+                     "Run the MooreToCore conversion and emit the resulting "
+                     "core dialect IR")),
       cl::init(LoweringMode::Full), cl::cat(cat)};
 
   //===--------------------------------------------------------------------===//
@@ -200,6 +215,23 @@ struct CLOptions {
 
 static CLOptions opts;
 
+/// Parse specified files that had been translated into Moore dialect IR. After
+/// that simplify these files like deleting local variables, and then emit the
+/// resulting Moore dialect IR .
+static LogicalResult populateMooreTransforms(mlir::PassManager &pm) {
+  pm.addPass(mlir::createMem2Reg());
+  // TODO: like dedup pass.
+
+  return success();
+}
+
+/// Convert Moore dialect IR into core dialect IR
+static LogicalResult populateMooreToCoreLowering(mlir::PassManager &pm) {
+  pm.addPass(createConvertMooreToCorePass());
+
+  return success();
+}
+
 //===----------------------------------------------------------------------===//
 // Implementation
 //===----------------------------------------------------------------------===//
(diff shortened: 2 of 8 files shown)


hailongSun2000:
I'm not sure the changes(like `a` is substituted with `x + 1`) occur at which stage. Thanks for anyone's help in advance.

fabianschuiki:
I've been digging into this a bit to see how Mem2Reg does its magic and why this isn't promoted. I've looked at the following:
```mlir
func.func @Foo() -> !moore.i32 {
  %x = moore.variable : !moore.i32
  %0 = moore.constant 42 : !moore.i32
  moore.blocking_assign %x, %0 : !moore.i32
  return %x : !moore.i32
}
```
My trace of Mem2Reg:
1. Collect `%x = moore.variable` into the initial set of allocators to promote
2. Go through the allocators to promote
3. Call `computeInfo` of `MemorySlotPromotionAnalyzer` on that `%x` variable
4. Call `computeBlockingUses` for `%x`
5. Visit all users of `%x` and make sure that they are all promotable
6. This visits `func.return %x`, which is not promotable, and aborts

So the problem is that we're using the variable `%x` directly in the return. This is definitely a design smell in the Moore dialect at the moment: we don't have any op that represents _reading_ from a variable. Most other IRs have dedicated _load_ and _store_ ops to interact with variables, and they only operations that are ever performed on a variable are loads and stores. For example, in MemRef you'd have something like this:

```mlir
func.func @Foo() -> i32 {
  %x = memref.alloca() : memref<i32>
  %0 = arith.constant 42 : i32
  memref.store %x, %0[] : memref<i32>
  %1 = memref.load %x[] : memref<i32>
  return %1 : i32
}
```

Instead of having variable/alloca directly return an `i32`, the MemRef dialect returns a reference type `memref<i32>`, which indicates that you're not dealing with an `i32` value itself, but rather with a _variable_/_memory reference_ that contains an `i32`. If you want to change `%x`, you have to use a `memref.store`. And if you want to know the value of `%x`, you have to use a `memref.load` to read its value.

We should probably do the same thing in the Moore dialect: `moore.variable` should probably return a `!moore.ref<i32>` type instead of `i32` directly. `moore.blocking_assign` would then require the LHS to be a ref type and act like a `memref.store`. And we'd need something like `moore.read_var` that takes a ref and returns the current value, like a `memref.load`. (I think we already have a `moore.read_lvalue` that signals a value read, just not with a ref/lvalue type wrapper yet.)

hailongSun2000:
I referred to `llvm. load` and `llvm. store`, their types are i32 or i* without a wrapper. I think `Mem2Reg` pass is aborted due to `%x` that doesn't be substituted with `%0`.

hailongSun2000:
I'm not sure whether can we use `moore.read_lvalue` to replace directly return a variable. Like the following image
image

I'll attempt to verify my thoughts.

Moxinilian:
Hey! I designed the mem2reg pass. Fabian’s analysis is correct, the pointer must only be used by promotable operations for any promotion to happen. In particular, returning the pointer is not allowed (as it would escape). You need a dedicated operation to read from the value. In the LLVM implementation, the slots are pointed to via an llvm.ptr, not an element value directly.

hailongSun2000:
Thanks @Moxinilian :heart: !  Whether can I understand that we should use dedicated ops like `llvm. load` to read and `llvm,store` to write the value and a type wrapper like `memref<i32>` rather than `i32`? We need dedicated ops to read and write explicitly right?

hailongSun2000:
@fabianschuiki :tada: 
```
module Foo;
  int x, y, z;
  always_comb begin
    int a;
    a = x + 1;
    y = a;
    a = a + 1;
    z = a;
  end
endmodule
```
```
module {
  moore.module @Foo {
    %x = moore.variable  : !moore.i32
    %y = moore.variable  : !moore.i32
    %z = moore.variable  : !moore.i32
    moore.procedure always_comb {
      %0 = moore.read_lvalue %x : !moore.i32
      %1 = moore.constant 1 : !moore.i32
      %2 = moore.add %0, %1 : !moore.i32
      %3 = moore.read_lvalue %y : !moore.i32               // Please ignore this IR
      moore.blocking_assign %y, %2 : !moore.i32
      %4 = moore.constant 1 : !moore.i32
      %5 = moore.add %2, %4 : !moore.i32
      %6 = moore.read_lvalue %z : !moore.i32               // Please ignore this IR
      moore.blocking_assign %z, %5 : !moore.i32
    }
  }
}
```

Moxinilian:
I would recommend a dedicated type yes. I think you may technically not absolutely need it because the MemorySlot would model it already, I think it would be better to make sure your slots do not accidentally escape like in our case.

Load/store operations on the other hand are absolutely necessary.

hailongSun2000:
The `deleteLocalVar` pass and a lot of redundant code is removed. And `Mem2Reg` only works when enabling `--ir-moore`/`--ir-hw`, so I haven't added the related tests for the time being. And I'm not sure calling `erase()` is whether reasonable.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -78,7 +78,10 @@ struct ExprVisitor {
 
   // Handle blocking and non-blocking assignments.
   Value visit(const slang::ast::AssignmentExpression &expr) {
-    auto lhs = context.convertExpression(expr.left());
+    auto *readOp = context.convertExpression(expr.left()).getDefiningOp();
+    auto lhs = dyn_cast<moore::ReadLValueOp>(readOp).getOperand();
+    readOp->erase();
+
```
hailongSun2000, line 84: There is no way to estimate when to create `moore.read_lvalue`. So call `op->erase()` to remove it if it is created by `expr.left()`.

fabianschuiki: I think this is a good idea!

I have a feeling that we are going to introduce a dedicated expression lowering for left-hand sides of assignments. This is done in a similar fashion in C/C++:
```c
int x, y;
x = y
```
would lower to something like the following pseudo-code:
```mlir
%x = alloca : ptr<int>
%y = alloca : ptr<int>
%0 = load %y  // y lowered as rvalue
store %x, %0  // x lowered as lvalue
```

If arrays or structs are involved, you'd have dedicated ops that allow you to convert a pointer/reference to the struct into a reference to the field you're interested in:
```c
struct { int a, b; } x, y;
x.b = y.a;
```
would lower to something like this:
```mlir
%x = alloca : ptr<struct<a: int, b: int>>
%y = alloca : ptr<struct<a: int, b: int>>
// x.b lowered as lvalue
%0 = struct_field_ref %x, "b" : ptr<struct<a: int, b: int>> -> ptr<int>
// y.a lowered as rvalue
%1 = load %y : ptr<struct<a: int, b: int>> -> struct<a: int, b: int>
%2 = struct_field %x, "a" : struct<a: int, b: int> -> int
// assignment
store %0, %2 : ptr<int>, int
```

These ops would allow us to reason about assignable lvalues that are just references to a storage location, and rvalues that are actual values. We would probably create two separate expression lowering functions, `context.convertLvalueExpression` and `context.convertRvalueExpression`: the former would always return a `ref<T>` type, and the latter would always return an actual value `T`.

fabianschuiki:
> In order to prevent the accident slot escape, we can add a wrapper for the type. What do @fabianschuiki think?

I think that's a fantastic idea! All types are now in ODS, so it should be pretty easy to create these simple wrapper types. We could start with a `ref<T>` type, and later think about whether we'd want to distinguish between nets and variables if that simplifies some of the lowering passes.

hailongSun2000:
I'm trying to add a new type named `RefType` as a wrapper type. This reminds me of `MooreLValueType`, `MooreRValueType`, and `VariableDeclOp` had been removed.

fabianschuiki:
> This reminds me of `MooreLValueType`, `MooreRValueType`, and `VariableDeclOp`

Yeah absolutely… I think we anticipated that we would need an lvalue/rvalue split at some point, but it got in the way of the initial lowering work back then. Now that we know exactly what we need, we're in a much better position to add a ref type back in that handles this 😃

hailongSun2000:
In my private branch, the `Mem2Reg` pass can work. But this PR relies on #7095. Because we add two options(`--ir-moore` and `--ir-hw`) for `circt-verilog`. Only enable `ir-moore`, the `Mem2Reg` can work. So I want to know how do I add the related test case? Thanks in advance! :smile:

fabianschuiki:
> In my private branch, the `Mem2Reg` pass can work.

That is great to hear :heart_eyes:! Maybe once #7095 lands, you can rebase this PR onto that work and get everything up and running?

hailongSun2000:
Yeah, right! I'll just turn this draft into the PR later!

hailongSun2000:
Hey, @fabianschuiki. I added a new test file(optimization.sv) to check whether Mem2Reg can work.

hailongSun2000:
> I would recommend a dedicated type yes. I think you may technically not absolutely need it because the MemorySlot would model it already, but I think it would be better to make sure your slots do not accidentally escape like in our case.
> 
> Load/store operations on the other hand are absolutely necessary.
> 
> I’ll give a proper review to your code later when I have time if you want.

Hey, @Moxinilian. We designed a dedicated (reference)type wrapper to ensure slots don't accidentally escape. You can rest assured. :smiley:

test/Conversion/ImportVerilog/optimization.sv
mingzheTerapines, line 1: THis test contains import verilog and mem2reg 2passes. Personally think you just need to check only mem2reg pass.
Here is a HW test which just test canonicalize.
```
// RUN: circt-opt -canonicalize='top-down=true region-simplify=true' %s | FileCheck %s

// CHECK-LABEL: hw.module @extract_noop(in %arg0 : i3, out "" : i3) {
// CHECK-NEXT:    hw.output %arg0

hw.module @extract_noop(in %arg0 : i3, out "": i3) {
  %x = comb.extract %arg0 from 0 : (i3) -> i3
  hw.output %x : i3
}

// Constant Folding

// CHECK-LABEL: hw.module @extract_cstfold(out result : i3) {
// CHECK-NEXT:    %c-3_i3 = hw.constant -3 : i3
// CHECK-NEXT:    hw.output  %c-3_i3

hw.module @extract_cstfold(out result : i3) {
  %c42_i12 = hw.constant 42 : i12
  %x = comb.extract %c42_i12 from 3 : (i12) -> i3
  hw.output %x : i3
}
```

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -237,6 +238,33 @@ void VariableOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
...
+
+void VariableOp::handleBlockArgument(const MemorySlot &slot,
+                                     BlockArgument argument,
+                                     OpBuilder &builder) {}
+
+::std::optional<::mlir::PromotableAllocationOpInterface>
+VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
+                                    OpBuilder &builder) {
+  if (defaultValue.use_empty())
+    defaultValue.getDefiningOp()->erase();
```
Moxinilian, line 263: This is not required, the default value is lazily constructed so if it is unused it will not have been created.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -419,6 +447,79 @@ OpFoldResult BoolCastOp::fold(FoldAdaptor adaptor) {
...
+  return blockingUse == slot.ptr && getDst() == slot.ptr &&
+         getSrc() != slot.ptr &&
+         getSrc().getType() == cast<RefType>(slot.elemType).getNestedType();
+}
+
+DeletionKind BlockingAssignOp::removeBlockingUses(
+    const MemorySlot &slot, const SmallPtrSetImpl<OpOperand *> &blockingUses,
+    OpBuilder &builder, Value reachingDefinition,
+    const DataLayout &dataLayout) {
+
```
Moxinilian, line 483: Seems to be an unwanted newline here.
```suggestion
```

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -200,6 +215,25 @@ struct CLOptions {
 
 static CLOptions opts;
 
+/// Parse specified files that had been translated into Moore dialect IR. After
+/// that simplify these files like deleting local variables, and then emit the
+/// resulting Moore dialect IR .
+static LogicalResult populateMooreTransforms(mlir::PassManager &pm) {
+
```
Moxinilian, line 222:
```suggestion
```

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -200,6 +215,25 @@ struct CLOptions {
...
+  pm.addPass(mlir::createMem2Reg());
+  // TODO: like dedup pass.
+
+  return success();
+}
+
+/// Convert Moore dialect IR into core dialect IR
+static LogicalResult populateLowMooreToCore(mlir::PassManager &pm) {
+
```
Moxinilian, line 231:
```suggestion
```

Moxinilian (commented):
Looks good to me beyond the newline nits and the test.

hailongSun2000:
Oh! Thanks! I'll be careful later.

test/Conversion/ImportVerilog/optimization.sv
fabianschuiki: I agree with @mingzheTerapines :slightly_smiling_face:! Since you are specifically checking whether mem2reg works on the Moore dialect, I would add a `mem2reg.mlir` test instead, use something like `// RUN: circt-opt --mem2reg %s | FileCheck %s` and use Moore dialect ops instead of SV inputs there.

The reasoning behind this is that you want a test that only checks the Verilog-to-Moore conversion, and a test that only checks whether mem2reg works on the Moore dialect. But you don't want to do both in the same file, since it makes tests very brittle and easy to break.

hailongSun2000: Hey, @mingzheTerapines @fabianschuiki.  I added the `mem2reg` to `circt-opt`, which only verifies whether the `mem2reg` works.

mingzheTerapines: This test and tool option looks perfect!

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -200,6 +215,23 @@ struct CLOptions {
...
+/// resulting Moore dialect IR .
+static LogicalResult populateMooreTransforms(mlir::PassManager &pm) {
+  pm.addPass(mlir::createMem2Reg());
+  // TODO: like dedup pass.
+
+  return success();
+}
+
+/// Convert Moore dialect IR into core dialect IR
+static LogicalResult populateLowMooreToCore(mlir::PassManager &pm) {
```
fabianschuiki, line 229:
```suggestion
static LogicalResult populateMooreToCoreLowering(mlir::PassManager &pm) {
```

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -269,6 +301,23 @@ static LogicalResult executeWithSources(MLIRContext *context,
...
+  if (opts.loweringMode == LoweringMode::OutputIRMoore ||
+      opts.loweringMode == LoweringMode::OutputIRHW) {
+
+    // Simplify the Moore dialect IR.
+    if (failed(populateMooreTransforms(pm)))
+      return failure();
+
+    if (opts.loweringMode == LoweringMode::OutputIRHW)
+      // Convert Moore IR into core IR.
+      if (failed(populateLowMooreToCore(pm)))
```
fabianschuiki, line 314:
```suggestion
      if (failed(populateMooreToCoreLowering(pm)))
```

fabianschuiki (approved):
Very nice! Great to see mem2reg in action :sunglasses:

PR #7083, opened 2024-05-23
[ImportVerilog] Support parameter constants.

angelzzzzz:
Create a new op `NamedConstantOp` for three elaboration-time constants: `parameter`, `localparam`, and `specparam`, which distinguishes 'parameter' from constants and variables.


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -294,6 +294,37 @@ def ConstantOp : MooreOp<"constant", [Pure]> {
...
+def LocalParameter : I32EnumAttrCase<"LocalParameter", 1, "localparam">;
+def SpecParameter : I32EnumAttrCase<"SpecParameter", 2, "specparam">;
+
+def NamedConstAttr : I32EnumAttr<"NamedConst", "elaboration-time constants", 
+       [Parameter, LocalParameter, SpecParameter]>{
+    let cppNamespace = "circt::moore";
+}
+
+def NamedConstantOp : MooreOp<"named_constant", [
+  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>,
```
hailongSun2000, line 307: If this op is approved. Maybe we should add `ConstantLike` and `Pure` to its trait list.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -294,6 +294,37 @@ def ConstantOp : MooreOp<"constant", [Pure]> {
...
+  OptionalTypesMatchWith<"value and named constant types match",
+    "result", "value", "$_self">,
+]>{
```
hailongSun2000, line 310: The `OptionalTypesMatchWith` is different from `TypesMatchWith`. It's just used to either `lhs` or `rhs` are optional. `SameOperandsAndResultType` is a better choice.

fabianschuiki: 👍 for `SameOperandsAndResultType` 😃

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: Cool! It's great to have an op to materialize the compile-time constants and preserve them. I don't think you can use `ConstantLike` here since the op takes a single operand, unlike `ConstantOp` which has no operands and just a single constant value attribute. `Pure` could work, but that would mean that if the `named_constant` result is unused (which I think it will be in a lot of cases, for example if the constant value was only used to parametrize types) then CSE/dead code elimination will delete the `named_constant` op again. That might not be what we want.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -191,6 +191,53 @@ struct MemberVisitor {
...
+    auto type = context.convertType(*paramNode.getDeclaredType());
+    if (!type)
+      return failure();
+
+    Value initial;
+    if (const auto *init = paramNode.getInitializer()) {
+      initial = context.convertExpression(*init);
+      if (!initial)
+        return failure();
+    }
```
fabianschuiki, line 205: If there is no initializer, I don't think you can create a `NamedConstantOp`, because there is no value to pass to that constant. I think you can just skip/ignore parameters that have no assigned value. So something like:
```suggestion
    const auto *init = paramNode.getInitializer();
    if (!init)
      return success();  // skip parameters without value
    Value initial = context.convertExpression(*init);
    if (!initial)
      return failure();
```

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -191,6 +191,53 @@ struct MemberVisitor {
...
+    auto type = context.convertType(*spNode.getDeclaredType());
+    if (!type)
+      return failure();
+
+    Value initial;
+    if (const auto *init = spNode.getInitializer()) {
+      initial = context.convertExpression(*init);
+      if (!initial)
+        return failure();
+    }
```
fabianschuiki, line 230: Same as above: you probably want to skip specparams without any value.
```suggestion
    const auto *init = spNode.getInitializer();
    if (!init)
      return success();  // skip parameters without value
    Value initial = context.convertExpression(*init);
    if (!initial)
      return failure();
```

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -78,6 +78,25 @@ module Basic;
...
+  localparam lp1 = 2;
+  // CHECK: %lp2 = moore.named_constant localparam %lp1 : !moore.l32
+  localparam lp2 = lp1;
+  // CHECK: [[TMP1:%.+]] = moore.constant 3 : !moore.i32
+  // CHECK: [[TMP2:%.+]] = moore.conversion [[TMP1]] : !moore.i32 -> !moore.l32
+  // CHECK: %sp1 = moore.named_constant specparam [[TMP2]] : !moore.l32
+  specparam sp1 = 3;
+  // CHECK: %sp2 = moore.named_constant specparam %sp1 : !moore.l32
+  specparam sp2 = sp1;
+
```
fabianschuiki, line 99: It would be great to also test the code path where there is no value assigned to the parameter. To ensure those cases also work (and potentially don't produce any IR op).

angelzzzzz: Thoughtful idea! But this may not work. If no value is assigned to the parameter, Slang will throw an error that "error: parameter declaration is missing an initializer".

include/circt/Dialect/Moore/MooreOps.td
angelzzzzz: Good idea! I will add `SameOperandsAndResultType` for it.

include/circt/Dialect/Moore/MooreOps.td
hailongSun2000: I get it! Looks like I misunderstod.

angelzzzzz:
@fabianschuiki Thank you for your generous help! I have resolved the issues and look forward to hearing any other suggestions you may have.

fabianschuiki (approved):
Looks great to me! Thanks a lot for pushing this forward :partying_face:

PR #7085, opened 2024-05-23
[Seq] Fix incorrect folder

uenoku:
If the register has a const reset value, we can replace but we cannot replace it with a non-constant reset value.


dtzSiFive:
(The source of the test failure has been resolved, if you rebase this should be good to go!)

dtzSiFive (approved):
LGTM.

If you'd like this could maybe allow anything that's ConstantLike? But that might be too general, what's here is good as-is just mentioning it in case that seems better to you.

PR #7086, opened 2024-05-23
Calyx ConstantOp Support

jiahanxie353:
This PR stacks on #6952 and is the first part of #6953 , in which I was asked to break down to smaller patches.

This PR adds support to the creation of Calyx `ConstantOp` in order to support floating point calculation in Calyx. In the meantime, we can support creating a Calyx `RegisterOp` by passing in a `Type` as an argument. I have also included a test case.


include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -18,6 +20,42 @@ class CalyxPrimitive<string mnemonic, list<Trait> traits = []> :
...
+    floating-point constant specified by an attribute.
+
+    Example:
+
+    ```
+    // Integer constant
+    %1 = calyx.constant 42 : i32
+
+    // Floating point constant
+    %1 = calyx.constant 42.00+e00 : f32
```
cgyurgyik, line 40: In Calyx, we usually treat constants as bit vectors. I'm a bit hesitant to introduce a typed constant here, e.g., how would you convert this when emitting to native Calyx?

jiahanxie353: Aside from "bitnum" and "fixed_point", we can also add a "floating_point" type, which will be interpreted as regular, decimal, floating point values.
But I agree that in the end of the day, everything is just 0 and 1 bit vectors.

cgyurgyik: Those are just interpretations of the bit vectors. We have a native Calyx emitter in CIRCT; how would you convert, e.g., `%1 = calyx.constant 1.1 : f32` using the emitter?

In any case, we should update the emitter and a [test](https://github.com/llvm/circt/blob/main/test/Dialect/Calyx/emit.mlir) for these Calyx constants.

jiahanxie353: You are right, just added the emitter!

jiahanxie353:
I know that LLVM has tools ` llvm/utils/update_mir_test_checks.py ` to update the tests by adding `CHECK`, `CHECK-NEXT`, etc. automatically. Is there a similar tool in CIRCT? I searched but couldn't find any.

test/Dialect/Calyx/emit.mlir
```diff
@@ -241,3 +241,40 @@ module attributes {calyx.entrypoint = "main"} {
...
+    %ret_arg1_reg.in, %ret_arg1_reg.write_en, %ret_arg1_reg.clk, %ret_arg1_reg.reset, %ret_arg1_reg.out, %ret_arg1_reg.done = calyx.register @ret_arg1_reg : f32, i1, i1, i1, f32, i1
+    %ret_arg0_reg.in, %ret_arg0_reg.write_en, %ret_arg0_reg.clk, %ret_arg0_reg.reset, %ret_arg0_reg.out, %ret_arg0_reg.done = calyx.register @ret_arg0_reg : i32, i1, i1, i1, i32, i1
+    calyx.wires {
+      calyx.assign %out1 = %ret_arg1_reg.out : f32
+      calyx.assign %out0 = %ret_arg0_reg.out : i32
+
+      // CHECK-LABEL: group ret_assign_0 {
+      // CHECK-NEXT:   ret_arg0_reg.in = 32'd42;
+      // CHECK-NEXT:   ret_arg0_reg.write_en = 1'd1;
+      // CHECK-NEXT:   ret_arg1_reg.in = 32'd4.200000;
```
cgyurgyik, line 261: Wait this well formed in Calyx?

jiahanxie353: You are right that it's not well-formed currently. I've been using json data to read external memory data, which can be floating points.
I will support direct floating point representation in Calyx in the future.

rachitnigam: I think floating-point constants are tricky and we should be careful about adding them to the language. A particular challenge is that we'd need to have an IEEE-754 (floating point specification) compatible emitter which can generate the right bit patterns for a given floating-point number which can be easy to get wrong. I've created a new issue in the Calyx repo to discuss this: https://github.com/calyxir/calyx/issues/2061

In the meantime, one approach I recommend is adding the ability to generate the right bit patterns for the floating-point number and generating that in the emitter to native Calyx. Does that make sense?

jiahanxie353: > In the meantime, one approach I recommend is adding the ability to generate the right bit patterns for the floating-point number and generating that in the emitter to native Calyx

Do you mean something like [this](https://github.com/llvm/circt/blob/437693a508e974a43982a1a39d887feff7107342/test/Dialect/Calyx/emit.mlir#L261)?

rachitnigam:
I think before merging this, we need to merge https://github.com/calyxir/calyx/issues/2061. @jiahanxie353 can you get the changes on that resolved so we can move forward with this PR?

jiahanxie353:
Changed the corresponding Calyx Constant Floating Point part reflecting [the new changes](https://github.com/calyxir/calyx/pull/2316). Marking it ready for review

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -1953,6 +1953,74 @@ ParseResult GroupDoneOp::parse(OpAsmParser &parser, OperationState &result) {
...
+
+    // Otherwise, build a complex name with the value and type.
+    SmallString<32> specialNameBuffer;
+    llvm::raw_svector_ostream specialName(specialNameBuffer);
+    specialName << 'c' << intCst.getValue();
+    if (intType)
+      specialName << '_' << type;
+    setNameFn(getResult(), specialName.str());
+  } else {
+    setNameFn(getResult(), "cst");
```
cgyurgyik, line 1977: Nit: the `else` case should be placed first and return early to avoid the additional (unused) allocations above.

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -899,6 +907,20 @@ void Emitter::emitInvoke(InvokeOp invoke) {
...
+  TypedAttr attr = constantOp.getValueAttr();
+  assert(isa<FloatAttr>(attr) && "must be a floating point constant");
+  auto fltAttr = cast<FloatAttr>(attr);
+  APFloat value = fltAttr.getValue();
+  auto type = cast<FloatType>(fltAttr.getType());
+  double doubleValue = value.convertToDouble();
+  auto floatBits = value.getSizeInBits(type.getFloatSemantics());
+  indent() << constantOp.getName().str() << space() << equals() << space()
+           << "std_float_const";
+  os << LParen() << "0" /* IEEE754 */ << comma() << floatBits << comma()
```
cgyurgyik, line 920: Perhaps spell out this magic number a bit more explicitly, something like:
```suggestion
  // Currently defaults to IEEE-754 representation [1]. 
  // [1]: https://github.com/calyxir/calyx/blob/main/primitives/float.futil
  static constexpr int32_t IEEE754 = 0;
  os << LParen() << std::to_string(IEEE754) << comma() << floatBits << comma()
```

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -1953,6 +1953,74 @@ ParseResult GroupDoneOp::parse(OpAsmParser &parser, OperationState &result) {
...
+  // The value's type must match the return type.
+  if (getValue().getType() != type) {
+    return emitOpError() << "value type " << getValue().getType()
+                         << " must match return type: " << type;
+  }
+  // Integer values must be signless.
+  if (llvm::isa<IntegerType>(type) &&
+      !llvm::cast<IntegerType>(type).isSignless())
+    return emitOpError("integer return type must be signless");
+  // Any float or elements attribute are acceptable.
```
cgyurgyik, line 1992: `elements`? Did you mean "or integer attribute"?

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -1953,6 +1953,74 @@ ParseResult GroupDoneOp::parse(OpAsmParser &parser, OperationState &result) {
...
+  } else {
+    setNameFn(getResult(), "cst");
+  }
+}
+
+LogicalResult ConstantOp::verify() {
+  auto type = getType();
+  // The value's type must match the return type.
+  if (getValue().getType() != type) {
+    return emitOpError() << "value type " << getValue().getType()
```
cgyurgyik, line 1985:
```suggestion
  if (auto valType = getValue().getType(); valType != type) {
    return emitOpError() << "value type " << valType
```

cgyurgyik (approved):
A small few comments, but in general LGTM.

jiahanxie353:
Thank you @cgyurgyik ! I have made the changes based on the review.
I guess I can just click the button once all the tests finish running and pass?

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -1953,6 +1953,74 @@ ParseResult GroupDoneOp::parse(OpAsmParser &parser, OperationState &result) {
...
+  if (isa<FloatAttr>(getValue())) {
+    setNameFn(getResult(), "cst");
+    return;
+  }
+  auto intCst = llvm::dyn_cast<IntegerAttr>(getValue());
+  auto intType = llvm::dyn_cast<IntegerType>(getType());
+
+  // Sugar i1 constants with 'true' and 'false'.
+  if (intType && intType.getWidth() == 1)
+    return setNameFn(getResult(), (intCst.getInt() ? "true" : "false"));
```
cgyurgyik, line 1970: Super nit suggestion: explicitly cast to boolean. 
```suggestion
    return setNameFn(getResult(), intCst.getInt() > 0 ? "true" : "false");
```

cgyurgyik:
> Thank you @cgyurgyik ! I have made the changes based on the review. I guess I can just click the button once all the tests finish running and pass?

Yes once tests pass feel free to merge. Great job!

andrewb1999 (approved):
LGTM! Eventually we should maybe consider just using the calyx ConstantOp for everything instead of using both the calyx and hw constant ops.

PR #7087, opened 2024-05-25
[Arc] Modify VectorizeOp to support AnyType

elhewaty:


fabianschuiki (approved):
LGTM! Since we are planning to use `arc.vectorize` fairly early in the pipeline to mark vectorizable regions of the design, it makes sense to allow for non-integer types such as `!seq.clock`, `!hw.array`, `!hw.struct`, as in this PR. :+1:

PR #7088, opened 2024-05-26
[Moore] Move array types into ODS

fabianschuiki:
Move the definitions of all array types from C++ land into `MooreTypes.td`. This removes a significant amount of redundant code and simplifies the dialect's type system.

Replace packed and unpacked ranges (`T [4:3]` or `T [2:4]`) with a new type that discards the directionality and offset of the range. This information is no longer needed at the IR level. Any meaning that the offsets have can be encoded in the corresponding ops. Both ranges are now represented as `array<2 x T>` and `array<3 x T>`, respectively.

Combine unpacked ranges (`T foo [6:2]`) and unpacked arrays (`T foo [42]`) into a single `uarray<N x T>` type.

### DIFF
--- a/test/Conversion/ImportVerilog/basic.sv
+++ b/test/Conversion/ImportVerilog/basic.sv
@@ -292,7 +292,7 @@ module Expressions;
   // CHECK: %b = moore.variable : !moore.i32
   // CHECK: %c = moore.variable : !moore.i32
   // CHECK: %u = moore.variable : !moore.i32
-  // CHECK: %v = moore.variable : !moore.packed<range<i4, 1:0>>
+  // CHECK: %v = moore.variable : !moore.array<2 x !moore.i4>
   // CHECK: %d = moore.variable : !moore.l32
   // CHECK: %e = moore.variable : !moore.l32
   // CHECK: %f = moore.variable : !moore.l32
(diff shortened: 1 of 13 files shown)


test/Conversion/ImportVerilog/basic.sv
```diff
@@ -292,7 +292,7 @@ module Expressions;
   // CHECK: %b = moore.variable : !moore.i32
   // CHECK: %c = moore.variable : !moore.i32
   // CHECK: %u = moore.variable : !moore.i32
-  // CHECK: %v = moore.variable : !moore.packed<range<i4, 1:0>>
+  // CHECK: %v = moore.variable : !moore.array<2 x !moore.i4>
```
maerhart, line 295: I thought the long-term goal was to use the built-in integer type. What is the motivation for replacing `i4` with `!moore.i4`?

test/Conversion/ImportVerilog/types.sv
```diff
@@ -76,34 +76,39 @@ module IntAtoms;
...
+// CHECK-LABEL: moore.module @Dimensions
+module Dimensions;
+  // CHECK-NEXT: %p0 = moore.variable : !moore.l3
+  logic [2:0] p0;
+  // CHECK-NEXT: %p1 = moore.variable : !moore.l3
+  logic [0:2] p1;
+  // CHECK-NEXT: %p2 = moore.variable : !moore.array<6 x !moore.l3>
+  logic [5:0][2:0] p2;
+  // CHECK-NEXT: %p3 = moore.variable : !moore.array<6 x !moore.l3>
+  logic [0:5][2:0] p3;
```
maerhart, line 88: There's no `!moore.open_array` test because the lowering from slang is not implemented yet, right?

maerhart (approved):
LGTM!

test/Conversion/ImportVerilog/basic.sv
fabianschuiki: This is due to the printing now being done by the generated printer. I have a follow-up PR that enables stripped printing on `UnpackedType`, which is going to drop the `!moore.` prefix in almost all cases.

test/Conversion/ImportVerilog/types.sv
fabianschuiki: Yes :cry:. I think the only place where this is legal is in function declarations, and possibly only in extern "DPI" functions. We don't parse those yet unfortunately.

PR #7089, opened 2024-05-26
Calyx Binary Floating Point AddF Operator

jiahanxie353:
This patch stacks on #7086 .

Add support for floating point addition in Calyx

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -341,12 +341,12 @@ class ArithBinaryFloatingPointLibraryOp<string mnemonic> : ArithBinaryLibraryOp<
 def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {
   let results = (outs I1:$clk, I1:$reset, I1:$go, I1:$control, I1:$subOp,
     AnyFloat:$left, AnyFloat:$right, AnySignlessInteger:$roundingMode, AnyFloat:$out,
-    AnySignlessInteger:$execptionalFlags, I1:$done);
+    AnySignlessInteger:$exceptionalFlags, I1:$done);
 
   let extraClassDefinition = [{
     SmallVector<StringRef> $cppClass::portNames() {
       return {clkPort, resetPort, goPort, "control", "subOp",
-              "left", "right", "roundingMode", "out", "execptionalFlags", donePort
+              "left", "right", "roundingMode", "out", "exceptionalFlags", donePort
       };
     }
 
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -335,6 +335,49 @@ def AndLibOp  : CombinationalArithBinaryLibraryOp<"and"> {}
 def OrLibOp   : CombinationalArithBinaryLibraryOp<"or"> {}
 def XorLibOp  : CombinationalArithBinaryLibraryOp<"xor"> {}
 
+class ArithBinaryFloatingPointLibraryOp<string mnemonic> : ArithBinaryLibraryOp<mnemonic, [
+  SameTypeConstraint<"left", "out">]> {}
+
+def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {
+  let results = (outs I1:$clk, I1:$reset, I1:$go, I1:$control, I1:$subOp,
+    AnyFloat:$left, AnyFloat:$right, AnySignlessInteger:$roundingMode, AnyFloat:$out,
+    AnySignlessInteger:$execptionalFlags, I1:$done);
+
+  let extraClassDefinition = [{
+    SmallVector<StringRef> $cppClass::portNames() {
+      return {clkPort, resetPort, goPort, "control", "subOp",
+              "left", "right", "roundingMode", "out", "execptionalFlags", donePort
+      };
+    }
+
+    SmallVector<Direction> $cppClass::portDirections() {
+      return {Input, Input, Input, Input, Input, Input, Input, Input, Output, Output, Output};
+      }
+
+    void $cppClass::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
+      getCellAsmResultNames(setNameFn, *this, this->portNames());
+      }
+
+    bool $cppClass::isCombinational() { return false; }
+
+    SmallVector<DictionaryAttr> $cppClass::portAttributes() {
+      IntegerAttr isSet = IntegerAttr::get(IntegerType::get(getContext(), 1), 1);
+      NamedAttrList go, clk, reset, done;
+      go.append(goPort, isSet);
+      clk.append(clkPort, isSet);
+      reset.append(resetPort, isSet);
+      done.append(donePort, isSet);
+      return {clk.getDictionary(getContext()), reset.getDictionary(getContext()),
+              go.getDictionary(getContext()), DictionaryAttr::get(getContext()),
+              DictionaryAttr::get(getContext()), DictionaryAttr::get(getContext()),
+              DictionaryAttr::get(getContext()), DictionaryAttr::get(getContext()),
+              DictionaryAttr::get(getContext()), done.getDictionary(getContext()),
+              DictionaryAttr::get(getContext())
+      };
+    }
+  }];
+}
+
 def MuxLibOp : CalyxLibraryOp<"mux", [
   Combinational, SameTypeConstraint<"tru", "fal">, SameTypeConstraint<"tru", "out">
   ]> {
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -29,6 +29,7 @@
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Casting.h"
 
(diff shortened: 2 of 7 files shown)


rachitnigam:
This is interesting. I think we keep having to add ops because we don't have a good mechanism to talk about extern ops in the dialect. @cgyurgyik do you have thoughts on how best to encode external ops in the dialect?

rachitnigam:
@cgyurgyik exactly! We should build support for something like this in the Calyx dialect. A tricky thing is that we want to specify paths to Verilog files which implement the specific extern we're talking about. The upside would be that we don't have to keep adding new primitives to CIRCT ops definition (especially because we're not defining any optimizations over them).

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -142,6 +142,10 @@ struct ImportTracker {
           static constexpr std::string_view sMemories = "memories/seq";
           return {sMemories};
         })
+        .Case<AddFNOp>([&](auto op) -> FailureOr<StringRef> {
```
cgyurgyik, line 145: Could you add emitter tests for these please?

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -936,6 +944,37 @@ void Emitter::emitLibraryPrimTypedByFirstOutputPort(
...
+    sigWidth = 24;
+  } else if (bitWidth == 64) {
+    expWidth = 11;
+    sigWidth = 53;
+  } else if (bitWidth == 128) {
+    expWidth = 15;
+    sigWidth = 113;
+  } else {
+    op->emitError("Unsupported floating point width");
+  }
```
cgyurgyik, line 970:
```suggestion
  switch (bitWidth) {
    case 16:
      expWidth = 5;
      sigWidth = 24;
    // ...
    default: 
      op->emitError("Unsupported floating point width");
  }
```

cgyurgyik:
> @cgyurgyik exactly! We should build support for something like this in the Calyx dialect. A tricky thing is that we want to specify paths to Verilog files which implement the specific extern we're talking about. The upside would be that we don't have to keep adding new primitives to CIRCT ops definition (especially because we're not defining any optimizations over them).

Got it. Would this back us into a corner from ever defining optimizations over them? I could imagine a world where Calyx would allow the following,

```
mul = std_mul_pipe(32);

...

@promotable(3) group do_mul {
  mul.lhs = reg.in;
  mul.rhs = 32'd3;
  ...
}
```

Could be rewritten to:

```
mul = std_mul_const(32);

...

static<1> group do_mul {
  mul.lhs = reg.in;
  mul.rhs = 32'd3;
}
```

These "peephole" optimizations would be easier to implement with different `Operation`s and their typed input/outputs. (It isn't really peephole, since it would require some whole program analysis to ensure we aren't breaking some other constraints.)

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
cgyurgyik: Up to you, but I think a switch with `break`s is easier to parse here.

rachitnigam:
@cgyurgyik finally catching up to your comment. This is a good question about which ops we want to perform peephole optimizations on. In general, I think we should define them only on "simple" operations like addition. Something like a floating-point operation should probably be implemented as an external because the most optimized version is going to be some black-box RTL anyways.

(FWIW, this argument should also extend to really large adders! For example, it is not obvious to me that a 128-bit adder is best implemented automatically. For now, however, we'll leave this dragon sleeping.)

@jiahanxie353 How critical is this PR in the stack that you're trying to get approved? For example, if it is not yet super critical, I would lean towards redesigning this to use `extern`s instead so we can integrate with the HardFloat library.

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
jiahanxie353: done!

jiahanxie353:
> A tricky thing is that we want to specify paths to Verilog files which implement the specific extern we're talking about. The upside would be that we don't have to keep adding new primitives to CIRCT ops definition 

> I would lean towards redesigning this to use externs instead so we can integrate with the HardFloat library.

I believe I'm using externs, because it emits `import float/addFN` [here](https://github.com/llvm/circt/blob/deb4fab619ad473bf24e1b40747fc03172240f9f/test/Dialect/Calyx/emit.mlir#L287), which in turn uses `extern` [here](https://github.com/calyxir/calyx/blob/b3ee79a9a5ca1911f58684253c86399a7113900e/primitives/float/addFN.futil#L1)

Am I missing anything? Are you saying we should completely get rid of `AddFNOp` in TableGen [here](https://github.com/llvm/circt/blob/deb4fab619ad473bf24e1b40747fc03172240f9f/include/circt/Dialect/Calyx/CalyxPrimitives.td#L341), I don't see an obvious way to introduce floating point add/mul without adding new ops.

rachitnigam:
> Am I missing anything? Are you saying we should completely get rid of AddFNOp in TableGen [here](https://github.com/llvm/circt/blob/deb4fab619ad473bf24e1b40747fc03172240f9f/include/circt/Dialect/Calyx/CalyxPrimitives.td#L341)? I don't see an obvious way to introduce floating point add/mul without adding new ops.

Yeah, @andrewb1999 has been working on a idea for defining operator libraries that might eventually allow us to decouple the specification of new primitives from new ops in the compiler. For now, this seems fine. 

One nit about the new operator definitions (that you can address in a follow-on PR if it's a bigger change). Instead of `calyx.addFN`, we should be saying something like `calyx.ieee754.fadd` to make it explicit that this is not only a floating-point adder, but specifically an IEEE-754 compliant adder.

jiahanxie353:
> @andrewb1999 has been working on a idea for defining operator libraries that might eventually allow us to decouple the specification of new primitives from new ops in the compiler

sounds very neat, looking forward to it

> One nit about the new operator definitions (that you can address in a follow-on PR if it's a bigger change). Instead of `calyx.addFN`, we should be saying something like `calyx.ieee754.fadd` to make it explicit that this is not only a floating-point adder, but specifically an IEEE-754 compliant adder.

Agree. I'll address when there's a need in the future.

For now, I'll click the button if it looks good

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -335,6 +335,49 @@ def AndLibOp  : CombinationalArithBinaryLibraryOp<"and"> {}
...
+
+def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {
+  let results = (outs I1:$clk, I1:$reset, I1:$go, I1:$control, I1:$subOp,
+    AnyFloat:$left, AnyFloat:$right, AnySignlessInteger:$roundingMode, AnyFloat:$out,
+    AnySignlessInteger:$execptionalFlags, I1:$done);
+
+  let extraClassDefinition = [{
+    SmallVector<StringRef> $cppClass::portNames() {
+      return {clkPort, resetPort, goPort, "control", "subOp",
+              "left", "right", "roundingMode", "out", "execptionalFlags", donePort
```
cgyurgyik, line 349:
```suggestion
              "left", "right", "roundingMode", "out", "exceptionalFlags", donePort
```

cgyurgyik: here and elsewhere.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -434,6 +439,17 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
     // The group is done when the register write is complete.
     rewriter.create<calyx::GroupDoneOp>(loc, reg.getDone());
 
+    if (isa<calyx::AddFNOp>(opPipe)) {
+      auto opFN = cast<calyx::AddFNOp>(opPipe);
+      hw::ConstantOp subOp;
+      if (isa<arith::AddFOp>(op)) {
+        subOp = createConstant(loc, rewriter, getComponent(), 1, 0);
```
cgyurgyik, line 446: Suggestion: can we add `/*parameter_name=*/` for the magic constants here and below?

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -964,6 +972,44 @@ void Emitter::emitLibraryPrimTypedByFirstOutputPort(
            << LParen() << bitWidth << RParen() << semicolonEndL();
 }
 
+void Emitter::emitLibraryFloatingPoint(Operation *op) {
+  auto cell = cast<CellInterface>(op);
+  unsigned bitWidth =
+      cell.getOutputPorts()[0].getType().getIntOrFloatBitWidth();
+  assert(bitWidth > 0 && "bitWidth must be greater than zero.");
```
cgyurgyik, line 979: This is dead code, unsigned will always be > 0.

test/Dialect/Calyx/emit.mlir
```diff
@@ -280,3 +280,54 @@ module attributes {calyx.entrypoint = "main"} {
   } {toplevel}
 }
 
+
+// -----
+
+module attributes {calyx.entrypoint = "main"} {
+  // CHECKL: import "primitives/float/addFN.futil";
```
cgyurgyik, line 287: Is this a typo, or does `CHECKL` exist?

cgyurgyik (approved):
Some small nits, otherwise LGTM

test/Dialect/Calyx/emit.mlir
jiahanxie353: my bad, I wonder how it even passes the test check..

PR #7090, opened 2024-05-26
[Moore] Support stripped type parsing and printing

fabianschuiki:
Operations defined through TableGen will use a version of type parsing and printing that strips away the dialect prefix in case this does not lead to ambiguities. If an op specifies a type like `UnpackedType`, the generated parser and printer will call `UnpackedType::parse` and `UnpackedType::print` in order to parse and print that type. This method usually prints the type without any dialect prefixes, and often also without the type name itself (as it is obvious given the type).

Add an implementation for `parse` and `print` to `UnpackedType` which performs this stripped printing. This makes the assembly more compact and in-line with what MLIR would generate by default if a TableGen-specified op would use a TableGen-specified type.

This makes the integer flavor of `IntType` look exactly like the builtin integer type, which will facilitate a future change from Moore's integer type to the builtin integer. Parsing is unambiguous, since the stripped form is only used when the type is clear from the op context.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -236,7 +236,7 @@ def ReadLValueOp : MooreOp<"read_lvalue", [SameOperandsAndResultType]> {
 
 class AssignOpBase<string mnemonic, list<Trait> traits = []> :
     MooreOp<mnemonic, traits # [SameTypeOperands]> {
-  let arguments = (ins AnyType:$dst, AnyType:$src);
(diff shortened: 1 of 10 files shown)


hailongSun2000 (approved):

hailongSun2000:
LGTM! :tada:

maerhart (approved):
LGTM!

uenoku (approved):
Awesome!

fabianschuiki:
Thanks :partying_face:

PR #7091, opened 2024-05-26
[Moore] Move struct types into ODS

fabianschuiki:
Move the definitions of packed and unpacked struct types from C++ land into `MooreTypes.td`. This removes a significant amount of complexity.

With all types moved into the TableGen file, a lot of redundant code can finally be removed and the parsing/printing can be streamlined significantly.

This change also drops the `StructKind` enum which was used to discern structs, unions, and tagged unions, although the ImportVerilog conversion never generated anything besides structs. Once we add support for unions in the future, the intention is to define new types in ODS for the unions and reuse the `StructMember`.

### DIFF
--- a/include/circt/Dialect/Moore/MooreTypes.h
+++ b/include/circt/Dialect/Moore/MooreTypes.h
@@ -23,6 +23,23 @@
 namespace circt {
 namespace moore {
 
+class ArrayType;
+class AssocArrayType;
+class ChandleType;
+class EventType;
+class IntType;
+class OpenArrayType;
+class OpenUnpackedArrayType;
+class PackedType;
+class QueueType;
+class RealType;
+class StringType;
+class StructType;
+class UnpackedType;
+class UnpackedArrayType;
+class UnpackedStructType;
+class VoidType;
+
 /// The number of values each bit of a type can assume.
 enum class Domain {
   /// Two-valued types such as `bit` or `int`.
@@ -31,43 +48,9 @@ enum class Domain {
   FourValued,
 };
 
-/// Whether a type is signed or unsigned.
-enum class Sign {
-  /// An `unsigned` type.
-  Unsigned,
-  /// A `signed` type.
-  Signed,
-};
-
-/// Map a `Sign` to the corresponding keyword.
-StringRef getKeywordFromSign(const Sign &sign);
-/// Map the keywords `unsigned` and `signed` to the corresponding `Sign`.
-std::optional<Sign> getSignFromKeyword(StringRef keyword);
-
-template <typename Os>
-Os &operator<<(Os &os, const Sign &sign) {
-  os << getKeywordFromSign(sign);
-  return os;
-}
-
-namespace detail {
-struct StructTypeStorage;
-} // namespace detail
-
 //===----------------------------------------------------------------------===//
 // Unpacked Type
 //===----------------------------------------------------------------------===//
-
-class PackedType;
-class StringType;
-class ChandleType;
-class EventType;
-class RealType;
-class UnpackedArrayType;
-class OpenUnpackedArrayType;
-class AssocArrayType;
-class QueueType;
-class UnpackedStructType;
 
 /// An unpacked SystemVerilog type.
 ///
@@ -123,12 +106,6 @@ protected:
 // Packed Type
 //===----------------------------------------------------------------------===//
 
-class VoidType;
-class IntType;
-class ArrayType;
-class OpenArrayType;
-class PackedStructType;
-
 /// A packed SystemVerilog type.
 ///
 /// Packed types are the core types of SystemVerilog. They combine a core packed
@@ -153,8 +130,8 @@ class PackedStructType;
 class PackedType : public UnpackedType {
 public:
   static bool classof(Type type) {
-    return llvm::isa<VoidType, IntType, ArrayType, OpenArrayType,
-                     PackedStructType>(type);
+    return llvm::isa<VoidType, IntType, ArrayType, OpenArrayType, StructType>(
+        type);
   }
 
   /// Get the value domain of this type.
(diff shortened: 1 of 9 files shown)


include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -272,6 +272,40 @@ def QueueType : MooreTypeDef<"Queue", [], "moore::UnpackedType"> {
...
+  "UnpackedStruct", [], "moore::UnpackedType"
+> {
+  let mnemonic = "ustruct";
+  let summary = "an unpacked struct type";
+  let description = [{
+    An unpacked struct.
+  }];
+}
+
+
```
mingzheTerapines, line 308: So packed/unpacked union type will be declared separately after this?

include/circt/Dialect/Moore/MooreTypes.h
```diff
@@ -170,31 +147,9 @@ class PackedType : public UnpackedType {
...
 //===----------------------------------------------------------------------===//
 
-/// Whether a struct is a `struct`, `union`, or `union tagged`.
-enum class StructKind {
-  /// A `struct`.
-  Struct,
-  /// A `union`.
-  Union,
-  /// A `union tagged`.
-  TaggedUnion,
```
mingzheTerapines, line 183: After some tests, I found union-tag check will be done in Slang such as 
union.v:46:15: error: value of type 'unionTemp' cannot be assigned to type 'bit[3:0]'
msg1.halfMsg  = myUnion;
error message will be printed.
So pack/unpacked struct/union will be all right.

mingzheTerapines:
This looks perfect!!!
ImportVerilog pass seems much more lighter and clearer because we contributers can only do conversion from   Ast to moore out of concerning lowering jobs. 
Also other transforming pass more than Mem2Reg pass is being needed. Could you give some advice?

lib/Dialect/Moore/MooreTypes.cpp
```diff
@@ -20,49 +20,13 @@ using namespace circt;
...
-  bool allowUnpacked = true;
-};
+static LogicalResult parseMembers(AsmParser &parser,
+                                  SmallVector<StructMember> &members);
+static void printMembers(AsmPrinter &printer, ArrayRef<StructMember> members);
 
-static ParseResult parseMooreType(AsmParser &parser, Subset subset, Type &type);
-static void printMooreType(Type type, AsmPrinter &printer, Subset subset);
+static ParseResult parseMooreType(AsmParser &parser, Type &type);
+static void printMooreType(Type type, AsmPrinter &printer);
```
hailongSun2000, line 29: Could I consider deleting `Subset subset` here is different from #7090 which specifies `Subset::None` is due to manually skipping the estimation of packed and unpacked, and then doesn't print `packed<>` and `unpacked<>`, right?    However, in this PR, in order to move types into ODS and streamline the parsing/printing, the distinction of `Subset` is unnecessary.

hailongSun2000 (approved):

hailongSun2000:
> This looks perfect!!! ImportVerilog pass seems much more lighter and clearer because we contributers can only do conversion from Ast to moore out of concerning lowering jobs. Also other transforming pass more than Mem2Reg pass is being needed. Could you give some advice?

Maybe you can wait until #7082 lands, then implement the struct-related Ops. The implementation of `DestructurableAllocationOpInterface` can refer to `MemRef`.

mingzheTerapines:
> > This looks perfect!!! ImportVerilog pass seems much more lighter and clearer because we contributers can only do conversion from Ast to moore out of concerning lowering jobs. Also other transforming pass more than Mem2Reg pass is being needed. Could you give some advice?
> 
> Maybe you can wait until #7082 lands, then implement the struct-related Ops. The implementation of `DestructurableAllocationOpInterface` can refer to `MemRef`.

Good advice! I get it.

include/circt/Dialect/Moore/MooreTypes.h
mingzheTerapines: Also in mooreops.td definining ops of procedure also has these kinds
```
def ProcedureKindAttr: I32EnumAttr<"ProcedureKind", "Procedure kind",
            [Initial, Final, Always, AlwaysComb, AlwaysLatch, AlwaysFF]>{
}
```
Will it be better to create more ops like procedure_initial, procedure_final and so on ?
Some of them will be lowered to hw but procedure initial will be lowered to llhd.

mingzheTerapines: Personally, I suggest let proceudre as a abstract class and let Initial, Final, Always, AlwaysComb, AlwaysLatch, AlwaysFF procedure ops implements separately with def declaration.

include/circt/Dialect/Moore/MooreTypes.td
fabianschuiki: Yes exactly!

include/circt/Dialect/Moore/MooreTypes.h
fabianschuiki: This is a bit of a balancing act. For types it's often easier to split them apart into separate types, as long as there is no explosion of combinations that you would have to support. This helps with type checks, where you often would prefer to write `isa<UnionType>()` instead of `auto ty = dyn_cast<StructLikeType(); ty && ty.getKind() == StructKind::Union`.

Procedures have very similar structure and limitations. That suggests a single op with an enum kind. You are totally right though, we may want to have more specialized process ops and lower to them. For example, for simple `always_comb` which are the only driver to the assigned signals, we could create something like this:
```
moore.procedure always_comb {
  %0 = comb.add %a, %b
  moore.blocking_assign %x, %0
  %1 = comb.sub %0, %c
  moore.blocking_assign %y, %1
}
// ----- 8< ----- lowered into ----- 8< ----- 
%2, %3 = moore.execute_region {
  %0 = comb.add %a, %b
  %1 = comb.sub %0, %c
  moore.yield %0, %1
}
moore.assign %x, %2
moore.assign %y, %3
```

The idea is to start close to the AST, and then massage the IR to get closer to our goal (inlined always blocks, registers detected and made explicit, etc.) to make MooreToCore's life very easy :smiley:

lib/Dialect/Moore/MooreTypes.cpp
fabianschuiki: Yes exactly! Now that the types indicate in their name whether they are packed or not (`array` vs. `uarray`), that whole complicated `Subset`-based parsing/printing is no longer needed :smiley:

PR #7095, opened 2024-05-28
[Moore] Introduce RefType and tweak the related ops.

hailongSun2000:
Add a wrapper type(`moore::RefType`) to work for `VariableOp/NetOp`, `ReadOp`, and `AssignOpBase`, and it can wrap any SystemVerilog types. We added `RefType` in order to introduce `Mem2Reg` into `moore` to eliminate the local temporary variables. Meanwhile, we added a new op named `moore.extract_ref` which is the copy of `moore.extract` that explicitly works on the ref type. And `moore.concat_ref` is the copy of `moore.concat`. Later, we will tweak `union/struct` ops like `moore.struct_create` in order to support the `ref` type.

`moore.variable/moore.net` will return a result with `ref<T>`, it's similar to `memref.alloca`.

Change `moore::ReadLValueOp` to `moore::ReadOp`, it's similar to `memref.load`.  Initially, we need `moore.read_lvalue` only to service for assigmen operations(like `+=`, `*=`, etc).  However, `moore. read` is not only for assignment operations but also to take a value with `RefType`(ref<T>) and return the current value with the nested type(`T`). For example, `a = b + c`, `moore.read` will take `b` and `c`(both are `ref<T>`), and then return a value with the nested type(`T`).

We think the `MooreLValueType` and `VariableDeclOp`(https://github.com/llvm/circt/commit/52a71a681e8cdd1d38936b8f5e1bb5716babf9e5) that had been removed eventually found their proper site.

We also divide `convertExpression()` into `convertLvalueExpression()` for the LHS of assignments and `convertRvalueExpression()` for the RHS of assignments. The former will return a value with `RefType` like `ref<i32>`, but the latter returns a value without `RefType` like `i32`.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -182,14 +182,14 @@ def ProcedureOp : MooreOp<"procedure", [
 def VariableOp : MooreOp<"variable", [
   DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>,
   OptionalTypesMatchWith<"initial value and variable types match",
-    "result", "initial", "$_self">,
+    "result", "initial", "cast<RefType>($_self).getNestedType()">
 ]> {
   let summary = "A variable declaration";
   let description = [{
     See IEEE 1800-2017 § 6.8 "Variable declarations".
   }];
   let arguments = (ins StrAttr:$name, Optional<UnpackedType>:$initial);
-  let results = (outs Res<UnpackedType, "", [MemAlloc]>:$result);
+  let results = (outs Res<RefType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
     `` custom<ImplicitSSAName>($name) ($initial^)? attr-dict
     `:` type($result)
@@ -219,7 +219,7 @@ def NetKindAttr : I32EnumAttr<"NetKind", "Net type kind", [
 def NetOp : MooreOp<"net", [
   DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>,
   OptionalTypesMatchWith<"assigned value and variable types match",
-    "result", "assignment", "$_self">,
+    "result", "assignment", "cast<RefType>($_self).getNestedType()">,
 ]> {
   let summary = "A net declaration";
   let description = [{
@@ -243,21 +243,24 @@ def NetOp : MooreOp<"net", [
     NetKindAttr:$kind,
     Optional<UnpackedType>:$assignment
   );
-  let results = (outs UnpackedType:$result);
+  let results = (outs Res<RefType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
     ``custom<ImplicitSSAName>($name) $kind ($assignment^)? attr-dict
     `:` type($result)
   }];
 }
 
-def ReadLValueOp : MooreOp<"read_lvalue", [SameOperandsAndResultType]> {
+def ReadOp : MooreOp<"read", [
+  TypesMatchWith<"input and result types match",
+    "result", "input", "RefType::get(cast<UnpackedType>($_self))">
+]> {
   let summary = "Read the current value of a declaration";
   let description = [{
     Samples the current value of a declaration. This is a helper to capture the
-    exact point at which declarations that can be targeted by assignments are
-    read.
+    exact point at which declarations that can be targeted by all possible 
+    expressions are read. It's similar to llvm.load.
   }];
-  let arguments = (ins Arg<UnpackedType, "", [MemRead]>:$input);
+  let arguments = (ins Arg<RefType, "", [MemRead]>:$input);
   let results = (outs UnpackedType:$result);
   let assemblyFormat = [{
     $input attr-dict `:` type($result)
@@ -269,8 +272,9 @@ def ReadLValueOp : MooreOp<"read_lvalue", [SameOperandsAndResultType]> {
 //===----------------------------------------------------------------------===//
 
 class AssignOpBase<string mnemonic, list<Trait> traits = []> :
-    MooreOp<mnemonic, traits # [SameTypeOperands]> {
-  let arguments = (ins UnpackedType:$dst, UnpackedType:$src);
+    MooreOp<mnemonic, traits # [TypesMatchWith<"src and dst types match",
+    "src", "dst", "RefType::get(cast<UnpackedType>($_self))">]> {
+  let arguments = (ins RefType:$dst, UnpackedType:$src);
   let assemblyFormat = [{
     $dst `,` $src attr-dict `:` type($src)
   }];
@@ -297,7 +301,7 @@ def BlockingAssignOp : AssignOpBase<"blocking_assign"> {
     See IEEE 1800-2017 § 10.4.1 "Blocking procedural assignments".
   }];
   let arguments = (ins
-    Arg<UnpackedType, "", [MemWrite]>:$dst,
+    Arg<RefType, "", [MemWrite]>:$dst,
     UnpackedType:$src
   );
 }
@@ -795,6 +799,17 @@ def ConcatOp : MooreOp<"concat", [
   }];
 }
 
+def ConcatRefOp : MooreOp<"concat_ref", [
+    Pure, DeclareOpInterfaceMethods<InferTypeOpInterface>
+]> {
+  let summary = "The copy of concat that explicitly works on the ref type.";
+  let arguments = (ins Variadic<RefType>:$values);
+  let results = (outs RefType:$result);
+  let assemblyFormat = [{
+    $values attr-dict `:` `(` type($values) `)` `->` type($result)
+  }];
+}
+
 def ReplicateOp : MooreOp<"replicate", [
   Pure
 ]> {
@@ -850,6 +865,16 @@ def ExtractOp : MooreOp<"extract"> {
   }];
   let arguments = (ins UnpackedType:$input, UnpackedType:$lowBit);
   let results = (outs UnpackedType:$result);
+  let assemblyFormat = [{
+    $input `from` $lowBit attr-dict `:`
+    type($input) `,` type($lowBit) `->` type($result)
+  }];
+}
+
+def ExtractRefOp : MooreOp<"extract_ref"> {
+  let summary = "The copy of extract that explicitly works on the ref type.";
+  let arguments = (ins RefType:$input, UnpackedType:$lowBit);
+  let results = (outs RefType:$result);
   let assemblyFormat = [{
     $input `from` $lowBit attr-dict `:`
     type($input) `,` type($lowBit) `->` type($result)
--- a/include/circt/Dialect/Moore/MooreTypes.td
+++ b/include/circt/Dialect/Moore/MooreTypes.td
@@ -326,6 +326,29 @@ def UnpackedUnionType : StructLikeType<
 }
 
 //===----------------------------------------------------------------------===//
+// Reference type wrapper
+//===----------------------------------------------------------------------===//
+
+def RefType : MooreTypeDef<"Ref", [], "moore::UnpackedType">{
+  let mnemonic = "ref";
+  let description = [{
+    A wrapper is used to wrap any SystemVerilog type. It's aimed to work for
+    'moore.variable', 'moore.blocking_assign', and 'moore.read', which are
+    related to memory, like alloca/write/read.
+  }];
+  let parameters = (ins "UnpackedType":$nestedType);
+  let assemblyFormat = [{
+    `<` $nestedType `>`
+  }];
+
+  let builders = [
+    AttrBuilderWithInferredContext<(ins "UnpackedType":$nestedType), [{
+      return $_get(nestedType.getContext(), nestedType);
+    }]>
+  ];
+}
+
+//===----------------------------------------------------------------------===//
 // Constraints
 //===----------------------------------------------------------------------===//
 
--- a/lib/Conversion/ImportVerilog/ImportVerilogInternals.h
+++ b/lib/Conversion/ImportVerilog/ImportVerilogInternals.h
@@ -76,7 +76,8 @@ struct Context {
   LogicalResult convertStatement(const slang::ast::Statement &stmt);
 
   // Convert an expression AST node to MLIR ops.
-  Value convertExpression(const slang::ast::Expression &expr);
+  Value convertRvalueExpression(const slang::ast::Expression &expr);
+  Value convertLvalueExpression(const slang::ast::Expression &expr);
 
   mlir::ModuleOp intoModuleOp;
   const slang::SourceManager &sourceManager;
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -133,8 +133,10 @@ struct MemberVisitor {
       // either attach it to the instance as an operand (for input, inout, and
       // ref ports), or assign an instance output to it (for output ports).
       if (auto *port = con->port.as_if<PortSymbol>()) {
-        // TODO: Convert as rvalue for inputs, lvalue for all others.
-        auto value = context.convertExpression(*expr);
+        // Convert as rvalue for inputs, lvalue for all others.
+        auto value = (port->direction == slang::ast::ArgumentDirection::In)
+                         ? context.convertRvalueExpression(*expr)
+                         : context.convertLvalueExpression(*expr);
         if (!value)
           return failure();
         portValues.insert({port, value});
@@ -145,8 +147,8 @@ struct MemberVisitor {
       // it up into multiple sub-values, one for each of the ports in the
       // multi-port.
       if (const auto *multiPort = con->port.as_if<MultiPortSymbol>()) {
-        // TODO: Convert as lvalue.
-        auto value = context.convertExpression(*expr);
+        // Convert as lvalue.
+        auto value = context.convertLvalueExpression(*expr);
         if (!value)
           return failure();
         unsigned offset = 0;
@@ -157,9 +159,12 @@ struct MemberVisitor {
           auto sliceType = context.convertType(port->getType());
           if (!sliceType)
             return failure();
-          auto slice =
-              builder.create<moore::ExtractOp>(loc, sliceType, value, index);
-          // TODO: Read to map to rvalue for input ports.
+          Value slice = builder.create<moore::ExtractRefOp>(
+              loc, moore::RefType::get(cast<moore::UnpackedType>(sliceType)),
+              value, index);
+          // Read to map to rvalue for input ports.
+          if (port->direction == slang::ast::ArgumentDirection::In)
+            slice = builder.create<moore::ReadOp>(loc, sliceType, slice);
           portValues.insert({port, slice});
           offset += width;
         }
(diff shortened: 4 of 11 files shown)


hailongSun2000:
I think creating a new PR is more convenient to talk about.

hailongSun2000:
My idea form is 
```
// CHECK: %a = moore.variable : ref<i32>
/// CHECK: %b = moore.variable : i32 -> ref<i32>
int a, b = 1;

always_comb begin
  // CHECK : %0 = moore.constant 1 : i32
  // CHECK : moore.blocking_assign %a, %0 : ref<i32>, i32
  a = 1;
  
  // CHECK : %1 = moore.read %a : ref<i32> -> i32
  // CHECK : %2 = moore.add %1, %1 : i32
  // CHECK :  moore.blocking_assign %b, %2 : ref<i32>, i32
  b = a + a;
end
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -148,18 +148,18 @@ def ProcedureOp : MooreOp<"procedure", [
...
   let description = [{
     See IEEE 1800-2017 § 6.8 "Variable declarations".
   }];
   let arguments = (ins StrAttr:$name, Optional<UnpackedType>:$initial);
-  let results = (outs Res<UnpackedType, "", [MemAlloc]>:$result);
+  let results = (outs Res<RefType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
     custom<ImplicitSSAName>($name) ($initial^)? attr-dict
-    `:` type($result)
+    `:` type($initial) `->` type($result)
```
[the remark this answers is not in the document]
fabianschuiki: I don't think that's needed. You should be able to keep just the old `type($result)`, but update the `OptionalTypesMatchWith` to unpack/pack the RefType. The `"$_self"` there is actually a piece of C++ code that can transform the type between `result` and `initial`. You can probably put something like the following there (I don't know which direction is going to work):
- `cast<RefType>($_self).getNestedType()`
- `RefType::get($_self)`
One of them is probably going to work 😁

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -230,15 +230,27 @@ def ReadLValueOp : MooreOp<"read_lvalue", [SameOperandsAndResultType]> {
...
+  let description = [{
+    //TODO:
+  }];
+  let arguments = (ins Arg<RefType, "", [MemRead]>:$input);
+  let results = (outs UnpackedType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input) `->` type($result)
+  }];
+}
+
```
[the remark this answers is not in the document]
fabianschuiki: I think you need a `TypesMatchWith` constraint on the op here similar to `VariableOp` (just not optional), which takes the result type and wraps it in a ref type for the input. That would also allow you to omit the `type($input) ->` part, and it would add a verifier to the op which makes sure that the input is `ref<T>` and the result just `T`.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -230,15 +230,27 @@ def ReadLValueOp : MooreOp<"read_lvalue", [SameOperandsAndResultType]> {
...
 //===----------------------------------------------------------------------===//
 
 class AssignOpBase<string mnemonic, list<Trait> traits = []> :
-    MooreOp<mnemonic, traits # [SameTypeOperands]> {
-  let arguments = (ins UnpackedType:$dst, UnpackedType:$src);
+    MooreOp<mnemonic, traits # []> {
+  let arguments = (ins RefType:$dst, UnpackedType:$src);
   let assemblyFormat = [{
-    $dst `,` $src attr-dict `:` type($src)
+    $dst `,` $src attr-dict `:` type($dst) `,` type($src)
```
fabianschuiki, line 253: You can use `TypesMatchWith` constraints here to encode that `dst` has to have type `ref<T>` and src type `T`. Similar to the suggestions above. Check out the `sv.assign` op in the SV dialect which does a similar thing. That also allows you to drop `type($dst)`.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -305,6 +305,19 @@ def UnpackedStructType : StructLikeType<
...
+  let assemblyFormat = [{
+    `<` $nestedType `>`
+  }];
+
+  let builders = [
+    AttrBuilderWithInferredContext<(ins "::mlir::Type":$nestedType), [{
+      return $_get(nestedType.getContext(), nestedType);
+    }]>
+  ];
+}
```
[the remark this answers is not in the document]
fabianschuiki: A `getNestedType()` method should be generated for you automatically by TableGen 😃

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
```diff
@@ -64,6 +64,7 @@ struct Context {
 
   // Convert an expression AST node to MLIR ops.
   Value convertExpression(const slang::ast::Expression &expr);
+  Value &convertRHSExpression(const slang::ast::Expression &expr);
```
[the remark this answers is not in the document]
fabianschuiki: Yes I totally agree! In the software world these are usually called L-values and R-values, referring to whether the expression appeared on the **L**eft or **R**ight side of an assignment. This only talks about assignments, not other expressions that have an LHS and RHS operand.

The reason is that an addition like `a + b` just needs values for `a` and `b` that it can add. But an assignment like `c = d` requires _a pointer to where `c`'s new value can be stored_ and the value for `d`. This is what lvalue vs. rvalue lowerings handle: lvalue lowering gives you _an assignable version_ of the expression, while rvalue lowering gives you _the value_ of the exprression. And lvalue lowering usually only supports a small subset of all possible expressions , e.g. only variable names, array indexing, and struct fields. While rvalue lowering supports all possible expressions.

So I would call these specifically `convertLvalueExpression` and `convertRvalueExpression`, to differentiate lvalue/rvalues in assignments from LHS/RHS operands in arbitrary binary expressions. `convertLvalueExpression` would then only support variable names in the beginning and would just return the `ref<T>` of that variable. And in `convertRvalueExpression`, when you see a variable name, you'd actually `moore.read` that variable to turn it into a `T`.

include/circt/Dialect/Moore/MooreOps.td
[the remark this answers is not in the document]
hailongSun2000: `cast<RefType>($_self).getNestedType()` can work.

fabianschuiki:
> @hailongSun2000  My idea form is [...]

It would be cool if we could use type constraints on the op (`TypesMatchWith` and friends) to remove redundant types in the syntax. For example, assignment always requires `ref<T>, T` as types, so it would be great if the syntax only mentioned `T`. (MLIR can often infer types based on these constraints.)

Something like this would be very cool:
```mlir
// CHECK: %a = moore.variable : ref<i32>
// CHECK: %b = moore.variable %c1_i32 : ref<i32>  <--- I think MLIR can infer `i32` for the operand from `ref<i32>` with a constraint
int a, b = 1;

always_comb begin
  // CHECK : %0 = moore.constant 1 : i32
  // CHECK : moore.blocking_assign %a, %0 : i32  <--- MLIR can infer `ref<i32>` for the LHS from the `i32` of the RHS with a constraint
  a = 1;
  
  // CHECK : %1 = moore.read %a : i32  <--- MLIR can infer `ref<i32>` for the operand from the `i32` of the result with a constraint
  // CHECK : %2 = moore.add %1, %1 : i32
  // CHECK :  moore.blocking_assign %b, %2 : i32  <--- same as above, MLIR should be able to infer this
  b = a + a;
end
```

hailongSun2000:
Hey, @fabianschuiki. I added the related `TypeMatchWith` constraints for `variable`, `read`, and `assignment`. And I created a new function named `const Value &convertLvalueExpression()` to handle the LHS of assignments. But there is a problem with the `return` of this function(Reference to stack memory associated with local variable 'lValue' returned :sweat_smile: ).

hailongSun2000:
Example:
```
module Foo;
  int x; 
  int y, z;
  always_comb begin
    x = y;
    x = 1;
  end
endmodule
```
----
```
module {
  moore.module @Foo {
    %x = moore.variable  : <!moore.i32>
    %y = moore.variable  : <!moore.i32>
    %z = moore.variable  : <!moore.i32>                     // await to tweak
    moore.procedure always_comb {
      %0 = moore.read %y : i32
      moore.blocking_assign %x, %0 : <!moore.i32>
      %1 = moore.constant 1 : i32
      moore.blocking_assign %x, %1 : <!moore.i32>
    }
  }
}
```

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
```diff
@@ -63,7 +63,8 @@ struct Context {
   LogicalResult convertStatement(const slang::ast::Statement &stmt);
 
   // Convert an expression AST node to MLIR ops.
-  Value convertExpression(const slang::ast::Expression &expr);
+  Value convertRvalueExpression(const slang::ast::Expression &expr);
+  const Value &convertLvalueExpression(const slang::ast::Expression &expr);
```
fabianschuiki, line 67:
```suggestion
  Value convertLvalueExpression(const slang::ast::Expression &expr);
```

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -305,6 +305,19 @@ def UnpackedStructType : StructLikeType<
   }];
 }
 
+def RefType : MooreTypeDef<"Ref", []>{
+  let mnemonic = "ref";
+  let parameters = (ins "::mlir::Type":$nestedType);
```
fabianschuiki, line 310: You could probably make the nested type an `UnpackedType`, to accept more or less any SV type here. That would be more specific than MLIR `Type`, which could be any weird thing in MLIR that we don't know how to work with.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -230,15 +230,32 @@ def ReadLValueOp : MooreOp<"read_lvalue", [SameOperandsAndResultType]> {
...
+  let summary = "Like llvm.load";
+  let description = [{
+    //TODO:
+  }];
+  let arguments = (ins Arg<RefType, "", [MemRead]>:$input);
+  let results = (outs UnpackedType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($result)
+  }];
+  let hasVerifier = 1;
```
fabianschuiki, line 246: The `TypesMatchWith` you have added should already cover everything you have in your verifier: if you make RefType wrap around an `UnpackedType` instead of MLIR's generic `Type`, RefType will do the checking for UnpackedType for you. And the `TypesMatchWith` will do the checking that input's type is the result's type wrapped in RefType. Therefore the hand-written verifier should no longer be needed.

lib/Conversion/ImportVerilog/ImportVerilogInternals.h
hailongSun2000: In this case, we will no longer return a reference. Maybe I think we only need a wrapper for the type. Not `Value &`.

include/circt/Dialect/Moore/MooreOps.td
hailongSun2000: I get it!

include/circt/Dialect/Moore/MooreTypes.td
hailongSun2000: You're totally right! :smile:

hailongSun2000:
I'm curious why the generated type is `<i32>/<real>/etc` without `ref`. I checked the related .inc file, `generatedTypePrinter()` has a `case` branch to print `moore::RefType::getMnemonic()`. However, it doesn't work :thinking: .

fabianschuiki:
This is a feature of the MLIR assembly printer/parser: if the type is already "clear" from the operation, MLIR will not print the name of the type anymore, and instead directly print the parameters (`<...>`) of the type. In the case of `moore.variable`, because you are defining the `$result` to be of type `RefType`, MLIR abbreviates `%a = moore.variable : !moore.ref<T>` to `%a = moore.variable : <T>`.

You can see a similar effect when you specify a `$result` to be of type `UnpackedType`: MLIR abbreviates `!moore.l42` to `l42` because it's already clear that the result must be a Moore type, so the assembly text does not need to provide any additional information.

hailongSun2000:
I think whether can we directly print `%a = moore.variable : <i32>`? We limit the result type of variables that must be `RefType`.

hailongSun2000:
> This is a feature of the MLIR assembly printer/parser: if the type is already "clear" from the operation, MLIR will not print the name of the type anymore, and instead directly print the parameters (`<...>`) of the type. In the case of `moore.variable`, because you are defining the `$result` to be of type `RefType`, MLIR abbreviates `%a = moore.variable : !moore.ref<T>` to `%a = moore.variable : <T>`.
> 
> You can see a similar effect when you specify a `$result` to be of type `UnpackedType`: MLIR abbreviates `!moore.l42` to `l42` because it's already clear that the result must be a Moore type, so the assembly text does not need to provide any additional information.

I thought the only difference between AsmPrint and DialectAsmPrint was that the former does not print `!moore.` Like abbreviate `!moore.array<2xT>` to `array<2xT>`. I didn't realize that AsmPrint would also estimate the result type(`RefType`) of a variable

fabianschuiki:
> I think whether can we directly print `%a = moore.variable : <i32>`?

Yes this sounds perfect :partying_face:

hailongSun2000:
Thanks for the patient explanation.

fabianschuiki:
> I thought the only difference between AsmPrint and DialectAsmPrint was that the former does not print `!moore.` Like abbreviate `!moore.array<2xT>` to `array<2xT>`. I didn't realize that AsmPrint would also estimate the result type(`RefType`) of a variable

It all depends where you start the call to `print`. In the parser/printer of ops like `moore.variable`, a call to `resultType.print()` (or similar) is used to print the type. If your op specifies a result type `RefType`, that `resultType.print()` is actually a call to `RefType::print()`. If the result type is `UnpackedType` it's `UnpackedType::print()`. Otherwise it's plain old `mlir::Type::print()`. The op parser/printer calls the most specific parse/print function on the type it can.

The type printing then works like this:
- `mlir::Type::print()` prints `!moore.` and calls the dialect and/or generated type printer.
- `UnpackedType::print()` directly calls the generated type printer.
- The generated type printer prints `ref` and calls `RefType::print()`.
- `RefType::print()` prints `<T>` (and recursively calls `T.print()` with the most specific parse/print function).

So depending on which print you call, you get either `!moore.ref<i42>`, `ref<i42>`, or just `<i42>`.

hailongSun2000:
Oh~ Heavenly nectar pouring down! Before this, I thought there was only a print difference between `IntType`(`i/l` + `width`) and other types(`generatedTypePrinter()`).

hailongSun2000:
I'm not sure whether we need to limit `RefType` only to `moore.variable`, `moore.blocking_assign`, and `moore.read`. Variables will be used for `moore.assign` and `moore.nonblocking_assign`. However, due to `moore.blocking_assign` has the `MemWrite` feature, can I consider only `moore.blocking_assign` having the feature of memory store? Meanwhile, for example
```
int a, b;
assign b = a;
```

If `RefType` doesn't service for `moore.assign`(only `UnpackedType`),  we have to generate `moore.read %b` to convert the type of `int b;`(`<i32>`-->`i32`). I think this is an error action to read LHS.
### moore IR
```
  %a = moore.variable : <i32>
  %b = moore.variable : <i32>
  %0 = moore.read %a : i32
  %1 = moore.read %b : i32      // an error IR
  moore.assign %1, %0 : i32
```
If `RefType` service for `moore.assign`(`RefType:$dst`, `UnpackedType:$src`), for example
```
wire a, b;
assign b = a;
```
### moore IR
```
 %a = moore.net "wire" : <l1>
 %b = moore.net "wire" : <l1>
 %0 = moore.read %a : l1
 moore.assign %b, %0 : l1
```
I think only `moore.variable` has the `MemAlloca` feature, `moore.net` doesn't. So `%0 = moore.read %a : l1` looks like strange.

hailongSun2000:
Sorry that! I checked the definition of `moore.net` in your private repository(slang-fronted branch), and I found the `MemAlloca` feature of `moore.net` was deleted at this PR(https://github.com/llvm/circt/pull/6884). It's my fault.

fabianschuiki:
Your IR where `moore.assign` requires `(RefType, UnpackedType)` looks great! And the IR you posted looks correct to me:
```mlir
 %a = moore.net "wire" : <l1>
 %b = moore.net "wire" : <l1>
 %0 = moore.read %a : l1
 moore.assign %b, %0 : l1
```

I think it's okay to have `moore.read` here in the module for the right-hand side of the assignment.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -149,14 +149,14 @@ def ProcedureOp : MooreOp<"procedure", [
...
 ]> {
   let summary = "A variable declaration";
   let description = [{
     See IEEE 1800-2017 § 6.8 "Variable declarations".
   }];
   let arguments = (ins StrAttr:$name, Optional<UnpackedType>:$initial);
-  let results = (outs Res<UnpackedType, "", [MemAlloc]>:$result);
+  let results = (outs Res<RefType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
     custom<ImplicitSSAName>($name) ($initial^)? attr-dict
```
hailongSun2000, line 161: I think `` is ignored here. But #7112 complements it.

hailongSun2000:
Hey, @fabianschuiki. I finished introducing `RefType` to moore. If there is no problem, I think it can wait to be merged until #7112 lands.

fabianschuiki:
Sounds great, let me push #7112 over the finish line.

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: The `` prevents an additional space to be printed in case there is no name:
```
moore.variable  %0 : i32  // without ``
moore.variable %0 : i32   // with ``
```

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -325,6 +325,29 @@ def UnpackedUnionType : StructLikeType<
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// Type wrapper
```
fabianschuiki, line 329:
```suggestion
// Reference type wrapper
```

include/circt/Dialect/Moore/MooreOps.td
hailongSun2000: You're totally right! I remember you mentioned this difference in another PR.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -146,7 +148,7 @@ struct MemberVisitor {
       // multi-port.
       if (const auto *multiPort = con->port.as_if<MultiPortSymbol>()) {
         // TODO: Convert as lvalue.
-        auto value = context.convertExpression(*expr);
+        auto value = context.convertRvalueExpression(*expr);
```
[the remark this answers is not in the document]
fabianschuiki: I think we need copies of `extract` and the union/field access ops that explicitly work on ref types. Something like `extract_ref`.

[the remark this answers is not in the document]
hailongSun2000: Cool! You're all right! And I think maybe we also need copies of `concat`. For example :
### .sv
```
module top;
  int u, v, w;
  initial begin
  {u, v} = w;
  end
endmodule
```
### AST
```
"body": {
                  "expr": {
                    "kind": "Assignment",
                    "type": "bit[63:0]",
                    "left": {
                      "kind": "Concatenation",
                      "type": "bit[63:0]",
                      "operands": [
                        {
                          "kind": "NamedValue",
                          "type": "int",
                          "symbol": "2199025167712 u"
                        },
                        {
                          "kind": "NamedValue",
                          "type": "int",
                          "symbol": "2199025168088 v"
                        }
                      ]
                    },
                    "right": {
                      "kind": "Conversion",
                      "type": "bit[63:0]",
                      "operand": {
                        "kind": "Conversion",
                        "type": "bit signed[63:0]",
                        "operand": {
                          "kind": "NamedValue",
                          "type": "int",
                          "symbol": "2199025168464 w"
                        }
                      }
                    },
                    "isNonBlocking": false
                  }
```
So from my perspective, we need to implement something that will be regarded as LHS of assignments like `concat_ref`, `extract_ref`, and others(I'm not sure).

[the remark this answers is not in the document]
fabianschuiki: Great idea! 🥳

fabianschuiki (approved):
This is really really nice! Thanks a lot for working on this :partying_face:. Looks good to me!

PR #6115, opened 2023-09-12
[FIRRTL][HoistPassthrough] Add pass to hoist must-driven output ports.

dtzSiFive:
Basic driver analysis to detect must-driven values and hoist/rematerialize them using equivalent driver at instantiation site.

End-to-end example included for testing `-disable-opt` demonstrates why this is desirable -- here's a before/after diff:

```patch
--- /proc/self/fd/11	2023-09-12 15:22:35.254920380 -0500
+++ /proc/self/fd/12	2023-09-12 15:22:35.254920380 -0500
@@ -1,39 +1,10 @@
-// Generated by CIRCT firtool-1.54.0-30-g7bb1650e3
-module Child(
-  input  [4:0] x,
-  output [4:0] y
-);
-
-  assign y = x;
-endmodule
-
-module Mid(
-  input  [4:0] x,
-  output [4:0] y
-);
-
-  Child c (
-    .x (x),
-    .y (y)
-  );
-endmodule
-
+// Generated by CIRCT firtool-1.54.0-29-gc3332a678
 module HWMultiLevel(
   input  [4:0] x,
                x2,
   output [4:0] y
 );
 
-  wire [4:0] _m2_y;
-  wire [4:0] _m_y;
-  Mid m (
-    .x (x),
-    .y (_m_y)
-  );
-  Mid m2 (
-    .x (x2),
-    .y (_m2_y)
-  );
-  assign y = _m_y & _m2_y;
+  assign y = x & x2;
 endmodule
```

This is originally motivated by need to do bottom-up half of "de-squiggling" to handle input probe u-turns, context-sensitively, in an efficient way but leaves cleanup and probe-specific diagnostics to other passes (does not sink equivalently-driven "n-turns" that remain).

Current analysis is intended to be simple (selectively chase output ports through must-drive chains) as that is sufficient for probes ("static single connect"+strict semantics, can't drive partially re:indexing) while lending itself to HW signals naturally.  Support for more sophisticated scenarios and more powerful framing is left for future work if we like this direction and there's need (e.g., more effective when enabling aggregate preservation, ability to hoist logic as well, so on), lots of adventures to be had here! :grin:

Safety is paramount, especially regarding the interpretation of some darker corners of the semantics of FIRRTL re:what it means to annotate, dontTouch, so on, a signal.  Care was taken to align with rest of pipeline and existing use-cases/flows but critical eye appreciated.

[...]

PR #7097, opened 2024-05-28
[FIRRTL] Drop Hoistpassthroughs pass.

dtzSiFive:
This has only been used for probes, but without input probes this is no longer useful or necessary.

The HW-signal hoisting works well (if limited) and has no known issues[1] but has yet to be used in practice due primarily to passthroughs being intentional occasionally and lacking a mechanism to capture or distill this intent properly (it must be ""optimized"" yet not).

Since this is unused, and lacking traction on the above, remove from pipeline and delete this for now instead of having it bitrot and be a maintenance burden while adding/completing new features.

Lives on in version control history!

[1] FWIW, see https://github.com/llvm/circt/pull/6115 for both small example and before/after on chipyard, as well as testing internally back when this was introduced.


mikeurbach:
Cleanup sounds good to me. There may be a use case for hoisting property passthroughs some day, so perhaps this could be revived at that point.

PR #7098, opened 2024-05-28
[CombToSMT] Register dependency on func

### LINKED ISSUE #7096: [CombToSMT] Func dialect dependency is missing
```mlir
../circt//build/bin/circt-opt bar.mlir -lower-comb -convert-comb-to-smt -mlir-print-ir-after-all
// -----// IR Dump After LowerComb (lower-comb) //----- //
module {
  hw.module @test(in %arg0 : i1, in %arg1 : i1, out out : i1) {
    %true = hw.constant true
    %false = hw.constant false
    %0 = comb.mux %arg1, %false, %true : i1
    %1 = comb.mux %arg1, %false, %true : i1
    %2 = comb.mux %arg0, %0, %1 : i1
    hw.output %2 : i1
  }
}


LLVM ERROR: Building op `func.func` but it isn't known in this MLIRContext: the dialect may not be loaded or this operation hasn't been added by the dialect. See also https://mlir.llvm.org/getting_started/Faq/#registered-loaded-dependent-whats-up-with-dialects-management
```

maerhart:
Fixes #7096

### DIFF
--- a/include/circt/Conversion/Passes.td
+++ b/include/circt/Conversion/Passes.td
@@ -709,7 +709,9 @@ def ConvertCombToSMT : Pass<"convert-comb-to-smt", "mlir::ModuleOp"> {
   let summary = "Convert combinational ops and constants to SMT ops";
   // Need to depend on HWDialect because some 'comb' canonicalization patterns
   // build 'hw.constant' operations.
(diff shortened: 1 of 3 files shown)


teqdruid (approved):
lgtm

PR #7104, opened 2024-05-29
[Verif] Allow for `disable` to be optional and infer it from property

dobios:
This PR modifies the `verif::clocked_assertlike` ops to allow for the disable to be optional. In the case where it is omitted, it will be inferred from the property itself following the folding patterns defined in `LTLToCore`, i.e.  
- `property = disable: i1 OR prop : !ltl.property` for assume and assert ops
-  `property =  NOT(disable: i1) AND prop : !ltl.property` for cover ops.  

This should allow us to get rid of `ltl.disable` in the new redesign of ltl


lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4804,23 +4806,176 @@ LogicalResult StmtEmitter::emitVerifClockedAssertLike(
...
+          mlir::m_Op<ltl::OrOp>(mBool(&disableVal), mProperty(&property)));
+
+  if (!disablePattern) {
+    mlir::OpBuilder builder(op.getContext());
+    mlir::Value f = builder.createOrFold<hw::ConstantOp>(
+        op.getLoc(), builder.getI1Type(), 0);
+
+    return emitVerifClockedAssertLike(op, op.getProperty(), f, op.getClock(),
+                                      op.getEdge(), PPExtString("assume"));
+  }
```
uenoku, line 4931: Please don't mutate IR in verilog emission. IR might be read from a different thread so it's not thread safe.

uenoku (commented):
Can this transformation be a canonicalization for clocked verif ops? I'm more comfortable if ExportVerilog performs direct emission and not heavily pattern matching.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4804,23 +4806,176 @@ LogicalResult StmtEmitter::emitVerifClockedAssertLike(
...
+
+  // Try to identify a disable pattern in the property
+  Value disableVal, property, one;
+  bool disablePattern =
+      mlir::matchPattern(
+          op.getProperty(),
+          mlir::m_Op<comb::AndOp>(
+              mProperty(&property),
+              mlir::m_Op<comb::XorOp>(mBool(&disableVal), mOne(&one)))) ||
+      // Not sure if the pattern matchers are naturally commutative, so just to
```
uenoku, line 4951: Can we define "canonical order" of operands for `ltl::AndOp`? For example we can create canonicalizer  that moves every ltl.property operand to the head. This logic is similar to comb::XOrOp::isBinaryNot(https://github.com/llvm/circt/blob/76b65f8606de447a85ef63c9c98ae9b1e2bd4b7f/lib/Dialect/Comb/CombOps.cpp#L224-L231). MLIR canonicalizer pushes constant operands to the tail for a commutative operation.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
dobios: Ah ok, I guess I can simply omit the disable all together, that would actually make a lot more sense.

dobios:
> Can this transformation be a canonicalization for clocked verif ops? I'm more comfortable if ExportVerilog performs direct emission and not heavily pattern matching.
@uenoku  The goal with this is to eventually get rid of `ltl.disable` as it is honestly simply something that exists in SVA and not in the general concept of temporal properties, so one step for moving ltl away from being an SVA AST is getting of this op. So this is really just something that is done to make the emitted SVA more performant and not something we want or need to have as part of the IR. At least this is my goal with this PR, so I don't think that this can be a canonicalization pass

lib/Conversion/ExportVerilog/ExportVerilog.cpp
fabianschuiki: 👍  You can pass in a null value where you currently have `f`, and then check in `emitVerifClockedAssertLike` whether the disable is non-null, and only then emit the `disable iff`.

dobios: that's exactly what I just did lol

uenoku:
I'm not sure moving pattern matches to canonicalization contradicts with your goal? Canonicalizer wouldn't introduce ltl.disable and just set disable operand. I think you don't have to modify ExportVerilog if `disable` is set in canonicalizer like below:

```c++
caonicalizer:
LogicalResult ClockAssertOp::canonicalize(Rewriter &rewriter) {
  Value inferredDisable = // pattern match;
  if (inferredDisable) return failure();
  setDisable(inferredDisable);
  return success();
}
```

dobios:
I need to go back to the drawing board about this, thanks to a breaking case caught by @youngar

PR #7108, opened 2024-05-31
[FIRRTL] Generate memportaccess op index as UInt.

dtzSiFive:
Create these constants as unsigned of unknown width, instead of s64.

This way when connecting to the address, we don't emit a connnect between s64 and something like u3
(requiring casts/trim/etc).


dtzSiFive:
Thanks! Minor, just noticed while working on something else.

PR #7109, opened 2024-05-31
[FIRRTL] Emulate tap-as-passive for no-ref-ports option.

dtzSiFive:
Taps (using probes) support a non-passive source with a passive sink (by their nature), but the hidden option to avoid probes does not.

Detect this specifically and fix by using passive type for wiring from source.

This allows taps of non-ground (non-passive) that work with probes to also work with this option.


prithayan (approved):
Looks good.

dtzSiFive:
Thanks!

PR #7111, opened 2024-06-01
[FIRRTL][Sim] Add `fopen` and `fclose` intrinsics, add `fd` operand to `printf`

### LINKED ISSUE #7092: [SV] Add `fopen` and `fclose`
Adding `fopen` and `fclose` allows `fwrite` to real files instead of just standard streams.

According to SV spec 21.3.1, the return value `fd` of `fopen` is an `integer`. The CIRCT SV dialect does not seem to implement the `integer` type at the moment.

I am preparing PR to add these two system tasks. For `integer`, should I implement it in its entirety as a new type, or try to reuse the existing `Inout` type? If the former is better, that means I'll have to implement a new op like `sv.assign_integer` for it as well, right?

CC @seldridge @nandor

SpriteOvO:
This PR enables the SystemVerilog generated by Chisel&CIRCT can be `fwrite` to different files by the simulator.

FIRRTL Spec changes: chipsalliance/firrtl-spec#213 (some discussion already there)

Closes #7092, CC @seldridge @sequencer @uenoku

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -4334,7 +4334,8 @@ LogicalResult FIRRTLLowering::visitStmt(RefReleaseInitialOp op) {
 LogicalResult FIRRTLLowering::visitStmt(PrintFOp op) {
   auto clock = getLoweredNonClockValue(op.getClock());
   auto cond = getLoweredValue(op.getCond());
-  if (!clock || !cond)
+  auto fd = getLoweredValue(op.getFd());
+  if (!clock || !cond || !fd)
     return failure();
 
   SmallVector<Value, 4> operands;
@@ -4359,9 +4360,7 @@ LogicalResult FIRRTLLowering::visitStmt(PrintFOp op) {
       ifCond = builder.createOrFold<comb::AndOp>(ifCond, cond, true);
 
       addIfProceduralBlock(ifCond, [&]() {
-        // Emit the sv.fwrite, writing to stderr by default.
-        Value fdStderr = builder.create<hw::ConstantOp>(APInt(32, 0x80000002));
-        builder.create<sv::FWriteOp>(fdStderr, op.getFormatString(), operands);
+        builder.create<sv::FWriteOp>(fd, op.getFormatString(), operands);
       });
     });
   });
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -1634,6 +1634,7 @@ struct FIRRTLLowering : public FIRRTLVisitor<FIRRTLLowering, LogicalResult> {
   LogicalResult visitExpr(HasBeenResetIntrinsicOp op);
   LogicalResult visitStmt(UnclockedAssumeIntrinsicOp op);
   LogicalResult visitExpr(FOpenIntrinsicOp op);
+  LogicalResult visitExpr(FCloseIntrinsicOp op);
 
   // Other Operations
   LogicalResult visitExpr(BitsPrimOp op);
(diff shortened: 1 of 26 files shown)


lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -4669,6 +4670,26 @@ LogicalResult FIRRTLLowering::visitStmt(UnclockedAssumeIntrinsicOp op) {
...
+LogicalResult FIRRTLLowering::visitExpr(FOpenIntrinsicOp op) {
+  auto resultTy = lowerType(op.getType());
+  if (!resultTy)
+    return failure();
+
+  auto filename = builder.create<sv::ConstantStrOp>(op.getFilename());
+  auto mode = builder.create<sv::ConstantStrOp>(op.getMode());
+
+  return setLoweringTo<sv::SystemFunctionOp>(
+      op, resultTy, builder.getStringAttr("fopen"), ValueRange{filename, mode});
```
uenoku, line 4682: `$fopen` is `task` so you cannot put this into non-procedural region. Please put this to initial block and assign return value to a register.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -4669,6 +4670,26 @@ LogicalResult FIRRTLLowering::visitStmt(UnclockedAssumeIntrinsicOp op) {
...
+}
+
+LogicalResult FIRRTLLowering::visitExpr(FCloseIntrinsicOp op) {
+  auto fd = getLoweredValue(op.getFd());
+  auto result = builder.create<sv::SystemFunctionOp>(
+      NoneType::get(builder.getContext()), builder.getStringAttr("fclose"),
+      ValueRange{fd});
+  return success();
```
uenoku, line 4690: This also needs to be in a procedural region. I noticed currently there is no support for `final` statement. That seems reasonable addition to SV dialect.

uenoku (commented):
Generally the implementation looks great. Let's wait for how https://github.com/chipsalliance/firrtl-spec/pull/213 goes  :+1:

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -2824,14 +2824,15 @@ ParseResult FIRStmtParser::parseMemPort(MemDirAttr direction) {
...
+/// printf ::= 'printf(' exp exp exp StringLit exp* ')' name? info?
 ParseResult FIRStmtParser::parsePrintf() {
   auto startTok = consumeToken(FIRToken::lp_printf);
 
-  Value clock, condition;
+  Value clock, condition, fd;
   StringRef formatString;
   if (parseExp(clock, "expected clock expression in printf") ||
       parseExp(condition, "expected condition in printf") ||
+      parseExp(fd, "expected fd expression in printf") ||
```
uenoku, line 2835: It 's necessary to accept old syntal. Please add 
a condition for FIRRTL-spec version.

test/firtool/fopen-fclose.fir
```diff
@@ -0,0 +1,27 @@
...
+FIRRTL version 4.0.0
+circuit FOpenFCloseTest:
+  ; CHECK-LABEL: @FOpenFCloseTest
+  public module FOpenFCloseTest :
+    input clock : Clock
+    input cond : UInt<1>
+    input var : UInt<32>
+    node fd = intrinsic(circt_fopen<filename = "file.txt", mode = "a"> : SInt<32>)
+    printf(clock, cond, fd, "test %d\n", var)
+    intrinsic(circt_fclose, fd)
```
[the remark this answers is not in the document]
uenoku: Ah, I was assuming `fclose` intrinsic is lowered in `fclose` in `final` statement. Alternatively we could add `clock` and `enable` operands either.

SpriteOvO:
Closed as in favor of #7983.

PR #7112, opened 2024-06-01
[Moore] Add module and instance port support

fabianschuiki:
Add module types and support for ports to the `moore.module` and `moore.instance` operations. Also add a `moore.output` operation as a terminator for modules, similar to the HW dialect. Extend ImportVerilog to generate ports.

The ports on SVModuleOp follow a similar strategy as in HWModuleOp: all input, inout, and ref ports are passed to the module as inputs and are carried inside the module as block arguments, while output ports are assigned by the OutputOp terminator. The Moore dialect reuses the `hw::ModuleType` but does not use the `inout` direction. Instead, inout ports will be represented as inputs with a `net<T>` wrapper type, while ref ports will be wrapped as `ref<T>`.

Instances work identically to the HW dialect: input, inout, and ref port connections are carried as operands, while output ports are represented as results.

This commit also adds module and instance port support to the ImportVerilog conversion. Regular ports are mapped to corresponding ports on the module with the appropriate direction. Multi-ports, which are a weird quirk of SystemVerilog that allow multiple ports to be grouped up and presented to the outside as a single named port, are split up into individual module ports. This is necessary since this group can contain ports of different directions.

Inside a module Slang automatically generates local net or variable declarations for all ports. The user may specify these declarations themselves when using non-ANSI port lists, which Slang handles properly. ImportVerilog inserts appropriate continuous assignments to drive the actual input port value onto the local net or variable declaration, and to drive the local declaration's value onto the actual output port of the module. This properly adapts from SystemVerilog's assignable and connectable ports that feel like nets or variables, to the Moore dialect's by-value passing of inputs and outputs.

Instances in Slang have expressions connected to their ports. Input ports lower this expression and directly use the result as an operand. Output ports lower this expression and drive it through a continuous assignment inserted after the instance, with the instance's corresponding result on the right-hand side of the assignment.

Once we have a `ref<T>` type, and later potentially also a `net<T>` type, the port lowering shall be revisited to ensure that inout and ref ports are mapped to net and ref types, respectively. The lowering of expressions connected to ports requires more care to ensure that they are appropriately lowered to lvalues or rvalues, as needed by the port direction. A `moore.short_circuit` operation or similar would help to connect inout ports to the local net declarations in a module, and to capture `alias` statements.

---------

### DIFF
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -94,15 +94,110 @@ struct MemberVisitor {
   // Skip typedefs.
   LogicalResult visit(const slang::ast::TypeAliasType &) { return success(); }
 
+  // Skip ports which are already handled by the module itself.
+  LogicalResult visit(const slang::ast::PortSymbol &) { return success(); }
+  LogicalResult visit(const slang::ast::MultiPortSymbol &) { return success(); }
+
   // Handle instances.
   LogicalResult visit(const slang::ast::InstanceSymbol &instNode) {
-    auto targetModule = context.convertModuleHeader(&instNode.body);
-    if (!targetModule)
-      return failure();
+    using slang::ast::ArgumentDirection;
+    using slang::ast::AssignmentExpression;
+    using slang::ast::MultiPortSymbol;
+    using slang::ast::PortSymbol;
 
-    builder.create<moore::InstanceOp>(
-        loc, builder.getStringAttr(instNode.name),
-        FlatSymbolRefAttr::get(targetModule.getSymNameAttr()));
+    auto *moduleLowering = context.convertModuleHeader(&instNode.body);
+    if (!moduleLowering)
+      return failure();
+    auto module = moduleLowering->op;
+    auto moduleType = module.getModuleType();
+
+    // Prepare the values that are involved in port connections. This creates
+    // rvalues for input ports and appropriate lvalues for output, inout, and
+    // ref ports. We also separate multi-ports into the individual underlying
+    // ports with their corresponding connection.
+    SmallDenseMap<const PortSymbol *, Value> portValues;
+    portValues.reserve(moduleType.getNumPorts());
+
+    for (const auto *con : instNode.getPortConnections()) {
+      const auto *expr = con->getExpression();
+      if (!expr)
+        return mlir::emitError(loc)
+               << "unconnected port `" << con->port.name << "` not supported";
+
+      // Unpack the `<expr> = EmptyArgument` pattern emitted by Slang for
+      // output and inout ports.
+      if (const auto *assign = expr->as_if<AssignmentExpression>())
+        expr = &assign->left();
+
+      // Regular ports lower the connected expression to an lvalue or rvalue and
+      // either attach it to the instance as an operand (for input, inout, and
+      // ref ports), or assign an instance output to it (for output ports).
+      if (auto *port = con->port.as_if<PortSymbol>()) {
+        // TODO: Convert as rvalue for inputs, lvalue for all others.
+        auto value = context.convertExpression(*expr);
+        if (!value)
+          return failure();
+        portValues.insert({port, value});
+        continue;
+      }
+
+      // Multi-ports lower the connected expression to an lvalue and then slice
+      // it up into multiple sub-values, one for each of the ports in the
+      // multi-port.
+      if (const auto *multiPort = con->port.as_if<MultiPortSymbol>()) {
+        // TODO: Convert as lvalue.
+        auto value = context.convertExpression(*expr);
+        if (!value)
+          return failure();
+        unsigned offset = 0;
+        auto i32 = moore::IntType::getInt(context.getContext(), 32);
+        for (const auto *port : llvm::reverse(multiPort->ports)) {
+          unsigned width = port->getType().getBitWidth();
+          auto index = builder.create<moore::ConstantOp>(loc, i32, offset);
+          auto sliceType = context.convertType(port->getType());
+          if (!sliceType)
+            return failure();
+          auto slice =
+              builder.create<moore::ExtractOp>(loc, sliceType, value, index);
+          // TODO: Read to map to rvalue for input ports.
+          portValues.insert({port, slice});
+          offset += width;
+        }
+        continue;
+      }
+
+      mlir::emitError(loc) << "unsupported instance port `" << con->port.name
+                           << "` (" << slang::ast::toString(con->port.kind)
+                           << ")";
+      return failure();
+    }
+
+    // Match the module's ports up with the port values determined above.
+    SmallVector<Value> inputValues;
+    SmallVector<Value> outputValues;
+    inputValues.reserve(moduleType.getNumInputs());
+    outputValues.reserve(moduleType.getNumOutputs());
+
+    for (auto &port : moduleLowering->ports) {
+      auto value = portValues.lookup(&port.ast);
+      assert(value && "no prepared value for port");
+      if (port.ast.direction == ArgumentDirection::Out)
+        outputValues.push_back(value);
+      else
+        inputValues.push_back(value);
+    }
+
+    // Create the instance op itself.
+    auto inputNames = builder.getArrayAttr(moduleType.getInputNames());
+    auto outputNames = builder.getArrayAttr(moduleType.getOutputNames());
+    auto inst = builder.create<moore::InstanceOp>(
+        loc, moduleType.getOutputTypes(), builder.getStringAttr(instNode.name),
+        FlatSymbolRefAttr::get(module.getSymNameAttr()), inputValues,
+        inputNames, outputNames);
+
+    // Assign output values from the instance to the connected expression.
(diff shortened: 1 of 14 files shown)


fabianschuiki:
@cepheus69 @hailongSun2000 I collected the work you've been doing on module and instance ports and have combined all of that into a PR, with tests and a few tweaks.

hailongSun2000:
Sounds great! :tada:

cepheus69:
It is so great to see its landing. My implementations in my branch are not finished yet 🥲 which is not based on ModuleType. Because I am hesitant to utilize HW ModuleType, I designed a PortInfo Structure for the Moore dialect and added the corresponding builder, parser, verifier, and so on for SVModuleOp. I think I consume much time to struggle with the operation definition function rather than the TableGen file. So could I add these things that I have implemented after this PR?

mingzheTerapines:
Cool! So elegently using these traits and can all writen in tablegen file.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -94,15 +94,110 @@ struct MemberVisitor {
...
+    auto *moduleLowering = context.convertModuleHeader(&instNode.body);
+    if (!moduleLowering)
+      return failure();
+    auto module = moduleLowering->op;
+    auto moduleType = module.getModuleType();
+
+    // Prepare the values that are involved in port connections. This creates
+    // rvalues for input ports and appropriate lvalues for output, inout, and
+    // ref ports. We also separate multi-ports into the individual underlying
+    // ports with their corresponding connection.
```
hailongSun2000, line 117: From my perspective, in your case(module PortsAnsi), `input`--`%a`, `inout`--`%c`, and `ref`--`%d` are all regarded as the rhs of `moore.assign`. And only `output`--`%d` is the operand of `moore.output`.

hailongSun2000:
I read the code in detail. I don't have any questions. 🎉

hailongSun2000 (approved):

uenoku (approved):
Looks reasonable to me! Note that `printModuleSignatureNew` expects non-empty arrays for locations if `mlir-print-debuginfo` is provided so `circt-translate --import-verilog basic.sv -o test.mlir --
mlir-print-debuginfo` should cause assertion failure right now. 
```
circt-translate: /home/uenoku/dev/circt/llvm/llvm/include/llvm/ADT/ArrayRef.h:257: const T& llvm::ArrayRef<T>::operator[](size_t) const [with T = mlir::Location; size_t = long unsigned int]: Assertion `Index < Length && "Invalid index!"' failed.
PLEASE submit a bug report to https://github.com/llvm/circt and include the crash backtrace.
Stack dump:
0.      Program arguments: ./build/bin/circt-translate --import-verilog basic.sv -o test.mlir --mlir-print-debuginfo
 #0 0x000062a9a9caf59f llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /home/uenoku/dev/circt/llvm/llvm/lib/Support/Unix/Signals.inc:727:3
 #1 0x000062a9a9cacfff llvm::sys::RunSignalHandlers() /home/uenoku/dev/circt/llvm/llvm/lib/Support/Signals.cpp:105:20
 #2 0x000062a9a9cad356 SignalHandler(int) /home/uenoku/dev/circt/llvm/llvm/lib/Support/Unix/Signals.inc:413:1
 #3 0x000079699d242520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x000079699d2969fc __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
 #5 0x000079699d2969fc __pthread_kill_internal ./nptl/pthread_kill.c:78:10
 #6 0x000079699d2969fc pthread_kill ./nptl/pthread_kill.c:89:10
 #7 0x000079699d242476 gsignal ./signal/../sysdeps/posix/raise.c:27:6
 #8 0x000079699d2287f3 abort ./stdlib/abort.c:81:7
 #9 0x000079699d22871b _nl_load_domain ./intl/loadmsgcat.c:1177:9
#10 0x000079699d239e96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
#11 0x000062a9a9e674e2 decltype(auto) llvm::dyn_cast<mlir::DictionaryAttr, mlir::Attribute>(mlir::Attribute const&) /home/uenoku/dev/circt/llvm/llvm/include/llvm/Support/Casting.h:650:3
#12 0x000062a9a9e674e2 decltype(auto) llvm::dyn_cast<mlir::DictionaryAttr, mlir::Attribute>(mlir::Attribute const&) /home/uenoku/dev/circt/llvm/llvm/include/llvm/Support/Casting.h:649:37
#13 0x000062a9a9e674e2 circt::hw::module_like_impl::printModuleSignatureNew(mlir::OpAsmPrinter&, mlir::Region&, circt::hw::ModuleType, llvm::ArrayRef<mlir::Attribute>, llvm::ArrayRef<mlir::Location>) /home/uenoku/dev/circt/lib/Dialect/HW/ModuleImplementation.cpp:438:47
#14 0x000062a9a9eaecc2 circt::moore::SVModuleOp::print(mlir::OpAsmPrinter&) /home/uenoku/dev/circt/lib/Dialect/Moore/MooreOps.cpp:32:37
#15 0x000062a9a9e849aa void llvm::detail::UniqueFunctionBase<void, mlir::Operation*, mlir::OpAsmPrinter&, llvm::StringRef>::CallImpl<mlir::Op<circt::moore::SVModuleOp, mlir::OpTrait::OneRegion, mlir::OpTrait::ZeroResults, mlir::OpTrait::ZeroSuccessors, mlir::OpTrait::ZeroOperands, mlir::OpTrait::SingleBlock, mlir::OpTrait::SingleBlockImplicitTerminator<circt::moore::OutputOp>::Impl, mlir::OpTrait::OpInvariants, mlir::BytecodeOpInterface::Trait, mlir::OpTrait::IsIsolatedFromAbove, mlir::SymbolOpInterface::Trait, mlir::OpAsmOpInterface::Trait>::getPrintAssemblyFn()::'lambda'(mlir::Operation*, mlir::OpAsmPrinter&, llvm::StringRef) const>(void*, mlir::Operation*, mlir::OpAsmPrinter&, llvm::StringRef) /home/uenoku/dev/circt/llvm/llvm/include/llvm/ADT/FunctionExtras.h:222:3
#16 0x000062a9a9ea1e11 llvm::unique_function<void (mlir::Operation*, mlir::OpAsmPrinter&, llvm::StringRef) const>::~unique_function() /home/uenoku/dev/circt/llvm/llvm/include/llvm/ADT/FunctionExtras.h:390:7
#17 0x000062a9a9ea1e11 mlir::RegisteredOperationName::Model<circt::moore::SVModuleOp>::printAssembly(mlir::Operation*, mlir::OpAsmPrinter&, llvm::StringRef) /home/uenoku/dev/circt/llvm/mlir/include/mlir/IR/OperationSupport.h:555:39
#18 0x000062a9aa3eef66 llvm::ilist_node_base<true>::getNext() const /home/uenoku/dev/circt/llvm/llvm/include/llvm/ADT/ilist_node_base.h:43:45
#19 0x000062a9aa3eef66 llvm::ilist_node_impl<llvm::ilist_detail::node_options<mlir::Operation, true, false, void, false>>::getNext() /home/uenoku/dev/circt/llvm/llvm/include/llvm/ADT/ilist_node.h:94:66
#20 0x000062a9aa3eef66 llvm::ilist_iterator<llvm::ilist_detail::node_options<mlir::Operation, true, false, void, false>, false, false>::operator++() /home/uenoku/dev/circt/llvm/llvm/include/llvm/ADT/ilist_iterator.h:157:25
#21 0x000062a9aa3eef66 (anonymous namespace)::DummyAliasOperationPrinter::print(mlir::Block*, bool, bool) /home/uenoku/dev/circt/llvm/mlir/lib/IR/AsmPrinter.cpp:727:26
```

lib/Conversion/ImportVerilog/Structure.cpp
fabianschuiki: Yes absolutely! :slightly_smiling_face: My idea was to make almost everything an input and then use your cool new `ref<T>` type that you've been working on:

- `input T` becomes `T` input
- `inout T` becomes `ref<T>` input (maybe later `net<T>` if we decide we need to distinguish vars and nets)
- `ref T` becomes `ref<T>` input
- `output T` becomes `T` output

At the instance, the input would be lowered as rvalue to get a `T`, and all others would be lowered as lvalue to get a `ref<T>`. The refs for inout and ref ports are passed directly to the instance as operands, and outputs generate a `moore.assign` with the lvalue on the left and the instance's output value on the right:

```mlir
%a = moore.variable 
%0 = moore.read %a  // rvalue for `a`
%b = moore.variable  // lvalue for `b`
%c = moore.net  // lvalue for `c`
%d = moore.variable  // lvalue for `d`
%1 = moore.instance ...(a: %0: T, c: %c: net<T>, d: %d: ref<T>) -> (b: T)
moore.assign %b, %1
```

fabianschuiki:
Thanks for catching this @uenoku :heart_eyes:! I'll take another look to fix this.

PR #7113, opened 2024-06-01
[Arc] Add statistics to the FindInitialVectors pass

elhewaty:


include/circt/Dialect/Arc/ArcPasses.td
```diff
@@ -62,6 +62,16 @@ def FindInitialVectors : Pass<"arc-find-initial-vectors", "mlir::ModuleOp"> {
   let summary = "Find initial groups of vectorizable ops";
   let constructor = "circt::arc::createFindInitialVectorsPass()";
   let dependentDialects = ["arc::ArcDialect"];
+  let statistics = [
+    Statistic<"numOfVectorizedOps", "vectorOps",
```
maerhart, line 66: Nit: the name `vectorOps` suggests the same as `numOfVectorsCreated`, i'd rename it slightly:
```suggestion
    Statistic<"numOfVectorizedOps", "vectorizedOps",
```

include/circt/Dialect/Arc/ArcPasses.td
```diff
@@ -62,6 +62,16 @@ def FindInitialVectors : Pass<"arc-find-initial-vectors", "mlir::ModuleOp"> {
...
+      "Total number of ops that were vectorized">,
+    Statistic<"numOfSavedOps", "numOfSavedOps",
+      "Total number of ops saved after FindInitialVectors pass">,
```
maerhart, line 69: Isn't this always equivalent to `vectorOps - numOfVectorsCreated` and could, therefore, be removed?

lib/Dialect/Arc/Transforms/FindInitialVectors.cpp
```diff
@@ -130,6 +131,11 @@ struct Vectorizer {
   TopologicalOrder order;
   Block *block;
 };
+
+size_t vecOps = 0;
+size_t savedOps = 0;
+size_t bigSeedVec = 0;
+size_t vecCreated = 0;
```
maerhart, line 138: Please don't use global variables; it's not thread-safe since multiple instances of `FindInitialVectorsPass` would access the same variables. You can make them members of the pass class instead (potentially creating a struct that groups them and thus makes it easier to pass them as function arguments).

maerhart (commented):
Really useful statistics, thanks! 🎉 
Added a few comments, but looks great otherwise!

include/circt/Dialect/Arc/ArcPasses.td
elhewaty:
```
a = b + c
d = e + f
g = h + i

after vectorization: [a, d, g] = [b, e, h] + [c, f, i]

biggestSeedVector: 3
numOfSavedOps: 2 (we had 3 ops but now we have one)
numOfVectorsCreated: 1
numOfVectorizedOps: 3
```

maerhart: Exactly, 3-1=2. I don't mind having it, just wanted to point out the redundancy.

maerhart (approved):
Thanks! Do you want me to merge it?

elhewaty:
> Exactly, 3-1=2. I don't mind having it, just wanted to point out the redundancy.

I just wanted to make the number clear, what do you think?

> Thanks! Do you want me to merge it?

If you think we don't need to remove `numOfSavedOps` please do.

PR #7116, opened 2024-06-03
[NFC, FIRRTL] Rename StrictConnect to MatchingConnect.  

darthscsi:
This is to reserve "Strict" for LHSType connects in the future.


prithayan (approved):
Looks good.

PR #7117, opened 2024-06-03
[FIRRTL] LHSType wrapper to indicate writable values. 

darthscsi:
This is to enable a split between writable identifiers and readable identifiers. This removed flow from the dialect, eventually.


prithayan (approved):
The `LHSType` make sense, can an `LHSType` `Value` be used in multiple strict connects, i.e., is there anything preventing multiple "writes" to an `LHSType` ?

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
```diff
@@ -417,6 +417,19 @@ def RefImpl : FIRRTLImplType<"Ref",
   }];
 }
 
+def LHSTypeImpl : FIRRTLImplType<"LHS", [], "::circt::firrtl::FIRRTLBaseType"> {
```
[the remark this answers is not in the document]
dtzSiFive: I think our type_isa checks if a thing statically derives from FIRRTLBaseType (`std::is_base_of`)? Might be worth a unit test to check this if this is carefully threading some details?

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
```diff
@@ -417,6 +417,19 @@ def RefImpl : FIRRTLImplType<"Ref",
...
+  let summary = "A wrapper for LHS types.";
+  let description = [{
+    A LHS type is a type usable for the destination of a strict connect
+    and for field indexing.  No other operations are valid.
+  }];
```
dtzSiFive, line 425: Probably these cannot be const (or probed, so on)?

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
```diff
@@ -417,6 +417,19 @@ def RefImpl : FIRRTLImplType<"Ref",
...
+  let parameters = (ins "FIRRTLBaseType":$type);
```
dtzSiFive, line 426: Should this be constrained to be passive, no analog?

Judging by direction re:strictconnect -- probably also only usable for known-width and so on?

EDIT: maybe that's what the modified SizedPassiveType is for?

LHS-of-const seems not-the-most-useful but maybe not inherently invalid (should be able to at least have LHS of a bundle that has const in it for writing to the non-const portions...?), WDYT?

include/circt/Dialect/FIRRTL/FIRRTLTypes.td
```diff
@@ -90,14 +90,13 @@ def SizedType : FIRRTLDialectType<CPred<"type_isa<FIRRTLBaseType>($_self) && "
     "a sized type (contains no uninferred widths)", "::circt::firrtl::FIRRTLType">;
 def SizedOrForeignType : AnyTypeOf<[SizedType, ForeignType]>;
 def SizedPassiveType : FIRRTLDialectType<And<[SizedType.predicate,PassiveType.predicate]>,
-    "a sized passive base type (contains no uninferred widths, or flips)", "::circt::firrtl::FIRRTLType">;
+    "a sized passive base type (contains no uninferred widths, or flips)", "::circt::firrtl::FIRRTLBaseType">;
```
dtzSiFive, line 93: Good change but appears unrelated.

lib/Dialect/FIRRTL/FIRRTLTypes.cpp
```diff
@@ -114,6 +114,11 @@ static LogicalResult customTypePrinter(Type type, AsmPrinter &os) {
           os << ", " << layer;
         os << '>';
       })
+      .Case<LHSType>([&](LHSType lhstype) {
+        os << "lhs<";
+        printNestedType(lhstype.getType(), os);
```
dtzSiFive, line 119: Please add printing/parsing test snippets for this?

As-is I think it's allowed in a wire, but probably that's a short-term thing (?).

Anyway if there's an IR snippet that uses these that would be great -- but adding it later sounds good if that doesn't really make sense until there are other bits.

dtzSiFive (commented):
Looks good, thanks for breaking this apart.

Haven't gone through the rest yet, but small question: We use wire's to plumb foreign and non-base types around -- properties, probes, so on.  Looks like these won't be convertible to strictwire since they require BaseType.  This seems reasonable re:strictwire, but means we may not be able to only have strictwire's until late in the pipeline (presently).  Just thinking this through -- any thoughts/plans here?

Left some feedback.

include/circt/Dialect/FIRRTL/FIRRTLTypes.td
```diff
@@ -110,6 +109,14 @@ def RWProbeConcreteReset : FIRRTLDialectType<
...
    "rwprobe type (with concrete resets only)", "::circt::firrtl::RefType">;
 
+def LHSType : FIRRTLDialectType<CPred<"isa<LHSType>($_self)">,
+                                      "writable type", "::circt::firrtl::LHSType">;
+
+class LHSTypeConstraint<string result, string input>
+  : TypesMatchWith<"Result must be lhs of input type",
+                   input, result,
+                   "type_cast<LHSType>($_self).getType()">;
+
```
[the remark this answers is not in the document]
darthscsi: They are in the ops and can I guess wait until that PR.

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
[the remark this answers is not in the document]
darthscsi: IIRC, to get sizing and other recursive type properties.

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
darthscsi: LHS-of-const seems a natural fallout of using connect to assign values to const.  Which isn't any different from now.

lib/Dialect/FIRRTL/FIRRTLTypes.cpp
darthscsi: Yea, wire needs to handle its type with it's constraints/verifier.  I've narrowed down the places which can go wrong by moving LHSType to FIRRTLType from FIRRTLBaseType.  Doesn't fix all issues, but does many.

darthscsi:
> The `LHSType` make sense, can an `LHSType` `Value` be used in multiple strict connects, i.e., is there anything preventing multiple "writes" to an `LHSType` ?

No, it certainly isn't the concern of the type to enforce this property.  But I also think that property is a complete artifact of the inner details of SFC implementation and doesn't actually matter.  There is the behavioral requirements specifying how values are set, but that isn't the same thing as internal representation.

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
darthscsi: I don't think known-width adds anything.

dtzSiFive: Oh, not sure what I was thinking.  So the source/destination are same type /now/ but may not be in the future -- so we'll insert the pads or whatever is needed when, e.g., doing width inference.  I think that's what we do already for matchingconnect?

darthscsi: yes, matching/strict connect have type identical constraints, including pre-inference.  a z/sext : known -> unknown captures the extension semantics, not connect.

darthscsi:
> Haven't gone through the rest yet, but small question: We use wire's to plumb foreign and non-base types around -- properties, probes, so on. Looks like these won't be convertible to strictwire since they require BaseType

I've been leaning towards foreign types being responsible for their own variable ops.  Ports need to support foreign types, but nothing in the strict.* does.  We can leave wires around and most passes just ignore them, but that isn't very satisfying.  Thus my thoughts that foreign dialects should have their own storage ops.

I do want to add a foreign by-val type wrapper which lets any foreign type which behaves as a simple typed collection of bits to usable in firrtl.  I think this would cover some of the space of opened type system in a more constrained way.  Likewise a connectable-only wrapper (which is I think more what the firesim types are, which is what the dialect was opened for).

PR #7120, opened 2024-06-03
 [FIRRTL] Add a new op interface for combinational loop detection.

prithayan:
Add a new `CombDataflow` op interface to `FIRRTL`. 
This interface is used for specifying the combinational dataflow that exists
 in the results and operands of an operation. Any operation that doesn't
 implement this interface is assumed to have a  combinational dependence
 from each operand to each result. 

Only `FIRRTL` register and memory ops implement this interface, but it can be used in other
dialects that intend to use the `CheckCombCycles` pass.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
@@ -330,8 +330,7 @@ def MemOp : HardwareDeclOp<"mem", [CombDataflow]> {
     // Extract the relevant attributes from the MemOp and return a FirMemory object.
     FirMemory getSummary();
 
-    SmallVector<std::pair<circt::FieldRef, circt::FieldRef>>
-      computeDataFlow ();
+    SmallVector<std::pair<circt::FieldRef, circt::FieldRef>> computeDataFlow();
   }];
 }
 
--- a/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
@@ -235,7 +235,7 @@ def InstanceChoiceOp : HardwareDeclOp<"instance_choice", [
 }
 
 
-def MemOp : HardwareDeclOp<"mem"> {
+def MemOp : HardwareDeclOp<"mem", [CombDataflow]> {
   let summary = "Define a new mem";
   let arguments =
     (ins ConfinedAttr<I32Attr, [IntMinValue<0>]>:$readLatency,
@@ -329,6 +329,9 @@ def MemOp : HardwareDeclOp<"mem"> {
 
     // Extract the relevant attributes from the MemOp and return a FirMemory object.
     FirMemory getSummary();
+
+    SmallVector<std::pair<circt::FieldRef, circt::FieldRef>>
+      computeDataFlow ();
   }];
 }
 
@@ -388,7 +391,7 @@ def NodeOp : HardwareDeclOp<"node", [
   let hasFolder = 1;
 }
 
-def RegOp : HardwareDeclOp<"reg", [Forceable, Sequential]> {
+def RegOp : HardwareDeclOp<"reg", [Forceable, CombDataflow]> {
   let summary = "Define a new register";
   let description = [{
     Declare a new register:
@@ -480,7 +483,7 @@ def RegOp : HardwareDeclOp<"reg", [Forceable, Sequential]> {
   let hasCanonicalizeMethod = true;
 }
 
-def RegResetOp : HardwareDeclOp<"regreset", [Forceable, Sequential]> {
+def RegResetOp : HardwareDeclOp<"regreset", [Forceable, CombDataflow]> {
   let summary = "Define a new register with a reset";
   let description = [{
     Declare a new register:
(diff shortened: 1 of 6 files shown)


include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
```diff
@@ -329,6 +329,9 @@ def MemOp : HardwareDeclOp<"mem"> {
 
     // Extract the relevant attributes from the MemOp and return a FirMemory object.
     FirMemory getSummary();
+
+    SmallVector<std::pair<circt::FieldRef, circt::FieldRef>>
+      computeDataFlow ();
```
uenoku, line 334: nit: format

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -113,6 +113,28 @@ bool firrtl::isDuplexValue(Value val) {
...
+  // exists.
+  if (getReadLatency() > 0)
+    return {};
+
+  SmallVector<std::pair<circt::FieldRef, circt::FieldRef>> deps;
+  // Add a dependency from the enable and address fields to the data field.
+  for (auto memPort : getResults())
+    if (auto type = type_dyn_cast<BundleType>(memPort.getType())) {
+      auto enableFieldId = type.getFieldID((unsigned)ReadPortSubfield::en);
+      auto addressFieldId = type.getFieldID((unsigned)ReadPortSubfield::addr);
```
uenoku, line 128: Could you use consistent cast style?

uenoku (commented):
Is this interface flexible enough to model dependency for probe?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
prithayan: Thanks, fixed the `static_cast`.

prithayan (commented):
I'm not sure how to handle Reftypes, since it cannot be determined locally, it requires a lot of state information like, "What the Ref type is pointing to ?", "What are the aliases for FieldRefs". I don't think this interface can handle refs.

uenoku (approved):
> I'm not sure how to handle Reftypes, since it cannot be determined locally, it requires a lot of state information like, "What the Ref type is pointing to ?", "What are the aliases for FieldRefs". I don't think this interface can handle refs.

That's fair. LGTM.

dtzSiFive:
The pairs of dataflow between FieldRef's are required to be leaf-most/ground types, is that right?  This would be good to note.

Having interface for this is great! I have some doubts about default being assuming the operation is indeed "combinatorial" (since that means mixing in operations we don't understand will cause this pass to /error out/ until those operations implement this interface to opt-out appropriately)-- that is to say, from a "safety" perspective it's conservatively more correct to over-record the dataflow (well, from operands -> results at least, it may have more complicated behavior re:side-effects beyond flowing to its results locally) through unknown operations but from a "don't abort the compilation if we really don't know" perspective it seems like maybe default should be to ignore unknown operations?  If Dialects wish to participate in /FIRRTL/'s notion of comb-loop-checking, they can opt-in.

Just some thoughts, probably this serves our needs for now without overdoing it.

Implementing it for our operations is a nice way to show this working, neat!  Looking at this a bit, it seems for registers this interface only tells part of the story -- looks like the pass still hardcodes knowledge of those operations (reg/regreset) -- can this be refactored? (If a new op wants to be register-like, can it use this interface to accomplish that?)

prithayan:
@dtzSiFive , thanks for your comments, 
> The pairs of dataflow between FieldRef's are required to be leaf-most/ground types, is that right? This would be good to note.
> 

Right, I will add a comment with this note.

> Having interface for this is great! I have some doubts about default being assuming the operation is indeed "combinatorial" (since that means mixing in operations we don't understand will cause this pass to /error out/ until those operations implement this interface to opt-out appropriately)-- that is to say, from a "safety" perspective it's conservatively more correct to over-record the dataflow (well, from operands -> results at least, it may have more complicated behavior re:side-effects beyond flowing to its results locally) through unknown operations but from a "don't abort the compilation if we really don't know" perspective it seems like maybe default should be to ignore unknown operations? If Dialects wish to participate in /FIRRTL/'s notion of comb-loop-checking, they can opt-in.
> 

Actually, here I took the shortcut, that is the assumption is that every operation, other than just the 3 in FIRRTL, have the default behavior. So, you should need to annotate only the very few special ops. That's why initially I was suggesting a `Sequential` interface, that denotes just the ops that should be excluded from comb loop checks.
I agree with you here, any non-FIRRTL dialect that wishes to participate has to explicitly opt-in, again the assumption is that very few ops will have a special combinational dataflow, so fewer updates are required. It is the responsibility of the client to ensure special ops are skipped and false loops are not reported.

> Just some thoughts, probably this serves our needs for now without overdoing it.
> 
> Implementing it for our operations is a nice way to show this working, neat! Looking at this a bit, it seems for registers this interface only tells part of the story -- looks like the pass still hardcodes knowledge of those operations (reg/regreset) -- can this be refactored? (If a new op wants to be register-like, can it use this interface to accomplish that?)

That's a good idea, so you mean a new `RegisterLike` op interface  ?

PR #7122, opened 2024-06-04
[ImportVerilog][MooreToCore]Lower moore.namedConstant to hw.param.value

mingzheTerapines:
Refer to https://github.com/llvm/circt/pull/7083
For [ImportVerilog] this PR changed value to attribute for builder of `moore.named_constant`. Because for lower pass case it is hard to transform an value to an attribute.
For [MooreToCore] this PR learned how `moore.constant`  and `hw.param.value` build and have this lowering method.
String attribute will be supported later.


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -74,6 +74,24 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
+struct NamedConstantOpConv : public OpConversionPattern<NamedConstantOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(NamedConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // make it as global if it is a parameter kind
+    if (op.getKind() == moore::NamedConst::Parameter) {
+      rewriter.setInsertionPointToStart(
+          op->getParentOfType<mlir::ModuleOp>()->getBlock());
```
uenoku, line 86: It's not legal to refer values in the global scope from IsolatedAbove op. I think you can just create the op in the local scope.

mingzheTerapines: Thanks a lot. I  just met a crash when iserting.

mingzheTerapines:
@fabianschuiki 
For case like this, int_t d will be generated for several times.
```
module top();
    genvar i;
    parameter a=2;
     typedef struct{
  int   INT_VAL;}int_t;

    generate
            for(i=0;i<a;i=i+1) begin
            int_t d;
        end
    endgenerate
endmodule
```
I will also add SSA-related trait and verify for union and struct in this PR.

mingzheTerapines:
@fabianschuiki  Still having some problem when comparing moore to llhd. 
%g = hw.param.value i42 = **#hw.param.expr.shrs<4, #hw.param.decl.ref<"xx">>**
value-like attribute may be needed. Do you think it is proper?
Now I can translte it into %p1 = moore.named_constant parameter : !moore.l32 = 1
which 1 is not the whole initializer(value)

mingzheTerapines:
Thanks @cepheus69 , I need to check attribute getvalue that I used.

mingzheTerapines:
@fabianschuiki @cepheus69  Sorry but still having literal problem while doing mooretocore literal check.
```
 circt-opt alc.mlir --convert-moore-to-core
alc.mlir:12:37: error: expected non-function type
%p1 = moore.named_constant parameter  1 : !moore.l32
```
image
I tried to let it do nothing while conversion. But still got this error. It seems this literal check start right before the conversion.
```
struct NamedConstantOpConv : public OpConversionPattern<NamedConstantOp> {
  using OpConversionPattern::OpConversionPattern;

  LogicalResult
  matchAndRewrite(NamedConstantOp op, OpAdaptor adaptor,
    // Type resultType = typeConverter->convertType(op.getResult().getType());
    // resultType.dump();
    //  rewriter.replaceOpWithNewOp<hw::ParamValueOp>(
    //      op, typeConverter->convertType(op.getResult().getType()),
    //      adaptor.getValue());
    //rewriter.eraseOp(op);
    return success();
  }
};
```
What's weired is constant op could be parsed.
image

mingzheTerapines:
> @fabianschuiki @cepheus69 Sorry but still having literal problem while doing mooretocore literal check.
> 
> ```
>  circt-opt alc.mlir --convert-moore-to-core
> alc.mlir:12:37: error: expected non-function type
> %p1 = moore.named_constant parameter  1 : !moore.l32
> ```
> 
> image I tried to let it do nothing while conversion. But still got this error. It seems this literal check start right before the conversion.
> 
> ```
> struct NamedConstantOpConv : public OpConversionPattern<NamedConstantOp> {
>   using OpConversionPattern::OpConversionPattern;
> 
>   LogicalResult
>   matchAndRewrite(NamedConstantOp op, OpAdaptor adaptor,
>                   ConversionPatternRewriter &rewriter) const override {
>     // Type resultType = typeConverter->convertType(op.getResult().getType());
>     // resultType.dump();
>     //  rewriter.replaceOpWithNewOp<hw::ParamValueOp>(
>     //      op, typeConverter->convertType(op.getResult().getType()),
>     //      adaptor.getValue());
>     //rewriter.eraseOp(op);
>     return success();
>   }
> };
> ```
> 
> What's weired is constant op could be parsed. image

Thanks @hailongSun2000 ,it is a problem of function parsing, I will look at it.

mingzheTerapines:
@fabianmcg  Still having problem with parsing. Because value.type is i32 but the result type is !moore.l32. They are of two different type which will fail when parsing.
```
static ParseResult parseParamValue(OpAsmParser &p, Attribute &value,
                                   Type &resultType) {
  if (p.parseAttribute(value, resultType) || p.parseColon() ||
      p.parseType(resultType))
    return failure();
  return success();
}
```

mingzheTerapines:
Because following varifying, mooretocore is stopped. When blocking this code will success.
```
LogicalResult ParamValueOp::verify() {
  // Check that the attribute expression is valid in this module.
  return checkParameterInContext(
      getValue(), (*this)->getParentOfType<hw::HWModuleOp>(), *this);
}
```

mingzheTerapines:
@fabianschuiki Hailong and I were wondering why hw.param.value has traits constant like and pure.
Also I found name information missed when lowering to hw.param.value, is that acceptable?
Test failed because I changed param.value varifying the parent module type.

fabianschuiki (commented):
Sorry for taking so long to look at this!

I think the `hw.param.*` are designed to work with parameters defined on the `hw.module`. These would then be emitted in Verilog as `module Foo #(<params>) ...`. If I'm not mistaken, `hw.param.value` is used to take one of those module parameters (which is an attribute) and make it available as an SSA value. So this might not do what you need :thinking:.

Have you tried lowering the named constants to an `hw.wire`? If you give the wire an inner symbol name, it will not be deleted. For example:
```
%p1 = moore.named_constant parameter 1 : !moore.i32
// lower to
%0 = hw.constant 1 : i32
%p1 = hw.wire %0 sym @someUniqueSymbolName : i32
```

mingzheTerapines:
> Sorry for taking so long to look at this!
> 
> I think the `hw.param.*` are designed to work with parameters defined on the `hw.module`. These would then be emitted in Verilog as `module Foo #(<params>) ...`. If I'm not mistaken, `hw.param.value` is used to take one of those module parameters (which is an attribute) and make it available as an SSA value. So this might not do what you need 🤔.
> 
> Have you tried lowering the named constants to an `hw.wire`? If you give the wire an inner symbol name, it will not be deleted. For example:
> 
> ```
> %p1 = moore.named_constant parameter 1 : !moore.i32
> // lower to
> %0 = hw.constant 1 : i32
> %p1 = hw.wire %0 sym @someUniqueSymbolName : i32
> ```

Thanks for your advice. In specialize pass in hw, hw.param.value with integer will be automatically replaced hw.constant. So it is the same for integer case. What's more, parameter in IEEE standard can accept more than integer attribute, so I think translate sv.parameter to hw.param.value is Okay. But other attribute type case this time will work around. Because we focus on basic.
And as your concern of hw.module, new patch by @cepheus69  will solve this problem to ensure hw.param.value in every hw.moduleop region.

mingzheTerapines:
@fabianschuiki  Thanks @cepheus69 , module to core has been available now. 
And I checked json below string will be implictly transformed to an integer. Any attribute could be handled in namedconstant.
```
{
              "name": "p",
              "kind": "Parameter",
              "addr": 2797836494528,
              "type": "logic[23:0]",
              "initializer": {
                "kind": "Conversion",
                "type": "logic[23:0]",
                "operand": {
                  "kind": "StringLiteral",
                  "type": "bit[23:0]",
                  "literal": "str",
                  "constant": "24'd7566450"
                },
                "constant": "24'd7566450"
              },
              "value": "24'd7566450",
              "isLocal": false,
              "isPort": false,
              "isBody": true
            }
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -345,23 +345,26 @@ def NamedConstAttr : I32EnumAttr<"NamedConst", "elaboration-time constants",
...
     See IEEE 1800-2017 § 6.20 "Constants".
   }];
-  let arguments = (ins StrAttr:$name, 
-                       NamedConstAttr:$kind, 
-                       UnpackedType:$value);
-  let results = (outs UnpackedType:$result);
+  let arguments = (ins StrAttr:$name,
+                       NamedConstAttr:$kind,
+                       AnyAttr:$value);
+  let results = (outs IntType:$result);
```
fabianschuiki, line 362: Since you know that the `named_constant` will always return an integer, `$value` could be an `IntegerAttr` as well :smiley:

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -155,6 +155,44 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
+      symStr = "specparameter";
+      break;
+    }
+    symStr += ":";
+    symStr += op.getNameAttr().str();
+    auto symAttr = rewriter.getStringAttr(symStr);
+    if (isa<IntegerAttr>(value)) {
+      auto valueInt = cast<IntegerAttr>(value);
+      auto paramOp =
+          rewriter.create<hw::ParamValueOp>(op->getLoc(), resultType, valueInt);
```
fabianschuiki, line 185: I would make this an `hw.constant`, since you're basically using the `ParamValueOp` to materialize a constant attribute.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -329,6 +329,49 @@ void ConstantOp::build(OpBuilder &builder, OperationState &result, IntType type,
...
+             << typeInt;
+
+    valueInt = valueInt.trunc(typeInt.getWidth());
+  }
+
+  auto attrType = IntegerType::get(p.getContext(), typeInt.getWidth());
+  value = IntegerAttr::get(attrType, valueInt);
+  resultType = typeInt;
+  return success();
+}
```
fabianschuiki, line 366: I'm wondering if we really want to duplicate this code from `moore.constant`. I have a feeling that things will be more maintainable if we just make `named_constant` accept an SSA value, and if we really need that to be a constant, to add a verifier that checks whether that SSA value's defining op is a `moore.constant` op. That will prevent quite a bit of duplication, and only `moore.constant` would have to know how to map a constant to the HW dialect.

Especially since we currently don't _need_ the `named_constant` to be a constant value per se: when lowering to HW, you're mapping it to an `hw.wire`, which is great! That can deal even with non-constant values for now, so it might make sense to just not bother checking whether the value is constant at all.

hailongSun2000: There is an example:
```
module Foo;
  parameter p1 = 32;
  parameter p2 = p1;
endmdule
```
lower to moore IR
```
moore.module (){
  %0 = moore.constant 32 : i32
  %p1 = moore.named_constant %0 : i32
  %p2 = moore.named_constant %p1 : i32
}
```
then lower to core IR
```
hw.module (){
  %c32_i32 = hw.constant 32 : i32
  %p1 = hw.wire %c32_i32 : i32
  %p2 = hw.wire %p1 : i32 
}
```
I think the above is what @fabianschuiki means.
We needn't lower `moore.named_constant` into `hw.para.value`, and also don't need `parseParamValue` that looks like parse `moore.constant`. And this lowering route can only lower `moore.constant` into `hw.constant` one-to-one.

mingzheTerapines: Thanks @fabianschuiki  @hailongSun2000 , I got it and updated my code. plz have a look =)

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -155,6 +155,36 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
+    case NamedConst::LocalParameter:
+      symStr = "localparameter";
+      break;
+    case NamedConst::SpecParameter:
+      symStr = "specparameter";
+      break;
+    }
+    symStr += ":";
+    symStr += op.getNameAttr().str();
+    auto symAttr = rewriter.getStringAttr(symStr);
```
hailongSun2000, line 180: nit: Could there use Twine()(lightweight) and llvm::StringRef() to connect string literal? Like
` auto symAttr = rewriter.getStringAttr(symStr + Twine(":") + adaptor.getName());`

mingzheTerapines: Sorry I will be careful.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -350,22 +350,23 @@ def NamedConstAttr : I32EnumAttr<"NamedConst", "elaboration-time constants",
 
 def NamedConstantOp : MooreOp<"named_constant", [
   SameOperandsAndResultType,
-  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>
+  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>,
```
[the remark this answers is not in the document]
mingzheTerapines: Sorry I will be careful

hailongSun2000 (approved):

fabianschuiki (approved):
Really nice!

Issue #7023, opened 2024-05-10
[HW] HWModule (et. al.) should have separate symbol and name to be emitted


teqdruid:
Given all the talk about the mistake of conflating symbols with names, I had assumed that this had been fixed in the HW dialect. Is there some reason it hasn't? Or is this just waiting on someone to do it?

teqdruid:
@darthscsi @seldridge Since you two were the two recommending to Morten and myself about names vs. symbols.

teqdruid:
Discussed this at the ODM. HWModuleOp should have a separate module name which gets unique'd instead of using the symbol by default.

Separately, everything should be using the `hw.verilogName` attribute to tell ExportVerilog "I know what I'm doing, use this name"

PR #7123, opened 2024-06-04
[Ibis] Divorce symbol and actual names in class and container ops

mortbopet:
... while factoring out `InnerSymbol`+name logic into a new interface that is also used by Ibis port-like ops. This model could probably be factored into the `hw` dialect, seeing as we're also eventually going to divorce symbol names and actual names in `hw.module`.

Also adds de-aliasing logic to `ContainersToHW`, which is required given that `hw.module` doesn't yet split its symbol and name (#7023).


blakep-msft (approved):
LGTM, thanks for doing this.

include/circt/Dialect/Ibis/IbisInterfaces.td
```diff
@@ -18,7 +18,38 @@ include "circt/Dialect/HW/HWOpInterfaces.td"
...
+  let cppNamespace = "circt::ibis";
+  let description = [{
+    An interface for operations which define inner symbols with an optional
+    namehint.
+  }];
+
+  let methods = [
+    InterfaceMethod<
+      "Returns the namehint attribute of this op",
+      "mlir::StringAttr", "getNameHintAttr",
```
teqdruid, line 31: Let's not call this a 'namehint'. Too easily confused with the sv.namehint attribute. Is there any reason why you don't just call it 'name'?

teqdruid (approved):

include/circt/Dialect/Ibis/IbisInterfaces.td
mortbopet: There's a crazy amount of aliasing going on here in the `name`-related namespace. if one adds an attribute in ODS as `name`, that'll define `getName` functions, which the interface would then alias with... it feels like `getName` is almost _too_ generic to be a valid interface function name. Maybe `getSymbolName`? - ideas are welcome.

Issue #6523, opened 2023-12-13
[Comb] Canonicalization blocking is too pessimistive?


uenoku:
```
  hw.module @Foo(in %a : i1) {
    %true = hw.constant true
    %r = sv.reg : !hw.inout<i1> 
    sv.initial {
      %0 = comb.and %a, %true : i1
      sv.passign %r, %0 : i1
    }
    hw.output
  }
```

We expect `%0` is canonicalized into `%a` but https://github.com/llvm/circt/pull/6235 blocks this kind of optimization.

mortbopet:
This is indeed somewhat conservative, and also highlights a potential issue between canonicalization barriers and constant hoisting... #6235 was known to be a conservative approach.
Personally, I think the next step is to implement an optional interface/trait for non-isolated-from-above region-defining operations (like `sv.initial`), e.g. `PreservesCombSemantics`, which the comb canonicalizers can inspect and thus allowing operations to opt-in to cross-region/block optimization.

uenoku:
Fixed by #8607

PR #6235, opened 2023-10-03
[Comb] Disallow canonicalization across MLIR blocks

mortbopet:
This is probably the most conservative implementation of this, but should suffice for the current requirements to this change. For any comb operation that has a canonicalizer, guard canonicalization on whether any operands are define outside of the current block. This still allows for (constant) folding, which (to me) still should be a safe thing to do across blocks - if that was not the case, then that would be a strange abstraction (i'd expect an op to be `IsolatedFromAbove` if it wants to prevent constant folding).

PR #7124, opened 2024-06-04
[CombFolds] Relax block optimization barrier by pulling constants

fzi-hielscher:
This is an attempt at reducing the conservativeness of the cross-block optimization barrier introduced in #6235. See also #6523.

If canonicalization of an operation is prevented due to an operand defined outside of its block, check if all of these operands are `ConstantLike`. If so, create a copy of them inside of the operation's block, in the hope that this enables optimization in the next iteration.
I do not love this approach, as it causes movements/duplications even if doesn't have any benefit. But I feel this is the lesser evil compared to leaving chunks of dead code float around because a trivial fold like `a & false -> false`  is being prevented.

Is this relaxation acceptable for your use case, @mortbopet @teqdruid ? Am I missing a more elegant solution?

### DIFF
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -36,6 +36,56 @@ static bool hasOperandsOutsideOfBlock(Operation *op) {
   });
 }
 
+/// If _all_ of the operation's operands, which are defined outside of the
+/// operation's own block, are ConstantLike, rematerialize them inside of the
+/// operation's block. This relaxes the conservativeness incurred by the
+/// 'hasOperandsOutsideOfBlock' checks.
+static LogicalResult
+materializeConstantOperandsLocally(Operation *op, PatternRewriter &rewriter) {
+  SmallVector<Value> newOperands;
+  newOperands.reserve(op->getNumOperands());
+  bool doOpRewrite = false;
+
+  for (auto operand : op->getOperands()) {
+    auto defOp = operand.getDefiningOp();
+
+    if (!defOp || defOp->getBlock() == op->getBlock()) {
+      // Keep local operands
+      newOperands.push_back(operand);
+      continue;
+    }
+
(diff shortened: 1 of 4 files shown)


lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -36,6 +36,54 @@ static bool hasOperandsOutsideOfBlock(Operation *op) {
...
+
+    if (!defOp || defOp->getBlock() == op->getBlock()) {
+      // Keep local operands
+      newOperands.push_back(operand);
+      continue;
+    }
+
+    if (!defOp->hasTrait<OpTrait::ConstantLike>())
+      // Optimization will remain blocked, so don't bother changing the op.
+      return failure();
```
dtzSiFive, line 60: This looks like it can make changes and then return failure, which should not happen.

fzi-hielscher: Ah, for some reason I assumed the rewriter would take care of rolling back the changes if the pattern fails...

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -36,6 +36,69 @@ static bool hasOperandsOutsideOfBlock(Operation *op) {
...
+
+    for (unsigned resultIdx = 0; resultIdx < defOp->getNumResults();
```
uenoku, line 64: I don't know if there is a multi-results ConstantLike operation. I think it's reasonable to restrict this to single result.

fzi-hielscher:
Thank you for your comments.
After giving some more thought to this I actually don't like this approach at all anymore. 😅
Firstly, there is the problem @mortbopet mentioned over in #6523: If the constant does not fold after moving, the hoisting and pulling get into a tug of war, which repeatedly causes pointless iterations.
Secondly, `a & false -> false` still won't fold if `a` is non-local. So, it kind of misses the point.
Let's try harder...

fzi-hielscher:
Fortunately made obsolete by #8611 .

PR #7126, opened 2024-06-04
[FIRRTL][DropConst] Fix performance with many extmodule's.

dtzSiFive:
Before, DropConst on FIRRTL w/300k extmodule's took 221s. Now it's < 1s.


prithayan (approved):
Nice, good to know that skipping over ext modules can avoid significant redundant work.

dtzSiFive:
> Nice, good to know that skipping over ext modules can avoid significant redundant work.

Apparently, yep!

Same idea basically as #6850/#6851, FWIW.  getNumRegions vs filtering on FModuleOp specifically.

dtzSiFive:
Thanks for the review!

Issue #7125, opened 2024-06-04
[FIRRTL] LowerClasses doesn't actually copy the leading part of the path from the owning module to the start of the NLA


mikeurbach:
We document that this is done in a comment: https://github.com/llvm/circt/blob/b39f9ecd9241a7aee15d043c31315fc77884c8da/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp#L499-L500

However, nothing is actually modifying `path` to include the intermediate hierarchies between the owning module and the start of the NLA.

I tried a small change, and I think this is easy to fix, but this will need to work with how we now cache the computation of `getOrComputeAltBasePath`. Basically, in the `while(true)` loop, all we need to do is append `InnerRefAttr`s for each instance we visit along the way up the instance graph from `moduleName` to the owning module. Or, we can do something similar that achieves the same effect.

I think it would probably be good to do this during the traversal in `getOrComputeAltBasePath`, but then we need to add the array of `InnerRefAttr`s to the cached state.

mikeurbach:
Note that after https://github.com/llvm/circt/pull/7129, "each instance we visit along the way" may actually be multiple instances of the same module. In that case, just using `usesBegin()` after emitting the warning produces the same results as EmitOMIR, so I think that is a fine short-term solution to get something working. I have confirmed this on a local patch, but that was before we refactored LowerClasses recently. Once we finally update the user code, and turn this back to an error (https://github.com/llvm/circt/issues/7128), picking `usesBegin()` is no problem at all.

Issue #7128, opened 2024-06-04
[FIRRTL] LowerClasses should enforce single instantiation


mikeurbach:
This was removed in https://github.com/llvm/circt/pull/7129, because this requirement never existed in EmitOMIR, and in real designs coming from Chisel
today, we are not yet able to enforce single instantiation. This is fine in ResolvePaths, it is sensible and correct to simply convey the local or hierarchical path the user requested. But in LowerClasses, we have to build up the path prefix from the owning module to the start of the hierpath op created by ResolvePaths.

Note that this is currently not being done at all, which is tracked in https://github.com/llvm/circt/issues/7125. When we do implement this, in the case of multiple instances, we will just have to pick one. Ideally, Chisel users wouldn't generate such paths to begin with, and we could enforce single instantiation. For now, just picking the first one in the case of multiple instances will produce the same output as EmitOMIR would.

PR #7129, opened 2024-06-04
[FIRRTL] Simplify path handling in ResolvePaths and LowerClasses.

mikeurbach:
This turns the multiple-instantiation error into a warning in ResolvePaths and LowerClasses. In real designs coming from Chisel today, we are not yet able to enforce single instantiation. This was never a requirement of the original way that we handled hierarchical paths in EmitOMIR, so this removes the requirement for now. Adding it back is tracked in https://github.com/llvm/circt/issues/7128.

With this change, the ResolvePaths logic was simplified to stop trying to disambiguate paths in some cases, and instead allow the annotations to simply convey the user's requested local or hierarchical path. In LowerClasses, if there are multiple instances, this means we have ambiguity. In practice, this logic will produce the same outputs as EmitOMIR, once we fix https://github.com/llvm/circt/issues/7125.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -377,14 +377,16 @@ PathTracker::getOrComputeNeedsAltBasePath(Location loc, StringAttr moduleName,
       break;
     }
     // If there is more than one instance of this module, then the path
-    // operation is ambiguous, which is an error.
+    // operation is ambiguous, which is a warning.  This should become an error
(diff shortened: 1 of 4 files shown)


prithayan (approved):
Looks good, If I understand this correctly, seems like we will never hit the ambiguous path case in the current designs.

mikeurbach:
We can have ambiguous paths, but it doesn't affect anything until we fix #7125. But in that change (https://github.com/llvm/circt/pull/7130) it's at least possible to handle it in a way that produces the same output in the properties flow as EmitOMIR does.

PR #7130, opened 2024-06-04
[FIRRTL] Actually copy the leading part of the path in LowerClasses.

mikeurbach:
We stated in a comment that we copy the leading part of the hierarchical path from the owning module to the start of the annotation's NLA. But we never actually did that, so this adds (back) that logic.

Fixes https://github.com/llvm/circt/issues/7125.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -377,14 +377,16 @@ PathTracker::getOrComputeNeedsAltBasePath(Location loc, StringAttr moduleName,
       break;
     }
     // If there is more than one instance of this module, then the path
-    // operation is ambiguous, which is an error.
+    // operation is ambiguous, which is a warning.  This should become an error
+    // once user code is properly enforcing single instantiation, but in
+    // practice this generates the same outputs as the original flow for now.
+    // See https://github.com/llvm/circt/issues/7128.
     if (!node->hasOneUse()) {
-      auto diag = mlir::emitError(loc)
+      auto diag = mlir::emitWarning(loc)
                   << "unable to uniquely resolve target due "
                      "to multiple instantiation";
       for (auto *use : node->uses())
         diag.attachNote(use->getInstance().getLoc()) << "instance here";
-      return diag;
     }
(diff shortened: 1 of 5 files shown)


mikeurbach:
This is orthogonal to, but stacked on https://github.com/llvm/circt/pull/7129.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -496,10 +499,29 @@ PathTracker::processPathTrackers(const AnnoTarget &target) {
...
+          break;
+
+        // Append the next level of hierarchy to the path. Note that if there
+        // are multiple instances, which we warn about, this is where the
+        // ambiguity manifests. In practice, just picking usesBegin generates
+        // the same output as EmitOMIR would for now.
+        InstanceRecord *inst = *node->usesBegin();
+        path.push_back(
+            OpAnnoTarget(inst->getInstance<InstanceOp>())
+                .getNLAReference(namespaces[inst->getParent()->getModule()]));
```
mikeurbach, line 518: We need to use the inner symbol namespace here from the current instances parent, so we can't use `moduleNamespace`, since that's not the right namespace. This is why I added the full inner symbol namespace collection as an instance variable. I think this should be safe, but wanted to call it out.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -496,10 +499,29 @@ PathTracker::processPathTrackers(const AnnoTarget &target) {
       } else {
         moduleName = cast<hw::InnerRefAttr>(oldPath.front()).getModule();
       }
+
+      // Copy the leading part of the hierarchical path from the owning module
+      // to the start of the annotation's NLA.
+      InstanceGraphNode *node = instanceGraph.lookup(moduleName);
+      while (true) {
```
mikeurbach, line 506: In this patch, I have just added a fresh instance graph walk up for each hierpath op. We could coordinate this with the walk in getOrComputeAltBasePath, but then we'd need to update the state to cache the array of paths as well. In large designs, I didn't observe any noticeable performance regression from this.

PR #7131, opened 2024-06-04
[Docs][LTL] Add SVA Encodings to LTL rationale

dobios:
This PR updates the LTL rationale to include all of the encodings I could think of for SVA in various CIRCT dialects (mostly LTL).   

Let me know what you think!


docs/Dialects/LTL.md
```diff
@@ -167,8 +174,240 @@ property p1; eventually p0; endproperty
...
+```mlir
+%zero = hw.constant 0 : i1
+%true = hw.constant 1 : i1
+%1 = seq.shitreg n, %a, %clk, %true, powerOn %zero : i1
+``` 
+
+> The following functions should be encoded in the front-ends directly:  
+> - **`$onehot(a)`**: `a.reduce(_+_) == 1`
+> - **`$onehot0(a)`**: `a.reduce(_+_) <= 1`
+> - **`$isunknown(a)`**: `a == X || a == Z`
```
sequencer, line 217: Frontend(chisel) cannot represent Z/X? Maybe we need to add an intrinsic here?

dobios: Maybe nut for now they aren't supported, I guess I'll just write that then

dobios: Maybe X/Z should have clearer support in the front-ends

sequencer: Agree. That might be a non trivial work we should think carefully.

dobios: Yes I'm afraid that this kind of approach could bloat the intrinsics available in chisel. I would rather use easily identifiable patterns that we can then identify in exportVerilog to emit these functions only in SV and not have the IR behave as an AST for SVA.

uenoku: FWIW there is an intrinsic for X value. https://github.com/llvm/circt/blob/c76eb120d675fc093ca114d288ca25548be85e7a/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td#L175-L189

fabianschuiki (approved):
This is fantastic stuff! Thank you very much for putting in all the effort to write up such a nice piece of documentation :partying_face:

liuyic00:
Some questions I have about this and previous PR adding operators in the dialect:

1. `ltl.intersect` also works with properties. Are we considering that the properties somehow match a trace, but not hold at some point of time?  
2. The strong version of operators are "not really distinguishable from the weak version in CIRCT". I‘m wondering why. For example, `p1 s_until p2` required `p2` mast hold at some time in the limited simulation trace. But the week version does not required this.  
3. `nexttime p` and `always p` are mapped to the operators that do not accept a property as the parameter now in the doc. I'm not sure if this is a mistake or if we are going to extend these operators in the future.

`not s` is not a part of SVA. `[->n:m]` and `[=n:m]` only work with a boolean expression, not a sequence in SVA. However, they are reasonable to include in this dialect.

liuyic00:
> `ltl.intersect %a, %b` is basically equivalent to `(a or b) implies (a and b)` so if one of the two fires, both should fire or this does not hold, so this works with all types of properties.

What confuses me is the description of `ltl.intersect` in the [code](https://github.com/llvm/circt/blob/53abd04cace31412a740668eaaa1ad305ea63097/include/circt/Dialect/LTL/LTLOps.td#L54). The description seems more like the defination of `intersect` in SVA, but it is not equivalent to `(a or b) implies (a and b)`. "start and end times" are the concepts for sequnces, not for properties.

> There was a lot of discussion around this, but basically the idea is that weak or strong properties only make sense in certain contexts. If you are asserting or assuming a property, they will be weak, if you are covering a property it will be strong. Looking though use cases, we couldn't find enough to justify adding the capacity to specify weak properties for cover and strong for assert and assume, as this would require having some notion of simulation / formal bounds, which didn't seem worth the complication given the limited usefulness.

Dose this mean that in `%0 = ltl.delay %p, 1, 0; %1 = ltl.not %0`, whether the property in the assertion is `%1` or `%0`, they are both week properties? It dosen't like `not` will filp the week/strong in SVA? 

Thanks for your answer!

PR #7136, opened 2024-06-05
[FIRRTL] Support abstract reset in RWProbeOp

dtzSiFive:
Simplification follows as no longer need forceable.

PR #6258, opened 2023-10-05
[FIRRTL][Parser] Improve rwprobe parsing, support inst results.

dtzSiFive:
Parse explicitly for rwprobe case for better handling and to avoid creating spurious indexing operations.

Instance results are supported by inserting a forceable bounce wire.  The wire replaces the instance port entirely (including its unbundled entry) locally, and is connected to/from the instance result.

Wires are created for entire result (port) only, and will be re-used by repeated rwprobe's to the result in part or whole. This means the entire port must be forceable and will go through the bounce wire.

Annotations will continue to resolve as they do today, not to the inserted bounce wire.

Indexing parsing based on code from #5809, thanks @youngar!

PR #5835, opened 2023-08-14
[FIRRTL][FIRParser] Prefer RWProbe op as much as possible.

dtzSiFive:
Keep forceable for now-- it's needed for uninferred resets presently, and has slightly better (if not ideal) behavior when attempting to rwprobe a non-passive aggregate.

Note: force_nonpassive.fir doesn't work with symbols, check.

We can't rwprobe() an aggregate that /must/ be lowered.  Forceable tries to allow this, but when using symbols this is just rejected.

Probably this should not be allowed.

PR #7137, opened 2024-06-05
[FIRRTL][Parser] Never use forceable

dtzSiFive:
Builds on #7136.


dtzSiFive:
Mostly stopped with #5835, and #6258.

dtzSiFive:
Most important part of this change is not creating every single declaration as "forceable" and rewriting them all to not be this way, in the parser.

As for not parsing into forceable, this is already the case with edge case of uninferred reset.  This portion is changed to use `rwprobe` op + inner symbol like everything else here: https://github.com/llvm/circt/pull/7137/commits/ba40b7edb674617f3b63a231509d7055c6ea634b .

dtzSiFive:
Post-merge review welcome!

PR #7140, opened 2024-06-05
Capitalize Mode in github workflow.  

darthscsi:
This might fix a new breakage on the gcc build complaining about DCMAKE_BUILD_TYPE

Should fix:
```
CMake Error at CMakeLists.txt:88 (message):
  

  No build type selected.  You need to pass -DCMAKE_BUILD_TYPE=<type> in
  order to configure LLVM.

  Available options are:

    * -DCMAKE_BUILD_TYPE=Release - For an optimized build with no assertions or debug info.
    * -DCMAKE_BUILD_TYPE=Debug - For an unoptimized build with assertions and debug info.
    * -DCMAKE_BUILD_TYPE=RelWithDebInfo - For an optimized build with no assertions but with debug info.
    * -DCMAKE_BUILD_TYPE=MinSizeRel - For a build optimized for size instead of speed.

```

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -145,12 +145,12 @@ jobs:
           # gcc   + debug   + assert   + shared
           - cc: clang
             cxx: clang++
-            mode: debug
+            mode: Debug
             assert: ON
             shared: ON
           - cc: gcc
             cxx: g++
-            mode: release
+            mode: Release
             assert: OFF
             shared: OFF
 
--- a/.github/workflows/uploadReleaseArtifacts.yml
+++ b/.github/workflows/uploadReleaseArtifacts.yml
@@ -29,10 +29,10 @@ on:
         required: true
         type: choice
         options:
-          - release
-          - relwithdebinfo
-          - debug
-        default: release
+          - Release
+          - RelWithDebInfo
+          - Debug
+        default: Release
 
   # Run every day at 0700 UTC which is:
   #   - 0000 PDT / 2300 PST


seldridge (approved):
I see no problem with changing this. I didn't think this made a difference, though. 🧐

dtzSiFive (approved):
LGTM!

I'm curious and will go chase down what this is about since not sure what the thinking is (I thought they were force-capitalized or whichever -- canonicalized?), but that does seem to break now and looks like a good change anyway :+1:.

fabianschuiki (approved):
LGTM!

dtzSiFive:
Build failure says:

>  No build type selected. 

And LLVM CMake suggests that means it thinks we didn't set one, not that we set an unknown one?

https://github.com/llvm/llvm-project/blob/a9342f2eab7993a3be1d1156070c0d2de21ed973/llvm/CMakeLists.txt#L87-L97

dtzSiFive:
Earlier it says:
```
-- Configuring done
You have changed variables that require your cache to be deleted.
Configure will be re-run and you may have to reset some variables.
The following variables have changed:
CMAKE_C_COMPILER= gcc
CMAKE_CXX_COMPILER= g++
CMAKE_CXX_COMPILER= g++
```

Anyway this seems like a good change anyway! But have yet to sort this out, and may be something else weird? (not sure!)

darthscsi:
Also, why are we using gcc 9?

darthscsi:
> And LLVM CMake suggests that means it thinks we didn't set one, not that we set an unknown one?

yes, but it seems we did from the yml.  I wish we could see the cmdline.

PR #7141, opened 2024-06-06
[FIRRTL][InferResets] Learn how to trace through nodes.

dtzSiFive:
Trace reset networks through nodes.

This seems like it was always reasonable to do but is now much more reachable with wire-to-node conversion happening.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
@@ -767,6 +767,8 @@ void InferResetsPass::traceResets(CircuitOp circuit) {
                         op.getLoc());
           })
           .Case<Forceable>([&](Forceable op) {
+            if (auto node = dyn_cast<NodeOp>(op.getOperation()))
+              traceResets(node.getResult(), node.getInput(), node.getLoc());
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -767,6 +767,8 @@ void InferResetsPass::traceResets(CircuitOp circuit) {
                         op.getLoc());
           })
           .Case<Forceable>([&](Forceable op) {
+            if (auto node = dyn_cast<NodeOp>(op.getOperation()))
+              traceResets(node.getResult(), node.getInput(), node.getLoc());
```
seldridge, line 771: If this is mutually exclusive with the following, then `else if` would be more appropriate.

seldridge (approved):
Nice find and fix!

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
dtzSiFive: Thanks! Nodes may be forceable, and if so we do also want to trace between the normal result and the rwprobe result, so I think this is correct as-is.

PR #7145, opened 2024-06-08
[Verif] Add formal test intent ops

dobios:
The goal of this PR is to introduce a first round of new test intent operations, for formal tests this time, to the verif dialect following [this proposal document](https://docs.google.com/document/d/1Ez6YosgVYuMtDg1MvBL8u9AMmntMARXhXofn_BcXlA4/edit?usp=sharing).  
  
This first PR will start by simply adding the ops with some verifier tests. Future PRs on this will have to then add a `PrepareForFormal` pass, and make sure that only `verif.formal` ops are used for formal verification and ignored elsewhere.  
  
The ops that have been added in this PR are:  
- [x] `verif.formal`  
- [x] `verif.symbolic_val`  
- [x] `verif.concrete_val`   

Feel free to let me know what you think!


include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
 }
 
+//===----------------------------------------------------------------------===//
+// Verification intent ops
+//===----------------------------------------------------------------------===//
+
+/// BMC -- short for bounded model check.
+def BMC : I32EnumAttrCase<"BMC", 0, "bmc">;
+/// KIP -- short for k-Inductive-Proof.
+def KIP : I32EnumAttrCase<"KIP", 0, "induction">;
```
maerhart, line 222: Maybe spell out the attribute names, e.g., `KInductiveProof` instead of `KIP`. Otherwise, people will probably have to frequently look up the meaning in the documentation here.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+    $test_region
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// Formal Verification Ops
+//===----------------------------------------------------------------------===//
+
+def SymbolicValOp : VerifOp<"symbolic_val", [Pure]> {
+  let summary = "declare a sybolic value for formal verification";
```
maerhart, line 262:
```suggestion
  let summary = "declare a symbolic value for formal verification";
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+    attr-dict  (`sym` $inner_sym^)? `(` (`k` `:` $k^)? (`type` `:` $test_type^)? `)` 
+    $test_region
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// Formal Verification Ops
+//===----------------------------------------------------------------------===//
+
+def SymbolicValOp : VerifOp<"symbolic_val", [Pure]> {
```
maerhart, line 261: Declaring it `pure` can lead to issues when you have multiple `symbolic_val` with the same result type since CSE would be allowed to merge them which would be equivalent to adding a constraint `symbolic_val1 == symbolic_val2`. To fix this, I think an allocation side-effect on the result value should be enough (which would still allow DCE to remove it when the result is never used).

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+  let summary = "declare a sybolic value for formal verification";
+  let description = [{
+    This operation declares a symbolic value that can then be used to reason 
+    about hw.instance inputs in a symbolic manner in future assertions and
```
maerhart, line 265:
```suggestion
    about `hw.instance` inputs in a symbolic manner in future assertions and 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+//===----------------------------------------------------------------------===//
+// Formal Verification Ops
+//===----------------------------------------------------------------------===//
+
+def SymbolicValOp : VerifOp<"symbolic_val", [Pure]> {
+  let summary = "declare a sybolic value for formal verification";
+  let description = [{
+    This operation declares a symbolic value that can then be used to reason 
+    about hw.instance inputs in a symbolic manner in future assertions and 
+    assumptions. The type result must be explicitly marked.
```
maerhart, line 266:
```suggestion
    assumptions. The result type must be explicitly marked.  
```

lib/Dialect/Verif/VerifOps.cpp
```diff
@@ -58,6 +58,26 @@ LogicalResult LogicEquivalenceCheckingOp::verifyRegions() {
...
 }
 
+//===----------------------------------------------------------------------===//
+// Formal Ops
+//===----------------------------------------------------------------------===//
+
+LogicalResult SymbolicValOp::verify() {
+  if (!isa<FormalOp>(getParentOp()))
+    return emitOpError()
+           << "Symbolic values can only be used in a formal test!";
```
maerhart, line 68: There is a `HasParent` constraint defined in tablegen that you could use to auto-generate this.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+
+  let assemblyFormat = "attr-dict `:` type($result)";
+
+  let hasVerifier = 1;
+}
+
+def ConcreteValOp : VerifOp<"concrete_val", [
+  Pure, AllTypesMatch<["init", "update", "result"]>
+]> {
+  let summary = "declare a concrete value for formatl verification";
```
maerhart, line 278:
```suggestion
  let summary = "declare a concrete value for formal verification";
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+  let description = [{
+    This opertaion declares a concrete value that can then be used in the
```
maerhart, line 280:
```suggestion
    This operation declares a concrete value that can then be used in the 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+}
+
+def ConcreteValOp : VerifOp<"concrete_val", [
+  Pure, AllTypesMatch<["init", "update", "result"]>
+]> {
+  let summary = "declare a concrete value for formatl verification";
+  let description = [{
+    This opertaion declares a concrete value that can then be used in the 
+    reasoning surrounding symbolic values, allowing for a form of concolic 
+    reasoning to take place in a verif.formal block. Concrete values are
```
maerhart, line 282:
```suggestion
    reasoning to take place in a `verif.formal` block. Concrete values are 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+]> {
+  let summary = "declare a concrete value for formatl verification";
+  let description = [{
+    This opertaion declares a concrete value that can then be used in the 
+    reasoning surrounding symbolic values, allowing for a form of concolic 
+    reasoning to take place in a verif.formal block. Concrete values are 
+    defined by an initial value and an update function.
+  }];
+
+  let arguments = (ins AnyType:$init, AnyType:$update);
```
maerhart, line 286: How will the update function be represented and what signature must it have? Should it be an SSA value of `FunctionType` or a symbol to a `func.func` ?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+//===----------------------------------------------------------------------===//
+
+def SymbolicValOp : VerifOp<"symbolic_val", [Pure]> {
+  let summary = "declare a sybolic value for formal verification";
+  let description = [{
+    This operation declares a symbolic value that can then be used to reason 
+    about hw.instance inputs in a symbolic manner in future assertions and 
+    assumptions. The type result must be explicitly marked.  
+  }];
+  let results = (outs AnyType:$result);
```
maerhart, line 268: If I declare an input to a module symbolic with this operation, will it have the same value across clock cycles, can it change any time, can it change according to some clock event?

maerhart (commented):
Nice to see all this progress!
Just added some (nit-picky) comments about things I saw while quickly reading through it, maybe they are useful.

include/circt/Dialect/Verif/VerifOps.td
TaoBi22: Seconded, and I wonder if a slight tweak to the name would be helpful here - when you talk about a symbolic value in z3, for example, that value can typically only have one assignment. Obviously that setting's a little different since there's no native sense of time, but I wonder if there's another name that could avoid confusion.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -212,4 +212,85 @@ def YieldOp : VerifOp<"yield", [
...
+}
+
+def FormalOp : VerifOp<"formal"> {
+  let summary = "defines a formal verification test";
+  let description = [{
+    This operation defines a formal verification test with potentially additional
+    assertions and assumptions. This should be written by declaring symbolic 
+    values that are then connected to a hardware instance. These symbols can then
+    be reasoned about in additional assertions. The region will then be converted 
+    into btor2, SMTLib, or SystemVerilog and ignored during regular SystemVerilog
```
TaoBi22, line 236: Nit: technically the output of an SMT lowering won't be pure SMTLib but a verif op containing SMT dialect - not the end of the world but another phrasing might be helpful in case anyone comes in expecting an exhaustive pure SMTLIB model of their design and can't figure out how to get one.

TaoBi22 (commented):
Excited to see where this leads!

include/circt/Dialect/Verif/VerifOps.td
dobios: Good point, I'll update that!

include/circt/Dialect/Verif/VerifOps.td
dobios: Ah right I didn't think about that, thanks for catching it!

lib/Dialect/Verif/VerifOps.cpp
dobios: Oh that's neat! I didn't know about that, thanks!

include/circt/Dialect/Verif/VerifOps.td
dobios: I was following the structure used for register `next` operands. So I was thinking just any operation that yields the same result as the result type of the concrete value. Seemed simpler, as this is something that should be easy to describe in Chisel (or any front-end). Do you think that using a function would be a better approach @maerhart ?

include/circt/Dialect/Verif/VerifOps.td
dobios: Yes, this would simply map to an input in btor so it would in theory not have more than one assignment. I think that symbolic_input would be a better name yeah, thanks !

include/circt/Dialect/Verif/VerifOps.td
dobios: Oh my bad, I will rephrase that

include/circt/Dialect/Verif/VerifOps.td
maerhart: So, essentially a register with an implicit clock? I was thinking of `%update` as a lambda function because you need some information about the previous cycle or the current cycle count to compute the input for the next cycle, but that also works with registers.

Maybe we could make this implicit concrete clock explicit (probably a block argument of `verif.formal`?) and just use the `seq.compreg` op instead of this op? That way we could just forward it to the DUT if we want a concrete clock. This should synergize way better with the BTOR backend than a lambda function. The SMT backend doesn't care too much about it since the update function would just be lowered to LLVM IR.

include/circt/Dialect/Verif/VerifOps.td
maerhart: Hmm, isn't it one assignment per clock cycle in BTOR internally? One assignment across all clock cycles seems overly restrictive to me. If we go the route of having a single concrete clock (as discussed above), this operation could be specified to hold the same value within a clock cycle of this concrete clock, but (potentially) another value in another cycle?

TaoBi22: I was under the same impression re: BTOR2 inputs - I think I was a bit unclear in the comment above, but what I was getting at was that an SMT variable can only have one assignment ever, whereas BTOR2 inputs can have different assignments on different clock cycles. I think the specification you're talking about makes the most sense @maerhart - maybe something like an SMT symbolic val (that can only have one assignment ever) would also be useful in some setting, but none spring to my mind

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
   ];
 }
 
+//===----------------------------------------------------------------------===//
+// Verification intent ops
+//===----------------------------------------------------------------------===//
+
+/// BMC -- short for bounded model check.
+def BMC : I32EnumAttrCase<"BMC", 0, "bmc">;
```
maerhart, line 270: Maybe use `BoundedModelCheck` for the TableGen and C++ name. The comment won't be necessary then.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+/// KIP -- short for k-Inductive-Proof.
```
maerhart, line 271: Strange comment, because you're not using this acronym anymore.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+def KInductiveProof : I32EnumAttrCase<"KInductiveProof", 1, "kinduction">;
+
+def FormalTestTypeAttr : I32EnumAttr<"FormalTestType", "formal test type",
```
maerhart, line 274: Nit: attribute summary could be a bit more descriptive.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+                                [BMC, KInductiveProof]> {
+  let cppNamespace = "circt::verif";
+}
+
+def FormalOp : VerifOp<"formal", [NoTerminator, RegionKindInterface]> {
+  let summary = "defines a formal verification test";
+  let description = [{
+    This operation defines a formal verification test with potentially additional
+    assertions and assumptions. This should be written by declaring symbolic
```
maerhart, line 283: `potentially additional assertions and assumptions` might be a bit confusing. Are they any different from regular assertions and assumptions in the design? AFAIU you could arbitrarily pull out assertions from the top level module or pull them in. You could also just inline the whole design into the formal op, right?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+
+def FormalOp : VerifOp<"formal", [NoTerminator, RegionKindInterface]> {
+  let summary = "defines a formal verification test";
+  let description = [{
+    This operation defines a formal verification test with potentially additional
+    assertions and assumptions. This should be written by declaring symbolic 
+    values that are then connected to a hardware instance. These symbols can then
+    be reasoned about in additional assertions. The region will then be converted 
+    into btor2, SystemVerilog, or used for verification directly in circt-bmc. 
+    The operations in this regiion are ignored during regular SystemVerilog
```
maerhart, line 287:
```suggestion
    The operations in this region are ignored during regular SystemVerilog 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+//===----------------------------------------------------------------------===//
+// Formal Verification Ops
+//===----------------------------------------------------------------------===//
+
+def SymbolicInputOp : VerifOp<"symbolic_input", [HasParent<"verif::FormalOp">]>{
+  let summary = "declare a symbolic input for formal verification";
+  let description = [{
+    This operation declares a symbolic input that can then be used to reason 
+    about `hw.instance` inputs in a symbolic manner in future assertions and 
+    assumptions. The result type must be explicitly marked.
```
maerhart, line 324: Is this really specific to `hw.instance` or could it be any instance or a fully inlined design without and instance operation in the `verif.formal`?
It's probably a good idea to describe a bit more precisely how this symbolic value behaves over time. Is it fixed over all clock cycles, can it change at a clock edge, can it change at any time?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+    This operation declares a concrete input that can then be used in the 
+    reasoning surrounding symbolic inputs, allowing for a form of concolic 
+    reasoning to take place in a `verif.formal` block. Concrete inputs are 
+    defined by an initial value and an update function.
+  }];
+
+  let arguments = (ins AnyType:$init, AnyType:$update);
+  let results = (outs AnyType:$result);
+
+  // Manual definition of a functional type is wanted
```
maerhart, line 345: So, `functional-type(operands, results)` is not working?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+
+def ConcreteInputOp : VerifOp<"concrete_input", [
+  AllTypesMatch<["init", "update"]>, HasParent<"verif::FormalOp">
+]> {
+  let summary = "declare a concrete input for formal verification";
+  let description = [{
+    This operation declares a concrete input that can then be used in the 
+    reasoning surrounding symbolic inputs, allowing for a form of concolic 
+    reasoning to take place in a `verif.formal` block. Concrete inputs are 
+    defined by an initial value and an update function.
```
maerhart, line 339: Can we specify when this update function is executed? Does this op effectively behave like a register with an implicit clock?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+      
+    Additional information about how the test should be conducted can be passed 
+    using the two supported attributes. These attributes are `k`, which defines 
+    the bounds of the test, and `type` which defines the type of formal test 
+    being done, i.e. Bounded Model Checking (BMC) or k-Induction.
+  }];
+
+  let arguments = (ins OptionalAttr<APIntAttr>:$k,
+                       OptionalAttr<FormalTestTypeAttr>:$test_type,
+                       OptionalAttr<InnerSymAttr>:$inner_sym);
```
maerhart, line 298: I think this operation should have a non-optional symbol. Not sure if we should add a regular MLIR symbol or make the inner symbol mandatory, though. I'd just follow what other ops do.

test/Dialect/Verif/basic.mlir
```diff
@@ -35,6 +35,27 @@ verif.cover %true label "foo3" : i1
...
+//===----------------------------------------------------------------------===//
+// Formal
+//===----------------------------------------------------------------------===//
+hw.module @Foo(in %0 "0": i1, in %clk "clk": !seq.clock, out "" : i1, out "1" : i1) {
+  %1 = seq.from_clock %clk
+  hw.output %0 , %1: i1, i1
+ }
+
+// CHECK: verif.formal sym @formal1(k : 20 : i64 type : bmc) {
+verif.formal sym @formal1(k : 20 type : bmc) {
```
maerhart, line 47: The assembly format in the parentheses is a bit weird. Maybe something like `(k = 20, type = bmc)` would be easier to understand.

include/circt/Dialect/Verif/VerifOps.td
TaoBi22: > It's probably a good idea to describe a bit more precisely how this symbolic value behaves over time. Is it fixed over all clock cycles, can it change at a clock edge, can it change at any time?

Agreed. Guessing from when we spoke about it before that it can only change on a clock-edge - we might want to add another op for a value that's fixed over all cycles (e.g. for defining BMC inputs while making sure we're adhering to the right semantics but not having to create a new BV every cycle when it's not necessary), but that's probably a matter for later.

TaoBi22 (approved):
Looking good! @maerhart covered most of my thoughts but a couple of other small points!

include/circt/Dialect/Verif/VerifOps.td
dobios: lol true

include/circt/Dialect/Verif/VerifOps.td
dobios: no it wasn't accepting multiple operands in a list, it only wants variadic operands, so something like `(ins:$args AnyType:$init Anytype:$update)` then `functional-type($args, $results)` was invalid because it was a list of arguments.

include/circt/Dialect/Verif/VerifOps.td
dobios: I see this as the op is executed on every "time-step" so it would be converted into a register with an implicit clock yes. I'm open to other ideas if you have any

include/circt/Dialect/Verif/VerifOps.td
dobios: I see, so have a symbol be auto-generated if not given by the user? The goal is to basically allow for someone to expree 
```
formal(k=20, type=Formal.BMC) { ... }  
```
in Chisel without requiring it to be stored anywhere -- but I guess we need to be able to refer to it somehow in the error report so maybe you're right

test/Dialect/Verif/basic.mlir
dobios: yeah I'm just not very good at writing the assembly regex format and couldn't figure out how to make the commas only needed if the two arguments were given. Any idea?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,91 @@ def YieldOp : VerifOp<"yield", [
...
+]> {
+  let summary = "declare a concrete input for formal verification";
+  let description = [{
+    This operation declares a concrete input that can then be used in the 
+    reasoning surrounding symbolic inputs, allowing for a form of concolic 
+    reasoning to take place in a `verif.formal` block. Concrete inputs are 
+    defined by an initial value and an update function.
+  }];
+
+  let arguments = (ins AnyType:$init, AnyType:$update);
```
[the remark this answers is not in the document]
dobios: The update function currently is designed to have similar semantics to the next function for registers -- this was because my goal was to have this concrete op define an implicit register with an implicit clock I think -- I'm open to other suggestions if you think there would be a better way to design this !

include/circt/Dialect/Verif/VerifOps.td
dobios: So would it be a good idea to split this into e.g. `symbolic_input` and `symbolic_const` or rather simply have `symbolic_input` take a boolean attribute `const` that can be defined using a special `const` keyword (similar to sync /async for registers?

maerhart: Both approaches are fine, but we don't necessarily have to do that right now. I just wanted to make sure that the semantics of the op is well described in the documentation to avoid any confusions in the future.

include/circt/Dialect/Verif/VerifOps.td
maerhart: Yeah, the `$args` looks invalid, but did you try literally `functional-type(operands, results)` (without any $ before the operands and results)?

include/circt/Dialect/Verif/VerifOps.td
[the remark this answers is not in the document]
TaoBi22: Ah, I see, so it would just be an SSA value providing the next value to be assigned? That sounds like a good approach, but in that case I'm not sure I'd call it an update function, as to me that implies the presence of some region or `func.func` that explicitly takes operands and returns an updated value. Maybe it would be worth saying that it's an update value, and explicitly outlining that it's just an SSA value that can be dependent on the previous value by having a path back from the output?

include/circt/Dialect/Verif/VerifOps.td
TaoBi22: Good point - personally I lean in favour of separate ops as a) I wouldn't say `symbolic_input` isn't a super clear name for a constant value (and it would nice to keep the clear mapping to SMT vs BTOR) and b) having two separate ops may just leave less room for confusion, but I'm very happy to be overruled!

include/circt/Dialect/Verif/VerifOps.td
maerhart: In its current state one could argue that it's not really necessary because you could just use the regular register operation, but I think it's still useful to have it so we still know that it is an input and not just part of the design. Also, providing the clock for the regular register would probably require a block argument from `verif.formal`. So, I'd leave it as is. I just think, it'd just be nice to describe this implicit clock and how the value updating works a bit more in the description. Basically, `verif.formal` provides an implicit base clock and the symbolic and concrete operations provide values for all inputs (including the actual clocks) based on this implicit clock.

include/circt/Dialect/Verif/VerifOps.td
maerhart: I'd make it mandatory for the user to provide a unique name for the test. Or is there a particular reason that shouldn't be the case?
E.g., in Rust (and other languages) a unit test is also a function that has a unique name with an attribute/annotation that it is a test.

test/Dialect/Verif/basic.mlir
maerhart: Why are they optional? If they have a default value, you can make them default-valued and just print them in the assembly format regardless.

include/circt/Dialect/Verif/VerifOps.td
dobios: Wouldn't we also be able to define a value that is only set once by using a symbolic input as the init value of a concrete input and then have the update value being simply constant?

maerhart: Good point. By doing the following, right?
```
%0 = verif.symbolic_input : i32
%1 = verif.concrete_input %0, %1 : i32
```

include/circt/Dialect/Verif/VerifOps.td
dobios: ah I was doing `functional-type(arguments, results)`, with operands it worked thanks!

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,85 @@ def YieldOp : VerifOp<"yield", [
...
+  let summary = "defines a formal verification test";
+  let description = [{
+    This operation defines a formal verification test. This should be written 
+    by declaring symbolic values that are then connected to a hardware instance. 
+    These symbols can then be used in additional assertions that are defined outside 
+    of the instanciated `hw.module` but inside of this region. Assertions/Assumptions 
+    defined within the instanciated module will also be used for verification. 
+    The region will then be converted into btor2, SystemVerilog, or used for verification 
+    directly in circt-bmc. The operations in this region are ignored during regular SystemVerilog 
+    emission.
```
maerhart, line 284: * 80 cols
* Can we briefly say why this is ignored in regular SystemVerilog? Is it because it cannot be expressed in SV or for another reason?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,85 @@ def YieldOp : VerifOp<"yield", [
...
+  IsolatedFromAbove, Symbol
+]> {
+  let summary = "defines a formal verification test";
+  let description = [{
+    This operation defines a formal verification test. This should be written 
+    by declaring symbolic values that are then connected to a hardware instance. 
+    These symbols can then be used in additional assertions that are defined outside 
+    of the instanciated `hw.module` but inside of this region. Assertions/Assumptions 
+    defined within the instanciated module will also be used for verification. 
+    The region will then be converted into btor2, SystemVerilog, or used for verification
```
maerhart, line 282: Nit: it's also fine to lower it to other backends, those are just options.
```suggestion
    The region can then be converted into btor2, SystemVerilog, or used for verification 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,85 @@ def YieldOp : VerifOp<"yield", [
...
+  NoTerminator, RegionKindInterface, HasParent<"mlir::ModuleOp">, 
+  IsolatedFromAbove, Symbol
+]> {
+  let summary = "defines a formal verification test";
+  let description = [{
+    This operation defines a formal verification test. This should be written 
+    by declaring symbolic values that are then connected to a hardware instance. 
+    These symbols can then be used in additional assertions that are defined outside 
+    of the instanciated `hw.module` but inside of this region. Assertions/Assumptions 
+    defined within the instanciated module will also be used for verification.
```
maerhart, line 281:
```suggestion
    of the instantiated `hw.module` but inside of this region. Assertions/Assumptions 
    defined within the instantiated module will also be used for verification. 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,85 @@ def YieldOp : VerifOp<"yield", [
...
+    by declaring symbolic values that are then connected to a hardware instance. 
+    These symbols can then be used in additional assertions that are defined outside 
+    of the instanciated `hw.module` but inside of this region. Assertions/Assumptions 
+    defined within the instanciated module will also be used for verification. 
+    The region will then be converted into btor2, SystemVerilog, or used for verification 
+    directly in circt-bmc. The operations in this region are ignored during regular SystemVerilog 
+    emission.   
+      
+    Additional information about how the test should be conducted can be passed 
+    using the `k` attribute, which defines the bounds of the test.
```
maerhart, line 287: Nit:
```suggestion
    The attribute `k` defines the upper bound of cycles to check for this test w.r.t. the implicit clock defined by this operation within its region.
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,85 @@ def YieldOp : VerifOp<"yield", [
...
+
+//===----------------------------------------------------------------------===//
+// Formal Verification Ops
+//===----------------------------------------------------------------------===//
+
+def SymbolicInputOp : VerifOp<"symbolic_input", [HasParent<"verif::FormalOp">]>{
+  let summary = "declare a symbolic input for formal verification";
+  let description = [{
+    This operation declares a symbolic input that can then be used to reason 
+    about `hw.instance` inputs in a symbolic manner in future assertions and
```
maerhart, line 315: Does `future` have a specific meaning here w.r.t. the implicit clock? Maybe remove this word?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,85 @@ def YieldOp : VerifOp<"yield", [
...
+// Formal Verification Ops
+//===----------------------------------------------------------------------===//
+
+def SymbolicInputOp : VerifOp<"symbolic_input", [HasParent<"verif::FormalOp">]>{
+  let summary = "declare a symbolic input for formal verification";
+  let description = [{
+    This operation declares a symbolic input that can then be used to reason 
+    about `hw.instance` inputs in a symbolic manner in future assertions and 
+    assumptions. The result type must be explicitly marked. The resulting value 
+    is a new non-deterministic value for every implicit clock cycle.
```
maerhart, line 317:
```suggestion
    is a new non-deterministic value for every clock cycle of the implicit clock defined by the parent `verif.formal`. 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,85 @@ def YieldOp : VerifOp<"yield", [
...
+def ConcreteInputOp : VerifOp<"concrete_input", [
+  AllTypesMatch<["init", "update"]>, HasParent<"verif::FormalOp">
+]> {
+  let summary = "declare a concrete input for formal verification";
+  let description = [{
+    This operation declares a concrete input that can then be used in the 
+    reasoning surrounding symbolic inputs, allowing for a form of concolic 
+    reasoning to take place in a `verif.formal` block. Concrete inputs are 
+    defined by an initial value and an update SSA value. This is equivalent 
+    to a register with an implicit clock, which is initialized with the init
```
maerhart, line 333:
```suggestion
    to a register with the parent `verif.formal`'s implicit clock, which is initialized with the init 
```

maerhart (approved):

include/circt/Dialect/Verif/VerifOps.td
dobios: Yes exactly

TaoBi22: This would work functionally - the disadvantage to me is that we'd have an unnecessary `symbolic_input` floating around, which I guess semantically in the case of model checking would require a new SMT var to be created for that input each cycle, though it wouldn't actually be used. We could pattern match for this case and optimize it away in VerifToSMT I guess, but it would mean we couldn't immediately reject `symbolic_input` ops which would be nice to do (since the meaning of that might get a little messy with the init/loop regions). Happy to just worry about this in the future though!

PR #7146, opened 2024-06-08
[Arc] Add VectorizeOp canonicalization

elhewaty:


lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -577,6 +583,63 @@ CompRegCanonicalizer::matchAndRewrite(seq::CompRegOp op,
...
+
+      if (!operand.hasOneUse())
+        return failure();
+
+      uniqueOperands.insert(operand);
+    }
+
+    if (allFromTheSameGroup) {
+      if (uniqueOperands.size() != groupSize)
+        return failure();
```
TaoBi22, line 624: Nit, but will `uniqueOperands` ever not just be a copy of groups? As far as I can tell, the only case in which they would be different is where you break from the for loop over `group` early, in which case you'll return failure anyway as the size won't equal groupSize, so you could avoid a little bit of overhead with
```suggestion
    for (auto operand : group) {
      if (operand.getDefiningOp() != defOp)
        return failure();

      if (!operand.hasOneUse())
        return failure();
    }
```
or maybe this is just for type conversion?

elhewaty: Thank you very much for the review.
> Will `uniqueOperands` ever not just be a copy of groups?

It will be a subset of `groups` of course but I need to make sure that all the operands of the definedOp appear in this operation and only have one use because this is one of the conditions to merge ops (This is an initial thoughts)

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -577,6 +583,63 @@ CompRegCanonicalizer::matchAndRewrite(seq::CompRegOp op,
...
+      // VectorizeOp and they are forming a full vector parameter.
+      
+      // Now we need to pull the `defOp`'s body into `op`'s body
+      // approach:
+      // get the operands of the `defOp` and devide them by two
+      // now we have LHS and RHS, pass the LHS and RHS as a parameter
+      // to the newOp (op) do the calculation (is there a way to know it?)
+      // after this do the calculation of the current op (the op we pulled
+      // the first op into) and boooom! we did it!!
+      // will this work??? am I on the right path so far???
```
elhewaty, line 636: @TaoBi22 @fabianschuiki @maerhart Can `RewriterBase::inlineRegionBefore` help with this?
or `RewriterBase::inlineBlockBefore` is better?

fabianschuiki: Yeah the inlining should be able to do this I believe. You might have to set up an `IRMapping` with mappings from the other vector's block args to the block args in the vector you're merging into.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
fabianschuiki: You can probably very quickly check whether one of the vector inputs is defined by another vector by comparing the `OperandRange`s:
```
// This gives you an OperandRangeRange, i.e. a 2D array of operands.
// The outer dimension corresponds to arguments to the body block.
// The inner dimension corresponds to individual vector elements.
auto inputVecs = vecOp.getInputs();

// This iterates over the vectors flowing into the vectorize op:
for (auto inputVec : inputVecs) {
  // inputVec is an OperandRange, where each element corresponds to a vector element.
  // You can pick the first element, get its defining op, check if it's a vectorize op, and check if
  // all elements in `inputVec` are exactly the elements from that other vectorize op.
  auto otherVecOp = inputVec[0].getDefiningOp<VectorizeOp>();
  if (!otherVecOp) continue;
  if (inputVec != otherVecOp.getResults()) continue;

  // Now you know that inputVec matches the results of otherVecOp exactly.
  // Now you can check if the current vecOp is the sole user of otherVecOp.
  if (!llvm::all_of(otherVecOp.getResults(), [](auto result) { return result.hasOneUse(); }))
    continue;

  // At this point you know that inputVec is exactly the result from otherVecOp, and
  // that this is the only use of otherVecOp. Now you should be able to merge the two.
}
```

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -577,6 +584,83 @@ CompRegCanonicalizer::matchAndRewrite(seq::CompRegOp op,
...
+    for (auto otherArg : otherBlock.getArguments()) {
+      auto newArg = currentBlock.insertArgument(
+          argIdx + paddedBy, otherArg.getType(), otherArg.getLoc());
+      argMapping.map(otherArg, newArg);
+      ++paddedBy;
+    }
+
+    rewriter.setInsertionPointToStart(&currentBlock);
+    for (auto &op : otherBlock.without_terminator())
+      rewriter.clone(op, argMapping);
```
maerhart, line 630: Using `rewriter.inlineBlockBefore` would avoid cloning operations since the original ops are removed anyways.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -577,6 +584,83 @@ CompRegCanonicalizer::matchAndRewrite(seq::CompRegOp op,
...
+    rewriter.replaceAllUsesWith(currentBlock.getArgument(argNewPos),
+                                argMapping.lookupOrDefault(retOp.getValue()));
+    currentBlock.eraseArgument(argNewPos);
+    vecOpsToRemove.push_back(otherVecOp);
+    // We erased an arg so the padding decreased by 1
+    paddedBy--;
+  }
+
+  if (!canBeMerged)
+    return success();
```
maerhart, line 645: If the pattern cannot be applied, it should return `failure()`, but you need to make sure that no modification to the IR has already happened in that case.
Returning `success()` in such a case means that the operation is pushed onto the worklist again since the driver thinks the pattern was applied successfully and modified something. This leads to unnecessary inefficiencies.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -577,6 +584,83 @@ CompRegCanonicalizer::matchAndRewrite(seq::CompRegOp op,
...
+
+  // Set the new inputOperandSegments value
+  SmallVector<int32_t> newAttr;
+  unsigned groupSize = vecOp.getResults().size();
+  unsigned numOfGroups = newOperands.size() / groupSize;
+  for (unsigned groupIdx = 0; groupIdx < numOfGroups; ++groupIdx)
+    newAttr.push_back(groupSize);
```
maerhart, line 652: Nit:
```suggestion
  unsigned groupSize = vecOp.getResults().size();
  unsigned numOfGroups = newOperands.size() / groupSize;
  SmallVector<int32_t> newAttr(numOfGroups, groupSize);
```

test/Dialect/Arc/arc-canonicalizer.mlir
```diff
@@ -295,3 +295,52 @@ hw.module @DontSinkDifferentConstants1(in %x: i4, out out0: i4, out out1: i4, ou
   hw.output %0, %1, %2 : i4, i4, i4
 }
 // CHECK-NEXT: }
+
+//===----------------------------------------------------------------------===//
+// MergeVectorizeOps
+//===----------------------------------------------------------------------===//
+
+hw.module @VecOpTest(in %b: i8, in %e: i8, in %h: i8, in %k: i8, in %c: i8, in %f: i8,
```
maerhart, line 303: Maybe add a test where multiple input vectors of a single `arc.vectorize` come from other vectorize ops and can be merged?

maerhart: Also, it'd be nice to have a test for the case where the results of the vectorize op to be merged into the current one has other users.

maerhart: And also one where the vector is not taken 1-1 but some reshuffling would need to happen.

maerhart (commented):
Great job! So cool that this is working now.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -577,6 +584,83 @@ CompRegCanonicalizer::matchAndRewrite(seq::CompRegOp op,
...
+    }
+
+    canBeMerged = true;
+    newOperands.insert(newOperands.end(), otherVecOp.getOperands().begin(),
+                       otherVecOp.getOperands().end());
+
+    auto &otherBlock = otherVecOp.getBody().front();
+    for (auto otherArg : otherBlock.getArguments()) {
+      auto newArg = currentBlock.insertArgument(
+          argIdx + paddedBy, otherArg.getType(), otherArg.getLoc());
```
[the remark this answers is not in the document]
elhewaty: I think there are no `insertArguments` in `mlir::Block` ? I think I will need to iterate over them for mapping ?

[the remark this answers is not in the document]
elhewaty: I think `inlineBlockBefore` will handle this.

test/Dialect/Arc/arc-canonicalizer.mlir
elhewaty: > And also one where the vector is not taken 1-1 but some reshuffling would need to happen.

```
auto otherVecOp = inputVec[0].getDefiningOp<VectorizeOp>();
    auto resultVec = otherVecOp.getResults();
    auto comparator = [](const mlir::Value &a, const mlir::Value &b) {
      return a.getAsOpaquePointer() < b.getAsOpaquePointer();
    };
    std::sort(inputVec.begin(), inputVec.end(), comparator);
    std::sort(resultVec.begin(), resultVec.end(), comparator);
```

Can we sort both? if not do you have any other way?

elhewaty: > And also one where the vector is not taken 1-1 but some reshuffling would need to happen.

Is there an easy way to sort `inputVec` and `resVec = otherVecOp.getResults()`?
Maybe I can store inputVec in `DensMap` and check but that will need to specialize the operators.

maerhart: I don't understand why you would need to sort something here. For this test, your patten should not apply, i.e., it should not merge the two `arc.vectorize` ops.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
maerhart: The body blocks don't have any successors since they are the only blocks in the region.
You'd need to modify the block arguments of the `otherVecOp` instead of the current block and pass the adequate values as the last argument of `inlineBlockBefore`. I don't think the code will get more complex by doing so, but it requires a few changes to your code. I'm also fine with leaving it as is for now. We can come back and get rid of the unnecessary cloning if we observe performance issues.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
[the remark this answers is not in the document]
maerhart: Ah sry, I thought there exists a `insertArguments` because there also exists an `addArguments`, but you're right.

fabianschuiki:
This is starting to look really nice! Any idea how well this works on the cores in the arc-tests repository? Are there any cases that don't work yet?

elhewaty:
> This is starting to look really nice! Any idea how well this works on the cores in the arc-tests repository? Are there any cases that don't work yet?

It works well on the dual rocket core we vectorized before, but just a few merges!!
The problem here is that if we allow `ClockTypes` it crashes when at `otherArg.getType()` but after some work I noticed the problem is not that `ClockTypes` crashes but they may with some args that will make it crash as if we allow all args to be of the same type there's no crash happens!! so what can make `otherArg.getType` crash?!! that is the problem I can allow Clocks, Ints, and floats this may work with the tests but fail in some other cases.

fabianschuiki (approved):
Looks good to me! Nice job getting this up and running :slightly_smiling_face:.

I liked @maerhart's suggestion of having a test where one vectorize op feeds into another, but with the operands shuffled, such that the pattern should not apply. That might be worth adding before merging.

elhewaty:
> I liked @maerhart's suggestion of having a test where one vectorize op feeds into another, but with the operands shuffled, such that the pattern should not apply. That might be worth adding before merging.

I think there is a test for this, there is a function called `Needs_Shuffle` at the enf of the tests? Or do you mean something else?

fabianschuiki:
Oh you're totally right, I overlooked that. Very nice 😃

test/Dialect/Arc/arc-canonicalizer.mlir
```diff
@@ -295,3 +295,183 @@ hw.module @DontSinkDifferentConstants1(in %x: i4, out out0: i4, out out1: i4, ou
...
+  %4 = arc.state @FooMux(%en, %C#0, %4) clock %clock latency 1 : (i1, i8, i8) -> i8
+}
+
+// CHECK-LABEL: hw.module @More_Than_One_Use(in %b : i8, in %e : i8, in %h : i8, in %k : i8, in %c : i8, in %f : i8, in %i : i8, in %l : i8, in %n : i8, in %p : i8, in %r : i8, in %t : i8, in %en : i1, in %clock : !seq.clock) {
+// CHECK-NEXT:    [[VEC0:%.+]]:4 = arc.vectorize (%b, %e, %h, %k), (%c, %f, %i, %l) : (i8, i8, i8, i8, i8, i8, i8, i8) -> (i8, i8, i8, i8) {
+// CHECK-NEXT:    ^[[BLOCK:[[:alnum:]]+]](%arg0: i8, %arg1: i8):
+// CHECK-NEXT:      [[ADD:%.+]] = comb.add %arg0, %arg1 : i8
+// CHECK-NEXT:      arc.vectorize.return [[ADD]] : i8
+// CHECK-NEXT:    }
+// CHECK-NEXT:    [[VEC1:%.+]]:4 = arc.vectorize ([[VEC0]]#0, [[VEC0]]#1, [[VEC0]]#2, [[VEC0]]#3), (%n, %p, %r, %t), ([[VEC0]]#0, [[VEC0]]#1, [[VEC0]]#2, [[VEC0]]#3) : (i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8) -> (i8, i8, i8, i8) {
```
maerhart, line 406: This could be another optimization pattern: if a vector is given multiple times as input to the vectorize op, we can remove one of them and replace usages with the other occurrence.
But that's something for another PR.

test/Dialect/Arc/arc-canonicalizer.mlir
```diff
@@ -295,3 +295,183 @@ hw.module @DontSinkDifferentConstants1(in %x: i4, out out0: i4, out out1: i4, ou
...
+// CHECK-NEXT:    }
+// CHECK-NEXT:    [[STATE:%.+]] = arc.state @FooMux(%en, [[VEC1]]#0, [[STATE]]) clock %clock latency 1 : (i1, i8, i8) -> i8
+// CHECK-NEXT:    hw.output
+// CHECK-NEXT:  }
+
+arc.define @TLMonitor_14_arc(%arg0: i3) -> i3 {
+  arc.output %arg0 : i3
+}
+
+hw.module private @Self_Use(in %clock : !seq.clock, in %reset : i1, in %io_in_a_ready : i1, in %io_in_a_valid : i1, in %io_in_a_bits_opcode : i3, in %io_in_a_bits_param : i3, in %io_in_a_bits_size : i3, in %io_in_a_bits_source : i3, in %io_in_a_bits_address : i17, in %io_in_a_bits_mask : i8, in %io_in_a_bits_corrupt : i1, in %io_in_d_ready : i1, in %io_in_d_valid : i1, in %io_in_d_bits_size : i3, in %io_in_d_bits_source : i3) {
```
maerhart, line 420: Nit: there are many unused input ports, we could trim them a bit.

maerhart (approved):
Nice work! 🎉

test/Dialect/Arc/arc-canonicalizer.mlir
elhewaty: I thought of this too. I will

test/Dialect/Arc/arc-canonicalizer.mlir
elhewaty: Done! Please merge if it's ok

PR #7150, opened 2024-06-10
[FIRRTL][Verif][LTL] Replace `ltl.disable` with an enable folded into `verif.assert`

dobios:
~This PR fixes a bug in ExpandWhens that incorrectly folded the when condition into the property.~
~We should really have a better system than calling everything a `uint1` to avoid errors like this in the future ^^"~

Edit: This PR gets rid of the `ltl.disable` op and intrinsic in favor of having an `enable` operand on the verif assert like ops. This made it trivial to fix an incorrect `when` condition folding for `AssertProperty`. Most of the PR relates to fixing the chaos that was caused by removing `ltl.disable`.


fabianschuiki:
> We should really have a better system than calling everything a `uint1` to avoid errors like this in the future ^^"

I totally agree :grinning:. @uenoku had some thoughts a while back on adding intrinsic types that served a similar role as intrinsic ops. Types are a lot trickier to get right though -- I remember there were a lot of concerns about a generic opaque intrinsic type. Although it would solve this particular problem here.

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
```diff
@@ -535,19 +535,52 @@ class WhenOpVisitor : public LastConnectResolver<WhenOpVisitor> {
...
+      }
+      foldedWhen = builder.create<LTLClockIntrinsicOp>(
+          op.getLoc(), foldedWhen.getType(), foldedWhen, clockOp.getClock());
+    }
+    if (foldedWhen) {
+      op.getPropertyMutable().assign(foldedWhen);
+
+      // destroy the old clock and disable ops
+      clockOp.erase();
+      disable.erase();
```
fabianschuiki, line 582: I think you have to individually check if these are now dead, because the clocked properties could theoretically be used by multiple assert/assumes. Something like:
```suggestion
      if (clockOp && isOpTriviallyDead(clockOp))
        clockOp.erase();
      if (disable && isOpTriviallyDead(disable))
        disable.erase();
```

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
```diff
@@ -535,19 +535,52 @@ class WhenOpVisitor : public LastConnectResolver<WhenOpVisitor> {
...
+    OpBuilder builder(op);
+    Value foldedWhen;
+    LTLClockIntrinsicOp clockOp;
+    LTLDisableIntrinsicOp disable;
+    // Extract the clock and disable from the UInt<1>
+    if ((clockOp =
+             dyn_cast<LTLClockIntrinsicOp>(op.getProperty().getDefiningOp()))) {
+      if ((disable = dyn_cast<LTLDisableIntrinsicOp>(
+               clockOp.getInput().getDefiningOp()))) {
+
```
fabianschuiki, line 569: What happens if the nesting is flipped, e.g. `disable(clock(P))`? The condition should probably be folded into the disable here as well, shouldn't it?

fabianschuiki: Also, what happens if this is just `assert(P)` without a disable? Should this create a disable like `assert(disable(P))` with the when condition?

dobios: For the case where assert has no disable, I think that it's totally legal to add a disable if it's wrapped in a when

dobios: I added a canonical ordering for clock and disable and renamed disable's operands so that they make sense and are not just `lhs` and `rhs`, but instead `input` and `disable`

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
```diff
@@ -535,19 +535,54 @@ class WhenOpVisitor : public LastConnectResolver<WhenOpVisitor> {
...
+    }
+    if (foldedWhen) {
+      op.getPropertyMutable().assign(foldedWhen);
+
+      // destroy the old clock and disable ops
+      if (clockOp && isOpTriviallyDead(clockOp))
+        clockOp.erase();
+      if (disable && isOpTriviallyDead(disable))
+        disable.erase();
+    }
```
uenoku, line 585: I'm worried about the pattern matching affecting semantics of when, e.g. inserting a node for clock intrinsic changes a semantics of two apparently equivalent IR. 

```mlir
firrtl.circuit "verifassert" {
firrtl.module @verifassert(
  in %clock : !firrtl.clock, in %cond : !firrtl.uint<1>, in %prop : !firrtl.uint<1>, in %prop1 : !firrtl.uint<1>,
  in %disable : !firrtl.uint<1>, in %disable1 : !firrtl.uint<1>, in %reset : !firrtl.uint<1>
) {
  firrtl.when %cond : !firrtl.uint<1> {

    %dis_prop = firrtl.int.ltl.disable %prop, %disable : (!firrtl.uint<1>, !firrtl.uint<1>) -> !firrtl.uint<1>
    %clk_prop = firrtl.int.ltl.clock %dis_prop, %clock : (!firrtl.uint<1>, !firrtl.clock) -> !firrtl.uint<1>
    %clk_prop_node = firrtl.node %clk_prop: !firrtl.uint<1>
    firrtl.int.verif.assert %clk_prop_node : !firrtl.uint<1>
    firrtl.int.verif.assert %clk_prop : !firrtl.uint<1>
  } else {
  }
}
}
```

```verilog
  assert property (@(posedge clock) disable iff (disable_0) prop);      // bar.mlir:9:17, :10:17, :12:5
  wire _GEN = cond | disable_0; // bar.mlir:4:33
  assert property (@(posedge clock) disable iff (_GEN) prop);   // bar.mlir:4:33, :13:5
```

We could look through nodes/wires but I guess it's still fragile (for example running  canonicalization before ExpandWhen might change the semantics). I agree that intrinsic/foreign type is actually good solution here.

dobios: I'm not entirely sure I see what the issue is here, could you elaborate @uenoku ?

uenoku: In the first assert ` firrtl.int.verif.assert %clk_prop_node` a condition (`%cond`) is dropped in ExpandWhen.     Whereas for the second assert `firrtl.int.verif.assert %clk_prop`, `%cond` is preserved.  I think it's problematic that existence of node affects the semantics.

dobios:
So after some discussion, this PR will shift focus towards replacing disable intrinsics with an enable that is folded into the verif intrinsics. This will make the fix in expandWhens as simple as for the other assertion type.

include/circt/Dialect/LTL/LTLOps.td
```diff
@@ -160,10 +160,7 @@ def ConcatOp : LTLOp<"concat", [Pure]> {
   }];
 }
 
-def RepeatOp : LTLOp<"repeat", [
-  Pure,
-  PredOpTrait<"'base' must be >= 0", CPred<"$base >= 0">>
```
uenoku, line 165: Why is this trait removed?

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -2033,6 +2032,10 @@ Value FIRRTLLowering::getPossiblyInoutLoweredValue(Value value) {
 /// This returns a null value for FIRRTL values that cannot be lowered, e.g.
 /// unknown width integers.
 Value FIRRTLLowering::getLoweredValue(Value value) {
+  // Allow for optional parameters
+  if (!value)
+    return value;
```
uenoku, line 2037: Inserting an additional null check just for optional enable seems over kill. Can we do this in the call site `op.getEnable()? getLoweredValue(op.getEnable()): Value()`?

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -3791,26 +3794,23 @@ LogicalResult FIRRTLLowering::visitExpr(LTLClockIntrinsicOp op) {
...
 
-LogicalResult FIRRTLLowering::visitExpr(LTLDisableIntrinsicOp op) {
-  return setLoweringToLTL<ltl::DisableOp>(
-      op,
-      ValueRange{getLoweredValue(op.getLhs()), getLoweredValue(op.getRhs())});
-}
-
 LogicalResult FIRRTLLowering::visitStmt(VerifAssertIntrinsicOp op) {
   builder.create<verif::AssertOp>(getLoweredValue(op.getProperty()),
+                                  getLoweredValue(op.getEnable()),
```
uenoku, line 3799: nit: Can we avoid calling (potentially) side-effecting functions in function arguments since their evaluation order is undefined? 

getLoweredValue
```suggestion
auto property = getLowerdValue(op.getProperty());
auto enable = op.getEnable() ? getLoweredValue(op.getEnable()): Value();
builder.create<verif::AssertOp>(property, enable);
```

lib/Dialect/Verif/VerifOps.cpp
```diff
@@ -42,6 +54,46 @@ OpFoldResult HasBeenResetOp::fold(FoldAdaptor adaptor) {
...
+  // assertlike(ltl.clock(prop, clk), en) -> clocked_assertlike(prop, en, clk)
+  template <typename TargetOp, typename Op>
+  static LogicalResult canonicalize(Op op, PatternRewriter &rewriter) {
+    // If the property is a block argument, then no canonicalization is possible
+    Value property = op.getProperty();
+    if (!property.getDefiningOp())
+      return success();
+
+    // Check for clock operand
+    if (auto clockOp = dyn_cast<ltl::ClockOp>(property.getDefiningOp()))
```
uenoku, line 74: nit: you can simplify this with getDefiningOp<..>. Also returning `success` is incorrect in the first condition.
```suggestion
    auto clockOp = op.getProperty.getDefiningOp<ltl::ClockOp>();
    if (!clockOp) return failure();
    ...
```

lib/Dialect/Verif/VerifOps.cpp
```diff
@@ -42,6 +54,46 @@ OpFoldResult HasBeenResetOp::fold(FoldAdaptor adaptor) {
...
+//===----------------------------------------------------------------------===//
+// AssertLikeOps Canonicalizations
+//===----------------------------------------------------------------------===//
+
+namespace {
+struct AssertLikeOp {
+public:
+  // Pattern:
+  // assertlike(ltl.clock(prop, clk), en) -> clocked_assertlike(prop, en, clk)
+  template <typename TargetOp, typename Op>
```
uenoku, line 66: nit: Can we remove outer struct if it's used as just a namespace? Just defining static function looks OK to me.

uenoku (commented):
Generally looks great! Thank you for cleaning up!

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -3513,10 +3509,20 @@ class PropertyEmitter : public EmitterBase,
...
   assert(localTokens.empty());
-  // Wrap to this column.
+
+  // If the property is tied to an enable, emit that.
+  if (enable) {
+    ps << "disable iff" << PP::nbsp << "(~(";
+    ps.scopedBox(PP::ibox2, [&] {
+      emitNestedProperty(enable, PropertyPrecedence::Lowest);
+      ps << "))";
+    });
```
uenoku, line 3522: I think we can set `Unary` precedence. 
```suggestion
    ps << "disable iff" << PP::nbsp << "(~(";
    ps.scopedBox(PP::ibox2, [&] {
      emitNestedProperty(enable, PropertyPrecedence::Unary);
      ps << ")";
    });
```

lib/Conversion/LTLToCore/LTLToCore.cpp
```diff
@@ -147,68 +90,6 @@ struct HasBeenResetOpConversion : OpConversionPattern<verif::HasBeenResetOp> {
...
-    // Generate the parenting sv.always posedge clock from the ltl
-    // clock, containing the generated sv.assert
-    rewriter.create<sv::AlwaysOp>(
-        clockOp.getLoc(), ltlToSVEventControl(clockOp.getEdge()), ltlClock,
-        [&] {
-          // Generate the sv assertion using the input to the
-          // parenting clock
-          rewriter.replaceOpWithNewOp<sv::AssertOp>(
-              op, disableVal,
-              sv::DeferAssertAttr::get(getContext(),
```
uenoku, line 200: Great to see we don't need complicated pattern matching anymore!

lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
```diff
@@ -447,16 +461,22 @@ class CirctVerifConverter : public IntrinsicConverter {
   using IntrinsicConverter::IntrinsicConverter;
 
   bool check(GenericIntrinsic gi) override {
-    return gi.hasNInputs(1) || gi.sizedInput<UIntType>(0, 1) ||
+    return gi.hasNInputs(1, 2) || gi.sizedInput<UIntType>(0, 1) ||
```
uenoku, line 464: Could you update FIRRTLIntrinsic.md as well?

lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
```diff
@@ -25,6 +25,20 @@ ParseResult GenericIntrinsic::hasNInputs(unsigned n) {
   return success();
 }
 
+// Checks for a number of operands between n and m (allows for (n-m) optional
+// parameters)
+ParseResult GenericIntrinsic::hasNInputs(unsigned n, unsigned m) {
```
darthscsi, line 30: Please give a better name.  the behavior is not has n inputs.

darthscsi: Also, isn't this something which is already handled by variadic constraints in td?

include/circt/Dialect/LTL/LTLOps.td
dobios: it was generating a lot of warnings because the generated code is >= 0 by default

lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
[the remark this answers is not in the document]
dobios: my idea was to have something that said "between n and m parameters are allowed" rather than having "n params with c optional params". Both work, which one would you prefer?

lib/Dialect/Verif/VerifOps.cpp
dobios: that doesn't work apparently, like `getDefiningOp<ltl::ClockOp>()` doesn't compile anymore

lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
dobios: yes but it wasn't supported for generic intrinsics. This adds support for that. I can name it something like `hasNToMInputs`

lib/Dialect/Verif/Transforms/VerifyClockedAssertLike.cpp
```diff
@@ -86,12 +86,7 @@ struct VerifyClockedAssertLikePass
       if (!defOp || handledOps.contains(defOp))
         continue;
 
-      // This is triggered if our operand is already in the worklist and
-      // wasn't handled
-      if (!worklist.insert({defOp, defOp->operand_begin()}).second) {
```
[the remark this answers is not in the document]
dobios: A bit of a subtlety of this verfier, the one way I usually get rid of dependency cycles is by pre-visiting the this that would support that, however here that isn't really necessary, as we just want to make sure that none of the children ops are clocks, so I removed that to avoid having errors thrown when the IR was valid.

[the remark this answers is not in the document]
dobios: basically dependency cycles aren't an issue here afaik

lib/Dialect/Verif/VerifOps.cpp
dobios: it's not invalid to have a block argument passed to an assertion, so the verifier shouldn't fail in that case, it should simply not have to check for a clock

dobios: unless failing in a canonicalizer doesn't fail the entire compilation?

uenoku: Ah maybe I was missing `()`. Does ` op.getProperty().getDefiningOp<ltl::ClockOp>()` work?

uenoku: > it's not invalid to have a block argument passed to an assertion, so we shouldn't fail in that case, it should simply not have to check for a clock

`op.getProperty().getDefiningOp<ltl::ClockOp>()` is equivalent to `isa_and_nonnull<ltl::ClockOp>(op.getProperty().getDefiningOp())` so it works fine even when property is an argument

dobios: no it doesn't

fabianschuiki (approved):
It's great to see `ltl.disable` go 🥳 Thanks for going through all this effort!

The logic of the changes looks good to me!

lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
dobios: It seems that generic intrinsics aren't mentioned in `FIRRTLIntrinsics.md`, what did you want me to update?

lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
dobios: I gave it the same signature as `hasNParams`

lib/Dialect/FIRRTL/FIRRTLIntrinsics.cpp
uenoku: Ah ok. I thought FIRRTL LTL intrinsics are documented there as well as other intrinsics. Then it’s fine not to add documents within the PR.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -3513,10 +3509,20 @@ class PropertyEmitter : public EmitterBase,
...
-void PropertyEmitter::emitProperty(
-    Value property, PropertyPrecedence parenthesizeIfLooserThan) {
+void PropertyEmitter::emitEnabledProperty(
+    Value property, Value enable, PropertyPrecedence parenthesizeIfLooserThan) {
   assert(localTokens.empty());
-  // Wrap to this column.
+
+  // If the property is tied to an enable, emit that.
+  if (enable) {
+    ps << "disable iff" << PP::nbsp << "(~(";
```
darthscsi, line 3518: This negation doesn't run into any `x` or `z` issues?

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4886,8 +4882,8 @@ LogicalResult StmtEmitter::visitSV(CoverConcurrentOp op) {
 
 /// Emit an assert-like operation from the `verif` dialect. This covers
 /// `verif.assert`, `verif.assume`, and `verif.cover`.
-LogicalResult StmtEmitter::emitVerifAssertLike(Operation *op, Value property,
-                                               PPExtString opName) {
+template <typename Op>
```
darthscsi, line 4885: Does the op type actually matter here?  other than getProperty, I don't think it does.  In which case, taking an Operation* will result in smaller code.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4926,22 +4924,22 @@ LogicalResult StmtEmitter::emitVerifAssertLike(Operation *op, Value property,
...
-  return emitVerifAssertLike(op, op.getProperty(), PPExtString("cover"));
+  return emitVerifAssertLike(op, PPExtString("cover"));
 }
 
 /// Emit an assert-like operation from the `verif` dialect. This covers
 /// `verif.clocked_assert`, `verif.clocked_assume`, and `verif.clocked_cover`.
-LogicalResult StmtEmitter::emitVerifClockedAssertLike(
-    Operation *op, Value property, Value disable, Value clock,
-    verif::ClockEdge edge, PPExtString opName) {
+template <typename Op>
```
darthscsi, line 4940: Again, there is a code-size tradeoff doing this instad of just taking the arguments.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -2033,6 +2035,7 @@ Value FIRRTLLowering::getPossiblyInoutLoweredValue(Value value) {
 /// This returns a null value for FIRRTL values that cannot be lowered, e.g.
 /// unknown width integers.
 Value FIRRTLLowering::getLoweredValue(Value value) {
+
```
darthscsi, line 2038: nit: stray white space

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1632,7 +1632,9 @@ struct FIRRTLLowering : public FIRRTLVisitor<FIRRTLLowering, LogicalResult> {
   LogicalResult visitExpr(LTLUntilIntrinsicOp op);
   LogicalResult visitExpr(LTLEventuallyIntrinsicOp op);
   LogicalResult visitExpr(LTLClockIntrinsicOp op);
-  LogicalResult visitExpr(LTLDisableIntrinsicOp op);
+
+  template <typename TargetOp, typename IntrinsicOp>
```
darthscsi, line 1636: minor: Maybe give this a different name?  giving it a similar name to the visitor-derived functions is slightly misleading.

darthscsi (approved):
Generally LGTM

lib/Conversion/ExportVerilog/ExportVerilog.cpp
uenoku: I forget to mention but could you remove extra parentheses?

include/circt/Dialect/LTL/LTLOps.td
uenoku: Please feel free to push unrelated change separately to main.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
dobios: I added them to properly scope the not, I was running into issues without it

lib/Conversion/ExportVerilog/ExportVerilog.cpp
dobios: not syntactically, but behaviorally maybe, I guess that if an X or Z is encounter, we don't really give any guarantees on the behavior anyway no?

uenoku (approved):
LGTM!

PR #7154, opened 2024-06-11
[Moore] Add evenOp to handle event controls.

hailongSun2000:
The `moore.wait_event`(EventOp) is aimed at handling event control, the delay control like #1ns will be implemented at another PR. And something like `@(posedge clk) a <= b + c;` will be handled by `Value visit(const slang::ast::AssignmentExpression &expr){...}`, so there are no error tests.

Thanks for the code review in advance! :smiley:

### DIFF
--- a/lib/Conversion/ImportVerilog/TimingControls.cpp
+++ b/lib/Conversion/ImportVerilog/TimingControls.cpp
@@ -1,0 +1,64 @@
+//===- TimingControl.cpp - Slang timing control conversion ----------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "ImportVerilogInternals.h"
+#include "slang/ast/TimingControl.h"
+using namespace circt;
+using namespace ImportVerilog;
+namespace {
+struct TimingCtrlVisitor {
+  Context &context;
+  Location loc;
+  OpBuilder &builder;
+
+  TimingCtrlVisitor(Context &context, Location loc)
+      : context(context), loc(loc), builder(context.builder) {}
+
+  LogicalResult visit(const slang::ast::SignalEventControl &ctrl) {
+    // TODO: When updating slang to the latest version, we will handle
+    // "iffCondition".
+    auto loc = context.convertLocation(ctrl.sourceRange.start());
+    auto input = context.convertRvalueExpression(ctrl.expr);
+    builder.create<moore::EventOp>(loc, static_cast<moore::Edge>(ctrl.edge),
+                                   input);
+    return success();
+  }
+
+  LogicalResult visit(const slang::ast::ImplicitEventControl &ctrl) {
+    return success();
+  }
+
+  LogicalResult visit(const slang::ast::EventListControl &ctrl) {
+    for (auto *event : ctrl.as<slang::ast::EventListControl>().events) {
+      if (failed(context.convertTimingControl(*event)))
+        return failure();
+    }
+    return success();
+  }
+
+  /// Emit an error for all other timing controls.
+  template <typename T>
+  LogicalResult visit(T &&node) {
+    mlir::emitError(loc, "unspported timing control: ")
+        << slang::ast::toString(node.kind);
+    return failure();
+  }
+
+  LogicalResult visitInvalid(const slang::ast::TimingControl &ctrl) {
+    mlir::emitError(loc, "invalid timing control");
+    return failure();
+  }
+};
+} // namespace
+
+LogicalResult
+Context::convertTimingControl(const slang::ast::TimingControl &timingControl) {
+  auto loc = convertLocation(timingControl.sourceRange.start());
+  TimingCtrlVisitor visitor{*this, loc};
+  return timingControl.visit(visitor);
+}
--- a/test/Conversion/ImportVerilog/basic.sv
+++ b/test/Conversion/ImportVerilog/basic.sv
@@ -1161,3 +1161,50 @@ module MultiPorts(
   // CHECK: [[C1_READ:%.+]] = moore.read %c1
   // CHECK: moore.output [[V1_READ]], [[C1_READ]]
 endmodule
+
+// CHECK-LABEL: moore.module @EventControl(in %clk : !moore.l1)
+module EventControl(input clk);
+  // CHECK: %clk_0 = moore.net name "clk" wire : <l1>
+
+  int a1, a2, b, c;
+
+  // CHECK: moore.procedure always
+  // CHECK:   [[CLK_READ:%.+]] = moore.read %clk_0 : l1
+  // CHECK:   moore.wait_event posedge [[CLK_READ]] : l1
+  always @(posedge clk) begin end;
+
+  // CHECK: moore.procedure always
+  // CHECK:   [[CLK_READ:%.+]] = moore.read %clk_0 : l1
+  // CHECK:   moore.wait_event negedge [[CLK_READ]] : l1
+  always @(negedge clk) begin end;
+
+  // CHECK: moore.procedure always
+  // CHECK:   [[CLK_READ:%.+]] = moore.read %clk_0 : l1
+  // CHECK:   moore.wait_event edge [[CLK_READ]] : l1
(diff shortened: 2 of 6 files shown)


lib/Conversion/ImportVerilog/TimingControls.cpp
```diff
@@ -0,0 +1,62 @@
...
+    return success();
+  }
+
+  LogicalResult visit(const slang::ast::EventListControl &ctrl) {
+    for (auto *event : ctrl.as<slang::ast::EventListControl>().events) {
+      if (failed(context.convertTimingControl(*event)))
+        return failure();
+    }
+    return success();
+  }
```
fabianschuiki, line 40: Could we merge this into `Statements.cpp` into the statement visitor? If I remember correctly, event and delay control are always associated with a statement, so it would make sense to just have that there.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -1161,3 +1161,50 @@ module MultiPorts(
...
+
+  // CHECK: moore.procedure always
+  // CHECK:   [[CLK_READ:%.+]] = moore.read %clk_0 : l1
+  // CHECK:   moore.wait_event negedge [[CLK_READ]] : l1
+  always @(negedge clk) begin end;
+
+  // CHECK: moore.procedure always
+  // CHECK:   [[CLK_READ:%.+]] = moore.read %clk_0 : l1
+  // CHECK:   moore.wait_event edge [[CLK_READ]] : l1
+  always @(edge clk) begin end;
```
fabianschuiki, line 1184: Very cool! Can you add a test for something like `always @(posedge x, negedge y iff z)`? If we don't support those yet -- which would be totally fine -- we should at least emit an error message. I think this might be the case already, but it would be great to check that.

fabianschuiki (approved):
LGTM!

test/Conversion/ImportVerilog/basic.sv
hailongSun2000: I notice the `iff`(conditional event controls) is supported at slang v0.6, but nothing in the slang AST. For example:
```
module Top(input x, input y);
  logic cond;
  reg r;
  always @(posedge x, negedge y iff cond) begin
          r <= 1;
  end
endmodule
```
```
 {
              "name": "",
              "kind": "ProceduralBlock",
              "addr": 2199025173672,
              "procedureKind": "Always",
              "body": {
                "kind": "Timed",
                "timing": {
                  "kind": "EventList",
                  "events": [
                    {
                      "kind": "SignalEvent",
                      "expr": {
                        "kind": "NamedValue",
                        "type": "logic",
                        "symbol": "2199025174120 x"
                      },
                      "edge": "PosEdge"
                    },
                    {
                      "kind": "SignalEvent",
                      "expr": {
                        "kind": "NamedValue",
                        "type": "logic",
                        "symbol": "2199025174696 y"
                      },
                      "edge": "NegEdge"
                    }
                  ]
                },
                "stmt": {
                  "kind": "Block",
                  "blockKind": "Sequential",
                  "body": {
                    "expr": {
                      "kind": "Assignment",
                      "type": "reg",
                      "left": {
                        "kind": "NamedValue",
                        "type": "reg",
                        "symbol": "2199025173296 r"
                      },
                      ...
```

lib/Conversion/ImportVerilog/TimingControls.cpp
hailongSun2000: Yeah, you're right! However, due to slang separating `timing controls` and `statements` into different classes, the former belongs to `slang::ast::TimingControl`, and the latter is `slang::ast::Statement`. Sure, the `TimedStatement` inherits from `Statement`, but we need to go via `stmt.timing` to visit the `timing control list`

test/Conversion/ImportVerilog/basic.sv
likeamahoney: There is a bug in the `iff` condition processing, and there is no printing support for it in `slang` before the patch that was introduced after the 6th release.

Currently AST for your example looks like this:

```json
{
  "design": {
    "name": "$root",
    "kind": "Root",
    "addr": 4792547490880,
    "members": [
      {
        "name": "",
        "kind": "CompilationUnit",
        "addr": 4792546809056
      },
      {
        "name": "Top",
        "kind": "Instance",
        "addr": 4792546810408,
        "body": {
          "name": "Top",
          "kind": "InstanceBody",
          "addr": 4792546809352,
          "members": [
            {
              "name": "x",
              "kind": "Port",
              "addr": 4792546810536,
              "type": "logic",
              "direction": "In",
              "internalSymbol": "4792546810664 x"
            },
            {
              "name": "x",
              "kind": "Net",
              "addr": 4792546810664,
              "type": "logic",
              "netType": {
                "name": "wire",
                "kind": "NetType",
                "addr": 4792547488640,
                "type": "logic"
              }
            },
            {
              "name": "y",
              "kind": "Port",
              "addr": 4792546811032,
              "type": "logic",
              "direction": "In",
              "internalSymbol": "4792546811160 y"
            },
            {
              "name": "y",
              "kind": "Net",
              "addr": 4792546811160,
              "type": "logic",
              "netType": {
                "name": "wire",
                "kind": "NetType",
                "addr": 4792547488640,
                "type": "logic"
              }
            },
            {
              "name": "cond",
              "kind": "Variable",
              "addr": 4792546809608,
              "type": "logic",
              "lifetime": "Static"
            },
            {
              "name": "r",
              "kind": "Variable",
              "addr": 4792546809952,
              "type": "reg",
              "lifetime": "Static"
            },
            {
              "name": "",
              "kind": "ProceduralBlock",
              "addr": 4792546810296,
              "procedureKind": "Always",
              "body": {
                "kind": "Timed",
                "timing": {
                  "kind": "EventList",
                  "events": [
                    {
                      "kind": "SignalEvent",
                      "expr": {
                        "kind": "NamedValue",
                        "type": "logic",
                        "symbol": "4792546810664 x"
                      },
                      "edge": "PosEdge"
                    },
                    {
                      "kind": "SignalEvent",
                      "expr": {
                        "kind": "NamedValue",
                        "type": "logic",
                        "symbol": "4792546811160 y"
                      },
                      "edge": "NegEdge",
                      "iff": {
                        "kind": "NamedValue",
                        "type": "logic",
                        "symbol": "4792546809608 cond"
                      }
                    }
                  ]
                },
                "stmt": {
                  "kind": "Block",
                  "blockKind": "Sequential",
                  "body": {
                    "expr": {
                      "kind": "Assignment",
                      "type": "reg",
                      "left": {
                        "kind": "NamedValue",
                        "type": "reg",
                        "symbol": "4792546809952 r"
                      },
                      "right": {
                        "kind": "Conversion",
                        "type": "reg",
                        "operand": {
                          "kind": "Conversion",
                          "type": "logic signed[31:0]",
                          "operand": {
                            "kind": "IntegerLiteral",
                            "type": "int",
                            "value": "1",
                            "constant": "1"
                          },
                          "constant": "1"
                        },
                        "constant": "1'b1"
                      },
                      "isNonBlocking": true
                    }
                  }
                }
              }
            }
          ],
          "definition": "4792547151360 Top"
        },
        "connections": [
        ]
      }
    ]
  },
  "definitions": [
    {
      "name": "Top",
      "kind": "Definition",
      "addr": 4792547151360,
      "defaultNetType": "4792547488640 wire",
      "definitionKind": "Module",
      "defaultLifetime": "Static",
      "unconnectedDrive": "None"
    }
  ]
}
```

likeamahoney: patch which fixed this is here - https://github.com/llvm/circt/pull/7154#discussion_r1637365910

hailongSun2000: oh~ Thanks! I used slang that I had installed locally. It's my mistake.

hailongSun2000: By the way, we use slang v3.0 as the CIRCT library at present. :sweat_smile:

fabianschuiki: 😅 Yeah, once we get a reasonably complete SV pipeline together, we can start to look into gradually bumping to newer versions of Slang. I tried going to v4 at some point, but the required C++ version was too high compared to the CI runners that CIRCT CI ran on. It's probably going to be easier to work against v3 until we have something that we're happy with, and then make up for the difference to v4/5/6 after that.

fabianschuiki (approved):
LGTM!

PR #7155, opened 2024-06-11
[ESI] Ensure esi-cosim.py ends in build/bin folder

mortbopet:
Previously, `esi-cosim.py` would only get copied when CIRCT was being installed, or the integration test targets were being built. This change makes it so that, if the ESI runtime is enabled, `esi-cosim.py` will always get copied.


teqdruid:
This was obviated by #7238

PR #7157, opened 2024-06-11
[firtool] Remove `LTLToCore` pass from `verification-falvor=immediate` pipeline

dobios:
The LTLToCore pass uses a flawed disable lowering, which I am working on fixing. In the meantime, we should remove the pass from the immediate lowering.


fabianschuiki (approved):
LGTM!

PR #7158, opened 2024-06-12
[ImportVerilog] [SROA] [Mem2Reg][Canonicalizers]Support Passes for Nested Type

mingzheTerapines:
For the new concept of Moore dialect, some operations will be defined as memory-related operations. Modeling `memref` Dialect and `LLVM` dialect, the operation relationship is as follows:
- `ReadOp` and `blockingAssignOp` are related to `loadOp` and `storeOp`.
- `VariableOp` is related to `allocaOp`.

However, the operations mentioned below are for basic types. This PR will support nested types in the following way:
- `VariableOp` with nested types is still related to `allocaOp` (will be replaced with `structCreateOp` and `UnionCreateOp`).
- `structExtractRefOp` is related to `storeOp`.
- `structExtractOp` is related to `loadOp`.

To implement this:
- Since these operations will be lowered to the `hw` dialect, the design largely refers to the `hw` dialect.

- Add the trait `DestructurableAllocationOpInterface` for `VariableOp`.
- Add the trait `DestructurableAccessorOpInterface` for `structExtractOp` and `structExtractRefOp`.
- Implement the `DestructurableTypeInterface` for `structLikeType` and the `reftype` of `structLikeType`.

For local variables:
- Use the SROA (Scalar Replacement of Aggregates) Pass to destructure all nested-type variables into basic-type variables.
- Use the Mem2Reg (Memory to Register) Pass to replace variables imported by SROA with constants.

For global/module-level variables:
- When importing Verilog, use `structInjectOp` rather than `blockingAssignOp`, because `structExtractRefOp` has the Destructurable trait, but global variables should not be destructured.
- `structInjectOp` means creating a new struct with new values and other old values.
- Use the canonicalizer Pass to fold duplicate injecting same field operations.
- Use the canonicalizer Pass to explicitly show new struct creation.
- Use the canonicalizer Pass to send source values directly and remove `structExtractOp`.

Also, remove some unnecessary spaces in other code.

What's more:
- Verify that the input of nested-type-related operations should match the field type defined.

To do:
- Update the use of struct SSA values referring to the latest `structInjectOp` SSA values.
- Design the method for union types.
- Add and support the `dbg` dialect to keep local variables visible after SROA & Mem2Reg.


mingzheTerapines:
@hailongSun2000 Could you give me some advice? I will appreciate it.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -885,7 +885,11 @@ def ExtractRefOp : MooreOp<"extract_ref"> {
   }];
 }
 
-def StructCreateOp : MooreOp<"struct_create", [SameOperandsAndResultType]> {
+def StructCreateOp : MooreOp<"struct_create", [
+  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>,
+  OptionalTypesMatchWith<"initial value and variable types match",
+    "result", "initial", "cast<RefType>($_self).getNestedType()">
+]> {
```
hailongSun2000, line 892: #7039 <-- At this PR, @fabianschuiki mentioned `DestructurableAllocationOpInterface` and `DestructurableAccessorOpInterface` interfaces. So I think maybe you can flag `moore.struct_create` with the former and the `moore.struct_extract` and `moore.struct_extract_ref` with the latter.

hailongSun2000: Search `MemorySlotInterfaces.td` to view the details.

mingzheTerapines: Thanks for reminding, I will check it.

fabianschuiki:
I think you probably want to have the `DestructurableAllocationOpInterface` implemented on `moore.variable` for the case where the variable has a struct type. But I would keep `moore.struct_create` as a simple op that just returns a struct value.

It generally makes life easier in the future if ops do only one thing, and you then compose multiple ops to do more complicated things. Here for example, you'd want `moore.struct_create` to form a `struct`-typed SSA value, and `moore.variable` if you want to create a variable of type `T` and get a `ref<T>` back. Then, if you want a struct reference, you'd create a `moore.variable` of type `struct` and get a `ref<struct>` back. You can then user `moore.struct_extract_ref` on that `ref<struct>` you got back from the variable -- which means that the variable probably has to know about struct types and has to be a _destructurable allocation_.

mingzheTerapines:
@fabianschuiki I was wondering use structtype to limit result type of createOp. But I find hard because the result could be either of packed or unpacked. So I have to use unpacked type. It seems thoses structlike types are useless. I think it will be better if we have :
struct unpackedstruct -> anyStruct unpackedstruct packedstruct
union unpackedunion -> anyUnion unpackedunion packedunion

fabianschuiki:
> @fabianschuiki I was wondering use structtype to limit result type of createOp. But I find hard because the result could be either of packed or unpacked.

Yeah you are right, that is indeed very annoying. I think you could create a type constraint like `def AnyStructType : AnyOf<[PackedStructType, UnpackedStructType]>` (or similar, I don't remember the precise syntax). But it would still be difficult to work with the struct types... I'm wondering if there might be a way to do this with LLVM's casting infrastructure: the struct types inherit from `PackedType` and `UnpackedType`, but it might be possible to make them additionally inherit from `StructLikeType` (which we could define separately).

Alternatively we could think about whether we actually use the PackedType and UnpackedType base classes often... If not, we might be able to change the type hierarchy such that we don't primarily distinguish between packed/unpacked.

mingzheTerapines:
> I think you probably want to have the `DestructurableAllocationOpInterface` implemented on `moore.variable` for the case where the variable has a struct type. But I would keep `moore.struct_create` as a simple op that just returns a struct value.
> 
> It generally makes life easier in the future if ops do only one thing, and you then compose multiple ops to do more complicated things. Here for example, you'd want `moore.struct_create` to form a `struct`-typed SSA value, and `moore.variable` if you want to create a variable of type `T` and get a `ref<T>` back. Then, if you want a struct reference, you'd create a `moore.variable` of type `struct` and get a `ref<struct>` back. You can then user `moore.struct_extract_ref` on that `ref<struct>` you got back from the variable -- which means that the variable probably has to know about struct types and has to be a _destructurable allocation_.

Thanks @fabianschuiki . It is very professional design. As I understanding HWAggregates Op designing, these kind of ops are all with Pure trait which means memory operations are with other Ops such as hw.wire or something.
Maybe I can also add  Pure trait to moore.union/sturct Ops. 
And thanks @hailongSun2000 's message, I was suggested to do memory-allocation need varify in moore.VarialbeOp, with trait DestructurableAllocationOpInterface.
I am also hoping to hear your suggestion @fabianschuiki . Thanks again!

mingzheTerapines:
@fabianschuiki  What's more,  
1.I can not find union_inject op and assigning some value to struct/union variable example in hw Dialect. Why is that?
2.can transform like this proper?
systemverilog code
```
  module top;
  int a;
  struct packed {
    int a, b;
  } ii;
  initial begin
    ii.a = a;
    a =ii.b;
   end
endmodule
```
After importverilog pass, variable has trait DestructurableAllocationOpInterface
```
module {
  moore.module @top() {
  %x = moore.variable : <i32>;
  %ii = moore.variable : !moore.packed<struct<{a: i32, b: i32}>>
  moore.procedure initial {
    %0 = moore.read %x :<i32>
    %1 = moore.struct_inject %ii, "a", %0, <struct<{a: i32, b: i32}>>
    moore.blocking_assign %ii, %1 : <struct<{a: i32, b: i32}>> -> <i32>

    %2 = moore.read %ii : <struct<{a: i32, b: i32}>>
    %3 = moore.struct_extract %2, "b" : struct<{a: i32, b: i32}> -> i32
    moore.blocking_assign %x, %3 : <struct<{a: i32, b: i32}>> -> <i32>
  }
  moore.output}
}
```

After conversion pass
```
module {
  moore.module @top() {
  %x = moore.variable : <i32>
  %c0_i32 = moore.constant 0 : !moore.i32
  %ii = moore.struct_create(%c0_i32 , %%c0_i32) :  <struct<{a: i32, b: i32}>>

  moore.procedure initial {
    %1 = moore.read %x :<i32>
    %2 = moore.read %ii : <struct<{a: i32, b: i32}>>
    %3 = moore.struct_inject %2, "a", %1, <struct<{a: i32, b: i32}>>
    moore.blocking_assign %ii, %3 : <struct<{a: i32, b: i32}>> -> <i32>


    %3 = moore.read %ii : <struct<{a: i32, b: i32}>>
    %4 = moore.struct_extract %3, "b" : struct<{a: i32, b: i32}> -> i32
    moore.blocking_assign %x, %4 : <struct<{a: i32, b: i32}>> -> <i32>
    }
  moore.output
  }
}
```
MooreToCore Pass
```
module {
  hw.module @top() {
  %x = hw.wire : @x <i32>
  %c0_i32 = = hw.wire : @c0_i32 <i32>
  %ii = hw.struct_create(%c0_i32 , %%c0_i32) :  !hw.struct<a: i32, y: i32>

  llhd.proc @initial () {
    %1 = moore.read %x :<i32>  //to do
    %2 = moore.read %ii : <struct<{a: i32, b: i32}>>  //to do
    %3 = hw.struct_inject %2, "a", %1, !hw.struct<a: i32, y: i32>
    moore.blocking_assign %ii, %3 : !hw.struct<a: i32, y: i32> -> <i32> //to do


    %3 = moore.read %ii : <struct<{a: i32, b: i32}>>  //to do
    %4 = hw.struct_extract %3, "b" : !hw.struct<a: i32, y: i32> -> i32
    moore.blocking_assign %x, %4 : <struct<{a: i32, b: i32}>> -> <i32>  //to do
    }
  hw.output
  }
}
```

hailongSun2000:
I think except `moore. varaible` shoude display `<T>`, others are `T`.

hailongSun2000:
> After conversion pass
module {
  moore.module @top() {
  %x = moore.variable : <i32>
  %c0_i32 = moore.constant 0 : !moore.i32
  %ii = moore.struct_create(%c0_i32 , %%c0_i32) :  <struct<{a: i32, b: i32}>>
  moore.procedure initial {
    %1 = moore.read %x :<i32>
    %2 = moore.read %ii : <struct<{a: i32, b: i32}>>
    %3 = moore.struct_inject %2, "a", %1, <struct<{a: i32, b: i32}>>
    moore.blocking_assign %ii, %3 : <struct<{a: i32, b: i32}>> -> <i32>
    %3 = moore.read %ii : <struct<{a: i32, b: i32}>>
    %4 = moore.struct_extract %3, "b" : struct<{a: i32, b: i32}> -> i32
    moore.blocking_assign %x, %4 : <struct<{a: i32, b: i32}>> -> <i32>
    }
  moore.output
  }
}

If `%ii` doesn't have an initial value, maybe we can lower it into ` %ii = moore.struct_create :  <struct<{a: i32, b: i32}>>`
For
 ```
  %2 = moore.read %ii : <struct<{a: i32, b: i32}>>  
  %3 = moore.struct_inject %2, "a", %1, <struct<{a: i32, b: i32}>> 
```
Maybe we can keep the form the same as before the conversion. Like `%1 = moore.struct_inject %ii, "a", %0, struct<{a: i32, b: i32}>`

**Comment by mingzheTerapines** [comment]
> > After conversion pass
> > module {
> > moore.module @top() {
> > %x = moore.variable : 
> > %c0_i32 = moore.constant 0 : !moore.i32
> > %ii = moore.struct_create(%c0_i32 , %%c0_i32) :  <struct<{a: i32, b: i32}>>
> > moore.procedure initial {
> > %1 = moore.read %x :
> > %2 = moore.read %ii : <struct<{a: i32, b: i32}>>
> > %3 = moore.struct_inject %2, "a", %1, <struct<{a: i32, b: i32}>>
> > moore.blocking_assign %ii, %3 : <struct<{a: i32, b: i32}>> -> 
> > %3 = moore.read %ii : <struct<{a: i32, b: i32}>>
> > %4 = moore.struct_extract %3, "b" : struct<{a: i32, b: i32}> -> i32
> > moore.blocking_assign %x, %4 : <struct<{a: i32, b: i32}>> -> 
> > }
> > moore.output
> > }
> > }
> 
> If `%ii` doesn't have an initial value, maybe we can lower it into ` %ii = moore.struct_create : <struct<{a: i32, b: i32}>>` For
> 
> ```
>  %2 = moore.read %ii : <struct<{a: i32, b: i32}>>  
>  %3 = moore.struct_inject %2, "a", %1, <struct<{a: i32, b: i32}>> 
> ```
> 
> Maybe we can keep the form the same as before the conversion. Like `%1 = moore.struct_inject %ii, "a", %0, struct<{a: i32, b: i32}>`

Thanks I got it.

**Comment by hailongSun2000** [comment]
And I think we can lower `moore.read` into `llhd.prb`, meanwhile, we need to lower `moore.variable` into `llhd.sig`.
I'm not sure whether we exactly need something like `moore.blocking_assign %ii, %1 : <!moore.packed<struct<{a: i32, b: i32}>>> -> <i32>`. I'm also curious why `hw.struct_inject` has the trait of `pure`. I think `hw.struct_inject` will modify the value of its member variables. Maybe I have misunderstood something.

**Comment by fabianschuiki** [comment]
> I'm also curious why `hw.struct_inject` has the trait of `pure`. I think `hw.struct_inject` will modify the value of its member variables. Maybe I have misunderstood something.

`hw.struct_inject` can be pure because it does not _technically_ modify the fields of its input struct, but instead it copies the input struct, modifies a field, and then returns the new struct. So roughly like this:
```mlir
%oldStruct = hw.struct_create("x": %a, "y": %b)
%newStruct = hw.struct_inject %oldStruct, "y", %c
// oldStruct remains unchanged, newStruct looks like this:
// %newStruct = hw.struct_create("x": %a, "y": %c)
```
And I agree with you: we should add the `Pure` trait to `moore.struct_inject` as well (and the union modification ops), since these never mutate the input struct (and they also can't operate on a `ref<struct>` type). If you wanted to "inject" a value into a struct variable, you'd use `moore.struct_extract_ref` to get a ref to the field, and then use one of the `moore.*assign` ops to assign to that field ref :smiley:

mingzheTerapines:
@fabianschuiki But hw dialect does not have the op like union_inject, which is weired.

fabianschuiki:
That is probably because unions don't really have fields that you want to change individually. The fields in a union are just different ways to look at the exact same data. Assigning a union's field is the same thing as replacing the union with a new union which has this field set. So instead of `union_inject` you'd just do `union_create` again 🙂

hailongSun2000:
> And I agree with you: we should add the `Pure` trait to `moore.struct_inject` as well (and the union modification ops), since these never mutate the input struct (and they also can't operate on a `ref<struct>` type). If you wanted to "inject" a value into a struct variable, you'd use `moore.struct_extract_ref` to get a ref to the field, and then use one of the `moore.*assign` ops to assign to that field ref 😃

Thanks!

mingzheTerapines:
Thanks @hailongSun2000  @fabianschuiki 
Final version of transform will be like this
```
  module top;
  int a;
  struct packed {
    int a, b;
  } ii;
  initial begin
    ii.a = a;
    a =ii.b;
   end
endmodule
```
After importverilog pass, variable has trait DestructurableAllocationOpInterface
```
module {
  moore.module @top() {
  %x = moore.variable : <i32>;
  %ii = moore.variable : !moore.packed<struct<{a: i32, b: i32}>>
  moore.procedure initial {
    %0 = moore.read %x :<i32>
    %1 = moore.struct_extract_ref %ii, "a", %0, <struct<{a: i32, b: i32}>>
    moore.blocking_assign %ii, %1 : <struct<{a: i32, b: i32}>> -> <i32>

    %2 = moore.read %ii : <struct<{a: i32, b: i32}>>
    %3 = moore.struct_extract %2, "b" : struct<{a: i32, b: i32}> -> i32
    moore.blocking_assign %x, %3 : <struct<{a: i32, b: i32}>> -> <i32>
  }
  moore.output}
}
```

After conversion pass
```
module {
  moore.module @top() {
  %x = moore.variable : <i32>
  %ii = moore.struct_create() :  <struct<{a: i32, b: i32}>>

  moore.procedure initial {
    %0 = moore.read %x :<i32>
    %1 = moore.struct_extract_ref %ii, "a", %0, <struct<{a: i32, b: i32}>>
    %2 = moore.struct_inject %1, "a", %0, <struct<{a: i32, b: i32}>>
    moore.blocking_assign %1, %2 : <struct<{a: i32, b: i32}>> -> <i32>


    %3 = moore.read %ii : <struct<{a: i32, b: i32}>>
    %4 = moore.struct_extract %3, "b" : struct<{a: i32, b: i32}> -> i32
    moore.blocking_assign %x, %4 : <struct<{a: i32, b: i32}>> -> <i32>
    }
  moore.output
  }
}
```
MooreToCore Pass
```
module {
  hw.module @top() {
  %x = llhd.sig : @x <i32>

  %ii = hw.struct_create() :  !hw.struct<a: i32, y: i32>

  llhd.proc @initial () {
    %1 = llhd.prb %x :<i32>
    %2 =llhd.prb %ii : <struct<{a: i32, b: i32}>>
    %3 = hw.struct_inject %2, "a", %1, !hw.struct<a: i32, y: i32>
    moore.blocking_assign %ii, %3 : !hw.struct<a: i32, y: i32> -> <i32> //to do some memory store op


    %3 = moore.read %ii : <struct<{a: i32, b: i32}>>  //to do
    %4 = hw.struct_extract %3, "b" : !hw.struct<a: i32, y: i32> -> i32
    moore.blocking_assign %x, %4 : <struct<{a: i32, b: i32}>> -> <i32>  //to do some memory store op
    }
  hw.output
  }
}
```

fabianschuiki:
Cool! In your example MLIR after ImportVerilog you have this:
> ```
>     %0 = moore.read %x :<i32>
>     %1 = moore.struct_extract_ref %ii, "a", %0, <struct<{a: i32, b: i32}>>
>     moore.blocking_assign %ii, %1 : <struct<{a: i32, b: i32}>> -> <i32>
> ```
The `struct_extract_ref` operates on a ref type, so you'll want to pass in the variable `%x` directly (the ref) instead of its value (the read):
```
    %1 = moore.struct_extract_ref %ii, "a", %x, <struct<{a: i32, b: i32}>>
    moore.blocking_assign %ii, %1 : <struct<{a: i32, b: i32}>> -> <i32>
```

mingzheTerapines:
@fabianschuiki sorry I have fixed it

hailongSun2000:
Hey, @mingzheTerapines. For `assign`, `blocking_assign`, and `nonblocking_assign`, their arguments-field all are `ref<T>` and `T`. Therefore, I think the LHS has a ref type, but the RHS does not. And as discussed before at this(7039) PR. We will transform `variable` with struct type into `struct_create`, and (`struct_extract` and `assign`) into `struct_inject`. So from my perspective, I think the following code segment should be changed:
from
### moore IR:
```
    %0 = moore.read %x :<i32>
    %1 = moore.struct_extract_ref %ii, "a", %0, <struct<{a: i32, b: i32}>>
    moore.blocking_assign %ii, %1 : <struct<{a: i32, b: i32}>> -> <i32>
```
to
```
  %0 = moore.read %x : i32
  %1 = moore.struct_extract_ref %ii, "a" : <struct<{a : i32, b : i32}>> -> <i32>
  moore.blocking_assign %1, %0 : i32
```
And then after performing a MooreTransform Pass:
```
   %0 = moore.read %x : i32
   %1 = moore.struct_inject %ii, "a", %0 : <struct<{a : i32, b : i32}>>, i32 -> i32
```
nit: you should be careful with the `assemblyFormat` with different ops. Especially types of printing.

mingzheTerapines:
@hailongSun2000 You are right, now I have improved it.

mingzheTerapines:
@fabianschuiki @hailongSun2000 Seems this PR has already apply the importVerilog pass. Mem2reg and Moore2Core passes can be applied in other PR.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -970,11 +972,11 @@ def StructCreateOp : MooreOp<"struct_create", [SameOperandsAndResultType]> {
     ```
     See IEEE 1800-2017 § 7.2. "Structures".
   }];
-  let arguments = (ins UnpackedType:$input);
-  let results = (outs UnpackedType:$result);
-  let hasCustomAssemblyFormat = 1;
+  let arguments = (ins StrAttr:$name,  Variadic<UnpackedType>:$input);
```
fabianschuiki, line 975: I don't think `struct_create` should have a name. The op creates a new struct SSA value, but that doesn't necessarily mean it has a corresponding Verilog name. The name is more likely to be attached to a `moore.variable`. For the `struct_create` it's okay to just have `(ins Variadic<UnpackedType>:$fields)`.

The `OpAsmOpInterface` then also shouldn't be needed.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -305,6 +305,8 @@ def UnpackedStructType : StructLikeType<
   }];
 }
 
+def AnyStructType : AnyTypeOf<[StructType, UnpackedStructType]>;
+
```
fabianschuiki, line 309: Very nice! 😃

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -970,11 +972,11 @@ def StructCreateOp : MooreOp<"struct_create", [SameOperandsAndResultType]> {
...
+  let results = (outs AnyStructType:$result);
   let assemblyFormat = [{
-  $input attr-dict `:` type($input)
+    `` custom<ImplicitSSAName>($name) $input attr-dict
+    `:` type($input) `->` type($result)
   }];
```
fabianschuiki, line 980: Can you add a constraint like `TypesMatchWith<...>` or a custom verifier that checks if the types of the `input`s are exactly equal to the types of the result struct fields? Otherwise this would allow you to do something like `moore.struct_create %a, %b : i42, string -> struct<{a: event, b: l9001}>`.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1006,12 +1008,21 @@ def StructExtractOp : MooreOp<"struct_extract"> {
...
-  let arguments = (ins StrAttr:$memberName, UnpackedType:$input);
+  let arguments = (ins StrAttr:$fieldName, UnpackedType:$input);
   let results = (outs UnpackedType:$result);
   let assemblyFormat = [{
-  $input `,`   $memberName  attr-dict `:`
-  type($input) `->`
-  type($result)
+  $input `,`   $fieldName  attr-dict `:`
+  type($input) `->` type($result)
+  }];
```
fabianschuiki, line 1016: Can you add a constraint like `TypesMatchWith<...>` that ensures that the type of `$result` matches the type of the accessed field in `$input`? This will then allow you to remove the `-> type($result)` in the assembly format, since the constraint allows MLIR to determine the type automatically.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1006,12 +1008,21 @@ def StructExtractOp : MooreOp<"struct_extract"> {
     ```
     See IEEE 1800-2017 § 7.2.1 "Assigning to structures".
   }];
-  let arguments = (ins StrAttr:$memberName, UnpackedType:$input);
+  let arguments = (ins StrAttr:$fieldName, UnpackedType:$input);
```
fabianschuiki, line 1011: Can you also make the `$input` an `AnyStructType`?

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1006,12 +1008,21 @@ def StructExtractOp : MooreOp<"struct_extract"> {
...
-  type($result)
+  $input `,`   $fieldName  attr-dict `:`
+  type($input) `->` type($result)
+  }];
+}
+
+def StructExtractRefOp : MooreOp<"struct_extract_ref"> {
+  let summary = "Struct Extract operation";
+  let arguments = (ins StrAttr:$fieldName, RefType:$input);
+  let results = (outs RefType:$result);
```
fabianschuiki, line 1022: Can you make `$input` an `AnyStructRefType`, and can you add a constraint like `TypesMatchWith<...>` that ensures the type of `$result` is the type of the accessed field in `$input` wrapped in a `ref<...>`? This will then allow you to remove the `-> type($result)` in the assembly format, since the constraint allows MLIR to determine the type automatically.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1039,16 +1050,18 @@ def StructInjectOp : MooreOp<"struct_inject"> {
     ```
     See IEEE 1800-2017 § 7.2. "Assigning to structures".
   }];
-  let arguments = (ins UnpackedType:$LHS, StrAttr:$memberName,
-  UnpackedType:$RHS);
+  let arguments = (ins UnpackedType:$input, StrAttr:$fieldName,
+  UnpackedType:$newValue);
   let results = (outs UnpackedType:$result);
```
fabianschuiki, line 1055: Can you make `$input` and `$result` an `AnyStructType`? And maybe also add a `TypesMatch<...>` constraint between `$input` and `$result`. (The result struct has the same type as the input struct.) That again lets you remove the `-> type($result)` part of the assembly format.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1073,11 +1086,11 @@ def UnionCreateOp : MooreOp<"union_create"> {
...
   }];
-  let arguments = (ins StrAttr:$unionName, UnpackedType:$input);
-  let results = (outs UnpackedType:$result);
+  let arguments = (ins StrAttr:$name, UnpackedType:$input);
+  let results = (outs AnyUnionType:$result);
   let assemblyFormat = [{
-  $unionName `,` $input attr-dict `:`
-  type($input) `->` type($result)
+    `` custom<ImplicitSSAName>($name) $input attr-dict
+    `:` type($input) `->` type($result)
```
fabianschuiki, line 1093: Same as with `struct_create`: the op only creates an SSA union value in the IR, but that doesn't have a name itself. Names would be provided by a `moore.variable` op that uses that SSA union value as an initial value, or an assignment that stores this SSA value into a variable.

However, I agree that you _do_ need a name here: the name of the union variant that should be used to create the union value. For example, a union `union<a: i42, b: i9001>` has two variants, "a" and "b". When you construct such a union, you have to pick which one of the variants you are choosing. The options would be:
```mlir
moore.union_create %0, "a" : i42 -> union<a: i42, b: i9001>
moore.union_create %1, "b" : i9001 -> union<a: i42, b: i9001>
```
So depending on which `variantName` the operation picks, the operand has to have a type corresponding to that variant in the union type. ("a" needs i42, "b" needs i9001).

You can use `TypesMatchWith<...>` constraints again to ensure that the type of the operand `%0` or `%1` matches the type of the field `$variantName` in `$result`.

The `OpAsmOpInterface` is then no longer needed.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1104,17 +1117,27 @@ def UnionExtractOp : MooreOp<"union_extract"> {
...
+
+def UnionExtractRefOp : MooreOp<"union_extract_ref"> {
+  let summary = "Union Extract operation";
+  let arguments = (ins StrAttr:$fieldName, RefType:$input);
+  let results = (outs RefType:$result);
+  let assemblyFormat = [{
+  $input `,`   $fieldName  attr-dict `:`
   type($input) `->` type($result)
   }];
 }
```
fabianschuiki, line 1136: Same as for the struct case: it would be great to have `$input` be a `AnyUnionType` and `AnyUnionRefType`, and to have `TypesMatchWith<...>` constraints that check if the `$result` type is the same as the accessed variant/field of the `$input` union. (Wrapped in `ref<...>` for the `*_ref` op.)

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -663,6 +663,24 @@ struct LvalueExprVisitor {
...
+      return builder.create<moore::StructExtractRefOp>(
+          loc, moore::RefType::get(cast<moore::UnpackedType>(type)),
+          builder.getStringAttr(expr.member.name), value);
+    }
+    if (valueType->isPackedUnion() || valueType->isUnpackedUnion()) {
+      return builder.create<moore::UnionExtractRefOp>(
+          loc, moore::RefType::get(cast<moore::UnpackedType>(type)),
+          builder.getStringAttr(expr.member.name), value);
+    }
+    llvm_unreachable("unsupported symbol kind");
```
fabianschuiki, line 682: Can you turn this into an error diagnostic since there doesn't seem to be a strict guarantee on Slang's side that it won't produce something beside `struct` and `union` types. (It's likely to produce these also for classes and mabye interfaces.)

Something like:
```c++
mlir::emitError(value.getLoc()) << "cannot access member";
return {};
```

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -943,7 +943,50 @@ module Expressions;
     // CHECK: [[TMP2:%.+]] = moore.add [[A_ADD]], [[TMP1]]
     // CHECK: moore.blocking_assign %a, [[TMP2]]
     a += (a *= a--);
- 
+
+    // CHECK: [[TMP1:%.+]] = moore.struct_extract_ref %struct0, "a" : <struct<{a: i32, b: i32}>> -> <i32>
+    // CHECK: [[TMP2:%.+]] = moore.read %a : i32
+    // CHECK: moore.blocking_assign [[TMP1]], [[TMP2]] : i32
+    struct0.a = a;
```
fabianschuiki, line 950: Very very nice! 🥳

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -943,7 +943,50 @@ module Expressions;
...
+    // CHECK: [[TMP2:%.+]] = moore.struct_extract_ref [[TMP1]], "a" : <struct<{a: i32, b: i32}>> -> <i32>
+    // CHECK: [[TMP3:%.+]] = moore.read %a : i32
+    // CHECK: moore.blocking_assign [[TMP2]], [[TMP3]] : i32
+    struct1.c.a = a;
+
+    // CHECK: [[TMP1:%.+]] = moore.read %struct1 : struct<{c: struct<{a: i32, b: i32}>, d: struct<{a: i32, b: i32}>}>
+    // CHECK: [[TMP2:%.+]] = moore.struct_extract [[TMP1]], "d" : struct<{c: struct<{a: i32, b: i32}>, d: struct<{a: i32, b: i32}>}> -> struct<{a: i32, b: i32}>
+    // CHECK: [[TMP3:%.+]] = moore.struct_extract [[TMP2]], "b" : struct<{a: i32, b: i32}> -> i32
+    // CHECK: moore.blocking_assign %b, [[TMP3]] : i32
+    b = struct1.d.b;
```
fabianschuiki, line 967: You probably don't need these nesting checks because they don't exercise any new code paths.

mingzheTerapines: Thanks for advice

lib/Conversion/ImportVerilog/Expressions.cpp
mingzheTerapines: You are certainly right. 
But "error: unsupported construct: ClassType" this error will emitted firstly. Seems I can not add related error examples in errors.sv =(

include/circt/Dialect/Moore/MooreOps.td
mingzheTerapines: It seems TypeMatchWith<> can not fit this case, I just wrote a verifier for union and struct. Please have a look, thanks a lot.

include/circt/Dialect/Moore/MooreOps.td
mingzheTerapines: Since we already have a verifier that verify the type of inputs, we do not need to have this exact type declartion. Types like anyStructType could be replaced to unpacked type.

include/circt/Dialect/Moore/MooreOps.td
mingzheTerapines: Since we already have a verifier that verify the type of inputs, we do not need to have this exact type declartion. Types like anyStructType could be replaced to unpacked type.

include/circt/Dialect/Moore/MooreOps.td
mingzheTerapines: I just add verifier for this case

include/circt/Dialect/Moore/MooreOps.td
mingzheTerapines: Thanks for the advice. I have just add a verifier, please check.

include/circt/Dialect/Moore/MooreOps.td
mingzheTerapines: As mentioned, I wrote a verifier to fix this.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -265,6 +266,64 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
...
+    // context.valueSymbols.insertIntoScope(context.valueSymbols.getCurScope(),&var,
+    // varOp);
+    inputs.push_back(varOp);
+    slotMap.try_emplace<MemorySlot>(usedIndex, {varOp, elemType});
+  }
+  builder.create<StructCreateOp>(getLoc(), getType(), inputs);
+
+  return slotMap;
+}
+
```
mingzheTerapines, line 318: @Moxinilian I just support DestructurableAllocationOpInterface to variableOp, but it seems no working for struct. Could you have a look, thanks a lot.

mingzheTerapines:
@hailongSun2000 Is this right?
```
  module Foo;
  int x, y, z;
  typedef struct packed {
    int a, b;
  } ab;
  always_comb begin
     ab ii;
   ii.a = 32'b1;
   ii.b = 32'b100;
    ii.a = x + 1;
    y = ii.a;
    ii.a = ii.a + 1;
    z = ii.a;
  end
endmodule
```

After importverilog pass

```
module {
  moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    moore.procedure always_comb {
      %ii = moore.variable : <struct<{a: i32, b: i32}>>
      %0 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %1 = moore.constant 1 : i32
      %2 = moore.conversion %1 : !moore.i32 -> !moore.i32
      moore.blocking_assign %0, %2 : i32
      %3 = moore.struct_extract_ref %ii, "b" : <struct<{a: i32, b: i32}>> -> <i32>
      %4 = moore.constant 4 : i32
      %5 = moore.conversion %4 : !moore.i32 -> !moore.i32
      moore.blocking_assign %3, %5 : i32
      %6 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %7 = moore.read %x : i32
      %8 = moore.constant 1 : i32
      %9 = moore.add %7, %8 : i32
      moore.blocking_assign %6, %9 : i32
      %10 = moore.read %ii : struct<{a: i32, b: i32}>
      %11 = moore.struct_extract %10, "a" : struct<{a: i32, b: i32}> -> i32
      moore.blocking_assign %y, %11 : i32
      %12 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %13 = moore.read %ii : struct<{a: i32, b: i32}>
      %14 = moore.struct_extract %13, "a" : struct<{a: i32, b: i32}> -> i32
      %15 = moore.constant 1 : i32
      %16 = moore.add %14, %15 : i32
      moore.blocking_assign %12, %16 : i32
      %17 = moore.read %ii : struct<{a: i32, b: i32}>
      %18 = moore.struct_extract %17, "a" : struct<{a: i32, b: i32}> -> i32
      moore.blocking_assign %z, %18 : i32
    }
    moore.output
  }
}
```
p
After Mem2Reg pass, added traits helps mem2reg

```
module {
  moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    moore.procedure always_comb {
       %0 = moore.read %x : i32
       %1 = moore.constant 1 : i32
       %2 = moore.constant 4 : i32
      //do we need moore.struct_create here?
       %3 = moore.add %0, %1 : i32
       moore.blocking_assign %y, %1 : i3
       %4 = moore.constant 1 : i32
       %5 = moore.add %4, %3 : i32
       moore.blocking_assign %z, %5 : i3
    }
    moore.output
  }
}
```

mingzheTerapines:
@Moxinilian Could you have a look? Thanks a lot =)

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -519,6 +818,43 @@ ReadOp::removeBlockingUses(const MemorySlot &slot,
...
+//         for (const auto &member : members) {
+//           auto index = StringAttr::get(getContext(), member.name);
+//           const auto &memorySlot = subslots.at(index);
+//           setOperand(memorySlot.ptr);
+//         }
+//         return DeletionKind::Keep;
+//       })
+//       .Default([]() { return DeletionKind::Keep; });
+// }
+
```
fabianschuiki, line 857: Commented out code slipped into the PR 🙂

mingzheTerapines: I was wonder do I need adding Trait DestructurableAccessorOpInterface to readOp.

mingzheTerapines:
@fabianschuiki Can we simplify code by replacing readOp for structType with structExtractOp while importVerilog?
before
```
 %ii = moore.variable : <struct<{a: i32, b: i32}>>
%0 = moore.constant 1 : i32
 %1 = moore.conversion %0 : !moore.i32 -> !moore.i32
%2 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
   moore.blocking_assign %2, %1 : i32
%3 = moore.read %ii : struct<{a: i32, b: i32}>
%4 = = moore.struct_extract %3, "a" : struct<{a: i32, b: i32}> -> i32
 moore.blocking_assign %z, %4 : i32
```
after

```
 %ii = moore.variable : <struct<{a: i32, b: i32}>>
%0 = moore.constant 1 : i32
 %1 = moore.conversion %0 : !moore.i32 -> !moore.i32
%2 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
   moore.blocking_assign %2, %1 : i32
%3  = moore.struct_extract %ii, "a" : struct<{a: i32, b: i32}> -> i32
 moore.blocking_assign %z, %3 : i32
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -970,15 +973,17 @@ def StructCreateOp : MooreOp<"struct_create", [SameOperandsAndResultType]> {
...
-  $input attr-dict `:` type($input)
+    $input attr-dict `:` type($input) `->` type($result)
   }];
+  let hasVerifier = 1;
 }
 
-def StructExtractOp : MooreOp<"struct_extract"> {
+def StructExtractOp : MooreOp<"struct_extract", [
+  DeclareOpInterfaceMethods<DestructurableAccessorOpInterface>
+  ]> {
```
Moxinilian, line 986:
```suggestion
]> {
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1006,16 +1011,28 @@ def StructExtractOp : MooreOp<"struct_extract"> {
...
     See IEEE 1800-2017 § 7.2.1 "Assigning to structures".
   }];
-  let arguments = (ins StrAttr:$memberName, UnpackedType:$input);
+  let arguments = (ins StrAttr:$fieldName, UnpackedType:$input);
   let results = (outs UnpackedType:$result);
   let assemblyFormat = [{
-  $input `,`   $memberName  attr-dict `:`
-  type($input) `->`
-  type($result)
+  $input `,`   $fieldName  attr-dict `:` type($input) `->` type($result)
```
Moxinilian, line 1017:
```suggestion
    $input `,` $fieldName attr-dict `:` type($input) `->` type($result)
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1006,16 +1011,28 @@ def StructExtractOp : MooreOp<"struct_extract"> {
...
+  }];
+  let hasVerifier = 1;
+}
+
+def StructExtractRefOp : MooreOp<"struct_extract_ref", [
+  DeclareOpInterfaceMethods<DestructurableAccessorOpInterface>
+  ]> {
```
Moxinilian, line 1024:
```suggestion
]> {
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1006,16 +1011,28 @@ def StructExtractOp : MooreOp<"struct_extract"> {
...
+  let summary = "Struct Extract operation";
+  let arguments = (ins StrAttr:$fieldName, RefType:$input);
+  let results = (outs RefType:$result);
+  let assemblyFormat = [{
+  $input `,`   $fieldName  attr-dict `:`
+  type($input) `->` type($result)
```
Moxinilian, line 1030:
```suggestion
    $input `,` $fieldName attr-dict `:`
    type($input) `->` type($result)
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1039,16 +1056,17 @@ def StructInjectOp : MooreOp<"struct_inject"> {
...
   }];
-  let arguments = (ins UnpackedType:$LHS, StrAttr:$memberName,
-  UnpackedType:$RHS);
+  let arguments = (ins UnpackedType:$input, StrAttr:$fieldName,
+  UnpackedType:$newValue);
   let results = (outs UnpackedType:$result);
   let assemblyFormat = [{
-    $LHS `,`   $memberName `,` $RHS
-    attr-dict `:` type($LHS) type($RHS) `->` type($result)
+    $input `,`   $fieldName `,` $newValue
```
Moxinilian, line 1063:
```suggestion
    $input `,` $fieldName `,` $newValue
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1104,17 +1122,29 @@ def UnionExtractOp : MooreOp<"union_extract"> {
     ```
     See IEEE 1800-2017 § 7.3.1 "Packed unions"
   }];
-  let arguments = (ins StrAttr:$memberName, UnpackedType:$input);
+  let arguments = (ins StrAttr:$fieldName, UnpackedType:$input);
   let results = (outs UnpackedType:$result);
   let assemblyFormat = [{
-  $input `,`   $memberName  attr-dict `:`
+  $input `,`   $fieldName  attr-dict `:`
+  type($input) `->` type($result)
```
Moxinilian, line 1129:
```suggestion
    $input `,`   $fieldName  attr-dict `:`
    type($input) `->` type($result)
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1104,17 +1122,29 @@ def UnionExtractOp : MooreOp<"union_extract"> {
...
+  let hasVerifier = 1;
+}
+
+def UnionExtractRefOp : MooreOp<"union_extract_ref"> {
+  let summary = "Union Extract operation";
+  let arguments = (ins StrAttr:$fieldName, RefType:$input);
+  let results = (outs RefType:$result);
+  let assemblyFormat = [{
+  $input `,`   $fieldName  attr-dict `:`
   type($input) `->` type($result)
```
Moxinilian, line 1140:
```suggestion
    $input `,`   $fieldName  attr-dict `:`
    type($input) `->` type($result)
```

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -284,6 +285,9 @@ class StructLikeType<
   let assemblyFormat = [{
     `<` custom<Members>($members) `>`
   }];
+   let extraClassDeclaration = [{
+   Attribute getMemorySpace();
```
Moxinilian, line 289:
```suggestion
  let extraClassDeclaration = [{
    Attribute getMemorySpace();
```

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -265,6 +266,61 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
   return std::nullopt;
 }
 
+SmallVector<DestructurableMemorySlot> VariableOp::getDestructurableSlots() {
+  auto refType = getType();
```
Moxinilian, line 270: By ref type here do you mean the type of the value you are referring or the type of the reference itself?

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -265,6 +266,61 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
...
+
+  DenseMap<Attribute, MemorySlot> slotMap;
+
+  auto destructurable = llvm::cast<DestructurableTypeInterface>(getType());
+  llvm::ArrayRef<StructLikeMember> members;
+  TypeSwitch<Type>(getType().getNestedType())
+      .Case<StructType, UnpackedStructType>(
+          [&members](auto &type) { members = type.getMembers(); })
+      .Default([this](auto &) {
+        emitOpError("Result type must be StructType or UnpackedStructType");
```
Moxinilian, line 296: Why? Could you not handle any destructurable type?

Moxinilian: If you cannot handle the other cases, you should either:
- Assert it if the verifier already ensures that the only destructurable types are structs.
- Not emit a destructurable slot in the first place if you cannot actually destructure it.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -401,6 +457,252 @@ LogicalResult ConcatRefOp::inferReturnTypes(
...
+            if (slot.ptr != getInput())
+              return false;
+            auto index = getFieldNameAttr();
+            if (!index)
+              return false;
+            usedIndices.insert(index);
+            return true;
+          })
+      .Default([](auto) { return false; });
+}
```
Moxinilian, line 530: I assume all struct field references are safely used by design, right? (from a reference to a field of a struct, one cannot access other fields of the same struct)

Maybe add a comment about this if it is correct.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -401,6 +457,252 @@ LogicalResult ConcatRefOp::inferReturnTypes(
...
+
+DeletionKind StructExtractOp::rewire(const DestructurableMemorySlot &slot,
+                                     DenseMap<Attribute, MemorySlot> &subslots,
+                                     OpBuilder &builder,
+                                     const DataLayout &dataLayout) {
+  return TypeSwitch<Type, DeletionKind>(getType())
+      .Case<StructType, UnpackedStructType>([this, &subslots](auto &type) {
+        auto index = getFieldNameAttr();
+        if (!index)
+          return DeletionKind::Keep;
```
Moxinilian, line 540: This has already been asserted when checking whether the operation can be rewired, no need to do it twice I think.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -401,6 +457,252 @@ LogicalResult ConcatRefOp::inferReturnTypes(
...
+          })
+      .Default([](auto) { return false; });
+}
+
+DeletionKind StructExtractOp::rewire(const DestructurableMemorySlot &slot,
+                                     DenseMap<Attribute, MemorySlot> &subslots,
+                                     OpBuilder &builder,
+                                     const DataLayout &dataLayout) {
+  return TypeSwitch<Type, DeletionKind>(getType())
+      .Case<StructType, UnpackedStructType>([this, &subslots](auto &type) {
```
Moxinilian, line 537: You probably do not need to type switch again here, as you do not appear to be using the type and the check has already been done in `canRewrite`. I realize this is not documented but `rewire` is indeed only called after `canRewire`, as it is fair to expect.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -401,6 +457,252 @@ LogicalResult ConcatRefOp::inferReturnTypes(
...
+        auto index = getFieldNameAttr();
+        if (!index)
+          return DeletionKind::Keep;
+        const auto &memorySlot = subslots.at(index);
+        setOperand(memorySlot.ptr);
+        return DeletionKind::Keep;
+      })
+      .Default([](auto) { return DeletionKind::Keep; });
+}
```
Moxinilian, line 607: Same comments as for ExtractOp.

Moxinilian (commented):
Neat! I'm not super familiar with the moore dialect so hopefully I understood its semantics right.

Are you testing the SROA feature anywhere?

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: Thanks again for so careful comments

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: Yes, the result of the variableOp is a value refering to this variable, and the refType is the type of this value.

mingzheTerapines: There is also a structCreateOp in Moore Dialect, I am not sure @fabianschuiki suggesting whether using structCreateOp replace variableOp

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: Other cases are basic value which will apply PromotableAllocationOpInterface trait.
Only nested types will apply DestructurableAllocationOpInterface trait.
As mentioned, I can be sure structCreatOp only handle for nested types. But @fabianschuiki suggested not.

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: Yes you are right, it is safe as far as I know

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: Sorry, I am using typeSwitch to do a type converting maybe it is unproper?

Moxinilian: But yet you appear to not use the result of the type cast, do you?

lib/Dialect/Moore/MooreOps.cpp
Moxinilian: Okay good. I was just asking because if it is not safe you need to add it to the safely used check so I was not sure if it was intended or forgotten.

lib/Dialect/Moore/MooreOps.cpp
Moxinilian: Okay let me phrase it another way: in what circumstances would this check fail? Would that be considered a problem with the user input or with the compiler logic?

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: You are certainlyright. That is useless judgement

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: That's my bad, I used it but abandoned. I will have a check.

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: I did notice memrefOp did some mutable-clear operation. I think the point is its Ops index several fields, but this Op only index one field.

lib/Dialect/Moore/MooreOps.cpp
mingzheTerapines: You are right, that should not emit an error.

mingzheTerapines:
@Moxinilian Sorry, I still met some problems. Mem2Reg is not calling getDestructurableSlots function, and I found no explict trait verifier for DestructurableAllocationOpInterface or DestructurableAccessorOpInterface in Mem2Reg. In what condition, Mem2Reg could call VariableOp::getDestructurableSlot?

lib/Dialect/Moore/MooreOps.cpp
Moxinilian: Yes, I was more thinking about LLVM-like use cases where pointers point to somewhere in the bit byte array that is memory. All good.

Moxinilian:
You need to invoke the SROA pass before Mem2Reg. Mem2Reg only handles promotion while SROA handles destructuring.

mingzheTerapines:
@Moxinilian Thanks, it works, and I am implementing DestructurableTypeInterface.

mingzheTerapines:
@fabianschuiki pass SROA has this method:only handling Op that used this destructable variable.
```
  // Initialize the analysis with the immediate users of the slot.
  for (OpOperand &use : slot.ptr.getUses()) {
    if (auto accessor =
            dyn_cast<DestructurableAccessorOpInterface>(use.getOwner())) {
      if (accessor.canRewire(slot, info.usedIndices, usedSafelyWorklist,
                             dataLayout)) {
        info.accessors.push_back(accessor);
        continue;
      }
    }
```
However, codes like this 
```
module {
  moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    %dd = moore.variable : <struct<{a: i32, b: i32}>>
    moore.procedure always_comb {
      %d = moore.variable : <i32>
      %3 = moore.struct_extract_ref %dd, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %4 = moore.read %d : i32
      moore.blocking_assign %3, %4 : i32
    }
    moore.output
  }
}
```
moore.blocking_assign does not directly used %dd, so destructable-related function was not called.
Yet I can not call destructable-related function at  moore.struct_extract_ref because I do not know which value should assign to this pointer.
One method is using pass to convert assign to structinject or I directly generate structinject while importing. I was thinking it is needless to have a so easy conversation by a pass.

mingzheTerapines:
@fabianschuiki @Moxinilian @hailongSun2000 Now conversation is like that. Sorry I am not sure it is suitable. please give me some advice, I would appreciate it.
sv
```
  module Foo;
  int x, y, z;
  typedef struct packed {
    int a, b;
  } ab;
  always_comb begin
     ab ii;
   ii.a = 32'b1;
   ii.b = 32'b100;
    ii.a = x + 1;
    y = ii.a;
    ii.a = ii.a + 1;
    z = ii.a;
  end
endmodule
```
importverilog
```
module {
  moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    moore.procedure always_comb {
      %ii = moore.variable : <struct<{a: i32, b: i32}>>
      %0 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %1 = moore.constant 1 : i32
      %2 = moore.conversion %1 : !moore.i32 -> !moore.i32
      moore.struct_inject %ii, "a", %2 : <struct<{a: i32, b: i32}>> i32
      %3 = moore.struct_extract_ref %ii, "b" : <struct<{a: i32, b: i32}>> -> <i32>
      %4 = moore.constant 4 : i32
      %5 = moore.conversion %4 : !moore.i32 -> !moore.i32
      moore.struct_inject %ii, "b", %5 : <struct<{a: i32, b: i32}>> i32
      %6 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %7 = moore.read %x : i32
      %8 = moore.constant 1 : i32
      %9 = moore.add %7, %8 : i32
      moore.struct_inject %ii, "a", %9 : <struct<{a: i32, b: i32}>> i32
      %10 = moore.struct_extract %ii, "a" : <struct<{a: i32, b: i32}>> -> i32
      moore.blocking_assign %y, %10 : i32
      %11 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %12 = moore.struct_extract %ii, "a" : <struct<{a: i32, b: i32}>> -> i32
      %13 = moore.constant 1 : i32
      %14 = moore.add %12, %13 : i32
      moore.struct_inject %ii, "a", %14 : <struct<{a: i32, b: i32}>> i32
      %15 = moore.struct_extract %ii, "a" : <struct<{a: i32, b: i32}>> -> i32
      moore.blocking_assign %z, %15 : i32
    }
    moore.output
  }
}
```
sroa
```
module {
  moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    moore.procedure always_comb {
      %0 = moore.constant 0 : i32
      %a = moore.variable %0 : <i32>
      %1 = moore.constant 0 : i32
      %b = moore.variable %1 : <i32>
      %2 = moore.struct_create %0, %1 : !moore.i32, !moore.i32 -> struct<{a: i32, b: i32}>
      %3 = moore.constant 1 : i32
      %4 = moore.conversion %3 : !moore.i32 -> !moore.i32
      moore.blocking_assign %a, %4 : i32
      %5 = moore.constant 4 : i32
      %6 = moore.conversion %5 : !moore.i32 -> !moore.i32
      moore.blocking_assign %b, %6 : i32
      %7 = moore.read %x : i32
      %8 = moore.constant 1 : i32
      %9 = moore.add %7, %8 : i32
      moore.blocking_assign %a, %9 : i32
      %10 = moore.read %a : i32
      moore.blocking_assign %y, %10 : i32
      %11 = moore.read %a : i32
      %12 = moore.constant 1 : i32
      %13 = moore.add %11, %12 : i32
      moore.blocking_assign %a, %13 : i32
      %14 = moore.read %a : i32
      moore.blocking_assign %z, %14 : i32
    }
    moore.output
  }
}
```
mem2reg
```
module {
  moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    moore.procedure always_comb {
      %0 = moore.constant 0 : i32
      %1 = moore.constant 0 : i32
      %2 = moore.struct_create %0, %1 : !moore.i32, !moore.i32 -> struct<{a: i32, b: i32}>
      %3 = moore.constant 1 : i32
      %4 = moore.conversion %3 : !moore.i32 -> !moore.i32
      %5 = moore.constant 4 : i32
      %6 = moore.conversion %5 : !moore.i32 -> !moore.i32
      %7 = moore.read %x : i32
      %8 = moore.constant 1 : i32
      %9 = moore.add %7, %8 : i32
      moore.blocking_assign %y, %9 : i32
      %10 = moore.constant 1 : i32
      %11 = moore.add %9, %10 : i32
      moore.blocking_assign %z, %11 : i32
    }
    moore.output
  }
}
```

hailongSun2000:
Hey, @mingzheTerapines! Could you display a case with the module-level/global variable with the struct type?

mingzheTerapines:
> Hey, @mingzheTerapines! Could you display a case with the module-level/global variable with the struct type?

Thanks for reminding, I have fixed it. SROA will do nothing for this case, because other assigning is non-blocking.

hailongSun2000:
My thought is that the `Mem2Reg` pass can eliminate the local variable including the variable with struct type, right? So maybe after executing the `Mem2Reg` pass, the `struct_create` unnecessarily exists, right?

hailongSun2000:
I'm not an expert on `SROA` and `Mem2Reg`. But from the previous discussion, we want to lower `variable` into `struct_create` and `struct_extract` and `*assign` together into `struct_inject`. In this case, we can directly lower them into hw dialect one-to-one.

However, we introduce the `moore.read`, `struct_extract_ref`, etc. So we need to properly tweak the strategy. I'm not sure whether we must directly generate `struct_inject`, rather than `*assign`.

Assume through `SROA` can achieve the previous discussion. There are two options:
1. Go through `SROA` and `Mem2Reg` to eliminate the local variables including `IntType`, `StructType`, etc. And lower the global/module-level `variable` into `struct_create` and `struct_extract` and `*assign` into `struct_inject`.
2. Only eliminate the local variable, don't operate on the module-level variables with struct type. Then add a new pass to implement the rest of the lowering.

What do @fabianschuiki @mingzheTerapines think?

Moxinilian:
> @fabianschuiki pass SROA has this method:only handling Op that used this destructable variable.
> 
> ```
>   // Initialize the analysis with the immediate users of the slot.
>   for (OpOperand &use : slot.ptr.getUses()) {
>     if (auto accessor =
>             dyn_cast<DestructurableAccessorOpInterface>(use.getOwner())) {
>       if (accessor.canRewire(slot, info.usedIndices, usedSafelyWorklist,
>                              dataLayout)) {
>         info.accessors.push_back(accessor);
>         continue;
>       }
>     }
> ```
> 
> However, codes like this
> 
> ```
> module {
>   moore.module @Foo() {
>     %x = moore.variable : <i32>
>     %y = moore.variable : <i32>
>     %z = moore.variable : <i32>
>     %dd = moore.variable : <struct<{a: i32, b: i32}>>
>     moore.procedure always_comb {
>       %d = moore.variable : <i32>
>       %3 = moore.struct_extract_ref %dd, "a" : <struct<{a: i32, b: i32}>> -> <i32>
>       %4 = moore.read %d : i32
>       moore.blocking_assign %3, %4 : i32
>     }
>     moore.output
>   }
> }
> ```
> 
> moore.blocking_assign does not directly used %dd, so destructable-related function was not called. Yet I can not call destructable-related function at moore.struct_extract_ref because I do not know which value should assign to this pointer. One method is using pass to convert assign to structinject or I directly generate structinject while importing. I was thinking it is needless to have a so easy conversation by a pass.

I don’t understand what is wrong with this code. Can you not make the struct extract an SROA accessor similar to LLVM’s GEPOp?

mingzheTerapines:
@hailongSun2000 You are right, Struct Type could be eliminated for local variable.

mingzheTerapines:
@Moxinilian The point is SROA I need rewire blockingAssign Op, but canrewire function will be called if this Op directly used destrucatable VaraibleOp(which means%dd). However this Op used %3  %4 and %3(extract_refOp) used this variableOp(%dd) which means used undirectly.
Problem has been solved by generating code like this.
```
module {
  moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    %dd = moore.variable : <struct<{a: i32, b: i32}>>
    moore.procedure always_comb {
      %d = moore.variable : <i32>
      %0 = moore.struct_extract_ref %dd, "a" : <struct<{a: i32, b: i32}>> -> <i32>
      %1 = moore.read %d : i32
      moore.struct_inject %dd, "a", %1 : <struct<{a: i32, b: i32}>> i32
    }
    moore.output
  }
}
```

Moxinilian:
No I don’t think you needed to reword the blocking assign, why would you?

Moxinilian:
When you rewire the struct extract ref op, the blocking assign becomes the user of an integer variable directly. This can be taken care of by Mem2Reg, if appropriate. No need to apply SROA to an assignment of to a value that is already a scalar, no need to do scalar replacement of aggregates when there is no aggregate :)

Moxinilian:
To be more specific: the point of SROA is to take an allocation/variable of a type that is an aggregate, and turn that allocation/variable into one allocation/variable for each of its fields. The main effect happens at the definition site of the allocation/variable, with the rest being about rewiring all the ops so they use the individual allocations themselves.

Here, what you want in the big picture with SROA is to replace all uses of a field reference to the newly created allocation. Nothing more, elimination of fields themselves is Mem2Reg’s job.

mingzheTerapines:
@Moxinilian That make sense, I will have a check.

mingzheTerapines:
@Moxinilian @hailongSun2000 
Here is the new version:
1. extractrefOp replace all use witch realted variableOp
2. only hanlde local struct variable for SROA

mingzheTerapines:
@fabianschuiki  I have the same question with Hailong. What should we do with Global Struct Variable? 
Here is one proposal: Destructe in several variable Ops. And use structCreateop to gather them. It seems HW dialect is doing like that. 
The chanllenges are:  
a. New varaible Ops may have naming problem with SSA rules. We need be careful for maintaining value symbol table and a method to name.
b.  some variable-optimized related pass may modify these varaibles unexpectedly. It can be solved if we have a chain-chekc at structCreateOp which can get all relating variableOp.
b. Ops like structInjectOp may be useless.
Another proposal is no destrucatable, only do Ops replacement like structInjectOp makes assigning looks more clearly.
The challenges ares:
a. Hard to do MooretoCore pass since hw.structCreate using several inputs.
b. Variable-optimized related pass will never walk for these Ops.
b. Just like you mentioned, some operations that only write part of the destrucatable memory -> creating new local variable and  write it to the whole destrucatable memory.
```
%oldStruct = hw.struct_create("x": %a, "y": %b)
%newStruct = hw.struct_inject %oldStruct, "y", %c
// oldStruct remains unchanged, newStruct looks like this:
// %newStruct = hw.struct_create("x": %a, "y": %c)
```
We can do some optimization for writing to struct.x and struct.y by do writing parallely, but we can not write struct parallely which has to be in order or we only can write successfully once.
Combing together I have proposal like this:
SV
```
 module Foo;
  int x, y, z;
  typedef struct packed {
    int a, b;
  } ab;
     ab ii;

   assign ii.b = 32'b100;
    assign ii.a = x + 1;
    assign y = ii.a;
    assign z = ii.a;
 endmodule
```
importverilog
```
 moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    %ii = moore.variable : <struct<{a: i32, b: i32}>>
    %0 = moore.struct_extract_ref %ii, "b" : <struct<{a: i32, b: i32}>> -> <i32>
    %1 = moore.constant 4 : i32
    %2 = moore.conversion %1 : !moore.i32 -> !moore.i32
    moore.assign %0, %2 : i32
    %3 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
    %4 = moore.read %x : i32
    %5 = moore.constant 1 : i32
    %6 = moore.add %4, %5 : i32
    moore.assign %3, %6 : i32
    %7 = moore.struct_extract %ii, "a" : <struct<{a: i32, b: i32}>> -> i32
    moore.assign %y, %7 : i32
    %8 = moore.struct_extract %ii, "a" : <struct<{a: i32, b: i32}>> -> i32
    moore.assign %z, %8 : i32
    moore.output
  }
```
SROA
```
 moore.module @Foo() {
    %x = moore.variable : <i32>
    %y = moore.variable : <i32>
    %z = moore.variable : <i32>
    %0 = moore.struct_extract_ref %ii, "a" : <struct<{a: i32, b: i32}>> -> <i32>
    %1 = moore.struct_extract_ref %ii, "b" : <struct<{a: i32, b: i32}>> -> <i32>
   2  = moore.struct_create %0, %1 : <i32> , <i32> -> struct<{a: i32, b: i32}>
    %3 = moore.constant 4 : i32
    %4 = moore.conversion %1 : !moore.i32 -> !moore.i32
    moore.assign %1, %4 : i32
    %6 = moore.read %x : i32
    %7 = moore.constant 1 : i32
    %8 = moore.add %1, %2 : i32
    moore.assign %0, %8 : i32
    moore.assign %y, %0: i32
    moore.assign %z, %0 : i32
    moore.output
  }

**Comment by mingzheTerapines** [comment]
@fabianschuiki It seems it is not creating new struct when lowering to LLVM, refering to circt/test/Conversion/HWToLLVM/convert_aggregates.mlir
image

**Comment by fabianschuiki** [comment]
> @mingzheTerapines It seems it is not creating new struct when lowering to LLVM

Yes, it lowers directly to the corresponding LLVM `{extract,insert}value` operations. What kind of LLVM did you expect to be generated?

**Comment by fabianschuiki** [comment]
Doing SROA for module-level variables sounds like a reasonable thing to do! Or maybe that's not even necessary, and Mem2Reg can create the corresponding `moore.struct_create` ops instead? I'm not too familiar with the detailed implementation of SROA and Mem2Reg. Pretty sure `moore.struct_inject` doesn't really work for module-level variables, since the assignments in a module all occur in parallel, so there is no order in the sense of old struct before the assignment and new struct after the assignment.

I wouldn't be too concerned about the naming of variables and structs: the `moore.variable` and `moore.net` ops are more about how we represent variables in the IR, and not what the variables were that the user typed. So if we have to split a single struct variable up into 100 separate variables, that is totally fine and the user will never see the names of those variables. Instead, once we care about preserving the variable names of the original Verilog source code, we'll want to start adding `dbg.variable` and `dbg.struct` and `dbg.array` operations in the `ImportVerilog` code (at least when some form of debug information is enabled). These `dbg.*` operations are designed to track what the original variables were in the source language, and they allow dialects like Moore and HW/Comb/Seq/Arc/FIRRTL to aggressively restructure the IR and the declarations without having to worry about maintaining any user-defined structure 😃

**Comment by mingzheTerapines** [comment]
> > @mingzheTerapines It seems it is not creating new struct when lowering to LLVM
> 
> Yes, it lowers directly to the corresponding LLVM `{extract,insert}value` operations. What kind of LLVM did you expect to be generated?

It may be lowered to llvm.mlir.undef : !llvm.struct<(i32, i2, i1)> which really create a struct
As Hailong reminded, canonicalize can convert hw.struct_inject to struct_create explictly which will be lowered to like this.
```
// CHECK-NEXT: [[V3:%.*]] = llvm.mlir.undef : !llvm.struct<(i32, i2, i1)>
  // CHECK-NEXT: [[V4:%.*]] = llvm.insertvalue %arg0, [[V3]][0] : !llvm.struct<(i32, i2, i1)>
  // CHECK-NEXT: [[V5:%.*]] = llvm.insertvalue %arg4, [[V4]][1] : !llvm.struct<(i32, i2, i1)>
  // CHECK-NEXT: [[V6:%.*]] = llvm.insertvalue %arg3, [[V5]][2] : !llvm.struct<(i32, i2, i1)>
  %3 = hw.struct_create (%arg3, %arg4, %arg0) : !hw.struct<foo: i1, bar: i2, baz: i32>
```
and this new struct will replace use with old struct

**Review by fabianschuiki — COMMENTED** [review]
Nice! Can you add a test for the folder of the StructInjectOp? I was wondering whether we have folders and constant materialization for structs already. It would be nice to have :smiley:

**Comment by mingzheTerapines** [comment]
> Nice! Can you add a test for the folder of the StructInjectOp? I was wondering whether we have folders and constant materialization for structs already. It would be nice to have 😃

Yes, I have added relating test for struct_inject folding test.
Also I supported struct_createOp folding. But it relates to other input Ops fold supporting, so there are no test demo for it.

File: lib/Dialect/Moore/MooreTypes.cpp [hunk 30]
```diff
@@ -29,6 +29,16 @@ static void printMembers(AsmPrinter &printer,
...
+static std::optional<DenseMap<Attribute, Type>>
+getAllSubelementIndexMap(ArrayRef<StructLikeMember> members);
+static Type getTypeAtAllIndex(ArrayRef<StructLikeMember> members,
+                              Attribute index);
+
+static std::optional<DenseMap<Attribute, Type>>
+getAllSubelementIndexMap(ArrayRef<StructLikeMember> members);
+static Type getTypeAtAllIndex(ArrayRef<StructLikeMember> members,
+                              Attribute index);
+
```
fabianschuiki, line 41: I don't think you need these forward declarations :slightly_smiling_face:

fabianschuiki (approved):
This looks nice! Thanks for doing all this hard work :partying_face: Just a minor comment on a bit of possibly unneeded code.

lib/Dialect/Moore/MooreTypes.cpp
mingzheTerapines: Sorry, I should have removed it but forgot.

PR #7160, opened 2024-06-12
[FIRRTL] Add pass to specialize layers

youngar:


lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
```diff
@@ -0,0 +1,725 @@
...
+  }
+
+  /// Specialize a block of operations, removing any probes which are
+  /// disabled, and moving all operations to the insertion point.
+  void specializeBlock(Block *block, InsertionPoint &insertionPoint,
+                       DenseSet<Attribute> &removedSyms) {
+    // Since this can erase operations that deal with disabled probes, we walk
+    // thee block in reverse to make sure that we erase uses before defs.
+    for (auto &op : llvm::make_early_inc_range(llvm::reverse(*block)))
+      specializeOp(&op, insertionPoint, removedSyms);
```
seldridge, line 451: This made me very confused for a while because I thought this was doing function overloading _at runtime_... I missed the `specializedOp` with the `TypeSwitch` above.

Personally, this would be more readable with that function inlined here as the indirection to only a `TypeSwitch` seems a bit too finely grained.

test/Dialect/FIRRTL/specialize-layers-errors.mlir
```diff
@@ -0,0 +1,28 @@
...
+
+// -----
+
+// expected-error @below {{unknown layer @A}}
+firrtl.circuit "Test" attributes {
+  firrtl.disable_layers = [@A]
+} {
+  firrtl.module @Test() {}
+}
+
```
seldridge, line 18: Duplicate test?

seldridge (commented):
This is shaping up great!

One question that I haven't fully though through: if there is a module with layers enabled and we disable those layers, what do we do with that module? I almost think that the entire module gets removed, as opposed to what is the current code where just that layer is removed from being enabled. My reasoning here is that enabling a layer on a module means that the module can now read freely from any probes colored with the enabled layer. However, this is now creating a large number of likely dead references.

I do note that this could result in very aggressive deletions. E.g., we have the `Verification` layer enabled on an internal testharness. If we disable that layer and there are no public modules left, then the circuit just got whacked.

Concrete example. What happens if `A` is disabled?

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  layer A, bind:

  extmodule Bar:
    input in: UInt<1>
    output a: Probe<UInt<1>, A>

  public module Foo enablelayer A:

    inst bar of Bar

    connect bar.in, read(bar.a)
```

WDYT?

What may have made this easier was to decouple the layer symbol from the layer name. That would avoid some of the need to do symbol updates. That said, the current approach does keep the IR looking tidier. That may be an anti-pattern. Regardless, good work on this logic. The tests look very correct to me.

youngar:
Thanks for taking a look @seldridge.

>  if there is a module with layers enabled and we disable those layers, what do we do with that module?

Right now the code is deleting the module, there is a test here:

https://github.com/llvm/circt/pull/7160/files#diff-aab1abaeffdac59090e622bc6af09a5ca9489665e6cfad4fc7710332d482aec9R220

The logic is that it becomes impossible to instantiate the module, so implicitly every instance of the module would have been deleted by specialization, so its safe to remove the module. It might seem a aggressive to delete the module instead of letting something like IMDeadCodeElim remove it, but since we actually remove the layer declarations when they get specialized the enabled layer would become a dangling reference and fails to verify.

>  If we disable that layer and there are no public modules left, then the circuit just got whacked.

Yeah, had that problem when I was writing the test, there is a comment describing why there is a useless top level module :).  I think the way things are going, we are eventually going to remove the concept of the top level module and then having an empty circuit would be fine.

test/Dialect/FIRRTL/specialize-layers-errors.mlir
youngar: Difference is testing `disable_layers` and `enable_layers`

seldridge:
Awesome. That logic makes exact sense to me. I didn't read the tests closely enough.

Agree about multi-public making empty circuits sane constructions.

lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
```diff
@@ -0,0 +1,725 @@
...
+    op->moveBefore(block, it);
+    it = Block::iterator(op);
+  }
+
+private:
+  InsertionPoint(Block *block, Block::iterator it) : block(block), it(it) {}
+
+  Block *block;
+  Block::iterator it;
+};
```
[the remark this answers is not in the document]
youngar: Everyone find this confusing but I have a suggestion from @rwy7 which will hopefully make it clearer, but I am starting to think it is not really an abstraction which makes sense to other people and I should delete it.  I was sort of thinking of this as an insertion point that points between operations instead of an operation.  Basically this is an insertion point that modifies itself every time you move an operation after it, to point to the moved operation.  The reason I need it is a little involved but it at a high level its because I am both enabling and disabling layers at the same time. When we are disabling layers, we have to delete operations, and so we walk the blocks in reverse to make sure that we removed uses of values before the value itself.  When we enable a layer, we inline the contents of the layerblock into the outer context, which turns out to be a little difficult to do in reverse.  When we inline an operation, we need to make sure that the insertion point is updated so that the next operation inlined goes before the previously inlined operation.

lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
```diff
@@ -0,0 +1,725 @@
...
+                    DenseSet<Attribute> &removedSyms) {
+    for (size_t i = 0, e = match.getNumRegions(); i < e; ++i) {
+      auto *caseBlock = &match.getRegion(i).front();
+      auto caseIP = InsertionPoint::atBlockEnd(caseBlock);
+      specializeBlock(caseBlock, caseIP, removedSyms);
+    }
+    insertionPoint.moveOpBefore(match);
+  }
+
+  void specializeOp(InstanceOp instance, InsertionPoint &insertionPoint,
```
dtzSiFive, line 343:
```suggestion
  void specializeOp(InstanceOp instance, InsertionPoint &insertionPoint,
```

Can these re-use the same implementation? Not sure if there's an interface they share, but maybe if templated? Your call if you'd prefer it this way :+1:.  Looking around, less sure but leaving for your consideration.

lib/Firtool/Firtool.cpp
```diff
@@ -116,6 +116,8 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
   pm.addNestedPass<firrtl::CircuitOp>(firrtl::createLowerFIRRTLTypesPass(
       opt.getPreserveAggregate(), firrtl::PreserveAggregate::None));
 
+  pm.addNestedPass<firrtl::CircuitOp>(firrtl::createSpecializeLayersPass());
```
dtzSiFive, line 119: Just wondering -- why at this point in the pipeline? Would it make sense to run this earlier? Motivation for this question is I'm wondering if we're deleting things doing it earlier is less work for everything after that, as one consideration anyway.

Disabling layers (and thereby dropping ports, modules, so on) might mask errors that would otherwise occur if not specializing, so dunno.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5598,6 +5598,34 @@ ParseResult FIRCircuitParser::parseCircuit(
...
+    SmallVector<FlatSymbolRefAttr> nestedRefs;
+    while (!rest.empty()) {
+      StringRef next;
+      std::tie(next, rest) = rest.split("::");
+      nestedRefs.push_back(FlatSymbolRefAttr::get(getContext(), next));
+    }
+    return SymbolRefAttr::get(getContext(), head, nestedRefs);
+  };
+
+  auto parseLayers = [&](const std::vector<std::string> &layers) {
```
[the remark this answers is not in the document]
youngar: Swapped to `auto parseLayers = [&](const auto &layers) {`

lib/Firtool/Firtool.cpp
youngar: We talked about this offline: we decided that it would be good to move this earlier in the pipeline but it must run after all the diagnostic passes have run.  We ended up moving it later in the pipeline, after expand whens and check comb loops.  There is some work to create a diagnostic only version of expand whens that is diagnostic only and runs very early in the pipeline and we can maybe move this pass earlier once that is merged.

PR #7161, opened 2024-06-12
[Moore] Add the SimplifyProcedures pass.

hailongSun2000:
Cause we intend to entirely inline the `always_comb` body into the module body(`moore.module`). Due to the difference way between continuous and blocking assignments, we decided to create a new pass named `SimplifyProcedures` to insert a local "shadow" variable in always blocks for every module-level/global variable that the process modifies. Quote @fabianschuiki 's example:
### origin:
```
module Foo;
  int a;
  int x, y, z;
  always_comb begin
    a = 1;
    x = a;
    a += 1;
    y = a;
    a += 1;
    z = a;
  end
endmodule
```
### after performing SimplifyProcedures pass:
```
module Foo;
  int a;
  int x, y, z;
  always_comb begin
    int local_a = a;
    local_a = 1; a = local_a;
    x = local_a;
    local_a += 1; a = local_a;
    y = local_a;
    local_a += 1; a = local_a;
    z = local_a;
  end
endmodule
```

### DIFF
--- a/include/circt/Dialect/Moore/CMakeLists.txt
+++ b/include/circt/Dialect/Moore/CMakeLists.txt
@@ -15,3 +15,7 @@ mlir_tablegen(MooreAttributes.cpp.inc -gen-attrdef-defs
   -attrdefs-dialect MooreDialect)
 add_public_tablegen_target(CIRCTMooreAttributesIncGen)
 add_dependencies(circt-headers CIRCTMooreAttributesIncGen)
+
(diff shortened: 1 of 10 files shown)


hailongSun2000:
Thanks for the code review in advance! :heart:

fabianschuiki (approved):
LGTM! Thanks for adding all the infrastructure for Moore dialect passes in this PR :+1:

PR #7163, opened 2024-06-12
[FIRRTL][ExportVerilog] Emit integers on DPI function as two state C-compatible types and clarify ABI

uenoku:
This PR modifies ExportVerilog to emit two state types (`bit` in general) for DPI import op. Furthermore, for specific bit width (8, 16, 32 and 64) it emits C-types (byte, shortint, int and longint). 

This PR also rejects small integer types other than 8, 16, 32 and 64 bit width since  otherwise we have to use`bit` but they are passed by references in DPI. So this PR defines the ABI as "small integers are passed by values, and larger integers are passed by references.

### DIFF
--- a/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
+++ b/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
@@ -242,3 +242,46 @@ not called. Hence their values will not be modify in that clock.
 | enable            | input     | UInt<1>  | Enable signal                   |
 | ...               | input     | Signals  | Arguments to DPI function call  |
 | result (optional) | output    | Signal   | Optional result of the dpi call |
+
+##### DPI Intrinsic ABI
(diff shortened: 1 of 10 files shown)


lib/Conversion/ExportVerilog/PrepareForEmission.cpp
```diff
@@ -1336,7 +1336,6 @@ struct PrepareForEmissionPass
   void runOnOperation() override {
     auto module = getOperation();
     LoweringOptions options(cast<mlir::ModuleOp>(module->getParentOp()));
-    module.dump();
```
uenoku, line 1339: this is due to stacked pr.

fabianschuiki (approved):
I love the idea of limiting arguments to the types which get passed by-value according to the SV spec. We can always make this more general later. :partying_face:

Issue #4831, opened 2023-03-15
[SCFToCalyx] Use `ref` construct when lowering `memref`


rachitnigam:
The current `scf` to `calyx` conversion handles `memref` arguments by directly inlining all the ports for the memory into the component's interface. Instead of doing this, we should use the new [`ref` keyword](https://docs.calyxir.org/lang/memories-by-reference.html) from Calyx and let lower passes handle the transformation of a `ref` cell by inlining its ports into the component.

The `ref` keyword is more general–it allows you to pass in any component by reference into another component to express component sharing–and it allows the compiler to simultaneously generate simulatable and synthesizable code. In the current state of affairs, the compiler cannot simulate designs that only have toplevel `memref`s in them.

Note that this also requires changing the `invoke` construct to accept components by name (see documentation from Calyx above).

rachitnigam:
This also requires solving #1679 because the `ref` syntax only works with `invoke`

cgyurgyik:
Addressed by #7164.

PR #7164, opened 2024-06-12
[Calyx] Passing memories by reference

jiahanxie353:
This patch addresses #4831 .

If the `main` `FuncOp` has `memref`s in its arguments, we create a new `main` component and make it the new entry point.

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxControl.td
+++ b/include/circt/Dialect/Calyx/CalyxControl.td
@@ -450,7 +450,7 @@ def InvokeOp : CalyxOp<"invoke", [
   let arguments = (ins FlatSymbolRefAttr:$callee,
                    Variadic<AnyType>:$ports,
                    Variadic<AnyType>:$inputs,
-                   DictionaryAttr:$refCellsMap,
+                   ArrayAttr:$refCellsMap,
                    ArrayAttr:$portNames,
                    ArrayAttr:$inputNames);
   let results = (outs);
--- a/include/circt/Dialect/Calyx/CalyxControl.td
+++ b/include/circt/Dialect/Calyx/CalyxControl.td
@@ -426,26 +426,31 @@ def InvokeOp : CalyxOp<"invoke", [
     call, which invokes a given component.
 
     The 'callee' attribute is the name of the component,
-    the 'ports' attribute The ports attribute specifies 
-    the input port of the component when it is invoked, 
-    and the 'inputs' attribute specifies the assignment on 
-    the corresponding port.
+    the 'ports' attribute specifies the input port of the component when it is invoked,
+    the 'inputs' attribute specifies the assignment on the corresponding port,
+    and the 'refCellsMap' attribute maps the reference cells in the `callee` with
+    the original cells in the caller.
 
     ```mlir
+      calyx.component @identity {
+        %mem.addr0, %mem.clk, ... = calyx.seq_mem @mem <[1] x 32> [1] {external = false}
+      }
       %id.in, %id.out, ... = calyx.instance @id of @identity : i32, i32, ...
       %r.in, ... = calyx.register @r : i32, ...
+      %mem_1.addr0, %mem_1.clk, ... = calyx.seq_mem @mem_1 <[1] x 32> [1] {external = true}
       ...
       calyx.control {
         calyx.seq {
-          calyx.invoke @id(%id.in = %c1_10, %r.in = %id.out) -> (i32, i32)
+          calyx.invoke @id[mem = mem_1](%id.in = %c1_10, %r.in = %id.out) -> (i32, i32)
         }
       }
     ```
   }];
 
-  let arguments = (ins FlatSymbolRefAttr:$callee, 
+  let arguments = (ins FlatSymbolRefAttr:$callee,
                    Variadic<AnyType>:$ports,
                    Variadic<AnyType>:$inputs,
+                   DictionaryAttr:$refCellsMap,
                    ArrayAttr:$portNames,
                    ArrayAttr:$inputNames);
   let results = (outs);
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -84,6 +84,7 @@ void buildAssignmentsForRegisterWrite(OpBuilder &builder,
 // A structure representing a set of ports which act as a memory interface for
 // external memories.
 struct MemoryPortsImpl {
+  StringRef memName;
   std::optional<Value> readData;
   std::optional<Value> readOrContentEn;
   std::optional<Value> writeData;
@@ -103,6 +104,7 @@ struct MemoryInterface {
   explicit MemoryInterface(calyx::SeqMemoryOp memOp);
 
   // Getter methods for each memory interface port.
+  StringRef memName();
   Value readData();
   Value readEn();
   Value contentEn();
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -84,7 +84,7 @@ void buildAssignmentsForRegisterWrite(OpBuilder &builder,
 // A structure representing a set of ports which act as a memory interface for
 // external memories.
 struct MemoryPortsImpl {
-  StringRef memName;
+  std::string memName;
   std::optional<Value> readData;
   std::optional<Value> readOrContentEn;
   std::optional<Value> writeData;
@@ -104,7 +104,7 @@ struct MemoryInterface {
   explicit MemoryInterface(calyx::SeqMemoryOp memOp);
 
   // Getter methods for each memory interface port.
-  StringRef memName();
+  std::string memName();
   Value readData();
   Value readEn();
   Value contentEn();
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -2014,24 +2014,34 @@ private:
 
   /// Creates a new new top-level function based on `baseName`.
   FuncOp createNewTopLevelFn(ModuleOp moduleOp, std::string &baseName) {
-    std::string newName = baseName;
-    unsigned counter = 0;
-    while (SymbolTable::lookupSymbolIn(moduleOp, newName)) {
-      newName = llvm::join_items("_", baseName, std::to_string(++counter));
+    std::string newName = "main";
+
+    if (auto *existingMainOp = SymbolTable::lookupSymbolIn(moduleOp, newName)) {
+      if (auto existingMainFunc = dyn_cast<FuncOp>(existingMainOp)) {
+
+        unsigned counter = 0;
+        std::string newOldName = baseName;
+        while (SymbolTable::lookupSymbolIn(moduleOp, newOldName))
+          newOldName = baseName + "_" + std::to_string(++counter);
+
+        existingMainFunc.setName(newOldName);
+        if (baseName == "main")
+          baseName = newOldName;
+      } else {
+        moduleOp.emitError() << "Symbol 'main' exists but is not a function";
+        return nullptr;
+      }
     }
 
+    // Create the new "main" function
     OpBuilder builder(moduleOp.getContext());
     builder.setInsertionPointToStart(moduleOp.getBody());
 
     FunctionType funcType = builder.getFunctionType({}, {});
 
     if (auto newFunc =
-            builder.create<FuncOp>(moduleOp.getLoc(), newName, funcType)) {
-      baseName = newName;
+            builder.create<FuncOp>(moduleOp.getLoc(), newName, funcType))
       return newFunc;
-    }
-
-    moduleOp.emitError("Cannot create new top-level function.");
 
     return nullptr;
   }
@@ -2086,6 +2096,7 @@ private:
     // corresponding operations in the new top-level.
     builder.setInsertionPointToEnd(callerEntryBlock);
     for (auto *op : opsToModify) {
+      // TODO: copy attributes as well
       Value newOpRes;
       if (auto allocaOp = dyn_cast<memref::AllocaOp>(op)) {
         newOpRes = builder.create<memref::AllocaOp>(callee.getLoc(),
@@ -2162,19 +2173,21 @@ private:
                  hasMemrefArguments(funcOp);
         });
 
-    std::string oldName = topLevelFunction;
     if (hasMemrefArgsInTopLevel) {
       auto newTopLevelFunc = createNewTopLevelFn(moduleOp, topLevelFunction);
+      if (!newTopLevelFunc)
+        return failure();
 
       OpBuilder builder(moduleOp.getContext());
       Operation *oldTopLevelFuncOp =
-          SymbolTable::lookupSymbolIn(moduleOp, oldName);
+          SymbolTable::lookupSymbolIn(moduleOp, topLevelFunction);
       if (auto oldTopLevelFunc = dyn_cast<FuncOp>(oldTopLevelFuncOp))
         insertCallFromNewTopLevel(builder, newTopLevelFunc, oldTopLevelFunc);
       else {
         moduleOp.emitOpError("Original top-level function not found!");
         return failure();
       }
+      topLevelFunction = "main";
     }
 
     return success();
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -701,7 +701,8 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
 
 template <typename TAllocOp>
 static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
-                                  PatternRewriter &rewriter, TAllocOp allocOp) {
+                                  PatternRewriter &rewriter, TAllocOp allocOp,
+                                  bool isExternal) {
   rewriter.setInsertionPointToStart(
       componentState.getComponentOp().getBodyBlock());
   MemRefType memtype = allocOp.getType();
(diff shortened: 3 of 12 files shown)


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1616,7 +1659,8 @@ class SCFToCalyxPass : public SCFToCalyxBase<SCFToCalyxPass> {
         return failure();
       }
     }
-    return success();
+
+    return createOptNewTopLevelFcn(moduleOp, topLevelFunction);
```
cgyurgyik, line 1663: Nit: Fcn -> Fn

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1712,6 +1756,94 @@ class SCFToCalyxPass : public SCFToCalyxBase<SCFToCalyxPass> {
...
+
+    FunctionType funcType = builder.getFunctionType({}, {});
+
+    if (auto newFunc =
+            builder.create<FuncOp>(moduleOp.getLoc(), newName, funcType)) {
+      baseName = newName;
+      return newFunc;
+    }
+
+    moduleOp.emitError("Cannot create new top-level function.");
```
cgyurgyik, line 1778:
```suggestion
    return moduleOp.emitError("Cannot create new top-level function.");
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1712,6 +1756,94 @@ class SCFToCalyxPass : public SCFToCalyxBase<SCFToCalyxPass> {
 private:
   LogicalResult partialPatternRes;
   std::shared_ptr<calyx::CalyxLoweringState> loweringState = nullptr;
+
+  FuncOp createNewTopLevelFcn(ModuleOp moduleOp, std::string &baseName) {
+    std::string newName = baseName;
+    unsigned counter = 0;
+    while (SymbolTable::lookupSymbolIn(moduleOp, newName)) {
+      newName = baseName + "_" + std::to_string(++counter);
```
cgyurgyik, line 1764: let's use `llvm::join`, its more efficient.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1712,6 +1756,94 @@ class SCFToCalyxPass : public SCFToCalyxBase<SCFToCalyxPass> {
 private:
   LogicalResult partialPatternRes;
   std::shared_ptr<calyx::CalyxLoweringState> loweringState = nullptr;
+
+  FuncOp createNewTopLevelFcn(ModuleOp moduleOp, std::string &baseName) {
```
cgyurgyik, line 1760: Add a comment here please, explaining what this is doing.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1712,6 +1756,94 @@ class SCFToCalyxPass : public SCFToCalyxBase<SCFToCalyxPass> {
...
+    }
+
+    OpBuilder builder(moduleOp.getContext());
+    builder.setInsertionPointToStart(moduleOp.getBody());
+
+    FunctionType funcType = builder.getFunctionType({}, {});
+
+    if (auto newFunc =
+            builder.create<FuncOp>(moduleOp.getLoc(), newName, funcType)) {
+      baseName = newName;
```
cgyurgyik, line 1774: Instead of passing a mutable parameter `baseName` and updating it, can you just get the name from the returned `FuncOp`?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1103,36 +1094,38 @@ struct FuncOpConversion : public calyx::FuncOpPartialLoweringPattern {
     auto *compState = loweringState().getState<ComponentLoweringState>(compOp);
     compState->setFuncOpResultMapping(funcOpResultMapping);
 
+    unsigned extMemCounter = 0;
+    for (auto arg : enumerate(funcOp.getArguments())) {
+      if (isa<MemRefType>(arg.value().getType())) {
+        std::string memName = "arg_mem" + std::to_string(extMemCounter++);
```
cgyurgyik, line 1100: llvm::join

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1103,36 +1094,38 @@ struct FuncOpConversion : public calyx::FuncOpPartialLoweringPattern {
     auto *compState = loweringState().getState<ComponentLoweringState>(compOp);
     compState->setFuncOpResultMapping(funcOpResultMapping);
 
+    unsigned extMemCounter = 0;
```
cgyurgyik, line 1097: Can you write a unit CHECK test for this please?

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -2621,7 +2621,14 @@ ParseResult InvokeOp::parse(OpAsmParser &parser, OperationState &result) {
 }
 
 void InvokeOp::print(OpAsmPrinter &p) {
```
cgyurgyik, line 2623: Can you add a test for this please, maybe in round-trip?

include/circt/Dialect/Calyx/CalyxControl.td
```diff
@@ -426,26 +426,31 @@ def InvokeOp : CalyxOp<"invoke", [
...
+        %mem.addr0, %mem.clk, ... = calyx.seq_mem @mem <[1] x 32> [1] {external = false}
+      }
       %id.in, %id.out, ... = calyx.instance @id of @identity : i32, i32, ...
       %r.in, ... = calyx.register @r : i32, ...
+      %mem_1.addr0, %mem_1.clk, ... = calyx.seq_mem @mem_1 <[1] x 32> [1] {external = true}
       ...
       calyx.control {
         calyx.seq {
-          calyx.invoke @id(%id.in = %c1_10, %r.in = %id.out) -> (i32, i32)
+          calyx.invoke @id[mem = mem_1](%id.in = %c1_10, %r.in = %id.out) -> (i32, i32)
```
cgyurgyik, line 444: Can you update the native Calyx emitter and add a test please?

cgyurgyik:
Nice! A few small comments and request for tests.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: Not really, the function signature is `FuncOp createNewTopLevelFn...`. We can't really return `emitError`

cgyurgyik: You're right, apologies; I was thinking about LogicalResult or InFlightDiagnostic, where [you can do something like this](https://github.com/llvm/llvm-project/blob/39f78464d5f8794bc836b5e15c92fd1e162a3d0b/mlir/include/mlir/Dialect/Transform/Interfaces/TransformInterfaces.h#L1601).

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: We need to change it in-place so that after setting the top-level function [here](https://github.com/llvm/circt/blob/db509755c08048ecac1f740ed06e4ed6f2e88e5a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp#L1855), we can directly use `topLevelFunction` [in `labelEntryPoint`](https://github.com/llvm/circt/blob/db509755c08048ecac1f740ed06e4ed6f2e88e5a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp#L1861) as well

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: Can you elaborate a bit more on this?
Like when we have a mix of different types of function arguments, out of which some are `memrefs` and some are not? An how can I check an `unsigned` variable here?

cgyurgyik: I meant to say you've added changes to the SCF to Calyx pass, but is there a test that confirms the behavior of what you're trying to do?

jiahanxie353: No I haven't, will push soon!
I might have asked this, but I've been wondering if there is any tool to generate `CHECK: ...`/`CHECK-LABEL:...`?

lib/Dialect/Calyx/CalyxOps.cpp
jiahanxie353: Round-trip? From SCF to Calyx and from Calyx back to?

cgyurgyik: There should be a round-trip.mlir test file (or something similar, not on my laptop) that just verified that printing and parsing works as expected

include/circt/Dialect/Calyx/CalyxControl.td
cgyurgyik: Hey @jiahanxie353 do you mind resolving these comments if you addressed the issues, or following up if you want some clarification / disagree with something? Thank you!

jiahanxie353: Oops sorry I didn't check in in the first place!

Thanks for reviewing and reminding me @cgyurgyik !
Hmm, I thought I have emitter check [here](https://github.com/llvm/circt/blob/9af79669551d9b8bf09f258c5256df4d4a6494f9/test/Dialect/Calyx/emit-invoke-with-ref.mlir#L61). Is this not we are looking for?

cgyurgyik: Right, so if you addressed the comment, can you mark this conversation as resolved?

cgyurgyik: It just makes it easier to determine whether you've addressed it or not.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -84,6 +84,7 @@ void buildAssignmentsForRegisterWrite(OpBuilder &builder,
 // A structure representing a set of ports which act as a memory interface for
 // external memories.
 struct MemoryPortsImpl {
+  StringRef memName;
```
cgyurgyik, line 87: Can you guarantee that MemoryPortsImpl will never outlive memName? If not, I think the safe choice is just using a `std::string`.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1391,13 +1389,54 @@ class BuildControl : public calyx::FuncOpPartialLoweringPattern {
...
+                      instanceOpLoweringState->getMemoryInterface(argI)
+                          .memName()),
+                  memOpNameAttr);
+              refCells.push_back(
+                  DictionaryAttr::get(rewriter.getContext(), namedAttrList));
+            }
+          } else {
+            inputPorts.push_back(operand);
+          }
+        }
```
cgyurgyik, line 1430:
```suggestion
        for (auto operandEnum : enumerate(callSchedPtr->callOp.getOperands())) {
          auto operand = operandEnum.value();
          auto index = operandEnum.index();
          if (!isa<MemRefType>(operand.getType())) {
            inputPorts.push_back(operand);
            continue;
          }
          auto memOpName = getState<ComponentLoweringState>()
                                 .getMemoryInterface(operand)
                                 .memName();
          auto memOpNameAttr =
                SymbolRefAttr::get(rewriter.getContext(), memOpName);
          Value argI = calleeFunc.getArgument(index);
          if (isa<MemRefType>(argI.getType())) {
            NamedAttrList namedAttrList;
            namedAttrList.append(
                rewriter.getStringAttr(
                    instanceOpLoweringState->getMemoryInterface(argI)
                        .memName()),
                memOpNameAttr);
            refCells.push_back(
                DictionaryAttr::get(rewriter.getContext(), namedAttrList));
          }
        }
```

cgyurgyik: We should just use an early exist in the case where this isn't a MemRefType. 

Ref: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1723,6 +1763,177 @@ class SCFToCalyxPass : public circt::impl::SCFToCalyxBase<SCFToCalyxPass> {
...
+    Block *calleeFnBody = &callee.getBody().front();
+    unsigned originalCalleeArgNum = callee.getArguments().size();
+
+    SmallVector<Value, 4> extraMemRefArgs;
+    SmallVector<Type, 4> extraMemRefArgTypes;
+    SmallVector<Value, 4> extraMemRefOperands;
+    SmallVector<Operation *, 4> opsToModify;
+    for (auto &op : callee.getBody().getOps()) {
+      if (isa<memref::AllocaOp>(op) || isa<memref::AllocOp>(op) ||
+          isa<memref::GetGlobalOp>(op))
```
cgyurgyik, line 1834:
```suggestion
      if (isa<memref::AllocaOp, memref::AllocOp, memref::GetGlobalOp>(op))
```

^ I think this should work.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1723,6 +1763,177 @@ class SCFToCalyxPass : public circt::impl::SCFToCalyxBase<SCFToCalyxPass> {
...
+    /// function from the new one if the original top-level has `memref` in its
+    /// argument
+    bool hasMemrefArgsInTopLevel = false;
+    for (auto funcOp : moduleOp.getOps<FuncOp>()) {
+      if (funcOp.getName() == topLevelFunction) {
+        if (hasMemrefArguments(funcOp)) {
+          hasMemrefArgsInTopLevel = true;
+        }
+      }
+    }
```
cgyurgyik, line 1918: This can be simplified (and short-circuited) using std::any_of, e.g.,

```cpp
funcOps = moduleOp.getOps<FuncOp>();
bool hasMemrefArgsInTopLevel = 
  std::any_of(
    funcOps.begin(), 
    funcOps.end(), 
    [](auto funcOp) { 
      return funcOp.getName() == topLevelFunction && hasMemrefArguments(funcOp);
    });
```
Ref: https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1723,6 +1763,177 @@ class SCFToCalyxPass : public circt::impl::SCFToCalyxBase<SCFToCalyxPass> {
...
+    if (hasMemrefArgsInTopLevel) {
+      auto newTopLevelFunc = createNewTopLevelFn(moduleOp, topLevelFunction);
+
+      OpBuilder builder(moduleOp.getContext());
+      Operation *oldTopLevelFuncOp =
+          SymbolTable::lookupSymbolIn(moduleOp, oldName);
+      auto oldTopLevelFunc = dyn_cast_or_null<FuncOp>(oldTopLevelFuncOp);
+
+      if (!oldTopLevelFunc)
+        oldTopLevelFunc.emitOpError("Original top-level function not found!");
```
cgyurgyik, line 1930: 1. Can we be explicit and change this to `oldTopLevelFunc == nullptr` please? 
2. This should this return right? I don't think you want to pass a nullptr to `insertCallFromNewTopLevel`. Can you add an error test that catches this?

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -2747,10 +2783,12 @@ LogicalResult InvokeOp::verify() {
     return emitOpError() << "with instance '@" << callee
                          << "', which does not exist.";
   // The argument list of invoke is empty.
-  if (getInputs().empty())
+  if (getInputs().empty() && getRefCellsMap().empty()) {
     return emitOpError() << "'@" << callee
                          << "' has zero input and output port connections; "
+                            " and has not passing-by-reference cells; "
```
cgyurgyik, line 2789: I'm not totally following the logic here, but I also don't know why an invoke with empty argument is illegal either :)

cgyurgyik (commented):
Nice! A few smaller coding-related comments.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: I was imagining this case:
```mlir
module {
  func.func @main() : () -> i32 {
     %c0 = constant 0 : i32
     %c1 = constant 1 : i32
     %1 = arith.addi %c0, %c1 : i32
    return %1
  }
  func.func @func_1(%arg0 : i32, %mem0 : memref<8xi32>, %i : index, %mem1 : memref<8xi32>) {
    memref.store %arg0, %mem1[%i] : memref<8xi32>
    memref.store %arg0, %mem0[%i] : memref<8xi32>
    return
  }
}
```
where the `main` function has nothing to do with `func_1` and our command is `lower-scf-to-calyx="top-level-function=main`.

I was wondering, will `func_1` be executed after `main` same as how they are ordered?

cgyurgyik: Perhaps I'm misunderstanding, but one of these would be labeled as the entry point, presumably `main`. If `func_1` is not accessible from the entry point, then it is dead, i.e., it will never be executed. This is not much different than an imperative language today, e.g.,

```cpp
int main() {
  int c0 = 0, c1 = 1;
  return c0 + c1;
  // Since there is no explicit call to `func_1`, it is considered dead code.
} 

void func_1(int arg0, std::vector<int> mem0, size_t i, std::vector<int> mem1) {
  mem1[i] = arg0;
  mem0[i] = arg0;
  return;
}
```

jiahanxie353: I see, then I think I'll just do `assert(oldTopLevelFunc && "Original top-level function not found)` since the `moduleOp` will first try to catch it [here](https://github.com/llvm/circt/blob/17c036f87c4c9e103160b207e18ad595911945e8/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp#L1818) if the top-level function is not set when there are multiple functions. And [this line](https://github.com/llvm/circt/blob/ad89e27ebabb2e50120a3d7468b9e8d738f9c6d5/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp#L1981) will also catch the case when the new top-level is not created properly

So actually couldn't come up the edge case where it will raise `oldTopLevelFunc.emitOpError("Original top-level function not found!")`

lib/Dialect/Calyx/CalyxOps.cpp
jiahanxie353: done!

cgyurgyik (commented):
A few smaller comments! Let me know if I'm misunderstanding something.

Issue #3962, opened 2022-09-21
Move to new pass generation approach


Dinistro:
Recently, table-gen was changed to split the emissions of passes into different chunks: https://reviews.llvm.org/D131839

This change not only splits a large monolith into smaller pieces but also adds support for generating pass constructors that expect pass options. The latter will be useful for tools like `hlstool` as many passes do currently not have constructures that take their options.

PR #7168, opened 2024-06-13
[ExportVerilog] Avoid using interface pass for PrepareForEmission, NFCI

uenoku:
ODS InterfacePass generates `canScheduleOn` method with a specified interface in a header file and currently we import every pass declaration through PassesDetail.h. Gcc/clang seem to compile when there is an unknown interface class but MSVC emits an error. So this PR avoids `InterfacePass` in ODS and manually implements `canScheduleOn` in PrepareForEmission. 

Windows CI: https://github.com/llvm/circt/actions/runs/9497120978

### DIFF
--- a/include/circt/Conversion/Passes.td
+++ b/include/circt/Conversion/Passes.td
@@ -118,8 +118,7 @@ def HWLowerInstanceChoices : Pass<"hw-lower-instance-choices",
   ];
 }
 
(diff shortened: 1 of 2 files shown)


uenoku:
I'm going to merge the PR to fix windows CI. I think a right fix is to stop using `GEN_PASS_DECL` in `PassDetail.h`. See https://github.com/llvm/circt/issues/3962.

Issue #7169, opened 2024-06-13
[FIRRTL] Dialect/FIRRTL/lower-dpi-error.mlir failed in nightly integration tests


uenoku:
https://github.com/llvm/circt/actions/runs/9499212780/job/26179636823

uenoku:
Apparently there is an UAF for DPI call op ;)

PR #7139, opened 2024-06-05
[FIRRTL] Add DPI call intrinsic and lowering pass

uenoku:
This PR adds DPICallIntrinsicOp and its lowering pass. DPICallIntrinsicOp is lowered into sim.func.dpi.call and sim.func.dpi ops. At FIRRTL level DPICallIntrinsicOp doesn't have symbols and instead LowerDPI pass accumulates callsites and creates symbols for dpi functions. LowerDPI pass directly lowers FIRRTL intrinsic into Sim dialect since FIRRTL doesn't have DPI/Function construct. LowerDPI pass could be simplified (or migrated into LowerToHW) once FIRRTL gets 1st class support for Function. Unrealized conversion cast is used to mix FIRRTL and HW type values before LowerToHW.

PR #7170, opened 2024-06-13
[LowerDPI] Defer deletion of call ops to prevent inavlid access.

fzi-hielscher:
The [lower-dpi-error.mlir](https://github.com/llvm/circt/pull/7139/files#diff-b9efcb02f7785b4bd52265c695cc5f2a6b00750101ea9abcd346e8bc161cd6db) test added in #7139 crashes on my local macOS debug build. This appears to be caused by [an invalid access in the lowering loop](https://github.com/llvm/circt/pull/7139/files#diff-8897e1920c98c76b3006af5bfe97e5313243e455253463acb99ffa55b924ef96R137). The iterator stored in `inputTypes` references the operand range of the first call op. However, after calling `lowerCall(firstDPICallop);` this op has been deleted. Deferring op erasure until after the current batch of call operations has been checked and lowered has fixed the problem for me.

Close https://github.com/llvm/circt/issues/7169

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerDPI.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerDPI.cpp
@@ -126,7 +126,6 @@ void LowerDPIPass::runOnOperation() {
                           ->getResult(0);
         dpiOp.getResult().replaceAllUsesWith(result);
       }
-      dpiOp.erase();
     };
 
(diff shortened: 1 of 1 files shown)


uenoku (approved):
My bad, thanks! I have just noticed that issue :bow: https://github.com/llvm/circt/issues/7169

PR #7172, opened 2024-06-13
[NFC][ExportVerilog] Rename generated `options` member.

fzi-hielscher:
Side-step an issue that prevents transitioning the inclusion of TableGen defined passes to the new(er) style, see #7168 #3962.

Currently, in the generated constructor for `TestApplyLoweringOptionBase` the formal parameter is shadowing the `options` member. This won't compile if we try to include it:
```
TestApplyLoweringOptionBase(const TestApplyLoweringOptionOptions &options) : TestApplyLoweringOptionBase() {
    options = options.options;
  }
```

Arguably, this could be fixed in upstream by at least picking a less obvious formal parameter name. But I think this tiny change isn't hurting us.

### DIFF
--- a/include/circt/Conversion/Passes.td
+++ b/include/circt/Conversion/Passes.td
@@ -88,7 +88,7 @@ def TestApplyLoweringOption : Pass<"test-apply-lowering-options",
     "circt::sv::SVDialect", "circt::comb::CombDialect", "circt::hw::HWDialect"
   ];
   let options = [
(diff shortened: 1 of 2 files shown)


uenoku (approved):
Looks reasonable to me!

fzi-hielscher:
Thanks. And sorry for the barrage of PRs. I'm afraid I have some more lined up. 😅

PR #7173, opened 2024-06-13
[NFCI][Transforms] Refactor TableGen Pass includes

fzi-hielscher:
Refactor includes of TableGen passes in the "Transforms" sub-tree to the new style.
See #3962 , https://reviews.llvm.org/D143773

See also #7174


uenoku (approved):
LGTM

PR #7175, opened 2024-06-13
[Verif] Add PrepareForFormal pass

dobios:
This PR introduces a `PrepareForFormal` pass that tries to alleviate some of the heavy lifting being done in the btor emission. For now this only flattens wires, but in the future in will also handle formal contracts and anything else that I do to make emission easier, which is why it's part of the `verif` dialect and not `hw`.

This is derived from #7150


lib/Dialect/Verif/Transforms/PrepareForFormal.cpp
```diff
@@ -0,0 +1,69 @@
...
+/// through canonicalization at some point.Some wires are however
+/// maintained.This pass unconditionally replaces all wires with their inputs,
+/// making it easier to reason about in contexts where wires don't exist.
+struct WireOpConversionPattern : OpConversionPattern<hw::WireOp> {
+  using OpConversionPattern<hw::WireOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(hw::WireOp wire, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOp(wire, adaptor.getInput());
```
uenoku, line 40: I'm not blocking but note that a wire could have an inner symbol. Verifier after the pass will fail if there is inner symbol user of it so it will be necessary to remove them as well in this pass.

lib/Dialect/Verif/Transforms/PrepareForFormal.cpp
```diff
@@ -0,0 +1,69 @@
+//===- FlattenWires.cpp - Flattens all wires ------*- C++ -*-===//
```
uenoku, line 1: Nit: file name.

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -616,13 +584,10 @@ struct ConvertHWToBTOR2Pass
...
-  // Wires can generally be ignored in bto2, however we do need
-  // to keep track of the new alias it creates
+  // Wires should have been removed in PrepareForFormal
   void visit(hw::WireOp op) {
-    // Retrieve the aliased operation
-    Operation *defOp = op.getOperand().getDefiningOp();
-    // Wires don't output anything so just record alias
-    setOpAlias(op, defOp);
+    op->emitError("Wires are not supported in btor!");
+    return signalPassFailure();
```
uenoku, line 590: Not blocking but it would be nicer to return LogicalResult::failure() instead of void in the future. Currently pass continues even after signalPassFailure();

uenoku (approved):
LGTM

test/Dialect/Verif/prepare-for-formal.mlir
```diff
@@ -0,0 +1,45 @@
...
+    //CHECK:  %c-2_i2 = hw.constant -2 : i2
+    %c-2_i2 = hw.constant -2 : i2
+    //CHECK:  %count = seq.compreg %6, %clock reset %reset, %c0_i32 : i32  
+    %count = seq.compreg %6, %clock reset %reset, %c0_i32 : i32
+    %wf = hw.wire %false sym @f : i1
+    %b = hw.wire %2 sym @b : i32
+    %d = hw.wire %b sym @d : i32
+    %c = hw.wire %5 sym @c : i32
+    %w_in = hw.wire %in : i1
+    %ww_in = hw.wire %w_in sym @in_w : i1
```
uenoku, line 21: Please add `CHECK-NOT: hw.wire`.

lib/Dialect/Verif/Transforms/PrepareForFormal.cpp
dtzSiFive: (nit: after fixing filename, pad with dashes until hits 80chars like the others-- sometimes helps to just copy a different one next to this and pad until they line up 👍 )

lib/Dialect/Verif/Transforms/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 add_circt_dialect_library(CIRCTVerifTransforms
   VerifyClockedAssertLike.cpp
+  PrepareForFormal.cpp
```
dtzSiFive, line 3: Nit: alpha-sort.

lib/Dialect/Verif/Transforms/PrepareForFormal.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "mlir/Pass/Pass.h"
+
+using namespace circt;
+using namespace mlir;
+using namespace verif;
+
+namespace {
+/// Flatten wires by unconditionally replacing them with their inputs
+/// Wires create an alias for a set of operations, they are usually removed
+/// through canonicalization at some point.Some wires are however
```
dtzSiFive, line 31:
```suggestion
/// through canonicalization at some point. Some wires are however
```
(idc how many spaces but should have some after a sentence-terminating '.')

"Flatten" seems a little overloaded -- not here, but across other passes (FlattenIO, FlattenModules, ..).  Works for me but just a small note.

test/Dialect/Verif/prepare-for-formal.mlir
```diff
@@ -0,0 +1,45 @@
+// RUN: circt-opt --prepare-for-formal %s | FileCheck %s
+module {
+  // CHECK-LABEL: @FlattenWires
+  hw.module @FlattenWires(in %clock : !seq.clock, in %reset : i1, in %in : i1, out out : i32) {
+
+    //CHECK:  %false = hw.constant false
```
dtzSiFive, line 6: Hyper-nit: Please change this to `// CHECK`, for consistency with the `CHECK-LABEL` above and in other tests.

A quick grep shows there's 412 instances of that (`//CHECK` presently vs 27745 with `// CHECK` .

include/circt/Dialect/Verif/Passes.td
```diff
@@ -27,4 +27,19 @@ def VerifyClockedAssertLike : Pass<"verify-clocked-assert-like", "hw::HWModuleOp
   let constructor = "circt::verif::createVerifyClockedAssertLikePass()";
 }
 
+/**
```
dtzSiFive, line 30: Nit: Alpha-sort?

I'm not sure where this pass best belongs, but it's not obvious it's in the `Verif` dialect since it presently doesn't touch/use anything from that dialect.  No objection but just a note.

dobios: It doesn't now, but my goal for this pass is for it to also handle the verif contract replacement stuff, which which will be most of what it does. Also it seemed weird to have something specific to modifying a core representation to make it easier for formal verification to be anywhere other than verif.

dtzSiFive: 100% sounds great to me! Thanks for the consideration / care :+1:.

PR #7177, opened 2024-06-14
[FIRRTL][ExpandWhens] Add StmtExprVisitor to Visitor and Support DPI intrinsic in ExpandWhens

uenoku:
This adds `StmtExprVisitor` struct for visitor to handle operations with an optional result. Currently `GenericIntrinsicOp `and `DPICallIntrinsicOp` are added.

Besides that `ExpandWhens` is modified to handle DPI intrinsic.


dtzSiFive (approved):
LGTM!

Whether this is how when + DPI enable "should" work I don't know but seems reasonable and aligns with assert/assume/cover/printf so works for me.

PR #7178, opened 2024-06-14
[NFCI][FIRRTL] Refactor TableGen Pass includes

fzi-hielscher:
Refactor includes of TableGen passes in the FIRRTL dialect to the new style.

Some manual code movements in the Passes header. Other than that purely mechanical changes.


uenoku (approved):
LGTM, thanks!

PR #7179, opened 2024-06-14
[NFCI][ESI][Ibis][MSFT] Refactor TableGen Pass includes

fzi-hielscher:
Refactor includes of TableGen passes in the ESI, Ibis and MSFT dialects to the new style.

Mostly mechanical changes. The `PassDetails.h` for ESI and MSFT are retained, as they contain some non-trivial code. We might consider renaming them.


teqdruid (approved):
Thanks!

PR #7180, opened 2024-06-14
[NFCI][Comb][HW][Seq] Refactor TableGen Pass includes

fzi-hielscher:
Refactor includes of TableGen passes in the Comb, HW and Seq dialects to the new style.


teqdruid (approved):
lgtm. thanks!

PR #7182, opened 2024-06-14
[NFCI][Calyx] Refactor TableGen Pass includes

fzi-hielscher:
Refactor includes of TableGen passes in the Calyx dialect to the new style.


cgyurgyik (approved):
Thank you!

PR #7186, opened 2024-06-15
[ESI Runtime] Read ports now invoke callbacks

teqdruid:
We've switched from a polling 'pull' method to a callback-based 'push'
mechanism for read ports. Polling (via std::futures) is built on top of
the push mechanism.

The read-ordering problem has also been fixed by using std::futures
exclusively for polling schemes. They also allow for poll-wait-notify
schemes without any changes on our part.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
+++ b/lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
@@ -18,7 +18,9 @@
 
 #include "esi/Common.h"
 #include "esi/Types.h"
+#include "esi/Utils.h"
 
+#include <cassert>
 #include <future>
 
 namespace esi {
@@ -31,15 +33,19 @@ namespace esi {
 class ChannelPort {
 public:
   ChannelPort(const Type *type) : type(type) {}
-  virtual ~ChannelPort() = default;
+  virtual ~ChannelPort() { disconnect(); }
 
-  virtual void connect() {}
+  virtual void connect() { connectImpl(); }
   virtual void disconnect() {}
 
   const Type *getType() const { return type; }
 
 private:
   const Type *type;
+
+  /// Called by all connect methods to let backends initiate the underlying
+  /// connections.
+  virtual void connectImpl() {}
 };
 
 /// A ChannelPort which sends data to the accelerator.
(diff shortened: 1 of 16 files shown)


mortbopet (commented):
AFAICT this solves the multiple-readers problem, but doesn't solve the function service call problem.

The following code still does not guarantee that calls to a function is ordered wrt. the order of which the argument port was called
https://github.com/llvm/circt/blob/18d2872d70dfbc69e0e5bd597d6e89fc51a214e0/lib/Dialect/ESI/runtime/cpp/lib/Services.cpp#L118-L119

I.e. the following interleaving of two calls C1 and C2 is possible
> C1 arg.write
> C2 arg.write
> C2 result.readAsync
> C1 result.readAsync

So there must be a lock added to `FuncService::Function::call` to ensure that `readAsync` is always called for function call `n` before function call `n+1` occurs.

lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
```diff
@@ -54,18 +55,59 @@ class WriteChannelPort : public ChannelPort {
...
+    if (!promiseQueue.push(p)) {
+      throw std::runtime_error("Channel is disconnected");
+      delete p;
+    }
+    return p->get_future();
+  }
+
+  /// Specify a buffer to read into. Blocking. Basic API, will likely change
+  /// for performance and functionality reasons.
+  virtual void read(MessageData &outData) {
```
mortbopet, line 90: `bool read(MessageData &)` -> `void read(MessageData &outData)` API change makes this PR not compile when ESI runtime is built w/ the new callback service (used here: https://github.com/llvm/circt/pull/7153/files#diff-0432614eceb5e1876ca0ea5dbda0d10f9d36c4a5e0e66e8d7c96380bd618256eR116)

mortbopet: Here's a fix i'm using internally:

```c++
  std::map<ReadChannelPort *, std::future<MessageData>> activeRead;

  while (!shutdown) {
    std::this_thread::sleep_for(std::chrono::microseconds(100));
    {
      std::lock_guard<std::mutex> g(listenerMutex);
      for (auto &[channel, cb] : listeners) {
        assert(channel && "Null channel in listener list");
        auto it = activeRead.find(channel);
        if (it == activeRead.end()) {
          activeRead[channel] = channel->readAsync();
        } else {
          auto &f = it->second;
          bool is_ready =
              f.wait_for(std::chrono::seconds(0)) == std::future_status::ready;
          if (!is_ready)
            continue;
          portUnlockWorkList.emplace_back(channel, cb, std::move(f.get()));
          activeRead.erase(it);
        }
      }
    }

...

Comment (teqdruid) [reply lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h:90]: Yeah, there's definitely at least head of line blocking in the service thread of the other PR.

**Comment by teqdruid** [comment]
@mortbopet: if you're not blocked on this, I'm actually gonna finish the gRPC conversion first so that I don't have to screw around with the capnp backend to adapt it to the changes we discussed.

**Comment by teqdruid** [comment]
I think switching to callbacks made things more elegant.

File: lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h [hunk 2]
```diff
@@ -51,21 +57,77 @@ class WriteChannelPort : public ChannelPort {
...
+/// A ChannelPort which reads data from the accelerator. It has two modes:
+/// Callback and Polling which cannot be used at the same time. The mode is set
+/// at connect() time. To change the mode, disconnect() and then connect()
+/// again.
 class ReadChannelPort : public ChannelPort {
+
 public:
-  using ChannelPort::ChannelPort;
+  ReadChannelPort(const Type *type) : ChannelPort(type) {}
+  virtual void disconnect() override { mode = Mode::Disconnected; }
```
mortbopet, line 68: With these split modes, and different modes having different member variables associated with them, i'd personally split those into two separate `ReadChannelPort`-derived classes to avoid any confusion/misuse/leak between the two API modes.

mortbopet (approved):

lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
teqdruid: This is something I considered. The problem is that these channel ports are constructed when the system gets built from the manifest, not when the user locates them. So the user doesn't have the chance to specify the type before getting the reference to them.

PR #7187, opened 2024-06-15
Add missing header includes

rwy7:
To fix our windows builds.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRParser.h
+++ b/include/circt/Dialect/FIRRTL/FIRParser.h
@@ -15,6 +15,8 @@
 #define CIRCT_DIALECT_FIRRTL_FIRPARSER_H
 
 #include "circt/Support/LLVM.h"
+#include <string>
+#include <vector>
 
 namespace llvm {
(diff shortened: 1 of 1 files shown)


seldridge (commented):
This makes sense. However, can the changes introduced in c93ca4ecf4cf5d707530435c3f614732309afa42 instead LLVM-internal libraries and _not_ use `std::vector` and `std::string`?

seldridge (approved):
Regardless of the above, this fixes the Windows build. This is fine to land and sort out the use of internal libraries later.

include/circt/Dialect/FIRRTL/FIRParser.h
```diff
@@ -15,6 +15,8 @@
 #define CIRCT_DIALECT_FIRRTL_FIRPARSER_H
 
 #include "circt/Support/LLVM.h"
+#include <string>
+#include <vector>
```
seldridge, line 19: One alternative to this is to forward declare these as opposed to an actual include. That may work here.

PR #7189, opened 2024-06-16
[ExportVerilog] Fix two state type emission of aggregate types

uenoku:
This PR fixes an issue that `emitAsTwoStateType` flag was not propagated for aggregate types.

This PR also fixes a bug that ctypes (int/shortint etc) are used as an inner type of packed types (e.g. `int [2:0]` is invalid) according to SV spec 6.8.


uenoku:
Since this is small fix I'm going to merge. Post-commit review is welcomed!

PR #7195, opened 2024-06-17
[HW] Move the CombDataFlow op interface from FIRRTL to HW

prithayan:
Move the `CombDataFlow` op interface from `FIRRTL` to `HW` dialect.
The op interface is better suited to reside in the `HW` dialect, along with other interfaces like the `HWModuleLike`.
This makes it more convenient for non-`FIRRTL` dialects to implement ops using this interface, without introducing a new dependence on the `FIRRTL` dialect, assuming  `HW` dependence already exists.


uenoku (commented):
Looks fine. I'm bit worried that this belongs to HW since the interface is only used by FIRRTL passes and FieldRef is not actually used in HW but I agree the concept itself would fit better in HW.

uenoku (approved):
Could you open an issue for modifying default implementation not to assume sequential?  LGTM otherwise

prithayan:
Thanks @uenoku , I will create the issue.

PR #7196, opened 2024-06-17
Update clang to 17.

darthscsi:
clang-format-17 makes a few different choices than clang-format-13.  v13 is almost 3 years old.  Given that slang builds are already installing v16, let's just unify on v17 for CI.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -14,7 +14,7 @@ jobs:
     name: Sanity Check
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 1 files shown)


teqdruid:
I think it's important to run as old a version of the compiler which we want to support. I run clang-10 and Python 3.8 for exactly that reason, so we are already running a newer version than we support. If we want to require a newer version, that's a discussion. I would like to support ancient versions since some EDA build farms run pretty ancient distros.

darthscsi:
We are still running gcc-11 which, while not EDA-build-farm-ancient, is still older than clang-13.  I suspect the primary control on older compilers is the language level we require.

PR #7198, opened 2024-06-18
[FIRRTL][CheckCombLoops] Verify that detection works with region ops

youngar:


test/Dialect/FIRRTL/check-comb-loops.mlir
```diff
@@ -1053,3 +1053,57 @@ firrtl.circuit "Issue6820" {
...
+    firrtl.layerblock @A {
+      %w = firrtl.wire  : !firrtl.uint<8>
+      firrtl.connect %w, %w : !firrtl.uint<8>, !firrtl.uint<8>
+    }
+  }
+}
+
+// -----
+
+// Single-element combinational loop with multiple connects.
```
dtzSiFive, line 1101: This is more about last-connect than checking re:region ops but good to have it as well.

Distills that behavior separate from interaction/handling of things like `when`.  :+1:

dtzSiFive (approved):
LGTM, thanks for putting this behavior under test!

darthscsi (approved):

PR #7200, opened 2024-06-18
[FIRRTL][SpecializeLayers] Fix incorrect CF leading to double free

youngar:
As a part of specializing layers we have to remove any HierPathOps which included a reference to deleted instances.  If any member of the path array is contained in the deleted references list, we need to delete the op.  There was incorrect use of `continue` which caused us to continue processing the path instead of skipping to the next path operation, which could lead to a double free when multiple instances in the path were removed.


dtzSiFive (approved):
LGTM!

PR #7202, opened 2024-06-18
[Moore] Support unconnected behavior

cepheus69:
See IEEE 1800 - 2017  § 23.3.3 Port connection rules. We could get behaviors corresponding to different port types including the rule of initiating default value for unconnected ports. 
| Port Type        | Unconnected Behavior       |
| ---------------- | -------------------------- |
| Input (Net)      | High-impedance value ('Z)  |
| Input (Variable) | Default initial value      |
| Output           | No effect on Simulation    |
| Inout (Net)      | High-impedance value ('Z)  |
| Inout (Variable) | Default initial value      |
| Ref              | Cannot be left unconnected |
| Interconnect     | Cannot be left unconnected |
| Interface        | Cannot be left unconnected |


fabianschuiki (commented):
Great idea to support this! I don't think you need a new `UnconnectedOp` though: you should be able to just create a `moore.net` for net-typed ports (input net, inout), or a `moore.variable` for var-typed ports (input var, ref). You should be able to create a `moore.{net,variable}` without a name, and without any initial assignment. That will create a default value according to the table you've documented -- and it will allow you to connect inout stuff as well.

Your table documenting default values is great by the way. You could add that to `docs/Dialects/Moore.md` 👍

cepheus69:
> Great idea to support this! I don't think you need a new `UnconnectedOp` though: you should be able to just create a `moore.net` for net-typed ports (input net, inout), or a `moore.variable` for var-typed ports (input var, ref). You should be able to create a `moore.{net,variable}` without a name, and without any initial assignment. That will create a default value according to the table you've documented -- and it will allow you to connect inout stuff as well.
> 
> Your table documenting default values is great by the way. You could add that to `docs/Dialects/Moore.md` 👍

Okay. I'll take it in my mind.

cepheus69:
So additional logic is needed by net / varOp to decide how to deal with two meanings of themselves.

cepheus69:
Already changed to net/var creation

fabianschuiki (approved):
Very nice! Thanks for adding this 🥳

PR #7203, opened 2024-06-18
[FIRRTL][NFC] Fast-path removeAnnotations for operations having none.

dtzSiFive:
Common pattern of walk/visit many operations looking for annotations and removing along the way is taking considerable amount of time in ArrayAttr::get.  Optimize the fast path for operations having no annotations.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLAnnotations.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLAnnotations.cpp
@@ -25,8 +25,12 @@
 using namespace circt;
 using namespace firrtl;
 
+static ArrayAttr getAnnotationsIfPresent(Operation *op) {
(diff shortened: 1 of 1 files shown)


rwy7 (approved):

darthscsi (approved):
LGTM

youngar:
I think there are a couple of places that have this code inlined, as not doing this attaches an empty annotation attribute which I usually only catch when I am writing tests. e.g. very recently: https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp#L66

youngar:
I think I misunderstood what this code is doing - just creating an empty annotation set creates an empty array attribute, and that is what we're trying to avoid? 🤦

PR #7204, opened 2024-06-18
[Seq] Add a new op interface for clock transformation ops. NFC

prithayan:
Add the clock gate, clock inverter and clock divider ops to a new clock transformation interface.
This makes it easier to handle them and get the input clock directly.

### DIFF
--- a/include/circt/Dialect/Seq/SeqOpInterfaces.td
+++ b/include/circt/Dialect/Seq/SeqOpInterfaces.td
@@ -24,6 +24,15 @@ def Clocked : OpInterface<"Clocked"> {
   ];
 }
 
+def ClockTransformation : OpInterface<"ClockTransformation"> {
+  let cppNamespace = "circt::seq";
+  let description = "Trait for a clock transformation operation";
+
+  let methods = [
+    InterfaceMethod<"Get the input clock signal", "Value", "getClk", (ins)>
+  ];
+}
+
 def Resettable : OpInterface<"Resettable"> {
   let cppNamespace = "circt::seq";
   let description = "Trait for Resettable elements";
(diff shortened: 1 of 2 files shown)


uenoku (commented):
Does `ClockMuxOp` need to have this interface?  In that case `getClk` returning a single value seems not composable.

prithayan:
> Does `ClockMuxOp` need to have this interface? In that case `getClk` returning a single value seems not composable.

Yeah, I was not sure about clock mux, for the use case that I was considering, `ClockMuxOp`  needs to be handled separately, unlike the others .  But it makes sense to include it as a clock transformation op, and let the client deal with it, if it returns multiple input clocks. I will update the interface to include mux ops.

include/circt/Dialect/Seq/SeqOpInterfaces.td
```diff
@@ -24,6 +24,15 @@ def Clocked : OpInterface<"Clocked"> {
   ];
 }
 
+def ClockTransformation : OpInterface<"ClockTransformation"> {
+  let cppNamespace = "circt::seq";
+  let description = "Trait for a clock transformation operation";
+
+  let methods = [
+    InterfaceMethod<"Get the input clock signal", "Value", "getClk", (ins)>
```
darthscsi, line 32: Is a single value here correct?

include/circt/Dialect/Seq/SeqOpInterfaces.td
```diff
@@ -24,6 +24,15 @@ def Clocked : OpInterface<"Clocked"> {
   ];
 }
 
+def ClockTransformation : OpInterface<"ClockTransformation"> {
```
darthscsi, line 27: Do we need a trait or just a function?  Are other dialects getting in on the action?  Are there times we just care about `ClockTransformation` ops but not what they do?

prithayan:
Closing, as this can be accomplished via a simple function, the itnerface is not relevant.

PR #7206, opened 2024-06-19
[FIRRTL][FIRParser] Tokenize Commas

mmaloney-sf:
Due to historical circumstances, the SFC parser for FIRRTL treated commas (`,`) as whitespace. This behavior was carried over into the CIRCT implementation of `firtool`.

This behavior is utterly bizarre. 

Moreover, the FIRRTL spec has indicated comma tokens (`,`) for at least every version since last year. 

This PR removes this quirk and properly tokenizes commas, requiring them in the places dictated by the FIRRTL spec.

The impact of this PR should be low, but it may break code (in CIRCT, in Chisel, and elsewhere) in places where FIRRTL was being emitted with a loose interpretation, or where the spec was ambiguous.

Changes:
* Update several tests which were leaning into the "commas-are-whitespace" behavior.
* Removed `','` from the list of "horizontal whitespace" and added an explicit `FIRToken::comma` token.
* Made changes to the various `parse*()` methods to consume the tokens.
* Note the use of a pattern I found useful in several places where I use a boolean `first` to skip parsing comma tokens on the first iteration. Please check me here, since I learned C++ before lambdas were a thing.
* Made a few judgment calls around unspeced constructs (`smem`).
* * I changed the FIRRTL emitter in one place where the syntax was unspeced, and there seemed to be conflicting examples.
* @seldridge For some reason, it seemed ambiguous whether `layer` decls need commas. I opted *for* commas. I can swap if I got this backwards.
* I added `parseRUW()` as a non-optional variant of `parseOptionalRUW()`, since the comma token can be used to determine optionality in one case.

### DIFF
--- a/lib/Dialect/FIRRTL/Import/FIRParser.cpp
+++ b/lib/Dialect/FIRRTL/Import/FIRParser.cpp
@@ -248,7 +248,8 @@ struct FIRParser {
   /// output a diagnostic and return failure.
   ParseResult parseToken(FIRToken::Kind expectedToken, const Twine &message);
 
-  /// Parse a list of elements, terminated with an arbitrary token.
+  //// Parse a comma-separated list of elements, terminated with an arbitrary
+  /// token.
   ParseResult parseListUntil(FIRToken::Kind rightToken,
                              const std::function<ParseResult()> &parseElement);
 
@@ -337,14 +338,25 @@ ParseResult FIRParser::parseToken(FIRToken::Kind expectedToken,
   return emitError(message);
 }
 
-/// Parse a list of elements, terminated with an arbitrary token.
+/// Parse a comma-separated list of zero or more elements, terminated with an
+/// arbitrary token.
 ParseResult
 FIRParser::parseListUntil(FIRToken::Kind rightToken,
                           const std::function<ParseResult()> &parseElement) {
+  if (consumeIf(rightToken)) {
+    return success();
+  }
 
-  while (!consumeIf(rightToken)) {
+  if (parseElement())
+    return failure();
+
+  while (consumeIf(FIRToken::comma)) {
     if (parseElement())
       return failure();
+  }
+
+  if (parseToken(rightToken, "expected ','")) {
+    return failure();
   }
   return success();
 }
@@ -791,13 +803,7 @@ ParseResult FIRParser::parseEnumType(FIRRTLType &result) {
                  "expected leading '{|' in enumeration type"))
     return failure();
   SmallVector<FEnumType::EnumElement> elements;
-  bool first = true;
   if (parseListUntil(FIRToken::r_brace_bar, [&]() -> ParseResult {
-        if (!first) {
-          if (parseToken(FIRToken::comma, "expected ','")) {
-            return failure();
-          }
-        }
         auto fieldLoc = getToken().getLoc();
 
         // Parse the name of the tag.
@@ -819,7 +825,6 @@ ParseResult FIRParser::parseEnumType(FIRRTLType &result) {
           type = UIntType::get(getContext(), 0);
         }
         elements.emplace_back(StringAttr::get(getContext(), name), type);
-        first = false;
         return success();
       }))
     return failure();
@@ -964,18 +969,16 @@ ParseResult FIRParser::parseType(FIRRTLType &result, const Twine &message) {
 
     SmallVector<StringRef> layers;
     if (consumeIf(FIRToken::comma)) {
+      if (requireFeature({3, 2, 0}, "colored probes"))
+        return failure();
       // Probe Color
-      if (getToken().getKind() == FIRToken::identifier) {
-        if (requireFeature({3, 2, 0}, "colored probes"))
+      do {
+        StringRef layer;
+        loc = getToken().getLoc();
+        if (parseId(layer, "expected layer name"))
           return failure();
-        do {
-          StringRef layer;
-          loc = getToken().getLoc();
-          if (parseId(layer, "expected layer name"))
-            return failure();
-          layers.push_back(layer);
-        } while (consumeIf(FIRToken::period));
-      }
+        layers.push_back(layer);
+      } while (consumeIf(FIRToken::period));
     }
 
     if (!consumeIf(FIRToken::greater))
@@ -1012,14 +1015,7 @@ ParseResult FIRParser::parseType(FIRRTLType &result, const Twine &message) {
 
     SmallVector<OpenBundleType::BundleElement, 4> elements;
     bool bundleCompatible = true;
-    bool first = true;
     if (parseListUntil(FIRToken::r_brace, [&]() -> ParseResult {
-          if (!first) {
-            if (parseToken(FIRToken::comma, "expected ','")) {
-              return failure();
-            }
-          }
-
           bool isFlipped = consumeIf(FIRToken::kw_flip);
 
           StringRef fieldName;
@@ -1034,7 +1030,6 @@ ParseResult FIRParser::parseType(FIRRTLType &result, const Twine &message) {
               {StringAttr::get(getContext(), fieldName), isFlipped, type});
           bundleCompatible &= isa<BundleType::ElementType>(type);
 
-          first = false;
           return success();
         }))
       return failure();
@@ -2251,15 +2246,8 @@ ParseResult FIRStmtParser::parsePrimExp(Value &result) {
   // Parse the operands and constant integer arguments.
   SmallVector<Value, 3> operands;
   SmallVector<int64_t, 3> integers;
-  bool first = true;
 
   if (parseListUntil(FIRToken::r_paren, [&]() -> ParseResult {
-        if (!first) {
-          if (parseToken(FIRToken::comma, "expected ','")) {
-            return failure();
-          }
-        }
-
         // Handle the integer constant case if present.
         if (getToken().isAny(FIRToken::integer, FIRToken::signed_integer,
                              FIRToken::string)) {
@@ -2279,8 +2267,6 @@ ParseResult FIRStmtParser::parsePrimExp(Value &result) {
         locationProcessor.setLoc(loc);
 
         operands.push_back(operand);
-
-        first = false;
 
         return success();
       }))
@@ -2445,14 +2431,7 @@ ParseResult FIRStmtParser::parseListExp(Value &result) {
     return failure();
 
   SmallVector<Value, 3> operands;
-  bool first = true;
   if (parseListUntil(FIRToken::r_paren, [&]() -> ParseResult {
-        if (!first) {
-          if (parseToken(FIRToken::comma, "expected ','")) {
-            return failure();
-          }
-        }
-
         Value operand;
         locationProcessor.setLoc(loc);
         if (parseExp(operand, "expected expression in List expression"))
@@ -2468,7 +2447,6 @@ ParseResult FIRStmtParser::parseListExp(Value &result) {
         }
 
         operands.push_back(operand);
-        first = false;
         return success();
       }))
     return failure();
@@ -3427,24 +3405,21 @@ ParseResult FIRStmtParser::parseIntrinsic(Value &result, bool isStatement) {
     return emitError("expected ':' in intrinsic expression");
 
   SmallVector<Value> operands;
-  bool first = true;
   auto loc = startTok.getLoc();
-  if (parseListUntil(FIRToken::r_paren, [&]() -> ParseResult {
-        if (first) {
-          if (parseToken(FIRToken::comma, "expected ','")) {
+  if (consumeIf(FIRToken::comma)) {
+    if (parseListUntil(FIRToken::r_paren, [&]() -> ParseResult {
+          Value operand;
+          if (parseExp(operand, "expected operand in intrinsic"))
             return failure();
-          }
-        }
-
-        Value operand;
-        if (parseExp(operand, "expected operand in intrinsic"))
-          return failure();
-        operands.push_back(operand);
-        locationProcessor.setLoc(loc);
-        return success();
-        first = false;
-      }))
-    return failure();
+          operands.push_back(operand);
+          locationProcessor.setLoc(loc);
+          return success();
+        }))
+      return failure();
+  } else {
+    if (parseToken(FIRToken::r_paren, "expected ')' in intrinsic"))
+      return failure();
+  }
 
   if (isStatement)
     if (parseOptionalInfo())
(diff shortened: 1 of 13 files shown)


lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -790,7 +791,13 @@ ParseResult FIRParser::parseEnumType(FIRRTLType &result) {
                  "expected leading '{|' in enumeration type"))
     return failure();
   SmallVector<FEnumType::EnumElement> elements;
+  bool first = true;
   if (parseListUntil(FIRToken::r_brace_bar, [&]() -> ParseResult {
+        if (!first) {
+          if (parseToken(FIRToken::comma, "expected ','")) {
+            return failure();
+          }
+        }
```
uenoku, line 800: I think every parseListUntil use is modified in the PR, can we change `parseListUntil` to something like `parseCommaSeparatedListUntil` and move `first` stuff to that function?

uenoku (commented):
The implementation generally looks good to me! I agree that comma as whitespace is weird and should be fixed but practically I guess the hack improved parser performance quite a bit :)  I'm curious how parser performance changes due to increased toke size.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: Good catch. I didn't realize this was locally-defined.

I plan on bringing the parser into alignment with the grammar in the spec, so I expect this to change. As long as this looks correct, I will probably leave it as it is for now.

dtzSiFive:
Technically this is only mandatory as of 4.0.0, https://github.com/chipsalliance/firrtl-spec/pull/191 .  However I think just requiring this regardless is fine.

cc https://github.com/chipsalliance/firrtl-spec/issues/188 .

mmaloney-sf:
@dtzSiFive Thanks for pointing that out. I feel like I removed that verbiage and then forgot it.

And thank you for pointing out that I didn't feature-gate this. I'm not sure how I should proceed on that point. I had plans to add a lot of small tweaks to the syntax, but it could easily make it a bit of spaghetti to keep strict backwards compat. But then again, I don't know much people care about that compatibility.

seldridge:
Note: no version of Chisel that I have ever seen has ever _not_ emitted commas unconditionally. There should be no risk in changing this and it should be backwards compatible.

test/firtool/phase-ordering.fir
```diff
@@ -31,7 +31,8 @@ circuit Issue794: %[[{
       data-type => UInt<8>
       depth => 4
       reader => r
-      writer => w0,w1
```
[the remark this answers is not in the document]
mmaloney-sf: I was quite confused when I saw this.

Some mysteries will never be revealed.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -4085,10 +4171,14 @@ ParseResult FIRStmtParser::parseSeqMem() {
...
       parseToken(FIRToken::colon, "expected ':' in smem") ||
-      parseType(type, "expected smem type") || parseOptionalRUW(ruw) ||
-      parseOptionalInfo())
+      parseType(type, "expected smem type"))
     return failure();
 
+  if (consumeIf(FIRToken::comma)) {
+    if (parseRUW(ruw) || parseOptionalInfo())
+      return failure();
+  }
```
uenoku, line 4180: I think parseOptionalInfo needs to be called regardless of RUW.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -336,15 +338,30 @@ ParseResult FIRParser::parseToken(FIRToken::Kind expectedToken,
   return emitError(message);
 }
 
-/// Parse a list of elements, terminated with an arbitrary token.
+/// Parse a comma-separated list of elements, terminated with an arbitrary
+/// token.
 ParseResult
 FIRParser::parseListUntil(FIRToken::Kind rightToken,
                           const std::function<ParseResult()> &parseElement) {
+  bool first = false;
```
dtzSiFive, line 346: It might be clearer to replace this boolean state with 1-2 lines that basically unroll that path?

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -336,15 +338,30 @@ ParseResult FIRParser::parseToken(FIRToken::Kind expectedToken,
...
 
   while (!consumeIf(rightToken)) {
+    // on the second iteration and thereafter, parse a comma.
+    if (first) {
+      if (parseToken(FIRToken::comma, "expected ','")) {
+        return failure();
+      }
+      if (consumeIf(rightToken)) {
```
dtzSiFive, line 354: Can we do this change separately? (trailing comma)

I also am not sold it makes sense everywhere we use this?

For example, a trailing comma in a prim op seems weird.  It's nice in things like bundles/enums/types, but seems a little weird in statements -- intrinsics, variadic things like printf/assert, and such.

At least that's my take right now (WDYT? :grin:) but main point is let's tackle that separately.

And wherever it's added, it should have tests please.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -954,18 +971,20 @@ ParseResult FIRParser::parseType(FIRRTLType &result, const Twine &message) {
...
-    if (getToken().getKind() == FIRToken::identifier) {
-      if (requireFeature({3, 2, 0}, "colored probes"))
-        return failure();
-      do {
-        StringRef layer;
-        loc = getToken().getLoc();
-        if (parseId(layer, "expected layer name"))
+    if (consumeIf(FIRToken::comma)) {
+      // Probe Color
+      if (getToken().getKind() == FIRToken::identifier) {
```
dtzSiFive, line 977: Trailing comma is kinda weird here, isn't it?

dtzSiFive (commented):
Thanks for pushing on this! Left some feedback, still going through but wanted to submit that much now.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: Fair point on all accounts.

test/Dialect/FIRRTL/parse-errors.fir
```diff
@@ -1371,55 +1371,55 @@ circuit Foo:
...
-  layer A bind:
+  layer A, bind:
   public module Foo:
     output a: Probe<
     ; expected-error @below {{expected probe data type}}
 ;// -----
 FIRRTL version 4.0.0
 circuit Foo:
-  layer A bind:
+  layer A, bind:
```
darthscsi, line 1381: Why this change?

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5190,8 +5285,11 @@ ParseResult FIRCircuitParser::parseLayer(CircuitOp circuit) {
     StringRef id, convention;
     LocWithInfo info(getToken().getLoc(), this);
     consumeToken();
-    if (parseId(id, "expected layer name") || parseGetSpelling(convention))
+    if (parseId(id, "expected layer name") ||
+        parseToken(FIRToken::comma, "expected ','") ||
```
darthscsi, line 5289: Is a comma here in the spec?  It looks a little confusing and I don't know that we should add it if it wasn't already.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5201,13 +5299,16 @@ ParseResult FIRCircuitParser::parseLayer(CircuitOp circuit) {
     consumeToken();
 
     hw::OutputFileAttr outputDir;
-    if (getToken().getKind() == FIRToken::string) {
-      auto text = getToken().getStringValue();
-      if (text.empty())
-        return emitError() << "output directory must not be blank";
-      outputDir = hw::OutputFileAttr::getAsDirectory(getContext(), text);
-      consumeToken(FIRToken::string);
+    if (consumeIf(FIRToken::comma)) {
```
darthscsi, line 5302: I don't think we want a trailing comma here.  `,:` shouldn't be valid I think.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
darthscsi: Yea, this both allows a dangling comma and doesn't properly parse optional parts.  I think most consumeIf calls I saw are probably wrong.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -336,15 +338,30 @@ ParseResult FIRParser::parseToken(FIRToken::Kind expectedToken,
...
 ParseResult
 FIRParser::parseListUntil(FIRToken::Kind rightToken,
                           const std::function<ParseResult()> &parseElement) {
+  bool first = false;
 
   while (!consumeIf(rightToken)) {
+    // on the second iteration and thereafter, parse a comma.
+    if (first) {
+      if (parseToken(FIRToken::comma, "expected ','")) {
+        return failure();
```
darthscsi, line 352: This is allowing dangling commas.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
darthscsi: yea.

darthscsi (commented):
Please add explicit tests that dangling commas in lists and before optional components are rejected.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -336,15 +338,30 @@ ParseResult FIRParser::parseToken(FIRToken::Kind expectedToken,
   return emitError(message);
 }
 
-/// Parse a list of elements, terminated with an arbitrary token.
+/// Parse a comma-separated list of elements, terminated with an arbitrary
+/// token.
```
darthscsi, line 342: Now that there are explicit commas, isn't there an existing delimiter separated list parse function we can just use?

test/Dialect/FIRRTL/parse-errors.fir
mmaloney-sf: Despite the grammar in the spec, there are mixed examples: some with an interleaving comma and some without. I made the call here to require the comma rather than to forbid it.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: If someone knows of such a function, I can switch it out.
Considering the whole parser is hand-written, I wasn't expecting there to be much in the way of parsing utilities.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: I'll probably just leave this as is unless there's a need for a change. This function does what it says on the tin, and the current spelling isn't so bad.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: I backed out of this change. Trailing commas are back to disallowed.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: I backed out of allowing trailing commas.

I found a bit of a quirk with the version gating on colors. But I think I got it.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: The spec doesn't have it. The tests had it both ways.

I decided to require the comma here. I'll update the spec accordingly after this change goes through.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: I backed out the trailing commas change.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: I backed out of the trailing commas change. Resolved.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
mmaloney-sf: Thanks for the catch, @uenoku.

@darthscsi I'm not sure what you mean. Feel free to add more information. If you have an example that should be added to the parse-errors file, that would be great.

mmaloney-sf:
> Please add explicit tests that dangling commas in lists and before optional components are rejected.

Done.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -248,7 +248,8 @@ struct FIRParser {
   /// output a diagnostic and return failure.
   ParseResult parseToken(FIRToken::Kind expectedToken, const Twine &message);
 
-  /// Parse a list of elements, terminated with an arbitrary token.
+  //// Parse a comma-separated list of elements, terminated with an arbitrary
```
uenoku, line 251: nit:
```suggestion
  /// Parse a comma-separated list of elements, terminated with an arbitrary
```

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -336,15 +338,26 @@ ParseResult FIRParser::parseToken(FIRToken::Kind expectedToken,
...
 
-/// Parse a list of elements, terminated with an arbitrary token.
+/// Parse a comma-separated list of zero or more elements, terminated with an
+/// arbitrary token.
 ParseResult
 FIRParser::parseListUntil(FIRToken::Kind rightToken,
                           const std::function<ParseResult()> &parseElement) {
+  if (consumeIf(rightToken)) {
+    return success();
+  }
```
uenoku, line 348: nit: please remove braces https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
  if (consumeIf(rightToken))
    return success();
```

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -336,15 +338,26 @@ ParseResult FIRParser::parseToken(FIRToken::Kind expectedToken,
...
+  if (parseElement())
+    return failure();
 
-  while (!consumeIf(rightToken)) {
+  while (consumeIf(FIRToken::comma)) {
     if (parseElement())
       return failure();
   }
+
+  if (parseToken(rightToken, "expected ','")) {
```
uenoku, line 358: ditto

mmaloney-sf:
If there are no other comments, could someone merge?

mmaloney-sf:
Sounds good to me, @uenoku

PR #7207, opened 2024-06-19
[ImportVerilog] Dedup moduleOp

mingzheTerapines:
Dedup module op generating while importing Verilog


mingzheTerapines:
@fabianschuiki 
This PR can only dedup same name module OP generating which is more lightweight.
https://github.com/llvm/circt/pull/7190/files That PR can dedup all equipllance module OP which is more powerful.

fabianschuiki:
I think you need more than just the name for deduplication: instances of the same module with different parameters will have the same name, but wildly different ports and bodies. I think you need a wrapper struct around the module and use that as a hash map key: that should hash the entire module if asked for a hash, and structurally check for equivalence if asked for equivalence.

mingzheTerapines:
@fabianschuiki @hailongSun2000 @cepheus69  Comparing name is dangerous, now I also compare parent's symbol.
```
if (instNode.body.name == module.first->name &&
          instNode.getParentScope() == module.getFirst()->getParentScope()) {
        duplicateInstanceBodySymbol = module.first;
        break;
      }
```

hailongSun2000:
Hey, @mingzheTerapines. There is an example with parameter:
```
```

mingzheTerapines:
@fabianschuiki I have supported TypeParameterSymbol dedup but construct is not ready yet. So I was wondering converting Typeparameter to hw.paramvalue, but found no mooreOPs support mlir::typeattribute declared in an Op. Maybe we need add NamedTypeConstant Op or something and typeType for Moore dialect.

PR #7209, opened 2024-06-19
[HW] Clean up HWTypes, NFC

uenoku:
* Port some of customAssemblyFormat to declarative assembly-format. HW uses unconventional printer so we still need to use custom directive though. 
* Use TypeBuilderWithInferredContext 
* Make banner consistent with others


dtzSiFive (approved):
LGTM!

Didn't super-carefully compare the before/after printing/parsing logic but the diff looks right and thanks for cleaning this up!

uenoku:
Thank you for review!

PR #7210, opened 2024-06-19
[FIRRTL] add default case for EventControl instantiation

sequencer:
This patch fix the compiler warning maybe-uninitialized

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -4580,6 +4580,9 @@ LogicalResult FIRRTLLowering::lowerVerificationStatement(
       case EventControl::AtNegEdge:
         event = circt::sv::EventControl::AtNegEdge;
         break;
+      default:
+        event = circt::sv::EventControl::AtPosEdge;
+        break;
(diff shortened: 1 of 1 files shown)


lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -4580,6 +4580,9 @@ LogicalResult FIRRTLLowering::lowerVerificationStatement(
       case EventControl::AtNegEdge:
         event = circt::sv::EventControl::AtNegEdge;
         break;
+      default:
+        event = circt::sv::EventControl::AtPosEdge;
+        break;
```
uenoku, line 4585:
```suggestion
      default:
        llvm::report_fatal_error("Invalid event");
```

Does this remove the warning? I think compile failure is better than implicitly rolling back to a random default value.

uenoku (commented):
Thanks! It's unfortunate that compiler is not smart enough :)

dtzSiFive:
https://llvm.org/docs/CodingStandards.html#don-t-use-default-labels-in-fully-covered-switches-over-enumerations

If it's covered please don't add a default.

PR #7216, opened 2024-06-20
[Moore] Add LowerConcatRef pass to handle concat_ref.

hailongSun2000:
Assignments like `{a, b} = c` are very tricky to handle directly. So we introduce a new Moore-to-Moore pass called `LowerConcatRef`. Which is used to disassemble the `moore.concat_ref`. And it runs before MooreToCore and guarantees never to see `moore.concat_ref` on the LHS of an assignment. For example:
```
{a, b} = c;
// becomes
a = c[9001:42];
b = c[41:0];
```

### DIFF
--- a/include/circt/Dialect/Moore/MoorePasses.td
+++ b/include/circt/Dialect/Moore/MoorePasses.td
@@ -29,4 +29,14 @@ def SimplifyProcedures : Pass<"moore-simplify-procedures", "moore::SVModuleOp"> 
     let constructor = "circt::moore::createSimplifyProceduresPass()";
 }
 
+def LowerConcatRef : Pass<"moore-lower-concatref", "moore::SVModuleOp"> {
+    let summary = "Lower moore.concat_ref ops";
+    let description = [{
+      It's used to disassemble the LHS of assignments that have a form like
+      "{a, b} = c" onto "a = c[9001:42];" and "b = c[41:0]". Aimed at
+      conveniently lowering this kind of assignment.
+    }];
+    let constructor = "circt::moore::createLowerConcatRefPass()";
+}
+
 #endif // CIRCT_DIALECT_MOORE_MOOREPASSES_TD
--- a/lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
+++ b/lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
@@ -1,0 +1,113 @@
+//===- LowerConcatRef.cpp - moore.concat_ref lowering ---------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file defines the LowerConcatRef pass.
+// It's used to disassemble the moore.concat_ref. Which is tricky to lower
+// directly. For example, disassemble "{a, b} = c" onto "a = c[7:3]"
+// and "b = c[2:0]".
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Moore/MooreOps.h"
+#include "circt/Dialect/Moore/MoorePasses.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+namespace circt {
+namespace moore {
+#define GEN_PASS_DEF_LOWERCONCATREF
+#include "circt/Dialect/Moore/MoorePasses.h.inc"
+} // namespace moore
+} // namespace circt
+
+using namespace circt;
+using namespace moore;
+using namespace mlir;
+
+namespace {
+
+// A helper function for collecting the non-concatRef operands of concatRef.
+static void collectOperands(Value operand, SmallVectorImpl<Value> &operands) {
+  if (auto concatRefOp = operand.getDefiningOp<ConcatRefOp>())
+    for (auto nestedOperand : concatRefOp.getValues())
+      collectOperands(nestedOperand, operands);
+  else
+    operands.push_back(operand);
+}
+
+template <typename OpTy>
+struct ConcatRefLowering : public OpConversionPattern<OpTy> {
+  using OpConversionPattern<OpTy>::OpConversionPattern;
+  using OpAdaptor = typename OpTy::Adaptor;
+
+  LogicalResult
+  matchAndRewrite(OpTy op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Use to collect the operands of concatRef.
+    SmallVector<Value, 4> operands;
+    collectOperands(op.getDst(), operands);
+    auto srcWidth =
+        cast<UnpackedType>(op.getSrc().getType()).getBitSize().value();
+
+    // Disassemble assignments with the LHS is concatRef. And create new
+    // corresponding assignments using non-concatRef LHS.
+    for (auto operand : operands) {
+      auto type = cast<RefType>(operand.getType()).getNestedType();
+      auto width = type.getBitSize().value();
+
+      rewriter.setInsertionPoint(op);
+      // FIXME: Need to estimate whether the bits range is from large to
+      // small or vice versa. Like "logic [7:0] or [0:7]".
+
+      // Only able to correctly handle the situation like "[7:0]" now.
+      auto i32 = moore::IntType::getInt(op.getContext(), 32);
+      auto lowBit =
+          rewriter.create<ConstantOp>(op.getLoc(), i32, srcWidth - width);
+      auto extract =
+          rewriter.create<ExtractOp>(op.getLoc(), type, op.getSrc(), lowBit);
+
+      // Update the real bit width of RHS of assignment. Like "c" the above
+      // description mentioned.
+      srcWidth = srcWidth - width;
+
+      rewriter.create<OpTy>(op.getLoc(), operand, extract);
+    }
+    rewriter.eraseOp(op);
+    return success();
+  }
+};
+
+struct LowerConcatRefPass
+    : public circt::moore::impl::LowerConcatRefBase<LowerConcatRefPass> {
+  void runOnOperation() override;
+};
+
+} // namespace
+
+std::unique_ptr<mlir::Pass> circt::moore::createLowerConcatRefPass() {
+  return std::make_unique<LowerConcatRefPass>();
+}
+
+void LowerConcatRefPass::runOnOperation() {
+  MLIRContext &context = getContext();
+  ConversionTarget target(context);
+
+  target.addDynamicallyLegalOp<ContinuousAssignOp, BlockingAssignOp,
+                               NonBlockingAssignOp>([](auto op) {
+    return !op->getOperand(0).template getDefiningOp<ConcatRefOp>();
+  });
+
+  target.addLegalDialect<MooreDialect>();
+  RewritePatternSet patterns(&context);
+  patterns.add<ConcatRefLowering<ContinuousAssignOp>,
+               ConcatRefLowering<BlockingAssignOp>,
+               ConcatRefLowering<NonBlockingAssignOp>>(&context);
+
+  if (failed(
+          applyPartialConversion(getOperation(), target, std::move(patterns))))
+    signalPassFailure();
+}
--- a/lib/Dialect/Moore/Transforms/SimplifyProcedures.cpp
+++ b/lib/Dialect/Moore/Transforms/SimplifyProcedures.cpp
@@ -53,7 +53,8 @@ void SimplifyProceduresPass::runOnOperation() {
         // Collect the users of the global variable that is mentioned above.
         DenseSet<Operation *> users;
         for (auto *user : nestedOp.getOperand(0).getUsers())
-          if (!users.contains(user))
+          // Ensuring don't handle the users existing in another procedure body.
+          if (user->getBlock() == procedureOp.getBody())
             users.insert(user);
 
         // Because the operand of moore.event_wait is net.
@@ -98,6 +99,5 @@ void SimplifyProceduresPass::runOnOperation() {
           assignOps.erase(assignOp);
         }
     }
-    return WalkResult::advance();
   });
 }
--- a/tools/circt-verilog/CMakeLists.txt
+++ b/tools/circt-verilog/CMakeLists.txt
@@ -7,6 +7,7 @@ llvm_update_compile_flags(circt-verilog)
 target_link_libraries(circt-verilog PRIVATE
   CIRCTImportVerilog
   CIRCTMooreToCore
+  CIRCTMooreTransforms
   CIRCTSupport
   MLIRIR
 )
--- a/tools/circt-verilog/circt-verilog.cpp
+++ b/tools/circt-verilog/circt-verilog.cpp
@@ -14,6 +14,7 @@
 
 #include "circt/Conversion/ImportVerilog.h"
 #include "circt/Conversion/MooreToCore.h"
+#include "circt/Dialect/Moore/MoorePasses.h"
 #include "circt/Support/Version.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Pass/PassManager.h"
@@ -219,6 +220,9 @@ static CLOptions opts;
 /// that simplify these files like deleting local variables, and then emit the
 /// resulting Moore dialect IR .
 static LogicalResult populateMooreTransforms(mlir::PassManager &pm) {
+  auto &modulePM = pm.nest<moore::SVModuleOp>();
+  modulePM.addPass(moore::createLowerConcatRefPass());
+  modulePM.addPass(moore::createSimplifyProceduresPass());
   pm.addPass(mlir::createMem2Reg());
   // TODO: like dedup pass.
 
--- a/include/circt/Dialect/Moore/MoorePasses.h
+++ b/include/circt/Dialect/Moore/MoorePasses.h
@@ -23,6 +23,7 @@ namespace moore {
 #include "circt/Dialect/Moore/MoorePasses.h.inc"
 
 std::unique_ptr<mlir::Pass> createSimplifyProceduresPass();
+std::unique_ptr<mlir::Pass> createLowerConcatRefPass();
 
 /// Generate the code for registering passes.
 #define GEN_PASS_REGISTRATION
--- a/lib/Dialect/Moore/Transforms/CMakeLists.txt
+++ b/lib/Dialect/Moore/Transforms/CMakeLists.txt
@@ -1,4 +1,5 @@
 add_circt_dialect_library(CIRCTMooreTransforms
+LowerConcatRef.cpp
 SimplifyProcedures.cpp
 
 
--- a/test/Dialect/Moore/lower-concatref.mlir
+++ b/test/Dialect/Moore/lower-concatref.mlir
@@ -1,0 +1,87 @@
+// RUN: circt-opt --moore-lower-concatref %s | FileCheck %s
+
+// CHECK-LABEL: moore.module @Foo()
+ moore.module @Foo() {
+    %a = moore.variable : <i8960>
+    %b = moore.variable : <i42>
+    %c = moore.variable : <i9002>
+    %u = moore.variable : <l8960>
+    %v = moore.variable : <l42>
+    %w = moore.variable : <l9002>
+
+    // CHECK: moore.concat_ref %a, %b
+    %0 = moore.concat_ref %a, %b : (!moore.ref<i8960>, !moore.ref<i42>) -> <i9002>
+    // CHECK: %[[C_READ:.+]] = moore.read %c : i9002
+    %1 = moore.read %c : i9002
+    // CHECK: %[[CONST_42:.+]] = moore.constant 42 : i32
+    // CHECK: %[[TMP1:.+]] = moore.extract %[[C_READ]] from %[[CONST_42]] : i9002, i32 -> i8960
+    // CHECK: moore.assign %a, %[[TMP1]] : i8960
+    // CHECK: %[[CONST_0:.+]] = moore.constant 0 : i32
+    // CHECK: %[[TMP2:.+]] = moore.extract %[[C_READ]] from %[[CONST_0]] : i9002, i32 -> i42
+    // CHECK: moore.assign %b, %[[TMP2]] : i42
+    moore.assign %0, %1 : i9002
+    moore.procedure always {
+      // CHECK: moore.concat_ref %u, %v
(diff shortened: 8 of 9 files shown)


hailongSun2000:
Thanks for the code review in advance! And Thanks @fabianschuiki's suggestion! :heart:

lib/Dialect/Moore/Transforms/SimplifyProcedures.cpp
```diff
@@ -53,7 +53,8 @@ void SimplifyProceduresPass::runOnOperation() {
         // Collect the users of the global variable that is mentioned above.
         DenseSet<Operation *> users;
         for (auto *user : nestedOp.getOperand(0).getUsers())
-          if (!users.contains(user))
+          // Ensuring don't handle the users existing in another procedure body.
+          if (user->getBlock() == procedureOp.getBody())
             users.insert(user);
```
hailongSun2000, line 58: I learned that the underlying implementation of the `insert()` of `llvm::DenseSet` is through calling the `try_emplace` of `llvm::DensMap`. So I think we needn't estimate whether the user exists in the `DenseSet`.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
```diff
@@ -0,0 +1,83 @@
...
+      auto srcWidth = cast<UnpackedType>(src.getType()).getBitSize().value();
+
+      // Handle all operands of moore.concat_ref.
+      for (auto operand : concatRefOp.getValues()) {
+        auto type = cast<RefType>(operand.getType()).getNestedType();
+        auto width = type.getBitSize().value();
+        OpBuilder::InsertionGuard guard(builder);
+        builder.setInsertionPoint(user);
+        auto i32 = moore::IntType::getInt(&getContext(), 32);
+        auto lowBit = builder.create<ConstantOp>(concatRefOp.getLoc(), i32,
```
uenoku, line 56: Could you leave a comment regarding endian? I think the implementation is correct but the endian in CIRCT is always confusing :)

uenoku (commented):
Generally looks great! Please consider using mlir OpConversionPattren.

tools/circt-verilog/CMakeLists.txt
```diff
@@ -8,5 +8,6 @@ target_link_libraries(circt-verilog PRIVATE
   CIRCTImportVerilog
   CIRCTMooreToCore
   CIRCTSupport
+  CIRCTMooreTransforms
```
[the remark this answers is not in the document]
hailongSun2000: I get it.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
```diff
@@ -0,0 +1,83 @@
...
+        // Estimate which assignment needs to be created.
+        if (isa<ContinuousAssignOp>(user))
+          builder.create<ContinuousAssignOp>(
+              user->getLoc(), operand.getDefiningOp()->getResult(0), extract);
+        else if (isa<BlockingAssignOp>(user))
+          builder.create<BlockingAssignOp>(
+              user->getLoc(), operand.getDefiningOp()->getResult(0), extract);
+        else
+          builder.create<NonBlockingAssignOp>(
+              user->getLoc(), operand.getDefiningOp()->getResult(0), extract);
```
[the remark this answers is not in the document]
hailongSun2000: I'll tweak it.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
```diff
@@ -0,0 +1,83 @@
...
+            builder.create<ExtractOp>(src.getLoc(), type, src, lowBit);
+
+        // Update the real bit width of RHS of assignment. Like "c" the above
+        // description mentioned.
+        srcWidth = srcWidth - width;
+
+        // Estimate which assignment needs to be created.
+        if (isa<ContinuousAssignOp>(user))
+          builder.create<ContinuousAssignOp>(
+              user->getLoc(), operand.getDefiningOp()->getResult(0), extract);
```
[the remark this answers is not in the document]
hailongSun2000: You're totally right! I was stupid. :sweat_smile:

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
```diff
@@ -0,0 +1,83 @@
+//===- DeleteLocalVar.cpp - Delete local temporary variables --------------===//
```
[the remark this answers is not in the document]
hailongSun2000: I'll be careful later.

lib/Dialect/Moore/Transforms/SimplifyProcedures.cpp
```diff
@@ -53,7 +53,8 @@ void SimplifyProceduresPass::runOnOperation() {
         // Collect the users of the global variable that is mentioned above.
         DenseSet<Operation *> users;
         for (auto *user : nestedOp.getOperand(0).getUsers())
-          if (!users.contains(user))
+          // Ensuring don't handle the users existing in another procedure body.
+          if (user->getBlock() == procedureOp.getBody())
```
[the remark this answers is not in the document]
hailongSun2000: I'll add a test for this later.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
```diff
@@ -0,0 +1,83 @@
...
+        else
+          builder.create<NonBlockingAssignOp>(
+              user->getLoc(), operand.getDefiningOp()->getResult(0), extract);
+      }
+
+      // Delete the original moore.concat_ref and its user--moore.*assign.
+      user->erase();
+      concatRefOp->erase();
+    }
+    return WalkResult::advance();
```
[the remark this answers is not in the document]
hailongSun2000: It looks like I misunderstood this. I think I also need to remove the same thing in another pass.

test/Dialect/Moore/simplify-assigns.mlir
```diff
@@ -0,0 +1,52 @@
+// RUN: circt-opt --moore-simplify-assigns %s | FileCheck %s
+
+ moore.module @Foo() {
```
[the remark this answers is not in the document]
hailongSun2000: I'll keep it in mind.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
hailongSun2000: No problem! I'll add a comment for endian.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
```diff
@@ -0,0 +1,83 @@
...
+          builder.create<BlockingAssignOp>(
+              user->getLoc(), operand.getDefiningOp()->getResult(0), extract);
+        else
+          builder.create<NonBlockingAssignOp>(
+              user->getLoc(), operand.getDefiningOp()->getResult(0), extract);
+      }
+
+      // Delete the original moore.concat_ref and its user--moore.*assign.
+      user->erase();
+      concatRefOp->erase();
```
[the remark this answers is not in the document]
hailongSun2000: Thanks so much! You're right to be worried. I'll tweak it using mlir::OpConversionPattern.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
```diff
@@ -0,0 +1,83 @@
...
+std::unique_ptr<mlir::Pass> circt::moore::createSimplifyAssignsPass() {
+  return std::make_unique<SimplifyAssignsPass>();
+}
+
+void SimplifyAssignsPass::runOnOperation() {
+  getOperation()->walk([&](ConcatRefOp concatRefOp) {
+    mlir::OpBuilder builder(&getContext());
+
+    // moore.concat_ref only as the LHS of assignments.
+    if (auto *user = *concatRefOp->getUsers().begin()) {
```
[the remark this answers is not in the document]
hailongSun2000: I was thoughtless. You're absolutely right! The `moore.instance` also uses `moore.concat_ref`, such as inout ports.

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -219,6 +220,8 @@ static CLOptions opts;
 /// that simplify these files like deleting local variables, and then emit the
 /// resulting Moore dialect IR .
 static LogicalResult populateMooreTransforms(mlir::PassManager &pm) {
+  pm.addNestedPass<moore::SVModuleOp>(moore::createSimplifyAssignsPass());
+  pm.addNestedPass<moore::SVModuleOp>(moore::createSimplifyProceduresPass());
```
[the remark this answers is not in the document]
hailongSun2000: Cool!

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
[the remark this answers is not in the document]
hailongSun2000: I can promise every `moore.concat_ref` only has a single user, because we will always create a new `moore.concat_ref` when we need it. This can be checked in the `Expressions.cpp`-- `Value Context::convertLvalueExpression(const slang::ast::Expression &expr){...}`

hailongSun2000:
Hey, @uenoku! I'm having trouble with nested `moore.concat_ref`. For example,
```
moore.module @top() {
  %a = moore.variable : <i32>
  %b = moore.variable : <i32>
  %c = moore.variable : <i32>
  moore.procedure initial {
    %0 = moore.concat_ref %a, %a : (!moore.ref<i32>, !moore.ref<i32>) -> <i64>
    %1 = moore.concat_ref %a, %b : (!moore.ref<i32>, !moore.ref<i32>) -> <i64>
    %2 = moore.concat_ref %0, %1 : (!moore.ref<i64>, !moore.ref<i64>) -> <i128>
    %3 = moore.read %c : i32
    %4 = moore.conversion %3 : !moore.i32 -> !moore.i128
    %5 = moore.conversion %4 : !moore.i128 -> !moore.i128
    moore.blocking_assign %2, %5 : i128
  }
  moore.output
}
```
I attempted to use OpConversionPattern to rewrite `moore.concat_ref`, I want to rewrite this case to become 
```
// All `moor.concat_ref` are erased.
moore.blokcing_assign %a, %10      // %10 = extract 32 bits[127:96] from %5
moore.blokcing_assign %a, %11      // %11 = extract 32 bits[95:64] from %5 
moore.blokcing_assign %a, %12      // %12 = extract 32 bits[63:32] from %5
moore.blokcing_assign %b, %13      // %13 = extract 32 bits[31:0] from %5
```
However, I can only rewrite `%2 = moore.concat_ref %0, %1... ` and `moore.blocking_assign %2, %5 : i128` to become 
```
%0 = moore.concat_ref %a, %a : (!moore.ref<i32>, !moore.ref<i32>) -> <i64>
%1 = moore.concat_ref %a, %b : (!moore.ref<i32>, !moore.ref<i32>) -> <i64>
moore.blokcing_assign %0, %20      // %0 = extract 64 bits[127:64] from %5
moore.blokcing_assign %1, %21      // %1 = extract 64 bits[63:0] from %5
```
It's more complicated than I thought. So, does the `OpConversionPattern` have a way(like a loop) to handle this situation?

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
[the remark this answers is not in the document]
uenoku: No, `concat_ref` is marked as `Pure` so CSE would freely eliminate duplicated values (so multiple users). If we want to keep a single user as an invariant we need to eliminate `Pure` and add a verifier to verify that but I don't think it's reasonable.

uenoku:
I think the pattern should be recursively applied to `%0 = moore.concat_ref %a, %a` and `%1 = moore.concat_ref %a, %b` if we let OpConversion Lister know the operation is replaced. Could you push a change so I take a look?

hailongSun2000:
I pushed the changed version.

hailongSun2000:
> I think the pattern should be recursively applied to `%0 = moore.concat_ref %a, %a` and `%1 = moore.concat_ref %a, %b` if we let OpConversion Lister know the operation is replaced. Could you push a change so I take a look?

Maybe I need to think about another way to implement the OpConversionPattern. I use the old implementation.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
```diff
@@ -0,0 +1,128 @@
...
+  return std::make_unique<LowerConcatRefPass>();
+}
+
+void LowerConcatRefPass::runOnOperation() {
+  MLIRContext &context = getContext();
+  ConversionTarget target(context);
+
+  target.addDynamicallyLegalOp<ConcatRefOp>([](ConcatRefOp op) {
+    auto *user = *op->getUsers().begin();
+    return !isa<BlockingAssignOp>(user);
```
uenoku, line 114: I think all concat ref operatins are illegal.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
```diff
@@ -0,0 +1,128 @@
...
+// };
+
+// template<typename OpTy>
+struct ConcatRefLowering : public OpConversionPattern<ConcatRefOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ConcatRefOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    if (auto *user = *op->getUsers().begin()) {
```
uenoku, line 54: I think you have to all users which include non-assignments (e.g. concat ref).

You can flatten concat_ref by flattening. 

`concat_ref(concat_ref(a, b), c) -> concat_ref(a, b, c)`.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
hailongSun2000: Yeah, you're right! I also think so. However, a present, I don't how to handle the case like the instance uses the `concat_ref`:
```
module {
  moore.module @top() {
    %w1 = moore.net wire : <l1>
    %w2 = moore.net wire : <l1>
    %w3 = moore.net wire : <l1>
    %w4 = moore.net wire : <l1>
    %0 = moore.concat_ref %w1, %w2, %w3, %w4 : (!moore.ref<l1>, !moore.ref<l1>, !moore.ref<l1>, !moore.ref<l1>) -> <l4>
    moore.instance "a" @A(a: %0: !moore.ref<l4>) -> ()
    moore.output
  }
  moore.module @A(in %a : !moore.ref<l4>) {
    %a_0 = moore.net name "a" wire : <l4>
    %0 = moore.read %a : l4
    moore.assign %a_0, %0 : l4
    moore.output
  }
}
```

uenoku: Oh, I see. That's tricky. In that case you can allow concat_ref whose users are instances.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
hailongSun2000: Exactly! I have thought about this you mentioned. But for the IRs, merge them
```
%0 = moore.concat_ref %a, %a : (!moore.ref<i32>, !moore.ref<i32>) -> <i64>
%1 = moore.concat_ref %a, %b : (!moore.ref<i32>, !moore.ref<i32>) -> <i64>
%2 = moore.concat_ref %0, %1 : (!moore.ref<i64>, !moore.ref<i64>) -> <i128>
```
into one op like `%0 = moore.concat_ref %a, %a, %a, %b : ...` looks like difficult.
When I visit `%0`, I must handle it immediately, right? If I don't handle it immediately. I must wait until visit `%2`, then recursively traverse its operands(`%0` and `%1`), and create a new `concat_ref`.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
hailongSun2000: Cool idea! I'll tweak it. Thanks! :smiley:

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
uenoku: > into one op like %0 = moore.concat_ref %a, %a, %a, %b : ... looks like difficult.

You already handled variadic operands so I think it works without any problem?

hailongSun2000: Maybe I can create a container like SmallVector, and then traverse the operands of `concat_ref`, and then collect them.

lib/Dialect/Moore/Transforms/SimplifyAssigns.cpp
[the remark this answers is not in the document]
hailongSun2000: Learned new knowledge.

include/circt/Dialect/Moore/MoorePasses.td
```diff
@@ -29,4 +29,14 @@ def SimplifyProcedures : Pass<"moore-simplify-procedures", "moore::SVModuleOp">
     let constructor = "circt::moore::createSimplifyProceduresPass()";
 }
 
+def LowerConcatRef : Pass<"lower-moore-concatref", "moore::SVModuleOp"> {
```
uenoku, line 32: nit: please start with `moore`.

hailongSun2000: I'll tweak it.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
```diff
@@ -0,0 +1,126 @@
...
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ConcatRefOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    // Use to collect the operands of concatRef.
+    SmallVector<Value> values;
+    for (auto operand : op.getValues()) {
+      // Handle the nested concatRef, and then collect its operands.
+      if (auto *defOp = operand.getDefiningOp(); isa<ConcatRefOp>(defOp)) {
```
uenoku, line 44:
```suggestion
      if (auto *defOp = operand.getDefiningOp<ConcatRefOp>()) {
```

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
```diff
@@ -0,0 +1,126 @@
...
+        for (auto nestedOperand : defOp->getOperands())
+          values.push_back(nestedOperand);
+        rewriter.eraseOp(defOp);
```
uenoku, line 47: I think it's illegal to delete defOp at this point

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
```diff
@@ -0,0 +1,126 @@
...
+          // Update the real bit width of RHS of assignment. Like "c" the above
+          // description mentioned.
+          srcWidth = srcWidth - width;
+          rewriter.create<OpTy>(user->getLoc(), operand, extract);
+        }
+
+        // Erase the old assignment if its LHS is concatRef.
+        rewriter.eraseOp(user);
+        // Because we needn't keep concatRef except it exists in instance.
+        rewriter.eraseOp(newOp);
```
uenoku, line 90: I think it must be after for loop?

fabianschuiki (commented):
One thing that might make the implementation easier is to make the `OpConversionPattern` not operate on `ConcatRefOp`, but on the assignment ops `ContinuousAssignOp`, `BlockingAssignOp`, and `NonBlockingAssignOp`.

For example, you could mark assignment ops with a `ConcatRefOp` LHS as illegal, and then have a pattern that splits the assignment up. You could even make the conversion pattern only consider _one single_ level of `ConcatRefOp`, and the conversion framework will then apply the pattern multiple times in case you have a nested concat.

That would also allow you to not care about other users of the `ConcatRefOp`: in case all users were assignments, they all get converted and the `ConcatRefOp` will have no more users and be deleted. Otherwise it remains with a few leftover users, like instances 😃

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
hailongSun2000: Sounds great! I'll keep in mind!

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
hailongSun2000: Yeah, you're right! I also think so! However, I don't know why print the same op three times. For example:
```
module {
  moore.module @top() {
    %a = moore.variable : <l5>
    %b = moore.variable : <l3>
    %c = moore.variable : <l8>
    %0 = moore.concat_ref %a, %b : (!moore.ref<l5>, !moore.ref<l3>) -> <l8>
    %1 = moore.read %c : l8
    moore.assign %0, %1 : l8
    moore.output
  }
}
```
When I add `op->dump()` at the start of the body of `matchAndRewrite()`, will dump
```
%0 = moore.concat_ref %a, %b : (!moore.ref<l5>, !moore.ref<l3>) -> <l8>
%0 = moore.concat_ref %a, %b : (!moore.ref<l5>, !moore.ref<l3>) -> <l8>
%0 = moore.concat_ref %a, %b : (!moore.ref<l5>, !moore.ref<l3>) -> <l8>
```
So when I move `rewriter.eraseOp(newOp);` after for loop, will throw an error:
```
failed to legalize operation 'moore.concat_ref' marked as erased
  assign {a,b} = c;
         ^
/home/phoenix/work/HDLBits/concatRef01.sv:5:10: note: see current operation: %3 = "moore.concat_ref"(%0, %1) : (!moore.ref<l5>, !moore.ref<l3>) -> !moore.ref<l8>
/home/phoenix/work/HDLBits/concatRef01.sv:5:10: note: found live user of result #0: moore.assign %0, %2 : l8
```
I'm not sure why this(dump three times) will occur.

hailongSun2000:
Good idea! My mind is solidified. Concentrate on assignments if its LHS is concatRef.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
hailongSun2000: I'm not sure. But I think you're right!

hailongSun2000:
Hey, @uenoku! I changed the implementation of `ConcatRefLowering` by using @fabianschuiki's suggestion. And don't erase the unused `defOp` that you point out.

uenoku (approved):
LGTM, thank you for iterating a lot on this!

hailongSun2000:
Don't be sorry. It's an honor for me to do something for CIRCT. Also very appreciative that you can give some nice suggestions. :smile:

PR #7217, opened 2024-06-20
[ESI Runtime] Replace Cap'nProto with gRPC

teqdruid:
gRPC is a lot easier to use owing to Cap'nProto's C++ implementation using likj.

Breaks MMIO, but we're not currently using it.


utils/get-grpc.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
mortbopet, line 1: I reeeeally have a hate for these fetch/install scripts, and would have a strong preference towards doing it in any other way, if possible.

Did you look into a fetchcontent-based approach? Looks like it's supported: https://github.com/grpc/grpc/tree/b217753062370b8f12e72fc3414eabe0a6f9f16e/src/cpp#fetchcontent

teqdruid: Yeah. Tried that and it failed spectacularly. gRPC pulls in like 16 submodules for 16 different dependencies. Some of them use the basel build system, so that gets built as well... One of the dependencies failed to build and I just knew that after I fix that, another would fail and didn't want to go down that rabbit hole.

I'm hoping that most distros will have a recent enough version of gRPC that we can just use the system-installed one.

teqdruid:
image

yep. that's what I expected.

utils/get-grpc.sh
mortbopet: Yikes, that sounds horrible... then this does seem like the lesser evil. Thank you for also investigating that!

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
```diff
@@ -0,0 +1,355 @@
...
+  server = nullptr;
+}
+
+Impl::~Impl() {
+  if (server)
+    stop();
+  for (auto &port : readPorts)
+    delete port.second;
+  for (auto &port : writePorts)
+    delete port.second;
```
mortbopet, line 168: What's the argument for not using smart pointers for these?

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
```diff
@@ -0,0 +1,355 @@
...
+  ServerUnaryReactor *reactor = context->DefaultReactor();
+  reactor->Finish(Status::OK);
+  return reactor;
+}
+
+/// Load the list of channels into the response and fire it off.
+ServerUnaryReactor *Impl::ListChannels(CallbackServerContext *context,
+                                       const VoidMessage *,
+                                       ListOfChannels *channelsOut) {
+  for (auto [name, port] : readPorts) {
```
mortbopet, line 198: ->`for(auto &[...]`)?

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
```diff
@@ -0,0 +1,355 @@
...
+    auto *channel = channelsOut->add_channels();
+    channel->set_name(name);
+    channel->set_type(port->getType()->getID());
+    channel->set_dir(ChannelDesc::Direction::ChannelDesc_Direction_TO_SERVER);
+  }
+  for (auto [name, port] : writePorts) {
```
mortbopet, line 204: as above

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
```diff
@@ -0,0 +1,355 @@
...
+      : writePort(writePort), sentSuccessfully(SendStatus::UnknownStatus),
+        shutdown(false) {
+    myThread = std::thread(&RpcServerWriteReactor::threadLoop, this);
+  }
+  ~RpcServerWriteReactor() {
+    shutdown = true;
+    myThread.join();
+  }
+
+  void OnDone() override { delete this; }
```
mortbopet, line 234: What's the argument for doing it like this, as opposed to more traditional (re: safe) memory management techniques?
I can tell that this is connected to the logic in Impl::ConnectToClientChannel. In the spirit of measureability and general awareness of what's going on in the system, it seems counterintuitive to spawn off these unmanaged objects out into the eather, with no ties (re: management) from the outside world. I'd rather see a design where such write reactors signal back to their parent management objects that they have finished doing whatever they need to do. Their managers can then react accordingly (log, trace, reorder, do nothing, ...) and safely destruct the reactor.

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
```diff
@@ -0,0 +1,355 @@
...
+      msg.set_data(reinterpret_cast<const char *>(data.getBytes()),
+                   data.getSize());
+
+      // Get a lock, reset the flag, start sending the message, and wait for the
+      // write to complete or fail. Be mindful of the shutdown flag.
+      std::unique_lock<std::mutex> lock(sentMutex);
+      sentSuccessfully = SendStatus::UnknownStatus;
+      StartWrite(&msg);
+      while (!shutdown && sentSuccessfully == SendStatus::UnknownStatus)
+        sentSuccessfullyCV.wait_for(lock, std::chrono::milliseconds(10));
```
mortbopet, line 288: Why poll? Seems like you have full control of when `sentSuccessfully` and `shutdown` are assigned; and as such, it should be possible to do the following, assuming the `shutdown` assignment on line 130 is placed behind the mutex and a cv wakeup:
```c++
sentSuccessfullyCV.wait(lock, [&](){return shutdown || sentSuccessfully != SendStatus::UnknownStatus});
```

lib/Dialect/ESI/runtime/cpp/include/esi/Utils.h
```diff
@@ -17,12 +17,80 @@
...
 void encodeBase64(const void *data, size_t size, std::string &out);
+
+/// Thread safe queue. Just wraps std::queue protected with a lock. Long term,
+/// we need to avoid copying data. It has a lot of data copies currently.
+template <typename T>
+class TSQueue {
+  using Lock = std::lock_guard<std::mutex>;
+
+  /// The queue and its mutex.
+  mutable std::mutex qM;
```
mortbopet, line 38: Why mutable?

lib/Dialect/ESI/runtime/cpp/include/esi/Utils.h
```diff
@@ -17,12 +17,80 @@
...
+  std::queue<T> q;
+
+  /// A mutex to ensure that only one 'pop' operation is happening at a time. It
+  /// is critical that locks be obtained on this and `qM` same order in both pop
+  /// methods. This lock should be obtained first since one of the pop methods
+  /// must unlock `qM` then relock it.
+  mutable std::mutex popM;
```
mortbopet, line 45: as above

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
```diff
@@ -25,9 +32,30 @@
...
 using namespace esi::backends::cosim;
 
+using grpc::Channel;
+using grpc::ClientContext;
+using grpc::ClientReader;
+using grpc::ClientReaderWriter;
+using grpc::ClientWriter;
+using grpc::Status;
+
+static void checkStatus(Status s, string msg) {
```
mortbopet, line 46: pass by const reference

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
```diff
@@ -25,9 +32,30 @@
 using namespace std;
```
mortbopet, line 32: ughhhhhhh

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
```diff
@@ -80,147 +108,212 @@ CosimAccelerator::CosimAccelerator(Context &ctxt, string hostname,
...
       throw runtime_error("Channel '" + name + "' has wrong type. Expected " +
-                          getType()->getID() + ", got " + ep->getRecvTypeId());
-    ep->setInUse();
+                          getType()->getID() + ", got " + desc.type());
+    if (desc.dir() != ChannelDesc::Direction::ChannelDesc_Direction_TO_CLIENT)
+      throw runtime_error("Channel '" + name + "' is not a to server channel");
+    assert(desc.name() == name);
+
+    // Initiate a stream of messages from the server.
+    context = new ClientContext();
```
mortbopet, line 260: as above - is there a particular good argument for having raw pointers here? Else, smart pointers should be the default.

mortbopet (commented):
Great to see that the diff in the non-cosim-backend code was relatively minor.

Based on the implementation and testing you've done on this, how is your initial gut feeling for gRPC vs capnproto? I'm thinking both in terms of compile-time and runtime-debugging that you went through.

utils/get-grpc.sh
teqdruid: I'll add that if someone wants to get this working in the future (as an alternative to a separate install), I have no major objections. (Just philosophical ones related to the very idea of FetchContent.)

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
teqdruid: Yeah, this is definitely an odd paradigm. It's copied from one of the gRPC examples: https://github.com/grpc/grpc/blob/4795c5e69b25e8c767b498bea784da0ef8c96fd5/examples/cpp/route_guide/route_guide_callback_server.cc#L120 so I assume it's the only way to know that gRPC has release the reference to this object. (I'll add a note to that effect.)

This model is equivalent letting gRPC do the memory management of this object, so it's not unmanaged or thrown into the ether. My gut actually agrees with yours, but I can't think of a tangible reason to add that complexity. In terms of measurability, what thread deletes the code isn't terribly relevant as the object could contain a reference to `Impl` or whatever is responsible for measurement, logging, or something else. If the management class wants to delete this object before gRPC is done with it, it can't -- that would be unsafe since gRPC isn't done with it.

I think this is the right way to go for now, and can be changed later on if there's a specific reason. There's no reason to prematurely add complexity.

lib/Dialect/ESI/runtime/CMakeLists.txt
```diff
@@ -175,6 +163,14 @@ if (XRT_PATH)
...
     ${XRT_PATH}/lib
   )
+  get_filename_component(XRT_SO ${XRT_PATH}/lib/libxrt_coreutil.so REALPATH)
+  install(FILES
+      ${XRT_SO}
+      ${XRT_PATH}/lib/libxrt_coreutil.so.2
+      ${XRT_PATH}/lib/libxrt_coreutil.so
+    DESTINATION lib
+    COMPONENT ESIRuntime
+  )
```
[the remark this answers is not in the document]
teqdruid: Yeah, I'm choosing to package the shared object rather than rely on a locally installed one so if someone ends up with a binary runtime distribution, they won't run into dynamic linking issues if they don't install the xrt runtime. Ideally, we'd have a plugin scheme instead. But you're right, it shouldn't be in this PR. I'll remove it.

lib/Dialect/ESI/runtime/cosim/cosim_dpi_server/DpiEntryPoints.cpp
```diff
@@ -61,7 +62,8 @@ static void log(char *epId, bool toClient,
 static int findPort() {
   const char *portEnv = getenv("COSIM_PORT");
   if (portEnv == nullptr) {
-    printf("[COSIM] RPC server port not found. Letting CapnpRPC select one\n");
+    printf(
```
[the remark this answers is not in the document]
teqdruid: I use it here only because the rest of the file uses printf. Originally, it was legacy coming from code which Lenharth wrote while still at Microsoft. I stuck with it because avoiding including <iostream> is a good thing. Since then, I've given in.

lib/Dialect/ESI/runtime/cosim/cosim_dpi_server/DpiEntryPoints.cpp
```diff
@@ -299,47 +332,50 @@ DPI int sv2cCosimserverMMIORegister() {
...
-    return -1;
-  *address = reqAddress.value();
-  ll->readsOutstanding++;
-  return 0;
+  // assert(server);
+  // LowLevel *ll = server->getLowLevel();
+  // std::optional<int> reqAddress = ll->readReqs.pop();
+  // if (!reqAddress.has_value())
+  return -1;
+  // *address = reqAddress.value();
```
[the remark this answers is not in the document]
teqdruid: Calling this function is not actually a no-no since I didn't want to break compatibility. Plus, this function is called every cycle, so... Writing a response would be, however, so that's the right place for a message and there's an assert there.

lib/Dialect/ESI/runtime/cosim/include/esi/cosim/RpcServer.h
```diff
@@ -0,0 +1,53 @@
...
+
+class RpcServer {
+public:
+  ~RpcServer();
+
+  /// Set the manifest and version. There is a race condition here in that the
+  /// RPC server can be started and a connection from the client could happen
+  /// before the manifest is set. TODO: rework the DPI API to require that the
+  /// manifest gets set first.
+  void setManifest(int esiVersion, std::vector<uint8_t> compressedManifest);
```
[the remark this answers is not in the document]
teqdruid: It's gotta be copied somewhere so I was coping at the call site then moving it... which is fairly unintuitive behavior. Change it.

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
teqdruid: IIRC, I couldn't get them to compile at some point though now I can. Updated.

lib/Dialect/ESI/runtime/cpp/include/esi/Utils.h
teqdruid: `empty()` needs to lock this and it's a const function.

lib/Dialect/ESI/runtime/cpp/include/esi/Utils.h
teqdruid: this one was a copy-paste. fixed.

lib/Dialect/ESI/runtime/cosim/lib/RpcServer.cpp
teqdruid: Actually, I don't think I need to lock it given that shutdown is an atomic assignment. Great suggestion.

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
teqdruid: Yeah, though this pre-dates this change. The original thought was that I didn't want to keep writing `std::string`, but I should just import string into the local namespace. I'll come back and change this after this pr. (If I remember.)

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
```diff
@@ -80,147 +108,212 @@ CosimAccelerator::CosimAccelerator(Context &ctxt, string hostname,
...
+    StartCall();
+    StartRead(&incomingMessage);
   }
-  virtual void disconnect() override {
-    if (ep)
-      ep->returnForUse();
+
+  /// Gets called when there's a new message from the server. It'll be stored in
+  /// `incomingMessage`.
+  void OnReadDone(bool ok) override {
```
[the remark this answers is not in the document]
teqdruid: This is dictated by gRPC.

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
```diff
@@ -277,11 +395,11 @@ Service *CosimAccelerator::createService(Service::Type svcType,
   }
 
   if (svcType == typeid(services::MMIO)) {
-    return new CosimMMIO(rpcClient->getLowLevel());
+    // return new CosimMMIO(rpcClient->getLowLevel());
```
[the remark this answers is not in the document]
teqdruid: Debated that, but returning nullptr is acceptable and indicates that a service isn't supported. Crashing is inconsistent with that.

teqdruid:
> Great to see that the diff in the non-cosim-backend code was relatively minor.

Actually, it was non-existent!

> Based on the implementation and testing you've done on this, how is your initial gut feeling for gRPC vs capnproto? I'm thinking both in terms of compile-time and runtime-debugging that you went through.

It was definitely easier to debug both compilation and runtime bugs; BUT, only because I'm more familiar with standard threading systems than libkj's async model. The 'googleability' aspect surprisingly didn't help much (the runtime errors I ran into didn't come up with any relevant results), though the larger number of examples did help.

Capnp RPC is more expressive, which I miss. It supports non-unary arguments/results and returning a capability (i.e. API), both of which help model a moderately complex system in a cleaner way. Also, capnp proto doesn't have NEARLY as many complex dependencies which probably would have made a `FetchContent` approach feasible. I would probably prefer Capnp if not for the C++ RPC implementation using libkj.

Actually, we would probably still be using capnp if libkj documentation was much better (notably the [tour](https://github.com/capnproto/capnproto/blob/v2/kjdoc/tour.md) document) years ago when I first started using it and libkj had evolved to use newer C++ APIs instead of sticking with its own stuff (which pre-dated the newer C++ APIs). I just got fed up over the years dealing with libkj and didn't take the time to properly learn it based on the code documentation when I had time at the outset.

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
teqdruid: 90954e2ae8cd7c230755e83754a87af52bfb8a6e

teqdruid:
Note that the commit message on this is wrong. It does not break the build in a horrible way. I just forgot to remove that part from the message when merging. 🤦

PR #7219, opened 2024-06-20
[Moore] Fix the stacking fault caused by cast and remove unused headers

hailongSun2000:
Cause every variable/net has been used by other ops. We must first read it. For example:
```
module foo(input clk, en, cin);
  reg q;
  always@(posedge clk) begin
          if(en) begin
             q <= cin;
          end
  end
endmodule      
```
```
moore.module @foo(in %clk : !moore.l1, in %en : !moore.l1, in %cin : !moore.l1) {
    %clk_0 = moore.net name "clk" wire : <l1>
    %en_1 = moore.net name "en" wire : <l1>
    %cin_2 = moore.net name "cin" wire : <l1>
    %q = moore.variable : <l1>
    moore.procedure always {
      %0 = moore.read %clk_0 : l1
      moore.wait_event posedge %0 : l1
      %1 = moore.read %en_1 : l1
      %2 = moore.conversion %1 : !moore.l1 -> i1
      scf.if %2 {
        %3 = moore.read %cin_2 : l1
        moore.nonblocking_assign %q, %3 : l1
      }
    }
    moore.assign %clk_0, %clk : l1
    moore.assign %en_1, %en : l1
    moore.assign %cin_2, %cin : l1
    moore.output
  }
```
`SimplifyProcedures` is through visit `moore.read` to get the user, but the user may be `moore.net` such as "clk". This situation will cause a segmentation fault.

### DIFF
--- a/include/circt/Dialect/Moore/MoorePasses.td
+++ b/include/circt/Dialect/Moore/MoorePasses.td
@@ -15,10 +15,10 @@
 
 include "mlir/Pass/PassBase.td"
 
-def SimplifyProcedures : Pass<"simplify-procedures", "moore::SVModuleOp"> {
+def SimplifyProcedures : Pass<"moore-simplify-procedures", "moore::SVModuleOp"> {
     let summary = "Simplify procedures";
     let description = [{
-      Cause we want to introduce mem2reg in the moore dialect to eliminate the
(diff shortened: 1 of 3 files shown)


fabianschuiki (commented):
Awesome, thanks!

PR #7221, opened 2024-06-20
[LowerClass] Run path tracking sequentially

uenoku:
This fixes a race condition was introduced for now by removing pallallesim


mikeurbach (approved):
Makes sense to me, thanks for the fix.

PR #7222, opened 2024-06-20
README.md: remove link to dead build

youngar:
The build flow which attempted to build CIRCT against LLVM head was deleted a long time ago, and our README should not be referencing it.


dtzSiFive (approved):

youngar:
(manually cancelled the builds)

PR #7223, opened 2024-06-20
Bump LLVM 

girishpai:
Notable commits:
https://github.com/llvm/llvm-project/pull/95700
https://github.com/llvm/llvm-project/pull/95057


tools/llhd-sim/llhd-sim.cpp
```diff
@@ -156,6 +157,7 @@ static LogicalResult applyMLIRPasses(ModuleOp module) {
   PassManager pm(module.getContext());
 
   pm.addPass(createConvertLLHDToLLVMPass());
+  pm.addPass(::mlir::createReconcileUnrealizedCastsPass());
```
youngar, line 161: cc @maerhart

lib/Conversion/HWToLLVM/HWToLLVM.cpp
```diff
@@ -675,9 +675,6 @@ void HWToLLVMLoweringPass::runOnOperation() {
   populateHWToLLVMTypeConversions(converter);
 
   LLVMConversionTarget target(getContext());
-  target.addLegalOp<UnrealizedConversionCastOp>();
-  target.addLegalOp<ModuleOp>();
-  target.addLegalDialect<LLVM::LLVMDialect>();
   target.addIllegalDialect<hw::HWDialect>();
```
youngar, line 678: These were already implicit in the LLVMConversionTarget - no need to set them up again. For ModuleOp we don't need to mark it legal since we are doing a `applyPartialConversion`.  Not technically needed for the bump to work, the real change is modifying the test to run the `reconcile-unrealized-casts` afterwords.

lib/Conversion/LLHDToLLVM/LLHDToLLVM.cpp
```diff
@@ -1884,14 +1884,13 @@ void LLHDToLLVMLoweringPass::runOnOperation() {
 
   LLVMConversionTarget target(getContext());
   target.addIllegalOp<InstOp>();
-  target.addLegalOp<UnrealizedConversionCastOp>();
```
youngar, line 1887: Implicit in the LLVMConversionTarget.

lib/Conversion/LLHDToLLVM/LLHDToLLVM.cpp
```diff
@@ -1884,14 +1884,13 @@ void LLHDToLLVMLoweringPass::runOnOperation() {
...
   cf::populateControlFlowToLLVMConversionPatterns(converter, patterns);
   arith::populateArithToLLVMConversionPatterns(converter, patterns);
 
   // Apply the partial conversion.
   if (failed(
           applyPartialConversion(getOperation(), target, std::move(patterns))))
-    signalPassFailure();
+    return signalPassFailure();
```
youngar, line 1893: Started early returning here - doesn't seem like a good idea to keep trying to convert the operations.

lib/Conversion/LLHDToLLVM/LLHDToLLVM.cpp
```diff
@@ -1907,21 +1906,10 @@ void LLHDToLLVMLoweringPass::runOnOperation() {
   populateCombToArithConversionPatterns(converter, patterns);
   arith::populateArithToLLVMConversionPatterns(converter, patterns);
 
-  target.addLegalDialect<LLVM::LLVMDialect>();
```
youngar, line 1910: LLVMDialect is already marked legal.

lib/Conversion/LLHDToLLVM/LLHDToLLVM.cpp
```diff
@@ -1907,21 +1906,10 @@ void LLHDToLLVMLoweringPass::runOnOperation() {
...
-  patterns.clear();
-
-  mlir::populateReconcileUnrealizedCastsPatterns(patterns);
-  target.addIllegalOp<UnrealizedConversionCastOp>();
-
-  // Apply the full conversion.
-  if (failed(applyFullConversion(getOperation(), target, std::move(patterns))))
-    signalPassFailure();
+    return signalPassFailure();
 }
```
youngar, line 1913: `populateReconcileUnrealizedCastsPatterns` was removed and we have to instead rely on running the `reconcile-unrealized-casts` pass separately.

test/Conversion/HWToLLVM/convert_aggregates.mlir
```diff
@@ -1,9 +1,9 @@
-// RUN: circt-opt %s --convert-hw-to-llvm | FileCheck %s
+// RUN: circt-opt %s --convert-hw-to-llvm -reconcile-unrealized-casts | FileCheck %s
```
rwy7, line 1:
```suggestion
// RUN: circt-opt %s --convert-hw-to-llvm --reconcile-unrealized-casts | FileCheck %s
```

tools/llhd-sim/llhd-sim.cpp
youngar: I did not add the `reconcile-unrealized-casts` pass in to the `circt-lec` pipeline after `convert-verif-to-smt`.  I am guessing that eitherit  doesn't need the pass because the canonicalizer cleans these up, or we just don't have a test that is catching the issue.
```
  pm.addPass(createConvertVerifToSMT());
  pm.addPass(createSimpleCanonicalizerPass());
```

rwy7 (approved):

lib/Conversion/LLHDToLLVM/LLHDToLLVM.cpp
```diff
@@ -1907,21 +1906,10 @@ void LLHDToLLVMLoweringPass::runOnOperation() {
   populateCombToArithConversionPatterns(converter, patterns);
   arith::populateArithToLLVMConversionPatterns(converter, patterns);
 
-  target.addLegalDialect<LLVM::LLVMDialect>();
   target.addLegalOp<ModuleOp>();
 
-  // Apply a full conversion to remove unrealized conversion casts.
   if (failed(applyFullConversion(getOperation(), target, std::move(patterns))))
```
rwy7, line 1911: IMO this should be a partial conversion (and the LLHD dialect ought to be marked illegal), so that this pass doesn't choke on unknown ops.

maerhart: This pass needs to be refactored anyway, so anything that makes the tests succeed is fine for now.

tools/llhd-sim/llhd-sim.cpp
maerhart: Right now, I cannot think of a case where the canonicalizer is not enough. The cases I'm aware of in the LEC lowering are already dealt with in the pass itself.

maerhart (approved):
Thanks!

youngar:
we have an unsolved issue if we bump LLVM too far, so I changed this PR to bump to right before the issue: https://github.com/llvm/circt/issues/7240

PR #7234, opened 2024-06-25
[FIRRTL] Allow layers under when and match.

dtzSiFive:
Behavior is same as-if the contained operations were not under a layer.

Some layer-related passes such as LayerMerge (for completeness) and LowerLayers (for correctness) now explicitly require ExpandWhen's to execute first.

The FIRRTL specification does not yet permit this.

### DIFF
--- a/test/Dialect/FIRRTL/expand-whens.mlir
+++ b/test/Dialect/FIRRTL/expand-whens.mlir
@@ -592,11 +592,11 @@ firrtl.module @WhenInGroup(in %cond : !firrtl.uint<1>) {
 // Check that expand whens works for layers under when's.
 firrtl.layer @Layer bind {}
 // CHECK-LABEL: firrtl.module @LayerUnderWhen(
-// CHECK-NOT:   firrtl.when
-// CHECK: firrtl.printf %clock, %cond
+// CHECK-NEXT:  firrtl.layerblock @Layer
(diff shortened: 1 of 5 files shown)


youngar (approved):
💯

seldridge (commented):
Can you include tests of a layer block under a when under a layer block? Also, can you include a test of an assert under a when under a layer block?

Do these both work?

test/Dialect/FIRRTL/expand-whens.mlir
```diff
@@ -589,6 +589,20 @@ firrtl.module @WhenInGroup(in %cond : !firrtl.uint<1>) {
   }
 }
 
+// Check that expand whens works for layers under when's.
+firrtl.layer @Layer bind {}
+// CHECK-LABEL: firrtl.module @LayerUnderWhen(
+// CHECK-NOT:   firrtl.when
+// CHECK: firrtl.printf %clock, %cond
```
seldridge, line 596: Can this also check that the `printf` remains under the layerblock?

seldridge (approved):
LGTM

test/Dialect/FIRRTL/expand-whens.mlir
dtzSiFive: You got it, done!

PR #7241, opened 2024-06-26
[ESI Runtime][NFC] Incorporate RPC server into ESICppRuntime

teqdruid:
Since the gRPC server is now going to be used in multiple places AND the RPC server now uses ports 'n' stuff from ESICppRuntime, it is appropriate to move RpcServer into ESICppRuntime proper. This also significantly simplifies the build.

No new code, just movement and CMake changes.


teqdruid:
Doing this now before we do more work on rpc server and merging becomes a mess.

mortbopet (approved):

PR #7243, opened 2024-06-27
[ImportVerilog] Support Generate constructs

angelzzzzz:
Support loop generate constructs and conditional generate constructs.
Because `genvar` only exists in the declaration and the variable exists as a parameter in GenerateBlock, genvar can be skipped.

### DIFF
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -98,6 +98,11 @@ struct MemberVisitor {
   LogicalResult visit(const slang::ast::PortSymbol &) { return success(); }
   LogicalResult visit(const slang::ast::MultiPortSymbol &) { return success(); }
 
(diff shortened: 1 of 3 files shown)


lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -368,6 +373,26 @@ struct MemberVisitor {
...
+    }
+    return success();
+  }
+
+  // Handle generate block array.
+  LogicalResult visit(const slang::ast::GenerateBlockArraySymbol &genArrNode) {
+    for (const auto *member : genArrNode.entries) {
+      if (failed(member->asSymbol().visit(MemberVisitor(context, loc))))
+        return failure();
+    }
```
uenoku, line 392: Oh interesting. Does Slang resolve generate statements in their side?

uenoku (approved):
LGTM 👍

PR #7244, opened 2024-06-27
[Moore] Add extra class declaration for RefType.

hailongSun2000:
Add `getDomain()` and `getBitSize()` for `RefType`, now can directly get the domain and size of its nested type.


hailongSun2000:
Thanks for the code review in advance! :smile:

uenoku (approved):
LGTM

PR #7245, opened 2024-06-27
[ImportVerilog]Dedup module Op

mingzheTerapines:
Dedup module Op while instance generating.


mingzheTerapines:
@fabianschuiki @hailongSun2000  Sorry, last dedup PR was ruined after git rebase, please have a look at this one, thanks.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -130,6 +130,7 @@ struct MemberVisitor {
       // connection for the port.
       if (!expr) {
         auto *port = con->port.as_if<PortSymbol>();
+        port = context.updatePortsSymbol(moduleLowering, port);
```
fabianschuiki, line 133: Instead of using `updatePortsSymbol` which linearly scans through the ports, could you create a `DenseMap` lookup table right after getting `moduleLowering` which associates the ports with their `SyntaxNode *`? Something like:
```
DenseMap<const slang::syntax::SyntaxNode *, const Symbol *> portsBySyntaxNode;
for (auto port : moduleLowering->ports)
  portsBySyntaxNode.insert({port.getSyntax(), port});
```
We could also store that in the lowering directly.

Then instead of calling `updatePortsSymbol` multiple times, you can map from the instance's port to the module lowering's port once at the beginning of the for loop after `expr = ...`:
```
const auto *expr = con->getExpression();
const auto *port = portsBySyntaxNode.lookup(con->port.getSyntax());
assert(port);
```
After that the original code should just work without requiring more port remapping.

In case the ports don't have a proper `SyntaxNode` associated with them, you can also use the port's name as a key for the lookup table. But the syntax node would be cleaner and faster 🙂

mingzheTerapines: Cool! That will avoid duplicate mapping.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -188,6 +195,10 @@ struct MemberVisitor {
                          : context.convertLvalueExpression(*expr);
         if (!value)
           return failure();
+        port = moduleLowering->portsBySyntaxNode.contains(con->port.getSyntax())
+                   ? moduleLowering->portsBySyntaxNode.lookup(
+                         con->port.getSyntax())
+                   : port;
```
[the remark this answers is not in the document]
mingzheTerapines: Great! Elegent code!

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -130,6 +130,13 @@ struct MemberVisitor {
       // connection for the port.
       if (!expr) {
         auto *port = con->port.as_if<PortSymbol>();
+        for (const auto &existPort : moduleLowering->ports)
+          if (port->name == existPort.ast.name) {
+            moduleLowering->portsBySyntaxNode.insert(
+                {con->port.getSyntax(), &existPort.ast});
+            port = &existPort.ast;
+          }
```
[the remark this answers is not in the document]
mingzheTerapines: Sorry, modulerLowering is created in convertModuleHeader(), and this interface only handle isntanse body( = InstNode.body). However the ports information is in instNode.getPortConnections(). I am not sure if this information belongs to module header. 
Easily I can show you what it will be.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -459,15 +470,77 @@ Context::convertCompilation(slang::ast::Compilation &compilation) {
...
+  if (slot) {
+    if (portConnections) {
+      for (const auto *con : *portConnections) {
+        auto *port = con->port.as_if<PortSymbol>();
+        for (const auto &existingPort : slot->ports)
+          if (port->name == existingPort.ast.name)
+            slot->portsBySyntaxNode.insert(
+                {port->getSyntax(), &existingPort.ast});
+      }
+    }
```
fabianschuiki, line 540: I don't think you need the `portConnections` for this. The `portsBySyntaxNode` should contain a mapping from the syntax node of a port to the port of the module. Something like:
```suggestion
    for (const auto &port : slot->ports)
      slot->portsBySyntaxNode.insert({port.ast.getSyntax(), &port.ast});
```
When you're dealing with instances, you can use the `portsBySyntaxNode` map to map from the instance's duplicated port to the original port on the already-lowered module. Something like:
```
auto *modulePort = moduleLowering->portsBySyntaxNode.lookup(instancePort->getSyntax());
```
The instancePort is a copy of modulePort. They are separate objects, but I think their syntax is a pointer to the exact same AST node. That's what we use as a key to map from the instance's duplicated port to the original port of the lowering.

mingzheTerapines: That's a really deep understanding which I can not get this information from AST.json! I was thinking to mapping duplicate instance port symbol to existing instance's relating module port symbol, which is somehow complex. I have imporve my code.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -470,11 +480,64 @@ ModuleLowering *
...
+        break;
+      }
+    }
+  }
 
   auto &slot = modules[module];
-  if (slot)
+  if (slot) {
+    for (const auto &port : slot->ports)
+      slot->portsBySyntaxNode.insert({port.ast.getSyntax(), &port.ast});
```
fabianschuiki, line 537: You don't have to do this every time a module is looked up. Instead, you can move this all the way to the bottom, just before `return &lowering;`. That way the `portsBySyntaxNode` is populated once when the lowering is created, and it remains valid. Then this line of code here can just be `if (slot) return slot.get();` again :slightly_smiling_face:

mingzheTerapines: Sorry, I was still thinking mapping  port symbol to symbol which should be done every time a module is looked up. You are very right, this map could be generated just when new lowering is generated. This will imporve running a lot.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -20,10 +20,39 @@ module Empty;
...
+module Dedup;
+  module NestedA(input wire a,
+  input wire b,
+  output wire [3:0] c);
+  endmodule
+  module NestedB #(parameter p = 32)
+  (input wire a,
+  input wire b,
+  output wire [3:0] c);
+  endmodule
```
fabianschuiki, line 33: Could you move these module declarations outside of the `Dedup` module, and name them something like `DedupB` and `DedupC`, or `DedupInnerA` and `DedupInnerB`? I don't think you need the nesting to show that dedup works, and we already have a separate test further down where we check the nesting.

fabianschuiki (approved):
Very cool! Minor comment regarding the nested test, but besides that this is very nice!

test/Conversion/ImportVerilog/basic.sv
mingzheTerapines: Thanks for your advice, I will be careful for other PRs.

Issue #6983, opened 2024-05-03
 Elaborate chisel type annotation from firtool to generate debug information for the Tywaves project


rameloni:
@fabianschuiki I created this issue to keep talking about the conversation we started here: https://github.com/chipsalliance/chisel/issues/4015#issuecomment-2077714172

I added attributes to debug variable to store the chisel type name and, eventually, the constructor parameters in the debug dialect (and maybe in `dbg.struct` and `dbg.array` as well).

However, I may need to create a dbg variable for each sub field, recursively, of aggregates. 
Is there a reason why values in `dbg.struct`/`dbg.array` are of Firrtl types rather than other `dbg.variable`s in `MaterializeDebugInfo`?

sequencer:
I think we can to expose the chisel type via compiler plugin like what we did in the source info. But what on earth is a chisel type? That’s the fundamental question(we need specification)

rameloni:
> I think we can to expose the chisel type via compiler plugin like what we did in the source info. 

I already managed to expose source level information from chisel to firrtl and firtool. In the meantime I also managed to read this information from firtool, processing it into proper debug variables and updating hgldd/emitting a new file format.

> But what on earth is a chisel type? That’s the fundamental question(we need specification)

Whit chisel type refer to the type of the scala variable in the source chisel code.
In the following example:

```scala
class MyBundle(size: Int) extends Bundle {
   val a = Bool()
   val b = SInt(size.W)
   val nested = new Bundle {
      val x = UInt(8.W)
   }
}
val inBundle  = IO(Input(new MyBundle(10)))
val wireBundle = Wire(new MyBundle(7))
val outBundle = IO(Output(new MyBundle(9)))) 
```
The types would be
| Variable          | Type                  |
| ----------------- | --------------------- |
| inBundle          | `IO[MyBundle(10)]`    |
| wireBundle        | `Wire[MyBundle(7)]`   |
| outBundle         | `IO[MyBundle(9)]`     |
| inBundle.a        | `IO[Bool]`            |
| inBundle.b        | `IO[SInt<10>]`        |
| inBundle.nested   | `IO[AnonymousBundle]` |
| inBundle.nested.x | `IO[UInt<8>]`         |
| ...               | ...                   |

Namely, I refer to anything represented in this table, including user-defined types:
https://raw.githubusercontent.com/chipsalliance/chisel/main/docs/src/images/type_hierarchy.svg?sanitize=true
[...]

PR #7246, opened 2024-06-27
[DebugInfo][HGLDD] 

rameloni:
This PR adds new functionality to CIRCT for emitting source language type information in HGLDD 
https://github.com/llvm/circt/issues/6983.

This new functionality support the new backend of my project. [Tywaves: a type-based waveform viewer](https://github.com/rameloni/tywaves-chisel-demo/wiki). Tywaves adds support for Chisel to [Surfer](https://gitlab.com/surfer-project) through HGLDD, however it misses of Chisel-type information and a proper documentation.

So, in addition to the [changes I made to CIRCT](https://github.com/rameloni/tywaves-chisel-demo/wiki/Changes-in-CIRCT), I also created a library for reading and [documenting HGLDD in rust](https://github.com/rameloni/tywaves-chisel-demo/wiki/Waveform-GUI:-Tywaves%E2%80%90rs-and-Surfer#hgldd-documentation). I noticed that HGLDD misses a formal specification, so the rust documentation is a first attempt to provide more formal description of the fields contained in the file.

### DIFF
--- a/include/circt/Analysis/DebugInfo.h
+++ b/include/circt/Analysis/DebugInfo.h
@@ -23,6 +23,13 @@ namespace detail {
 struct DebugInfoBuilder;
 } // namespace detail
 
+struct DISourceLang {
+  /// The name of the type.
+  StringAttr typeName;
+  /// The constructor parameters of the type.
+  ArrayAttr params;
+};
+
 struct DIModule {
   /// The operation that generated this level of hierarchy.
   Operation *op = nullptr;
@@ -36,6 +43,10 @@ struct DIModule {
   bool isExtern = false;
   /// If this is an inline scope created by a `dbg.scope` operation.
   bool isInline = false;
+
+  // TODO: add support for source language type info to a module
+  // /// The source language type of this module.
+  // DISourceLang sourceLangType;
 };
 
 struct DIInstance {
@@ -54,6 +65,9 @@ struct DIVariable {
   LocationAttr loc;
   /// The SSA value representing the value of this variable.
   Value value = nullptr;
+
+  /// The source language type of this module.
+  DISourceLang sourceLangType;
 };
 
 /// Debug information attached to an operation and the operations nested within.
--- a/include/circt/Analysis/DebugInfo.h
+++ b/include/circt/Analysis/DebugInfo.h
@@ -44,9 +44,8 @@ struct DIModule {
   /// If this is an inline scope created by a `dbg.scope` operation.
   bool isInline = false;
 
-  // TODO: add support for source language type info to a module
-  // /// The source language type of this module.
-  // DISourceLang sourceLangType;
+  /// The source language type of this module.
+  DISourceLang sourceLangType;
 };
 
 struct DIInstance {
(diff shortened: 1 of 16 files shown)


rameloni:
The related PR in chisel: https://github.com/chipsalliance/chisel/pull/4224

darthscsi:
Some general comments:
We are trying very hard to not add new annotations.
I thought the current debug efforts were grabbing chisel types and only falling back to firrtl types if asked.  It might be this isn't done yet, in which case the efforts should be coordinated.  @fabianschuiki or @prithayan might be able to shed more light.

rameloni:
> Some general comments:
> We are trying very hard to not add new annotations.
> I thought the current debug efforts were grabbing chisel types and only falling back to firrtl types if asked.  It might be this isn't done yet, in which case the efforts should be coordinated.  @fabianschuiki or @prithayan might be able to shed more light.

@darthscsi thanks for the comment.
Yeah, I found that Intrinsics might be a better option to do that. However, I'm not sure if I can make it before the deadline of my thesis. That's why I'm using at the moment the annotations, since I've found they were slightly easier to manage on average between Chisel and CIRCT sides

rameloni:
I have also noticed another thing. HGLDD also missed reference to enum variants, so I'm trying to add that. 
I've seen that the `hw` dialect has a definition of `EnumType` however it seems not to be generated?!? I'm not sure, am I wrong?
Chisel passes the enum variants through 3 annotations (`EnumComponentAnnotation`, `EnumDefAnnotation`, and `EnumVecAnnotation`) but they are dropped in the LowerAnnotations pass, is there a reason for that? Apart from the fact you are trying to replace the annotation mechanism.

fabianschuiki:
Hey @rameloni, sorry for taking so long to respond. Your work is very exciting! Thanks for pushing on getting debug information communicated from Chisel to CIRCT :partying_face:.

I agree with @darthscsi that using intrinsics here would be very nice. They would allow you to directly annotate information onto Chisel and FIRRTL values directly in the IR, without going through the annotations mechanism which we are trying to phase out as much as possible. Intrinsics should be almost identical in handling as annotations, with the difference that you have more control over where they appear in the IR and you never have to magically materialize MLIR ops from a JSON data structure.

One concern regarding debug info output: the HGLDD format is used by Synopsys VCS/Verdi to ingest debug information. I'm pretty sure that changes to the format, especially things that add new constructs and fields and operators, will break the Synopsys-side implementation. The format is also tailored to work with VCS-internal code patterns and limitations. Since you are working on your own waveform viewer, would it be possible to simply create a new, more streamlined debug info format for your viewer? You could just copy-paste the HGLDD emission code with your changes, if you want to stick with those. But I'm pretty sure you've found a few things in HGLDD to be weird or cumbersome, and creating your own output format would be an opportunity to make things more streamlined for you :smiley:. But it would be great to keep the HGLDD format unchanged since we don't control the other half of its implementation.

rameloni:
Hey @fabianschuiki, thanks for the feedback. At the beginning I created a draft file format different from HGLDD but later I thought that extending HGLDD could've been also an opportunity to make the new features available to tools using that. I also though that two formats could create confusion :sweat_smile:. My thesis deadline is soon and, at this point, I am not sure if creating a new format is a good idea, emitting the updated HGLDD as "another" format is super fast instead.

Regarding intrinsics, I'd really like to replace my annotations with them, I'm trying to finish quickly my remaining tasks to do so by the end of July, but it is more likely that I'll make it after my defense.

Also, annotations are already used for the Chisel enums, which means that also those would need to be replaced.

PR #7247, opened 2024-06-27
[NFC][Seq] Remove unnecessary dependency of Seq on SV

TaoBi22:
Seq currently depends on SV according to CMake, which seems to be unnecessary (modulo a fix to the SVToSeq CMakeLists) and means adding certain dependencies will introduce unwanted cycles.

### DIFF
--- a/lib/Conversion/SeqToSV/CMakeLists.txt
+++ b/lib/Conversion/SeqToSV/CMakeLists.txt
@@ -13,4 +13,5 @@ add_circt_conversion_library(CIRCTSeqToSV
   LINK_LIBS PUBLIC
   CIRCTEmit
   CIRCTSeq
+  CIRCTSV
 )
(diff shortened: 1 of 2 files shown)


maerhart (approved):
LGTM

dtzSiFive:
The dependency cycle does need to be fixed, but presently it looks like Seq has transforms that depend on SV.
As a result, the link dependency is required for the Seq library to have what it needs to function.

See `SVPasses.td`  and grep for `sv::` in `lib/Dialect/Seq`, looks like `HWMemSimImpl` and `LowerSeqHLMem` each require the SV dialect, as they create operations from it.

One way to address this is to split our Dialect libraries a bit, in particular splitting Transforms from the IR portion specifically.
Following upstream MLIR's organization seems reasonable.

dtzSiFive (approved):
LGTM, the dependency is from the transforms and that still links in `CIRCTSV`.

I keep backfiring because I expect the library at one level to include its children (and perhaps would find pushing this into an `IR` subdirectory or something a bit clearer).

TaoBi22:
@dtzSiFive agreed, a clearer distinction between IR and transform would definitely be helpful here. Thanks for looking into this!!

PR #7248, opened 2024-06-27
[FIRRTL] Reject abstract resets on extmodule's.

dtzSiFive:
When parsing FIRRTL >= 4.0.0, check and reject use of abstract reset on extmodule's.

https://github.com/chipsalliance/firrtl-spec/pull/181

Previously implemented as part of #6731, but used in practice so deferred.  Let's revisit now.


dtzSiFive:
Draft only because not certain want to land this just yet (and when ready would be nice to land w/FIRRTL spec change, and if that's after 4.0.0 update this accordingly)

dtzSiFive:
Spec PR landed, keeping as draft for now pending cleaning out reliance on this.

PR #7249, opened 2024-06-27
[NFC][SV] Remove unnecessary dependency of SV on Verif

TaoBi22:
Same situation as #7247, SV depends on Verif but that seems to be unnecessary as long as om-linker and HWToBTOR2 CMakeLists are updated to link in Verif

### DIFF
--- a/tools/om-linker/CMakeLists.txt
+++ b/tools/om-linker/CMakeLists.txt
@@ -14,6 +14,7 @@ target_link_libraries(om-linker PRIVATE
   CIRCTOMTransforms
   CIRCTSV
   CIRCTSupport
+  CIRCTVerif
 
   MLIRBytecodeReader
   MLIRBytecodeWriter
--- a/lib/Conversion/HWToBTOR2/CMakeLists.txt
+++ b/lib/Conversion/HWToBTOR2/CMakeLists.txt
@@ -12,5 +12,6 @@ add_circt_conversion_library(CIRCTHWToBTOR2
   CIRCTSV
   CIRCTComb
   CIRCTSeq
+  CIRCTVerif
   MLIRTransforms
 )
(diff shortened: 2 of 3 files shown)


dtzSiFive (approved):
LGTM.

In ODS, SV is marked as depending on HW and Comb, which matches what we link in after this PR.

(Having these two aligned seems like a good rule of thumb)

dtzSiFive (changes requested):
Hold on, SVPasses.td mentions that extract-test-code depends on Verif.

dtzSiFive:
Sorry for the back-and-forth!

Looks like we do link the Verif dialect into SVTransforms, so maybe this is good to go.  Going to circle back (and maybe others take a look/have thoughts).

Thanks for pushing on this! :+1:.

TaoBi22:
> Sorry for the back-and-forth!
> 
> Looks like we do link the Verif dialect into SVTransforms, so maybe this is good to go. Going to circle back (and maybe others take a look/have thoughts).
> 
> Thanks for pushing on this! 👍.

No worries, thanks for taking a look! Strictly only one of this and #7247 is needed to break the cycle I need to break (so I can have Verif depend on Seq), just figured it was worth tidying them both up while I was at it (although #7247 has the same case with a LowerSeqHLMem dep on SV that's linked for CIRCTSeqTransforms).

maerhart:
> Looks like we do link the Verif dialect into SVTransforms, so maybe this is good to go. 

I agree with this analysis. Linking the transform/conversion libraries against Verif should be enough since the dialect library does not seem to use anything from Verif. Thus, the dependency of `CIRCTSV` on `CIRCTVerif` can go.

tools/om-linker/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ target_link_libraries(om-linker PRIVATE
   CIRCTOMTransforms
   CIRCTSV
   CIRCTSupport
+  CIRCTVerif
```
dtzSiFive, line 17: This makes sense, the om-linker itself is what needs the Verif dialect.  Thanks!

dtzSiFive (approved):
LGTM, thanks!

TaoBi22:
Thanks again for looking into this @dtzSiFive!

PR #7251, opened 2024-06-28
[Moore] Add AssignedVarOp and canonicalization for VariableOp.

hailongSun2000:
The canonicalization for VariableOp is aimed at merging the "easy use" of the variable and its value(continuous assignment) into a new op called `assigned_variable`. Don't handle `blocking_assign` and `nonBlocking_assign` due to the dominance. 
The "easy use" is assigning a value to a variable with the whole bit-width, in other words, exclude the bits slice, concatenation operator, etc.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -198,6 +198,7 @@ def VariableOp : MooreOp<"variable", [
     `` custom<ImplicitSSAName>($name) ($initial^)? attr-dict
     `:` type($result)
   }];
+  let hasCanonicalizeMethod = true;
 }
 
 def NetKindAttr : I32EnumAttr<"NetKind", "Net type kind", [
@@ -249,8 +250,21 @@ def NetOp : MooreOp<"net", [
   );
   let results = (outs Res<RefType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
-    ``custom<ImplicitSSAName>($name) $kind ($assignment^)? attr-dict
+    `` custom<ImplicitSSAName>($name) $kind ($assignment^)? attr-dict
     `:` type($result)
+  }];
+}
+
+def AssignedVarOp : MooreOp<"assigned_variable", [
+  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmResultNames"]>,
+  TypesMatchWith<"initial value and variable types match",
+    "result", "initial", "cast<RefType>($_self).getNestedType()">
+]> {
+  let summary = "The copy of variable must have the initial value";
+  let arguments = (ins StrAttr:$name, UnpackedType:$initial);
+  let results = (outs Res<RefType, "", [MemAlloc]>:$result);
+  let assemblyFormat = [{
+    `` custom<ImplicitSSAName>($name) $initial attr-dict `:` type($result)
   }];
 }
 
--- a/lib/Dialect/Moore/MooreOps.cpp
+++ b/lib/Dialect/Moore/MooreOps.cpp
@@ -265,11 +265,41 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
   return std::nullopt;
 }
 
+LogicalResult VariableOp::canonicalize(VariableOp op,
+                                       ::mlir::PatternRewriter &rewriter) {
+  Value initial;
+  for (auto *user : op->getUsers())
+    if (isa<ContinuousAssignOp>(user) &&
+        (user->getOperand(0) == op.getResult())) {
+      // Don't canonicalize the multiple continuous assignment to the same
+      // variable.
+      if (initial)
+        return failure();
+      initial = user->getOperand(1);
+    }
+
+  if (initial) {
+    rewriter.replaceOpWithNewOp<AssignedVarOp>(op, op.getType(), op.getName(),
+                                               initial);
+    return success();
+  }
+
+  return failure();
+}
+
 //===----------------------------------------------------------------------===//
 // NetOp
 //===----------------------------------------------------------------------===//
 
 void NetOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
+  setNameFn(getResult(), getName());
+}
+
+//===----------------------------------------------------------------------===//
+// AssignedVarOp
+//===----------------------------------------------------------------------===//
+
+void AssignedVarOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
   setNameFn(getResult(), getName());
 }
 
--- a/test/Dialect/Moore/canonicalizers.mlir
+++ b/test/Dialect/Moore/canonicalizers.mlir
@@ -9,3 +9,31 @@ func.func @Casts(%arg0: !moore.i1) -> (!moore.i1, !moore.i1) {
   // CHECK: return %arg0, %arg0
   return %0, %1 : !moore.i1, !moore.i1
 }
+
+// CHECK-LABEL: moore.module @SingleAssign
+moore.module @SingleAssign() {
+  // CHECK-NOT: moore.variable
+  // CHECK: %a = moore.assigned_variable %0 : <i32>
+  %a = moore.variable : <i32>
+  // CHECK: %0 = moore.constant 32 : i32
+  %0 = moore.constant 32 : i32
+  // CHECK: moore.assign %a, %0 : i32
+  moore.assign %a, %0 : i32
+  moore.output
+}
+
+// CHECK-LABEL: moore.module @MultiAssign
+moore.module @MultiAssign() {
+  // CHECK-NOT: moore.assigned_variable
+  // CHECK: %a = moore.variable : <i32>
+  %a = moore.variable : <i32>
+  // CHECK: %0 = moore.constant 32 : i32
+  %0 = moore.constant 32 : i32
+  // CHECK: moore.assign %a, %0 : i32
+  moore.assign %a, %0 : i32
+  // CHECK: %1 = moore.constant 64 : i32
+  %1 = moore.constant 64 : i32
+  // CHECK: moore.assign %a, %1 : i32
+  moore.assign %a, %1 : i32
+  moore.output
+}
--- a/tools/circt-verilog/circt-verilog.cpp
+++ b/tools/circt-verilog/circt-verilog.cpp
@@ -224,6 +224,7 @@ static LogicalResult populateMooreTransforms(mlir::PassManager &pm) {
   modulePM.addPass(moore::createLowerConcatRefPass());
   modulePM.addPass(moore::createSimplifyProceduresPass());
   pm.addPass(mlir::createMem2Reg());
+
   // TODO: like dedup pass.
 
   return success();


hailongSun2000:
Appendix:
```
/home/phoenix/work/HDLBits/test01.sv:4:10: error: Unsupported drive the same wire 'l' two times now
  assign l = 1'b0;
         ^
module {
  moore.module @Foo() {
    %l = moore.net wire : <l1>
    %l_0 = moore.assigned_variable name "l" %1 : <l1>
    %0 = moore.constant true : i1
    %1 = moore.conversion %0 : !moore.i1 -> !moore.l1
    moore.assign %l, %1 : l1
    %2 = moore.constant false : i1
    %3 = moore.conversion %2 : !moore.i1 -> !moore.l1
    moore.assign %l, %3 : l1
    moore.output
  }
}
```

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
```diff
@@ -0,0 +1,79 @@
...
+      }
+    });
+  });
+}
+
+void MergeAssignmentsPass::runOnOperation() {
+  OpBuilder builder(&getContext());
+
+  // Use to collect the easy declaration and its value.
+  DenseMap<Value, Value> assignments;
```
uenoku, line 67: Iteration order of `DenseMap` is non-deterministic so we should avoid using it when creating IR. You can use MapVector instead.

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
```diff
@@ -0,0 +1,79 @@
...
+  return std::make_unique<MergeAssignmentsPass>();
+}
+
+// TODO: The net can be driven multiple times. However, the related rule is
+// complicated. So we will implement this in the future.
+
+// Only collect the easy declaration and its value at module level.
+static void collectAssignmets(SVModuleOp moduleOp,
+                              DenseMap<Value, Value> &assignments) {
+  moduleOp->walk([&](Operation *op) {
```
uenoku, line 48: You don't need inner TypeSwitch. 

```suggestion
  moduleOp->walk([&](ContinuousAssignOp op) {
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -246,11 +249,23 @@ def NetOp : MooreOp<"net", [
   );
   let results = (outs Res<RefType, "", [MemAlloc]>:$result);
   let assemblyFormat = [{
-    ``custom<ImplicitSSAName>($name) $kind ($assignment^)? attr-dict
+    `` custom<ImplicitSSAName>($name) $kind ($assignment^)? attr-dict
     `:` type($result)
   }];
 }
 
+def AssignedVarOp : MooreOp<"assigned_variable", [
```
uenoku, line 257: Do we need a new op? NetOp seems to have an `assignment` operand so is that same?

uenoku (commented):
I think it's reasonable to make SVModuleOp graph region op for progressive lowering to core dialects. However it's a big change for Moore dialect design so would be nice to have consensus on this change.

hailongSun2000:
I'll separate the changing of `SVModule` into a different PR.

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
hailongSun2000: I'll tweak it later. And keep it in my mind.

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
hailongSun2000: From the perspective of the design of `AssignedVarOp`, I think we can only handle continuous assignments. And due to the dominance of non-continuous assignments. We can not merge variable/net and continuous assignment into one op called `assigned_var`. I'll tweak it later.

include/circt/Dialect/Moore/MooreOps.td
hailongSun2000: You're totally right! `NetOp` has an `assignment` operand. Maybe I think I need to rename it. Like `def AssignedDeclOp : MooreOp<"assigned_decl(aration)", [`, what do @fabianschuiki  @uenoku  think? By the way, slang will throw an error, if we drive the same variable multiple times.

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
uenoku: > From the perspective of the design of AssignedVarOp, I think we can only handle continuous assignments.

I meant ` moduleOp->walk([&](ContinuousAssignOp op) {` is equivalent to `moduleOp->walk([&](Operation*op) {
TypeSwith(op)...`.

hailongSun2000: I understood. I added TypeSwitch here to reserve a case branch to handle some situations I have not considered now. Because I'm not sure whether I think enough. But then I found that non-continuous assignments would trigger the dominance error. So I left this comment.

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: Hmmm yeah good question... SV differentiates between nets and variables in quite a few places. My gut feeling is that turning a `VariableOp` into a `NetOp` with an assignment isn't going to be correct in general :thinking:. The `AssignedVariableOp` feels like a good step, because it retains the variable-ness but makes it explicit that it is continuously assigned a specific value.

I'm pretty sure we'll be able to write some simplification passes that take `always_comb` procedures that write to variables through `moore.blocking_assign` and convert that into a simple `moore.assigned_variable` for simplicity. Then only the truly annoying uses of variables would remain (the ones interacting with the event queue, have delays, etc.).

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
```diff
@@ -0,0 +1,79 @@
...
+    TypeSwitch<Operation *>(op).Case<ContinuousAssignOp>([&](auto op) {
+      if (isa<VariableOp, NetOp>(op.getDst().getDefiningOp())) {
+        if (!assignments.lookup(op.getDst()))
+          assignments[op.getDst()] = op.getSrc();
+        else
+          mlir::emitError(op.getLoc())
+              << "Unsupported drive the same wire '"
+              << op.getDst().template getDefiningOp<NetOp>().getName()
+              << "' two times now";
```
fabianschuiki, line 57: Instead of emitting an error here, could we just skip this variable/net op and not try to merge its assignment? It's legal for a net op to have multiple continuous assignments for example, and having this code here would cause the compiler to error out despite the legal use case.

The assignment merging feels like an opportunistic optimization that brings the IR into a form that's easier to analyze, but it's technically not needed for correctness. We could always keep the original var/net op and assignment op and lower them into `llhd.sig` and `llhd.drv`.

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  collectAssignmets(getOperation(), assignments);
+  for (auto assignment : assignments) {
+    auto varName =
+        assignment.first.getDefiningOp()->getAttrOfType<StringAttr>("name");
+
+    builder.setInsertionPointAfterValue(assignment.first);
+    builder.create<AssignedVarOp>(assignment.first.getLoc(),
+                                  assignment.first.getType(), varName,
+                                  assignment.second);
```
fabianschuiki, line 77: I think this might turn a `NetOp` into an `AssignedVarOp` :thinking:, because you are collecting `NetOp` and `VariableOp`s into the `assignments` map.

fabianschuiki:
I'm wondering if this could work as a canonicalization pattern defined on `VariableOp` :thinking:. The pattern could theoretically go through all users of the variable op and if there is a single continuous assignment, the assigned value could be merged into a new `AssignedVariableOp`. But I'm not sure how idiomatic and scalable it is to scan an entire user list in a canonicalization... @uenoku may have an opinion on that :smirk:

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
hailongSun2000: Have substance! I'll tweak this error.

Yeah, I keep the original var/net and assignment ops according to what you meant.  However, I have a question about lowering them into `llhd.sig` and `llhd.drv`.  If we keep them, they also slow the speed of the simulation similar to directly lower into `llhd.sig` and `llhd.drv` without `assigned_variable`, right?

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
hailongSun2000: Yeah, `NetOp` also is merged into `AssignedVarOp`. What you meant was only merge `variable` and `assignment` into assigned_variable, and don't handle the `net` and its `assignment`, right?

fabianschuiki: Yes exactly, we shouldn't change nets into variables or vice versa. I think it's still valuable to look at `NetOp`s and continuous assignments to them and if there is one single unique assignment to fold that into the `NetOp` itself, otherwise we'll just leave it as it is. That could be its own canonicalization though 😃.

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
fabianschuiki: Yes I think the simulation will be slower. But it will still be correct. From that point of view, merging `var+assign` into `assigned_var` is a pure optimization that doesn't affect correctness. It just makes the simulation a bit faster. Therefore it should never emit an error, because it is optional. Instead, if the merging cannot happen for whatever reason, it should just silently not happen 🙂

hailongSun2000: Thanks for the detailed explanation.

lib/Dialect/Moore/Transforms/MergeAssignments.cpp
hailongSun2000: Let me attempt to add this canonicalization to it.

hailongSun2000:
Hey, @fabianschuiki @uenoku!  I add the canonicalization pattern for `moore.variable`. But it seems can not keep the original variable, if not call `replaceOp()` /`replaceOpWithNewOp()`, the `canonicalize()` will not take effect.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -265,6 +265,19 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
...
+  for (auto *user : op->getUsers())
+    if (isa<ContinuousAssignOp>(user) &&
+        (user->getOperand(0) == op.getResult())) {
+      auto assignVarOp = rewriter.create<AssignedVarOp>(
+          op->getLoc(), op.getType(), op.getName(), user->getOperand(1));
+      rewriter.replaceOp(op, assignVarOp);
+      return success();
+    }
+  return failure();
+}
```
uenoku, line 279: I think you must ensure that  there is no multiple `ContinuousAssignOp`.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -265,6 +265,19 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
...
 }
 
+LogicalResult VariableOp::canonicalize(VariableOp op,
+                                       ::mlir::PatternRewriter &rewriter) {
+  for (auto *user : op->getUsers())
+    if (isa<ContinuousAssignOp>(user) &&
+        (user->getOperand(0) == op.getResult())) {
+      auto assignVarOp = rewriter.create<AssignedVarOp>(
+          op->getLoc(), op.getType(), op.getName(), user->getOperand(1));
+      rewriter.replaceOp(op, assignVarOp);
```
uenoku, line 275: nit: you can use replaceOpWithNewOp.

uenoku:
Can we remove mergeAssignments pass if we implemented it as a canonicalization?

lib/Dialect/Moore/MooreOps.cpp
hailongSun2000: Yeah, you're totally right!

lib/Dialect/Moore/MooreOps.cpp
hailongSun2000: Slang has guaranteed that the variable has only one continuous assignment. For example:
```
int x;
assign x = 32;
assign x = 64;
```
```
source.sv:4:9: error: cannot have multiple continuous assignments to variable 'x'
 assign x = 64;
        ^
source.sv:3:9: note: also assigned here
 assign x = 32;
 ```

**Comment by hailongSun2000** [comment]
I have removed the `MergeAssignments` pass.

File: lib/Dialect/Moore/MooreOps.cpp (hunk 6 above)
Comment (uenoku) [reply lib/Dialect/Moore/MooreOps.cpp]: I think we shouldn't rely on implicit assumption of slang. We should at least check that in our verifier.

Comment (hailongSun2000) [reply lib/Dialect/Moore/MooreOps.cpp]: Ok, I'll add this.

Comment (hailongSun2000) [reply lib/Dialect/Moore/MooreOps.cpp]: I understood why we need to check whether there are multiple continuous assignments for the same variable. Because the `--canonicalize` operates on the IR, not from AST to IR, right?

File: lib/Dialect/Moore/MooreOps.cpp [hunk 8]
```diff
@@ -265,6 +265,30 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
...
+    if (isa<ContinuousAssignOp>(user) &&
+        (user->getOperand(0) == op.getResult())) {
+      // Don't canonicalize the multiple continuous assignment to the same
+      // variable.
+      if (variables.contains(op))
+        return failure();
+
+      variables.insert(op);
+      initial = user->getOperand(1);
+    }
```
uenoku, line 282: I think we don't need `variables` set.
```suggestion
  Value initial;
  for (auto *user : op->getUsers())
    if (isa<ContinuousAssignOp>(user) &&
        (user->getOperand(0) == op.getResult())) {
      // Don't canonicalize the multiple continuous assignment to the same
      // variable.
      if (initial)
        return failure();

      initial = user->getOperand(1);
    }
```

uenoku (approved):
One nit, otherwise LGTM!

lib/Dialect/Moore/MooreOps.cpp
uenoku: Yes, I think we should able to use Moore dialect not only for slang.

hailongSun2000:
Thanks for the code review :smiley: .

fabianschuiki:
Very cool! 🥳

PR #7253, opened 2024-07-01
[MooreToCore] Fix parse error for parameter

mingzheTerapines:
img_v3_02cc_cfdc615a-56e2-43ae-ab4e-f56bcd8c03dg
Fix parser error like this. Parser will find type after the ":", while I naming parameter with "@typekind:name". So replace ":" with "_" will solve this problem.


uenoku (commented):
Why are we adding inner symbols to constants in the first place 🤔 ?

mingzheTerapines:
@uenoku , SystemVerilog provides  three elaboration-time constants: parameter, localparam, and specparam. Conversation should not lose these key words, so I added it at the firstplace, which won't duplicate with each other.

fabianschuiki:
If I recall correctly, the idea is to have the MooreToCore lowering preserve the constants as `hw.wire` such that they show up in simulations. It's pretty hacky. Ideally we'd do this with a `dbg.variable`, but there isn't very good support for those in Arcilator and other CIRCT lowerings yet.

fabianschuiki (approved):
Fix LGTM!

PR #7256, opened 2024-07-01
[ESI Runtime] Avoid using CIRCT_* cmake variables

### LINKED ISSUE #7254: [ESI] ESI runtime relies on CIRCT-defined CMake variables
E.g.
https://github.com/llvm/circt/blob/20e08c092c67adf20d6b4114d06e45d5d8dc4965/lib/Dialect/ESI/runtime/cosim/CMakeLists.txt#L21-L24

relies on `CIRCT_TOOLS_DIR`. When ESI Runtime is built out-of-tree, `CIRCT_TOOLS_DIR` is undefined, resulting in an attempted copy of `esi-cosim.py` to `/esi-cosim.py`, which is illegal.

Some possible offending locations:
- https://github.com/llvm/circt/blob/5950674a5f32b31820fc734b0da4aa549a3b9c68/lib/Dialect/ESI/runtime/cosim/MtiPliStub/CMakeLists.txt#L3
- https://github.com/llvm/circt/blob/20e08c092c67adf20d6b4114d06e45d5d8dc4965/lib/Dialect/ESI/runtime/cosim/CMakeLists.txt#L24
- https://github.com/llvm/circt/blob/20e08c092c67adf20d6b4114d06e45d5d8dc4965/lib/Dialect/ESI/runtime/cosim/CMakeLists.txt#L29

teqdruid:
Fixes #7254


teqdruid:
Those are the only two places I could find after the refactor I'm merging into.

teqdruid:
No CI checks test this currently.

PR #7259, opened 2024-07-02
[Verif][circt-bmc] Add `verif.bmc` op and `circt-bmc` tool

TaoBi22:
This PR (based on an initial draft by @maerhart, thanks!) introduces a bounded model checking tool based on the SMT dialect, similar to the updated circt-lec tool.

To achieve this, a BMC op is added to the verif dialect - in order to support multiple clocks in the near future, this has init and loop regions that allow the manipulation of clocks to be specified (e.g. so that specific interleavings can be explored). These regions are allowed some extra state arguments, as is likely necessary for specifying clock interleavings (e.g. to say which clock's turn it is to be toggled next).

Sorry for the big diff!


lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -40,11 +45,74 @@ struct VerifAssertOpConversion : OpConversionPattern<verif::AssertOp> {
...
+  using OpConversionPattern<verif::ClockedAssumeOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(verif::ClockedAssumeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    return op.emitError(
+        "Conversion of ClockedAssumeOps to SMT not yet supported");
+  };
+};
+
```
TaoBi22, line 115: Not totally sure if this is necessary, but my thinking is that without these patterns the errors produced will likely be pretty opaque to someone not familiar with CIRCT/MLIR internals, which maybe isn't ideal for this tool.

darthscsi:
Can this be split up into 
* new op
* pr per pass being changed/added (looks like 2 changes in conversion?)
* tool addition

TaoBi22:
> Can this be split up into
> 
> * new op
> * pr per pass being changed/added (looks like 2 changes in conversion?)
> * tool addition

That should be fine, thanks for the suggestion!

TaoBi22:
New op PR now up at #7263

Issue #7229, opened 2024-06-23
[FIRRTL] When passing Reg to DPI, it will trigger the comb logic cycle error.


sequencer:
I use `WireDefault` to work around, but I think this is a bug which should be fixed by CIRCT:)

PR #7267, opened 2024-07-02
[FIRRTL][HW] Change default implementation of CombDataFlow and add it to DPI intrinsic 

uenoku:
This PR removes default implementation of CombDataFlow which indicate an op was sequential. Also implement it for DPI intrinsic.

Close https://github.com/llvm/circt/issues/7229

### DIFF
--- a/include/circt/Dialect/HW/HWOpInterfaces.td
+++ b/include/circt/Dialect/HW/HWOpInterfaces.td
@@ -561,12 +561,9 @@ def CombDataflow : OpInterface<"CombDataFlow"> {
      This returns a pair of ground type fieldrefs. The first element is the destination
      and the second is the source of the dependence. The default implementation returns
      an empty list, which implies that the operation is not combinational.}],
-     "void", "computeDataFlow",
-     (ins "llvm::function_ref<void(circt::FieldRef, circt::FieldRef)>":$callback)>,
+     "SmallVector<std::pair<circt::FieldRef, circt::FieldRef>>", "computeDataFlow",
+     (ins)>,
   ];
-  let extraSharedClassDeclaration = [{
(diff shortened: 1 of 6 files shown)


darthscsi (commented):
It's an alternate implementation, sure.  It's not obviously faster.  Control flow and indirect calls are quite expensive, possibly more so than dense storage passed by value.

uenoku:
> It's an alternate implementation, sure. It's not obviously faster. Control flow and indirect calls are quite expensive, possibly more so than dense storage passed by value.

I don't have a strong opinion TBH. CombDataFlow interface is opt-in and not heavily used right now (only used by MemoryOp) so it would be hard to do microbenchmark

include/circt/Dialect/HW/HWOpInterfaces.td
```diff
@@ -556,17 +556,17 @@ def CombDataflow : OpInterface<"CombDataFlow"> {
...
-     "the dependence."
-     "The default implementation returns an empty list, which implies that the"
-     "operation is not combinational.",
-     "SmallVector<std::pair<circt::FieldRef, circt::FieldRef>>",
-     "computeDataFlow", (ins), [{}], /*defaultImplementation=*/[{
-      return {};
-    }]>,
+    InterfaceMethod<[{
+     Get the combinational dataflow relations between the operands and the results.
+     This returns a pair of ground type fieldrefs. The first element is the destination
```
prithayan, line 561:
```suggestion
     The callback is invoked with a pair of ground type fieldrefs. The first element is the destination
```
The comment needs to be updated as it no longer returns it, something along the lines of the suggestion.

prithayan (approved):
The updated interface looks good to me.

uenoku:
I reverted a change for using a callback for now. I'll follow up if necessary.

PR #7272, opened 2024-07-02
[FIRRTL] Ensure hierpath considers owning module in LowerClasses.

mikeurbach:
We already had logic to detect if the owning module was in the middle of a hierarchical path, and build up the prefix just to the owning module in this case. However, we still used the entire original hierarchical path in the new path, which caused the prefix to the owning module to be duplicated. This is fixed by not only setting the moduleName to the owning module, but also trimming the prefix to the owning module in the original path.


mikeurbach:
I'm gonna go ahead and merge this, but happy to incorporate any post-commit review. It's unfortunately not something you can catch until you actually run the evaluator and the basepaths are concatenated with the hierarchical paths. But I've tested this on several downstream designs and it seems to be behaving correctly now. I also think the risk of this accidentally regressing something and creating invalid hierpath ops is low--in the case that changed, we were already not prefixing anything to the path, the difference is we now also snip off the extra prefix.

PR #7273, opened 2024-07-02
[FIRRTL] Fix use-after-free in InferReset

### LINKED ISSUE #7225: [FIRRTL] InferResets: deleting op used to drive reset network results in use-after-free
```mlir
firrtl.circuit "MovableNodeShouldDominate" {
  firrtl.module @MovableNodeShouldDominate(in %clock: !firrtl.clock) {
    %child_clock = firrtl.instance child @Child(in clock: !firrtl.clock)
    firrtl.connect %child_clock, %clock : !firrtl.clock
    %ui1 = firrtl.constant 1 : !firrtl.uint<1>
    %0 = firrtl.asAsyncReset %ui1 : (!firrtl.uint<1>) -> !firrtl.asyncreset
    %localReset = firrtl.node %0 {annotations = [{class = "sifive.enterprise.firrtl.FullAsyncResetAnnotation"}]} : !firrtl.asyncreset
  }
  firrtl.module @Child(in %clock: !firrtl.clock) {
    %reg = firrtl.reg %clock : !firrtl.clock, !firrtl.uint<8>
  }
}
```
```
./bin/circt-opt --pass-pipeline='builtin.module(firrtl.circuit(firrtl-infer-resets))' reset.mlir
```

InferResets needs to wire the result of the NodeOp to the instance of Child, and to do so [replaces it with a wire](https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/Transforms/InferResets.cpp#L1724).  This results in a dangling reference to the deleted node stored in the ResetDomain.

Running it under valgrind yields many errors, such as the one below, but my line numbers won't line up with `main` due to a lot of printf debugging:
```
==176563== Invalid read of size 8
==176563==    at 0x1C35A53: mlir::Value::getLoc() const (llvm/mlir/lib/IR/Value.cpp:0)
==176563==    by 0x12A6F98: implementAsyncReset (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:1688)
==176563==    by 0x12A6F98: implementAsyncReset (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:1647)
==176563==    by 0x12A6F98: (anonymous namespace)::InferResetsPass::runOnOperationInner() (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:545)
==176563==    by 0x12A4465: (anonymous namespace)::InferResetsPass::runOnOperation() (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:513)

==176563==  Address 0x613f688 is 40 bytes inside a block of size 128 free'd
==176563==    at 0x4C39A93: free (vg_replace_malloc.c:872)
==176563==    by 0x18E466D: llvm::iplist_impl<llvm::simple_ilist<mlir::Operation>, llvm::ilist_traits<mlir::Operation> >::erase(llvm::ilist_iterator<llvm::ilist_detail::node_options<mlir::Operation, true, false, void, false>, false, false>) (ilist.h:205)
==176563==    by 0x12A7755: erase (OpDefinition.h:131)
==176563==    by 0x12A7755: implementAsyncReset (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:1742)
==176563==    by 0x12A7755: implementAsyncReset (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:1647)
==176563==    by 0x12A7755: (anonymous namespace)::InferResetsPass::runOnOperationInner() (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:545)
==176563==    by 0x12A4465: (anonymous namespace)::InferResetsPass::runOnOperation() (lib/Dialect/FIRRTL/Transforms/InferResets.cpp:513)
```

darthscsi:
As @youngar explains well, a node was being deleted when it had references.  Just route the node through the bounce wire instead of trying to replace it.

Closes #7225

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
@@ -1723,6 +1723,11 @@ LogicalResult InferResetsPass::implementAsyncReset(FModuleOp module,
         // Don't delete the node, since it might be in use in worklists.
         nodeOp->replaceAllUsesWith(wireOp);
         nodeOp->removeAttr(nodeOp.getInnerSymAttrName());
+        nodeOp.setName("");
+        // Leave forcable alone, since we cannot remove a result.  It will be
+        // cleaned up in canonicalization since it is dead.  As will this node.
+        nodeOp.setNameKind(NameKindEnum::DroppableName);
+        nodeOp.setAnnotationsAttr(ArrayAttr::get(builder.getContext(), {}));
         builder.setInsertionPointAfter(nodeOp);
         emitConnect(builder, wireOp.getResult(), nodeOp.getResult());
         resetOp = wireOp;
--- a/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
@@ -1714,14 +1714,17 @@ LogicalResult InferResetsPass::implementAsyncReset(FModuleOp module,
       if (nodeOp && !dom.dominates(nodeOp.getInput(), opsToUpdate[0])) {
         LLVM_DEBUG(llvm::dbgs()
                    << "- Promoting node to wire for move: " << nodeOp << "\n");
-        ImplicitLocOpBuilder builder(nodeOp.getLoc(), nodeOp);
+        auto builder = ImplicitLocOpBuilder::atBlockBegin(nodeOp.getLoc(),
+                                                          nodeOp->getBlock());
         auto wireOp = builder.create<WireOp>(
             nodeOp.getResult().getType(), nodeOp.getNameAttr(),
             nodeOp.getNameKindAttr(), nodeOp.getAnnotationsAttr(),
             nodeOp.getInnerSymAttr(), nodeOp.getForceableAttr());
-        emitConnect(builder, wireOp.getResult(), nodeOp.getInput());
+        // Don't delete the node, since it might be in use in worklists.
         nodeOp->replaceAllUsesWith(wireOp);
-        nodeOp.erase();
+        nodeOp->removeAttr(nodeOp.getInnerSymAttrName());
+        builder.setInsertionPointAfter(nodeOp);
+        emitConnect(builder, wireOp.getResult(), nodeOp.getResult());
         resetOp = wireOp;
         actualReset = wireOp.getResult();
         domain.existingValue = wireOp.getResult();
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -1714,14 +1714,17 @@ LogicalResult InferResetsPass::implementAsyncReset(FModuleOp module,
...
+                                                          nodeOp->getBlock());
         auto wireOp = builder.create<WireOp>(
             nodeOp.getResult().getType(), nodeOp.getNameAttr(),
             nodeOp.getNameKindAttr(), nodeOp.getAnnotationsAttr(),
             nodeOp.getInnerSymAttr(), nodeOp.getForceableAttr());
-        emitConnect(builder, wireOp.getResult(), nodeOp.getInput());
+        // Don't delete the node, since it might be in use in worklists.
         nodeOp->replaceAllUsesWith(wireOp);
-        nodeOp.erase();
+        nodeOp->removeAttr(nodeOp.getInnerSymAttrName());
```
youngar, line 1725: Should probably remove the annotations as well, e.g. if the node has an OMIR path tracker we will end up having two things with the same path tracker.

Not as important but might want to make the Node non-forcible.

darthscsi: yea working on it.

dtzSiFive:
Thanks for getting us past this.

Sounds like some bits are still coming, but please add a test (such as in the linked issue).

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
darthscsi: Changing the node to be non-forcable is kind of a pain as it would requiring changing the number of results, which I don't think you can do.  The forcable result should be unused, so it should be cleared away with the rest of the node by CSE/Canonicalization.

dtzSiFive: Yes you can't change results, the way it's done is a bit rough but basically creates a new one and replaces it.

And even if you did that (there's a helper, but in the `detail` namespace) it'd be erasing the existing node so no-go.

Thanks!!

dtzSiFive (approved):
LGTM, we can fix the rest of InferResets later.

dtzSiFive:
Okay to land? :grin:

PR #7278, opened 2024-07-03
[Moore] SymbolVisibility attribute support for SVModuleOp

cepheus69:
We can append the appropriate symbol visibility to svmoduleOp according to the structures of instanceSymbol provided by the slang frontend. slang provides function to get the root of the design. Calling this method could get all top-level instanceSymbols and help determine which ones should be tagged. Note that the visibility attribute now used does not contain the `nested`.


cepheus69:
> Can we define visibility in ODS like https://github.com/llvm/llvm-project/blob/56f0ecd6db9219b7d14a8eda613d6b75060643eb/mlir/include/mlir/Dialect/Func/IR/FuncOps.td#L271? I know many CIRCT ops are not following this convention but probably it's better to use this style. We should also fix other CIRCT ops.

Well, it is a good choice to follow the upstream style. I'll take a look.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -25,19 +25,38 @@ using namespace mlir;
...
                                                 getModuleType(), {}, {});
-  p.printOptionalAttrDictWithKeyword(getOperation()->getAttrs(),
-                                     getAttributeNames());
   p << " ";
   p.printRegion(getBodyRegion(), /*printEntryBlockArgs=*/false,
                 /*printBlockTerminators=*/true);
+
+  SmallVector<StringRef, 4> omittedAttrs = {"sym_name", "module_type",
+                                            visibilityAttrName};
+  p.printOptionalAttrDictWithKeyword((*this)->getAttrs(), omittedAttrs);
```
uenoku, line 53: I think this change is not required anymore? Also this changes the position of attributes and is that intentional?

cepheus69: You are right. It is no need to preserve it.

cepheus69: I will delete it soon. Thanks so much.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -25,19 +25,34 @@ using namespace mlir;
...
+  // Print the visibility of the module.
+  StringRef visibilityAttrName = SymbolTable::getVisibilityAttrName();
+  if (auto visibility = (*this)->getAttrOfType<StringAttr>(visibilityAttrName))
+    p << visibility.getValue() << ' ';
+
   p.printSymbolName(SymbolTable::getSymbolName(*this).getValue());
   hw::module_like_impl::printModuleSignatureNew(p, getBodyRegion(),
                                                 getModuleType(), {}, {});
-  p.printOptionalAttrDictWithKeyword(getOperation()->getAttrs(),
-                                     getAttributeNames());
```
uenoku, line 34: We should keep this.

uenoku (approved):
Please revert the change regarding `printOptionalAttrDictWithKeyword` otherwise LGTM

cepheus69:
Thank you for the review!

lib/Dialect/Moore/MooreOps.cpp
cepheus69: I think I misunderstood your previous comments. It was already been reverted and applied the new way.

PR #7282, opened 2024-07-04
[ESI] Add read-side MMIO back

teqdruid:
Use ESI channels/bundles to implement MMIO reads. Replaces the low-level AXI interface. Keep the XRT AXI-based interface as-is for now.


teqdruid:
All the code added to xrt.py was copied from common.py. I did this to maintain XRT support as-is. I'll update it later to use ChannelMMIO.

mortbopet (approved):

PR #7284, opened 2024-07-05
[ARC] Keep just one result as a parameter if the result is given multiple times

elhewaty:
This pull request adds a new canonical form, if we have the following IR

```
%1:4 = arc.vectorize(%0#0, %0#1, %0#2, %0#3), (%a, %b, %c, %d), (%0#0, %0#1, %0#2, %0#3)
```﻿
we can remove the repeated input
```
%1:4 = arc.vectorize(%0#0, %0#1, %0#2, %0#3), (%a, %b, %c, %d)
```

This can reduce the time of the unpacking of vector elements.

### CODE REVIEW

**Comment by elhewaty** [comment]
This can be generalized to remove any repeated pattern in the input, what do you think?

File: lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp [hunk 1]
```diff
@@ -657,6 +663,46 @@ MergeVectorizeOps::matchAndRewrite(VectorizeOp vecOp,
   return success();
 }
 
+LogicalResult KeepOneVecOp::matchAndRewrite(VectorizeOp vecOp,
+                                            PatternRewriter &rewriter) const {
+  BitVector argsToRemove(vecOp.getInputs().size(), false);
+  DenseMap<VectorizeOp, SmallVector<unsigned>> opExist;
+  for (auto [argIdx, inputVec] : llvm::enumerate(vecOp.getInputs())) {
+    auto otherVecOp = inputVec[0].getDefiningOp<VectorizeOp>();
```
maerhart, line 671: For this canonicalization pattern, it doesn't matter where the values come from, so we can simplify (and at the same time generalize) this a bit.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -657,6 +663,46 @@ MergeVectorizeOps::matchAndRewrite(VectorizeOp vecOp,
...
+  for (auto [argIdx, inputVec] : llvm::enumerate(vecOp.getInputs()))
+    if (!argsToRemove[argIdx])
+      newOperands.insert(newOperands.end(), inputVec.begin(), inputVec.end());
+
+  // Set the new inputOperandSegments value
+  unsigned groupSize = vecOp.getResults().size();
+  unsigned numOfGroups = newOperands.size() / groupSize;
+  SmallVector<int32_t> newAttr(numOfGroups, groupSize);
+  vecOp.setInputOperandSegments(newAttr);
+  vecOp.getOperation()->setOperands(ValueRange(newOperands));
```
maerhart, line 701: Nit: since this is the same as in the other pattern, we could factor it out into a helper function.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -657,6 +663,46 @@ MergeVectorizeOps::matchAndRewrite(VectorizeOp vecOp,
...
+  for (auto [op, positions] : opExist)
+    for (unsigned idx = 1; idx < positions.size(); ++idx)
+      rewriter.replaceAllUsesWith(currentBlock.getArgument(positions[idx]),
+                                  currentBlock.getArgument(positions[0]));
+
+  currentBlock.eraseArguments(argsToRemove);
+  SmallVector<Value> newOperands;
+  for (auto [argIdx, inputVec] : llvm::enumerate(vecOp.getInputs()))
+    if (!argsToRemove[argIdx])
+      newOperands.insert(newOperands.end(), inputVec.begin(), inputVec.end());
```
maerhart, line 694: Nit: we could already build this new operand list during the first iteration over the inputs (instead of having `argsToRemove`)

test/Dialect/Arc/arc-canonicalizer.mlir
maerhart, line 1: Can we add a test where multiple duplicates are removed that are not the result of a vectorize op?

maerhart (commented):
Yes, I think this is a useful canonicalization regardless of how much you can save in vector packing. 👍 
It can reduce the use-count of values and thus enable more applications of your other canonicalization pattern, for example.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -285,6 +291,17 @@ LogicalResult canonicalizePassthoughCall(mlir::CallOpInterface callOp,
   return failure();
 }
 
+LogicalResult vecOpCanonicalizersHelper(VectorizeOp &vecOp,
+                                        const SmallVector<Value> &newOperands) {
```
fabianschuiki, line 295: How about calling this `updateInputOperands`?

fabianschuiki (approved):
This is a very useful canonicalization! LGTM, besides a minor comment about `SmallVector` maybe not being necessary.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -657,6 +669,63 @@ MergeVectorizeOps::matchAndRewrite(VectorizeOp vecOp,
...
+                      const SmallVector<Value> &rhs) {
+    return lhs == rhs;
+  }
+};
+} // namespace llvm
+
+LogicalResult KeepOneVecOp::matchAndRewrite(VectorizeOp vecOp,
+                                            PatternRewriter &rewriter) const {
+  BitVector argsToRemove(vecOp.getInputs().size(), false);
+  DenseMap<SmallVector<Value>, unsigned> inExists;
```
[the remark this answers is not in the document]
elhewaty: @fabianschuiki, I tried but `OperandRange` default constructor is deleted. 
how can I return an empty OpernadRange?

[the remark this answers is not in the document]
fabianschuiki: Ah hmmm, good point. That's annoying. My hope was that `OperandRange` would already have the DenseMapInfo stuff defined, but it probably doesn't. Feel free to stick with the `SmallVector<Value>` if it's not easy to switch 😃

fabianschuiki (approved):
LGTM!

PR #7285, opened 2024-07-05
[circt-bmc] Add ExternaliseRegisters Pass

TaoBi22:
This PR adds a pass (pass itself written by @maerhart, I'm just PRing as part of the splitting of #7259) that extracts `seq.compreg`s within modules, and replaces them with additional inputs and outputs such that register states are fed as module outputs, then fed back in as inputs. This simplifies model checking as the state can be provided directly as an input to a function checking the design, rather than having to walk the IR to calculate the state.

### DIFF
--- a/include/circt/Tools/CMakeLists.txt
+++ b/include/circt/Tools/CMakeLists.txt
@@ -1,2 +1,2 @@
-add_subdirectory(circt-lec)
 add_subdirectory(circt-bmc)
+add_subdirectory(circt-lec)
--- a/include/circt/Tools/CMakeLists.txt
+++ b/include/circt/Tools/CMakeLists.txt
@@ -1,1 +1,2 @@
 add_subdirectory(circt-lec)
+add_subdirectory(circt-bmc)
--- a/include/circt/Tools/circt-bmc/Passes.td
+++ b/include/circt/Tools/circt-bmc/Passes.td
@@ -20,8 +20,7 @@ def ExternalizeRegisters : Pass<"externalize-registers", "::mlir::ModuleOp"> {
   let summary = "Removes registers and adds corresponding input and output ports";
 
   let dependentDialects = [
-    "mlir::func::FuncDialect", "mlir::LLVM::LLVMDialect",
-    "circt::verif::VerifDialect"
+    "circt::seq::SeqDialect"
   ];
 }
 
--- a/include/circt/Tools/circt-bmc/Passes.td
+++ b/include/circt/Tools/circt-bmc/Passes.td
@@ -1,0 +1,29 @@
+//===-- Passes.td - BMC Transforms pass definition file ----*- tablegen -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file contains definitions for BMC transformation passes.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_TOOLS_CIRCT_BMC_PASSES_TD
+#define CIRCT_TOOLS_CIRCT_BMC_PASSES_TD
+
+include "mlir/Pass/PassBase.td"
+include "mlir/Rewrite/PassUtil.td"
+
+def ExternalizeRegisters : Pass<"externalize-registers", "::mlir::ModuleOp"> {
+  let summary = "Removes registers and adds corresponding input and output ports";
+
+  let dependentDialects = [
+    "mlir::func::FuncDialect", "mlir::LLVM::LLVMDialect",
+    "circt::verif::VerifDialect"
+  ];
+}
+
+#endif // CIRCT_TOOLS_CIRCT_BMC_PASSES_TD
+
--- a/include/circt/Tools/circt-bmc/Passes.td
+++ b/include/circt/Tools/circt-bmc/Passes.td
@@ -20,7 +20,7 @@ def ExternalizeRegisters : Pass<"externalize-registers", "::mlir::ModuleOp"> {
   let summary = "Removes registers and adds corresponding input and output ports";
 
   let dependentDialects = [
-    "circt::seq::SeqDialect"
+    "circt::seq::SeqDialect", "circt::hw::HWDialect"
   ];
 }
 
--- a/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
+++ b/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
@@ -138,6 +138,7 @@ void ExternalizeRegistersPass::runOnOperation() {
           numRegs += newInputs.size();
           instanceOp.replaceAllUsesWith(
               newInst.getResults().take_front(instanceOp->getNumResults()));
+          instanceGraph.replaceInstance(instanceOp, newInst);
           instanceOp->erase();
           return;
         }
--- a/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
+++ b/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
@@ -65,17 +65,23 @@ void ExternalizeRegistersPass::runOnOperation() {
       if (!module)
         continue;
 
+      mlir::Value firstClock;
       unsigned numRegs = 0;
       module->walk([&](Operation *op) {
         if (auto regOp = dyn_cast<seq::CompRegOp>(op)) {
           addedInputs[module.getSymNameAttr()].push_back(regOp.getType());
           addedOutputs[module.getSymNameAttr()].push_back(
               regOp.getInput().getType());
-
           OpBuilder builder(regOp);
           regOp.getResult().replaceAllUsesWith(
               module.appendInput("", regOp.getType()).second);
           module.appendOutput("", regOp.getInput());
+          if (numRegs == 0)
+            firstClock = regOp.getClk();
+          else if (firstClock != regOp.getClk())
+            regOp->emitWarning(
+                "multiple clocks not yet supported - all registers will be "
+                "assumed to be clocked together");
           regOp->erase();
           ++numRegs;
           return;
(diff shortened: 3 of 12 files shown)


lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -0,0 +1,119 @@
...
+          OpBuilder builder(regOp);
+          regOp.getResult().replaceAllUsesWith(
+              module.appendInput("", regOp.getType()).second);
+          module.appendOutput("", regOp.getInput());
+          if (numRegs == 0)
+            firstClock = regOp.getClk();
+          else if (firstClock != regOp.getClk())
+            regOp->emitWarning(
+                "multiple clocks not yet supported - all registers will be "
+                "assumed to be clocked together");
```
uenoku, line 84: Can we cause signal pass failure instead of just a warning?

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -0,0 +1,119 @@
...
+      unsigned numRegs = 0;
+      module->walk([&](Operation *op) {
+        if (auto regOp = dyn_cast<seq::CompRegOp>(op)) {
+          addedInputs[module.getSymNameAttr()].push_back(regOp.getType());
+          addedOutputs[module.getSymNameAttr()].push_back(
+              regOp.getInput().getType());
+          OpBuilder builder(regOp);
+          regOp.getResult().replaceAllUsesWith(
+              module.appendInput("", regOp.getType()).second);
+          module.appendOutput("", regOp.getInput());
```
uenoku, line 78: CompRegOp could have reset or preset as well so I think we should reject if it has reset or preset for now.

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -0,0 +1,119 @@
...
+      mlir::Value firstClock;
+      unsigned numRegs = 0;
+      module->walk([&](Operation *op) {
+        if (auto regOp = dyn_cast<seq::CompRegOp>(op)) {
+          addedInputs[module.getSymNameAttr()].push_back(regOp.getType());
+          addedOutputs[module.getSymNameAttr()].push_back(
+              regOp.getInput().getType());
+          OpBuilder builder(regOp);
+          regOp.getResult().replaceAllUsesWith(
+              module.appendInput("", regOp.getType()).second);
```
uenoku, line 77: nit: could you add port names?

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -0,0 +1,119 @@
...
+          SmallVector<Type> resTypes(instanceOp->getResultTypes());
+          resTypes.append(newOutputs);
+          auto newInst = builder.create<InstanceOp>(
+              instanceOp.getLoc(), resTypes, instanceOp.getInstanceNameAttr(),
+              instanceOp.getModuleNameAttr(), instanceOp.getInputs(),
+              instanceOp.getArgNamesAttr(), instanceOp.getResultNamesAttr(),
+              instanceOp.getParametersAttr(), instanceOp.getInnerSymAttr());
+          for (auto output : newInst->getResults().take_back(newOutputs.size()))
+            module.appendOutput("", output);
+          instanceOp.erase();
```
uenoku, line 109: I think we may need to update instance graph when replacing InstanceOp.

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -0,0 +1,119 @@
...
+          ++numRegs;
+          return;
+        }
+        if (auto instanceOp = dyn_cast<InstanceOp>(op)) {
+          auto newInputs =
+              addedInputs[instanceOp.getModuleNameAttr().getAttr()];
+          auto newOutputs =
+              addedOutputs[instanceOp.getModuleNameAttr().getAttr()];
+          addedInputs[module.getSymNameAttr()].append(newInputs);
+          addedOutputs[module.getSymNameAttr()].append(newOutputs);
```
uenoku, line 95: Do we need to update `numRegs` as well here?

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -0,0 +1,119 @@
...
+using namespace hw;
+using namespace igraph;
+
+namespace circt {
+#define GEN_PASS_DEF_EXTERNALIZEREGISTERS
+#include "circt/Tools/circt-bmc/Passes.h.inc"
+} // namespace circt
+
+//===----------------------------------------------------------------------===//
+// Convert Lower To BMC pass
```
uenoku, line 33: nit: Probably old comment.

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
TaoBi22: Sure, I'll change it over

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
TaoBi22: Good point, thanks!

TaoBi22: Actually, looks like CompRegs don't have preset values, just FirRegs. I'll add an error on reset value though,

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
uenoku: It would be tricky to compute instance paths to each register so it's not necessary to do that within this PR but I believe we'll eventually want to see correspondence between ports and original registers for debuggability :smile:

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
uenoku: Ah CompReg has `powerOnValue` as a preset value 🙇

TaoBi22:
Just realised there are no tests for nested modules so I'll add those on too

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
TaoBi22: Ah of course! Error and test added for powerOnValues :)

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
TaoBi22: The names of the inputs & outputs are now based on the original names of the registers if present (suffixed with `_state` or `_input` depending on whether they represent the input to the register or its current state), otherwise it's just `reg_<n>_state`. Agreed instance paths would be nice in the future for debuggability!

TaoBi22:
@uenoku sorry for the delay and thanks for the review - I think that should be all comments addressed now!

include/circt/Tools/CMakeLists.txt
```diff
@@ -1 +1,2 @@
 add_subdirectory(circt-lec)
+add_subdirectory(circt-bmc)
```
uenoku, line 2: super nit: please alpha-sort.
```suggestion
add_subdirectory(circt-bmc)
add_subdirectory(circt-lec)
```

include/circt/Tools/circt-bmc/Passes.td
```diff
@@ -0,0 +1,28 @@
...
+#define CIRCT_TOOLS_CIRCT_BMC_PASSES_TD
+
+include "mlir/Pass/PassBase.td"
+include "mlir/Rewrite/PassUtil.td"
+
+def ExternalizeRegisters : Pass<"externalize-registers", "::mlir::ModuleOp"> {
+  let summary = "Removes registers and adds corresponding input and output ports";
+
+  let dependentDialects = [
+    "circt::seq::SeqDialect"
```
uenoku, line 23: I think we also need HW as instnace op is created.

test/Tools/circt-bmc/externalize-registers.mlir
```diff
@@ -0,0 +1,61 @@
...
+// CHECK:  hw.module @nested_reg(in [[CLK:%.+]] : !seq.clock, in [[IN0:%.+]] : i32, in [[IN1:%.+]] : i32, in [[OLD_REG:%.+]] : i32, out {{.+}} : i32, out {{.+}} : i32) attributes {num_regs = 1 : i32} {
+// CHECK:    [[INSTOUT:%.+]], [[INSTREG:%.+]] = hw.instance "one_reg" @one_reg(clk: [[CLK]]: !seq.clock, in0: [[IN0]]: i32, in1: [[IN1]]: i32, {{.+}}: [[OLD_REG]]: i32) -> ({{.+}}: i32, {{.+}}: i32)
+// CHECK:    hw.output [[INSTOUT]], [[INSTREG]]
+// CHECK:  }
+hw.module @nested_reg(in %clk: !seq.clock, in %in0: i32, in %in1: i32, out out: i32) {
+  %0 = hw.instance "one_reg" @one_reg(clk: %clk: !seq.clock, in0: %in0: i32, in1: %in1: i32) ->  (out: i32)
+  hw.output %0 : i32
+}
+
+// CHECK:  hw.module @nested_nested_reg(in [[CLK:%.+]] : !seq.clock, in [[IN0:%.+]] : i32, in [[IN1:%.+]] : i32, in [[OLD_REG0:%.+]] : i32, in [[OLD_REG1:%.+]] : i32, out {{.+}} : i32, out {{.+}} : i32, out {{.+}} : i32) attributes {num_regs = 2 : i32} {
```
uenoku, line 53: Can we test names of ports? It's not necessary to test all ports but nice to check nested registers are named properly

uenoku (approved):
LGTM, nice work!

include/circt/Tools/circt-bmc/Passes.td
TaoBi22: Good point, thanks!

test/Tools/circt-bmc/externalize-registers.mlir
TaoBi22: Sure, I'll add a check on the regs!

PR #7287, opened 2024-07-08
[ESI] Add verify connections pass

### LINKED ISSUE #7286: [ESI] Add value validation to Channels and Bundles
ESI channel and bundle types have the restriction that they must only have at most one consumer and exactly one consumer, respectively. The wrap/pack ops' verifiers check this, but other things which produce them (e.g. modules) are not checked, leading to relatively obtuse errors.

Add something which checks this invariant early on.

teqdruid:
New pass to verify that channels and bundles are used the appropriate number of times. The wrap/unwrap, pack/unpack operations already have these checks, but when a channel/bundle is produced by a module (or some other op), it is not checked. A pass is the best way I could figure out how to implement the check.

Closes #7286.

### DIFF
--- a/include/circt/Dialect/ESI/ESIPasses.h
+++ b/include/circt/Dialect/ESI/ESIPasses.h
@@ -32,7 +32,7 @@ struct Platform {
 #define GEN_PASS_DECL
 #include "circt/Dialect/ESI/ESIPasses.h.inc"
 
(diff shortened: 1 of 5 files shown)


teqdruid:
This check is rather expensive. I'm open to other ideas.

Also, I was unable to specify that it can be run on HWModuleLike ops independently. Doesn't look like on can use an OpInterface in ODS that way.

teqdruid:
Verbal approval from Morten

PR #7288, opened 2024-07-08
CODEOWNERS: remove @dtzSiFive from most paths.

dtzSiFive:


teqdruid (approved):
So you want off of CI changes, huh?

dtzSiFive:
> So you want off of CI changes, huh?

Hahaha 😆 .

I haven't helped much there lately so it seems accurate but still care quite a lot about that and the project so don't worry 👍 . 

Quite open to assisting with anything on that front.  Thank you for your work maintaining that, if there are problems and there's a lack of bandwidth please cc me and I'm happy to help.

Issue #7266, opened 2024-07-02
[Seq] invalid register opt leads to comb loop


youngar:
h/t @dtzSiFive 
The first example from [3003](https://github.com/llvm/circt/issues/3003) results in a cycle that we spin infinitely on in HW,
```firrtl
FIRRTL version 4.0.0
circuit Test:
  public module Test:
    input clock : Clock
    input reset : UInt<1>
    output o : UInt<8>
    wire w : UInt<8>
    reg r : UInt<8>, clock with :
      reset => (reset, mux(reset, UInt<8>(3), w))
    w <= r
    o <= r
```
Before canonicalize in HW this is now:

```mlir
hw.module @Test(in %clock : !seq.clock, in %reset : i1, out o : i8) {
  %c3_i8 = hw.constant 3 : i8
  %w = hw.wire %r  : i8
  %0 = comb.mux bin %reset, %c3_i8, %w : i8
  %r = seq.firreg %r clock %clock reset sync %reset, %0 {firrtl.random_init_start = 0 : ui64} : i8
  hw.output %r : i8
}
```
Which canonicalize quickly reduces the register to a combinational loop
```mlir
hw.module @Test(in %clock : !seq.clock, in %reset : i1, out o : i8) {
  %c3_i8 = hw.constant 3 : i8
  %0 = comb.mux bin %reset, %c3_i8, %0 : i8
  hw.output %0 : i8
}
```

The register should not be optimized in this way.  We had a similar issue in the folder which was fixed (h/t @uenoku) https://github.com/llvm/circt/pull/7085.

The issue appears to be in this code:
https://github.com/llvm/circt/blob/cc51aa5dd03ca451e0329c4a5552565b106c0c0f/lib/Dialect/Seq/SeqOps.cpp#L571-L582

PR #7289, opened 2024-07-09
[Seq] Remove incorrect canonicalization

uenoku:
Fix https://github.com/llvm/circt/issues/7266

### DIFF
--- a/lib/Dialect/Seq/SeqOps.cpp
+++ b/lib/Dialect/Seq/SeqOps.cpp
@@ -547,6 +547,11 @@ void FirRegOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
 std::optional<size_t> FirRegOp::getTargetResultIndex() { return 0; }
 
 LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
+  // Don't canonicalize if there is a preset value for now.
+  // TODO: Handle a preset value.
+  if (op.getPresetAttr())
+    return failure();
+
   // If the register has a constant zero reset, drop the reset and reset value
   // altogether.
   if (auto reset = op.getReset()) {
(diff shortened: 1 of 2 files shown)


lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -577,9 +577,8 @@ LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
   };
 
   if (isConstant()) {
```
dtzSiFive, line 579: `if (isConstant() && !op.getResetValue()) {`
?

test/Dialect/Seq/canonicalization.mlir
```diff
@@ -42,7 +42,7 @@ hw.module @FirRegReset(in %clk : !seq.clock, in %in : i32, in %r : i1, in %v : i
   // value.
```
dtzSiFive, line 42: Please update the comment / test.

Just curious, does this have an impact on our designs in practice?

dtzSiFive (approved):
Thanks for fixing this.  I think LGTM but I think there's more work to do here.

Probably can just simplify this logic now (see comment)?

---

Not for this PR, but when looking at this myself earlier:

1) What role should preset values have re:folders/canonicalization? Not really asking for an answer but it does seem our folders/canonicalizers should be updated to handle that.
2) Folder and canonicalizer are not aligned.  A firreg with "trivial feedback" with a constant zero reset and a constant reset value-- canonicalizer sees constant zero reset and drops the reset and reset value, folder will replace it with a constant zero.  If folder goes first it replaces the register with the constant reset value.

Here's some IR for (2) that acts as described (point `circt-opt -canonicalize` at it, you get `123`, disable that part of the folder and it'll become zero!).

```mlir
hw.module.extern @Observe(in %x : i32)

hw.module @FirRegReset(in %clk : !seq.clock, in %r : i1) {
  %false = hw.constant false
  %constant = hw.constant 123 : i32

  %reg = seq.firreg %reg clock %clk reset sync %false, %constant : i32
  hw.instance "sink" @Observe(x: %reg: i32) -> ()
}
```

It would be good for us to sit down and revisit the justification and correctness for these.

lib/Dialect/Seq/SeqOps.cpp
uenoku: Yes, thanks!

uenoku:
> 1. What role should preset values have re:folders/canonicalization? Not really asking for an answer but it does seem our folders/canonicalizers should be updated to handle that.

Yeah good point. I agree we need to update canonicalizations. For now I added a condition to stop fold/canonicalization for a register with a preset value. Let's revisit later.

test/Dialect/Seq/canonicalization.mlir
uenoku: There was no change in large design 🤔

dtzSiFive:
Thanks, @uenoku !! 🚀

PR #7291, opened 2024-07-09
[NFC][HW] Fix parsing of nullary `hw.triggered` ops

fzi-hielscher:
Make the variadic  `inputs` argument of the `hw.triggered` operation optional in the declarative assembly format.  Without this change the generated parser fails on nullary instances due to the empty/missing type list.


fzi-hielscher:
I recon this is a pretty non-controversial change. As it is currently blocking another PR, I hope you don't mind me merging this without approval.

teqdruid (approved):
Lgtm

PR #7301, opened 2024-07-10
Lower MemRef GetGlobal and write data to json files

jiahanxie353:
This patch lowers `memref::GetGlobalOp` and write data that it carries to corresponding json files, so that they can be used later by Calyx. Json data will be written to the same directory as the source file. Since there might be situations where users don't wish to do so, writing to json will only be carried out if users specify `--lower-scf-to-calyx="write-json=true"`.

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -520,7 +520,8 @@ private:
   /// A mapping between the callee and the instance.
   llvm::StringMap<calyx::InstanceOp> instanceMap;
 
-  /// A json file to store external global memory data
+  /// A json file to store external global memory data. See
+  /// https://docs.calyxir.org/lang/data-format.html?highlight=json#the-data-format
   llvm::json::Value extMemData;
 };
 
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -33,6 +33,8 @@
 namespace circt {
 namespace calyx {
 
+using json = nlohmann::ordered_json;
+
 void appendPortsForExternalMemref(PatternRewriter &rewriter, StringRef memName,
                                   Value memref, unsigned memoryID,
                                   SmallVectorImpl<calyx::PortInfo> &inPorts,
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -450,6 +450,20 @@ public:
     return builder.create<TLibraryOp>(loc, getUniqueName(name), resTypes);
   }
 
+  json &getExtMemData() { return extMemData; }
+
+  const json &getExtMemData() const { return extMemData; }
+
+  void setDataField(StringRef name, const json::array_t &data) {
+    extMemData[name]["data"] = data;
+  }
+
+  void setFormat(StringRef name, std::string numType, bool isSigned,
+                 unsigned width) {
+    extMemData[name]["format"] = {
+        {"numeric_type", numType}, {"is_signed", true}, {"width", width}};
+  }
+
 private:
   /// The component which this lowering state is associated to.
   calyx::ComponentOp component;
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -27,13 +27,12 @@
 #include "mlir/IR/PatternMatch.h"
 #include "llvm/ADT/SmallPtrSet.h"
 #include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/JSON.h"
 
 #include <variant>
 
 namespace circt {
 namespace calyx {
-
-using json = nlohmann::ordered_json;
 
 void appendPortsForExternalMemref(PatternRewriter &rewriter, StringRef memName,
                                   Value memref, unsigned memoryID,
@@ -452,18 +451,36 @@ public:
     return builder.create<TLibraryOp>(loc, getUniqueName(name), resTypes);
   }
 
-  json &getExtMemData() { return extMemData; }
+  llvm::json::Value &getExtMemData() { return extMemData; }
 
-  const json &getExtMemData() const { return extMemData; }
+  const llvm::json::Value &getExtMemData() const { return extMemData; }
 
-  void setDataField(StringRef name, const json::array_t &data) {
-    extMemData[name]["data"] = data;
+  void setDataField(StringRef name, llvm::json::Array data) {
+    auto *extMemDataObj = extMemData.getAsObject();
+    assert(extMemDataObj && "extMemData should be an object");
+
+    auto &value = (*extMemDataObj)[name.str()];
+    llvm::json::Object *obj = value.getAsObject();
+    if (!obj) {
+      value = llvm::json::Object{};
+      obj = value.getAsObject();
+    }
+    (*obj)["data"] = llvm::json::Value(std::move(data));
   }
 
   void setFormat(StringRef name, std::string numType, bool isSigned,
                  unsigned width) {
-    extMemData[name]["format"] = {
-        {"numeric_type", numType}, {"is_signed", true}, {"width", width}};
+    auto *extMemDataObj = extMemData.getAsObject();
+    assert(extMemDataObj && "extMemData should be an object");
+
+    auto &value = (*extMemDataObj)[name.str()];
+    llvm::json::Object *obj = value.getAsObject();
+    if (!obj) {
+      value = llvm::json::Object{};
+      obj = value.getAsObject();
+    }
+    (*obj)["format"] = llvm::json::Object{
+        {"numeric_type", numType}, {"is_signed", isSigned}, {"width", width}};
   }
 
 private:
@@ -502,6 +519,9 @@ private:
 
   /// A mapping between the callee and the instance.
   llvm::StringMap<calyx::InstanceOp> instanceMap;
+
+  /// A json file to store external global memory data
+  llvm::json::Value extMemData;
 };
 
 /// An interface for conversion passes that lower Calyx programs. This handles
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -47,7 +47,7 @@ namespace circt {
 class ComponentLoweringStateInterface;
 namespace scftocalyx {
 
-using json = nlohmann::json;
+using json = nlohmann::ordered_json;
 
 //===----------------------------------------------------------------------===//
 // Utility types
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -47,6 +47,8 @@ namespace circt {
 class ComponentLoweringStateInterface;
 namespace scftocalyx {
 
+using json = nlohmann::json;
+
 //===----------------------------------------------------------------------===//
 // Utility types
 //===----------------------------------------------------------------------===//
@@ -283,7 +285,7 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
                              scf::ParallelOp, scf::ReduceOp,
                              /// memref
                              memref::AllocOp, memref::AllocaOp, memref::LoadOp,
-                             memref::StoreOp,
+                             memref::StoreOp, memref::GetGlobalOp,
                              /// standard arithmetic
                              AddIOp, SubIOp, CmpIOp, ShLIOp, ShRUIOp, ShRSIOp,
                              AndIOp, XOrIOp, OrIOp, ExtUIOp, ExtSIOp, TruncIOp,
@@ -341,6 +343,8 @@ private:
   LogicalResult buildOp(PatternRewriter &rewriter, IndexCastOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, memref::AllocOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, memref::AllocaOp op) const;
+  LogicalResult buildOp(PatternRewriter &rewriter,
+                        memref::GetGlobalOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, memref::LoadOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, memref::StoreOp op) const;
   LogicalResult buildOp(PatternRewriter &rewriter, scf::WhileOp whileOp) const;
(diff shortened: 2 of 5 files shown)


include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -33,6 +34,8 @@
 namespace circt {
 namespace calyx {
 
+using json = nlohmann::ordered_json;
```
cgyurgyik, line 37: Does this dependency already exist in CIRCT?

cgyurgyik: Oh, I see you've just created an additional header file. I don't think copying and pasting a third-party header file into CIRCT source code is a good idea. Is there a reason why this can't be done via some external API instead of introducing this dependency?

uenoku: Drive-by comment: I agree we should not copy and paste a third-party header to CIRCT. FYI ESI has an external dependency to nlohmann/json https://github.com/llvm/circt/blob/ffde8abb3a6c1848604ed7007be9d9943bb56228/lib/Dialect/ESI/runtime/CMakeLists.txt#L40 so you can do the similar thing. Also LLVM has a json library which is used extensively within CIRCT (e.g. https://github.com/llvm/circt/blob/ffde8abb3a6c1848604ed7007be9d9943bb56228/lib/Dialect/Arc/ModelInfo.cpp#L126-L129) so it would be great if we can use LLVM json library there is no particular reason.

cgyurgyik: Thanks @uenoku !

jiahanxie353:
I have switched to using `llvm::json`; not sure how to test the output flushed to a json file though..

jiahanxie353:
Test case included! Ready for another round of review!

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -486,6 +519,9 @@ class ComponentLoweringStateInterface {
 
   /// A mapping between the callee and the instance.
   llvm::StringMap<calyx::InstanceOp> instanceMap;
+
+  /// A json file to store external global memory data
```
cgyurgyik, line 523: Please provide some context as to why a JSON file is needed from someone not as familiar with the native Calyx compiler; perhaps there is some documentation you can link to.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +1027,74 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+  int totalSize = 1;
+  for (auto dim : shape) {
+    totalSize *= dim;
+    dimensions.push_back(dim);
+  }
+
+  std::vector<double> flattenedVals(totalSize, 0);
+
+  // Helper function to get the correct indices
+  auto getIndices = [&dimensions](int flatIndex) {
```
cgyurgyik, line 1044: Can you place this closer to where it is used please?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +1027,74 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+    std::vector<int> indices(dimensions.size(), 0);
```
cgyurgyik, line 1045: It seems a bit overkill to allocate a new vector O(flattenedVals.size()) times. Do you think it would be worth it just to inline this code, and create 
`std::vector<int> indices(dimensions.size(), 0);` outside of the loop where this is called?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -306,10 +316,35 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
                                  : WalkResult::interrupt();
     });
 
+    if (!writeJson.empty()) {
+      if (auto fileLoc = dyn_cast<mlir::FileLineColLoc>(funcOp->getLoc())) {
+        std::string filename = fileLoc.getFilename().str();
+        // Hack: remove unwanted prefixes during testing
```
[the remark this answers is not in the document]
cgyurgyik: Maybe we can remove split input files for now since there's only one module, and then add a comment in the test file explaining this?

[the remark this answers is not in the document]
jiahanxie353: sounds good!

test/Conversion/SCFToCalyx/write_memory.mlir
```diff
@@ -0,0 +1,65 @@
...
+// CHECK:           ],
+// CHECK:           "format": {
+// CHECK:             "is_signed": true,
+// CHECK:             "numeric_type": "ieee754_float",
+// CHECK:             "width": 32
+// CHECK:           }
+// CHECK:         }
+
+module {
+  memref.global "private" constant @constant_10xi32_0 : memref<10xi32> = dense<[43, 8, -39, -19, 70, -64, -7, -27, -57, 5]>
```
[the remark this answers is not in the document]
jiahanxie353: I have actually forced all incoming memories to be a one-dim, as I found this discussion: https://github.com/calyxir/calyx/issues/907
(which also means I have made the insert nested json recursive function non-recursive

what do you think?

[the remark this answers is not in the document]
cgyurgyik: What do you mean by "forced"?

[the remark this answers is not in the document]
jiahanxie353: by early returning `failure` if it's not one-dimension

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: not used anymore, resolving conversation

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: not used anymore, resolving conversation

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+        int64_t value = intAttr.getInt();
+        flattenedVals[sizeCount++] = static_cast<uint64_t>(value);
+      }
+    }
+
+    rewriter.eraseOp(globalOp);
+  }
+
+  llvm::json::Array result;
+  if (shape.empty()) {
```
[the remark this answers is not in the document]
jiahanxie353: this test: https://github.com/llvm/circt/blob/cf45d9410f94922d59fb994826c74931b949fe4b/test/Conversion/SCFToCalyx/convert_memory.mlir#L807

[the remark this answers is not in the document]
cgyurgyik: Got it. Can you add a test below for this then?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+    if (bitValue & (1ULL << 63))
+      isSigned = true;
+    llvm::json::Value value = 0;
+    if (isFloat) {
+      double floatValue;
+      std::memcpy(&floatValue, &bitValue, sizeof(double));
+      value = floatValue;
+    } else {
+      int64_t intValue = static_cast<int64_t>(bitValue);
+      value = intValue;
```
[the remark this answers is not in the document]
jiahanxie353: we want regular floating point/integer values when writing to json; writing a bitvector to json is just pain to read for Calyx users..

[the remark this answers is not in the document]
jiahanxie353: then it brings the question of "why did we cast it to bitvectors initially". I thought you were suggesting to have a uniform representation for float/integers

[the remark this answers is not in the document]
cgyurgyik: Ah, I had forgotten we represent the typed value in the data format now and not just raw bits.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+        globalOp.getConstantInitValue());
+    int sizeCount = 0;
+    for (auto attr : cstAttr.template getValues<Attribute>()) {
+      assert((isa<mlir::FloatAttr>(attr) || isa<mlir::IntegerAttr>(attr)) &&
+             "memory attributes must be float or int");
+      if (auto fltAttr = dyn_cast<mlir::FloatAttr>(attr)) {
+        double value = fltAttr.getValueAsDouble();
+        std::memcpy(&flattenedVals[sizeCount++], &value, sizeof(double));
+      } else {
+        auto intAttr = dyn_cast<mlir::IntegerAttr>(attr);
```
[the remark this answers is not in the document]
jiahanxie353: no there is no explicit guarantees from the `getGlobal` side; I just thought there would be no other attributes, like `StringAttr`, we should expect only numbers like integer and float to pass in right?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+  bool isFloat = !memtype.getElementType().isInteger();
+
+  auto shape = allocOp.getType().getShape();
+  int totalSize =
+      std::reduce(shape.begin(), shape.end(), 1, std::multiplies<int>());
+  // The `totalSize <= 1` check is a hack to:
+  // https://github.com/llvm/circt/pull/2661 We should instead never pass
+  // multi-dimensional memories to Calyx, as discussed in:
+  // https://github.com/calyxir/calyx/issues/907
+  if (!(shape.size() <= 1 || totalSize <= 1)) {
```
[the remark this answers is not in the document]
jiahanxie353: `totalSize` is for:
https://github.com/llvm/circt/blob/cf45d9410f94922d59fb994826c74931b949fe4b/test/Conversion/SCFToCalyx/convert_memory.mlir#L769

where its `shape.size()` is 4, but it can be collapsed to a `memref<1xi32>`

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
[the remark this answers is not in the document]
cgyurgyik: Mistakes happen, I'd personally feel most comfortable with some assertion or error returned, because right now it's UB if it is not an Integer/Float

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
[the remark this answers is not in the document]
cgyurgyik: Great, could you add that as an example in the comment above? It makes it much clearer.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+    auto *symbolTableOp =
+        getGlobalOp->template getParentWithTrait<mlir::OpTrait::SymbolTable>();
+    auto globalOp = dyn_cast_or_null<memref::GlobalOp>(
+        SymbolTable::lookupSymbolIn(symbolTableOp, getGlobalOp.getNameAttr()));
+    // Flatten the values in the attribute
+    auto cstAttr = llvm::dyn_cast_or_null<DenseElementsAttr>(
+        globalOp.getConstantInitValue());
+    int sizeCount = 0;
+    for (auto attr : cstAttr.template getValues<Attribute>()) {
+      assert((isa<mlir::FloatAttr>(attr) || isa<mlir::IntegerAttr>(attr)) &&
```
jiahanxie353, line 1030: > Mistakes happen, I'd personally feel most comfortable with some assertion or error returned, because right now it's UB if it is not an Integer/Float

can you explain what does `UB` stand for? @cgyurgyik 
and here is the assertion right?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+    }
+  }
+
+  bool isSigned = false;
+  // Put the flattened values in the multi-dimensional structure
+  for (size_t i = 0; i < flattenedVals.size(); ++i) {
+    uint64_t bitValue = flattenedVals[i];
+    // checks whether the MSB of the `uint64_t` type `bitValue` is set
+    if (bitValue & (1ULL << 63))
+      isSigned = true;
```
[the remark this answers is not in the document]
jiahanxie353: I went to verify the result of [`isSignlessInteger`](https://mlir.llvm.org/doxygen/classmlir_1_1Type.html#a53765d45f36d2bbff81bb81b286a9525), I believe it always returns `signless = true` for all `iN` type; and it only returns `signless = false` for `siN` types: https://mlir.llvm.org/docs/Dialects/Builtin/#integertype

so I think it's really up to the specific operation to determines the semantics, and we are determining it via checking the most significant bit here, which I think makes sense?

jiahanxie353:
fixed the signless integer problem!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+      APInt apInt(/*numBits=*/elmTyBitWidth, bitValue);
+      value =
+          isSignlessOrUnsigned ? apInt.getZExtValue() : apInt.getSExtValue();
+    }
+    result[i] = std::move(value);
+  }
+
+  componentState.setDataField(memoryOp.getName(), result);
+
+  auto width = memtype.getElementType().getIntOrFloatBitWidth();
```
cgyurgyik, line 1075: Above you call:

```
unsigned elmTyBitWidth = memtype.getElementTypeBitWidth();
``` 

I think that's the same right?

cgyurgyik (commented):
Nice! A few smaller style comments, but overall looks great.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +997,88 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+
+  auto shape = allocOp.getType().getShape();
+  int totalSize =
+      std::reduce(shape.begin(), shape.end(), 1, std::multiplies<int>());
+  // The `totalSize <= 1` check is a hack to:
+  // https://github.com/llvm/circt/pull/2661, where a multi-dim memory whose
+  // size in some dimension equals 1, e.g. memref<1x1x1x1xi32>, will be
+  // collapsed to `memref<1xi32>`, whose `totalSize` equals 1. We should instead
+  // never pass multi-dimensional memories to Calyx, as discussed in:
+  // https://github.com/calyxir/calyx/issues/907
```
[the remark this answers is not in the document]
jiahanxie353: I prefer update the comment for now to respect that contributer's work; and potentially delete it in the future if it becomes necessary. If you agree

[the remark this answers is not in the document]
cgyurgyik: SGTM

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: sounds good!

jiahanxie353:
Made changes based on the feedback, thank you @cgyurgyik !
And now including all floating point type, signless integer type, and signed integer type in testing to test comprehensively.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1035,49 +1039,44 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
             bit_cast<uint64_t>(fltAttr.getValueAsDouble());
       } else {
         auto intAttr = dyn_cast<mlir::IntegerAttr>(attr);
-        int64_t value = intAttr.getInt();
```
jiahanxie353: because `getInt()` only works for signless or index type. So switching to use getting an `APInt value` to handle signed and signless uniformly

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +1001,83 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+      // value's precision handling
+      value = bit_cast<double>(bitValue);
+    } else {
+      APInt apInt(/*numBits=*/elmTyBitWidth, bitValue,
+                  /*isSigned=*/!isSignlessOrUnsigned);
+      if (!isSignlessOrUnsigned) {
+        // need to explicitly assign it to an int variable before storing to a
+        // json Value, otherwise it will interpret the raw data
+        int64_t signedValue = apInt.getSExtValue();
+        value = signedValue;
```
cgyurgyik, line 1068: can also do:
```suggestion
        value = static_cast<int64_t>(apInt.getSExtValue())
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +1001,83 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+  llvm::json::Array result;
+  result.reserve(std::max(static_cast<int>(shape.size()), 1));
+
+  bool isSignlessOrUnsigned = memtype.getElementType().isSignlessInteger() ||
+                              memtype.getElementType().isUnsignedInteger();
+  for (uint64_t bitValue : flattenedVals) {
+    llvm::json::Value value = 0;
+    if (isFloat) {
+      // we cast to `double` and let downstream calyx to deal with the actual
+      // value's precision handling
```
cgyurgyik, line 1059: Nit: Please use proper punctuation in your comments, i.e., capitalization and end with period.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +1001,83 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+                  /*isSigned=*/!isSignlessOrUnsigned);
+      if (!isSignlessOrUnsigned) {
+        // need to explicitly assign it to an int variable before storing to a
+        // json Value, otherwise it will interpret the raw data
+        int64_t signedValue = apInt.getSExtValue();
+        value = signedValue;
+      } else
+        value = apInt.getZExtValue();
+    }
+    result.push_back(value);
```
cgyurgyik, line 1072: std::move

> They can be copied, but should generally be moved.

Ref: https://llvm.org/doxygen/classllvm_1_1json_1_1Value.html

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -962,6 +1001,83 @@ static LogicalResult buildAllocOp(ComponentLoweringState &componentState,
...
+      }
+    }
+
+    rewriter.eraseOp(globalOp);
+  }
+
+  llvm::json::Array result;
+  result.reserve(std::max(static_cast<int>(shape.size()), 1));
+
+  bool isSignlessOrUnsigned = memtype.getElementType().isSignlessInteger() ||
```
cgyurgyik, line 1053: Suggestion: it seems that every place you use this its prefixed by `!`. Maybe just do

```cpp
Type elemType = memtype.getElementType();
bool isSigned = !elemType.isSignlessInteger() && !elemType.isUnsignedInteger();
```

(Although isn't that just the same as `isSignedInteger()`?)

cgyurgyik (approved):
A final few nits, please address them as you see fit. LGTM once tests pass. Nice job!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: sounds good!

> Although isn't that just the same as isSignedInteger()?

`isSigned` should be true when `isSignedIntegerOrFloat`, but there is no such built-in check; only `isSignlessOrFloat` is present, but that's not what we want. 
Or we can use `isSignedInteger || isFloat`, but there is again no built-in `isFloat` check, only those specific checks like `isFloat8E3M4()`.
So I have to use the negation logic..

jiahanxie353:
Planning to squash and merge once the CI passes. Thank you for the review!

PR #7302, opened 2024-07-10
[SV] Add Intermediary Assert Op for better enable polarity flip

dobios:
This PR introduces a new assert property op to the sv dialect and uses that as an intermediary for property assertion emission. This should solve the issue of the polarity being different in SV and in verif for the enable signals ( enable in verif, disable in sv ).


include/circt/Dialect/SV/SVVerification.td
```diff
@@ -159,3 +159,58 @@ def CoverConcurrentOp : ConcurrentVerifOp<"cover.concurrent",
...
+  let assemblyFormat = [{
+    $property (`on` $event^)? ($clock^)? (`disable_iff` $disable^)?
+    (`label` $label^)? attr-dict
+  }];
+
+  let builders = [
+    // no clock or event
+    OpBuilder<(ins "mlir::Value":$property,
+      "mlir::Value":$disable,
+      StrAttr.storageType:$label),
```
dtzSiFive, line 182: Just curious, what's going on here?

include/circt/Dialect/SV/SVVerification.td
```diff
@@ -159,3 +159,58 @@ def CoverConcurrentOp : ConcurrentVerifOp<"cover.concurrent",
...
+      [{ build(odsBuilder, odsState, property, EventControlAttr{}, Value(), disable, label); }]>,
+    // no clock or event or label
+    OpBuilder<(ins "mlir::Value":$property,
+      "mlir::Value":$disable),
+      [{ build(odsBuilder, odsState, property, EventControlAttr{}, Value(), disable, StringAttr{}); }]>
+  ];
+}
+
+def AssertPropertyOp : PropertyVerifOp<"assert_property"> {
+  let summary = "property assertion -- can be disbled and clocked";
```
dtzSiFive, line 192:
```suggestion
  let summary = "property assertion -- can be disabled and clocked";
```

include/circt/Dialect/SV/SVVerification.td
```diff
@@ -159,3 +159,58 @@ def CoverConcurrentOp : ConcurrentVerifOp<"cover.concurrent",
...
+  let description = [{
+    Assert that a given SVA-style property holds. This is only checked when 
+    the disable signal is low and a clock event occurs. This is analogous to 
+    the verif.assert operation, but with a flipped enable polarity.
+  }];
+}
+
+def AssumePropertyOp : PropertyVerifOp<"assume_property"> {
+  let summary = "property assumption -- can be disbled and clocked";
```
dtzSiFive, line 201:
```suggestion
  let summary = "property assumption -- can be disabled and clocked";
```

include/circt/Dialect/SV/SVVerification.td
```diff
@@ -159,3 +159,58 @@ def CoverConcurrentOp : ConcurrentVerifOp<"cover.concurrent",
...
+  let description = [{
+    Assume that a given SVA-style property holds. This is only considered when 
+    the disable signal is low and a clock event occurs. This is analogous to 
+    the verif.assume operation, but with a flipped enable polarity.
+  }];
+}
+
+def CoverPropertyOp : PropertyVerifOp<"cover_property"> {
+  let summary = "property cover point -- can be disbled and clocked";
```
dtzSiFive, line 210:
```suggestion
  let summary = "property cover point -- can be disabled and clocked";
```

lib/Conversion/VerifToSV/VerifToSV.cpp
```diff
@@ -128,6 +128,126 @@ struct HasBeenResetConversion : public OpConversionPattern<HasBeenResetOp> {
...
+  llvm_unreachable("Unknown event control kind");
+}
+
+// Convert the verif.assert like op that uses an enable, into an equivalent
+// sv.assert_property op that has the negated enable as its disable.
+template <typename TargetOp, typename Op, typename Adaptor>
+static LogicalResult convert(Op op, Adaptor operands,
+                             ConversionPatternRewriter &rewriter) {
+  // negate the enable if it exists
+  Value enable, wdisable;
```
dtzSiFive, line 151: Nit: `if (auto enable = operands.getEnable())` since it's not used outside the if block? And as a result can drop the parens re:assignment.

lib/Conversion/VerifToSV/VerifToSV.cpp
```diff
@@ -128,6 +128,126 @@ struct HasBeenResetConversion : public OpConversionPattern<HasBeenResetOp> {
...
+  if ((enable = operands.getEnable())) {
+    Value disable = rewriter.create<comb::XorOp>(op.getLoc(), enable, enable);
```
dtzSiFive, line 153: Consider `createOrFold` (similarly variant below)?

lib/Conversion/VerifToSV/VerifToSV.cpp
```diff
@@ -128,6 +128,126 @@ struct HasBeenResetConversion : public OpConversionPattern<HasBeenResetOp> {
...
+      op.getContext(), verifToSVEventControl(operands.getEdge()));
+
+  rewriter.replaceOpWithNewOp<TargetOp>(op, operands.getProperty(), eventattr,
+                                        operands.getClock(), wdisable,
+                                        operands.getLabelAttr());
+  return success();
+}
+} // namespace VerifAssertLikeOp
+
+struct VerifAssertConversion : public OpConversionPattern<verif::AssertOp> {
```
dtzSiFive, line 181: Are these conversion patterns used? (maybe not yet, hence draft/WIP?)

include/circt/Dialect/SV/SVVerification.td
dobios: just some custom builders to make it easier to use the optional parameters in any order.

lib/Conversion/VerifToSV/VerifToSV.cpp
```diff
@@ -128,6 +128,128 @@ struct HasBeenResetConversion : public OpConversionPattern<HasBeenResetOp> {
...
+
+template <typename TargetOp, typename Op, typename Adaptor>
+static LogicalResult convertClocked(Op op, Adaptor operands,
+                                    ConversionPatternRewriter &rewriter) {
+  // negate the enable if it exists
+  Value wdisable;
+  if (auto enable = operands.getEnable()) {
+    Value disable =
+        rewriter.createOrFold<comb::XorOp>(op.getLoc(), enable, enable);
+    wdisable = rewriter.createOrFold<hw::WireOp>(op.getLoc(), disable);
```
uenoku, line 171: I think a wire will be immediately folded away. We should insert a wire in PrepareForEmission.

test/Conversion/ExportVerilog/verif.mlir
```diff
@@ -261,18 +261,18 @@ hw.module @SystemVerilogSpecExamples(in %clk: i1, in %a: i1, in %b: i1, in %c: i
   %b1 = ltl.concat %b, %b0, %a1 : i1, !ltl.sequence, !ltl.sequence
   %b2 = ltl.implication %a, %b1 : i1, !ltl.sequence
   %b3 = ltl.clock %b2, posedge %clk : !ltl.property
-  verif.assert %b3 : !ltl.property
+  sv.assert_property %b3 : !ltl.property
 
-  // CHECK: assert property (disable iff (~(e)) @(posedge clk) a |-> not b ##1 c ##1 d);
+  // CHECK: assert property (disable iff (e) @(posedge clk) a |-> not b ##1 c ##1 d);
```
uenoku, line 266: Please add a test case that checks `~e` is properly spilled to a wire.

lib/Conversion/VerifToSV/VerifToSV.cpp
```diff
@@ -128,6 +128,128 @@ struct HasBeenResetConversion : public OpConversionPattern<HasBeenResetOp> {
...
+      op.getContext(), verifToSVEventControl(operands.getEdge()));
+
+  rewriter.replaceOpWithNewOp<TargetOp>(op, operands.getProperty(), eventattr,
+                                        operands.getClock(), wdisable,
+                                        operands.getLabelAttr());
+  return success();
+}
+} // namespace VerifAssertLikeOp
+
+struct VerifAssertConversion : public OpConversionPattern<verif::AssertOp> {
```
uenoku, line 183: Could you please combine 6 conversion pattern structs into into a struct with a template?

lib/Conversion/VerifToSV/VerifToSV.cpp
```diff
@@ -128,6 +128,128 @@ struct HasBeenResetConversion : public OpConversionPattern<HasBeenResetOp> {
...
+}
+
+template <typename TargetOp, typename Op, typename Adaptor>
+static LogicalResult convertClocked(Op op, Adaptor operands,
+                                    ConversionPatternRewriter &rewriter) {
+  // negate the enable if it exists
+  Value wdisable;
+  if (auto enable = operands.getEnable()) {
+    Value disable =
+        rewriter.createOrFold<comb::XorOp>(op.getLoc(), enable, enable);
```
uenoku, line 170: The second operand must be 1?

lib/Conversion/VerifToSV/VerifToSV.cpp
uenoku: I think the right place to do this would be here https://github.com/llvm/circt/blob/961bad58f4f904f34b103b8072748e6441f53717/lib/Conversion/ExportVerilog/ExportVerilog.cpp#L784

uenoku (approved):
LGTM, I would like to follow up PrepareForEmission since we should do the spilling at the same place as `allowExprInEventControl` handling but I don't block the PR.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -3450,28 +3438,22 @@ class PropertyEmitter : public EmitterBase,
...
-  /// Emit the specified value as an SVA property or sequence with an enable
-  /// signal. This is the entry point to print an entire tree of property or
-  /// sequence expressions in one go.
-  void emitEnabledProperty(
-      Value property, Value enable,
-      PropertyPrecedence parenthesizeIfLooserThan = PropertyPrecedence::Lowest);
+  void emitAssertPropertyBody(
+      Value property, Value disable,
+      PropertyPrecedence parenthesizeIfLooserThan = PropertyPrecedence::Lowest,
+      bool ignoreNonEmptyTokens = false);
```
dtzSiFive, line 3444: The boolean flag is a bit rough here I think, and it's basically "I'm calling this as the tail of a different entry-point-y function" (since it does the flush) and is for internal use only as a result.

Additionally, functionally that boolean is dead in non-assert builds.

Putting the common functionality in a helper method the entry-ish methods (the public ones) call to implement their behavior is the usual pattern and avoids the need to answer questions about the flag.  This would mean the entry-points do the buffer management dance.

Thanks!

PR #7303, opened 2024-07-11
[SV] Add UnpackedArrayCreateOp

uenoku:
This commit adds unpacked array create op that will be lowered into array literal
 `'{..}`.

Since array literal must be used in assignment-like context,`unpacked_array_create` is always spilled to a temporary wire.

The operand ordering follows HW array representation. This is currently tricky since we emit unpacked types slightly differently (`downTo`) but the correct ordering is emitted in ExportVeriliog.


fabianschuiki (approved):
Very nice! Having an op to explicitly construct unpacked arrays is going to help with lowering debug hierarchies to SV :+1:

PR #7305, opened 2024-07-11
[FIRRTL] Add a intrinsic for defining DPI argument with open array type

uenoku:
This commit adds `int.dpi.open_uarray_cast` operation that is used to indicate DPI argument to have an open unpacked array type instead of a packed fixed-size array type. FIRRTL lowering by default lowers FIRRTL vector into packed arrays but in DPI it's very common to use open arrays to send dynamic length of data to DPI functions. The user of this operation must be DPI call intrinsic.


fabianschuiki (approved):
I think this makes sense!

If I recall correctly, passing packed arrays through DPI directly can have weird memory layouts that makes it difficult to extract elements on the C side (because of the bit-dense packing in SV). Making arrays always go through an open array conversion where the C side can easily reason about the dimensions sounds like a good idea :+1:

sequencer:
Looking forward to adopt this PR:) wish it can be used for deduping a massive DPI call of interface with different parameters.

PR #7306, opened 2024-07-11
[ESI] MMIO read service implementation in PyCDE

teqdruid:
Implements a channel-based MMIO read service generator. Changes the offset type on the MMIO std service to `ui32` from `i32`.

### DIFF
--- a/frontends/PyCDE/src/pycde/esi.py
+++ b/frontends/PyCDE/src/pycde/esi.py
@@ -664,7 +664,7 @@ def ChannelDemux2(data_type: Type):
 
 def ChannelDemux(input: ChannelSignal, sel: BitsSignal,
                  num_outs: int) -> List[ChannelSignal]:
-  """Build a pipelined demultiplexer of ESI channels. Ideally, this would be a
+  """Build a demultiplexer of ESI channels. Ideally, this would be a
   parameterized module with an array of output channels, but the current ESI
   channel-port lowering doesn't deal with arrays of channels. Independent of the
   signaling protocol."""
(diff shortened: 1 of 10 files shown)


frontends/PyCDE/src/pycde/esi.py
```diff
@@ -624,3 +627,120 @@ def package(sys: System):
     circt_lib_dir = build_dir / "tools" / "circt" / "lib"
     esi_lib_dir = circt_lib_dir / "Dialect" / "ESI"
   # shutil.copy(esi_lib_dir / "ESIPrimitives.sv", sys.hw_output_dir)
+
+
+def ChannelDemux2(data_type: Type):
+
+  class ChannelDemux2(Module):
+    """Combinational 2-way channel demultiplexer for valid/ready signaling."""
+
```
mortbopet, line 636: For any practical application, won't this completely kill fmax (under the assumption of a sufficiently large demux tree). I'm in favor of doing this for the initial POC, but this'll obviously have to change in the future...

frontends/PyCDE/src/pycde/esi.py
```diff
@@ -624,3 +627,120 @@ def package(sys: System):
...
+      input_ready.assign((output0_ready & (ports.sel == Bits(1)(0))) |
+                         (output1_ready & (ports.sel == Bits(1)(1))))
+
+  return ChannelDemux2
+
+
+def ChannelDemux(input: ChannelSignal, sel: BitsSignal,
+                 num_outs: int) -> List[ChannelSignal]:
+  """Build a pipelined demultiplexer of ESI channels. Ideally, this would be a
```
mortbopet, line 667: pipelined?

mortbopet (approved):

frontends/PyCDE/src/pycde/esi.py
teqdruid: The thought is to add buffering at a higher level in the instance hierarchy. (While assembling the tree.)

teqdruid: I plan on adding buffers on the input and output of the Channel[De]mux function after the next PR.

Issue #7078, opened 2024-05-22
[OM] Rework Class and Object op to finish the move away from fields being symbols


mikeurbach:
Since the OM dialect's inception, fields were symbols in the Class. This was problematic, because Classes themselves are symbols in the builtin Module symbol table, so they can be referenced by Object ops and Object field ops.

As we have found time and again, nested symbol tables don't work, so we made fields no longer a symbol. But this meant we had to scan the class body while verifying fields. That became a performance problem, so we moved this logic out of a verifier, and instead ran a global Class/Object verification pass at certain points in the pipeline: https://github.com/llvm/circt/pull/7026

This issue is to track the final resolution of this saga, which will probably encompass a few concrete steps.

In short, I think it makes sense for Class / Object to behave more like HWModule / Instance in the HW dialect. This will necessitate a few changes:

* Rather than om.field ops in the class body, we can move to a terminator op like om.fields, analogous to the hw.output op
* This will require an attribute somewhere, probably on the om.class, to hold the field names, and verify that the names array matches the number of operands to the om.fields op

This is sufficient to get away from the half-baked sort of symbolic representation of om.field. But we need to address the core issue, which was verifier performance.

It might be possible to keep the om.class type as just a symbol reference to the om.class op with the above changes. The om.object op can verify the class exists and takes the specified inputs. The om.object.field op may be able to efficiently verify that the requested field exists with the correct type. One thing we can change is the om.object.field doesn't need to support a path of fields--it turned out we never needed this ability--it is only ever one field at a time.

If it would be helpful for verification performance, we could also consider augmenting the om.class type to include the fields, more like the hw.module_type. This would still need to be verified against the om.class, probably in the om.object verifier, but then the om.object.field verifier could check against the type directly.

mikeurbach:
One thing @uenoku brought up is we will have cyclic references. So putting the array of field types into the `om.class` type might not be feasible.

PR #7312, opened 2024-07-12
[OM] Introduce API to interact with class fields

leonardt:
The goal is to introduce an abstraction that allows us to change how fields are represented under the hood.  This is a step towards addressing https://github.com/llvm/circt/issues/7078 which aims to use a terminator op to declare and store the field name/type mapping.  With this API in place, we can change how the underlying fields are accessed without having to update the producers/consumers of this information.

Issues:
* We introduce a `Field` struct to provide the information required by the evaluator.  We could return the FieldOp instead since this just passes through the relevant data, but if we change to fetching this information from the terminator op instead, we'd need to update the consumer code, so introducing a new intermediate type abstracts how the information is stored/fetched (at the cost of introducing a layer of indirection).  If we think this will be a performance issue, we may want to passthrough the underlying representation (currently FieldOp) and update the references when the internal implementation is changed.
* The `addField` API will provide a hook to record field names and types for emitting the final terminator op, however is there a way for the `ClassOp` to run internal code at completion of the `lowerClass` logic? Otherwise, we will need to update the user code to run some kind of terminator function to tell the class to construct the final terminator op. E.g. some call here to "emit fields" https://github.com/llvm/circt/blob/e616cf1d327bd6748304f78755b034ba037224f0/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp#L972
* Is there a way for `getFields` to have a type specified as an "iterator emitting type Field"? That is, the interface should be any range iterator over a collection of `Field` members (not necessarily the mapped iterator implementation used here).  Hiding it behind a name helps enforce the interface, but if someone were to expect that mapped iterator type, it would fail if we changed the underlying implementation to emit the fields differently.  I think the appropriate type would be some `Iterator<Field>` type where any iterator producing a Field could be used.  Exploring various ways to approach this using strategies such as iterator_traits etc... did not reveal anything useful yet.

### DIFF
--- a/include/circt/Dialect/OM/OMOps.td
+++ b/include/circt/Dialect/OM/OMOps.td
@@ -82,6 +82,15 @@ def ClassOp : OMClassLike<"class"> {
     static mlir::RegionKind getRegionKind(unsigned index) {
       return mlir::RegionKind::Graph;
     }
+
+    void addField(mlir::OpBuilder &builder, mlir::Location loc,
+                  llvm::StringRef name, mlir::Value src);
+
+    using fields_iterator =
+        llvm::mapped_iterator<mlir::detail::op_iterator<circt::om::ClassFieldOp,
+                                                        mlir::Region::OpIterator>,
+                              circt::om::Field (*)(circt::om::ClassFieldOp)>;
+    llvm::iterator_range<fields_iterator> getFields();
   }];
 }
 
--- a/include/circt/Dialect/OM/OMTypes.h
+++ b/include/circt/Dialect/OM/OMTypes.h
@@ -21,6 +21,12 @@ namespace circt::om {
 // integer.
 bool isMapKeyValuePairType(mlir::Type);
 
+struct Field {
+  mlir::StringAttr name;
+  mlir::Value value;
+  mlir::Location loc;
+};
+
 } // namespace circt::om
 
 #define GET_TYPEDEF_CLASSES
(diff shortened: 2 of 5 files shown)


mikeurbach:
Nice, thanks for starting to put up some kind of abstraction over fields here. A few ideas inline...

> The addField API will provide a hook to record field names and types for emitting the final terminator op, however is there a way for the ClassOp to run internal code at completion of the lowerClass logic?

Maybe in this future version, we would make the terminator op always required, but with a variadic number of operands that is allowed to be zero. And in the builder for ClassOp, we would always create the terminator just with no operands. Then, addField would add the operand to the terminator.

That should work, but I haven't looked into the performance of iteratively adding operands to the same operation one by one. It might be that performance is a concern here, and we'll just want a bulk addFields API anyway.

There might be clever ways to call addField multiple times and batch up internally before bulk updating (or creating) the terminator, but I think it's reasonable to push the complexity to the compiler author to keep track of all the fields they want to add.

> I think the appropriate type would be some Iterator<Field>

Hmm, not sure what the best type for this would be... Often we return an ArrayRef if we are just providing a view into storage we own, but I'm not sure the best way to return an `ArrayRef<Field>` in this case. There must be some kind of base iterator type we could use, instead of the specific map_range iterator.

leonardt:
Incrementally building the terminator makes sense, I can take a stab at that since it would preserve this natural incremental `addField` API and requires minimal changes to the client code, but switching to a bulk operation should be easy if performance does become an issue.

leonardt:
Closing in favor of https://github.com/llvm/circt/pull/7383 I ended up using a terminator style single op constructor pattern since incrementally modifying a terminator op (and locations) seemed to not be working well.

PR #7313, opened 2024-07-12
[Moore] Extend the operand of AssignOp and add PullNonBlockingUp pass.

hailongSun2000:
`moore.*assign` has a new extra operand--`enable` to capture the if condition, like `moore.assign %dst, %src if %enable`
`PullNonBlockingUp` pass is aimed at pulling non-blokcing assignments upwards.
From:
```
moore.procedure {
  scf.if not(%rst) {
    ...
  } else {
    scf.if eq(cond1, a1) {
      %0 = moore.constant 1
      %1 = moore.extract_ref %q, %0
      moore.nonblokcing %1, %0
    }

    scf.if eq(cond2, a2) {
       %0 = moore.constant 1
       %1 = moore.extract_ref %q, %0
       moore.nonblokcing %1, %0
    }
   }
}
```
To:
```
moore.procedure {
  scf.if not(%rst) {
    ...
  } else {
    %0 = moore.constant 1
    %1 = moore.extract_ref %q, %0
    moore.nonblokcing %1, %0 if eq(cond1, a1)
    %2 = moore.constant 1
    %3 = moore.extract_ref %q, %0
    moore.nonblokcing %3, %2 if eq(cond2, a2)
   }
}
```
@fabianschuiki mentioned at this PR(https://github.com/llvm/circt/pull/7075).

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -296,9 +296,13 @@ def ReadOp : MooreOp<"read", [
 class AssignOpBase<string mnemonic, list<Trait> traits = []> :
     MooreOp<mnemonic, traits # [TypesMatchWith<"src and dst types match",
     "src", "dst", "RefType::get(cast<UnpackedType>($_self))">]> {
-  let arguments = (ins RefType:$dst, UnpackedType:$src);
+  let arguments = (ins
+    RefType:$dst, UnpackedType:$src,
+    Optional<AnySingleBitType>:$enable
+  );
   let assemblyFormat = [{
-    $dst `,` $src attr-dict `:` type($src)
+    $dst `,` $src (`if` $enable^)? attr-dict 
+    `:` type($src) (`,` type($enable)^)?
   }];
 }
 
@@ -326,7 +330,8 @@ def BlockingAssignOp : AssignOpBase<"blocking_assign", [
   }];
   let arguments = (ins
     Arg<RefType, "", [MemWrite]>:$dst,
-    UnpackedType:$src
+    UnpackedType:$src,
+    Optional<AnySingleBitType>:$enable
   );
 }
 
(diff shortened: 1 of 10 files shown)


fabianschuiki:
I'm curious what @maerhart, @uenoku, and @dtzSiFive think of this. This is essentially a conditional store operation in order to pull them out of control flow regions, in the hopes that you'll end up with a single store at the exit of the procedure.

uenoku (commented):
I'm not sure adding a condition to assignment is a tractable solution here. Can we eliminate the condition by lowering a condition to a mux in a similar way to FIRRTL ExpandWhens?

maerhart:
I don't know how ExpandWhens works, but there's a difference between not driving any value on a signal and driving the current value again. So, using muxes is probably not always possible.

I'm not sure if this increased complexity at the Moore level is worth it. If I understand correctly, the motivation here is to simplify pattern matching the procedure to replace it with a register, right?
However, we already have a working flow using LLHD, so we could lower directly to `llhd.proc` which allows arbitrary basic blocks, so we can just reuse the SCF to CF lowering and then run the regular behavioral LLHD to structural LLHD pipeline.
Are there any optimizations on the Moore level that would benefit from this change?

fabianschuiki:
@maerhart Lowering to LLHD and then relying on the desequentialization/register detection pass there would definitely work. I think the initial idea was to see if there are simple patterns we can already detect at the Moore dialect level, and then lower them directly to `seq.compreg` or `seq.firreg`. But you're right, LLHD would be the path that should always work. So maybe trying to do this at the Moore level is not worth it?

maerhart:
Yeah, the question is where to draw the line and use the full LLHD lowering instead of the simple pattern matching. Maybe it's better to collapse assigns using muxes as @uenoku suggested when possible, but stay away from the optional condition operand and use the LLHD lowering instead when this would be necessary.

PR #719, opened 2021-03-04
[SV] Verify fwrite/bpassign/passign are in a procedural region

uenoku:
This PR introduces `ProceduralOp` trait which ensures that operations with this trait must be in Procedural Region.
Currently, I added the trait to fwrite/bpassign/passign since it was unclear to me for other ops.

Will close #628

PR #7314, opened 2024-07-12
[HW][SV] Allow procedural SV ops in hw::TriggeredOp

fzi-hielscher:
Adapt the SV dialect `ProceduralOp` / `NonProceduralOp` traits to consider `hw::TriggeredOp` a procedural region. This should enable us to lower the procedural operations within a `hw::TriggeredOp` region before lowering the entire module.

See the discussion in #7292. We currently do not have a cross-dialect notion of procedural regions. However, the description of `hw::TriggeredOp` states bluntly that it creates a procedural region, so I think this is a relatively intuitive change.

Piggybacked are two sanity checks for `hw::TriggeredOp`:
- Don't allow nesting of TriggeredOps
- Ensure the number of input arguments matches the number of block arguments

### DIFF
--- a/include/circt/Dialect/HW/HWStructure.td
+++ b/include/circt/Dialect/HW/HWStructure.td
@@ -730,6 +730,9 @@ def TriggeredOp : HWOp<"triggered", [
   let builders = [
     OpBuilder<(ins "EventControlAttr":$event, "Value":$trigger, "ValueRange":$inputs)>
   ];
+
+  let hasVerifier = 1;
+  let hasRegionVerifier = 1;
 }
 
 #endif // CIRCT_DIALECT_HW_HWSTRUCTURE_TD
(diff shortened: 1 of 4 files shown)


uenoku (commented):
Hmm, I'm not sure it's reasonable change to regard regions hw::TriggeredOp as ProceduralRegion in verifier. I'm more comfortable if hw::TriggeredOp just implements `ProceduralRegion` trait in that case. In order to do that I think it's necessary to promote the notion of procedural regions from SV to HW but it will require more broader discussion.

fzi-hielscher:
I agree that this is far from ideal. I had, in fact, tried to move `ProceduralRegion` into HW, but then stumbled over the question on how to deal with this transitively w.r.t. upstream dialects (notably SCF). CC #719.

So, I went with the quick and dirty solution allowing me to temporarily have a `sv.if` in a `hw.triggered`. 😬

fzi-hielscher:
Superseded by #7335 .

Issue #6517, opened 2023-12-12
[Scheduling] Rethink use of macros in problem classes


jopperm:
Problem classes in the scheduling infra currently use a macro to define a protected constructor and common members. The original rationale was to be able to use default constructors throughout the hierarchy. However, we should replace the macro with an inheritance-based approach.

PR #7320, opened 2024-07-14
[Scheduling] Replace macro use in problem definitions

jopperm:
Removes the `DEFINE_COMMON_MEMBERS` macro from the definitions of the built-in scheduling problems to address #6517. Inheriting the base class' constructors keeps the boilerplate code small, but unfortunately I still have to declare the default constructor as protected in each class in order to prevent user code from creating problem instances without the `containingOp`.

A useful side-effect of exposing the constructor is that is now possible to create problem instances on the heap.

### DIFF
--- a/include/circt/Dialect/SSP/Utilities.h
+++ b/include/circt/Dialect/SSP/Utilities.h
@@ -142,7 +142,7 @@ ProblemT loadProblem(InstanceOp instOp,
                      std::tuple<OperatorTypePropertyTs...> oprProps,
                      std::tuple<DependencePropertyTs...> depProps,
                      std::tuple<InstancePropertyTs...> instProps) {
-  auto prob = ProblemT::get(instOp);
+  ProblemT prob(instOp);
 
   loadInstanceProperties<ProblemT, InstancePropertyTs...>(
       prob, instOp.getSspPropertiesAttr());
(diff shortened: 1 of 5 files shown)


mikeurbach (approved):
Makes sense to me, thanks for the cleanup!

7FM (approved):
Sorry for the late review, but I was on vacation.

LGTM, thx!

I find the logging sometimes a bit too verbose, especially when the `containingOp` is very large. A workaround that I sometimes use OoT is something like `mlir::emitError(OpBuilder{mlirContext}.getUnknownLoc()) << "Some context info:" << "your error message"`.
Some examples that might benefit from this approach:
- https://github.com/llvm/circt/blob/main/lib/Scheduling/Problems.cpp#L124C12-L127
- https://github.com/llvm/circt/blob/main/lib/Scheduling/Problems.cpp#L183-L187
- https://github.com/llvm/circt/blob/main/lib/Scheduling/Problems.cpp#L282-L285

Most of the time, the error messages are not directly related to the containing operation itself but rather to the way the scheduling problem was constructed.
But that is just personal preference and probably out of scope of this PR.

jopperm:
Thanks for the reviews!

> I find the logging sometimes a bit too verbose, especially when the containingOp is very large.

Good point, filed a follow-up issue for that.

PR #7321, opened 2024-07-15
[ImportVerilog] Tweak the implementation of for loops.

hailongSun2000:
If we want to unroll the loops that contain the non-blocking assignments, I think adding the block arguments for `scf.while` is easier to capture the `lower and upper bound and the step`. But the `step and execute body` are mixed. So I have another plan is to lower `for` statements into `scf.for`.

### DIFF
--- a/lib/Conversion/ImportVerilog/Statements.cpp
+++ b/lib/Conversion/ImportVerilog/Statements.cpp
@@ -179,32 +179,49 @@ struct StmtVisitor {
       return mlir::emitError(loc,
                              "variables in for loop initializer not supported");
 
-    // Generate the initializers.
-    for (auto *initExpr : stmt.initializers)
-      if (!context.convertRvalueExpression(*initExpr))
-        return failure();
+    // Use to collect the types and locations of initializer.
+    SmallVector<Type> initTypes;
+    SmallVector<Location> locs;
+
+    // Generate the initializers. Initializer may be more than one.
+    auto inits =
+        llvm::to_vector(llvm::map_range(stmt.initializers, [&](auto *initExpr) {
+          return context.convertRvalueExpression(*initExpr);
+        }));
+    if (llvm::any_of(inits, [](auto init) { return !init; }))
+      return failure();
+    for (auto init : inits) {
+      initTypes.push_back(init.getType());
+      locs.push_back(init.getLoc());
+    }
 
     // Create the while op.
-    auto whileOp = builder.create<scf::WhileOp>(loc, TypeRange{}, ValueRange{});
+    auto whileOp = builder.create<scf::WhileOp>(loc, initTypes, inits);
     OpBuilder::InsertionGuard guard(builder);
 
     // In the "before" region, check that the condition holds.
-    builder.createBlock(&whileOp.getBefore());
+    builder.createBlock(&whileOp.getBefore(), {}, initTypes, locs);
     auto cond = context.convertRvalueExpression(*stmt.stopExpr);
     if (!cond)
       return failure();
     cond = builder.createOrFold<moore::BoolCastOp>(loc, cond);
     cond = builder.create<moore::ConversionOp>(loc, builder.getI1Type(), cond);
-    builder.create<mlir::scf::ConditionOp>(loc, cond, ValueRange{});
+    builder.create<mlir::scf::ConditionOp>(loc, cond,
+                                           whileOp.getBeforeArguments());
 
     // In the "after" region, generate the loop body and step expressions.
-    builder.createBlock(&whileOp.getAfter());
+    builder.createBlock(&whileOp.getAfter(), {}, initTypes, locs);
     if (failed(context.convertStatement(stmt.body)))
       return failure();
-    for (auto *stepExpr : stmt.steps)
-      if (!context.convertRvalueExpression(*stepExpr))
-        return failure();
-    builder.create<mlir::scf::YieldOp>(loc);
+
+    // Step may be more than one.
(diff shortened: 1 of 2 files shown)


lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -179,32 +179,49 @@ struct StmtVisitor {
...
-    for (auto *initExpr : stmt.initializers)
-      if (!context.convertRvalueExpression(*initExpr))
-        return failure();
+    // Use to collect the types and locations of initializer.
+    SmallVector<Type> initTypes;
+    SmallVector<Location> locs;
+
+    // Generate the initializers. Initializer may be more than one.
+    auto inits =
+        llvm::to_vector(llvm::map_range(stmt.initializers, [&](auto *initExpr) {
```
hailongSun2000, line 188:
```suggestion
   
// Handle `for` loops.
  LogicalResult visit(const slang::ast::ForLoopStatement &stmt) {
    if (!stmt.loopVars.empty())
      return mlir::emitError(loc,
                             "variables in for loop initializer not supported");
    if (stmt.initializers.size() != 1)
      return mlir::emitError(loc,
                             "exactly one initializer expression expected");

    auto convertAndCast = [&](auto *expr, auto type) -> std::optional<Value> {
      if (auto value = context.convertRvalueExpression(*expr)) {
        type = type ? type
                    : builder.getIntegerType(
                          cast<moore::UnpackedType>(value.getType())
                              .getBitSize()
                              .value());
        return builder.create<moore::ConversionOp>(loc, type, value);
      }
      return std::nullopt;
    };

    auto lowerBound = *convertAndCast(stmt.initializers.front(), Type());
    auto upperBound = *convertAndCast(stmt.stopExpr, lowerBound.getType());
    auto step = *convertAndCast(stmt.steps.front(), lowerBound.getType());
    if (!lowerBound || !upperBound || !step)
      return mlir::emitError(loc, "failed to convert loop bounds or step");

    auto forOp = builder.create<scf::ForOp>(loc, lowerBound, upperBound, step);
    OpBuilder::InsertionGuard guard(builder);
    builder.setInsertionPointToStart(forOp.getBody());
    if (failed(context.convertStatement(stmt.body)))
      return failure();
    return success();
  }
```

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -299,16 +299,17 @@ module Statements;
...
-    // CHECK:   scf.condition([[COND]])
+    // CHECK:   scf.condition([[COND]]) %arg0 : !moore.i1
     // CHECK: } do {
+    // CHECK: ^bb0(%arg0: !moore.i1):
     // CHECK:   [[TMP3:%.+]] = moore.read %y : i1
     // CHECK:   moore.blocking_assign %x, [[TMP3]] : i1
     // CHECK:   [[TMP4:%.+]] = moore.read %z : i1
     // CHECK:   moore.blocking_assign %x, [[TMP4]] : i1
-    // CHECK:   scf.yield
+    // CHECK:   scf.yield [[TMP4]] : !moore.i1
```
hailongSun2000, line 312: If we lower into `scf.for`, get the follow IR:
```
%0 = moore.read %x : i1
moore.blocking_assign %y, %0 : i1
%1 = moore.conversion %0 : !moore.i1 -> i1
%2 = moore.read %x : i1
%3 = moore.conversion %2 : !moore.i1 -> i1
%4 = moore.read %z : i1
moore.blocking_assign %x, %4 : i1
%5 = moore.conversion %4 : !moore.i1 -> i1
scf.for %arg0 = %1 to %3 step %5  : i1 {
  %6 = moore.read %y : i1
  moore.blocking_assign %x, %6 : i1
}
```

hailongSun2000: I prefer to lower into `scf.for`. Because it execute-body only performs `x=y`.

fabianschuiki: I'm not sure if this is a correct lowering of the for loop below :thinking:. Written as pseudo-code, I think your lowering does the following:
```mlir
scf.for %arg0 = x to x step z {
  ...
}
```
But the for loop as written does not have a clear iteration variable and clear bounds. It behaves like a while loop like the following:
```mlir
assign y = x
scf.while x {
  assign x = z
  assign x = y
}
```
C-style for loops are very annoying for this reason :cry:... they aren't really clean loops with a well-known iteration/induction variable and iteration bounds. Those are a subset of all possible loops you can write.

To deal with this, could we add a canonicalization or lowering pass that maps loops with a clean iteration variable and loop bounds to `scf.for`? That would allow us to accept all possible loops as `scf.while` and opportunistically optimize them into the more constrained `scf.for` where possible.

hailongSun2000: I noticed `ForToWhile` pass in the SCF dialect, I wonder whether we can apply this pass to convert `scf.for` without non-blocking assignments into `scf.while`.

fabianschuiki: `scf.for` to `scf.while` should always work I think. But from the SystemVerilog AST we currently get while loops (they are technically called `for`, but in practice they are `while` loops with an initial statement and an increment/step statement). So I think we'd need the other way around? `WhileToFor`?

fabianschuiki: Good question... are there cases where you would want or need arguments?

hailongSun2000: No need now! If we lower `always@(*edge)` directly into the `llhd.proc`, I think it is unnecessary to modify this.

PR #7322, opened 2024-07-15
[FIRRTL] Add lowering support for inline layers

rwy7:
Update the firrtl-lower-layers pass to lower layers with the inline convention. Layers with an inline convention will have a macro declaration created for them, and any layerblock will be lowered to an sv.macro.ifdef operation.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -122,6 +122,14 @@ static SmallString<32> fileNameForLayer(StringRef circuitName,
   return result;
 }
 
+/// For a layerblock `@A::@B::@C`, the verilog macro is `A_B_C`.
+static SmallString<32> macroNameForLayer(ArrayRef<StringAttr> layerName) {
+  SmallString<32> result;
+  for (auto part : layerName)
+    appendName(part, result);
+  return result;
+}
+
 //===----------------------------------------------------------------------===//
 // LowerLayersPass
 //===----------------------------------------------------------------------===//
@@ -129,8 +137,8 @@ static SmallString<32> fileNameForLayer(StringRef circuitName,
 class LowerLayersPass
     : public circt::firrtl::impl::LowerLayersBase<LowerLayersPass> {
   hw::OutputFileAttr getOutputFile(SymbolRefAttr layerName) {
-    auto layer = dyn_cast<LayerOp>(
-        symbolTable->lookupSymbolIn(getOperation(), layerName));
+    auto layer =
+        cast<LayerOp>(SymbolTable::lookupSymbolIn(getOperation(), layerName));
     if (!layer)
       return nullptr;
     return layer->getAttrOfType<hw::OutputFileAttr>("output_file");
@@ -170,16 +178,25 @@ class LowerLayersPass
   /// remove the cast itself from the IR.
   void removeLayersFromRefCast(RefCastOp cast);
 
+  /// Lower an inline layerblock to an ifdef block.
+  void lowerInlineLayerBlock(LayerOp layer, LayerBlockOp layerBlock);
+
+  /// Create macro declarations for a given layer, and its child layers.
+  void createMacroDecls(CircuitNamespace &ns, OpBuilder &b, LayerOp layer,
+                        SmallVector<StringAttr> &stack);
+  void createMacroDecls(CircuitNamespace &ns, LayerOp layer);
+
   /// Entry point for the function.
   void runOnOperation() override;
 
   /// Indicates exclusive access to modify the circuitNamespace and the circuit.
   llvm::sys::SmartMutex<true> *circuitMutex;
 
-  SymbolTable *symbolTable;
-
   /// A map of layer blocks to module name that should be created for it.
   DenseMap<LayerBlockOp, StringRef> moduleNames;
+
+  /// A map from inline layers to their macro names.
+  DenseMap<LayerOp, FlatSymbolRefAttr> macroNames;
 };
 
 /// Multi-process safe function to build a module in the circuit and return it.
@@ -279,6 +296,15 @@ InnerRefMap LowerLayersPass::runOnModuleLike(FModuleLike moduleLike) {
   return innerRefMap;
 }
 
+void LowerLayersPass::lowerInlineLayerBlock(LayerOp layer,
+                                            LayerBlockOp layerBlock) {
+  OpBuilder builder(layerBlock);
+  auto macroName = macroNames[layer];
+  auto ifDef = builder.create<sv::IfDefOp>(layerBlock.getLoc(), macroName);
+  ifDef.getBodyRegion().takeBody(layerBlock.getBodyRegion());
+  layerBlock.erase();
+}
+
 void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
                                       InnerRefMap &innerRefMap) {
   CircuitOp circuitOp = moduleOp->getParentOfType<CircuitOp>();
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -130,7 +138,7 @@ class LowerLayersPass
     : public circt::firrtl::impl::LowerLayersBase<LowerLayersPass> {
   hw::OutputFileAttr getOutputFile(SymbolRefAttr layerName) {
     auto layer = dyn_cast<LayerOp>(
-        symbolTable->lookupSymbolIn(getOperation(), layerName));
+        SymbolTable::lookupSymbolIn(getOperation(), layerName));
```
rwy7, line 141: We ought to switch this pass over to use a SymbolTableCollection. In the meantime, I dropped the symbolTable because we weren't using it.

youngar (approved):

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -634,22 +698,26 @@ void LowerLayersPass::runOnOperation() {
...
+      });
+      continue;
+    }
+    // Build verilog macro declarations for each inline layer declarations.
+    if (auto layerOp = dyn_cast<LayerOp>(op)) {
+      OpBuilder b(layerOp);
+      createMacroDecls(ns, b, layerOp);
+      continue;
+    }
+  }
```
seldridge, line 720: This would be more future-proof as a walk or something which does not assume that `LayerOp` is at the top-level of a module.

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -350,6 +350,75 @@ firrtl.circuit "Test" {
     %foo_o = firrtl.instance foo @Foo(out o : !firrtl.probe<uint<1>, @A>)
     %x = firrtl.ref.resolve %foo_o : !firrtl.probe<uint<1>, @A>
   }
+
+  //===--------------------------------------------------------------------===//
+  // Inline Layers
+  //===--------------------------------------------------------------------===//
+
+  // CHECK:  sv.macro.decl @Inline_0["Inline"]
```
seldridge, line 358: Use of `Inline_0` isn't great as it encodes an assumption about mangling. This would be better as either a capture or as a unique name.

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -350,6 +350,75 @@ firrtl.circuit "Test" {
...
+  firrtl.layer @Inline inline {
+    // CHECK:  sv.macro.decl @Inline_Inline
+    firrtl.layer @Inline inline {}
+    firrtl.layer @Bound bind {
+    // CHECK:  sv.macro.decl @Inline_Bound_Inline
```
seldridge, line 363: Nit (here and after): indentation.

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -350,6 +350,75 @@ firrtl.circuit "Test" {
...
+  // CHECK:   }
+  // CHECK: }
+
+  // CHECK: firrtl.module @ModuleWithInlineLayerBlocks() {
+  // CHECK:   sv.ifdef  @Inline_0 {
+  // CHECK:     %w1 = firrtl.wire : !firrtl.uint<1>
+  // CHECK:     sv.ifdef  @Inline_Inline {
+  // CHECK:       %w2 = firrtl.wire : !firrtl.uint<2>
+  // CHECK:     }
+  // CHECK:     firrtl.instance inline_bound {lowerToBind, output_file = #hw.output_file<"layers_Test_Inline_Bound.sv", excludeFromFileList>} @ModuleWithInlineLayerBlocks_Inline_Bound()
```
seldridge, line 397: The ABI of the FIRRTL spec should be updated to match this file name. Note: I think that including the inline layer in the name is fine. The alternative of not including it seems less fine. Regardless, this should be rationalized.

seldridge (approved):
Looks great!

My comments are mostly around use of `walk` to future-proof this for situations where regions may show up. This will already not do the right thing for `when` blocks even though whens are supposed to be removed at this point. Regardless, use of `walk` doesn't hurt anything.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -623,6 +658,35 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+    auto macNameAttr = StringAttr();
+    if (macName != symName)
+      macNameAttr = StringAttr::get(ctx, macName);
+
+    b.create<sv::MacroDeclOp>(layer->getLoc(), symNameAttr, ArrayAttr(),
+                              macNameAttr);
+    macroNames[layer] = FlatSymbolRefAttr::get(&getContext(), symNameAttr);
+  }
+  for (auto child : layer.getOps<LayerOp>())
+    createMacroDecls(ns, b, child, stack);
```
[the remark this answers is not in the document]
seldridge: As pointed out to me elsewhere, this is confusing layers and layer blocks. When's can't appear here.

There may be a refactoring here that is cleaner that _does_ use a walk as opposed to doing manual filtering. However, that is a stylistic concern.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -623,6 +658,35 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+  stack.pop_back();
+}
+
+void LowerLayersPass::createMacroDecls(CircuitNamespace &ns, OpBuilder &b,
+                                       LayerOp layer) {
+  SmallVector<StringAttr> stack;
+  createMacroDecls(ns, b, layer, stack);
+}
```
seldridge, line 688: This function has some strict expectations on the insertion point. Also, it has one usage where the `OpBuilder` is constructed and passed in. This could save an argument and avoid the strict expectations by creating the `OpBuilder` insider this function.

```suggestion
void LowerLayersPass::createMacroDecls(CircuitNamespace &ns, LayerOp layer) {
  OpBuilder b(layer);
  SmallVector<StringAttr> stack;
  createMacroDecls(ns, b, layer, stack);
}
```

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -279,6 +296,15 @@ InnerRefMap LowerLayersPass::runOnModuleLike(FModuleLike moduleLike) {
   return innerRefMap;
 }
 
+void LowerLayersPass::lowerInlineLayerBlock(LayerOp layer,
+                                            LayerBlockOp layerBlock) {
+  OpBuilder builder(layerBlock);
+  auto macroName = macroNames[layer];
+  auto ifDef = builder.create<sv::IfDefOp>(layerBlock.getLoc(), macroName);
+  ifDef.getBodyRegion().takeBody(layerBlock.getBodyRegion());
+  layerBlock.erase();
```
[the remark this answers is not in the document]
seldridge: I believe this is safe _even if it doesn't advance_ because this is walking post-order. See: https://mlir.llvm.org/doxygen/classmlir_1_1Operation.html#a59740592240b950b8c8afcf4a2eb4113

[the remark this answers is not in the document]
dtzSiFive: Sure, I'm familiar and yep you're right (and I did miss that detail, whoops!).  Anyway the point is this is fine even if it's structured such that the legality isn't locally evident.

dtzSiFive:
Woohoo!! Great work! 🚀 🎉 🎉

PR #7325, opened 2024-07-16
[Verif] Introduce Formal Contracts

dobios:
This PR proposes a new set of ops aimed at enabling the use of formal contracts to abstract away module instances during verification. This is described in [this proposal](https://docs.google.com/document/d/1Ez6YosgVYuMtDg1MvBL8u9AMmntMARXhXofn_BcXlA4/edit?usp=sharing).   
  
There is still some work needed to handle these ops in PrepareForFormal, which will be done in a later PR.

Depends on #7145 .


include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,146 @@ def YieldOp : VerifOp<"yield", [
...
+
+//===----------------------------------------------------------------------===//
+// Formal Contract Ops
+//===----------------------------------------------------------------------===//
+
+def ContractOp : VerifOp<"contract", [Pure]>{
+  let summary = "declare formal contract";
+  let description = [{
+    This operation declares a sformal contract that is used to create precondition 
+    and postconditions on a parent module. These are used as an abstraction to
```
TaoBi22, line 360: Do we want a HasParent trait to ensure this is within a `hw.module`?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,146 @@ def YieldOp : VerifOp<"yield", [
...
+    Preconditions are assumed during the verification of a module and 
+    asserted during the verification of an instance of a module.
+  }];
+  let arguments = (ins LTLAnyPropertyType:$property,
+                      ClockEdgeAttr:$edge, I1:$clock,
+                      Optional<I1>:$enable,
+                      OptionalAttr<StrAttr>:$label);
+
+  let assemblyFormat = [{
+    $property (`if` $enable^)? `,` $edge $clock
```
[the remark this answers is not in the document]
TaoBi22: Similarly, is the clock attribute to allow the conditions to only apply on a given clock edge? If so, do we know how this lowers to our backends (e.g. can btor2 encode checking a property on a clock negedge?)

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,146 @@ def YieldOp : VerifOp<"yield", [
...
+// Formal Contract Ops
+//===----------------------------------------------------------------------===//
+
+def ContractOp : VerifOp<"contract", [Pure]>{
+  let summary = "declare formal contract";
+  let description = [{
+    This operation declares a sformal contract that is used to create precondition 
+    and postconditions on a parent module. These are used as an abstraction to 
+    better modularize formal verification such that each module containing a contract 
+    is checked exactly once.
```
TaoBi22, line 362: If a user specifies insufficient postconditions, do we allow whatever verification tool this ends up in to refine its model by exploring the module itself rather than just the post-conditions? Or do we strictly forbid this and just fail to verify?

TaoBi22 (commented):
Very excited to see this happening! :tada:

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +262,146 @@ def YieldOp : VerifOp<"yield", [
...
+def BMC : I32EnumAttrCase<"BMC", 0, "bmc">;
+/// KIP -- short for k-Inductive-Proof.
+def KInductiveProof : I32EnumAttrCase<"KInductiveProof", 1, "kinduction">;
+
+def FormalTestTypeAttr : I32EnumAttr<"FormalTestType", "formal test type",
+                                [BMC, KInductiveProof]> {
+  let cppNamespace = "circt::verif";
+}
+
+def FormalOp : VerifOp<"formal", [NoTerminator, RegionKindInterface]> {
```
uenoku, line 279: Do we need `IsolatedFromAbove` and `SymbolOpInterface`? Or do we define Formal op within HWModule?

include/circt/Dialect/Verif/VerifOps.td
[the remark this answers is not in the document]
dobios: the idea with using enable signals and clocks is that these should be able to support SVA properties, which are inherently invalid before the a given state has been reached, and this needs to be specified somewhere, e.g. `a |=> b` is invalid until the system has been reset at least once since a and b both have garbage values at that point, and that condition needs to be defined somewhere. 

Basically I want to give the same modeling power as writing assertions and assumptions by hand

include/circt/Dialect/Verif/VerifOps.td
dobios: I would say no, the guarantees we give are supposed to be "the module is correct wrt its specification" and not just "the module is correct". So if the user chooses to give a contract, then it is up to them to make it sufficiently tight to be a good abstraction of their system.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,197 @@ def YieldOp : VerifOp<"yield", [
...
+    $init `,` $update attr-dict `:` 
+    functional-type(operands, results)
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// Formal Contract Ops
+//===----------------------------------------------------------------------===//
+
+def ContractOp : VerifOp<"contract", [Pure, HasParent<"hw::HWModuleOp">]>{
```
maerhart, line 351: In the ops current form, the user can use arbitrary SSA values from the hw module body in this op which means the contract and design are not entirely separated. Did you consider isolating this op from above and providing the module ports as block arguments within the region? Additionally, the contract could live next to a module instead of inside of it which could be useful for external modules, but not sure if that's needed. Just wondering if you thought about this?

include/circt/Dialect/Verif/VerifOps.td
maerhart: I agree with @dobios take there.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,197 @@ def YieldOp : VerifOp<"yield", [
...
+    ```
+  }];
+  let regions = (region SizedRegion<1>:$contract_region);
+
+  let assemblyFormat = [{
+    attr-dict-with-keyword $contract_region
+  }];
+}
+
+def RequireOp : VerifOp<"require", [
```
maerhart, line 402: Are require and ensure structurally equivalent with assume and assert? If yes, we could create a common base class in tablegen.

dobios: Yes they are, they are currently just an op that describes "something that's sometimes an assert and other times an assume", I can refactor the tablegen a bit

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,197 @@ def YieldOp : VerifOp<"yield", [
...
+    %c0_8 = hw.constant 0 : i8 
+    %prec = comb.icmp bin neq %c0_8, %bar : i8
+    verif.assert %precond : i1
+
+    %out0 = verif.symbolic_input : i8
+    %out1 = verif.symbolic_input : i8
+    %post = comb.icmp bin ugt %out0, %bar : i8
+    %post1 = comb.icmp bin ult %out1, %bar : i8
+    verif.assume %post : i1
+    verif.assume %post1 : i1
```
TaoBi22, line 392: I think you swapped %foo for %bar between here and the contract above?
```suggestion
    %c0_8 = hw.constant 0 : i8 
    %prec = comb.icmp bin neq %c0_8, %foo : i8
    verif.assert %precond : i1

    %out0 = verif.symbolic_input : i8
    %out1 = verif.symbolic_input : i8
    %post = comb.icmp bin ugt %out0, %foo : i8
    %post1 = comb.icmp bin ult %out1, %foo : i8
    verif.assume %post : i1
    verif.assume %post1 : i1
```

Also, I'm guessing %out0 and %out1 replace all uses of %bar.0 and %bar.1? It could be clearer to just name them %bar.0 and %bar.1 in the after PrepareForFormal section, as that implies the replacement?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,197 @@ def YieldOp : VerifOp<"yield", [
...
+  }];
+}
+
+def ResultOp : VerifOp<"result", [HasParent<"verif::ResultOp">]>{
+  let summary = "define a symbolic reference to the outputs of the module";
+  let description = [{
+    This operation defines a set of symbolic references to each output of the 
+    module for which this contract is defined. This means that a single reference 
+    of result will result in `#outputs` SSA values which refer to each output of 
+    the module. This is needed for when the instance of the module will be replaced
```
TaoBi22, line 448: This might be slightly easier to read:
```suggestion
    This operation defines a set of symbolic references to each output of the 
    module within which this contract is defined. This means that a single 
    reference of result will result in as many SSA values as the module has 
    outputs, with each result corresponding to one output. This is needed 
    for when the instance of the module will be replaced 
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -260,4 +263,197 @@ def YieldOp : VerifOp<"yield", [
...
+                      Optional<I1>:$enable,
+                      OptionalAttr<StrAttr>:$label);
+
+  let assemblyFormat = [{
+    $property (`if` $enable^)? (`,` $edge^ $clock)? 
+    (`label` $label^)? attr-dict `:` type($property)
+  }];
+}
+
+def ResultOp : VerifOp<"result", [HasParent<"verif::ResultOp">]>{
```
uenoku, line 442: `HasParent<"verif::ResultOp">` may be incorrect?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -345,4 +350,164 @@ def ConcreteInputOp : VerifOp<"concrete_input", [
...
+    e.g. 
+    ```
+    hw.module @Bar(in %clk : !seq.clock, in %foo : i8, out "" : i8, out "1" : i8) { 
+      verif.contract (%clk, %foo) : (i8, i8) {
+        ^bb0(%clk : !seq.clock, %foo : i8): 
+          %c0_8 = hw.constant 0 : i8 
+          %prec = comb.icmp bin neq %c0_8, %foo : i8
+          verif.require %precond : i1
+
+          %out0, %out1 = verif.result : i8
```
fabianschuiki, line 404: I don't think you need a `verif.result` operation. Instead, you could just use the actual SSA values of the module output in the contract here as well. (Instead of `%out0` and `%out1`, use the values passed to the module's `hw.output` operation directly.)

The problem with using something like `verif.result` is that it is basically an SSA edge where your contract gains access to the SSA values of the module outputs, but this dependency edge is not captured in the SSA graph. If an optimization were to remove an output port from the module (because it is unused, or similar), that optimization would have no idea that it would be breaking the `verif.result` op in the contract. However, if you directly use the module output's SSA value, all the stuff in the contract would actually be visible to the optimization as a dependency. Furthermore, if an output got modified or dropped, the optimization pass would still be able to update the uses in your contract such that it would be a non-breaking change.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -345,4 +350,164 @@ def ConcreteInputOp : VerifOp<"concrete_input", [
...
+        verif.assert %precond : i1
+
+        %bar.0 = verif.symbolic_input : i8
+        %bar.1 = verif.symbolic_input : i8
+        %post = comb.icmp bin ugt %bar.0, %foo : i8
+        %post1 = comb.icmp bin ult %bar.1, %foo : i8
+        verif.assume %post : i1
+        verif.assume %post1 : i1
+        verif.yield %bar.0, %bar.1 : i8, i8
+    }
```
fabianschuiki, line 433: I'm wondering if you need a `verif.instance` op at all. Could you instead just inline the contents of the contract? I think it only makes sense to add an operation if a pass is going to use that operation later. At first glance, `verif.instance` looks like it migth just be inlined during the formal checking of the contract. Or are there cases where you'd ignore the contents of `verif.instance`, or treat them differently than in other cases?

If the formal contract checking more or less inlines the contents of `verif.instance`, and it has to deal with things like `verif.symbolic_input` within the context of the surrounding module anyway, then you could simplify the IR by inlining everything directly and allowing `verif.symbolic_input` to be used anywhere in the module. WDYT?

dobios: Given that symbolic inputs really only make sense in the context of verification, I really want to avoid having them be usable outside of explicit verification contexts. This will guarantee that they are always used legally in a way that doesn't require expensive verification logic. It is also a neat way to encapsulate the contract in a way that keeps the transformed IR similar to the base IR, thus making it easier to debug -- i.e. it is obvious that a verif.instance is the transformed result of the now missing hw.instance.

include/circt/Dialect/Verif/VerifOps.td
maerhart: Couldn't we add block arguments representing the outputs? That would simplify the operation verifier, make it easier for transformation passes to change the module signature, and you'd not have to define another op.

dobios: that could work yeah

dobios: After playing around with this idea for a bit I think that it would make the subsequent transformation more difficult than it needs to be, because I would need to look at the rest of the module's internal logic to figure out which of my block arguments are outputs and which are inputs, so I honestly think that having a `verif.result` would be the easiest solution

maerhart: I think you should preserve the same order of arguments as the parent modules `!hw.modty` attribute. Then you just need to iterate over this attribute's port list to know if it's an output, input, or inout.

test/Dialect/Verif/basic.mlir
```diff
@@ -53,9 +53,69 @@ verif.formal @formal1(k = 20) {
...
   // CHECK: %foo.0, %foo.1 = hw.instance "foo" @Foo("0": %6: i1, "1": %8: i1) -> ("": i1, "1": i1)
-  %foo.0, %foo.1 = hw.instance "foo" @Foo("0": %sym: i1, "1": %8 : i1)  -> ("" : i1, "1" : i1)
+  %foo.0, %foo.1 = hw.instance "foo" @Foo("0": %sym: i1, "1": %8 : i1) -> ("" : i1, "1" : i1)
 }
 
+hw.module @Bar(in %foo : i8, out "" : i8, out "1" : i8) { 
+  // CHECK: verif.contract(%foo) : (i8) {
+  verif.contract (%foo) : (i8) {
+    // CHECK: ^bb0(%[[ARG:.+]]: i8, %[[OUT:.+]]: i8, %[[OUT1:.+]]: i8):
+    ^bb0(%arg1 : i8, %bar.0 : i8, %bar.1 : i8):
```
uenoku, line 63: Maybe i'm missing something but what do the second and third block argument mean?

uenoku: Ah ok I get it. Could you clarify that arguments correspond to output values in tablegen?

test/Dialect/Verif/basic.mlir
```diff
@@ -53,9 +53,69 @@ verif.formal @formal1(k = 20) {
...
+      verif.ensure %post : i1
+      // CHECK: verif.ensure %[[P1]] : i1
+      verif.ensure %post1 : i1
+  }
+  %c1_8 = hw.constant 1 : i8
+  %o0 = comb.add bin %foo, %c1_8 : i8
+  %o1 = comb.sub bin %foo, %c1_8 : i8
+  hw.output %o0, %o1 : i8, i8
+}
+
```
uenoku, line 85: nit: Please add `CHECK-LABEL`.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -345,4 +350,146 @@ def ConcreteInputOp : VerifOp<"concrete_input", [
...
+    This operation declares a formal contract that is used to create precondition 
+    and postconditions on a parent module. These are used as an abstraction to 
+    better modularize formal verification such that each module containing a contract 
+    is checked exactly once.
+          
+    e.g. 
+    ```
+    hw.module @Bar(in %clk : !seq.clock, in %foo : i8, out "" : i8, out "1" : i8) { 
+      verif.contract (%clk, %foo) : (i8, i8) {
+        ^bb0(%clk : !seq.clock, %foo : i8):
```
uenoku, line 399: nit: please update IR in the description.

uenoku (approved):
LGTM!

PR #7326, opened 2024-07-16
[Moore] Add MergeExtractRef pass.

hailongSun2000:
This pass is used to merge `moore.extract_ref` together in the little-endian without
existing in `if/for` statements and nested `moore.extract_ref`.
For example:
```
bit [31:0] arr;
assign arr[23:16] = 8'd5;
always_comb begin
arr[15:8] = 8'd10;
end
```
After running this pass
```
bit [31:0] arr;
...
always_comb begin
...
arr[15:8] = 8'd10;
%concat = {arr[31:24], 8'd5, 8'd10, arr[7:0]} // concat
arr = %concat
end
```
However, cannot handle the complex situations of missing some middle bits, such as:
```
bit [3:0] arr;
arr[3:2] = 2'b0;
arr[0] = 1'b0;
```
`arr[1]` don't be assigned a value, this situation belongs to missing some middle bits.

And if the same slice is assigned a value repeatedly, will emit an error and interrupt.

### DIFF
--- a/lib/Dialect/Moore/Transforms/MergeExtractRef.cpp
+++ b/lib/Dialect/Moore/Transforms/MergeExtractRef.cpp
@@ -1,0 +1,184 @@
+//===- MergeExtractRef.cpp - Merge moore.extract_ref ----------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file defines the MergeExtractRef pass.
+// It's used to merge moore.extract_ref together in the big-endian order without
+// existing in if/for statements and nested moore.extract_ref.
+// For example:
+// ```
+// bit [31:0] arr;
+// assign arr[23:16] = 8'd5;
+// always_comb begin
+// arr[15:8] = 8'd10;
+// end
+// ```
+// After running this pass:
+// ```
+// bit [31:0] arr;
+// ...
+// always_comb begin
+// ...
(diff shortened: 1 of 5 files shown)


lib/Dialect/Moore/Transforms/MergeExtractRef.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    } else {
+      // TODO:
+      // Like 4-bits arr, arr[3:2] = 2'b0; arr[0] = 1'b0;. "arr[1]" don't be
+      // assigned a value, this situation belongs to missing some middle bits.
+      emitError(loc)
+          << "Unsupported the complex situations of missing some middle bits ";
+      return;
+    }
+    builder.create<VariableOp>(loc, type, name, concat);
```
hailongSun2000, line 182: There should be `blocking_assign` or `nonblocking_assign`.

lib/Dialect/Moore/Transforms/MergeExtractRef.cpp
```diff
@@ -0,0 +1,184 @@
...
+    auto width = cast<RefType>(type).getBitSize().value();
+    auto domain = cast<RefType>(type).getDomain();
+    auto loc = extrRef.getLoc();
+
+    // Collect and then concat all values together for the same variable.
+    SmallVector<Value> values;
+    for (const auto &indexedSrc : indexedSrcs)
+      values.push_back(indexedSrc.src);
+
+    builder.setInsertionPointAfterValue(indexedSrcs.back().src);
```
hailongSun2000, line 141: Maybe set the insertion point at the end of the body of the procedure.

hailongSun2000:
Maybe this pass should be removed, but I think some thought of this pass can be applied when after executing 
> @fabianschuiki `combine non-blocking assignments to the same variable` (https://github.com/llvm/circt/pull/7075) .

Then analyze and map to a `moore.register` op.

Issue #7327, opened 2024-07-16
[FIRRTL] Crash in LowerClasses related to StringMap buckets


mikeurbach:
I just observed this in an internal design:

```

[2024-07-16T19:06:09.356Z] Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
[2024-07-16T19:06:09.356Z] 0  firtool         0x00000000006ddac7 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 39
[2024-07-16T19:06:09.356Z] 1  firtool         0x00000000006db87e llvm::sys::RunSignalHandlers() + 238
[2024-07-16T19:06:09.356Z] 2  firtool         0x00000000006de3ef
[2024-07-16T19:06:09.356Z] 3  libpthread.so.0 0x00007f794e6f0cf0
[2024-07-16T19:06:09.356Z] 4  firtool         0x0000000000699053 llvm::StringMapImpl::LookupBucketFor(llvm::StringRef, unsigned int) + 307
[2024-07-16T19:06:09.356Z] 5  firtool         0x000000000075bddc
[2024-07-16T19:06:09.356Z] 6  firtool         0x000000000075ed5e
[2024-07-16T19:06:09.356Z] 7  firtool         0x0000000000e04076 circt::firrtl::getOrAddInnerSym(mlir::MLIRContext*, circt::hw::InnerSymAttr, unsigned long, llvm::function_ref<circt::hw::InnerSymbolNamespace& ()>) + 198
[2024-07-16T19:06:09.356Z] 8  firtool         0x0000000000e044a3 circt::firrtl::getOrAddInnerSym(circt::hw::InnerSymTarget const&, llvm::function_ref<circt::hw::InnerSymbolNamespace& ()>) + 131
[2024-07-16T19:06:09.356Z] 9  firtool         0x0000000000e04654 circt::firrtl::getOrAddInnerSym(circt::hw::InnerSymTarget const&, llvm::function_ref<circt::hw::InnerSymbolNamespace& (circt::firrtl::FModuleLike)>) + 292
[2024-07-16T19:06:09.356Z] 10 firtool         0x0000000000e04707 circt::firrtl::getInnerRefTo(circt::hw::InnerSymTarget const&, llvm::function_ref<circt::hw::InnerSymbolNamespace& (circt::firrtl::FModuleLike)>) + 151
[...]

PR #7328, opened 2024-07-16
Reland "[FIRRTL][LowerClass] Pre-allocate namespaces before capturing refs (#7102)".

mikeurbach:
This was originally added to avoid a use-after-free. However, it got removed as part of f4920aa. It seems this is still required, even when we process PathTrackers sequentially. I confirmed there was a use-after-free without this change, which is fixed after adding this back.

Closes https://github.com/llvm/circt/issues/7327.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -302,6 +302,12 @@ PathTracker::run(CircuitOp circuit, InstanceGraph &instanceGraph,
                  const DenseMap<DistinctAttr, FModuleOp> &owningModules) {
   SmallVector<PathTracker> trackers;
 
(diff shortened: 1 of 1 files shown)


dtzSiFive (approved):
LGTM, especially if its fixes the issue found.

Doing this lazily is nice, although perhaps as we lean more on these things we're likely to need this for all modules anyway.

I'd like to root this out a bit more thoroughly (is there a move constructor missing? maybe the collection uses unique_ptr's? or just a comment?) but this looks correct and reliable way to fix this with what we have now, thanks!

mikeurbach:
Yeah, I will say that I haven't yet dug more deeply into this, just confirmed that @uenoku's original fix here does fix the issue we see in single threaded PathTracker processing. I'll go ahead and merge this for now so main is clean.

mikeurbach:
Opened the linked ticket to continue the investigation: https://github.com/llvm/circt/issues/7329

PR #7333, opened 2024-07-17
[ExportVerilog] Add "context" to imported DPI functions

uenoku:
This commit adds "context" property to DPI import statements to enable us to call exported functions from DPI functions. Ideally it's better to create an attribute to specify the property "context" or "pure" but "context" by default should be ok for now.


fabianschuiki (approved):
LGTM! If I recall correctly, leaving `context` away is a performance optimization that can make some DPI calls faster, but having `context` should always be correct. So this feels like the right, conservative thing to do.

uenoku:
Yes, removing `context` improves simulation performances so eventually it would be nice to remove `context` by opt-in.

PR #7334, opened 2024-07-17
[ImportVerilog] Accept empty packages

fabianschuiki:
Add handling for package definitions and accept empty packages for the time being. Later on we'll want to support parameters, variables, functions, and other things that can appear in packages. These require a mechanism to refer to things outside of an `SVModuleOp` though, which does not yet exist.


hailongSun2000 (approved):
LGTM! :partying_face:

PR #7336, opened 2024-07-17
[ImportVerilog] Fix unbased unsized literals of packed aggregate type

fabianschuiki:
Fix an issue where ImportVerilog would currently crash if an unbased, unsized literal like `'0` is used to initialize or assign to a packed struct. Instead, the integer literal should be constructed and then converted to the packed aggregate type. In a later pass, this conversion can be lowered to a more careful `moore.struct_create`.


maerhart (approved):
LGTM

mingzheTerapines:
@fabianschuiki Good idea, may I do the lower job for `variabelOp` lowering to `structCreateOp` by pass canonicalize like `assignedVarOp` does.

PR #7337, opened 2024-07-17
[ImportVerilog] Handle $signed/$unsigned system tasks

fabianschuiki:
Handle calls to the `$signed` and `$unsigned` system tasks by simply passing through the argument. The casting is handled by Slang during type checking and is materialized as a type conversion node in the AST. The call itself has no function after that.

The `visitCall` function is going to be the place to add handling for other system calls in the future. IEEE 1800-2017 section 20 defines a fairly substantial list of builtin system tasks. Most of these will want to have a dedicated `moore.builtin.*` op in the future.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "ImportVerilogInternals.h"
+#include "slang/ast/SystemSubroutine.h"
 #include "slang/syntax/AllSyntax.h"
 
 using namespace circt;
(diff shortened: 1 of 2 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -580,6 +581,38 @@ struct RvalueExprVisitor {
...
+  Value visitCall(const slang::ast::CallExpression &expr,
+                  const slang::ast::CallExpression::SystemCallInfo &info) {
+    const auto &subroutine = *info.subroutine;
+
+    if (subroutine.name == "$signed" || subroutine.name == "$unsigned")
+      return context.convertRvalueExpression(*expr.arguments()[0]);
+
+    mlir::emitError(loc) << "unsupported system call `" << subroutine.name
+                         << "`";
+    return {};
```
hailongSun2000, line 613: I viewed the source code in slang. I understood that `SubroutineSymbol*` is used to handle the functions defined by users like `function ... endfunction`, but the `SystemCallInfo` handles the system functions like `$signed`, `$display`, etc, right?

hailongSun2000 (approved):
I have an unknown knowledge. Other than that, it's very nice! :+1:

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Yes exactly! 🙂

hailongSun2000: Oh! I learn it! :smile:

uenoku (approved):
LGTM

PR #7339, opened 2024-07-17
[ImportVerilog] Allow enum variants in expressions

fabianschuiki:
Add support for enum variants in expressions. Until now, ImportVerilog would emit a "unknown name" diagnostic when an enum variant is used in an expression. This is due to the lowering of named value expressions simply checking whether an MLIR value is present in the `valueSymbols` table for the referenced symbol name. This works for variables which create dedicated declarations, but not for enum variants which potentially live outside the current module's scope.

Instead, ImportVerilog will now check if the named value is a constant, for example a parameter or an enum variant, and materialize that constant directly as a `moore.constant` op. This will also be able to deal with parameters declared inside of packages, for which the table of value symbols of the current module does not contain any value. (Local parameters inside the module generate a `moore.named_constant` op that is present in the table.)

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -60,13 +60,26 @@ struct RvalueExprVisitor {
 
   // Handle named values, such as references to declared variables.
   Value visit(const slang::ast::NamedValueExpression &expr) {
-    if (auto value = context.valueSymbols.lookup(&expr.symbol))
-      return isa<moore::NamedConstantOp>(value.getDefiningOp())
-                 ? value
-                 : builder.create<moore::ReadOp>(
-                       loc,
-                       cast<moore::RefType>(value.getType()).getNestedType(),
-                       value);
+    if (auto value = context.valueSymbols.lookup(&expr.symbol)) {
+      if (auto refType = dyn_cast<moore::RefType>(value.getType()))
+        value =
+            builder.create<moore::ReadOp>(loc, refType.getNestedType(), value);
+      return value;
+    }
+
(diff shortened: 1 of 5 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -60,13 +60,26 @@ struct RvalueExprVisitor {
...
-                       cast<moore::RefType>(value.getType()).getNestedType(),
-                       value);
+    if (auto value = context.valueSymbols.lookup(&expr.symbol)) {
+      if (auto refType = dyn_cast<moore::RefType>(value.getType()))
+        value =
+            builder.create<moore::ReadOp>(loc, refType.getNestedType(), value);
+      return value;
+    }
+
+    // Try to materialize constant values directly.
```
hailongSun2000, line 70: I noticed that you mentioned 
> for example a parameter or an enum variant, and materialize that constant directly as a moore.constant op.

The `Parameter` is still stored in the `ValueSymbol`, then return it(`namedConstant`), rather than its constant value.

hailongSun2000: As if I misunderstand again, that is materializing the constant directly as a `moore.constant`. Sorry that! :sweat_smile:

hailongSun2000 (approved):

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Yeah this is basically a fall-back for when constants are used that have not been lowered into a `moore.named_constant`. But if there is a `named_constant` available, this code will still prefer that :smiley:

PR #7340, opened 2024-07-18
[Moore] Distinguish the dynamic and constant extract.

hailongSun2000:
We divide the extract into a dynamic named `moore.dyn_extract` and a constant extract named `moore.extract`, the former has a variable as its low bit like `arr[x]`, but the latter has a constant value as its low bit like `arr[1]`.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -440,24 +440,47 @@ struct RvalueExprVisitor {
   Value visit(const slang::ast::ElementSelectExpression &expr) {
     auto type = context.convertType(*expr.type);
     auto value = context.convertRvalueExpression(expr.value());
-    auto lowBit = context.convertRvalueExpression(expr.selector());
-
-    if (!type || !value || !lowBit)
+    if (!type || !value)
       return {};
-    return builder.create<moore::ExtractOp>(loc, type, value, lowBit);
+    if (auto *constValue = expr.selector().constant) {
+      assert(!constValue->hasUnknown());
+      assert(constValue->size() <= 32);
+
+      auto lowBit = constValue->integer().as<uint32_t>().value();
+      return builder.create<moore::ExtractOp>(loc, type, value, lowBit);
+    }
+    auto lowBit = context.convertRvalueExpression(expr.selector());
+    if (!lowBit)
+      return {};
+    return builder.create<moore::DynExtractOp>(loc, type, value, lowBit);
   }
 
   // Handle range bits selections.
   Value visit(const slang::ast::RangeSelectExpression &expr) {
     auto type = context.convertType(*expr.type);
     auto value = context.convertRvalueExpression(expr.value());
-    Value lowBit;
+    if (!type || !value)
+      return {};
+
+    Value dynLowBit;
+    uint32_t constLowBit;
+    auto *leftConst = expr.left().constant;
+    auto *rightConst = expr.right().constant;
+    if (leftConst) {
+      assert(!leftConst->hasUnknown());
+      assert(leftConst->size() <= 32);
+    }
+    if (rightConst) {
+      assert(!rightConst->hasUnknown());
+      assert(rightConst->size() <= 32);
+    }
+
     if (expr.getSelectionKind() == slang::ast::RangeSelectionKind::Simple) {
-      if (expr.left().constant && expr.right().constant) {
-        auto lhs = expr.left().constant->integer().as<uint64_t>().value();
-        auto rhs = expr.right().constant->integer().as<uint64_t>().value();
-        lowBit = lhs < rhs ? context.convertRvalueExpression(expr.left())
-                           : context.convertRvalueExpression(expr.right());
+      if (leftConst && rightConst) {
+        // Estimate whether is big endian or little endian.
+        auto lhs = leftConst->integer().as<uint32_t>().value();
+        auto rhs = rightConst->integer().as<uint32_t>().value();
+        constLowBit = lhs < rhs ? lhs : rhs;
       } else {
         mlir::emitError(loc, "unsupported a variable as the index in the")
             << slang::ast::toString(expr.getSelectionKind()) << "kind";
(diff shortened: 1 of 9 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -427,24 +427,35 @@ struct RvalueExprVisitor {
   Value visit(const slang::ast::ElementSelectExpression &expr) {
     auto type = context.convertType(*expr.type);
     auto value = context.convertRvalueExpression(expr.value());
+    if (!type || !value)
+      return {};
+    if (auto *constValue = expr.selector().constant) {
+      auto lowBit = constValue->integer().as<uint32_t>().value();
+      return builder.create<moore::ExtractOp>(loc, type, value, lowBit);
+    }
```
fabianschuiki, line 435: Can you add a check that the `constValue` fits into 32 bits, and only pick this path if that's the case? Otherwise we can still lower to the dynamic extract. Just to make sure that we don't silently overflow or fail if somebody types `arr[123456789123456789123456789123456789]` in the SystemVerilog input :grin:

fabianschuiki: Also, you probably want to check that `constValue->isUnknown()` is false, meaning that the value does not have any X or Z in them. Otherwise we would silently drop X or Z when converting to the `lowBit` integer value.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -427,24 +427,35 @@ struct RvalueExprVisitor {
...
-      if (expr.left().constant && expr.right().constant) {
-        auto lhs = expr.left().constant->integer().as<uint64_t>().value();
-        auto rhs = expr.right().constant->integer().as<uint64_t>().value();
-        lowBit = lhs < rhs ? context.convertRvalueExpression(expr.left())
-                           : context.convertRvalueExpression(expr.right());
+      if (leftConst && rightConst) {
+        // Estimate whether is big endian or little endian.
+        auto lhs = leftConst->integer().as<uint32_t>().value();
+        auto rhs = rightConst->integer().as<uint32_t>().value();
+        constLowBit = lhs < rhs ? lhs : rhs;
```
fabianschuiki, line 458: Can you add the same check here, to make sure that we at least error out if the constant is too large to fit into 32 bits?

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -286,19 +286,19 @@ struct ModuleVisitor : public BaseVisitor {
...
         unsigned offset = 0;
-        auto i32 = moore::IntType::getInt(context.getContext(), 32);
+        // auto i32 = moore::IntType::getInt(context.getContext(), 32);
         for (const auto *port : llvm::reverse(multiPort->ports)) {
           if (auto *existingPort = moduleLowering->portsBySyntaxNode.lookup(
                   con->port.getSyntax()))
             port = existingPort;
           unsigned width = port->getType().getBitWidth();
-          auto index = builder.create<moore::ConstantOp>(loc, i32, offset);
+          // auto index = builder.create<moore::ConstantOp>(loc, i32, offset);
```
fabianschuiki, line 295: No longer needed :smiley:?

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
```diff
@@ -64,11 +64,8 @@ struct ConcatRefLowering : public OpConversionPattern<OpTy> {
       // small or vice versa. Like "logic [7:0] or [0:7]".
 
       // Only able to correctly handle the situation like "[7:0]" now.
-      auto i32 = moore::IntType::getInt(op.getContext(), 32);
-      auto lowBit =
-          rewriter.create<ConstantOp>(op.getLoc(), i32, srcWidth - width);
-      auto extract =
-          rewriter.create<ExtractOp>(op.getLoc(), type, op.getSrc(), lowBit);
+      auto extract = rewriter.create<ExtractOp>(op.getLoc(), type, op.getSrc(),
+                                                srcWidth - width);
```
fabianschuiki, line 68: Cool! Looks like this helps make things simpler in a lot of cases :smiley:

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: > Also, you probably want to check that `constValue->isUnknown()` is false, meaning that the value does not have any X or Z in them. Otherwise we would silently drop X or Z when converting to the `lowBit` integer value.

You're totally right! I'll add this check.

hailongSun2000: > Can you add a check that the `constValue` fits into 32 bits, and only pick this path if that's the case? Otherwise we can still lower to the dynamic extract. Just to make sure that we don't silently overflow or fail if somebody types `arr[123456789123456789123456789123456789]` in the SystemVerilog input 😁

slang will check the bit width and overflow like 
```
source.sv:2:6: warning: signed integer literal overflows 32 bits, will be truncated to -1097262461 [-Wint-overflow]
bit [123456789123:0] arr;
     ^
source.sv:2:6: error: packed type of 1097262462 bits is too large; maximum width is 16777215 bits
bit [123456789123:0] arr;
```

lib/Conversion/ImportVerilog/Structure.cpp
hailongSun2000: Yeah, right! Because it is used to handle the `multiport`, the number of ports must be exact. So I think we can remove these annotation codes.

lib/Dialect/Moore/Transforms/LowerConcatRef.cpp
hailongSun2000: Yeah, it can help reduce code size :smile: .

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: As above mentioned, slang will truncate the integer literal that overflows 32 bits, so we also need to add this check :thinking:?

hailongSun2000:
I want to add this constraint for `dyn_extract` op, but I failed, could @fabianschuiki teach me how to add this or where I can refer to it? Thanks so much :smile:!
```
class SigPtrIndexBitWidthConstraint<string index, string input>
  : TypesMatchWith<"Index width should be exactly clog2 (size of array)",
                   input, index,
                   [{
                     IntegerType::get($_self.getContext(), std::max<uint64_t>(
                       llvm::Log2_64_Ceil(llhd::getLLHDTypeWidth($_self)), 1))
                   }]>;
```
I know one way is to add `parse` and `print` like `constant` op to trunc the width of `lowBit`.

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: Slang is always checking the `constValue` whether is `x/z`. Like
```
source.sv:6:11: error: cannot refer to element 1'bx of 'logic[3:0]' [-Windex-oob]
    b = a[1'bx];
          ^~~~
source.sv:8:14: error: value must not have any unknown bits
    b = a[3-:1'bx];
             ^~~~
source.sv:9:10: error: value must not have any unknown bits
    a[3-:1'bz] = b;
```

fabianschuiki: Ah that's nice! Thanks a lot. I didn't realize that SV array dimensions would be limited to 32 bits by Slang. That's great though, in that case we don't really need to generate diagnostic errors in the code. Maybe just an `assert(...)` to encode our assumption about <=32bits and no X/Z in the code somewhere.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Nice find! :smile:

fabianschuiki:
@hailongSun2000 Do we need that clog2 constraint on the index somewhere? I'm asking because I think the type constraint on the index only makes sense if it ensures that no out-of-bounds elements can be accessed. However, if you have an array with 9 elements (0 to 8), you'd require a 4 bit index, which means you could still access elements 9 to 15. So it looks like we are going to have to deal with out-of-bounds accesses anyway, and so we might as well make the op easier to use by not constraining the index. What do you think?

hailongSun2000:
It makes sense. Maybe we have a `9-bit` array. I think we needn't `clog2` here.

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: I only add `assert()` for no X/Z. For `<= 32 bits`, I'll add later.

fabianschuiki (approved):
LGTM!

PR #7343, opened 2024-07-18
[circt-bmc] Add LowerToBMC Pass

TaoBi22:
This adds a pass (again based on a first draft by @maerhart) that takes a `hw.module` and converts it into a corresponding `verif.bmc` operation. Currently only supports modules with one clock.

### DIFF
--- a/include/circt/Tools/circt-bmc/Passes.td
+++ b/include/circt/Tools/circt-bmc/Passes.td
@@ -16,6 +16,28 @@
 include "mlir/Pass/PassBase.td"
 include "mlir/Rewrite/PassUtil.td"
 
+def LowerToBMC : Pass<"lower-to-bmc", "::mlir::ModuleOp"> {
+  let summary = "Lower CIRCTs core dialects to a BMC problem statement";
+  let description = [{
+  }];
+  let options = [
+    Option<"topModule", "top-module", "std::string",
+           /*default=*/"",
+           "Name of the top module to verify.">,
+    Option<"bound", "bound", "unsigned",
+           /*default=*/"",
+           "Cycle bound.">,
+    Option<"insertMainFunc", "insert-main", "bool",
+           /*default=*/"false",
+           "Whether a main function should be inserted for AOT compilation.">,
+  ];
+
+  let dependentDialects = [
+    "mlir::func::FuncDialect", "mlir::LLVM::LLVMDialect",
+    "circt::verif::VerifDialect"
+  ];
+}
+
 def ExternalizeRegisters : Pass<"externalize-registers", "::mlir::ModuleOp"> {
   let summary = "Removes registers and adds corresponding input and output ports";
 
--- a/lib/Tools/circt-bmc/LowerToBMC.cpp
+++ b/lib/Tools/circt-bmc/LowerToBMC.cpp
@@ -108,9 +108,9 @@ void LowerToBMCPass::runOnOperation() {
 
   // Double the bound given to the BMC op, as a clock cycle takes 2 BMC
   // iterations
-  verif::BMCOp bmcOp;
+  verif::BoundedModelCheckingOp bmcOp;
   if (auto numRegs = hwModule->getAttr("num_regs"))
-    bmcOp = builder.create<verif::BMCOp>(
+    bmcOp = builder.create<verif::BoundedModelCheckingOp>(
         loc, 2 * bound, cast<IntegerAttr>(numRegs).getValue().getZExtValue());
   else {
     hwModule->emitOpError(
@@ -122,14 +122,14 @@ void LowerToBMCPass::runOnOperation() {
   // Check that there's only one clock input to the module
   // TODO: supporting multiple clocks isn't too hard, an interleaving of clock
   // toggles just needs to be generated
-  std::optional<unsigned long> clkIndex;
+  bool hasClk;
   for (auto [i, input] : llvm::enumerate(hwModule.getInputTypes())) {
     if (isa<seq::ClockType>(input)) {
-      if (clkIndex) {
+      if (hasClk) {
         hwModule.emitError("Designs with multiple clocks not yet supported.");
         return signalPassFailure();
       }
-      clkIndex = i;
+      hasClk = true;
     }
   }
   {
@@ -137,7 +137,7 @@ void LowerToBMCPass::runOnOperation() {
     // Initialize clock to 0 if it exists, otherwise just yield nothing
     auto *initBlock = builder.createBlock(&bmcOp.getInit());
     builder.setInsertionPointToStart(initBlock);
-    if (clkIndex) {
+    if (hasClk) {
       auto initVal =
           builder.create<hw::ConstantOp>(loc, builder.getI1Type(), 0);
       auto toClk = builder.create<seq::ToClockOp>(loc, initVal);
@@ -149,12 +149,10 @@ void LowerToBMCPass::runOnOperation() {
     // Toggle clock in loop region if it exists, otherwise just yield nothing
     auto *loopBlock = builder.createBlock(&bmcOp.getLoop());
     builder.setInsertionPointToStart(loopBlock);
-    SmallVector<Location> locs(
-        hwModule.getBodyBlock()->getArgumentTypes().size(), loc);
-    loopBlock->addArguments(hwModule.getBodyBlock()->getArgumentTypes(), locs);
-    if (clkIndex) {
-      auto fromClk = builder.create<seq::FromClockOp>(
-          loc, loopBlock->getArgument(clkIndex.value()));
+    if (hasClk) {
+      loopBlock->addArgument(seq::ClockType::get(&getContext()), loc);
+      auto fromClk =
+          builder.create<seq::FromClockOp>(loc, loopBlock->getArgument(0));
       auto cNeg1 = builder.create<hw::ConstantOp>(loc, builder.getI1Type(), -1);
       auto nClk = builder.create<comb::XorOp>(loc, fromClk, cNeg1);
       auto toClk = builder.create<seq::ToClockOp>(loc, nClk);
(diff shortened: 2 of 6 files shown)


lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,188 @@
...
+    global = b.create<LLVM::GlobalOp>(
+        loc, arrayTy, /*isConstant=*/true, LLVM::linkage::Linkage::Private, str,
+        StringAttr::get(b.getContext(), Twine(str).concat(Twine('\00'))));
+  }
+
+  // FIXME: sanity check the fetched global: do all the attributes match what
+  // we expect?
+
+  return builder.create<LLVM::AddressOfOp>(loc, global);
+}
```
maerhart, line 62: Given that we always only call this exactly twice, we could also just create new unique globals each time and simplify this (i.e., the FIXME would also be dealt with). But also great as is for now.

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+void LowerToBMCPass::runOnOperation() {
+  Namespace names;
+
+  // Fetch the 'hw.module' operation to model check.
+  Operation *expectedModule = getOperation().lookupSymbol(topModule);
+  if (!expectedModule) {
+    getOperation().emitError("module named '") << topModule << "' not found";
```
maerhart, line 70: Can we add an error test for this message?

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,188 @@
...
+    return signalPassFailure();
+  }
+  auto hwModule = dyn_cast<hw::HWModuleOp>(expectedModule);
+  if (!hwModule) {
+    expectedModule->emitError("must be a 'hw.module'");
```
maerhart, line 75: Maybe also an error test for this?

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,188 @@
...
+    return signalPassFailure();
+  }
+
+  if (hwModule.getOps<verif::AssertOp>().empty()) {
+    hwModule.emitError("no property provided to check in module");
```
maerhart, line 80: What if an instantiated module as asserts but the top module doesn't? That would still be fine, right?

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,188 @@
...
+  builder.setInsertionPointToEnd(getOperation().getBody());
+  auto ptrTy = LLVM::LLVMPointerType::get(ctx);
+  auto voidTy = LLVM::LLVMVoidType::get(ctx);
+
+  // Lookup or declare printf function.
+  auto printfFunc =
+      LLVM::lookupOrCreateFn(getOperation(), "printf", ptrTy, voidTy, true);
+
+  // Replace the top-module with a function performing the BMC
+  Type i32Ty = builder.getI32Type();
```
maerhart, line 97: Nit: we could move this further down, just before its actual use sites.

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+  // Double the bound given to the BMC op, as a clock cycle takes 2 BMC
+  // iterations
+  verif::BoundedModelCheckingOp bmcOp;
+  if (auto numRegs = hwModule->getAttr("num_regs"))
+    bmcOp = builder.create<verif::BoundedModelCheckingOp>(
+        loc, 2 * bound, cast<IntegerAttr>(numRegs).getValue().getZExtValue());
+  else {
+    hwModule->emitOpError(
+        "No num_regs attribute found - please run externalize "
```
maerhart, line 118: An error test for this might also be nice 🙂

test/Tools/circt-bmc/lower-to-bmc.mlir
```diff
@@ -0,0 +1,67 @@
+// RUN: circt-opt --externalize-registers --lower-to-bmc="top-module=comb bound=10" %s | FileCheck %s
```
maerhart, line 1: It'd be great if we don't have to run the externalize registers pass in this regression test for lower-bmc. Maybe you could just run it manually and copy-paste the results of it into this file?

maerhart (approved):
Amazing! Thanks for pushing this forward! 🚀

include/circt/Tools/circt-bmc/Passes.td
```diff
@@ -16,6 +16,28 @@
 include "mlir/Pass/PassBase.td"
 include "mlir/Rewrite/PassUtil.td"
 
+def LowerToBMC : Pass<"lower-to-bmc", "::mlir::ModuleOp"> {
+  let summary = "Lower CIRCTs core dialects to a BMC problem statement";
+  let description = [{
+  }];
```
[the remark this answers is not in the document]
TaoBi22: Oops, good point, thanks for the catch!

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include "circt/Dialect/Verif/VerifOps.h"
+#include "circt/Support/Namespace.h"
+#include "circt/Tools/circt-bmc/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/LLVMIR/FunctionCallUtils.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/Location.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
```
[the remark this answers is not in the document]
TaoBi22: Good point, thanks! VSCode inserting overeagerly :)

test/Tools/circt-bmc/lower-to-bmc.mlir
TaoBi22: Yeah, that's a good point! Fixed now!

lib/Tools/circt-bmc/LowerToBMC.cpp
TaoBi22: Yeah, that's a good point! I'll get that fixed in this PR since it's a quick one

lib/Tools/circt-bmc/LowerToBMC.cpp
TaoBi22: Great point! I guess the low-cost way to handle this would just be to not throw this error if there's an instance op, but I think it's probably nice to catch this early on. If you agree then I'll add a traversal over instances to check whether there's an assertion anywhere?

maerhart: Either way sounds good to me. You'll have to be conservative when hitting an external module or function, though. In that case we could still do a check at runtime, however. Could also be worth skipping for now and adding in a later PR once the general flow is operational 😄

TaoBi22: Good point - for the time being I'll just leave set it to not throw if there are instance ops

uenoku (approved):
Several nits, otherwise LGTM!

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,191 @@
...
+                        Twine(successStr).concat(Twine('\00'))));
+
+    StringRef failureStr = "Assertion can be violated!\n";
+    auto failArrayTy =
+        LLVM::LLVMArrayType::get(builder.getI8Type(), failureStr.size() + 1);
+    failureGlobal = builder.create<LLVM::GlobalOp>(
+        loc, failArrayTy, /*isConstant=*/true, LLVM::linkage::Linkage::Private,
+        failureStr,
+        StringAttr::get(builder.getContext(),
+                        Twine(failureStr).concat(Twine('\00'))));
```
maerhart, line 170: I didn't really mean to inline the whole function, sorry if that wasn't clear in my previous comment. Having the duplicate logic in a lambda would still be nice. I meant we could just generate a unique symbol and create new global ops each time instead of checking if a matching one already exists (i.e., not just use the content string as symbol).

maerhart (approved):

lib/Tools/circt-bmc/LowerToBMC.cpp
TaoBi22: Ah my mistake! Yeah I misunderstood, I see your point now. I'll get that updated!

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,184 @@
...
+  // we expect?
+
+  return builder.create<LLVM::AddressOfOp>(loc, global);
+}
+
+void LowerToBMCPass::runOnOperation() {
+  Namespace names;
+
+  // Fetch the 'hw.module' operation to model check.
+  Operation *expectedModule = getOperation().lookupSymbol(topModule);
```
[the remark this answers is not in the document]
TaoBi22: Good point, this clears up an error too. Thanks!

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -0,0 +1,191 @@
...
+        "registers pass first");
+    return signalPassFailure();
+  }
+
+  // Check that there's only one clock input to the module
+  // TODO: supporting multiple clocks isn't too hard, an interleaving of clock
+  // toggles just needs to be generated
+  bool hasClk = false;
+  for (auto input : hwModule.getInputTypes()) {
+    if (isa<seq::ClockType>(input)) {
```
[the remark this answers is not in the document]
TaoBi22: Oh yeah, great point! I've added a check for that

lib/Tools/circt-bmc/LowerToBMC.cpp
TaoBi22: @maerhart does the lambda version in the latest commit look good to you?

maerhart: Looks great! Thanks!

Issue #7347, opened 2024-07-18
[FIRRTL] Incorrect Output Directory for Multiply Instantiate Module in Layer


seldridge:
I'm seeing what I think is incorrect behavior for output directories for the following circuit:

``` firrtl
FIRRTL version 4.0.0
circuit Foo: %[[
  {
    "class": "firrtl.transforms.DontTouchAnnotation",
    "target": "~Foo|Bar>b"
  },
  {
    "class": "firrtl.transforms.DontTouchAnnotation",
    "target": "~Foo|Baz>b"
  },
  {
    "class": "sifive.enterprise.firrtl.TestBenchDirAnnotation",
    "dirname": "verification/testbench"
  },
  {
    "class": "sifive.enterprise.firrtl.MarkDUTAnnotation",
    "target": "~Foo|Bar"
  }
]]
  layer A, bind, "verification/testbench":

  module Baz:
    wire b: UInt<1>
    invalidate b

  module Bar:
    wire b: UInt<1>
    invalidate b

    layerblock A:
      inst baz_0 of Baz
      inst baz_1 of Baz

  public module Foo:

    inst bar of Bar
```

What I am seeing is the following:

```
# rm -rf design && firtool TestBenchDirAnnotation.fir -o design -split-verilog && tree design
design
├── Bar.sv
├── filelist.f
└── verification
    ├── Baz.sv
    └── testbench
        ├── Bar_A.sv
        ├── Foo.sv
        └── layers_Foo_A.sv
```

It is unexpected that `Baz.sv` is located under `design/verification/`. It is expected that this should be under `design/verification/testbench/`.

Interestingly, if I instead _singly-instantiate_ Baz, I get the right output:

``` firrtl
FIRRTL version 4.0.0
circuit Foo: %[[
  {
    "class": "firrtl.transforms.DontTouchAnnotation",
    "target": "~Foo|Bar>b"
  },
  {
    "class": "firrtl.transforms.DontTouchAnnotation",
    "target": "~Foo|Baz>b"
  },
  {
    "class": "sifive.enterprise.firrtl.TestBenchDirAnnotation",
    "dirname": "verification/testbench"
  },
  {
    "class": "sifive.enterprise.firrtl.MarkDUTAnnotation",
    "target": "~Foo|Bar"
  }
]]
  layer A, bind, "verification/testbench":

  module Baz:
    wire b: UInt<1>
    invalidate b

  module Bar:
    wire b: UInt<1>
    invalidate b

[...]

PR #7348, opened 2024-07-18
[FIRRTL] AssignOutputDirs: Ensure absolute dirs end with trailing slash

rwy7:
When making a directory absolute, the remove_dots helper canonicalizes away any trailing slash. Add the slash back on, which ensures that we interpret the path as a directory correctly.

This fixes a bug where a path, when LCA'd with itself, drops the last segment.

An example of the problem: given a path foo/bar/, if we (accidentally) drop the trailing slash, foo/bar is interpreted as a file called bar in a directory called foo. When we LCA foo/bar with itself, the common parent directory is determined to be foo/. By retaining the trailing slash, we can ensure that the correct LCA path foo/bar/ is computed.

Fixes: #7347

### DIFF
--- a/test/Dialect/FIRRTL/assign-output-dirs.mlir
+++ b/test/Dialect/FIRRTL/assign-output-dirs.mlir
@@ -93,3 +93,38 @@ firrtl.circuit "AssignOutputDirs" {
     firrtl.instance byYZ @ByYZ()
   }
 }
+
+// https://github.com/llvm/circt/issues/7347
+firrtl.circuit "SameOutputDirTwice" {
+  // CHECK: firrtl.module private @Bar() attributes {output_file = #hw.output_file<"verification{{/|\\\\}}testbench{{/|\\\\}}">} {
+  firrtl.module private @Bar() {
+  }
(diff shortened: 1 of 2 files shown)


test/Dialect/FIRRTL/assign-output-dirs.mlir
```diff
@@ -93,3 +93,15 @@ firrtl.circuit "AssignOutputDirs" {
     firrtl.instance byYZ @ByYZ()
   }
 }
+
+// https://github.com/llvm/circt/issues/7347
+firrtl.circuit "SameOutputDirTwice" {
+  // CHECK: firrtl.module private @Bar() attributes {output_file = #hw.output_file<"verification/testbench/">} {
+  firrtl.module private @Bar() {
+  }
+  firrtl.module @Foo() attributes {convention = #firrtl<convention scalarized>, output_file = #hw.output_file<"verification/testbench/">} {
```
seldridge, line 102: Nit: 
```suggestion
  firrtl.module @Foo() attributes {output_file = #hw.output_file<"verification/testbench/">} {
```

seldridge (approved):
LGTM

PR #7349, opened 2024-07-19
[ImportVerilog] Add basic function support

fabianschuiki:
Add basic support for function definitions and calls to such functions.

Input function arguments are currently passed into the function by value, while output, inout, ref, and const ref arguments are passed by reference at the moment. Later on we may want to be more faithful to the SV spec and make output and inout arguments also pass by value.

Calls to void functions are slightly strange in that they don't really have a result value, but the `convertExpression` function really wants a valid `Value` to return. To work around this, these calls generate an additional `!moore.void` value using `unrealized_conversion_cast`, which gets deleted again by the statement conversion code.

### DIFF
--- a/lib/Conversion/ImportVerilog/CMakeLists.txt
+++ b/lib/Conversion/ImportVerilog/CMakeLists.txt
@@ -42,6 +42,7 @@ add_circt_translation_library(CIRCTImportVerilog
   LINK_LIBS PUBLIC
   CIRCTHW
   CIRCTMoore
+  MLIRFuncDialect
   MLIRSCFDialect
   MLIRTranslateLib
   PRIVATE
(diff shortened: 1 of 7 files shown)


hailongSun2000 (commented):
This is so cool :sunglasses:! I fully agree with this :hand:!

hailongSun2000 (approved):

hailongSun2000:
I noticed `verif` dialect has `assert`, `print` ops, when we handle the system functions like `assert` and `disply`, should we inline `verif` dialect into `moore`?

fabianschuiki:
> I noticed `verif` dialect has `assert`, `print` ops, when we handle the system functions like `assert` and `disply`, should we inline `verif` dialect into `moore`?

I think we'll want to make the `MooreToCore` conversion pass lower `moore.assert` and `moore.display` (do not exist yet) to the corresponding `verif.*` operations.

hailongSun2000:
I get it!

PR #7350, opened 2024-07-19
[Seq] Canonicalize firreg with preset 0

prithayan:
Canonicalize `FirReg` with a constant 0 preset. The pattern was earlier ignoring registers with any preset.

This builds on https://github.com/llvm/circt/pull/7289


lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -675,20 +684,25 @@ OpFoldResult FirRegOp::fold(FoldAdaptor adaptor) {
...
 
   // Otherwise we want to replace the register with a constant 0. For now this
   // only works with integer types.
   auto intType = dyn_cast<IntegerType>(getType());
   if (!intType)
     return {};
+  // If a non-zero preset present, then we cannot replace with zero.
+  if (presetAttr)
+    if (!presetAttr.getValue().isZero())
+      return {};
```
uenoku, line 705: I think we can replace it with `presetAttr` even when preset is not zero?

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -651,20 +657,23 @@ LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
...
   // value in silicon. If it is synchronous, the register value is undefined
   // until the first clock edge at which point it becomes the reset value, in
   // which case we simply define the initial value to already be the reset
-  // value.
-  if (auto reset = getReset())
-    if (auto constOp = reset.getDefiningOp<hw::ConstantOp>())
-      if (constOp.getValue().isOne())
-        return getResetValue();
+  // value. Works only if no preset.
+  if (!presetAttr)
```
davidbiancolin, line 672: This should also work fine if the preset value equals the reset value, right?

prithayan: This is actually checking if the reset signal is a constant constant high, then replace the register with the reset value. The reset value need not even be a constant. But if the reset is also a constant and its value matches with the preset, then yes we can replace it.

lib/Dialect/Seq/SeqOps.cpp
prithayan: You are right, addressed it.

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -651,20 +657,23 @@ LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
...
+    if (auto reset = getReset())
+      if (auto constOp = reset.getDefiningOp<hw::ConstantOp>())
+        if (constOp.getValue().isOne())
+          return getResetValue();
```
uenoku, line 676: Not directly related but I feel this canonicalization is not correct either... I think reset value needs to constant

test/Dialect/Seq/canonicalization.mlir
```diff
@@ -79,6 +79,14 @@ hw.module @FirRegReset(in %clk : !seq.clock, in %in : i32, in %r : i1, in %v : i
...
   // CHECK: %reg_preset = seq.firreg
   %reg_preset = seq.firreg %reg_preset clock %clk reset sync %r, %c0_i32 preset 3: i32
+
+  // A register with 0 preset value is folded.
+  // CHECK-NOT: %reg_preset_0 = seq.firreg
+  %reg_preset_0 = seq.firreg %reg_preset_0 clock %clk preset 0: i32
+
+  // A register with const false reset and 0 preset value is folded.
+  // CHECK-NOT: %reg_preset_1 = seq.firreg 
+  %reg_preset_1 = seq.firreg %reg_preset_1 clock %clk reset sync %false, %c0_i32 preset 0: i32
```
uenoku, line 89: Could you add a test for non-zero preset and observe folded values?

uenoku (approved):
A few comments but otherwise LGTM

lib/Dialect/Seq/SeqOps.cpp
prithayan: okay, seemed fine to me, according to the comments in the code. But lets investigate this further, if its legal or not.

PR #7351, opened 2024-07-19
[LLHD] Remove llhd-sim

maerhart:
Getting an e2e flow from SV to simulation going requires Moore to be compiled through LLHD to Arcilator. In the process it would be beneficial to refactor some LLHD constructs to better integrate with the core dialects (e.g., remove `llhd.entity` in favour of `hw.module`, refactor `llhd.proc`). Keeping llhd-sim up-to-date during those big refactorings poses a lot of additional work. The initial assumption was that we can unify llhd-sim with arcilator by reusing parts of llhd-sim, however, given those LLHD refactorings and the fact that the LLHD to LLVM lowering is already partially broken and hard to read and maintain, it probably makes more sense to rewrite this part from scratch.


fabianschuiki (approved):
Wow thanks a lot for going through all these files, and even updating the dialect diagram! :partying_face: I totally agree -- evolving LLHD and extending Arcilator to simulate it is going to be easier without llhd-sim in the picture.

PR #7353, opened 2024-07-19
[LLHD] Use hw.inout instead of llhd.sig

maerhart:
This aims to better integrate LLHD into the core dialects and is a preparation for other LLHD dialect refactorings (e.g., removing `llhd.entity` and using `hw.module` instead).


include/circt/Dialect/LLHD/IR/LLHDTypesImpl.td
```diff
@@ -17,26 +17,6 @@ class LLHDType<string name> : TypeDef<LLHD_Dialect, name> { }
...
-
-  let builders = [
-    AttrBuilderWithInferredContext<(ins "::mlir::Type":$underlyingType), [{
-      return $_get(underlyingType.getContext(), underlyingType);
-    }]>
-  ];
-}
-
 // Declares the llhd::PtrType in C++.
 def PtrTypeImpl : LLHDType<"Ptr"> {
```
fabianschuiki, line 21: Do you think we might be able to use `memref`s as a replacement for `!llhd.ptr` in the future?

fabianschuiki (approved):
Really nice! I like the idea of reusing more things from the core dialects :partying_face:

include/circt/Dialect/LLHD/IR/LLHDTypesImpl.td
maerhart: In theory yes, but we might not want to pull in all the complexity of `memref` in practice.

PR #7355, opened 2024-07-19
[CI] Report failure of clang-tidy

rwy7:
Correctly detect that clang-tidy found problems, by looking at its status-code, not whether it produced any fixups.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -268,19 +268,19 @@ jobs:
       - name: clang-tidy
         if: ${{ always() }}
         run: |
-          git diff -U0 $DIFF_COMMIT...HEAD | \
-            clang-tidy-diff -path build -p1 -fix -j$(nproc)
-          git clang-format -f $DIFF_COMMIT
-          git diff --ignore-submodules > clang-tidy.patch
-          if [ -s clang-tidy.patch ]; then
-            echo "Clang-tidy problems in the following files. " \
-              "See diff in the clang-tidy.patch artifact."
-            git diff --ignore-submodules --name-only
-            git checkout .
+          if git diff -U0 $DIFF_COMMIT...HEAD | clang-tidy-diff -path build -p1 -fix -j$(nproc); then
+            exit 0
+          fi
+          if git diff --quiet --ignore-submodules; then
             exit 1
           fi
-          echo "Clang-tidy found no problems"
-          exit 0
+          git clang-format -f $DIFF_COMMIT
+          echo "Fixups available for the following files:"
+          git diff --ignore-submodules --name-only
+          echo "See diff in the clang-tidy.patch artifact."
+          git diff --ignore-submodules > clang-tidy.patch
+          git checkout .
+          exit 1
 
       # Upload the tidy patches to an artifact (zip'd) associated
       # with the workflow run. Only run this on a failure.


.github/workflows/buildAndTest.yml
```diff
@@ -266,21 +266,19 @@ jobs:
...
         run: |
-          git diff -U0 $DIFF_COMMIT...HEAD | \
-            clang-tidy-diff -path build -p1 -fix -j$(nproc)
-          git clang-format -f $DIFF_COMMIT
-          git diff --ignore-submodules > clang-tidy.patch
-          if [ -s clang-tidy.patch ]; then
-            echo "Clang-tidy problems in the following files. " \
-              "See diff in the clang-tidy.patch artifact."
-            git diff --ignore-submodules --name-only
-            git checkout .
```
teqdruid, line 279: IIRC we had `git checkout .` to put it back the way we found it. There may not be any steps after this which operate on the source code, but there could be eventually. I think it should be added back.

teqdruid (approved):
Thanks for doing this!

.github/workflows/buildAndTest.yml
rwy7: Sounds good!

.github/workflows/buildAndTest.yml
```diff
@@ -266,21 +266,19 @@ jobs:
       # Run clang-tidy against only the changes. The 'clang-tidy-diff' script
       # does this if supplied with the diff.
       - name: clang-tidy
-        if: ${{ always() }}
```
rwy7, line 269: @teqdruid do you know why this was here? IIUC It means that the job would run even if cancelled (or if an earlier job failed). Is it just that we want to run the clang-tidy checks even if other checks fail? Should I add it back in?

https://docs.github.com/en/actions/learn-github-actions/expressions#always

teqdruid: It was desirable to run the clang-tidy checks even when the check-circt test fails. Yeah, add it back.

.github/workflows/buildAndTest.yml
```diff
@@ -266,21 +266,19 @@ jobs:
...
-          git clang-format -f $DIFF_COMMIT
-          git diff --ignore-submodules > clang-tidy.patch
-          if [ -s clang-tidy.patch ]; then
-            echo "Clang-tidy problems in the following files. " \
-              "See diff in the clang-tidy.patch artifact."
-            git diff --ignore-submodules --name-only
-            git checkout .
+          FAILED=0
+          git diff -U0 $DIFF_COMMIT..HEAD | clang-tidy-diff -path build -p1 -fix -j$(nproc) || FAILED=1
+          git clang-format -f $DIFF_COMMIT || FAILED=1
```
rwy7, line 272: Does it actually make sense to run clang-format here? It seems clang-format already runs as part of Sanity Check, so I think this is just wasted effort. Thoughts on removing?

The clang-format job is here:
https://github.com/llvm/circt/blob/main/.github/workflows/buildAndTest.yml#L71

and an example run:
https://github.com/llvm/circt/actions/runs/10010063935/job/27670384820#step:7:1

.github/workflows/buildAndTest.yml
rwy7: Does anyone actually use the clang-tidy patch artifacts? I could get rid of it, while I'm here. I don't personally use it.  I'm happy to leave it in if it's useful (or if we're not sure).

.github/workflows/buildAndTest.yml
rwy7: What about if I just moved the clang-tidy step to before the main build. I think I can do this if I run cmake before clang-tidy. Then, that clang-tidy won't run if an earlier step, such as checking out the source code, fails. As well, if the lint fails, we won't bother running the build. It might be leaner on resources.  I can just leave this as-is, too... I'm just looking around for anything we might want to change, while I'm here

.github/workflows/buildAndTest.yml
uenoku: I remember clang-tidy fix is often formatted by clang-format. I think it's fine to remove though.

uenoku (approved):
Cool, I thought clang-tidy requires full build but seems header is sufficient. Thank you for cleaning up this!

dtzSiFive (approved):
LGTM, although I see some small considerations left but I'm good either way.  Thanks for fixing this!!

.github/workflows/buildAndTest.yml
rwy7: Leaving it in.

rwy7:
The change was snowballing so I cut it way back, it could use another look

PR #7356, opened 2024-07-19
[Handshake] Fix library dependencies

youngar:
The Handshake passes are including a header file from the Handshake conversion passes, but don't formally depend on the conversion passes.  This can lead to non-deterministic build errors, when a header file is being included which has not yet been generated by the build system, such as the following: 
```
  [ 67%] Building CXX object tools/circt/lib/Dialect/Handshake/Transforms/CMakeFiles/obj.CIRCTHandshakeTransforms.dir/LowerExtmemToHW.cpp.o
  In file included from /Users/runner/work/circt/circt/lib/Dialect/Handshake/Transforms/LowerExtmemToHW.cpp:13:
  /Users/runner/work/circt/circt/include/circt/Conversion/HandshakeToHW.h:28:10: fatal error: 'circt/Conversion/Passes.h.inc' file not found
  #include "circt/Conversion/Passes.h.inc"
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  1 error generated.
  make[3]: *** [tools/circt/lib/Dialect/Handshake/Transforms/CMakeFiles/obj.CIRCTHandshakeTransforms.dir/LowerExtmemToHW.cpp.o] Error 1
  make[2]: *** [tools/circt/lib/Dialect/Handshake/Transforms/CMakeFiles/obj.CIRCTHandshakeTransforms.dir/all] Error 2
  make[1]: *** [tools/circt/lib/Bindings/Python/CMakeFiles/install-CIRCTPythonModules.dir/rule] Error 2
  ```
  
  Instead of just making the passes depend on the conversion passes, I moved the common functionality to the Handshake dialect library in HandshakeUtils.h.  If a function had only a single user, I made the function static to that compilation unit.
At the same time, I moved a bunch of utility functionality out of the regular passes and in to HandshakeUtils.h, so no longer do the conversion passes need to depend on the regular passes.

I also added a dialect dependency from Handshake to ESI, as the helper `getPortInfoForOpTypes` seems to create types from the ESI dialect.

### DIFF
--- a/include/circt/Conversion/HandshakeToHW.h
+++ b/include/circt/Conversion/HandshakeToHW.h
@@ -27,32 +27,8 @@ namespace circt {
 #define GEN_PASS_DECL_HANDSHAKETOHW
 #include "circt/Conversion/Passes.h.inc"
 
-namespace esi {
-class ChannelType;
-} // namespace esi
-
 std::unique_ptr<mlir::Pass> createHandshakeToHWPass();
 
-namespace handshake {
-
-// Attribute name for the name of a predeclaration of the to-be-lowered
-// hw.module from a handshake function.
-static constexpr const char *kPredeclarationAttr = "handshake.module_name";
-
-// Converts 't' into a valid HW type. This is strictly used for converting
-// 'index' types into a fixed-width type.
-Type toValidType(Type t);
-
-// Wraps a type into an ESI ChannelType type. The inner type is converted to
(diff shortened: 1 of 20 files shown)


fabianschuiki (approved):
I'm not familiar with the Handshake dialect, but the changes look sensible and they clean up the dependencies nicely! :+1:

PR #7357, opened 2024-07-19
[LLHD] Refactor llhd.proc and remove llhd.inst

maerhart:
LLHD processes are now nested inside `hw.module`, thus `llhd.inst` is not needed anymore.


fabianschuiki (approved):
LGTM! This is really really awesome! :partying_face: :100: :rocket:

PR #7358, opened 2024-07-19
[FIRRTL] LowerToHW: guard against folded operations

youngar:
In LowerToHW we copy the name from FIRRTL ops to the lowered HW ops.  We often call `createOrFold` when creating the HW ops, which can mean that there is no operation to copy the name to.  We were not properly detecting this scenario, which could cause us to try to copy the name to a null operation.  Most ops were already covered under a guard in a generic code path, but div-like and subaccess-like operations needed to be guarded.  Some of these problem are impossible to test right now as we require folders to be implemented for some operations.

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -2729,7 +2729,8 @@ FailureOr<Value> FIRRTLLowering::lowerSubindex(SubindexOp op, Value input) {
     result = builder.createOrFold<sv::ArrayIndexInOutOp>(input, iIdx);
   else
     result = builder.createOrFold<hw::ArrayGetOp>(input, iIdx);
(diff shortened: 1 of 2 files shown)


seldridge (approved):
LGTM

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -2752,7 +2753,8 @@ FailureOr<Value> FIRRTLLowering::lowerSubaccess(SubaccessOp op, Value input) {
     result = builder.createOrFold<sv::ArrayIndexInOutOp>(input, valueIdx);
   else
     result = createArrayIndexing(input, valueIdx);
-  tryCopyName(result.getDefiningOp(), op);
+  if (auto definingOp = result.getDefiningOp())
```
seldridge, line 2756: PR is missing one `auto *definingOp` in the cleanup.

PR #7359, opened 2024-07-19
[firtool] Make layer specialization take a list of layers

youngar:
The `firtool` options `-enable-layers` and `-disable-layers` are supposed to take a list of layers to enable or disable.  The code forgot to include the comma-separated flag, which meant that it could only take one layer at a time.  You could, however, specify multiple layers by including the command line option multiple times, but this was never the intended behaviour.


seldridge (approved):

dtzSiFive (commented):
Thanks! This is more intuitive and convenient, awesome!

PR #7360, opened 2024-07-20
[Arc] Add Initial Cost Model

elhewaty:


elhewaty:
@maerhart @fabianschuiki.

elhewaty:
@maerhart @fabianschuiki, What is wrong with the build?

maerhart:
> @maerhart @fabianschuiki, What is wrong with the build?

You are using operations from the Comb and Func dialects but not linking against them in the `CMakeLists.txt` file. You'd need to add them as link targets there.

elhewaty:
@maerhart 

in `lib/Dialect/Arc/CMakeLists.txt`

```
add_circt_dialect_library(CIRCTArc
  ${CIRCT_Arc_Sources}

  ADDITIONAL_HEADER_DIRS
  ${CIRCT_MAIN_INCLUDE_DIR}/circt/Dialect/Arc

  DEPENDS
  CIRCTArcEnumsIncGen
  CIRCTArcInterfacesIncGen
  MLIRArcIncGen

  LINK_COMPONENTS
  Support

  LINK_LIBS PUBLIC
  CIRCTComb
  ...
  MLIRIR
  ...
  MLIRFunc
)
```

lib/Dialect/Arc/ArcCostModel.cpp
```diff
@@ -0,0 +1,106 @@
...
+    return size_t(OperationCost::CONCATCOST);
+  if (isa<circt::comb::ExtractOp>(op))
+    return size_t(OperationCost::EXTRACTCOST);
+  // We have some other functions that need to be handled in a different way
+  // arc::StateOp, arc::CallOp, mlir::func::CallOp and arc::VectorizeOp, each of
+  // these functions have bodies so the cost of the op equals the cost of its
+  // body.
+  if (isa<arc::StateOp>(op) || isa<arc::CallOp>(op) ||
+      isa<mlir::func::CallOp>(op))
+    return countBodyOps(op->getOperand(0).getDefiningOp()->getRegions());
```
fabianschuiki, line 32: `getOperand(0)` gives you the first call argument. I think what you want here is `getCallee()` or similar. I think the `CallOpInterface` that is implemented by these ops might have a few helpful functions :smiley:

lib/Dialect/Arc/ArcCostModel.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  if (isa<arc::VectorizeOp>(op))
+    return countBodyOps(op->getRegions()) +
+           getInputVectorsCost(op->getOperand(0).getDefiningOp<VectorizeOp>());
+
+  return size_t(OperationCost::NORMALCOST);
+}
+
+size_t ArcCostModel::countBodyOps(const MutableArrayRef<Region> &regions) {
+  size_t counter = 0;
```
fabianschuiki, line 42: It would be good to cache the result of `countBodyOps` in a `DenseMap` somewhere, such that when you call the function again you don't have to go and compute the operation costs again, but can re-use the last result. Otherwise you'll end up revisiting the IR a lot.

elhewaty:
> I was also interested with cost model for the whole project, this looks very nice. However, it seems to be the cost for time dimension, could you add cost for memory-space dimiension, which may be determined by data type or data amount?

@mingzheTerapines Can you please explain a little further? Honestly, I don't get this

mingzheTerapines:
@elhewaty I mean this is calculating static cost for the model, but if we know the size of input(such as vector) we can know the dynamcia memory/register space cost.

elhewaty:
@mingzheTerapines, It seems an interesting idea, I will try to investigate further. @fabianschuiki what do you think, will this affect the vectors' cost significantly??

I am working on minimizing the shuffling cost in vectors. and I will return to this PR.

elhewaty:
@fabianschuiki @maerhart PING

include/circt/Dialect/Arc/ArcCostModel.h
```diff
@@ -0,0 +1,67 @@
...
+  // 4: the input is a mix of some vectors:
+  //    a) same order we multiply by 2
+  //    b) shuffling we multiply by 3
+  size_t getInputVectorsCost(VectorizeOp vecOp);
+  size_t getShufflingCost(const ValueRange &inputVec, bool isSame = false);
+  DenseMap<Operation *, size_t> opCostCash;
+  size_t packingCost{0};
+  size_t shufflingCost{0};
+  size_t vectoroizeOpsBodyCost{0};
+  size_t allVectorizeOpsCost{0};
```
fabianschuiki, line 61: I would not keep a global cost here in the `ArcCostModel`. In the future, we will probably use `ArcCostModel` like an analysis or cost cache, and create one instance of `ArcCostModel` that is then used to query the costs of different operations. That becomes easier if you don't have global cost counts that only count the specific ops passed to `getCost()`.

If you'd like to keep separate track of the packing, shuffling, and vectorize body op costs, you could create a cost struct that has separate integers for these costs. You could then pass an entire `hw::HWModuleOp` or `arc::VectorizeOp` or `mlir::ModuleOp` to `getCost(...)` and get the detailed split costs as a result. And they would be cached 😃

include/circt/Dialect/Arc/ArcPasses.td
```diff
@@ -58,6 +58,24 @@ def Dedup : Pass<"arc-dedup", "mlir::ModuleOp"> {
   ];
 }
 
+def DummyAnalysisTester : Pass<"arc-dummy-analysis-tester", "mlir::ModuleOp"> {
+  let summary = "A dymmy pass to test analysis passes";
+  let constructor = "circt::arc::createDummyAnalysisTesterPass()";
+  let dependentDialects = ["arc::ArcDialect"];
```
fabianschuiki, line 64: Can you call this something like `PrintCostModel`, `arc-print-cost-model`? Dummy analysis tester is very generic.

lib/Dialect/Arc/ArcCostModel.cpp
```diff
@@ -0,0 +1,120 @@
...
+      isa<mlir::func::CallOp>(op)) {
+    size_t totalCost = 0;
+    const auto regions =
+        dyn_cast<CallOpInterface>(op).resolveCallable()->getRegions();
+    for (auto &region : regions)
+      for (auto &block : region)
+        for (auto &innerOp : block)
+          totalCost += computeOperationCost(&innerOp);
+    return opCostCash[op] = totalCost;
+  }
```
fabianschuiki, line 42: In this case I would do something like `getCost(cast<CallOpInterface>(op).resolveCallable())` and then make the `for region { for block { for innerOp { count cost }}}` pattern the default for any operation. That will allow you to call `getCost()` on an `hw::HWModuleOp` or `mlir::ModuleOp` or `arc::DefineOp` as well, and get the correct results.

include/circt/Dialect/Arc/ArcCostModel.h
```diff
@@ -0,0 +1,67 @@
...
+  NOCOST,
+  NORMALCOST,
+  PACKCOST = 2,
+  EXTRACTCOST = 3,
+  CONCATCOST = 3,
+  SAMEVECTORNOSHUFFLE = 0,
+  SAMEVECTORSHUFFLECOST = 2,
+  DIFFERENTVECTORNOSHUFFLE = 2,
+  DIFFERENTVECTORSHUFFLECOST = 3
+};
```
fabianschuiki, line 32: This does not show up as cost in the public interface of the cost model (which is great)! Therefore I would move this into the `*.cpp` file.

include/circt/Dialect/Arc/ArcPasses.td
elhewaty: you're right, what a dummy name!

lib/Dialect/Arc/ArcCostModel.cpp
elhewaty: I think I don't get this. should I call `getCost` as you suggest?
but this will make a recursive chain as `getCost` will call `computeOperationCost`?
so the `computeOperationCost` will just have the base cases and the `nested for loops`?
should we remove the check for `vectorizeOp` as well?

include/circt/Dialect/Arc/ArcCostModel.h
elhewaty: should I split getCost into separate interfaces? each one handles difference type?

lib/Dialect/Arc/ArcCostModel.cpp
fabianschuiki: Oh you can also call `computeOperationCost` instead, since `getCost` is just forwarding the call. But yeah, I would call `computeOperationCost` on the thing being called, and then add a way to compute the cost of function and arc definitions. That will make the cost model easier to use, because you can just ask for the cost of an entire function or arc.

I think you still need a bit of special handling for the vectorize op, since it has the shuffling cost you need to account for.

include/circt/Dialect/Arc/ArcCostModel.h
elhewaty: @fabianschuiki I think I don't get this, could you please explain a little further?

lib/Dialect/Arc/ArcCostModel.cpp
elhewaty: > I would call computeOperationCost on the thing being called, and then add a way to compute the cost of function and arc definitions

Please explain this part, it's a bit confusing to me.

include/circt/Dialect/Arc/ArcCostModel.h
elhewaty: do you mean something like this?
```
struct OperationCosts {
  size_t normalCost{0};
  size_t packingCost{0};
  size_t shufflingCost{0};
  size_t vectorizeOpsBodyCost{0};
  size_t totalCost() const {
    return normalCost + packingCost + shufflingCost + vectorizeOpsBodyCost;
  }
};
```

elhewaty: and calculate the costs with this struct instead of size_t?
like:
```
OperationCosts ArcCostModel::getInputVectorsCost
```

before doing this, do you need to save this information?

lib/Dialect/Arc/ArcCostModel.cpp
elhewaty: How about this?
```
size_t ArcCostModel::computeOperationCost(Operation *op) {
  if (opCostCash.count(op))
    return opCostCash[op];
  if (isa<circt::comb::ConcatOp>(op))
    return opCostCash[op] = size_t(OperationCost::CONCATCOST);
  if (isa<circt::comb::ExtractOp>(op))
    return opCostCash[op] = size_t(OperationCost::EXTRACTCOST);

  if (isa<arc::VectorizeOp>(op)) {
    size_t inputVecCost = getInputVectorsCost(dyn_cast<VectorizeOp>(op));
    size_t vecOpBodyCost = 0;
    auto regions = op->getRegions();
    for (auto &region : regions)
      for (auto &block : region)
        for (auto &innerOp : block)
          vecOpBodyCost += computeOperationCost(&innerOp);

    vectoroizeOpsBodyCost += vecOpBodyCost;
    allVectorizeOpsCost += inputVecCost + vecOpBodyCost;
    return opCostCash[op] = inputVecCost + vecOpBodyCost;
  }
  // We have some other functions that need to be handled in a different way
  // arc::StateOp, arc::CallOp, mlir::func::CallOp and arc::VectorizeOp, each of
  // these functions have bodies so the cost of the op equals the cost of its
  // body.
  if (auto *callableOp = cast<CallOpInterface>(op).resolveCallable()) {
    size_t totalCost = 0;
    const auto regions = callableOp->getRegions();
    for (auto &region : regions)
      for (auto &block : region)
        for (auto &innerOp : block)
          totalCost += computeOperationCost(&innerOp);
    return opCostCash[op] = totalCost;
  }

  return opCostCash[op] = size_t(OperationCost::NORMALCOST);
}
```

include/circt/Dialect/Arc/ArcCostModel.h
fabianschuiki: Yeah this looks great! You could then caches these `OperationCosts` in your dense map, and return it from `getCost(...)` as you already do. 😃

lib/Dialect/Arc/ArcCostModel.cpp
fabianschuiki: I would turn this:
```c++
  if (auto *callableOp = cast<CallOpInterface>(op).resolveCallable()) {
    size_t totalCost = 0;
    const auto regions = callableOp->getRegions();
    for (auto &region : regions)
      for (auto &block : region)
        for (auto &innerOp : block)
          totalCost += computeOperationCost(&innerOp);
    return opCostCash[op] = totalCost;
  }
```
into something like this:
```c++
  if (auto callableOp = dyn_cast<CallOpInterface>(op))
    if (auto *calledOp = callableOp.resolveCallable())
      return computeOperationCost(*calledOp);
  if (isa<func::FuncOp, arc::DefineOp, mlir::ModuleOp>(op)) {
    size_t totalCost = 0;
    for (auto &region : op.getRegions())
      for (auto &block : region)
        for (auto &innerOp : block)
          totalCost += computeOperationCost(&innerOp);
    return opCostCash[op] = totalCost;
  }
```
And then switch from `size_t` costs to the `OperationCosts` struct you've proposed above 😃. This makes call ops use `computeOperationCost` on the op they are calling. And the region->block->ops cost counting is done whenever you call `computeOperationCost` on a `func.func`, `arc.define`, or `builtin.module`. This will allow you to compute the total operation cost in your IR, e.g. for printing, by just asking your cost model for the cost of the root `builtin.module` op.

include/circt/Dialect/Arc/ArcCostModel.h
```diff
@@ -52,13 +51,9 @@ class ArcCostModel {
...
   //    b) shuffling we multiply by 3
-  size_t getInputVectorsCost(VectorizeOp vecOp);
+  OperationCosts getInputVectorsCost(VectorizeOp vecOp);
   size_t getShufflingCost(const ValueRange &inputVec, bool isSame = false);
-  DenseMap<Operation *, size_t> opCostCash;
-  size_t packingCost{0};
-  size_t shufflingCost{0};
-  size_t vectoroizeOpsBodyCost{0};
-  size_t allVectorizeOpsCost{0};
+  DenseMap<Operation *, OperationCosts> opCostCash;
```
fabianschuiki, line 56:
```suggestion
  DenseMap<Operation *, OperationCosts> opCostCache;
```

lib/Dialect/Arc/ArcCostModel.cpp
```diff
@@ -16,52 +16,66 @@ using namespace circt;
...
+};
 
-size_t ArcCostModel::computeOperationCost(Operation *op) {
+OperationCosts ArcCostModel::getCost(Operation *op) {
+  return computeOperationCost(op);
+}
+
+OperationCosts ArcCostModel::computeOperationCost(Operation *op) {
   if (opCostCash.count(op))
     return opCostCash[op];
```
fabianschuiki, line 38: You can often use iterators to combine checking if a value exists in a map, and then doing something with the value if it does exist:
```suggestion
  if (auto it = opCostCache.find(op); it != opCostCache.end())
    return it->second;
```

lib/Dialect/Arc/ArcCostModel.cpp
```diff
@@ -16,52 +16,66 @@ using namespace circt;
...
-          totalCost += computeOperationCost(&innerOp);
-    return opCostCash[op] = totalCost;
-  }
+    costs.normalCost = size_t(OperationCost::CONCATCOST);
+  else if (isa<circt::comb::ExtractOp>(op))
+    costs.normalCost = size_t(OperationCost::EXTRACTCOST);
+  else if (isa<arc::VectorizeOp>(op)) {
+    // VectorizeOpCost = packingCost + shufflingCost + bodyCost
+    OperationCosts inputVecCosts =
+        getInputVectorsCost(dyn_cast<VectorizeOp>(op));
```
fabianschuiki, line 49:
```suggestion
  else if (auto vecOp = dyn_cast<arc::VectorizeOp>(op)) {
    // VectorizeOpCost = packingCost + shufflingCost + bodyCost
    OperationCosts inputVecCosts = getInputVectorsCost(vecOp);
```

lib/Dialect/Arc/Transforms/PrintCostModel.cpp
```diff
@@ -0,0 +1,60 @@
...
+} // namespace
+
+void PrintCostModelPass::runOnOperation() {
+  for (auto moduleOp : getOperation().getOps<hw::HWModuleOp>()) {
+    ArcCostModel arcCostModel;
+    moduleOp.walk([&](Operation *op) {
+      OperationCosts curOpCosts = arcCostModel.getCost(op);
+      statVars += curOpCosts;
+    });
+  }
```
fabianschuiki, line 48: This might over-count operations, since `walk` will also walk into the body of vectorize ops and count each of the ops separately, but then also call `getCost` on the vectorize op itself which will count stuff in the body again. If you add `HWModuleOp` to the list of ops that `computeOperationCosts` handles as regions->blocks->ops counting, you can do:
```suggestion
  OperationCosts statVars; // no need to carry this in the pass struct
  ArcCostModel arcCostModel; // want to keep the cached results around
  for (auto moduleOp : getOperation().getOps<hw::HWModuleOp>()) {
    statVars += arcCostModel.getCost(moduleOp);
  }
```

elhewaty:
Thanks for the review, I will update and push it now

elhewaty:
@fabianschuiki What is wrong with CMakeList file?

lib/Dialect/Arc/Transforms/DummyAnalysisTester.cpp
```diff
@@ -0,0 +1,72 @@
+//===- DummyAnalysisTester.cpp --------------------------------------------===//
```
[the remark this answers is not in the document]
elhewaty: Oh, I will update now. I will finish this PR even it costs my life, it's 12:36 AM here :sob:

fabianschuiki (approved):
LGTM! And CI also seems to be happy 😃 Thanks a lot for iterating on this!

elhewaty:
Thanks for reviewing, Let's work on the splitting :rocket: :fire:

mingzheTerapines:
@elhewaty Could you  add a option for printing the cost explictly in Ops in the future? Maybe using `dbg` dialect or something.

PR #7367, opened 2024-07-22
[ESI runtime] Host memory service

teqdruid:
This service is responsible for managing host memory which is accessible from the accelerator.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
+++ b/lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
@@ -394,6 +394,35 @@ private:
   std::unique_ptr<FuncService::Function> readMMIO;
 };
 
+class CosimHostMem : public HostMem {
+public:
+  CosimHostMem() {}
+
+  struct CosimHostMemRegion : public HostMemRegion {
+    CosimHostMemRegion(std::size_t size) {
+      ptr = malloc(size);
+      this->size = size;
+    }
+    virtual ~CosimHostMemRegion() { free(ptr); }
+    virtual void *getPtr() const { return ptr; }
+    virtual std::size_t getSize() const { return size; }
+
+  private:
+    void *ptr;
+    std::size_t size;
+  };
+
+  virtual std::unique_ptr<HostMemRegion> allocate(std::size_t size,
+                                                  HostMem::Options opts) const {
+    return std::unique_ptr<HostMemRegion>(new CosimHostMemRegion(size));
+  }
+  virtual bool mapMemory(void *ptr, std::size_t size,
+                         HostMem::Options opts) const {
+    return true;
+  }
+  virtual void unmapMemory(void *ptr) const {}
+};
+
 } // namespace
 
 Service *CosimAccelerator::createService(Service::Type svcType,
@@ -420,6 +449,8 @@ Service *CosimAccelerator::createService(Service::Type svcType,
 
   if (svcType == typeid(services::MMIO)) {
     return new CosimMMIO(getCtxt(), rpcClient);
+  } else if (svcType == typeid(services::HostMem)) {
+    return new CosimHostMem();
   } else if (svcType == typeid(SysInfo)) {
     switch (manifestMethod) {
     case ManifestMethod::Cosim:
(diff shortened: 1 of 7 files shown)


lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
```diff
@@ -394,6 +394,35 @@ class CosimMMIO : public MMIO {
...
 
+class CosimHostMem : public HostMem {
+public:
+  CosimHostMem() {}
+
+  struct CosimHostMemRegion : public HostMemRegion {
+    CosimHostMemRegion(std::size_t size) {
+      ptr = malloc(size);
+      this->size = size;
+    }
```
mortbopet, line 405: Consider makign the constructor private if CosimHostMem is the only thing that should ever create these.

lib/Dialect/ESI/runtime/cpp/lib/backends/Trace.cpp
```diff
@@ -271,12 +275,63 @@ TraceAccelerator::requestChannelsFor(AppIDPath idPath,
...
 
+class TraceHostMem : public HostMem {
+public:
+  TraceHostMem(TraceAccelerator::Impl &impl) : impl(impl) {}
+
+  struct TraceHostMemRegion : public HostMemRegion {
+    TraceHostMemRegion(std::size_t size, TraceAccelerator::Impl &impl)
+        : impl(impl) {
+      ptr = malloc(size);
+      this->size = size;
```
mortbopet, line 286: As above, wrt. constructor visibility. Even more relevant for this one, since the trace `allocate` write is performed in `TraceHostMem::allocate` and not in the `TraceHostMemRegion` constructor.

lib/Dialect/ESI/runtime/cpp/lib/backends/Trace.cpp
```diff
@@ -271,12 +275,63 @@ TraceAccelerator::requestChannelsFor(AppIDPath idPath,
...
+    void *ptr;
+    std::size_t size;
+    TraceAccelerator::Impl &impl;
+  };
+
+  virtual std::unique_ptr<HostMemRegion> allocate(std::size_t size,
+                                                  HostMem::Options opts) const {
+    auto ret =
+        std::unique_ptr<HostMemRegion>(new TraceHostMemRegion(size, impl));
+    impl.write("HostMem") << ret->getPtr() << " allocate " << size
```
mortbopet, line 305: Nit: a void* is printed to a stream as hex, right? If so (and i'd prefer that), suggestion to prefix it with `0x` (i don't assume the stream itself infers the prefix).

lib/Dialect/ESI/runtime/cpp/lib/backends/Trace.cpp
```diff
@@ -271,12 +275,63 @@ TraceAccelerator::requestChannelsFor(AppIDPath idPath,
...
+                          << " bytes. Writeable: " << opts.writeable
+                          << ", useLargePages: " << opts.useLargePages
+                          << std::endl;
+    return ret;
+  }
+  virtual bool mapMemory(void *ptr, std::size_t size,
+                         HostMem::Options opts) const {
+    impl.write("HostMem") << "map " << ptr << " size " << size
```
mortbopet, line 313: As above wrt. pointer printout

lib/Dialect/ESI/runtime/cpp/lib/backends/Trace.cpp
```diff
@@ -271,12 +275,63 @@ TraceAccelerator::requestChannelsFor(AppIDPath idPath,
...
+                          << " bytes. Writeable: " << opts.writeable
+                          << ", useLargePages: " << opts.useLargePages
+                          << std::endl;
+    return true;
+  }
+  virtual void unmapMemory(void *ptr) const {
+    impl.write("HostMem") << "unmap " << ptr << std::endl;
```
mortbopet, line 320: as above.

mortbopet (approved):

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
teqdruid: It's in a cpp file, so nothing outside the file can construct it anyway.

PR #7368, opened 2024-07-22
[ImportVerilog] Fix use after free

fzi-hielscher:
Avoid accessing an erased operation in ImportVerilog. Also removed an unused `getInputMutable`.

Note that I don't really know what is happening here. I cannot judge how safe this erasure is. E.g., can we guarantee that `refOp` is not still in use somewhere else?


fabianschuiki (approved):
Thanks for the fix! Yeah we probably don't want to erase ops here. The `StructInjectOp` looks like it wants to be a canonicalization instead :+1:

mingzheTerapines:
@fabianschuiki Sorry for these code, and thansk for reminding. You are right, canonicalization can do this.

PR #7371, opened 2024-07-22
[FIRRTL] Support layer-colored probes in force, force_initial.

### LINKED ISSUE #7370: [FIRRTL] Fix use of layer-enabled rwprobe's with force, force_initial
```firrtl
FIRRTL version 4.0.0

circuit L:
  layer T, bind:

  public module L:
    output p : RWProbe<UInt<1>, T>

    layerblock T:
      wire w : UInt<1>
      connect w, UInt<1>(0)
      define p = rwprobe(w)
      force_initial(p, UInt<1>(1))
```

Fails while verifying constraint on `force_initial` which expects the destination type to be something that can be inferred from the source (rwprobe of that type):

```
layers-rwprobe-5.fir:13:7: error: 'firrtl.ref.force_initial' op failed to verify that reference type of dest should be RWProbe of src
      force_initial(p, UInt<1>(1))
      ^
layers-rwprobe-5.fir:13:7: note: see current operation: "firrtl.ref.force_initial"(%2, %arg0, %1) : (!firrtl.uint<1>, !firrtl.rwprobe<uint<1>, @T>, !firrtl.const.uint<1>) -> ()
```

I believe the introduction of colors means that these operations no longer can infer/omit the type of their rwprobe argument, since it's unknown the layers that may be present.

dtzSiFive:
With addition of probe colors, no longer can infer the rwprobe type so make it explicit and update tests and verify accordingly.

Fixes #7370.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLStatements.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStatements.td
@@ -312,17 +312,19 @@ def ForceOp : FIRRTLOp<"force", [SameTypeOperands]> {
 
 //===- Reference force/release --------------------------------------------===//
 
-class ForceRefTypeConstraint<string ref, string base>
-  : TypesMatchWith<"reference type of " # ref # " should be RWProbe of " # base,
-                   base, ref,
-                   "RefType::get(firrtl::type_cast<FIRRTLBaseType>($_self).getAllConstDroppedType(), true)">;
+class ForceRefTypeConstraint<string ref, string base> : PredOpTrait<
+        "reference type of " # ref # " should be RWProbe of " # base,
+        CPred<[{
+          firrtl::type_cast<RefType>($}] # ref # [{.getType()).getType() ==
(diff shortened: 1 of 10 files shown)


seldridge (approved):
LGTM

Tangentially, this makes more sense to me as a human writing the MLIR by hand.

dtzSiFive:
Yeah, it's a bit surprising.  It gets awfully painful to do this for larger types (especially manually eliding the flip/const), so actually this /was/ for benefit of humans hoping the machine can do fiddly verbose work for them when it's entirely something that can be inferred.

But yes remembering which of our operations have which policy for which types they want where is a bit 😞.  Moving to "list them all (not matter what) so no surprises" is one policy we could go with, but isn't the one we have presently.

Anyway all that said, this is a rarely used operation so while it's verbose to duplicate it textually it's used infrequently enough that making it /optional/ or something (vast majority of cases it can be inferred) is probably not well motivated?
And might be even more confusing than not doing so 🙃 .

dtzSiFive:
Thanks for taking a look! 🎉

PR #7372, opened 2024-07-22
[FIRRTL] Verify RWProbeOp target has layer requirements.

dtzSiFive:
RWProbe conservatively means a write to the target, so check that the target is indeed writeable from where the rwprobe is.

### DIFF
--- a/test/Dialect/FIRRTL/errors.mlir
+++ b/test/Dialect/FIRRTL/errors.mlir
@@ -2072,6 +2072,20 @@ firrtl.circuit "RWProbeUseDef" {
 
 // -----
 
+firrtl.circuit "RWProbeLayerRequirements" {
+  firrtl.layer @A bind { }
+  firrtl.module @RWProbeLayerRequirements(in %cond : !firrtl.uint<1>) {
+    // expected-note @below {{target is missing layer requirements: @A}}
+    %w = firrtl.wire sym @x : !firrtl.uint<1>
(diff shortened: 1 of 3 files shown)


test/Dialect/FIRRTL/errors.mlir
```diff
@@ -2072,6 +2072,20 @@ firrtl.circuit "RWProbeUseDef" {
...
+  firrtl.layer @A bind { }
+  firrtl.module @RWProbeLayerRequirements(in %cond : !firrtl.uint<1>) {
+    // expected-note @below {{target is missing layer requirements: @A}}
+    %w = firrtl.wire sym @x : !firrtl.uint<1>
+    firrtl.layerblock @A {
+      // expected-error @below {{target has insufficient layer requirements}}
+      %rw = firrtl.ref.rwprobe <@RWProbeLayerRequirements::@x> : !firrtl.rwprobe<uint<1>>
+    }
+  }
+}
```
seldridge, line 2085: Are there any checks of _legally_ doing this, i.e., what should be allowed here? If there are not, can one be added?

seldridge (approved):
LGTM

test/Dialect/FIRRTL/errors.mlir
dtzSiFive: Good call, added one.  Thanks!

dtzSiFive:
Thanks!

Issue #7361, opened 2024-07-20
[FIRRTL] Canonicalization missing for narrow dynamic subindex


uenoku:
Input fir file:
```
FIRRTL version 3.0.0
circuit stmts:
  module stmts :
    input reset : UInt<1>
    input clock : Clock
    input vec : UInt<8>[8]
    input index: UInt<2>
    output b : UInt<8> 
    b <= vec[index]
```
output:
```
// Generated by CIRCT unknown git version
module stmts(
  input        reset,
               clock,
  input  [7:0] vec_0,
               vec_1,
               vec_2,
               vec_3,
               vec_4,
               vec_5,
               vec_6,
               vec_7,
  input  [1:0] index,
  output [7:0] b
);

  wire [7:0][7:0] _GEN =
    {{vec_7}, {vec_6}, {vec_5}, {vec_4}, {vec_3}, {vec_2}, {vec_1}, {vec_0}};
  assign b = _GEN[{1'h0, index}];
endmodule
```

Since `index` is 2-bit integer the range is `[0, 4)` so we can drop upper elements (vec_4 ~ vec_7). 
We can perform this canonicalization in HW(hw.array_get) and FIRRTL (firrtl.multibit_mux)

PR #7373, opened 2024-07-23
[FIRRTL] Canonicalize multibit_mux with narrow index

uenoku:
This adds a canonicalization to optimize multibit_mux whose index has narrow width. Close https://github.com/llvm/circt/issues/7361

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -1660,7 +1660,7 @@ LogicalResult MultibitMuxOp::canonicalize(MultibitMuxOp op,
   // elements.
   auto indexWidth = op.getIndex().getType().getBitWidthOrSentinel();
   uint64_t inputSize = op.getInputs().size();
-  if (indexWidth < 64 && 1ull << indexWidth < inputSize) {
+  if (indexWidth >= 0 && indexWidth < 64 && 1ull << indexWidth < inputSize) {
     rewriter.modifyOpInPlace(op, [&]() {
       op.getInputsMutable().erase(0, inputSize - (1ull << indexWidth));
     });
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -1656,6 +1656,17 @@ LogicalResult MultibitMuxOp::canonicalize(MultibitMuxOp op,
     return success();
   }
 
+  // If the index width is narrower than the size of inputs, drop front
+  // elements.
+  auto indexWidth = op.getIndex().getType().getBitWidthOrSentinel();
+  uint64_t inputSize = op.getInputs().size();
+  if (indexWidth < 64 && 1ull << indexWidth < inputSize) {
+    rewriter.modifyOpInPlace(op, [&]() {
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1656,6 +1656,17 @@ LogicalResult MultibitMuxOp::canonicalize(MultibitMuxOp op,
     return success();
   }
 
+  // If the index width is narrower than the size of inputs, drop front
+  // elements.
+  auto indexWidth = op.getIndex().getType().getBitWidthOrSentinel();
+  uint64_t inputSize = op.getInputs().size();
+  if (indexWidth < 64 && 1ull << indexWidth < inputSize) {
```
seldridge, line 1663: Is `1ull` portable across windows and Linux?

seldridge (approved):
LGTM

While, this looks right to me, it is worthwhile to verify (locally test) that the behavior is doing to right thing for edge cases, e.g., zero width.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
uenoku: I think it's fine as there are several existing use cases in LLVM and CIRCT repo.

dtzSiFive: long long is at least 64bits since c++11, according to https://en.cppreference.com/w/cpp/language/types .

That said (only looking at the code added here):

indexWidth is signed (int32_t), and in particular can be negative (!).  Shift by a negative amount is UB! Please add a check for that, as even without UB it's logically not what we want either.

dtzSiFive (commented):
Great work identifying this and putting together the canonicalizer!

On my "TODO" is to bolster our knownbits-like capabilities, and with that hat on I can't help but wonder about generalizing this beyond index being narrow directly.  Future work, but since you're pushing on this just thought I'd mention it 👍 .

Anyway, generally looks good but there's an important issue that needs to be addressed before landing this, see my comment.  Thanks!

uenoku:
> On my "TODO" is to bolster our knownbits-like capabilities, and with that hat on I can't help but wonder about generalizing this beyond index being narrow directly. Future work, but since you're pushing on this just thought I'd mention it 👍 .

Yeah that sounds great idea. I think naively computing knownbits could be expensive (as shown in https://github.com/llvm/circt/blob/e8033f1928ca78d6fa35cdf9cc8bb809b99b5422/lib/Dialect/Comb/CombFolds.cpp#L618-L627) so I guess a dedicated pass would be needed. Though computing known bits is exactly bit-sensitive IMCP so it's worth extending the lattice in IMCP.

dtzSiFive (changes requested):
Friendly upgrade to "request changes" since this is reachable UB.  I don't have the details paged in but the expected change (dropping widths, I think?) to one of the tests was all it took 👍 .

Just making sure this gets addressed, thanks!

uenoku:
Thanks! I added a condition and test to ensure that the width is not negative number

dtzSiFive (approved):
LGTM! Thanks!

(sorry for delay, missed the update 👍 )

uenoku:
Np, thank you for review!

PR #7374, opened 2024-07-23
[FIRRTL] Add Formal Test intent op to FIRRTL

dobios:
The goal of this PR is to add a new formal test intent op to FIRRTL:
```firrtl
circuit c:
      formal test1, k = 20:
            ;; formal test
```
This will then be lowered to the `verif.formal` op.


lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5153,6 +5180,47 @@ ParseResult FIRCircuitParser::parseModule(CircuitOp circuit, bool isPublic,
...
+           failure();
+
+  // Build out the firrtl mlir op
+  auto builder = circuit.getBodyBuilder();
+  auto formal = builder.create<firrtl::FormalOp>(info.getLoc(), id, k);
+  auto &body = formal.getBody().emplaceBlock();
+
+  // Create the statement parser
+  hw::InnerSymbolNamespace ns(circuit);
+  SymbolTable circuitSymTbl(circuit);
```
uenoku, line 5213: SymbolTable computation walks entire top-level modules so it would eventually cause O(N^2) regression. Can you defer the actual parsing in a similar way to [Module parsing ](https://github.com/llvm/circt/blob/10180d3ca24f644dd6ef33868c1225a503ff57b1/lib/Dialect/FIRRTL/Import/FIRParser.cpp#L5569-L5577)? Basically we might want to separate the parser into `parseFormalBody` and `parseFormal`

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5153,6 +5180,47 @@ ParseResult FIRCircuitParser::parseModule(CircuitOp circuit, bool isPublic,
...
+    return emitError("Invalid parameter given to formal test: ") << kSpelling,
+           failure();
+
+  // Build out the firrtl mlir op
+  auto builder = circuit.getBodyBuilder();
+  auto formal = builder.create<firrtl::FormalOp>(info.getLoc(), id, k);
+  auto &body = formal.getBody().emplaceBlock();
+
+  // Create the statement parser
+  hw::InnerSymbolNamespace ns(circuit);
```
uenoku, line 5212:
```suggestion
  hw::InnerSymbolNamespace ns;
```

Maybe we can just use default constructor? I think this is a bit tricky since formal doesn't have InnerSymbolTable trait. For parsing I think it's ok to create a dummy table but we still need to check that there is no inner symbol produced within formal op

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
```diff
@@ -336,6 +336,45 @@ def MemOp : HardwareDeclOp<"mem", [DeclareOpInterfaceMethods<CombDataflow>]> {
...
+    OpBuilder<(ins "::mlir::Type":$elementType,
+                   "::mlir::StringAttr":$name), [{
+      $_state.addAttribute(getNameAttrName($_state.name), name);
+      $_state.addTypes(elementType);
+    }]>,
+    OpBuilder<(ins "::mlir::Type":$elementType,
+                   "::llvm::StringRef":$name), [{
+      return build($_builder, $_state, elementType, $_builder.getStringAttr(name));
+    }]>
+  ];
```
uenoku, line 371: I think these are not necessary anymore?

uenoku (commented):
Awesome! Parser and Op definitions generally look great. It would be also necessary to modify firrtl-spec. Regarding FIRRTL dialect representation there are several things to consider since FormalOp is a first operation which is not FModuleOp but could contain mots of FIRRTL ops:

1. InstanceGraph -- FormalOp has an instance op so we would need to create an edge from FormalOp to the referenced module. Probably need to implement ModuleOpInterface.
2. FModuleOp pass -- LowerIntrinsics, ExpandWhen etc. are currently FModuleOp pass so we need to find a way to run them on FormalOp

mmaloney-sf:
@uenoku commented:

> It would be also necessary to modify firrtl-spec.

Feel free to reach out to me if you want to offload the FIRRTL spec work to me.

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
```diff
@@ -36,10 +36,49 @@ class ReferableDeclOp<string mnemonic, list<Trait> traits = []> :
 /// located in a hardware-creating context, such as the body of a module.
 class HardwareDeclOp<string mnemonic, list <Trait> traits = []> :
   ReferableDeclOp<mnemonic, traits # [
-    ParentOneOf<[
+    ParentOneOf<[ 
       "firrtl::FModuleOp", "firrtl::LayerBlockOp",
       "firrtl::WhenOp", "firrtl::MatchOp", "sv::IfDefOp"]>]> {}
 
+def FormalOp : ReferableDeclOp<"formal", [
```
uenoku, line 43: Do we need ReferableDeclOp? I think FormalOp is one of top-level ops so InnerSymbol interface looks a bit weird to me. 

```
/// A declaration that can be referred to by it's name/identity.
class ReferableDeclOp<string mnemonic, list<Trait> traits = []> :
  FIRRTLOp<mnemonic, traits # [HasCustomSSAName,
                               DeclareOpInterfaceMethods<InnerSymbol,["getTargetResultIndex"]>,
                               FNamableOp]> {
```

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -3360,6 +3360,40 @@ void RegResetOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
   return forceableAsmResultNames(*this, getName(), setNameFn);
 }
 
+//===----------------------------------------------------------------------===//
+// FormalOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult
+FormalOp::verifySymbolUses(::mlir::SymbolTableCollection &symbolTable) {
+  auto referencedModule = symbolTable.lookupNearestSymbolFrom<FModuleLike>(
```
uenoku, line 3369: Do we want to allow an external module? I think it's reasonable to restrict this to FModuleOp.

uenoku (commented):
The new design looks fantastic! Thank you for iterating on this! The implementation looks great to me except for few nits regarding ReferableDeclOp.

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
```diff
@@ -36,10 +36,50 @@ class ReferableDeclOp<string mnemonic, list<Trait> traits = []> :
...
 class HardwareDeclOp<string mnemonic, list <Trait> traits = []> :
   ReferableDeclOp<mnemonic, traits # [
-    ParentOneOf<[
+    ParentOneOf<[ 
       "firrtl::FModuleOp", "firrtl::LayerBlockOp",
       "firrtl::WhenOp", "firrtl::MatchOp", "sv::IfDefOp"]>]> {}
 
+def FormalOp : FIRRTLOp<"formal", [
+    HasParent<"firrtl::CircuitOp">,
+    DeclareOpInterfaceMethods<SymbolUserOpInterface>,
```
uenoku, line 45: I think it needs Symbol as well.

uenoku (approved):
LGTM, can you check whether we should first change FIRRTL-spec or merge the PR?

PR #7375, opened 2024-07-23
[StringDebugInfoPred] Workaround FusedLoc bytecode issue

uenoku:
This PR changes to use a different `FusedLoc::get` to avoid a bug https://github.com/llvm/llvm-project/issues/99626.


mikeurbach (approved):
I am happy to avoid the issue in the bytecode reader, so this looks good to me. It's a little crazy that a different constructor with the same arguments, just in a different order, doesn't have this problem. Please leave a comment for future reference, since that might confuse the next person.

Issue #5272, opened 2023-05-26
circt-opt round-trip support (+add to CI?)


dtzSiFive:
https://github.com/llvm/llvm-project/commit/a6d09d4b1ac224ed90ee8ff6c964a2bff39421c7

Good for checking bytecode and textual forms.

Just trying to round-trip the MLIR in our lit tests has exposed bugs in the past, so this looks useful!

PR #7377, opened 2024-07-23
[CI][lit] Enable roundtrip tests in CI

uenoku:
An issue like llvm/llvm-project#99626 could have been found more easily with roundtrip tests in CI. This PR adds a round trip tests In CI. This could slightly increase the test time but it's ignorable compared to build time. 

c.f. https://github.com/llvm/llvm-project/commit/a6d09d4b1ac224ed90ee8ff6c964a2bff39421c7

Close https://github.com/llvm/circt/issues/5272

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -205,8 +205,6 @@ jobs:
       # --------
 
       - name: Build and Test CIRCT
-        env:
-          CIRCT_OPT_CHECK_IR_ROUNDTRIP: 1
         run: |
           mkdir build
           cd build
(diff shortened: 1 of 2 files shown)


uenoku:
https://github.com/llvm/circt/actions/runs/10054073138/job/27787994785?pr=7377 correctly catches https://github.com/llvm/circt/issues/7376 so it is working OK.

dtzSiFive (approved):
Awesome, thanks for tackling this! Been on the TODO for too long!

Changes LGTM!

Looks like this works for local testing as well, great!

uenoku:
Thank you for review!

PR #7381, opened 2024-07-23
[LLHD] Add TemporalCodeMotionPass

maerhart:
This pass can be run after early code motion and brings the IR in an easier from for desequentialization. Canonicalization should be run directly afterwards to significantly simplify the control flow. Once we also have the desequentialization pass some mem2reg pass for inout values, we should be able to transform most of the LLHD processes produced by the MooreToCore pass to pure HW/Comb/Seq IR suitable for simulation using Arcilator.


include/circt/Dialect/LLHD/Transforms/Passes.td
```diff
@@ -109,4 +109,21 @@ def EarlyCodeMotion : Pass<"llhd-early-code-motion", "hw::HWModuleOp"> {
...
 
+def TemporalCodeMotion : Pass<"llhd-temporal-code-motion", "hw::HWModuleOp"> {
+  let summary = "move drive operations to the exit basic block in processes";
+  let description = [{
+    This pass uses the temporal region analysis to transform the IR such that
+    every temporal region has a unique exit block and moves all 'llhd.drv'
+    operations in a temporal region into its exit block by adjusting the
+    enable operand.
+    Furthermore, it combines 'llhd.drv' operations driving the same signal with
+    with the same delay by multiplexing the driven value according to their
```
fabianschuiki, line 120:
```suggestion
    the same delay by multiplexing the driven value according to their
```

fabianschuiki (approved):
Really cool! :partying_face:

I'm wondering if instead of erroring out if the pass sees an input that has more than 2 TRs, the pass could just not perform the transformation and leave the process unmodified (or in a valid state). Simulations for example would still be able to run even if TCM did not run through, so it might be worthwhile not erroring out?

PR #7382, opened 2024-07-24
[ImportVerilog] Add HierarchicalNames.cpp to support hierarchical names.

hailongSun2000:
Based on the SystemVerilog IEEE Std 1800-2017 § 23.6 Hierarchical names. Hierarchical names are separated into **upward** and **downward**. For example:
```
module Top;
  int x;
  SubA subA();
  assign x = subA.a;  // upward: The Sub module's variable is used at the Top module.
endmodule

module SubA;
  int a;
  assign a = Top.x;  // downward: The Top module's variable is used at the Sub module.
endmodule
```
Therefore, we mark **upward** as `outputs` and **downward** as `inputs`, meanwhile, all hierarchical names are marked as `RefType`.

**Unsupported cases**:
However, we don't support hierarchical names invoked by two irrelevant modules at the same level. For example:
```
module A;
  int a = B.b;
endmodule

module B;
  int b = A.a;
endmodule
```
And we also don't support hierarchical names existing in the repeat modules. For example:
```
module Bar;
  SubC subC1();
  SubC subC2();
  int u = subC1.a;
  assign subC2.b = u;
endmodule

module SubC();
  int a, b;
endmodule
```


lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -549,13 +552,13 @@ LogicalResult Context::convertCompilation() {
   }
 
   // Prime the root definition worklist by adding all the top-level modules.
-  SmallVector<const slang::ast::InstanceSymbol *> topInstances;
```
hailongSun2000, line 552: unused container.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -722,7 +725,6 @@ Context::convertModuleBody(const slang::ast::InstanceBodySymbol *module) {
   builder.setInsertionPointToEnd(lowering.op.getBody());
 
   // Convert the body of the module.
-  ValueSymbolScope scope(valueSymbols);
   for (auto &member : module->members()) {
     auto loc = convertLocation(member.location);
     if (failed(member.visit(ModuleVisitor(*this, loc))))
```
hailongSun2000, line 730: From my perspective, when we create `scope` at traversing the `members` of the `instanceBody`, the variables stored int `valueSymbols` cannot be used by hierarchical name. An error like 
```
/home/phoenix/work/HDLBits/test09.sv:19:15: error: unknown name `data`
  assign w0 = sub1.data;
              ^
/home/phoenix/work/HDLBits/test09.sv:4:15: note: no value generated for Variable
  logic [7:0] data;
```
However, I moved the location of creating `scope` to the top/outermost level, and this error still exists.
I don't know how to deal with it. I need @fabianschuiki help. Thanks in advance :smiley:!

hailongSun2000: For the above example, we have the top module `m`, and submodule `m1`(instance `n`). I think we should create `scope` to contain 
```
        "name": "m",
        "kind": "InstanceBody",
        "addr": 2199025897488,
        "members": [
          {
            "name": "n",
            "kind": "Instance",
         .......
```
, then we can handle hierarchical values. But now, we will create `scope` when we visit `InstanceBody`.

hailongSun2000: Maybe should we need a new op to represent the hierarchical name, rather than get it from `valueSymbols`?

fabianschuiki:
Really cool seeing this PR touching a really annoying and difficult corner of SystemVerilog :partying_face:!

I agree with you @hailongSun2000: simply accessing hierarchical names through the `valueSymbols` table is not going to work well. One of the reasons is that you cannot directly access an SSA value defined in another module, which is why we use the scopes with `valueSymbols` to only make SSA values visible within the same block. I'm pretty sure we need an entirely different mechanism for hierarchical paths.

One thing that would be great is if our solution to hierarchical paths wouldn't just magically refer to something defined in another module, but if we actually used SSA values (and maybe special module ports) to represent the hierarchical path. For example, when you see a hierarchical path `a.b.c`, it's non-trivial to figure out whether `a` refers to an instance in the current module (and you're accessing the hierarchy _downwards_), or if `a` is one of the current module's parents (and you're accessing the hierarchy _upwards_). It would be great if our mechanism for hierarchical paths in the IR would make this obvious and unambiguous, for example by passing around hierarchical pointers or references to what is being accessed.

@dtzWill has done a lot of awesome work on FIRRTL references/probes that we could draw a lot of inspiration from :1st_place_medal:. We could take a similar approach and pass around references/pointers in the hierarchy:

```verilog
module A;
  B b();
  assign b.c.y = b.c.x + 1;
endmodule

module B;
  C c();
endmodule

module C;
  int x = 42;
  int y;
endmodule
```

```mlir
moore.module @A() {
  %b.href.c.x, %b.href.c.y = moore.instance "b" @B() -> (href.c.x: !moore.ref<i32>, href.c.y: !moore.ref<i32>)
  %0 = moore.read %b.href.c.x : <i32>  // read `b.c.x`
  %1 = moore.constant 1 : i32
  %2 = moore.add %0, %1 : i32
  moore.assign %b.href.c.y, %2 : <i32> // assign `b.c.y`
}

moore.module @B(
  out %href.c.x : !moore.ref<i32>,
  out %href.c.y : !moore.ref<i32>
) {
  %c.href.x, %c.href.y = moore.instance "c" @C() -> (href.x: !moore.ref<i32>, href.y: !moore.ref<i32>)
  moore.output %c.href.x, %c.href.y
}

moore.module @C(
  out %href.x : !moore.ref<i32>,
  out %href.y : !moore.ref<i32>
) {
  %0 = moore.constant 42 : i32
  %x = moore.variable %0 : i32
  %y = moore.variable : i32
  moore.output %x, %y  // return references to the hierarchically-accessed values
}
```

A potentially very nice part of materializing hierarchical references as actual references being passed around the hierarchy is that we may be able to write optimization passes later that turn these hierarchical references into regular ports. In the example above, we may later on figure out that `b.c.x` is only read, and `b.c.y` is only written, and therefore convert them accordingly:
- `out %href.c.x : !moore.ref<i32>` converted to `out %href.c.x : i32`
- `out %href.c.y : !moore.ref<i32>` converted to `in %href.c.y : i32`

To implement something like the reference-passing above, we would already have to know which variables are accessed via hierarchical paths when we convert a module body. Because if a variable _is_ accessed through a hierarchical path, we have to create additional ports to make sure we can pass its reference around. To know this, we may have to do a pre-pass over the entire AST to find all hierarchical names, and determine what exactly they access and through which modules the references have to pass. In the example above, the path `b.c.x` means that module `C` has to make `x` available as a ref port, but it also means that module `B` has to pass the reference through to its parent. So in a sense we need to precompute information about all hierarchical names and which paths they take through the hierarchy. And then teach the module body lowering to look at that information and generate the necessary ports and ops.

What do you think about that @hailongSun2000, @dtzSiFive?

hailongSun2000:
It looks like so reasonable! Let me get this straight. 
> To know this, we may have to do a pre-pass over the entire AST to find all hierarchical names, and determine what exactly they access and through which modules the references have to pass. 

This is the first step. 
Then make these hierarchical names as the outputs of the instance and module. 
And we must be careful with hierarchy _downwards_ or _upwards_

hailongSun2000:
A lot of cases are not considered. Such as `Top.sub.x = a`, `sub.x = a` & ` b = sub.x`, etc...
I just want to know whether my thought is correct. Only if it's right can I continue. @fabianschuiki Thanks for your guidance in advance :heart:!

hailongSun2000:
I think I cannot mark `Top.sub.var`(`Top.sub.var = x`) as an input port, because we cannot assign a value to an input port :sweat_smile:. I'll tweak it.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -688,6 +722,31 @@ Context::convertModuleHeader(const slang::ast::InstanceBodySymbol *module) {
...
+    if (auto hierName = hierPath.second.exportPath) {
+      hierPath.second.idx = outputIdx++;
+      modulePorts.push_back({*hierName, hierType, hw::ModulePort::Output});
+    }
+
+    if (auto hierName = hierPath.second.importPath) {
+      hierPath.second.idx = inputIdx++;
+      modulePorts.push_back({*hierName, hierType, hw::ModulePort::Input});
+      block->addArgument(hierType, hierLoc);
+    }
```
fabianschuiki, line 748: If I understand correctly, a hiername is marked as `exportPath` or `importPath` depending on whether it is used as an lvalue or rvalue (or both). This means you can get a hiername that is both imported and exported if some module both reads _and_ writes it.

I think you can get rid of that bidirectionality by _always_ passing variables targeted by hiernames around as ref types. So when a module reads and/or writes a variable, you could make the `ref<T>` of that variable available in the module (either as an input port of `ref<T>` type, or as an output port of a submodule of `ref<T>` type). Then you can both read and write from that ref.

A side-effect of this would be that a module never both imports _and_ outputs a variable at the same time. Instead, a module imports a variable ref if it is needed for a hiername locally or in a submodule, and it exports a variable ref if it is used in a hiername in one of the parent modules. But never both. So instead of having separate `exportPath` and `importPath` fields, your hierinfo could just contain a `direction` field that indicates whether it is being imported or exported by a module 😃. That would allow you to support all the weird use cases of hiernames! And we can add an optimization pass later on that converts `ref<T>` ports to `T` ports where possible.

hailongSun2000: Yeah, you're right! The `importPath` corresponds to an rvalue and `exportPath` corresponds to an lvalue, if these two exist, it will be marked as the ref port.

Great idea! Mark all hierarchical names as `ref` ports, and introduce a `direction` to indicate whether it's being imported or exported by a module. I'll apply this thought 🚀 !

hailongSun2000: I don't add a `direction` field to indicate the direction of a hierarchical name. We don't have to use it when we create a module or an instance. WDYT @fabianschuiki?

hailongSun2000: In order to collect all hierarchical names, we traverse the whole AST detailedly. The same traversing exists when creating Moore ops. Hierarchical names are involved everywhere, such as `foo.a = foo.arr[foo.b]`(maybe this case will be nested in `for{ if{always...}}`). So we must add different `visit(expr/stmt/node)`s for different statements, expressions, etc, which is similar to creating Moore ops.

test/Conversion/ImportVerilog/hierarchical-names.sv
```diff
@@ -0,0 +1,67 @@
...
+  // CHECK: %s = moore.variable %0 : <i32>
+  int s = subA.subB.y;
+  // CHECK: %r = moore.variable : <i32>
+  int r;
+  // CHECK: %1 = moore.read %r : <i32>
+  // CHECK: moore.assign %subA.subB.x, %1 : i32
+  assign subA.subB.x = r;
+  // CHECK: moore.instance "subA" @SubA(subB.y: %subA.subB.y: !moore.ref<i32>, subB.x: %subA.subB.x: !moore.ref<i32>) -> ()
+  SubA subA();
+endmodule
```
fabianschuiki, line 17: Hmmmm, I'm not sure we can make all hierarchical paths be ref-typed inputs to the top-level module. That begs the question where those references would come from. Does the user of the `Foo` top-level module have to provide a declaration for every single variable involved in a hierarchical path in the design, and pass that into the top-level?

Could we instead make `moore.instance "subA"` produce the `%subA.subB.x` ref value as an output? That way the declaration of the variables can stay where they are, and the modules/instances would just produce a reference to those declarations to be used in hierarchical paths.

hailongSun2000: Yeah, you're right! If we make hierarchical names be ref-typed inputs to the top-level module, it looks like we need to pass values corresponding to these inputs.
However, if we make them be outputs, for example:
```
  assign subA.subB.x = r;
  %subA.subB.x = moore.instance "subA" 
```
It looks like we assign `r` to an output(subA.subB.x). I implemented this form before, but I think assigning a value to an output is so strange, Haha.

fabianschuiki: 😃 It's like a function that returns a pointer in C 😏:
```c
int *x = someFunction();
*x = 42;
```
So you don't get the value itself as an output from "subA", but _a reference_ to the value. And because you have a reference to it, you can use it on the LHS of an assignent 😃

hailongSun2000: It makes sense!

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -694,6 +709,20 @@ Context::convertModuleHeader(const slang::ast::InstanceBodySymbol *module) {
...
+    if (!hierType)
+      return {};
+
+    // All hierarchical names are regarded as output ports with ref<T> types.
+    if (auto hierName = hierPath.hierName) {
+      hierType = moore::RefType::get(cast<moore::UnpackedType>(hierType));
+      hierPath.idx = outputIdx++;
+      modulePorts.push_back({hierName, hierType, hw::ModulePort::Output});
+    }
+  }
```
hailongSun2000, line 725: From yesterday to now, I've always solved an error that `hierPath.idx` is empty. Due to the acquiescently deduplicate module/instance, we cannot record **subC2**. In other words, `hierPaths[subC2]` is empty, therefore, `hierPath.idx` is also empty. For example:
```
module Bar;
  SubC subC1();
  SubC subC2();
  int u = subC1.subD.z;
  assign subC2.subD.z = u;
endmodule
```

test/Conversion/ImportVerilog/hierarchical-names.sv
```diff
@@ -0,0 +1,59 @@
...
+module Foo;
+  // CHECK: %subA.subB.y, %subA.subB.x = moore.instance "subA" @SubA() -> (subB.y: !moore.ref<i32>, subB.x: !moore.ref<i32>)
+  SubA subA();
+  // CHECK: [[RD_SA_SB_Y:%.+]] = moore.read %subA.subB.y : <i32>
+  int s = subA.subB.y;
+  int r;
+  // CHECK: [[RD_R:%.+]] = moore.read %r : <i32>
+  // CHECK: moore.assign %subA.subB.x, [[RD_R]] : i32
+  assign subA.subB.x = r;
+endmodule
```
hailongSun2000, line 15: After we marked all hierarchical names as output, we can't handle the case like 
```
int s = subA.subB.y; or int r; assign subA.subB.x = r;
SubA subA();
```
Because when we handle `int s...`, the `valueSymbol` doesn't record `subA.subB.y` or `subA.suB.x` which is generated by `moore.instance`. However, this rule is supported by SV IEEE Std.

lib/Conversion/ImportVerilog/Structure.cpp
hailongSun2000: By the way, when I use `llvm::errs()<<module->parentInstance->name` before executing deduplicate, the terminator will display `subC1` and `subC2`. If after executing deduplicate, it will display `subC1`, empty :thinking:.

fabianschuiki: Hmmm I don't think you can make all hierarchical paths output ports. I think there are two different kinds of hierarchical references a module has to deal with:

- **Imported references.** If you are inside a module `Top.subA.subB` and the module uses the hierarchical name `Top.subA.x`, then a reference to `x` is imported as an input ref port. The parent module has to know about this and has to forward the reference to `x` to its child `subB`.
- **Exported references.** If you are inside a module `Top.subA.subB` and the parent module accesses the hierarchical name `subB.x`, then a reference to `x` is exported as an output ref port. The parent module then uses that reference for its access to `subB.x`.

Basically, for every module in the design you need a list of hierarchical references that need to be imported or exported by that module. During your initial traversal of the design where you visit all hierarchical paths, you will need to iterate through every element of a hierarchical path, look up the corresponding module, and add an entry to its list of hierarchical references. Depending on whether the hierarchical path goes downwards into a child module, or starts further up in a parent, you will have to mark the hierarchical reference as either import or export. (Children need to export references as output ports to their parents. Hierarchical paths that start further up in the hierarchy have to first export references upwards to the root of the path, and then import that references down to the hierarchical path use.)

hailongSun2000: Really appreciate for the detailed explanation.

test/Conversion/ImportVerilog/hierarchical-names.sv
```diff
@@ -0,0 +1,69 @@
...
+module Bar(input a, b,
+           output c);
+  // CHECK: %subC1.c, %subC1.subD.z = moore.instance "subC1" @SubC(a: %0: !moore.l1, b: %1: !moore.l1) -> (c: !moore.l1, subD.z: !moore.ref<i32>)
+  SubC subC1(a, b, c);
+  // CHECK: %subC2.c, %subC2.subD.z = moore.instance "subC2" @SubC(a: %2: !moore.l1, b: %3: !moore.l1) -> (c: !moore.l1, subD.z: !moore.ref<i32>)
+  SubC subC2(a, b, c);
+  // CHECK: [[RD_SC1_SD_Z:%.+]] = moore.read %subC1.subD.z : <i32>
+  // CHECK: moore.variable [[RD_SC1_SD_Z]] : <i32>
+  int u = subC1.subD.z;
+  // assign subC2.subD.z = u;
```
hailongSun2000, line 52: I'll fix this later.

hailongSun2000:
Hey, @fabianschuiki! I update this PR! Please check it, thanks in advance :smiley:!

fabianschuiki (approved):
Really really cool! Thanks a lot for working this out :partying_face: And it's great to have all those tests around! :100:

PR #7383, opened 2024-07-24
(WIP) [OM] Rework ClassOp to use fields terminator

leonardt:
Addresses https://github.com/llvm/circt/issues/7078


leonardt:
Squashed in https://github.com/llvm/circt/pull/7537

PR #7386, opened 2024-07-25
[Arc][Sim] Lower Sim DPI func to func.func and support dpi call in Arc

uenoku:
This PR implements initial support for lowering Sim DPI operations to Arc. 

* sim::LowerDPIFuncPass implements lowering from `sim.dpi.func` to `func.func` that respects C-level ABI. 
* acc::LowerStatePass is modified to allocate states and call functions for `sim.dpi.call` op. 

In the follow-up I'll rename `sim.dpi.call` to `sim.func.call` since now `sim.dpi.call` accepts both `sim.dpi.func` and `func.func`.


lib/Dialect/Arc/ArcOps.cpp
```diff
@@ -142,25 +142,28 @@ void DefineOp::print(OpAsmPrinter &p) {
...
+  //   // printed. This can be switched of when creating the context, but one
+  //   // might not want to switch that off for other error messages. Here it's
+  //   // definitely not desirable as arcs can be very big and would fill up the
+  //   // error log, making it hard to read. Currently, only the signature
+  //   (first
+  //   // line) of the arc is printed.
+  //   auto diag = mlir::emitError(getLoc(), "body contains non-pure
+  //   operation"); diag.attachNote(op.getLoc()).append("first non-pure
+  //   operation here: "); return diag;
+  // }
```
uenoku, line 166: Clearly it's not correct. How can we model a state transfer function that has side-effect? Do we need a new op?

sequencer:
I think in the future. We may require the ClockGen intrinsic for triggering clock and reset. This is the only Verilog simulation blackbox that t1 use in the Verilator/VCS. It should be lowered to both Verilog and ArcOp.

integration_test/arcilator/JIT/dpi.mlir
```diff
@@ -0,0 +1,39 @@
...
+// CHECK:      c = 0
+// CHECK-NEXT: c = 5
+sim.func.dpi @dpi(in %a : i32, in %b : i32, out c : i32) attributes {verilogName = "adder_func"}
+func.func @adder_func(%arg0: i32, %arg1: i32, %arg2: !llvm.ptr) {
+  %0 = arith.addi %arg0, %arg1 : i32
+  llvm.store %0, %arg2 : i32, !llvm.ptr
+  return
+}
+hw.module @adder(in %clock : i1, in %a : i32, in %b : i32, out c : i32) {
+ %seq_clk = seq.to_clock %clock
```
maerhart, line 13: Nit:
```suggestion
  %seq_clk = seq.to_clock %clock
```

lib/Dialect/Sim/Transforms/LowerDPIFunc.cpp
```diff
@@ -0,0 +1,176 @@
...
+// Pass Implementation
+//===----------------------------------------------------------------------===//
+
+namespace {
+
+struct LoweringState {
+  DenseMap<StringAttr, func::FuncOp> dpiFuncDeclMapping;
+  circt::Namespace nameSpace;
+};
+struct LowerDPIFuncPass : public sim::impl::LowerDPIFuncBase<LowerDPIFuncPass> {
```
maerhart, line 60: Nit:
```suggestion
};

struct LowerDPIFuncPass : public sim::impl::LowerDPIFuncBase<LowerDPIFuncPass> {
```

lib/Dialect/Sim/Transforms/LowerDPIFunc.cpp
```diff
@@ -0,0 +1,176 @@
...
+  if (auto verilogName = simFunc.getVerilogName()) {
+    func = symbolTable.lookup<func::FuncOp>(*verilogName);
+    // TODO: Check if function type matches.
+  }
+
+  // If a referred function is not in the same module, create an external
+  // function declaration.
+  if (!func) {
+    func = builder.create<func::FuncOp>(simFunc.getVerilogName()
+                                            ? *simFunc.getVerilogName()
```
maerhart, line 106: Nit: these two cases do not seem to be covered with tests

lib/Dialect/Sim/Transforms/LowerDPIFunc.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+  // Map old symbol to a new func op.
+  loweringState.dpiFuncDeclMapping[simFunc.getSymNameAttr()] = funcOp;
+
+  builder.setInsertionPointToStart(funcOp.addEntryBlock());
+  SmallVector<Value> functionInputs;
+  SmallVector<LLVM::AllocaOp> functionOutputAllocas;
+
+  size_t inputIndex = 0;
+  for (auto [idx, arg] : llvm::enumerate(moduleType.getPorts())) {
```
maerhart, line 126: Isn't `idx` unused?

maerhart (approved):
LGTM! Really cool that arcilator can deal with DPI calls now! :tada:

lib/Dialect/Sim/SimOps.cpp
```diff
@@ -69,12 +70,12 @@ ParseResult DPIFuncOp::parse(OpAsmParser &parser, OperationState &result) {
...
 sim::DPICallOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
-  auto referencedOp = dyn_cast_or_null<sim::DPIFuncOp>(
-      symbolTable.lookupNearestSymbolFrom(*this, getCalleeAttr()));
+  auto referencedOp =
+      symbolTable.lookupNearestSymbolFrom(*this, getCalleeAttr());
   if (!referencedOp)
     return emitError("cannot find function declaration '")
            << getCallee() << "'";
-  return success();
+  return success(isa<func::FuncOp, sim::DPIFuncOp>(referencedOp));
```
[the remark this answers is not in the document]
uenoku: Thanks, added a message and test.

lib/Dialect/Sim/Transforms/LowerDPIFunc.cpp
uenoku: Added in lower-dpi.mlir, thanks!

integration_test/arcilator/JIT/dpi.mlir
```diff
@@ -0,0 +1,39 @@
...
+sim.func.dpi @dpi(in %a : i32, in %b : i32, out c : i32) attributes {verilogName = "adder_func"}
+func.func @adder_func(%arg0: i32, %arg1: i32, %arg2: !llvm.ptr) {
+  %0 = arith.addi %arg0, %arg1 : i32
+  llvm.store %0, %arg2 : i32, !llvm.ptr
+  return
+}
+hw.module @adder(in %clock : i1, in %a : i32, in %b : i32, out c : i32) {
+  %seq_clk = seq.to_clock %clock
+
+  %0 = sim.func.dpi.call @dpi(%a, %b) clock %seq_clk : (i32, i32) -> i32
```
fabianschuiki, line 15: Just out of curiosity: since the `sim.func.dpi.call` function seems like it accepts both `sim.func.dpi` _and_ `func.func` callables, would this code also work with a second `sim.func.dpi.call` that calls the `@adder_func` directly?

fabianschuiki (approved):
This is really really cool! Thanks for also adding an integration test with Arcilator's JIT to actually give this a practical check 🙇 🙌 🥳

integration_test/arcilator/JIT/dpi.mlir
uenoku: Directly calling `@adder_func` doesn't work since it's output value is passed by reference in argument. If `@adder_func` was a normal dataflow-ish func.func such as `func.func @adder_func_ok(%arg0: i32, %arg1: i32) -> i32`, we can call it directly.

fabianschuiki: Ooooh you're right, sorry, I overlooked that difference in the operand layout. Makes sense!

PR #7387, opened 2024-07-25
[ESI Runtime] Port connect: add optional buffer size arg

teqdruid:
It will not be unusual for customers to want to customize the size of the DMA host memory buffer. This provides the ability to do that, though none of the packaged backends use it.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
+++ b/lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
@@ -35,7 +35,12 @@ public:
   ChannelPort(const Type *type) : type(type) {}
   virtual ~ChannelPort() { disconnect(); }
 
-  virtual void connect() { connectImpl(); }
+  /// Set up a connection to the accelerator. The buffer size is optional and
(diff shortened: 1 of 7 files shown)


lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
```diff
@@ -35,7 +35,12 @@ class ChannelPort {
   ChannelPort(const Type *type) : type(type) {}
   virtual ~ChannelPort() { disconnect(); }
 
-  virtual void connect() { connectImpl(); }
+  /// Set up a connection to the accelerator. The buffer size is optional and
+  /// should be considered merely a hint. Individual implementations use it
+  /// however they like. The unit is number of messages of the port type.
+  virtual void connect(std::optional<size_t> bufferSize = std::nullopt) {
```
mortbopet, line 41: Suggesting changing from size_t to unsigned; size_t is usually used to indicate the size of something in memory - which, as the comment indicates, this isn't.

mortbopet (approved):

PR #7391, opened 2024-07-26
[ImportVerilog] Support large integer constants

fabianschuiki:
Remove the 64 bit limitation on SV integer constants and allow for arbitrary Slang `SVInt`s to be converted to MLIR `APInt`s.


uenoku (approved):
Awesome!

PR #7393, opened 2024-07-27
[ImportVerilog] Support for loop variables

fabianschuiki:
Add support for variables declared inside the for loop initializer, such as `for (int x = 0; ...)`. This is pretty straightforward, since Slang already creates corresponding variable declaration AST nodes in the surrounding scope. The only change needed is removing an error diagnostic.


maerhart (approved):
Cool that this doesn't require any additional work!

PR #7394, opened 2024-07-27
[Arc] Make the canonicalizer shuffle the input vector elements before merging

elhewaty:


test/Dialect/Arc/arc-canonicalizer.mlir
```diff
@@ -453,22 +453,14 @@ in %clock: !seq.clock, in %o: i8, in %v: i8, in %q: i8, in %s: i8) {
...
-// CHECK-NEXT:    [[VEC1:%.+]]:4 = arc.vectorize ([[VEC0]]#1, [[VEC0]]#0, [[VEC0]]#2, [[VEC0]]#3), (%n, %p, %r, %t) : (i8, i8, i8, i8, i8, i8, i8, i8) -> (i8, i8, i8, i8) {
-// CHECK-NEXT:    ^[[BLOCK:[[:alnum:]]+]](%arg0: i8, %arg1: i8):
-// CHECK-NEXT:      [[OUT:%.+]] = comb.and %arg0, %arg1 : i8
-// CHECK-NEXT:      arc.vectorize.return [[OUT]] : i8
-// CHECK-NEXT:    }
-// CHECK-NEXT:    [[VEC2:%.+]]:4 = arc.vectorize ([[VEC1]]#1, [[VEC1]]#0, [[VEC1]]#2, [[VEC1]]#3), (%o, %v, %q, %s) : (i8, i8, i8, i8, i8, i8, i8, i8) -> (i8, i8, i8, i8) {
-// CHECK-NEXT:    ^[[BLOCK:[[:alnum:]]+]](%arg0: i8, %arg1: i8):
-// CHECK-NEXT:      [[OUT:%.+]] = arc.call @Just_A_Dummy_Func(%arg0, %arg1) : (i8, i8) -> i8
-// CHECK-NEXT:      arc.vectorize.return [[OUT]] : i8
+// CHECK-NEXT:    [[VEC:%.+]]:4 = arc.vectorize (%b, %e, %h, %k), (%c, %f, %i, %l), (%n, %p, %r, %t), (%o, %v, %q, %s) : (i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8, i8) -> (i8, i8, i8, i8) {
```
maerhart, line 456: You need to make sure that all input vectors are re-ordered the same way. As we see here, the vector `(%n, %p, %r, %t)` didn't change, but should be `(%p, %n, %r, %t)` after the optimization.

elhewaty: Oops, I didn't notice this, I will think of a way to do this, maybe bind the operands to index and shuffle them?

We will need more optimizations here to reduce the shuffle cost. 
I think this optimization will affect the canonicalization of `keeping one vector if the pattern is repeated.` as it will treat `[a, b, c, d]` differently from `[a, c, b, d]`.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -619,18 +619,58 @@ MergeVectorizeOps::matchAndRewrite(VectorizeOp vecOp,
...
       newOperands.insert(newOperands.end(), inputVec.begin(), inputVec.end());
       continue;
     }
+
+    // Here, all elements are from the same `VectorizeOp`.
+    // If all elements of the input vector come from the same `VectorizeOp`
+    // sort the vectors by their indices
+    DenseMap<Value, size_t> resultIdxMap;
+    for (auto [resulrIdx, result] : llvm::enumerate(otherVecOp.getResults()))
+      resultIdxMap[result] = resulrIdx;
```
fabianschuiki, line 637: Nit: typo
```suggestion
    for (auto [resultIdx, result] : llvm::enumerate(otherVecOp.getResults()))
      resultIdxMap[result] = resultIdx;
```

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -619,18 +619,58 @@ MergeVectorizeOps::matchAndRewrite(VectorizeOp vecOp,
...
+
+    SmallVector<Value> tempVec(inputVec.begin(), inputVec.end());
+    llvm::sort(tempVec, [&](Value a, Value b) {
+      return resultIdxMap[a] < resultIdxMap[b];
+    });
+
+    // Check if inputVec matches the result after sorting.
+    if (tempVec != SmallVector<Value>(otherVecOp.getResults().begin(),
+                                      otherVecOp.getResults().end())) {
```
fabianschuiki, line 646: Does `tempVec != otherVecOp.getResults()` not work here?

fabianschuiki (approved):
Very neat! Cool to see that the shuffling allows you to merge more vectorize ops :smiley:.

Can you check whether this works on one of the cores in `circt/arc-tests`? Once it does I think this is ready to go! :partying_face:

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
elhewaty: I will try, but I think `getResults()` returns `result_range` which cannot be converted into a `SmallVector<Value>` but is worth trying!

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
elhewaty: Oh, I didn't notice this :sob:  thanks

elhewaty:
> Can you check whether this works on one of the cores in `circt/arc-tests`? Once it does I think this is ready to go! 🥳

The canonicalizer crashes :cry:, I will test it before pushing but I need to reduce the test case.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
fabianschuiki: There should be an implicit conversion from `SmallVector<Value>` to an iterator range, which should theoretically be able to compare against a result range. Worth trying, don't worry if it doesn't work :smiley:

elhewaty: I didn't work!
```
 error: invalid operands to binary expression ('SmallVector<mlir::Value>' and '::mlir::Operation::result_range' (aka 'mlir::ResultRange'))
    if (tempVec != otherVecOp.getResults()) 
```

fabianschuiki: Thanks for checking :smiley:

elhewaty: @fabianschuiki, should I land this now?

elhewaty: It's 12:17 AM here, so please land it if it's Okay.

PR #7401, opened 2024-07-29
[FIRRTL] default layer specialization

youngar:


lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
```diff
@@ -227,14 +229,14 @@ struct SpecializeLayers {
...
-          return true;
-        }
+      auto specialization = getSpecialization(oldRoot, oldNestedRefs);
+      if (specialization == LayerSpecialization::Enable) {
+        // We are enabling this layer, the next non-enabled layer should
+        // include this layer's name as a prefix.
+        prefix.append(ref.getValue());
+        prefix.append("_");
+        return true;
+      } else if (specialization == LayerSpecialization::Disable) {
```
seldridge, line 239: Nit: the `return` makes the `else` for this `if` unnecessary.

test/Dialect/FIRRTL/specialize-layers.mlir
```diff
@@ -29,6 +29,15 @@ firrtl.circuit "DisableTwice" attributes {
...
 }
 
+// CHECK-LABEL: firrtl.circuit "DefaultSpecialization"
+firrtl.circuit "DefaultSpecialization" attributes {
+    // This option does not get removed - it is idempotent to leave it.
+    // CHECK: default_layer_specialization = #firrtl<layerspecialization enable>
+    default_layer_specialization = #firrtl<layerspecialization enable>
+  } {
+  firrtl.extmodule @DefaultSpecialization()
+}
```
seldridge, line 39: I agree with this. It originally seemed incongruous to remove the attributes for enabling/disabling, but leave this one. However, this one does make sense to leave in.

lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
```diff
@@ -753,8 +752,22 @@ struct SpecializeLayersPass
...
+    if (auto specialization = circuit.getDefaultLayerSpecialization()) {
+      if (*specialization != LayerSpecialization::None) {
+        shouldSpecialize = true;
+        defaultSpecialization = *specialization;
+      }
+    }
+
+    // If we did not transform the circuit the circuit, return early.
+    // TODO: if both arrays are empty we could preserve specific analyses, but
+    // not all analyses since we have modified the circuit op.
```
seldridge, line 765: Duplicate `the circuit` in comment.

seldridge (approved):
The one thing that may be weird about this is that there are now a lot of different ways to do the same thing given that the options include both the ability to enable and disable layers as well as to set the default behavior. Use of mutually exclusive options instead of composable options may be simpler for users: (1) enable all, (2) disable all, and (3) enable specific.

lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
youngar: I wonder why this did not get caught by clang-tidy 🤔.

dtzSiFive (approved):
LGTM!

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -35,7 +35,8 @@ def CircuitOp : FIRRTLOp<"circuit",
     StrAttr:$name,
     DefaultValuedAttr<AnnotationArrayAttr, "{}">:$annotations,
     OptionalAttr<SymbolRefArrayAttr>:$enable_layers,
-    OptionalAttr<SymbolRefArrayAttr>:$disable_layers
+    OptionalAttr<SymbolRefArrayAttr>:$disable_layers,
+    OptionalAttr<LayerSpecializationAttr>:$default_layer_specialization
```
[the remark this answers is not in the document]
youngar: I made the attribute only have two cases, enable and disable, and kept it optional on the circuit.  I made the command line argument use a different 3 value enum, with None, Enable, Disable.  Unfortunately the `cl` framework doesn't seem to have magical handling for std::optional, and I would have had to use a custom parser to make `std::optional<LayerSpecialization>` work - but using a new 3 value enum value gives better help descriptions.

PR #7402, opened 2024-07-30
[FIRRTL] Convert FModule passes to FModuleLike interface passes

dobios:
This PR is derived from and depends on #7374 and replaces all FModuleOp passes with FModuleLike interface passes, allowing for arbitrary hardware to exist outside of modules.


lib/Dialect/FIRRTL/Transforms/LowerIntrinsics.cpp
```diff
@@ -53,14 +66,14 @@ LogicalResult LowerIntrinsicsPass::initialize(MLIRContext *context) {
...
-  auto result = lowering->lower(getOperation());
-  if (failed(result))
-    return signalPassFailure();
-
-  numConverted += *result;
 
-  if (*result == 0)
-    markAllAnalysesPreserved();
+  TypeSwitch<Operation *>(&(*getOperation()))
+      .Case<FModuleOp, FExtModuleOp, FIntModuleOp, FMemModuleOp, ClassOp,
```
darthscsi, line 71: The copy-and-pasting of this block is increasing the ops being visited by a lot of passes.  Very few passes deal with FExtModuleOp, for example, as they don't have bodies.

lib/Dialect/FIRRTL/Transforms/EliminateWires.cpp
```diff
@@ -45,54 +45,61 @@ namespace {
...
+    }
   }
+};
+} // end anonymous namespace
+
+void EliminateWiresPass::runOnOperation() {
+  LLVM_DEBUG(debugPassHeader(this) << "\n";);
+
+  TypeSwitch<Operation *>(&(*getOperation()))
+      .Case<FModuleOp, FExtModuleOp, FIntModuleOp, FMemModuleOp, ClassOp,
```
darthscsi, line 100: Do all these have wires in them?

darthscsi (changes requested):
FModuleLike doesn't mean "has logic in it".  Filtering ops later (and less efficiently) in the runtime may exacerbate parallel task balance issues.  Assuming a new top-level op which can contain logic makes sense (I haven't looked at that), the way to deal with it is not to make passes operate on FModuleLike.  (I also have looked to see if the new construct should even be FModuleLike.  Can it be instantiated with instance ops?)

dobios:
This is no longer needed since the FormalOp redesign.

PR #7403, opened 2024-07-30
[ImportVerilog] Support for String Types, String Literals

wenhu1024:
Support for String Types, String Literals.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -454,6 +454,21 @@ def NamedConstantOp : MooreOp<"named_constant", [
   }];
 }
 
+def StringConstantOp : MooreOp<"string", [Pure, ConstantLike]> {
+  let summary = "Produce a constant string value";
+  let description = [{
+    Produces a constant value of string type.
+
+    Example:
+    ```mlir
+    %0 = moore.string "hello world"
+    ```
+  }];
+  let arguments = (ins StrAttr:$value);
+  let results = (outs SimpleBitVectorType:$result);
+  let assemblyFormat = "$value attr-dict `:` type($result)";
+}
+
 def ConversionOp : MooreOp<"conversion", [Pure]> {
   let summary = "A type conversion";
   let description = [{
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -454,7 +454,7 @@ def NamedConstantOp : MooreOp<"named_constant", [
   }];
 }
 
-def StringConstantOp : MooreOp<"string", [Pure, ConstantLike]> {
+def StringConstantOp : MooreOp<"string_constant", [Pure, ConstantLike]> {
   let summary = "Produce a constant string value";
   let description = [{
     Produces a constant value of string type.
@@ -465,7 +465,7 @@ def StringConstantOp : MooreOp<"string", [Pure, ConstantLike]> {
     ```
   }];
   let arguments = (ins StrAttr:$value);
-  let results = (outs SimpleBitVectorType:$result);
+  let results = (outs IntType:$result);
   let assemblyFormat = "$value attr-dict `:` type($result)";
 }
 
--- a/test/Conversion/ImportVerilog/types.sv
+++ b/test/Conversion/ImportVerilog/types.sv
@@ -146,3 +146,9 @@ module Typedefs;
   myType1 v0;
   myType2 v1;
 endmodule
+
+// CHECK-LABEL: moore.module @String
+module String;
+  // CHECK-NEXT: %s = moore.variable : <string>
+  string s;
+endmodule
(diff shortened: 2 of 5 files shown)


test/Conversion/ImportVerilog/types.sv
```diff
@@ -146,3 +146,9 @@ module Typedefs;
   myType1 v0;
   myType2 v1;
 endmodule
+
+// CHECK-LABEL: moore.module @String
+module String;
+  // CHECK-NEXT: %s = moore.variable : <string>
+  string s;
+endmodule
```
hailongSun2000, line 154: nit: Could you add more tests for string type? Like `string s = "Welcome to Moore"` and `string s1; assign s1 = "Hello World"`. We have `basic.sv` to test expressions, statements, etc. And `types.sv` to test different types as you write.

wenhu1024: Ok, I will do that.

wenhu1024: Now, more tests have been added in basic.sv file.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -454,6 +454,21 @@ def NamedConstantOp : MooreOp<"named_constant", [
...
+  let description = [{
+    Produces a constant value of string type.
+
+    Example:
+    ```mlir
+    %0 = moore.string "hello world"
+    ```
+  }];
+  let arguments = (ins StrAttr:$value);
+  let results = (outs SimpleBitVectorType:$result);
```
uenoku, line 468: Can we restrict this to a StringType?

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -454,6 +454,21 @@ def NamedConstantOp : MooreOp<"named_constant", [
...
+  let assemblyFormat = "$value attr-dict `:` type($result)";
```
uenoku, line 469: You may not need type($result) once result is guaranteed to be StringType always.

include/circt/Dialect/Moore/MooreOps.td
wenhu1024: Maybe not. 

Here is the sv content
```
module test();
    string s="hello world";
endmodule
```


And slang ast json file generated below.
```
{
  "name": "$root",
  "kind": "Root",
  "addr": 100747055908128,
  "members": [
    {
      "name": "",
      "kind": "CompilationUnit",
      "addr": 100747056022832
    },
    {
      "name": "test",
      "kind": "Instance",
      "addr": 100747056023696,
      "body": {
        "name": "test",
        "kind": "InstanceBody",
        "addr": 100747056023104,
        "members": [
          {
            "name": "s",
            "kind": "Variable",
            "addr": 100747056023352,
            "type": "string",
            "initializer": {
              "kind": "Conversion",
              "type": "string",
              "operand": {
                "kind": "StringLiteral",
                "type": "bit[87:0]",
                "literal": "hello world",
                "constant": "88'h68656c6c6f20776f726c64"
              },
              "constant": "\"hello world\""
            },
            "lifetime": "Static"
          }
        ],
        "definition": "test"
      },
      "connections": [
      ]
    }
  ]
}
```
I reckon there is a Conversion between bits and string.

hailongSun2000: Hummm...It looks like slang turns the string into the bits array.

uenoku: Ok I see. In that case it looks reasonable to treat string literal as a packed bit.

fabianschuiki: Oh that's a nice find! So it looks like string literals are basically just a different way to create an integer constant?

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: Could we make this return an `IntType` instead? I'm pretty sure the string literal will always have a type `bit [N-1:0]`, which would be a `!moore.iN` integer type.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -454,6 +454,21 @@ def NamedConstantOp : MooreOp<"named_constant", [
   }];
 }
 
+def StringConstantOp : MooreOp<"string", [Pure, ConstantLike]> {
```
fabianschuiki, line 457: Minor nit: could we call the op something like `moore.constant_string` or `moore.string_constant` in the assembly? To line up with the C++ name `StringConstantOp` and to have the word "constant" in there :smiley:?

wenhu1024:
At present, I have revised the submission according to the given suggestions.Thanks a lot!

hailongSun2000 (approved):

fabianschuiki (approved):
Thanks a lot for adding this!

uenoku (approved):
LGTM!

Issue #7399, opened 2024-07-29
[ExportVerilog] Comb ExtractOp assertion "should be handled by isExpressionUnableToInline" failed with lowering option disallowLocalVariables


7FM:
To reproduce the failing assertion, use the following mlir code and run `circt-opt -hw-legalize-modules  -export-verilog <test.mlir>`:
```mlir
module @export_verilog_crash attributes {circt.loweringOptions = "disallowLocalVariables"} {

  hw.module @bitextract_crash(in %clk : i1, in %sel : i3, in %dummy : i32, out MyOutput : i17) {
    %res_reg = sv.reg sym @res_reg : !hw.inout<i17> 

    sv.alwaysff(posedge %clk) {
      %c-1209970334_i32 = hw.constant -1209970334 : i32
      %c-1028445891_i32 = hw.constant -1028445891 : i32

      %casez_tmp = sv.reg : !hw.inout<i32> 
      sv.case casez %sel : i3
      case b000: {
        sv.bpassign %casez_tmp, %c-1209970334_i32 : i32
      }
      default: {
        sv.bpassign %casez_tmp, %c-1028445891_i32 : i32
      }
      %17 = sv.read_inout %casez_tmp : !hw.inout<i32>
      %29 = comb.xor %dummy, %17 : i32
      %32 = comb.extract %29 from 3 : (i32) -> i17
      sv.passign %res_reg, %32 : i17
    }

    %res_reg_data = sv.read_inout %res_reg : !hw.inout<i17>
    hw.output %res_reg_data : i17
  }
}
```
Note that the `-hw-legalize-modules` is optional to reproduce the crash.
`circt.loweringOptions = "disallowLocalVariables"` is, however, required.

Here is the full stacktrace:
```
circt-opt: circt/lib/Conversion/ExportVerilog/ExportVerilog.cpp:2754: SubExprInfo (anonymous namespace)::ExprEmitter::visitComb(ExtractOp): Assertion `(x.precedence == Symbol || (x.precedence == Selection && isOkToBitSelectFrom(op.getInput()))) && "should be handled by isExpressionUnableToInline"' failed.
PLEASE submit a bug report to https://github.com/llvm/circt and include the crash backtrace.
 #0 0x000055589ad931bc llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) circt/llvm/llvm/lib/Support/Unix/Signals.inc:723:13
[...]

PR #7404, opened 2024-07-30
[PrepareForEmission] Hoist registers in a procedural region with `disallowLocalVariables`

uenoku:
Fix https://github.com/llvm/circt/issues/7399


uenoku:
The fixes the crash reported so I'm going to merge it.

uenoku:
Thank you for pointing this out! I was slightly surprised that reg op initial operand is not restricted to constant. I think for RegOp with initial operand under disallowLocalVariables, PrepareForEmission should drop initialization operand in reg op declaration but create sv.bpassign in the original position. WDYT?

PR #7417, opened 2024-07-31
[OM] Pass Python values back and forth, not Attributes.

mikeurbach:
Internally, primitive OM EvaluatorValues are represented as TypedAttributes. This is great internally, but when we pass these from C++ out to Python, we have to use a very inefficient method to pull the Python value out of the attribute.

This updates how primitives are handled at the Python <> C++ interface to directly construct the appropriate Python values and return them. Similarly, for top-level inputs to the Evaluator, Python values are directly accepted and converted to Attributes internally.

On large designs, this was shown to decrease single threaded CPU time to process large amounts of OM data by roughly 70%. There is no difference in the output.

### DIFF
--- a/lib/Bindings/Python/OMModule.cpp
+++ b/lib/Bindings/Python/OMModule.cpp
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "DialectModules.h"
+#include "circt-c/Dialect/HW.h"
 #include "circt-c/Dialect/OM.h"
 #include "mlir-c/BuiltinAttributes.h"
 #include "mlir-c/BuiltinTypes.h"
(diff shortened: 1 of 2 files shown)


lib/Bindings/Python/OMModule.cpp
```diff
@@ -349,6 +363,88 @@ Map::dunderGetItem(std::variant<intptr_t, std::string, MlirAttribute> key) {
   return dunderGetItemAttr(std::get<MlirAttribute>(key));
 }
 
+// Convert a generic MLIR Attribute to a PythonValue. This is basically a C++
+// fast path of the parts of attribute_to_var that we use in the OM dialect.
+static PythonPrimitive omPrimitiveToPythonValue(MlirAttribute attr) {
+  if (omAttrIsAIntegerAttr(attr)) {
+    auto strRef = omIntegerAttrToString(attr);
+    return py::int_(py::str(strRef.data, strRef.length));
```
prithayan, line 371: is there a reason, we need to go from int to str to `py::int_` ?
(Not super familiar with the pybind API)

prithayan (approved):
Minor comment, but overall looks good.

uenoku (approved):
LGTM!

leonardt (approved):

mikeurbach:
For the C++ -> Python direction, this should handle all the types of primitive attributes we are using. For the Python -> C++ direction, this only supports some of the basic int, float, str, bool. Some of the more complex primitives could be supported, but nothing was actually using that, so I'm leaving it to future work.

PR #7419, opened 2024-07-31
[docs] Remove confusing reset in Seq docs SV example

TaoBi22:
These docs give an example of a SV lowering of a firreg without a reset, but then the SV snippet has a reset in it, which is a bit confusing.


uenoku (approved):
LGTM!

PR #7420, opened 2024-07-31
[FIRRTL][Dedup] Rework hashing for perf and bug fixes.

### LINKED ISSUE #7415: [FIRRTL][Dedup] Incorrect deduplication of same operations in different blocks
Consider:
```firrtl
firrtl.circuit "Whens" {
  firrtl.module private @A(in %x: !firrtl.uint<1>) {
    firrtl.when %x : !firrtl.uint<1> {
      %0 = firrtl.constant 5 : !firrtl.uint<32>
      %n = firrtl.node %0 : !firrtl.uint<32>
    }
  }
  firrtl.module private @B(in %x: !firrtl.uint<1>) {
    firrtl.when %x : !firrtl.uint<1> {
    } else {
      %0 = firrtl.constant 5 : !firrtl.uint<32>
      %n = firrtl.node %0 : !firrtl.uint<32>
    }
  }
  firrtl.module @Whens(in %x: !firrtl.uint<1>) {
    %a_x = firrtl.instance a @A(in x: !firrtl.uint<1>)
    firrtl.matchingconnect %a_x, %x : !firrtl.uint<1>
    %b_x = firrtl.instance b @B(in x: !firrtl.uint<1>)
    firrtl.matchingconnect %b_x, %x : !firrtl.uint<1>
  }
}
```

Note that the first `firrtl.when` has two operations in the "true" block, the second has the same two in the "false" lblock.

`circt-opt --firrtl-dedup` on the above yields:

```firrtl
firrtl.circuit "Whens" {
  firrtl.module private @A(in %x: !firrtl.uint<1>) {
    firrtl.when %x : !firrtl.uint<1> {
      %c5_ui32 = firrtl.constant 5 : !firrtl.uint<32>
      %n = firrtl.node %c5_ui32 : !firrtl.uint<32>
    }
  }
  firrtl.module @Whens(in %x: !firrtl.uint<1>) {
    %a_x = firrtl.instance a @A(in x: !firrtl.uint<1>)
    firrtl.matchingconnect %a_x, %x : !firrtl.uint<1>
    %b_x = firrtl.instance b @A(in x: !firrtl.uint<1>)
    firrtl.matchingconnect %b_x, %x : !firrtl.uint<1>
  }
}
```

### LINKED ISSUE #7416: [FIRRTL][Dedup] Empty blocks are treated same as no block
Related to #7415, consider:

```
// Check dedup of an empty block vs no block.
// CHECK-LABEL: EmptyBlockVSNoBlock
firrtl.circuit "EmptyBlockVSNoBlock" {
  // CHECK: firrtl.module private @A(
  firrtl.module private @A(in %x: !firrtl.uint<1>) {
    firrtl.when %x : !firrtl.uint<1> {
    }
  }
  // CHECK-NOT: @B
  firrtl.module private @B(in %x: !firrtl.uint<1>) {
    firrtl.when %x : !firrtl.uint<1> {
    } else {
    }
  }
  firrtl.module @EmptyBlockVSNoBlock(in %x: !firrtl.uint<1>) {
    // CHECK: firrtl.instance a @A
    %a_x = firrtl.instance a @A(in x: !firrtl.uint<1>)
    firrtl.matchingconnect %a_x, %x : !firrtl.uint<1>
    // CHECK: firrtl.instance b @A
    %b_x = firrtl.instance b @B(in x: !firrtl.uint<1>)
    firrtl.matchingconnect %b_x, %x : !firrtl.uint<1>
  }
}
```

Presently dedup considers an empty block for "else" the same as no block, and these two dedup when they shouldn't.

This is fairly benign in the above example, but the difference may matter for other or foreign operations.

dtzSiFive:
Primary change is to only generate and populate mappings for sources of values, and not each value themselves.

Values are identified using these base numberings plus an appropriate offset.

The main benefit of this is to greatly reduce the number of entries in the `indices` map.
When handling operations with many block arguments (module-like's with many ports) or with many results (instances of those module-like's) this greatly reduces the pressure on the `indices` map.  For these designs, dedup now runs dramatically faster and uses significantly less memory.

Also separates location of the value impl, such that if a Value's impl is storage inline into an Operation or Block such that there is aliasing, the two are given different numbers (and especially the numbering isn't changed).

On a synthetic design containing a module with 2^20 ports and 256 instances of that module, this is the difference between completing in 20s and OOM'ing on my machine after running for 30 minutes.

Functional changes:

Fixes #7415.
Fixes #7416.
Also fixes deduping if block arg types are different (but unused).

This is done by hashing block count, and each block's numbering between as well as the types of its arguments before that block's operations.

Additionally fixes use of numberings (indices) before it was populated where attribute processing for inner symbol ports hashed using the block argument's numbering before it was populated.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -78,8 +78,17 @@ struct ModuleInfo {
   mlir::ArrayAttr referredModuleNames;
 };
 
-struct SymbolTarget {
+/// Unique identifier for a value.  All value sources are numbered by apperance,
+/// and values are identified using this numbering (`index`) and an `offset`.
+/// For BlockArgument's, this is the argument number.
+/// For OpResult's, this is the result number.
+struct ValueId {
   uint64_t index;
+  uint64_t offset;
+};
+
+struct SymbolTarget {
+  ValueId index;
   uint64_t fieldID;
 };
 
@@ -161,14 +170,19 @@ private:
 
   void record(void *address) {
     auto size = indices.size();
+    assert(!indices.contains(address));
     indices[address] = size;
   }
 
-  void update(BlockArgument arg) { record(arg.getAsOpaquePointer()); }
+  /// Get the unique id for the specified value.
+  ValueId getId(Value val) {
+    if (auto arg = dyn_cast<BlockArgument>(val))
+      return {indices.at(arg.getOwner()), arg.getArgNumber()};
+    auto result = cast<OpResult>(val);
+    return {indices.at(result.getOwner()), result.getResultNumber()};
+  }
 
   void update(OpResult result) {
-    record(result.getAsOpaquePointer());
-
     // Like instance ops, don't use object ops' result types since they might be
     // replaced by dedup. Record the class names and lazily combine their hashes
     // using the same mechanism as instances and modules.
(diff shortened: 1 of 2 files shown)


mikeurbach (approved):
This is great. You've clearly put a lot more thought into this than I, but after reviewing it no major concerns from me.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -78,8 +78,17 @@ struct ModuleInfo {
   mlir::ArrayAttr referredModuleNames;
 };
 
-struct SymbolTarget {
+/// Unique identifier for a value.  All value sources are numbered by apperance,
+/// and values are identified using this numbering (`index`) and an `offset`.
+/// For BlockArguments, this is the block number.
+/// For OpResults, this is the result number.
```
dtzSiFive, line 84:
```suggestion
/// For BlockArgument's, this is the argument number.
/// For OpResult's, this is the result number.
```

uenoku (approved):
Awesome!

dtzSiFive:
Landing, thanks for reviews folks!

PR #5793, opened 2023-08-07
[NFC][HW] Add container class for n-valued logic

fzi-hielscher:
This PR adds RLELogic,  a run-length encoding container class for arbitrary sequences of  n-valued logic digits within the set of `0,1,X,Z,U,W,L,H,-`. It is a prerequisite for integrating the `hw.logic` type of #5622. As it is not currently used anywhere it should not cause any functional changes, yet.

The motivation behind RLELogic is to provide a somewhat efficient representation for non-integer constants in hardware designs, which generally will be of `logic` or `std_logic` type for SystemVerilog and VHDL respectively. These can usually be assumed to be short and/or contain long runs of the same digit repeatedly. Thus, RLELogic uses run-length encoding to reduce their size in memory. 
Each byte of the encoded buffer represents one of the nine possible digits in its lower nibble, while encoding a run-length from one to sixteen in the upper nibble. The run-length of the most significant encoded digit is implicitly assumed to be infinite. Thus, all values which are just one specific digit repeated over and over share the same encoding independent of their 'actual' length. Since an RLELogic value will generally be associated with a known width instance of the `hw.logic` type class, storing the width in RLELogic would be redundant. However, it contains a bit mask indicating all the digits contained in its value at least once. This allows to quickly check, e.g., whether the encoded value contains an 'unknown' digit or represents an integer.

In general RLELogic should _not_ be used to store integers. In fact, it should be avoided to encode an integer value as RLELogic whenever possible and APInt should be used instead. This is a major departure from my previous attempts of creating an unified APLogic container for both integer and non-integer values. While APLogic  _technically_ worked, its integration did not pan out as I had hoped, and it would not have been particularly efficient. Also, its implementation became _way_ too complex to be worthwhile. RLELogic should be a much cleaner solution, and likely more efficient in real-world scenarios.

There is currently only a limited number of operations implemented on RLELogic. I intend to add more when the general type infrastructure for `hw.logic` is in place.

PR #7422, opened 2024-08-02
[Support] Add FVInt, a four-valued arbitrary precision integer

fabianschuiki:
Add the `FVInt` class to CIRCT's support library. This class can represent arbitrary precision integers where each bit can be one of the four values 0, 1, X, and Z. The name intends to suggest a *four-valued APInt*, with the option to also introduce a *nine-valued NVInt* in the future.

Internally, `FVInt` uses two `APInt`s to store its data: `value` stores whether a bit is 0/X or 1/Z, and `unknown` stores whether a bit is known (0 or 1) or unknown (X or Z). Together they allocate 2 bits of storage for each of the `FVInt`'s digits, which allows for four different values per digit. This representation as `value` and `unknown` makes many of the logical and arithmetic operations pretty straightforward to implement. Most four-valued operations can be trivially implemented by performing the equivalent two-valued operation on `value`, and then accounting for X and Z bits through a few logic operations on `unknown`.

Note that Slang defines its own version of this (`SVInt`). But since Slang is an optional dependency of CIRCT, it makes sense to have a CIRCT equivalent that is built around LLVM's `APInt`, for use in our dialects.

This first version of `FVInt` has a rather incomplete set of operations, but it covers basic AND, OR, XOR, NOT, negation, addition, subtraction, and multiplication as a proof-of-concept. The remaining operations will be added in future commits. We are also going to need a four-valued equivalent of `IntegerAttr` based on `FVInt`.

This commit is motivated by the Slang frontend, which now supports enough of SystemVerilog to make some test designs start to hit the lack of number literals with X and Z.

### DIFF
--- a/include/circt/Support/FVInt.h
+++ b/include/circt/Support/FVInt.h
@@ -1,0 +1,592 @@
+//===- FVInt.h - Four-valued integer ----------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements a class to represent arbitrary precision integers where
+// each bit can be one of four values. This corresponds to SystemVerilog's
+// four-valued `logic` type (originally defined in IEEE 1364, later merged into
+// IEEE 1800).
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_SUPPORT_FVINT_H
+#define CIRCT_SUPPORT_FVINT_H
+
+#include "circt/Support/LLVM.h"
+#include "llvm/ADT/APInt.h"
+#include "llvm/ADT/SmallString.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace circt {
+
+/// Four-valued arbitrary precision integers.
+///
+/// Each bit of the integer can be 0, 1, X, or Z. Internally the bits are stored
+/// in a pair of `APInt`s, one of which specifies the value of each bit (0/X or
+/// 1/Z), and the other whether the bit is unknown (X or Z).
+class FVInt {
+public:
+  /// Construct an `FVInt` from a 64-bit value. The result has no X or Z bits.
+  FVInt(unsigned numBits, uint64_t value, bool isSigned = false)
+      : FVInt(APInt(numBits, value, isSigned)) {}
+
+  /// Construct an `FVInt` from an `APInt`. The result has no X or Z bits.
+  FVInt(const APInt &value)
+      : value(value), unknown(APInt::getZero(value.getBitWidth())) {}
+
+  /// Construct an `FVInt` from two `APInt`s used internally to store the bit
+  /// data. The first argument specifies whether each bit is 0/X or 1/Z. The
+  /// second argument specifies whether each bit is 0/1 or X/Z. Both `APInt`s
+  /// must have the same bitwidth. The two arguments correspond to the results
+  /// of `getRawValue()` and `getRawUnknown()`.
+  FVInt(APInt &&rawValue, APInt &&rawUnknown)
+      : value(rawValue), unknown(rawUnknown) {
+    assert(rawValue.getBitWidth() == rawUnknown.getBitWidth());
+  }
+
+  /// Construct an `FVInt` with all bits set to 0.
+  static FVInt getZero(unsigned numBits) {
+    return FVInt(APInt::getZero(numBits));
+  }
+
+  /// Construct an `FVInt` with all bits set to 1.
+  static FVInt getAllOnes(unsigned numBits) {
+    return FVInt(APInt::getAllOnes(numBits));
+  }
+
+  /// Construct an `FVInt` with all bits set to X.
+  static FVInt getAllX(unsigned numBits) {
+    return FVInt(APInt::getZero(numBits), APInt::getAllOnes(numBits));
+  }
+
+  /// Construct an `FVInt` with all bits set to Z.
+  static FVInt getAllZ(unsigned numBits) {
+    return FVInt(APInt::getAllOnes(numBits), APInt::getAllOnes(numBits));
+  }
+
+  /// Return the number of bits this integer has.
+  unsigned getBitWidth() const { return value.getBitWidth(); }
+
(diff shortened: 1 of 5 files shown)


include/circt/Support/FVInt.h
```diff
@@ -0,0 +1,591 @@
...
+    assert(success && "radix cannot represent FVInt");
+    return str;
+  }
+
+  /// Print an `FVInt` to an output stream.
+  void print(raw_ostream &os) const;
+
+private:
+  APInt value;
+  APInt unknown;
```
TaoBi22, line 560: This is super exciting to see!! One nit: the Verilog standard document uses 'unknown' as the name specifically for an `x` bit:

> 0—represents a logic zero or a false condition
> 1—represents a logic one or a true condition
> x—represents an unknown logic value
> z—represents a high-impedance state

(from the [IEEE standard document](https://standards.ieee.org/ieee/1800/7743/))


 Is there a chance it could cause confusion that here 'unknown' refers to both `x` and `z` bits?

fzi-hielscher (approved):
Very nice! 🎉 
Just some nitpicks re zero width values.

When tinkering with nine-valued containers last year I spent a lot (arguably too much) time trying to compress the in-memory representation to avoid heap traffic. If, and only if, FVInt is intended as a general container for Verilog `logic` constants my intuition would still lead me to try and avoid allocating the `unknown` APInt for values that do not contain any X/Z. IIRC this is how slang's `SVInt` handles it. Of course it is only relevant for heap allocated APInts, i.e., values wider than 64 bit, so it may not actually matter in most cases.

I still have not entirely given up on my RLELogic class (#5793). I think it has some nice properties for dealing with parameterized-width IEEE 1164 values specifically. But for the slang integration your approach is a much better fit.

include/circt/Support/FVInt.h
fabianschuiki: That's a great point! I took the naming scheme from Slang's `SVInt`. This description of `x` as being _the_ unknown logic value is a bit unfortunate, because the SV standard in other places talks about functions like `$isunknown`, which checks for X _and_ Z. I guess that's also where Slang got its name from. My gut feeling would be to stick with Slang's naming scheme here.

include/circt/Support/FVInt.h
```diff
@@ -0,0 +1,591 @@
...
+
+  /// Zero-extend the integer to a new bit width. The additional high-order bits
+  /// are filled in with zero.
+  FVInt zext(unsigned bitWidth) const {
+    return FVInt(value.zext(bitWidth), unknown.zext(bitWidth));
+  }
+
+  /// Sign-extend the integer to a new bit width. The additional high-order bits
+  /// are filled in with the sign bit (top-most bit) of the original number. If
+  /// the sign bit is X or Z, fills in X or Z, respectively.
```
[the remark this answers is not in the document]
fabianschuiki: Good point. Added.

fabianschuiki:
@fzi-hielscher Thanks a lot for the detailed review! :partying_face: 

> [...] my intuition would still lead me to try and avoid allocating the `unknown` APInt for values that do not contain any X/Z. IIRC this is how slang's `SVInt` handles it.

That is definitely a good point. My gut feeling here was to try and avoid early optimization, and instead try to get something up and running that's correct but possibly suboptimal. I think it's going to be easier to come back a while later and add these optimizations as a separate PR. And I think you're right, in practice this might be unmeasurable in almost all cases, since most integers will be very small.

fzi-hielscher:
Rumor has it _certain_ HDL generators produce ROMs as ginormous muxes of literals. At that point I feel literal performance could become a concern. On a completely unrelated note, one of my far too many experimental branches adds support for `FirMemInitAttrs` to arcilator. Maybe I should spend more time giving that some spit and polish rather than trying to prematurely micro-"optimize" heap allocations. 😬

fabianschuiki:
:grimacing: That sounds like a pretty adventurous HDL generator :stuck_out_tongue:.

Your arcilator work sounds fantastic! Please hit me with all the PRs :grin:

include/circt/Support/FVInt.h
```diff
@@ -0,0 +1,592 @@
...
+    auto v = *this;
+    v.negate();
+    return v;
+  }
+
+  /// Compute the addition of this integer and another. If any bits in either
+  /// integer are unknown, the entire result is X.
+  FVInt &operator+=(const FVInt &other) {
+    value += other.value;
+    setAllXIfAnyUnknown(other);
```
uenoku, line 412: nit: It might be a bit too conservative to mark everything unknown if lower bits are known, e.g. `x10 + x10 = x00`. Does IEEE 1800 define 4-value logic arithmetic as well as bitwise logical operation?

uenoku (approved):
Awesome, this really looks great! Having data structure for 4 value logic does in CIRCT does make a lot sense :)

include/circt/Support/FVInt.h
fabianschuiki: Good point. I've found the following nugget in IEEE 1800-2017 "11.4.3 Arithmetic operators" hidden between tables:

> For the arithmetic operators, if any operand bit value is the unknown value x or the high-impedance value z, then the entire result value shall be x.

So I guess conservative is fine/desired?

fzi-hielscher: The "fun" really starts when you get to the unary arithmetic operators. What's the expected output here?
```
module test;
  logic [3:0] foo = 4'b01XZ;
  initial begin
    $display("None: %b", foo);
    $display("Plus: %b", +foo);
    $display("Minus: %b", -foo);
  end
endmodule
```
I've seen simulators disagree on the handling of the unary plus.

fabianschuiki: Nasty! I guess from that line in the standard you'd expect to see `foo` unchanged, and both `+foo` and `-foo` to be all X since they count as arithmetic operators? Unless there's a line in the standard that says unary `+` can be treated as a no-op… But if that doesn't exist, we'd need a `moore.x_if_any_unknown` style op 😅

fzi-hielscher: Check out Table 11-5 right above the sentence you quoted.
cantbeliveididthis
I'll see myself out....

But most simulators seem to opt for "same as m".

fabianschuiki: What a beautiful spec... But yeah, `+m = m` feels like the right thing to do intuitively, but it also seems to violate the spec. We can always drop an op that handles the X-ification for `+`.

uenoku: > So I guess conservative is fine/desired?

Oh, yeah. Thank you for pointing it out!

PR #7425, opened 2024-08-02
[Ibis] Don't include design name in namespace in IbisContainersToHW

mortbopet:
The ibis.design op will be removed after the IbisContainersToHW pass, and there may be ibis.component's inside the design that have the same name as the design; we want that name to persist, and not be falsely considered a duplicate.


blakep-msft (approved):
Thanks for doing this.

PR #7426, opened 2024-08-02
[FIRRTL][Dedup] Improve diagnostic for failing to dedup due to public.

dtzSiFive:


lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -46,6 +46,20 @@ using namespace circt;
 using namespace firrtl;
 using hw::InnerRefAttr;
 
+//===----------------------------------------------------------------------===//
+// Utility function for classifying a Symbol's dedup-ability.
+//===----------------------------------------------------------------------===//
+
+bool checkVisibility(mlir::SymbolOpInterface symbol) {
```
rwy7, line 53: nit: probably should be static

dtzSiFive: Yes, thanks!

PR #7428, opened 2024-08-02
[ImportVerilog] Add rvalue assignment pattern support

fabianschuiki:
Add support for assignment patterns like `'{42, 9001}` in rvalue position. These are roughly SystemVerilog's equivalent of `struct_create` and `array_create`. This commit also adds an `array_create` op to support assignment patterns for arrays.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -727,6 +727,70 @@ struct RvalueExprVisitor {
     return builder.create<moore::StringConstantOp>(loc, type, expr.getValue());
   }
 
+  /// Handle assignment patterns.
+  Value visitAssignmentPattern(
+      const slang::ast::AssignmentPatternExpressionBase &expr,
+      unsigned replCount = 1) {
+    auto type = context.convertType(*expr.type);
+
+    // Convert the individual elements first.
+    auto elementCount = expr.elements().size();
+    SmallVector<Value> elements;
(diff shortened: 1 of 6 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -727,6 +727,70 @@ struct RvalueExprVisitor {
...
+    elements.reserve(replCount * elementCount);
+    for (auto elementExpr : expr.elements()) {
+      auto value = context.convertRvalueExpression(*elementExpr);
+      if (!value)
+        return {};
+      elements.push_back(value);
+    }
+    for (unsigned replIdx = 1; replIdx < replCount; ++replIdx)
+      for (unsigned elementIdx = 0; elementIdx < elementCount; ++elementIdx)
+        elements.push_back(elements[elementIdx]);
```
hailongSun2000, line 748: nit: With the existing tests, the `elementCount` is always 1. I noticed that slang supports `arr = '{3{'{3{'0, '1}}}};`. I'm not sure whether `{'0, '1}` is considered two element counts. Maybe we can add this as a case.

hailongSun2000 (approved):
LGTM!

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Excellent point! I have adjusted the tests to cover the case you're describing :smiley:

PR #7429, opened 2024-08-03
[Arc] Fix crash in the arc canonicalizer produced by the KeepOneVecOp optimization

elhewaty:


lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
```diff
@@ -709,15 +708,14 @@ LogicalResult KeepOneVecOp::matchAndRewrite(VectorizeOp vecOp,
...
-      argsToRemove.set(argIdx);
       rewriter.replaceAllUsesWith(currentBlock.getArgument(argIdx - shuffledBy),
                                   currentBlock.getArgument(in->second));
       currentBlock.eraseArgument(argIdx - shuffledBy);
       ++shuffledBy;
       changed = true;
       continue;
     }
-    inExists[input] = argIdx;
+    inExists[input] = argIdx - shuffledBy;
```
maerhart, line 718: Nit: instead of incrementing `argIdx` every iteration and subtracting the number of args we have already deleted, we could only increment `argIdx` when we don't delete the argument.

test/Dialect/Arc/arc-canonicalizer.mlir
```diff
@@ -491,6 +491,48 @@ hw.module @Repeated_input(in %b: i8, in %e: i8, in %h: i8, in %k: i8, in %en: i1
...
+  ^bb0(%arg0: !seq.clock):
+    %4 = arc.state @TLXbar_8_arc() clock %arg0 latency 1 : () -> i1
+    arc.vectorize.return %4 : i1
+  }
+  %2:2 = arc.vectorize (%1#0, %1#1), (%1#0, %1#1), (%0, %0), (%0, %0) : (i1, i1, i1, i1, i1, i1, i1, i1) -> (i124, i124) {
+  ^bb0(%arg0: i1, %arg1: i1, %arg2: i1, %arg3: i1):
+    %4 = arc.call @Queue_24_arc(%arg1, %arg1, %arg3, %arg3) : (i1, i1, i1, i1) -> i124
+    arc.vectorize.return %4 : i124
+  }
+  %3 = arc.state @dummy(%2#0, %3) clock %clock latency 1 : (i124, i124) -> i124
```
maerhart, line 521: Nit: any particular reason this `state` (and the arc it's referring to) has to be here or could we just add `%2` as module output?

test/Dialect/Arc/arc-canonicalizer.mlir
```diff
@@ -491,6 +491,48 @@ hw.module @Repeated_input(in %b: i8, in %e: i8, in %h: i8, in %k: i8, in %en: i1
...
+  %0 = comb.or %arg0, %arg1 : i124
+  arc.output %0 : i124
+}
+
+hw.module @Repeated_again(in %clock : !seq.clock, in %0 "" : i1) {
+  %1:2 = arc.vectorize (%clock, %clock) : (!seq.clock, !seq.clock) -> (i1, i1) {
+  ^bb0(%arg0: !seq.clock):
+    %4 = arc.state @TLXbar_8_arc() clock %arg0 latency 1 : () -> i1
+    arc.vectorize.return %4 : i1
+  }
```
maerhart, line 515: Is this `arc.vectorize` necessary? Isn't just the other one triggering this issue?

maerhart (commented):
Good job finding this! Just some question about whether we couldn't simplify the regression test a little more.

lib/Dialect/Arc/Transforms/ArcCanonicalizer.cpp
elhewaty: Yes, that's better. can we handle this in `auto [argIdx, inputVec] : llvm::enumerate(vecOp.getInputs())`, should I refactor the loop?

test/Dialect/Arc/arc-canonicalizer.mlir
elhewaty: No particular reason for this. I just wanted to use `%2`, I don't know how to use it as a module output, if we don't use it, it will be eliminated by the optimizer.

test/Dialect/Arc/arc-canonicalizer.mlir
elhewaty: It isn't, I just needed multiple inputs to handle the shuffles. I will handle it with another input.

elhewaty: how about this? If there's a way to return the result from the module, that would be great as we will we git rid of the arc.

```
arc.define @dummy(%arg0: i1, %arg1: i1) -> i1 {
  %0 = comb.or %arg0, %arg1 : i1
  arc.output %0 : i1
}

hw.module @Repeat_agian(in %clock : !seq.clock, in %0 : i1, in %in1: i1, in %in2: i1) {
  %2:2 = arc.vectorize (%in1, %in2), (%in1, %in2), (%0, %0), (%0, %0) : (i1, i1, i1, i1, i1, i1, i1, i1) -> (i1, i1) {
  ^bb0(%arg0: i1, %arg1: i1, %arg2: i1, %arg3: i1):
    %4 = comb.or %arg0, %arg1, %arg2, %arg3 : i1
    arc.vectorize.return %4 : i1
  }
  %3 = arc.state @dummy(%2#0, %3) clock %clock latency 1 : (i1, i1) -> i1
  hw.output
}
```

fabianschuiki (approved):
Very nice find and fix! :partying_face: LGTM

maerhart (approved):
Thanks!

PR #7430, opened 2024-08-03
[OM] Add IntegerAttr -> Python int conversion

seldridge:
Add a missing conversion for OM to Python conversion where only OM integers and not arbitrary MLIR integers would be converted to Python. This could result in failures in Python scripts that needed to consume OM.

This fixes a bug introduced in: 17c036f87c4c9e103160b207e18ad595911945e8

### DIFF
--- a/lib/Bindings/Python/OMModule.cpp
+++ b/lib/Bindings/Python/OMModule.cpp
@@ -366,6 +366,15 @@ Map::dunderGetItem(std::variant<intptr_t, std::string, MlirAttribute> key) {
 // Convert a generic MLIR Attribute to a PythonValue. This is basically a C++
 // fast path of the parts of attribute_to_var that we use in the OM dialect.
 static PythonPrimitive omPrimitiveToPythonValue(MlirAttribute attr) {
+  if (mlirAttributeIsAInteger(attr)) {
+    MlirType type = mlirAttributeGetType(attr);
+    if (mlirTypeIsAIndex(type) || mlirIntegerTypeIsSignless(type))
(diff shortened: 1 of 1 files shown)


seldridge:
I'm testing this to see if it fixes the failure I observed...

lib/Bindings/Python/OMModule.cpp
```diff
@@ -366,6 +366,15 @@ Map::dunderGetItem(std::variant<intptr_t, std::string, MlirAttribute> key) {
...
 // fast path of the parts of attribute_to_var that we use in the OM dialect.
 static PythonPrimitive omPrimitiveToPythonValue(MlirAttribute attr) {
+  if (mlirAttributeIsAInteger(attr)) {
+    MlirType type = mlirAttributeGetType(attr);
+    if (mlirTypeIsAIndex(type) || mlirIntegerTypeIsSignless(type))
+      return py::int_(mlirIntegerAttrGetValueInt(attr));
+    if (mlirIntegerTypeIsSigned(type))
+      return py::int_(mlirIntegerAttrGetValueSInt(attr));
+    return py::int_(mlirIntegerAttrGetValueUInt(attr));
+  }
```
seldridge, line 376: This is copied from: https://github.com/llvm/llvm-project/blob/1a9acd786d493b00c08d1611f51420d421b74cf1/mlir/lib/Bindings/Python/IRAttributes.cpp

mikeurbach (approved):
Thanks, looks good to me. Sorry I missed this originally.

PR #7432, opened 2024-08-04
[ImportVerilog] Switch from SCF to CF dialect for control flow

fabianschuiki:
Instead of using the SCF dialect and its `scf.if` and `scf.while` operations for control flow, switch over to the CF dialect and its `cf.br` and `cf.cond_br`. This allows us to support SystemVerilog's `continue`, `break`, and `return` statements which don't map well to the Structured Control Flow dialect.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -131,8 +131,6 @@ def ProcedureKindAttr: I32EnumAttr<"ProcedureKind", "Procedure kind",
 }
 
 def ProcedureOp : MooreOp<"procedure", [
-  SingleBlock,
-  NoTerminator,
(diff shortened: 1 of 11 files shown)


hailongSun2000 (approved):
Very fine-grain code! I reviewed the code line-by-line against the tests. In this way, I'm just able to make clear the whole flow. The functions `isTerminated()` and `setTerminated()` must be paid attention. They're a key to guaranteeing the code fine-grain :rocket:.

fabianschuiki:
Thanks for the review @hailongSun2000. I'm going to land this such that we can look into reworking other parts of the pipeline to be compatible with CF and multiple blocks. For example, I'm pretty sure this will break the `SimplifyProcedures` pass, which doesn't expect to see multiple blocks inside a procedure.

PR #7433, opened 2024-08-04
[FIRRTL] Add CheckRecursiveInstantiation diagnostic pass

youngar:
This adds a pass to FIRRTL to check for recursive module instantiation,
which is illegal because it corresponds to infinitely sized hardware.
This uses the path based strongly connected component (SCC) algorithm
with some extra book keeping to track if there was a loop in the current
SCC or not.  This is especially helpful to detect if a single-element
SCC actually contains a loop or not.

This does not use the upstream SCCIterator because we hope that this
version of the algorithm is faster by a constant factor, we don't need
to visit the SCCs in post-order, and hopefully faster handling of cycles
in single-element SCCs (which requires re-scanning all edges from a node
using the upstream version).

### DIFF
--- a/include/circt/Dialect/FIRRTL/Passes.td
+++ b/include/circt/Dialect/FIRRTL/Passes.td
@@ -970,4 +970,18 @@ def ProbesToSignals : Pass<"firrtl-probes-to-signals", "firrtl::CircuitOp"> {
   let constructor = "circt::firrtl::createProbesToSignalsPass()";
 }
 
+def CheckRecursiveInstantiation : Pass<"firrtl-check-recursive-instantiation",
+    "firrtl::CircuitOp"> {
+  let summary = "Check for illegal recursive instantiation";
+  let description = [{
+    This pass checks for illegal recursive module instantion.  Recursive
+    instantiation is when a module instantiates itself, either directly or
+    indirectly through other modules it instantiates.  Recursive module 
+    instantiation is illegal because it would require infinite hardware to
+    synthesize. Recursive class instantiation is illegal as it would create an
+    infinite loop.
+  }];
+  let constructor = "circt::firrtl::createCheckRecursiveInstantiation()";
+}
+
 #endif // CIRCT_DIALECT_FIRRTL_PASSES_TD
--- a/lib/Dialect/FIRRTL/Transforms/CheckRecursiveInstantiation.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CheckRecursiveInstantiation.cpp
@@ -1,0 +1,66 @@
+//===- CheckRecursiveInstantiation.cpp - Check recurisve instantiation ----===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/FIRRTL/FIRRTLInstanceGraph.h"
+#include "circt/Dialect/FIRRTL/Passes.h"
+#include "mlir/Pass/Pass.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SCCIterator.h"
+
+namespace circt {
+namespace firrtl {
+#define GEN_PASS_DEF_CHECKRECURSIVEINSTANTIATION
+#include "circt/Dialect/FIRRTL/Passes.h.inc"
+} // namespace firrtl
+} // namespace circt
+
+using namespace circt;
+using namespace firrtl;
+
+static void printPath(InstanceGraph &instanceGraph,
+                      ArrayRef<InstanceGraphNode *> nodes) {
+  assert(nodes.size() > 0 && "an scc should have at least one node");
+  auto diag =
+      emitError(nodes.front()->getModule().getLoc(), "recursive instantiation");
+  llvm::SmallPtrSet<InstanceGraphNode *, 8> scc(nodes.begin(), nodes.end());
+  for (auto *node : nodes) {
+    for (auto *record : *node) {
+      auto *target = record->getTarget();
+      if (!scc.contains(target))
+        continue;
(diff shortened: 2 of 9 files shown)


youngar:
This does not support InstanceChoice, as the InstanceGraph itself does not support it yet.

test/Dialect/FIRRTL/check-recursive-modules-errors.mlir
```diff
@@ -0,0 +1,85 @@
...
+firrtl.circuit "MutualLoop1" {
+  firrtl.module public @MutualLoop1() {
+    firrtl.instance b @B()
+  }
+  firrtl.module @A() {
+    // expected-note @below {{A instantiates B here}}
+    firrtl.instance b @B()
+  }
+  // expected-error @below {{recursive instantiation}}
+  firrtl.module @B() {
```
dtzSiFive, line 68: Modules are public unless explicitly marked private, so I'm not sure what tests are intended but if these are meant to test re:private they should be updated, and if not a test or two for those cases seems warranted.

youngar: Updated the test case as mentioned, and discovered that the depth calculation was not working 😭  Hooray for tests.  Anyways, pushed a fix for that, and also renamed my test files to match the name of the test (went through a couple different names).

youngar:
I keep thinking about it, I think my depth heuristic can return the wrong results if there is an SCC instantiating another SCC.  I  might rework this later when I have time.

seldridge:
> This does not use the upstream SCCIterator because we hope that this
version of the algorithm is faster by a constant factor, we don't need
to visit the SCCs in post-order, and hopefully faster handling of cycles
in single-element SCCs (which requires re-scanning all edges from a node
using the upstream version).

This is worth weighing against the overhead of having another SCC implementation. E.g., if the single-element SCC is a concern, then that can be handled with a non-SCC-based pass over the modules to check for trivial self-instantiation. This could then fallback to the upstream algorithm for deeper analysis which can then assume no single-element SCCs.

More of a thought than anything.

I weighed similar concerns with using a new SCC for layer sink and settled on the opposite finding: it was better to struggle with upstream graph traits for reasons of maintainability.

lib/Dialect/FIRRTL/Transforms/CheckRecursiveInstantiation.cpp
```diff
@@ -0,0 +1,91 @@
...
+  auto *current = first;
+  while (true) {
+    // Iterate all instances in this module.
+    for (auto *record : *current) {
+      auto *target = record->getTarget();
+      // Filter out the edges that don't point in the SCC.
+      if (!scc.contains(target))
+        continue;
+      // If this node is already on our path, skip it.
+      if (visited.contains(target))
```
dtzSiFive, line 49:
```suggestion
      if (!visited.insert(target).second)
```
?

Since this is diagnostic code whatever you think is best.

lib/Dialect/FIRRTL/Transforms/CheckRecursiveInstantiation.cpp
```diff
@@ -0,0 +1,91 @@
...
+        continue;
+      auto &note = diag.attachNote(record->getInstance().getLoc());
+      note << record->getParent()->getModule().getModuleName();
+      note << " instantiates "
+           << record->getTarget()->getModule().getModuleName() << " here";
+      // If we have looped back to the starting point, return here.
+      if (target == first)
```
dtzSiFive, line 56: I'm not sure about this?

The following causes an infinite loop:
```
firrtl.circuit "SCCMore" {
  firrtl.module @SCCMore() {
    firrtl.instance a @A()
  }
  firrtl.module @A() {
    firrtl.instance b @B()
    firrtl.instance c @C()
  }
  firrtl.module @B() {
    firrtl.instance c @C()
    firrtl.instance d @D()
  }

  firrtl.module @C() {
    firrtl.instance a @A()
    firrtl.instance b @B()
  }

  firrtl.module @D() {
    firrtl.instance a @A()
    firrtl.instance c @C()
  }
}
```

(may not be minimal)

youngar: I had hastily adapted some code from clang-tidy, while trying to make everything simpler (replacing my proper depth first path search), which didn't work quite the way I expected.  I have decided I don't like how clang-tidy reports these errors, as the function they attach the error to does not necessarily show up in the recursive call chain they report.  So I ripped out this code and went with something even simpler, we just report all modules and instances which are involved in the SCC. I added a smaller version of your test case.

include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -968,4 +968,18 @@ def ProbesToSignals : Pass<"firrtl-probes-to-signals", "firrtl::CircuitOp"> {
   let constructor = "circt::firrtl::createProbesToSignalsPass()";
 }
 
+def CheckRecursiveInstantiation : Pass<"firrtl-check-recursive-instantiation",
+    "firrtl::CircuitOp"> {
+  let summary = "Check for illegal recursive instantiation";
+  let description = [{
+    This pass checks for illegal recursive module instantion. Recursive
```
dtzSiFive, line 975:
```suggestion
    This pass checks for illegal recursive module instantion.  Recursive
```
(I don't know what we standardize on, but later it's two spaces so went with that, just as long as it's consistent at least in same paragraph!)

test/Dialect/FIRRTL/check-recursive-instantiation-errors.mlir
```diff
@@ -0,0 +1,89 @@
...
+    firrtl.instance c @C()
+    // expected-note @below {{B instantiates A here}}
+    firrtl.instance a @A()
+  }
+  // expected-error @below {{recursive instantiation}}
+  firrtl.module @C() {
+    // expected-note @below {{C instantiates B here}}
+    firrtl.instance b @B()
+  }
+}
```
dtzSiFive, line 89: ... eof newline? 🥺

dtzSiFive (approved):
LGTM

Issue #7188, opened 2024-06-16
[CMake] Installation fails if SLANG_ENABLE_FRONTEND is enabled


cepheus69:
While I was trying to generate installation of CIRCT with enabled SLANG_ENABLE_FRONTEND option. The installation is failed. I went through the file `cmake_install.cmake`  and found snippets that should be relevant to this question.

```CMake
if(CMAKE_INSTALL_COMPONENT STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/include" TYPE DIRECTORY FILES "/home/ursa/Projects/circt/build/_deps/fmt-src/include/fmt")
endif()
```
It looks like an extra piece of error code was generated when installing the header files for the fmt library. The wrong code tries to find the fmt library headers in circt's include path.
```CMake
if(CMAKE_INSTALL_COMPONENT STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/include" TYPE FILE FILES
    "/home/ursa/Projects/circt/include/fmt/args.h"
    "/home/ursa/Projects/circt/include/fmt/chrono.h"
    "/home/ursa/Projects/circt/include/fmt/color.h"
    "/home/ursa/Projects/circt/include/fmt/compile.h"
    "/home/ursa/Projects/circt/include/fmt/core.h"
    "/home/ursa/Projects/circt/include/fmt/format.h"
    "/home/ursa/Projects/circt/include/fmt/format-inl.h"
    "/home/ursa/Projects/circt/include/fmt/os.h"
    "/home/ursa/Projects/circt/include/fmt/ostream.h"
    "/home/ursa/Projects/circt/include/fmt/printf.h"
    "/home/ursa/Projects/circt/include/fmt/ranges.h"
    "/home/ursa/Projects/circt/include/fmt/std.h"
    "/home/ursa/Projects/circt/include/fmt/xchar.h"
    )
endif()
```

cepheus69:
If remove the second piece of code in `cmake_install.cmake`. Then the installation can proceed normally.

cepheus69:
I'm trying to fix it. So are there any other details I'm missing? @fabianschuiki

SpriteOvO:
Can you provide the error message from CMake and commands you executed? Do you have the `fmt` package installed on your system?

[...]

PR #7437, opened 2024-08-05
[CMake] Fix install failure with CMAKE_SLANG_FRONTEND_ENABLED enabled

cepheus69:
Slang compiler relies on the third-party libraries like unordered_dense and fmt library. The fmt library provides two ways to integrated it:
  1.Headers-only
  2.Seperately compiled
The main purpose of this commit is to avoid installation failure of CIRCT project due to finding fmt header file in wrong path which is in circt `include` directory when CMake_slang_Frontend_enabled is turned on. We hope to not install header files coming from fmt library.


cepheus69:
This problem is mentioned previously in https://github.com/llvm/circt/issues/7188

fabianschuiki (approved):
Very nice! Can we do the same thing for `unordered_dense`?

cepheus69:
> Very nice! Can we do the same thing for `unordered_dense`?

Okay! I ‘ll add it😉

PR #7438, opened 2024-08-05
[Python][OM] Handle BoolAttr's before IntegerAttr's.

dtzSiFive:
Add integration test for bool constants that fails without this change.


dtzSiFive:
Haven't checked the integration test yet (before or after), marking draft until local build with python bindings enabled goes through (hopefully/probably our CI will come back first).

seldridge (approved):
LGTM

PR #7439, opened 2024-08-05
[FIRRTL] Verify main module is a module, and is public.

dtzSiFive:
Add tests for new errors and existing error (changed).

Op errors should remember to make sense when read prefixed with "mydialect.opname op ".


fabianschuiki (approved):
This feels like a reasonable set of constraints to put on the main module :smiley:

PR #7444, opened 2024-08-05
[Docs][FIRRTL] Build and include docs for intrinsic ops

### LINKED ISSUE #7443: [Docs][FIRRTL] Missing intrinsic ops docs
Documentation for ops defined in https://github.com/llvm/circt/blob/bec0deab4bfa79785e3027bc95548435f58bd294/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td is not shown on website https://circt.llvm.org/docs/Dialects/FIRRTL/.

SpriteOvO:
Fixes #7443.


dtzSiFive:
Thanks for the PR!!

Change LGTM from phone, haven't tried it to double-check but seems right to me!

Probably we should move more intrinsics into this section / file but that's not for this PR, just a detail now that we have this nicely integrated into the generated docs 😁.

SpriteOvO:
Thank @dtzSiFive for reviewing! I'm not sure how to invoke Hugo, so I didn't check HTML. At least in `./build/docs/Dialects/FIRRTLIntrinsicOps.md` I see the missing ops appearing.

PR #7449, opened 2024-08-06
[FIRRTL] Tweak printing of layers to avoid extra space.

dtzSiFive:
Before:

```mlir
firrtl.layer @A  inline
```

After:

```mlir
firrtl.layer @A inline
```


seldridge (approved):
Nice. Is it really that easy?

rwy7 (approved):

PR #7450, opened 2024-08-06
[circt-lec] Accept two MLIR inputs

uenoku:
Practically it is very useful to verify equivalence between modules in two different MLIR files. This commit changes `inputFilename` to a list and implements a very simple module merge that moves operations in the second module to the first by resolving the symbol.


tools/circt-lec/circt-lec.cpp
```diff
@@ -122,6 +122,65 @@ static cl::opt<OutputFormat> outputFormat(
...
+      if (oldSymbol == name) {
+        assert(!renamed && "symbol must be unique");
+        renamed = *result;
+      }
+    }
+  }
+
+  if (!name)
+    return src->emitError()
+           << "module " << name << " was not found in the second module";
```
TaoBi22, line 148: I'm a little confused on this error message - it looks like it's throwing when a name isn't given, rather than when a name isn't found? Maybe I missed some detail though!

tools/circt-lec/circt-lec.cpp
```diff
@@ -122,6 +122,65 @@ static cl::opt<OutputFormat> outputFormat(
 // Tool implementation
 //===----------------------------------------------------------------------===//
 
+// Move all operations in `src` to `dest`. Rename all symbols in `src` to avoid
+// conflict.
+static FailureOr<StringAttr> mergeModules(ModuleOp dest, ModuleOp src,
+                                          StringAttr name) {
+
+  SymbolTable destTable(dest), srcTable(src);
+  StringAttr renamed = {};
```
TaoBi22, line 131: Super-nit: Could we call this something like `newName`? To me `renamed` implies a boolean indicating whether it's been renamed.

TaoBi22 (approved):
Couple of nits, but looks good - thanks a lot for this!

tools/circt-lec/circt-lec.cpp
uenoku: Good catch, the condition must be `!newName`. I added a test :)

fabianschuiki:
Very nice! Thanks for adding this :partying_face:

PR #7453, opened 2024-08-07
[arcilator] Add an option to load shared lib into JIT engine

uenoku:
This PR adds an option `--shared-libs` to load shared lib in the JIT engine in a similar way to what circt-lec does for loading z3 shared lib.


maerhart (approved):
Really cool! Thanks! 🚀

fabianschuiki:
Awesome :heart_eyes:!

PR #7454, opened 2024-08-07
[ExportVerilog] Add a lowering option to fix up empty modules

uenoku:
This commit adds a new lowering option to sanitize empty modules by creating a dummy wire in it.


fabianschuiki (approved):
LGTM! I remember this coming up in various synthesis tools :see_no_evil:

PR #7455, opened 2024-08-07
[ESI] Add option to build runtime as a static library

mortbopet:


lib/Dialect/ESI/runtime/CMakeLists.txt
```diff
@@ -143,9 +164,11 @@ target_link_libraries(ESICppRuntime PUBLIC
   nlohmann_json::nlohmann_json
 )
 if(NOT MSVC)
-  target_link_libraries(ESICppRuntime PRIVATE
-    dl
-  )
+  if(NOT ESI_STATIC_RUNTIME)
+    target_link_libraries(ESICppRuntime PRIVATE
```
teqdruid, line 168: This is for loading other libraries (backends), so you still need it if we're compiling ESI Runtime as a static library.

lib/Dialect/ESI/runtime/cpp/lib/Accelerator.cpp
```diff
@@ -175,25 +175,48 @@ static void loadBackend(std::string backend) {
...
 namespace internal {
 
-static std::map<std::string, BackendCreate> backendRegistry;
-void registerBackend(std::string name, BackendCreate create) {
-  if (backendRegistry.count(name))
-    throw std::runtime_error("Backend already exists in registry");
-  backendRegistry[name] = create;
+class BackendRegistry {
+public:
+  static void registerBackend(const std::string &name, BackendCreate create) {
```
teqdruid, line 180: This (and `connect`) shouldn't be static. Only the `get` method should be.

lib/Dialect/ESI/runtime/cpp/lib/Accelerator.cpp
```diff
@@ -175,25 +175,48 @@ static void loadBackend(std::string backend) {
...
+      loadBackend(backend);
+      f = registry.find(backend);
+      if (f == registry.end())
+        throw std::runtime_error("Backend '" + backend + "' not found");
+    }
+    return f->second(ctxt, connection);
+  }
+
+private:
+  static BackendRegistry &get() {
```
teqdruid, line 203: Make this public.

lib/Dialect/ESI/runtime/cpp/lib/Accelerator.cpp
```diff
@@ -175,25 +175,48 @@ static void loadBackend(std::string backend) {
...
+    static BackendRegistry instance;
```
teqdruid, line 204: Does this not work as a private static pointer? (The usual way one would do this.) I'd hope that C++ would guarantee that all static fields in a particular class are initialized before any calls to static methods (but I may be giving C++ too much credit).

lib/Dialect/ESI/runtime/CMakeLists.txt
```diff
@@ -119,9 +132,17 @@ IF(MSVC)
 ENDIF(MSVC)
 
 # The core API. For now, compile the backends into it directly.
```
teqdruid, line 134: Can you fix this outdated comment while you're in here?

teqdruid (approved):

Issue #7051, opened 2024-05-16
[Calyx] Verified-valid IR crashes in canonicalizer


dtzSiFive:
Consider this snippet, created while trying to reduce example in #7050:

```mlir
module attributes {calyx.entrypoint = "main"} {
  calyx.component @main(%go: i1 {go}, %clk: i1 {clk}, %reset: i1 {reset}) -> (%done: i1 {done}) {
    %r.in, %r.write_en, %r.clk, %r.reset, %r.out, %r.done = calyx.register @r : i1, i1, i1, i1, i1, i1
    %eq.left, %eq.right, %eq.out = calyx.std_eq @eq : i1, i1, i1
    %c1_1 = hw.constant 1 : i1
    calyx.wires {
    }
    calyx.control {
      calyx.seq {
        calyx.if %eq.out {
          calyx.seq {
          }
        } else {
          calyx.seq {
          }
        }
      }
    }
  }
}
```

Feeding this through `circt-opt -canonicalize` crashes:

```
circt-opt: /home/will/src/sifive/circt/llvm/llvm/include/llvm/ADT/ilist_iterator.h:138: reference llvm::ilist_iterator<llvm::ilist_detail::node_options<mlir::Operation, true, false, void, false>, true, false>::operator*() const [OptionsT = llvm::ilist_detail::node_options<mlir::Operation, true, false, void, false>, IsReverse = true, IsConst = false]: Assertion `!NodePtr->isKnownSentinel()' failed.
PLEASE submit a bug report to https://github.com/llvm/circt and include the crash backtrace.
Stack dump:
0.	Program arguments: ./build/bin/circt-opt -canonicalize calyx-crash.mlir
 #0 0x00005560165c0346 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /home/will/src/sifive/circt/llvm/llvm/lib/Support/Unix/Signals.inc:723:13
 #1 0x00005560165be420 llvm::sys::RunSignalHandlers() /home/will/src/sifive/circt/llvm/llvm/lib/Support/Signals.cpp:106:18
 #2 0x00005560165c09eb SignalHandler(int) /home/will/src/sifive/circt/llvm/llvm/lib/Support/Unix/Signals.inc:413:1
 #3 0x00007f9bd7854eb0 __restore_rt (/nix/store/p3jshbwxiwifm1py0yq544fmdyy98j8a-glibc-2.38-27/lib/libc.so.6+0x3deb0)
[...]

PR #7456, opened 2024-08-07
[calyx] fix calyx canonicalization.

linuxlonelyeagle:
See [issue](https://github.com/llvm/circt/issues/7051).
Honestly, I saw this bug a few months ago and I think I fixed it, just like this PR did. But I don't think it was enough.
```
test.mlir:2:3: note: see current operation: 
"calyx.component"() ({
^bb0(%arg0: i1, %arg1: i1, %arg2: i1, %arg3: i1):
  %0:6 = "calyx.register"() {sym_name = "r"} : () -> (i1, i1, i1, i1, i1, i1)
  "calyx.wires"() ({
  ^bb0:
  }) : () -> ()
  "calyx.control"() ({
  ^bb0:
  }) : () -> ()
}) {function_type = (i1, i1, i1, i1) -> (), portAttributes = [{go}, {clk}, {reset}, {done}], portDirections = -8 : i4, portNames = ["go", "clk", "reset", "done"], sym_name = "main"} : () -> ()
```
The problem is as I said in the issue, solved he still reports an error because now `calyx.control` is empty after `-canonicalize`.
So a few months ago I tried to include stricter restrictions, but I didn't have too good an idea.
Now my idea, when checking calyx.control, that op must have `calyx.enable,calyx.invoke alive fsm.machine` present. but it also causes other problems.
On the one hand, I want to discuss it, and on the other hand, I think this PR makes sense.There is no test file because there are no examples of errors reported after -canonicalise in the existing test file.

### DIFF
--- a/lib/Dialect/Calyx/CalyxOps.cpp
+++ b/lib/Dialect/Calyx/CalyxOps.cpp
@@ -761,7 +761,7 @@ return WalkResult::interrupt();
                "The component currently does nothing. It needs to either have "
                "continuous assignments in the Wires region or control "
                "constructs in "
-               "the Control region.The Control region should involve ")
+               "the Control region. The Control region should contain at least one of ")
            << "'" << EnableOp::getOperationName() << "' , "
            << "'" << InvokeOp::getOperationName() << "' or "
            << "'" << fsm::MachineOp::getOperationName() << "'.";
--- a/lib/Dialect/Calyx/CalyxOps.cpp
+++ b/lib/Dialect/Calyx/CalyxOps.cpp
@@ -769,7 +769,6 @@ LogicalResult ComponentOp::verify() {
            << "'" << EnableOp::getOperationName() << "' , "
            << "'" << InvokeOp::getOperationName() << "' or "
            << "'" << fsm::MachineOp::getOperationName() << "'.";
-
   return success();
 }
 
--- a/lib/Dialect/Calyx/CalyxOps.cpp
+++ b/lib/Dialect/Calyx/CalyxOps.cpp
@@ -748,15 +748,11 @@ LogicalResult ComponentOp::verify() {
   // region, or continuous assignments.
   bool hasNoControlConstructs = true;
   getControlOp().walk<WalkOrder::PreOrder>([&](Operation *op) {
-    if (dyn_cast<EnableOp>(op))
-      hasNoControlConstructs = false;
-    if (dyn_cast<InvokeOp>(op))
-      hasNoControlConstructs = false;
-    if (dyn_cast<fsm::MachineOp>(op))
-      hasNoControlConstructs = false;
-    if (hasNoControlConstructs == false)
-      return WalkResult::interrupt();
-    return WalkResult::advance();
+if (isa<EnableOp, InvokeOp, fsm::MachineOp>(op)) {
+  hasNoControlConstructs = false;
+  return WalkResult::advance();
+}
+return WalkResult::interrupt();
   });
   bool hasNoAssignments =
       getWiresOp().getBodyBlock()->getOps<AssignOp>().empty();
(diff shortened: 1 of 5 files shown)


rachitnigam:
Thanks for creating this @linuxlonelyeagle! I think we should add a text with the three cases you mentioned so that future PRs don't break it. I'm also wondering if we should start canonicalizing away empty control programs (`if` with empty bodies etc.) so passes can make stronger assumptions about the shape of programs.

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -745,16 +746,29 @@ LogicalResult ComponentOp::verify() {
...
+  getControlOp().walk<WalkOrder::PreOrder>([&](Operation *op) {
+    if (dyn_cast<EnableOp>(op))
+      hasNoControlConstructs = false;
+    if (dyn_cast<InvokeOp>(op))
+      hasNoControlConstructs = false;
+    if (dyn_cast<fsm::MachineOp>(op))
+      hasNoControlConstructs = false;
+    if (hasNoControlConstructs == false)
+      return WalkResult::interrupt();
+    return WalkResult::advance();
```
cgyurgyik, line 759:
```suggestion
if (isa<EnableOp, InvokeOp, fsm::MachineOp>(op)) {
  hasNoControlConstructs = false;
  return WalkResult::advance();
}
return WalkResult::interrupt();
```

...would be the less verbose way of doing this.

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -745,16 +746,29 @@ LogicalResult ComponentOp::verify() {
...
   if (hasNoControlConstructs && hasNoAssignments)
     return emitOpError(
-        "The component currently does nothing. It needs to either have "
-        "continuous assignments in the Wires region or control constructs in "
-        "the Control region.");
-
+               "The component currently does nothing. It needs to either have "
+               "continuous assignments in the Wires region or control "
+               "constructs in "
+               "the Control region.The Control region should involve ")
```
cgyurgyik, line 768:
```suggestion
               "the Control region. The Control region should contain at least one of ")
```

test/Dialect/Calyx/errors.mlir
```diff
@@ -899,6 +899,30 @@ module attributes {calyx.entrypoint = "main"} {
...
+    %c1_1 = hw.constant 1 : i1
+    calyx.wires {
+    }
+    calyx.control {
+      calyx.seq {
+        calyx.if %eq.out {
+          calyx.seq {
+          }
+        } else {
+          calyx.seq {
```
cgyurgyik, line 918: For completeness, can you make this a `calyx.par`?

cgyurgyik (approved):
A few smaller comments, but otherwise LGTM.

cgyurgyik:
> I'm also wondering if we should start canonicalizing away empty control programs (`if` with empty bodies etc.) so passes can make stronger assumptions about the shape of programs.

+1

lib/Dialect/Calyx/CalyxOps.cpp
linuxlonelyeagle: Hahaha, graduated and played for a few months, I do remember a cast that could cast multiple types, but I forgot, thanks for the suggestion. It's a bit late now, I'll add what you said tomorrow.

lib/Dialect/Calyx/CalyxOps.cpp
```diff
@@ -745,16 +746,26 @@ LogicalResult ComponentOp::verify() {
...
       getWiresOp().getBodyBlock()->getOps<AssignOp>().empty();
   if (hasNoControlConstructs && hasNoAssignments)
     return emitOpError(
-        "The component currently does nothing. It needs to either have "
-        "continuous assignments in the Wires region or control constructs in "
-        "the Control region.");
-
+               "The component currently does nothing. It needs to either have "
+               "continuous assignments in the Wires region or control "
+               "constructs in "
```
cgyurgyik, line 763: Not can you reformat this manually so that it's not just two words on one line?

cgyurgyik (approved):
LGTM, I see this as an incremental improvement. Thanks!

linuxlonelyeagle:
Can you help me merge codes? Thanks!

Issue #7458, opened 2024-08-07
[SimToSV] Add includes guard to SV import


uenoku:
input fir file:
```firrtl
FIRRTL version 4.0.0
circuit DPI:
  module Child :
    input clock: Clock
    input enable: UInt<1>
    input in: UInt<8>[1]

    intrinsic(circt_dpi_call<isClocked = 1, functionName="clocked_void">, clock, enable, in[0])

  public module DPI:
    input clock: Clock
    input enable: UInt<1>
    input in: UInt<8>[1]

    intrinsic(circt_dpi_call<isClocked = 1, functionName="clocked_void">, clock, enable, in[0])
    inst child of Child
    child.clock <= clock
    child.enable <= enable
    child.in <= in
```
`firtool bar.fir -split-verilog -o design` produces following verilog files:

```verilog
$ cat design/DPI.sv
import "DPI-C" context function void clocked_void(
  input byte in_0
);


module DPI(
  input       clock,
              enable,
  input [7:0] in_0
);

  always @(posedge clock) begin
    if (enable)
      clocked_void(in_0);
  end // always @(posedge)
  Child child (
    .clock  (clock),
    .enable (enable),
    .in_0   (in_0)
  );
endmodule

$ cat design/Child.sv
// Generated by CIRCT firtool-1.80.1
import "DPI-C" context function void clocked_void(
  input byte in_0
);


module Child(
  input       clock,
              enable,
  input [7:0] in_0
);

  always @(posedge clock) begin
    if (enable)
      clocked_void(in_0);
  end // always @(posedge)
endmodule
```
However verilator complains duplicate DPI import declarations:
```
$ verilator --cc design/Child.sv design/DPI.sv  --top-module DPI
%Error: design/DPI.sv:2:38: Duplicate declaration of task: 'clocked_void'
    2 | import "DPI-C" context function void clocked_void(
      |                                      ^~~~~~~~~~~~
        design/Child.sv:2:38: ... Location of original declaration
    2 | import "DPI-C" context function void clocked_void(
      |                                      ^~~~~~~~~~~~
%Error: Exiting due to 1 error(s)
```

SV spec is a bit ambiguous about duplicate import statements that are exactly same but the behavior of verilator is reasonable. We should add include guards for import statements.

PR #7459, opened 2024-08-07
[SimToSV] Add include guards to DPI import

uenoku:
This adds include guards `__CIRCT_DPI_IMPORT_*` to DPI import statements generated in SimToSV. 

Fix https://github.com/llvm/circt/issues/7458.


fabianschuiki (approved):
LGTM!

It's pretty convenient that the `*.dpi` op is just a marker that points at a func declaration somewhere else. Moving the actual func with the `@...` name into would probably have been a lot more annoying, right?

uenoku:
> Moving the actual function with the actual @... symbol name into would probably have been a lot more annoying, right?

Yes the function declaration must be a child of SymbolTable op and we cannot nest that.

Issue #7457, opened 2024-08-07
[ESI][Runtime] Teardown: ownership


teqdruid:
When an `AcceleratorConnection` is torn down and destructed, accesses to its associated `Accelerator`s and their associated ports will be memory violations. This is (fairly) well documented in the API, but isn't ideal. `Accelerator` shouldn't be owned by `AcceleratorConnection`. The associated ports should merely throw exceptions on accesses to disconnected or destructed connections.

The basic idea is that `Accelerator` and everything accessible from it should merely be a facade.

PR #7460, opened 2024-08-07
[ESI][Runtime] Poll method and optional service thread polling

teqdruid:
Add a poll method to ports, a master poll method to the Accelerator, and the ability to poll from the service thread. Also, only spin up the service thread if it's requested.

The service thread polling (in particular) required some ownership changes: Accelerator objects now belong to the AcceleratorConnection so that the ports aren't destructed before the service thread gets shutdown (which causes an invalid memory access). This particular binding isn't ideal, is brittle, and will be an issue for anything doing the polling. Resolving #7457 should mitigate this issue.

Backends are now _required_ to call `disconnect` in their destructor.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
+++ b/lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
@@ -42,14 +42,14 @@ public:
   virtual void disconnect() = 0;
   virtual bool isConnected() const = 0;
 
-  /// Poll for incoming data. Returns true if data was read into a buffer as a
-  /// result of the poll. Calling the call back could (will) also happen in that
-  /// case. Some backends need this to be called periodically. In the usual
-  /// case, this will be called by a background thread, but the ESI runtime does
-  /// not want to assume that the host processes use standard threads. If the
-  /// user wants to provide their own threads, they need to call this on each
-  /// port occasionally. This is also called from the 'master' poll method in
-  /// the Accelerator class.
+  /// Poll for incoming data. Returns true if data was read or written into a
+  /// buffer as a result of the poll. Calling the call back could (will) also
+  /// happen in that case. Some backends need this to be called periodically. In
+  /// the usual case, this will be called by a background thread, but the ESI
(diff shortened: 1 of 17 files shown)


lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
```diff
@@ -33,21 +33,38 @@ namespace esi {
...
   /// Set up a connection to the accelerator. The buffer size is optional and
   /// should be considered merely a hint. Individual implementations use it
   /// however they like. The unit is number of messages of the port type.
-  virtual void connect(std::optional<unsigned> bufferSize = std::nullopt) {
-    connectImpl(bufferSize);
+  virtual void connect(std::optional<unsigned> bufferSize = std::nullopt) = 0;
+  virtual void disconnect() = 0;
+  virtual bool isConnected() const = 0;
+
+  /// Poll for incoming data. Returns true if data was read into a buffer as a
```
mortbopet, line 45: "read into a buffer" ... or "written to a buffer"? I'm assuming it's not just for read ports.

lib/Dialect/ESI/runtime/cpp/include/esi/Ports.h
```diff
@@ -33,21 +33,38 @@ namespace esi {
...
 
   const Type *getType() const { return type; }
 
-private:
+protected:
   const Type *type;
 
+  /// Method called by poll() to actually poll the channel if the channel is
+  /// connected.
+  virtual bool pollImpl() { return false; }
```
mortbopet, line 66: Shouldn't a user get some kind of feedback if they called this on something which they expected to have a poll implementation?

teqdruid: Feedback of what sort? Whether or not one has to run this function is very much backend dependent so I don't want like an exception being thrown if a user calls it on a backend which doesn't need it.

mortbopet (approved):

PR #7461, opened 2024-08-08
[Moore] Add FVIntegerAttr

fabianschuiki:
Add the `FVIntegerAttr`, an attribute containing an `FVInt` value. This allows four-valued integer constants to be used as attributes on operations and for constant folding to occur on such values. In contrast to the builtin `IntegerAttr`, `FVIntegerAttr` does not have a type yet. The type can be added later as soon as we have a concrete use case.

I expect us to eventually move `FVIntegerAttr` into the HW dialect once we are happy with its design. Other parts of CIRCT will eventually want to reason about four-valued integers as well.

The parsing and printing of the attribute tries to make the `FVInt` read like a plain old `APInt` when there are no X or Z bits present. Otherwise it falls back to printing as hexadecimal or binary number. To distinguish the different representations and to allow constants with X or Z to be parsed as keywords, a `h` and `b` prefix is used for the hexadecimal and binary formatting, respectively.

Examples of the attribute:

    #moore.fvint<42 : 32>
    #moore.fvint<-42 : 32>
    #moore.fvint<1234567890123456789012345678901234567890 : 131>
    #moore.fvint<hABCDEFXZ0123456789 : 72>
    #moore.fvint<b1010XZ01 : 8>

### DIFF
--- a/include/circt/Dialect/Moore/CMakeLists.txt
+++ b/include/circt/Dialect/Moore/CMakeLists.txt
@@ -9,10 +9,8 @@ mlir_tablegen(MooreEnums.cpp.inc -gen-enum-defs)
 add_public_tablegen_target(CIRCTMooreEnumsIncGen)
 add_dependencies(circt-headers CIRCTMooreEnumsIncGen)
 
-mlir_tablegen(MooreAttributes.h.inc -gen-attrdef-decls
-  -attrdefs-dialect MooreDialect)
-mlir_tablegen(MooreAttributes.cpp.inc -gen-attrdef-defs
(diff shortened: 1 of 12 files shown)


mikeurbach (approved):
I think the boilerplate to add the attribute looks good. The printer/parser is nice, looks good to me.

uenoku (approved):
Awesome!

hailongSun2000 (approved):
Learned some new coding. LGTM :tada:!

fabianschuiki:
Thanks for the reviews :smiley: Going to land this.

PR #7464, opened 2024-08-08
[FIRRTL] Don't capture non-passives in LowerLayers

seldridge:
Fix a bug where the LowerLayers pass could create output ports on a module lowered from a bind convention layer.  Avoid this entirely by moving all subfield, subindex, and subaccess operations (sub-* ops) out of layerblocks before modules are created.

LowerLayers works by converting "captured" values into ports.  However, the sub-* ops may capture non-passive types allowably, but then never drive them.  E.g., consider the following:

    %0 = firrtl.wire : !firrtl.bundle<a : uint<1>, b flip: uint<1>>
    firrtl.layerblock @A {
      %1 = firrtl.subfield %0[a] : !firrtl.bundle<a : uint<1>, b flip: uint<1>>
      %2 = firrtl.node %1 : !firrtl.uint<1>
    }

Naively, this "captures" the non-passive %0.  However, this is really only capturing the _passive portion_ of %0 through a subfield.  Without this commit, LowerLayers will try to create a port with the same type as %0 when it should be creating a port with the same type of %1.  In order to determine what the port is, LowerLayers needs to know what is actually captured and not blindly assume that anything captured needs to be a port.

Because this analysis may be tricky, instead solve this by moving the sub-* ops outside the layerblock before computing captures.  The captures can then _continue_ to be naively computed by seeing if a value is defined outside the layerblock.

#### Metadata

This is stacked on https://github.com/llvm/circt/pull/7462.

This fixes the failure observed on this PR: https://github.com/chipsalliance/chisel/pull/4323

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -497,37 +497,46 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
         continue;
       }
 
-      // If this operation has non-passive operands defined outside the
-      // layerblock, then it may result in a lowering that creates output ports
-      // on the module.  This is illegal in the FIRRTL spec since no layerblock
-      // may write to values outside it.  Try to fix this up if possible.
-      if (llvm::any_of(op.getOperands(), [&](Value a) {
-            auto type = type_dyn_cast<FIRRTLBaseType>(a.getType());
-            return type && !type.isPassive();
-          })) {
-        // If this is a subfield, subindex, or subaccess op, then try to fix the
-        // problem by moving the value outside the layerblock.  This always
-        // works for subfield and subindex, but may fail for subaccess if the
-        // index is defined inside the layerblock.
-        if (isa<SubfieldOp, SubindexOp, SubaccessOp>(op)) {
-          if (llvm::none_of(op.getOperands(), [&](Value a) {
-                return isAncestorOfValueOwner(layerBlock, a);
-              })) {
-            op.moveBefore(layerBlock);
-            continue;
-          }
-          auto diag =
-              op.emitOpError()
-              << "has a non-passive operand and captures a value defined "
-                 "outside its enclosing bind-convention layerblock.  The "
-                 "'LowerLayers' pass cannot lower this as it would create an "
-                 "output port on the resulting module.";
-          diag.attachNote(layerBlock.getLoc())
-              << "the layerblock is defined here";
-          return WalkResult::interrupt();
-        }
+      // Handle subfields, subindexes, and subaccesses which are indexing into
+      // non-passive values.  If these are kept in the module, then the module
+      // must create bi-directional ports.  This doesn't make sense as the
+      // FIRRTL spec states that no layerblock may write to values outside it.
+      // Fix this for subfield and subindex by moving these ops outside the
+      // layerblock.  Try to fix this for subaccess and error if the move can't
+      // be made because the index is defined inside the layerblock.  (This case
+      // is exceedingly rare given that subaccesses are almost always unexepcted
+      // when this pass runs.)
+      if (isa<SubfieldOp, SubindexOp>(op)) {
+        auto input = op.getOperand(0);
+        if (!firrtl::type_cast<FIRRTLBaseType>(input.getType()).isPassive() &&
+            !isAncestorOfValueOwner(layerBlock, input))
+          op.moveBefore(layerBlock);
+        continue;
       }
 
+      if (auto subOp = dyn_cast<SubaccessOp>(op)) {
+        auto input = subOp.getInput();
+        if (firrtl::type_cast<FIRRTLBaseType>(input.getType()).isPassive())
+          continue;
+
+        if (!isAncestorOfValueOwner(layerBlock, input) &&
+            !isAncestorOfValueOwner(layerBlock, subOp.getIndex())) {
+          subOp->moveBefore(layerBlock);
+          continue;
+        }
+        auto diag = op.emitOpError()
+                    << "has a non-passive operand and captures a value defined "
+                       "outside its enclosing bind-convention layerblock.  The "
+                       "'LowerLayers' pass cannot lower this as it would "
+                       "create an output port on the resulting module.";
+        diag.attachNote(layerBlock.getLoc())
+            << "the layerblock is defined here";
+        return WalkResult::interrupt();
+      }
+
+      // Beyond this point, we are handling operations which capture values
+      // defined outside the layerblock.  Whenever we see this, we need to
+      // create ports for the module that this layerblock will become.
       if (auto refSend = dyn_cast<RefSendOp>(op)) {
         auto src = refSend.getBase();
         if (!isAncestorOfValueOwner(layerBlock, src))


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -490,6 +490,17 @@ void LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+      // If this operation is a subfield op and it is capturing something
+      // non-passive outside the layerblock, then move this outside the
+      // layerblock.  This avoids creating output ports on the created module.
+      if (isa<SubfieldOp, SubindexOp, SubaccessOp>(op)) {
+        if (llvm::none_of(op.getOperands(),
+                          [&](Value a) { return isAncestor(layerBlock, a); })) {
+          op.moveBefore(layerBlock);
+          continue;
+        }
+      }
```
youngar, line 502: I don't think this does the right thing for subaccess ops when the index is defined inside the layerblock.
```
firrtl.circuit "Foo" {
  firrtl.layer @A bind { }
  firrtl.module @Foo(
    in %a: !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>,
    in %b: !firrtl.uint<1>
  ) {
    firrtl.layerblock @A {
      %n = firrtl.node %b : !firrtl.uint<1>
      %0 = firrtl.subaccess %a[%b] : !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, !firrtl.uint<1>
    }
  }
}
```
```
./layers.mlir:9:12: error: 'firrtl.matchingconnect' op operand #0 must be a sized passive base type (contains no uninferred widths, or flips) or foreign type, but got '!firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>'
      %0 = firrtl.subaccess %a[%b] : !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, !firrtl.uint<1>
           ^
./layers.mlir:9:12: note: see current operation: "firrtl.matchingconnect"(%0#1, %arg0) : (!firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>) -> ()
// -----// IR Dump After LowerLayers Failed (firrtl-lower-layers) //----- //
"firrtl.circuit"() ({
  "sv.verbatim"() {format_string = "`ifndef layers_Foo_A\0A`define layers_Foo_A", output_file = #hw.output_file<"layers_Foo_A.sv", excludeFromFileList>, symbols = []} : () -> ()
  "firrtl.module"() ({
  ^bb0(%arg2: !firrtl.uint<1>, %arg3: !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>):
    %1 = "firrtl.node"(%arg2) {annotations = [], name = "n", nameKind = #firrtl<name_kind droppable_name>} : (!firrtl.uint<1>) -> !firrtl.uint<1>
    %2 = "firrtl.subaccess"(%arg3, %arg2) : (!firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, !firrtl.uint<1>) -> !firrtl.vector<bundle<a: uint<1>, b flip: uint<2>>, 2>
  }) {annotations = [], convention = #firrtl<convention internal>, layers = [], portAnnotations = [], portDirections = array<i1: false, false>, portLocations = [loc("./layers.mlir":8:12), loc("./layers.mlir":9:12)], portNames = ["b", "a"], portSyms = [], portTypes = [!firrtl.uint<1>, !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>], sym_name = "Foo_A", sym_visibility = "private"} : () -> ()
  "firrtl.module"() ({
  ^bb0(%arg0: !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, %arg1: !firrtl.uint<1>):
    %0:2 = "firrtl.instance"() {annotations = [], layers = [], lowerToBind, moduleName = @Foo_A, name = "a", nameKind = #firrtl<name_kind droppable_name>, output_file = #hw.output_file<"layers_Foo_A.sv", excludeFromFileList>, portAnnotations = [[], []], portDirections = array<i1: false, false>, portNames = ["b", "a"]} : () -> (!firrtl.uint<1>, !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>)
    "firrtl.matchingconnect"(%0#1, %arg0) : (!firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>) -> ()
    "firrtl.matchingconnect"(%0#0, %arg1) : (!firrtl.uint<1>, !firrtl.uint<1>) -> ()
  }) {annotations = [], convention = #firrtl<convention internal>, layers = [], parameters = [], portAnnotations = [], portDirections = array<i1: false, false>, portLocations = [loc("./layers.mlir":4:8), loc("./layers.mlir":5:8)], portNames = ["a", "b"], portSyms = [], portTypes = [!firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, !firrtl.uint<1>], sym_name = "Foo"} : () -> ()
  "sv.verbatim"() {format_string = "`endif // layers_Foo_A", output_file = #hw.output_file<"layers_Foo_A.sv", excludeFromFileList>, symbols = []} : () -> ()
}) {annotations = [], name = "Foo"} : () -> ()
```

seldridge: For the example, this does error. I will take a look. However, I think the example is supposed to use `%n` as the index and not `%b`?

``` mlir
    firrtl.layerblock @A {
      %n = firrtl.node %b : !firrtl.uint<1>
      %0 = firrtl.subaccess %a[%n] : !firrtl.vector<vector<bundle<a: uint<1>, b flip: uint<2>>, 2>, 2>, !firrtl.uint<1>
    }
```

seldridge: I pushed a change which detects this situation and errors. I'm going to bucket this as improvements to the pass. Actually fixing this is going to be complicated, however.

Does that seem acceptable?

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -497,6 +497,37 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
         continue;
       }
 
+      // If this operation has non-passive operands defined outside the
+      // layerblock, then it may result in a lowering that creates output ports
+      // on the module.  This is illegal in the FIRRTL spec since no layerblock
+      // may write to values outside it.  Try to fix this up if possible.
+      if (llvm::any_of(op.getOperands(), [&](Value a) {
```
dtzSiFive, line 504: Nit: could probably check for the ops first to avoid scanning all operands but shrug.

dtzSiFive: (I don't need this addressed to land)

seldridge: Definitely better. Changed to do this in: https://github.com/llvm/circt/pull/7464/commits/d540ee6b412b2c578df0c50c5024b73cb0f4c12c

seldridge:
Landed in https://github.com/llvm/circt/commit/73c053de32941656332d500e957cf6bef89df91f and https://github.com/llvm/circt/commit/73c053de32941656332d500e957cf6bef89df91f. I fat fingered this one...

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: Good suggestion. Addressed in the fixup: https://github.com/llvm/circt/pull/7464/commits/d540ee6b412b2c578df0c50c5024b73cb0f4c12c

PR #7469, opened 2024-08-08
[FIRRTL][Dedup] Reduce size of integer data hashed.

dtzSiFive:
Don't always zext to size_t.

Use smaller integer sizes for ValueId and add to hash together.

Small performance win.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -97,8 +97,8 @@ struct ModuleInfo {
 /// For BlockArgument's, this is the argument number.
 /// For OpResult's, this is the result number.
 struct ValueId {
-  uint64_t index;
-  uint64_t offset;
+  unsigned index;
+  unsigned offset;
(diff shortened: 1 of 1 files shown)


mikeurbach (approved):
Makes sense to me.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -154,16 +154,20 @@ struct StructuralHasher {
...
 
 private:
-  void update(const void *pointer) {
-    auto *addr = reinterpret_cast<const uint8_t *>(&pointer);
-    sha.update(ArrayRef<uint8_t>(addr, sizeof pointer));
-  }
-
-  void update(size_t value) {
+  /// Hash value as array of bytes.
+  template <typename T>
```
dtzSiFive, line 158: FWIW:
If I could confidently enable_if this to ensure it doesn't get used with things that have possibly unset padding or short reads or something I would but instead I just wrap it for types that are definitely safe and cover our needs.

PR #7473, opened 2024-08-08
[MooreToCore] Fix conversion and dyn_extract operation lowering

maerhart:
The lowering pattern of these two operations crashed when used on aggregate types. This PR adds type conversions for array and struct types and fixes the two lowering patterns by adding support for these types


fabianschuiki (approved):
LGTM! Very cool :partying_face:

PR #7474, opened 2024-08-08
[Seq] Add a new pass to sink clock_gate ops to the users

prithayan:
This is a new pass that sinks `seq.clock_gate` into its users and  attempts to remove as many `seq.clock_gate` ops as possible.

Any operation that can potentially encode the enable directly, will replace its gated clock with the un-gated clock and encode the clock gate enable as the operation's enable signal.
Currently only the `CompRegOp` operation is handled, but similar transformation needs to be attempted on other supported operations also. 

The first phase of the pass is a global transformation that continues to push the clock gates into the referenced modules, until none of the instance ops are driven by a gated clock.
Next, best effort is made to remove the gated clock and encode the enable directly into the operation that is being driven by the clock.

### DIFF
--- a/include/circt/Dialect/HW/HWStructure.td
+++ b/include/circt/Dialect/HW/HWStructure.td
@@ -231,6 +231,10 @@ def HWModuleOp : HWModuleOpBase<"module",
 
     /// Verifies the body of the function.
     LogicalResult verifyBody();
+
+    void appendPorts(
+    ArrayRef<std::tuple<StringAttr, Type, Location>> insertInputs,
+    ArrayRef<std::pair<StringAttr, Value>> insertOutputs);
   }];
 
   let hasCustomAssemblyFormat = 1;
--- a/include/circt/Dialect/Seq/SeqPasses.td
+++ b/include/circt/Dialect/Seq/SeqPasses.td
@@ -107,4 +107,11 @@ def LowerSeqShiftReg : Pass<"lower-seq-shiftreg", "hw::HWModuleOp"> {
   let dependentDialects = ["circt::hw::HWDialect"];
 }
 
+def SinkClockGates : Pass<"sink-clock-gates", "mlir::ModuleOp"> {
+  let summary = "Sink the gated clock enables into the op that is driven by the clock.";
+  let constructor = "circt::seq::createSinkClockGatesPass()";
+  let dependentDialects = ["circt::hw::HWDialect", "circt::comb::CombDialect"];
+}
+
+
 #endif // CIRCT_DIALECT_SEQ_SEQPASSES
--- a/lib/Dialect/Seq/Transforms/SinkClockGates.cpp
+++ b/lib/Dialect/Seq/Transforms/SinkClockGates.cpp
@@ -1,0 +1,260 @@
+//===- SinkClockGates.cpp - Sink clock gates to its users. -----===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/HW/HWAttributes.h"
+#include "circt/Dialect/HW/HWInstanceGraph.h"
+#include "circt/Dialect/HW/HWOpInterfaces.h"
+#include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/Seq/SeqOps.h"
+#include "circt/Dialect/Seq/SeqPasses.h"
+#include "circt/Dialect/Seq/SeqTypes.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/MLIRContext.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringExtras.h"
+#include "llvm/ADT/Twine.h"
+#include "llvm/Support/Debug.h"
+#include <string>
+
+namespace circt {
+namespace seq {
+#define GEN_PASS_DEF_SINKCLOCKGATES
+#include "circt/Dialect/Seq/SeqPasses.h.inc"
+} // namespace seq
+} // namespace circt
+
+#define DEBUG_TYPE "sink-clock-gates"
+
+using namespace circt;
+using namespace seq;
+using namespace hw;
+
+namespace {
+struct SinkClockGatesPass
+    : public circt::seq::impl::SinkClockGatesBase<SinkClockGatesPass> {
+  using SinkClockGatesBase::SinkClockGatesBase;
+
+  // TODO: Can be moved to some utils !!
+  /// Append inputs to the instance, and also update the referenced modules.
+  void appendInputsToInstance(HWInstanceLike instance,
+                              ArrayRef<std::pair<StringAttr, Value>> inputs,
+                              ArrayRef<HWModuleOp> refModules) {
+
+    SmallVector<Attribute> argNames;
+    if (auto args = instance->getAttrOfType<ArrayAttr>("argNames"))
+      argNames.insert(argNames.begin(), args.begin(), args.end());
+
+    SmallVector<Value> newOperands;
+    newOperands.reserve(inputs.size());
+    SmallVector<std::tuple<StringAttr, Type, Location>> newInputs;
+    newInputs.reserve(inputs.size());
+    for (const auto &[name, value] : inputs) {
+      // If no argNames attribute, donot add it.
+      if (!argNames.empty())
+        argNames.push_back(name);
+      // Additional operands to the Instance op.
+      newOperands.push_back(value);
+      // Prepare the additional input ports that need to be added to the
+      // referenced module.
+      newInputs.emplace_back(name, value.getType(), value.getLoc());
+    }
+    // Add the new operands to the instance.
+    instance->insertOperands(instance->getNumOperands(), newOperands);
+    if (!argNames.empty())
+      instance->setAttr("argNames",
+                        ArrayAttr::get(instance->getContext(), argNames));
+    // Update all the modules with the new input ports.
+    for (auto refMod : refModules) {
+      refMod.appendPorts(newInputs, {});
+    }
+  }
+
+  /// Get the referenced modules for an instance.
+  SmallVector<HWModuleOp, 1> getReferencedModules(HWInstanceLike inst) {
+    SmallVector<HWModuleOp, 1> refMods;
+    for (auto refModName : inst.getReferencedModuleNamesAttr()) {
+      auto *node = graph->lookup(cast<StringAttr>(refModName));
+      if (auto instModule = dyn_cast_or_null<HWModuleOp>(*node->getModule()))
+        refMods.push_back(instModule);
+    }
+    return refMods;
+  }
+
+  /// If the `argNames` attribute exists, return it.
+  SmallVector<StringAttr> getInstanceArgNames(HWInstanceLike inst) {
+    SmallVector<StringAttr> argNames;
+    if (auto args = inst->getAttrOfType<ArrayAttr>("argNames")) {
+      auto attr = args.getAsRange<StringAttr>();
+      argNames.insert(argNames.begin(), attr.begin(), attr.end());
+    }
+    return argNames;
+  }
+
+  /// Combine two consecutive clock_gate ops, into one.
+  void collapseConsecutiveClockGates(ClockGateOp driverGate,
+                                     ClockGateOp userGate) {
+    mlir::OpBuilder builder(context);
+    builder.setInsertionPoint(userGate);
+    Value enable = getEnable(driverGate);
+    enable = builder.create<comb::AndOp>(userGate->getLoc(), enable,
+                                         userGate.getEnable());
+    userGate.getInputMutable().assign(driverGate.getInput());
+    userGate.getEnableMutable().assign(enable);
+  }
+
+  /// If the clock_gate op has `test_enable`, combine it with the `enable` and
+  /// return the final enable.
+  Value getEnable(ClockGateOp clkGate) {
+    Value enable = clkGate.getEnable();
+    // If there is testEnable, or it.
+    if (auto testEnable = clkGate.getTestEnable()) {
+      OpBuilder builder(clkGate.getContext());
+      builder.setInsertionPointAfter(clkGate);
+      enable =
+          builder.create<comb::OrOp>(clkGate->getLoc(), enable, testEnable);
+    }
+    return enable;
+  }
+
+  void runOnOperation() override;
+
+private:
+  InstanceGraph *graph;
+  DenseSet<Operation *> opsToErase;
+  mlir::MLIRContext *context;
+};
+
+void SinkClockGatesPass::runOnOperation() {
+  graph = &getAnalysis<circt::hw::InstanceGraph>();
+  DenseSet<HWInstanceLike> instancesWithGatedClk;
+  DenseSet<Operation *> opsWithGatedClock;
+  context = getOperation().getContext();
+
+  // Record all the instances that have clock_gate as an input. This will be
+  // used to push the clock_gate down through the hierarchy to its users.
+  // This also collapses consecutive clock gates.
+  auto findInstancesWithGatedClock = [&](HWModuleOp hwModule) {
+    hwModule->walk([&](ClockGateOp clkGate) {
+      bool canErase = true;
+      for (auto &use : clkGate->getUses()) {
+        auto *useOp = use.getOwner();
+        if (auto inst = dyn_cast<HWInstanceLike>(useOp))
+          instancesWithGatedClk.insert(inst);
+        else if (auto userGate = dyn_cast<ClockGateOp>(useOp)) {
+          collapseConsecutiveClockGates(clkGate, userGate);
+        } else {
+          canErase = false;
+          opsWithGatedClock.insert(useOp);
+        }
+      }
+      if (canErase)
+        opsToErase.insert(clkGate);
+    });
+  };
+  // Seed phase, find the instances to start the traversal from.
+  // These instances have a gated clock as input.
+  getOperation()->walk(
+      [&](hw::HWModuleOp hwModule) { findInstancesWithGatedClock(hwModule); });
+
+  // Transformation phase, push the clock_gate through the instances and find
+  // all the indirect instances that have a gated clock.
+
+  // Iterate until none of the instances have a gated clock as input.
+  while (!instancesWithGatedClk.empty()) {
+
+    // Remove an entry from the set.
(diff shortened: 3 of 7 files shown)


lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,260 @@
...
+  void collapseConsecutiveClockGates(ClockGateOp driverGate,
+                                     ClockGateOp userGate) {
+    mlir::OpBuilder builder(context);
+    builder.setInsertionPoint(userGate);
+    Value enable = getEnable(driverGate);
+    enable = builder.create<comb::AndOp>(userGate->getLoc(), enable,
+                                         userGate.getEnable());
+    userGate.getInputMutable().assign(driverGate.getInput());
+    userGate.getEnableMutable().assign(enable);
+  }
```
fabianschuiki, line 110: This is very neat! Might also make sense as a canonicalization on `ClockGateOp`.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,260 @@
...
+    Value enable = clkGate.getEnable();
+    // If there is testEnable, or it.
+    if (auto testEnable = clkGate.getTestEnable()) {
+      OpBuilder builder(clkGate.getContext());
+      builder.setInsertionPointAfter(clkGate);
+      enable =
+          builder.create<comb::OrOp>(clkGate->getLoc(), enable, testEnable);
+    }
+    return enable;
+  }
```
fabianschuiki, line 124: I'm wondering if we'd need to keep the regular `enable` and the `test_enable` separate… probably not?

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,260 @@
...
+    // Add the new operands to the instance.
+    instance->insertOperands(instance->getNumOperands(), newOperands);
+    if (!argNames.empty())
+      instance->setAttr("argNames",
+                        ArrayAttr::get(instance->getContext(), argNames));
+    // Update all the modules with the new input ports.
+    for (auto refMod : refModules) {
+      refMod.appendPorts(newInputs, {});
+    }
+  }
```
fabianschuiki, line 77: This is a very handy function! Should this potentially accept multiple (instance, inputs) tuples, since a module might be instantiated multiple times and you'd need to update each of the instances in turn?

fabianschuiki (commented):
Awesome! This is a very useful transform 😍. Does the code also work if a module is instantiated multiple times? And if one of the instances has a clock gate in front and the other does not?

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,260 @@
...
+    // The new enable ports that will be added to the instance.
+    SmallVector<std::pair<StringAttr, Value>> enablePorts;
+    // The instance may or maynot have argNames attribute.
+    SmallVector<StringAttr> argNames = getInstanceArgNames(inst);
+    // Iterate over all instance operands to find a gated clock.
+    for (auto [index, in] : llvm::enumerate(inst->getOperands())) {
+      if (!isa<seq::ClockType>(in.getType()))
+        continue;
+      auto clkGate = dyn_cast_or_null<ClockGateOp>(in.getDefiningOp());
+      if (!clkGate)
```
uenoku, line 192: This seems like a bit fragile since it could break analysis when there is wire between instance and clock.

uenoku:
> And if one of the instances has a clock gate in front and the other does not?

Yeah I think this transformation requires fixed-point analysis whose lattice is a pair of clock and enable.

prithayan:
Thanks a lot @fabianschuiki  and @uenoku for the feedback. 
The major issue in the previous iteration was multiple instances.
I have handled that in this version.
The other concern regarding the precision of this analysis, this is intended as a fast/simple and best-effort analysis, so its okay to miss a few clock_gates. We donot intend to eliminate all the clock gates.
Additionally clock_gates to some public-modules/extern-modules cannot be removed.
As Hideto suggested, I am also working on a fixed-point analysis that we might utilize if the need for it arises.

Please take another look at it.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: Good idea, will add another PR to add this pattern.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: I asked David, and seems like its okay to combine them. Not sure of other use cases, if that's required then we need two additional ports for every gated clock.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: You are right, fixed it.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: As you pointed out, this would require a fixed-point analysis to handle all the possible dataflow operations that can propagate clocks. I think such operations of passing a clock by a wire are very rare, and it might be fine to ignore them for this transformation. This transformation does not guarantee that all clock gates will be removed. I want to do some more experiments with this transformation enabled, and then based on the need, we can utilize a fixed-point analysis.

include/circt/Dialect/HW/HWStructure.td
```diff
@@ -231,6 +231,10 @@ def HWModuleOp : HWModuleOpBase<"module",
 
     /// Verifies the body of the function.
     LogicalResult verifyBody();
+
+    void appendPorts(
+    ArrayRef<std::tuple<StringAttr, Type, Location>> insertInputs,
```
dtzSiFive, line 236: nit: Can we tweak the formatting to not have these parameters on same column as the beginning of the method declaration? Thanks!

include/circt/Dialect/Seq/SeqPasses.td
```diff
@@ -107,4 +107,11 @@ def LowerSeqShiftReg : Pass<"lower-seq-shiftreg", "hw::HWModuleOp"> {
   let dependentDialects = ["circt::hw::HWDialect"];
 }
 
+def SinkClockGates : Pass<"sink-clock-gates", "mlir::ModuleOp"> {
+  let summary = "Sink the gated clock enables into the op that is driven by the clock.";
+  let constructor = "circt::seq::createSinkClockGatesPass()";
+  let dependentDialects = ["circt::hw::HWDialect", "circt::comb::CombDialect"];
```
dtzSiFive, line 113: Might be fun to throw a statistic on here to see how much it's doing, but just a thought.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,344 @@
+//===- SinkClockGates.cpp - Sink clock gates to its users. -----===//
```
dtzSiFive, line 1: nit: pad this out please

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,344 @@
...
+//
+// This file implements a transformation that "tries" to sink as many clock
+// gates as possible. This is a best-effort transformation, no assumptions can
+// be made post this transformation.
+// A clock gate can be sunk through an instance to a module, if all the
+// instances of the module are known and they reference a unique HWModuleOp. If
+// there are instances of a module, that have gated clock while others have
+// un-gated clock, then the un-gated versions add a constant true enable
+// condition. This analysis relies on the fact that the HW instance graph cannot
+// have cycles/recursion. It is guarunteed that any instance operand will be
```
dtzSiFive, line 17:
```suggestion
// have cycles/recursion. It is guaranteed that any instance operand will be
```

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,344 @@
...
+          collapseConsecutiveClockGates(clkGate, userGate);
+        } else {
+          opsWithGatedClock.insert(useOp);
+        }
+      }
+    });
+  };
+  // Seed phase, find the instances to start the traversal from.
+  // These instances have a gated clock as input.
+  getOperation()->walk(
```
dtzSiFive, line 219: Can this use an instance graph or something to find the modules, instead of `walk` through the entire design looking for HWModuleOp's?

dtzSiFive (commented):
Cool! This is neat! :grin: 🎉 

Left some minor feedback.

include/circt/Dialect/Seq/SeqPasses.td
fabianschuiki: That would be really neat!

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
fabianschuiki: Ah right, good point, forgot about that. I was just wondering of how a user might expect the pass to behave in CIRCT -- someone could expect that the enable inputs would be preserved as separate signals. But on the other hand this is also very easy to adjust once the pattern of the pass is established, and this comes up as a concrete need. Combining them feels like a good start!

include/circt/Dialect/HW/HWStructure.td
prithayan: Thanks for the suggestion, fixed.

include/circt/Dialect/Seq/SeqPasses.td
prithayan: Good idea, added a stat.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: Fixed.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: Thanks for catching this, applied it.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: Right, thanks for catching this, moved to iterate over instance graph nodes.

include/circt/Dialect/HW/HWStructure.td
```diff
@@ -231,6 +231,10 @@ def HWModuleOp : HWModuleOpBase<"module",
 
     /// Verifies the body of the function.
     LogicalResult verifyBody();
+
+    void appendPorts(
+                  ArrayRef<std::tuple<StringAttr, Type, Location>> insertInputs,
+                  ArrayRef<std::pair<StringAttr, Value>> insertOutputs);
```
uenoku, line 237: What's difference from `modifyPorts` which we are going to deprecate?

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+  /// If the `argNames` attribute exists, return it.
+  SmallVector<StringAttr> getInstanceArgNames(HWInstanceLike inst) {
+    SmallVector<StringAttr> argNames;
+    if (auto args = inst->getAttrOfType<ArrayAttr>("argNames")) {
+      auto attr = args.getAsRange<StringAttr>();
+      argNames.insert(argNames.begin(), attr.begin(), attr.end());
+    }
+    return argNames;
+  }
```
uenoku, line 155: Could you add `argNames` getter to HWInstanceLike. I guess all operations that implement HWInstanceLike have non-optional `argNames`

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,348 @@
...
+    HWModuleOp instModule = {};
+    auto refModNames = inst.getReferencedModuleNamesAttr();
+    if (refModNames.size() != 1) {
+      inst.emitWarning("expected an instance with a single reference");
+      return {};
+    }
+    auto *node = graph->lookup(cast<StringAttr>(*refModNames.begin()));
+    if (node)
+      instModule = dyn_cast_or_null<HWModuleOp>(*node->getModule());
+    if (instModule.isPublic())
```
uenoku, line 126: Please insert null-guard for instModule. I agree that we shouldn't optimize across public modules but please add a comment and test.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+  void runOnOperation() override;
+
+private:
+  using ClockEnablePortPairs = DenseMap<unsigned, unsigned>;
+  InstanceGraph *graph;
+  DenseSet<Operation *> opsToErase;
+  mlir::MLIRContext *context;
+  DenseMap<HWModuleOp, ClockEnablePortPairs> moduleClockGateEnableMap;
+};
```
uenoku, line 191: Please clean up data structures at the end of runOnOperation

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,348 @@
...
+      Value enable = getEnable(gate);
+      OpBuilder builder(context);
+      builder.setInsertionPoint(compReg);
+      auto compEn = builder.create<CompRegClockEnabledOp>(
+          compReg->getLoc(), compReg.getInput(), clk, enable,
+          compReg.getNameAttr(), compReg.getReset(), compReg.getResetValue(),
+          compReg.getPowerOnValue(), compReg.getInnerSymAttr());
+      compReg->replaceAllUsesWith(compEn);
+      opsToErase.insert(compReg);
+    }
```
uenoku, line 339: I think this should go to canonicalization if it's always beneficial.

uenoku (commented):
What do you think about decoupling canonicalization from this pass? I feel it looks more cleaner if `SinkClockGates` only sinks clock gates and not perform comp reg simplification.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,348 @@
...
+      // Replace the gated clock with the original base clock.
+      inst->setOperand(index, clkGate.getInput());
+      Value enable = getEnable(clkGate);
+      if (clockEnablePortIndices.contains(index)) {
+        // The enable for this clock is already added, (from another instance of
+        // the module).
+        auto oldEnablePort = clockEnablePortIndices[index];
+        auto oldEnableVal = inst->getOperand(oldEnablePort);
+        auto *def = oldEnableVal.getDefiningOp();
+        assert(llvm::isa_and_nonnull<ConstantOp>(def));
```
uenoku, line 275: Why is it guaranteed to be constant?

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
```diff
@@ -0,0 +1,348 @@
...
+// A clock gate can be sunk through an instance to a module, if all the
+// instances of the module are known and they reference a unique HWModuleOp. If
+// there are instances of a module, that have gated clock while others have
+// un-gated clock, then the un-gated versions add a constant true enable
+// condition. This analysis relies on the fact that the HW instance graph cannot
+// have cycles/recursion. It is guaranteed that any instance operand will be
+// updated only once. This assert checks for it:
+// `assert(llvm::isa_and_nonnull<ConstantOp>(def))`
+// This doesnot handle gated clocks returned from an output port yet.
+
```
uenoku, line 21: nit:
```suggestion
// 
//===----------------------------------------------------------------------===//```

File: lib/Dialect/Seq/Transforms/SinkClockGates.cpp [hunk 17]
```diff
@@ -0,0 +1,348 @@
...
+    return enable;
+  }
+
+  void runOnOperation() override;
+
+private:
+  using ClockEnablePortPairs = DenseMap<unsigned, unsigned>;
+  InstanceGraph *graph;
+  DenseSet<Operation *> opsToErase;
+  mlir::MLIRContext *context;
```
uenoku, line 189: nit: I think pass implements a context getter `getContext()`.

include/circt/Dialect/HW/HWStructure.td
prithayan: Good point, thanks for bringing it up, I will add comments here. 
Few reasons,
1. Stop using the deprecated methods.
2. Implement a simpler API, that only appends the input and then output ports, the most-common/only use case. Avoids the generic methods, that can insert/delete at random port indices. 
3. Avoid any performance foot-guns, the deprecated methods can have performance issues, implementing the fresh API that we discuss and properly inspect to ensure there are no obvious performance issues.
4. At least rhodium can stop using the deprecated API, and switch to this, if its approved.
I will commit this separately, so if it makes sense to discuss this more broadly, I will create a new PR with this change.
What do you think ?

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: This is a constant, because this is another instance of a module that is already visited. When the first instance was visited, we added the enable port to all the module instances. The enable port for all the other instances were set to a constant true, except the one that is being visited, which is set to the clock gate op's enable. This ensures that we donot add multiple enable ports for the same clock gate port when visiting multiple instances of a module.

lib/Dialect/Seq/Transforms/SinkClockGates.cpp
prithayan: Yeah, I thought about that, but as you pointed out, I don't think this is "beneficial" in all cases. This should only happen if the user explicitly enables this transformation. So, the transformation of "sink clock-gates" implies trying to combine the clock gate's enable into the operation that is driven by the gated clock. This might not be desirable in general, and if we add it to the canonicalization we can't disable it in general. 
But maybe this can be added to some op interface. A new op interface that any op can opt-in, and then we can invoke it from here.
My plan is to add the enable to FirRegOp next, similar transformation needs to happen to memory ops also, but its semantics are slightly non-trivial.

include/circt/Dialect/HW/HWStructure.td
uenoku: Sorry for delayed replay. TBH I don't find much difference between deprecated functions. The intention for the deprecation was "it doesn't make sense to have an API which takes input and output ports separately" (https://github.com/llvm/circt/blob/5d8cf69cf222ec8776b1680d4d1de7e0e2aa4a86/lib/Dialect/HW/HWOps.cpp#L675-L679) so I'm not sure it's a right direction.

prithayan: Thanks for your suggestions Hideto, I updated the interface, we need the output Values to update the `OutputOp` also added the input values, that can be set as the block arguments are added. This avoids the need to know the argument index of the input args.

PR #7476, opened 2024-08-08
[FIRRTL][InferResets] Generalize FART to support sync reset

jackkoenig:
The FullAsyncResetAnnotation is deprecated and replaced by FullResetAnnotation which includes a resetType argument that must be 'sync' or 'async'.  IgnoreFullAsyncResetAnnotation is deprecated and replaced by ExcludeFromFullResetAnnotation. The new annotations are in package circt.

The behavior of FullResetAnnotation with resetType = 'async' is identical to that of the old FullAsyncResetAnnotation. The behavior of FullResetAnnotation with resetType = 'sync' is very similar, except the type of the reset wired through will be UInt<1>, and any registers with an existing reset at all (both sync or async) will be left unchanged (resetType = 'async' will add async resets to registers with existing sync resets).

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
@@ -1,4 +1,4 @@
-//===- InferResets.cpp - Infer resets and add async reset -------*- C++ -*-===//
+//===- InferResets.cpp - Infer resets and add full reset --------*- C++ -*-===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -270,6 +270,15 @@ using ResetNetwork = llvm::iterator_range<
 /// Whether a reset is sync or async.
 enum class ResetKind { Async, Sync };
 
+static StringRef resetKindToStringRef(const ResetKind &kind) {
+  switch (kind) {
+  case ResetKind::Async:
+    return "async";
+  case ResetKind::Sync:
+    return "sync";
+  }
+  llvm_unreachable("unhandled reset kind");
+}
 } // namespace
 
 namespace llvm {
@@ -306,12 +315,11 @@ static T &operator<<(T &os, const ResetKind &kind) {
 //===----------------------------------------------------------------------===//
 
 namespace {
-/// Infer concrete reset types and insert full async reset.
+/// Infer concrete reset types and insert full reset.
 ///
 /// This pass replaces `reset` types in the IR with a concrete `asyncreset` or
-/// `uint<1>` depending on how the reset is used, and adds async resets to
-/// registers in modules marked with the corresponding
-/// `FullAsyncResetAnnotation`.
+/// `uint<1>` depending on how the reset is used, and adds resets to registers
+/// in modules marked with the corresponding `FullResetAnnotation`.
 ///
 /// On a high level, the first stage of the pass that deals with reset inference
 /// operates as follows:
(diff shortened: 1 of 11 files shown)


lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -1311,40 +1318,79 @@ InferResetsPass::collectAnnos(FModuleOp module) {
...
     }
     return false;
-  });
+  };
+
+  AnnotationSet::removePortAnnotations(module,
+                                       [&](unsigned argNum, Annotation anno) {
+                                         Value arg = module.getArgument(argNum);
+                                         return checkAnnotations(anno, arg);
+                                       });
```
jackkoenig, line 1384: clang-format did me dirty here, unclear why this is tabbed so weird whereas the very similar `AnnotationSet::removeAnnotations(op, [&](Annotation anno) {` below is tabbed more reasonably 🤷‍♂️

fabianschuiki (approved):
LGTM!

The new annotations and deprecation of the old ones looks good. The changes to the existing InferResets pass also look solid. Thanks for adding all the additional tests to ensure this works!

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
```diff
@@ -435,6 +435,45 @@ static LogicalResult applyOutputDirAnno(const AnnoPathValue &target,
...
+      << "'" << ignoreFullAsyncResetAnnoClass << "' is deprecated, use '"
+      << excludeFromFullResetAnnoClass << "' instead";
+
+  NamedAttrList newAnno(anno.getValue());
+  newAnno.set("class", StringAttr::get(context, excludeFromFullResetAnnoClass));
+
+  DictionaryAttr newDictionary = DictionaryAttr::get(op->getContext(), newAnno);
+
+  return applyWithoutTarget<true, FModuleOp>(target, newDictionary, state);
+}
```
dobios, line 475: You can probably avoid this much code reuse with a templated fucntion or simply one that takes the specifics as parameters and then does the annotation update separately.

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
dobios: The unlucky situation where you ever-so-slightly surpass 80 characters ://

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -1311,40 +1318,79 @@ InferResetsPass::collectAnnos(FModuleOp module) {
...
+        if (!isa<UIntType>(arg.getType()) ||
+            cast<UIntType>(arg.getType()).getWidth() != 1) {
+          badType = true;
+        }
+      }
+      if (badType) {
+        auto kind = resetKindToStringRef(expectedResetKind);
+        mlir::emitError(arg.getLoc(),
+                        "'FullResetAnnotation' with resetType == '")
+            << kind << "' must target " << kind << " reset, but targets "
```
dobios, line 1359: You can probably simplify this a bunch:
```suggestion
      // Check that the type is well-formed
      bool isAsync = expectedResetKind == ResetKind::Async;
      bool validUint = false;
      if (auto uintT = dyn_cast<UIntType>(arg.getType()))
        validUint = uintT.getWidth() == 1;

      if ((isAsync && !isa<AsyncResetType>(arg.getType())) ||
          (!isAsync && !validUint)) {
        auto kind = resetKindToStringRef(expectedResetKind);
        mlir::emitError(arg.getLoc(),
                        "'FullResetAnnotation' with resetType == '")
            << kind << "' must target " << kind << " reset, but targets "
```

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
jackkoenig: I'm not sure my C++ is quite at the level of templating a function sufficiently to deal with it needing to call differently templated functions (see 455 and 475), maybe that could be handled by accepting a thunk. It'd be easy for me in Scala 😂.

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
jackkoenig: Thanks!

dobios (approved):

PR #7477, opened 2024-08-08
[Moore] Add more AssignedVariableOp canonicalizations

fabianschuiki:
Change the `AssignedVariableOp` to directly return type `T` instead of the `ref<T>`. This removes the implied allocation and assignment, and makes this op behave essentially like `hw.wire`.

The canonicalizers for `VariableOp` and `NetOp` can now replace all reads from the old variable or net with the value of `AssignedVariableOp` directly, since there is no more `ref<T>` type involved. This also allows us to fully eliminate unnamed variables and nets that have a unique continuous assignment.

Also add canonicalizers that remove `AssignedVariableOp` if they shadow an input or output port of the same name, or if there are multiple such variables with the same name in a chain.

At a later stage we may want to replace `AssignedVariableOp` entirely with `dbg.variable`.

### DIFF
--- a/lib/Dialect/Moore/MooreOps.cpp
+++ b/lib/Dialect/Moore/MooreOps.cpp
@@ -289,25 +289,49 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
 }
 
 LogicalResult VariableOp::canonicalize(VariableOp op,
-                                       ::mlir::PatternRewriter &rewriter) {
-  Value initial;
-  for (auto *user : op->getUsers())
-    if (isa<ContinuousAssignOp>(user) &&
-        (user->getOperand(0) == op.getResult())) {
-      // Don't canonicalize the multiple continuous assignment to the same
-      // variable.
-      if (initial)
+                                       PatternRewriter &rewriter) {
+  // Check if the variable has one unique continuous assignment to it, all other
+  // uses are reads, and that all uses are in the same block as the variable
+  // itself.
+  auto *block = op->getBlock();
+  ContinuousAssignOp uniqueAssignOp;
+  for (auto *user : op->getUsers()) {
+    // Ensure that all users of the variable are in the same block.
+    if (user->getBlock() != block)
+      return failure();
+
+    // Ensure there is at most one unique continuous assignment to the variable.
+    if (auto assignOp = dyn_cast<ContinuousAssignOp>(user)) {
+      if (uniqueAssignOp)
         return failure();
-      initial = user->getOperand(1);
+      uniqueAssignOp = assignOp;
+      continue;
     }
 
-  if (initial) {
-    rewriter.replaceOpWithNewOp<AssignedVarOp>(op, op.getType(),
-                                               op.getNameAttr(), initial);
-    return success();
+    // Ensure all other users are reads.
+    if (!isa<ReadOp>(user))
+      return failure();
+  }
+  if (!uniqueAssignOp)
+    return failure();
+
+  // If the original variable had a name, create an `AssignedVariableOp` as a
+  // replacement. Otherwise substitute the assigned value directly.
+  Value assignedValue = uniqueAssignOp.getSrc();
+  if (auto name = op.getNameAttr(); name && !name.empty())
+    assignedValue = rewriter.create<AssignedVariableOp>(
+        op.getLoc(), name, uniqueAssignOp.getSrc());
+
+  // Remove the assign op and replace all reads with the new assigned var op.
+  rewriter.eraseOp(uniqueAssignOp);
+  for (auto *user : llvm::make_early_inc_range(op->getUsers())) {
+    auto readOp = cast<ReadOp>(user);
+    rewriter.replaceOp(readOp, assignedValue);
   }
 
-  return failure();
+  // Remove the original variable.
+  rewriter.eraseOp(op);
+  return success();
 }
 
 SmallVector<DestructurableMemorySlot> VariableOp::getDestructurableSlots() {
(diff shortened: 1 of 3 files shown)


lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -371,15 +395,116 @@ void NetOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
...
+  if (uniqueAssignOp && !op.getAssignment()) {
+    rewriter.modifyOpInPlace(
+        op, [&] { op.getAssignmentMutable().assign(uniqueAssignOp.getSrc()); });
+    rewriter.eraseOp(uniqueAssignOp);
+    modified = true;
+  }
+
+  // If all users of the net op are reads (besides the unique assignment),
+  // directly replace the reads with the net's assigned value.
+  if (allUsesAreReads && op.getAssignment()) {
```
maerhart, line 435: We also need to check here that `uniqueAssignOp` is not NULL

hailongSun2000, line 435: For example:
```
 wire w1 = 1'b0;
 assign w1 = 1'b1;
 wire w2;
 assign w2 = 1'b1;
 // read w1 and w2
```
`w2` will trigger these two `if` branches, but `w1` only triggers the latest `if` branch.
If we tweak these the same as the canonicalize of `variable`, will something go wrong?

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -371,15 +395,116 @@ void NetOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
...
+    auto moduleOp = dyn_cast<SVModuleOp>(outputOp.getParentOp());
+    if (!moduleOp)
+      break;
+    auto moduleType = moduleOp.getModuleType();
+    auto portName = moduleType.getOutputNameAttr(use.getOperandNumber());
+    if (portName == op.getNameAttr()) {
+      rewriter.replaceOp(op, op.getInput());
+      return success();
+    }
+  }
```
hailongSun2000, line 503: I have a question: The type of `use` is OpOperand, in your example `%b = moore.assigned_variable %3 : i42`, so I always think `use` == `%3`, and why `use.getOwner()` will return an output op :thinking:? Similarly, I view the annotation regarding `OpOperand.getOperandNumber()`, because `assigned_variable` only has one operand(`input`) , so `use.getOperandNumber()` must get zero, if we tweak the order of `moore.output %b, %w : !moore.i42, !moore.i42` like `moore.output %w, %b : !moore.i42, !moore.i42`. Is this canonicalization also triggered?

lib/Dialect/Moore/MooreOps.cpp
fabianschuiki: Good point! Thanks

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -289,25 +289,49 @@ VariableOp::handlePromotionComplete(const MemorySlot &slot, Value defaultValue,
...
+  if (!uniqueAssignOp)
+    return failure();
 
-  return failure();
+  // If the original variable had a name, create an `AssignedVariableOp` as a
+  // replacement. Otherwise substitute the assigned value directly.
+  Value assignedValue = uniqueAssignOp.getSrc();
+  if (auto name = op.getNameAttr(); name && !name.empty())
+    assignedValue = rewriter.create<AssignedVariableOp>(
+        op.getLoc(), name, uniqueAssignOp.getSrc());
```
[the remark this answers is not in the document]
fabianschuiki: Unfortunately there are cases where `getNameAttr()` returns a valid string attribute, but the string attribute contains an empty string. It's a bit unfortunate that we have to do this double checking, but it's a bit of a pattern in other parts of CIRCT 😢

lib/Dialect/Moore/MooreOps.cpp
fabianschuiki: Yeah very good point. The second if, which replaces the net with an `AssignedVariableOp`, should only be triggered if the wire has one unique assignment -- which is not the case for `w1`. I have fixed this.

lib/Dialect/Moore/MooreOps.cpp
fabianschuiki: You're right that `use` is a `OpOperand`, but importantly the loop goes through all `OpOperand`s of all the other ops that use `op`. In the example of `%b = moore.assigned_variable %3 : i42`, `getUses()` does not return the `%3`, but all places in other ops where a `%b` is used. If there's an op like `moore.output %b, %w` around, calling `getUses()` on `%b` will produce the `OutputOp`'s use of `%b`. If you had `moore.output %b, %b`, then `getUses()` would produce two `OpOperand`s, both with `OutputOp` as the owner, but one with operand number 0 and one with operand number 1.

Basically, `OpOperand` is the slot in an operation where a value can be plugged into. Asking for `getUses()` on an op returns you a list of all these slots where the op (or value if you call it on a value) is used.

fabianschuiki:
Applies your suggestions @maerhart @hailongSun2000.

test/Dialect/Moore/canonicalizers.mlir
```diff
@@ -10,32 +10,168 @@ func.func @Casts(%arg0: !moore.i1) -> (!moore.i1, !moore.i1) {
...
+  dbg.variable "b", %4 : !moore.i42
+
+  // Unique continuous assignments to nets should remove the `ref<T>`
+  // indirection and instead directly propagate the assigned value to readers.
+  // CHECK-NOT: moore.assign
+  // CHECK-NOT: moore.net wire
+  // CHECK: %c = moore.assigned_variable %w : i42
+  // CHECK: dbg.variable "c", %c : !moore.i42
+  moore.assign %c, %w : i42
+  %c = moore.net wire : <i42>
```
maerhart, line 45: Sorry, but a couple of dumb SV questions:

If we can replace a `net wire` with an `assigned_variable` can we also replace it with just a `variable`? Is this only allowed for uniquely driven wires? Otherwise, why are we distinguishing between wire and variable?

Also, for the cases where we have an initial value given to the variable or net operation, wouldn't that be noticeable in a simulation trace because at `t=0s` it would be the initial value and at time `t=0s,0d,1e`  it would be the driven value? Do we only care about the real time steps?

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -371,15 +395,118 @@ void NetOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
...
+  ContinuousAssignOp uniqueAssignOp;
+  bool allUsesAreReads = true;
+  for (auto *user : op->getUsers()) {
+    // Ensure that all users of the net are in the same block.
+    if (user->getBlock() != block)
+      return failure();
+
+    // Ensure there is at most one unique continuous assignment to the net.
+    if (auto assignOp = dyn_cast<ContinuousAssignOp>(user)) {
+      if (uniqueAssignOp || assignOp.getDst() != op)
```
maerhart, line 413: Isn't this overly conservative? It cannot be another operand because the type wouldn't match, right?
(same for VariableOp)
```suggestion
      if (uniqueAssignOp)
```

maerhart (approved):
Looks good to me!

lib/Dialect/Moore/MooreOps.cpp
fabianschuiki: Haha yeah that is actually true. Let me drop that.

test/Dialect/Moore/canonicalizers.mlir
fabianschuiki: I think in case there is a single unique driver to the net, the net is actually equivalent to the variable. If I recall correctly, the differences start when you have multiple drivers to the same net or variable, where a net will potentially have different driving conflict resolution compared to the variable. My guess is that we'll be able to make the conflict resolution explicit in an operation at some point, and then have optimization passes at the Moore level where we replace wires with simpler variables, or plain values directly, whenever the fancy features of nets aren't used. One example would be inout ports, or the nets we conjure up to model instance outputs driving values. A lot of these should turn out to be simplifiable to variables or even to plain values.

Good point about the assignments and delays. IIRC the SV spec allows for continuous assignments to be interpreted as aliases, such that they don't introduce a delta time delay. Which in this case would mean that there wouldn't be any difference observable at `t=0s`. I remember that some commercial tools make heavy use of this to reduce the number of iterations on events 🤔. I'll have to do some more digging in the SV spec on what the _exact_ semantics are, but I think most EDA tools apply optimizations that remove delta time steps.

fabianschuiki: Found the following in IEEE 1800-2017 "4.9.1 Continuous assignment":
> A continuous assignment process is also evaluated at time zero in order to propagate constant values. This includes implicit continuous assignments inferred from port connections (see 3.9.6).

lib/Dialect/Moore/MooreOps.cpp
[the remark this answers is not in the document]
hailongSun2000: Haha, I got it! Interesting empty string.

lib/Dialect/Moore/MooreOps.cpp
hailongSun2000: Thanks for the detailed explanation :smiley: !  Looks like I misunderstand `getUses()`.  But now I know.

PR #7478, opened 2024-08-08
[Moore] Add constant materialization, fold constant conversions

fabianschuiki:
Implement the `materializeConstant` function for the Moore dialect and mark the `ConstantOp` as a `ConstantLike` operation. This now allows us to write constant folders for various operations. As a first example, add a constant folder for `ConversionOp` that directly applies domain conversions (e.g., `i42` to `l42` or vice versa) to constants.

### DIFF
--- a/include/circt/Dialect/Moore/MooreDialect.td
+++ b/include/circt/Dialect/Moore/MooreDialect.td
@@ -38,6 +38,7 @@ def MooreDialect : Dialect {
   }];
   let useDefaultAttributePrinterParser = 1;
   let useDefaultTypePrinterParser = 0;
+  let hasConstantMaterializer = 1;
   let dependentDialects = ["hw::HWDialect"];
 }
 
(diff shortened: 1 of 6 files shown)


maerhart (approved):
Isn't it required to implement a fold function for operations that have the `ConstantLike` trait that just returns the attribute?

Also, ConstantStringOp already has ConstantLike added, so maybe also support that in the constant materializer?

fabianschuiki:
Oh good point, I forgot about the string constant op. Will add a fold method there. Apparently it's not technically required, or at least nothing complained so far about the string constant being `ConstantLike` but having no folder. Let me fix that.

fabianschuiki:
Ugh just realized that the `moore.string_constant` op returns an `IntType`, so the folder and constant materializer will have to convert the string constants to an `FVInt` and back. That's probably something for a separate PR.

maerhart:
> Ugh just realized that the `moore.string_constant` op returns an `IntType`, so the folder and constant materializer will have to convert the string constants to an `FVInt` and back. That's probably something for a separate PR.

Oh, yeah that sounds annoying.

hailongSun2000 (approved):
Awesome！

PR #7481, opened 2024-08-08
[MooreToCore] Fix variable op lowering of aggregate types

maerhart:


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -158,17 +158,26 @@ struct VariableOpConversion : public OpConversionPattern<VariableOp> {
   LogicalResult
   matchAndRewrite(VariableOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
     Type resultType = typeConverter->convertType(op.getResult().getType());
     Value init = adaptor.getInitial();
     // TODO: Unsupport x/z, so the initial value is 0.
     if (!init && cast<RefType>(op.getResult().getType()).getDomain() ==
                      Domain::FourValued)
       return failure();
```
fabianschuiki, line 167: I'm wondering whether we should remove this check and generally lower everything to an `i1` for now by mapping X and Z bits to 0. Verilator does a similar thing. Later, once we have proper support for four-valued types in HW/Comb/Seq, we could come back and adjust the lowering again.

fabianschuiki (approved):
Very nice!

lib/Conversion/MooreToCore/MooreToCore.cpp
cepheus69: I support to choose X / Z pessimistic strategy. Once we support multi-valued types in core IR. A new option could be added to enable the optimistic strategy - spread x / z value and preserve their value.

maerhart: Adding a pass option to let the user decide whether x/z should be converted to 0 sounds like a reasonable approach to me. I'll leave it for a follow-up PR.

PR #7486, opened 2024-08-09
[FIRRTL] Add list concatenation operation.

mikeurbach:
This operation produces a list by concatenating lists of the same
type. The operation definition and a simple round trip test have been
added. This is used to compose lists, which is a key feature to enable
hierarchical composition of Properties.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
@@ -1154,6 +1154,23 @@ def ListCreateOp : FIRRTLOp<"list.create", [Pure, SameTypeOperands]> {
   let hasVerifier = 1;
 }
 
+def ListConcatOp : FIRRTLOp<"list.concat", [Pure, SameOperandsAndResultType]> {
+  let summary = "Concatenate multiple lists to produce a new list";
+  let description = [{
+    Produces a value of list type by concatenating the provided lists.
+
+    Example:
+    ```mlir
+    %3 = firrtl.list.concat %0, %1, %2 : !firrtl.list<string>
+    ```
+  }];
+
+  let arguments = (ins Variadic<ListType>:$subLists);
+  let results = (outs ListType:$result);
+
+  let assemblyFormat = "$subLists attr-dict `:` type($result)";
+}
+
 def BoolConstantOp : FIRRTLOp<"bool", [Pure, ConstantLike]> {
   let summary = "Produce a constant boolean value";
   let description = [{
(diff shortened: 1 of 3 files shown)


prithayan (approved):

include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
```diff
@@ -1154,6 +1154,23 @@ def ListCreateOp : FIRRTLOp<"list.create", [Pure, SameTypeOperands]> {
   let hasVerifier = 1;
 }
 
+def ListConcatOp : FIRRTLOp<"list.concat", [Pure, SameOperandsAndResultType]> {
```
dtzSiFive, line 1157: Add to FIRRTLVisitor.h please.

include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
```diff
@@ -1154,6 +1154,23 @@ def ListCreateOp : FIRRTLOp<"list.create", [Pure, SameTypeOperands]> {
...
+    Example:
+    ```mlir
+    %3 = firrtl.list.concat %0, %1, %2 : !firrtl.list<string>
+    ```
+  }];
+
+  let arguments = (ins Variadic<ListType>:$subLists);
+  let results = (outs ListType:$result);
+
+  let assemblyFormat = "$subLists attr-dict `:` type($result)";
```
dtzSiFive, line 1171: Might be neat to give this a folder in the future for IR simplification (and use w/IMCP) but that can come later.

dtzSiFive (approved):

include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
mikeurbach: Oops, yep will do, thanks!

include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
mikeurbach: Could be interesting! I think there are lot of interesting optimizations we could explore on properties.

dtzSiFive: Yeah just thinking about cleaning up the trivial stuff that literally constant-folds.  I'm not sure we're likely to feed this constants at first/soon and yep lots of interesting things for sure! Look forward to it, LMK if want some assistance!

Looks like integer ops have issues filed for adding folders there, but quick check on some internal usage suggests these wouldn't help much just yet anyway.

Anyway stray thought, GLHF!

leonardt (approved):

PR #7489, opened 2024-08-09
[ESI][Manifest] Embed constants in manifest

teqdruid:
Users want to know what constants were used in a design. Plumb them through using the manifest. No runtime support, no pycde support.

Cleanups to the manifest as well. Update the runtime to support the schema changes to the manifest.

### DIFF
--- a/include/circt/Dialect/ESI/ESIManifest.td
+++ b/include/circt/Dialect/ESI/ESIManifest.td
@@ -169,6 +169,23 @@ def AppIDHierNodeOp : ESI_Op<"manifest.hier_node", [
   }];
 }
 
+def SymbolConstantsOp : ESI_Op<"manifest.consts", [
+  DeclareOpInterfaceMethods<IsManifestData>]> {
+  let summary = "Constant values associated with a symbol";
+
+  let arguments = (ins FlatSymbolRefAttr:$symbolRef,
+                       DictionaryAttr:$constants);
+  let assemblyFormat = [{
+    $symbolRef $constants attr-dict
+  }];
+
+  let extraClassDeclaration = [{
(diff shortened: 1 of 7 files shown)


lib/Dialect/ESI/Passes/ESIBuildManifest.cpp
```diff
@@ -165,15 +174,34 @@ std::string ESIBuildManifestPass::json() {
...
+    DenseMap<SymbolRefAttr, SmallVector<IsManifestData>> symbolInfoLookup;
+    for (auto symInfo : mod.getBody()->getOps<IsManifestData>()) {
+      FlatSymbolRefAttr sym = symInfo.getSymbolRefAttr();
+      if (!sym || !symbols.contains(sym))
         continue;
+      symbolInfoLookup[sym].push_back(symInfo);
+    }
+
+    // Now, emit a JSON object for each symbol.
+    for (const auto &symNameInfo : symbolInfoLookup) {
```
mortbopet, line 187: Ideally, some sort of human-readable documentation would also have been modified with this change to the manifest builder (in other words, where is the manifest spec?)

include/circt/Dialect/ESI/ESIManifest.td
```diff
@@ -169,6 +169,23 @@ def AppIDHierNodeOp : ESI_Op<"manifest.hier_node", [
   }];
 }
 
+def SymbolConstantsOp : ESI_Op<"manifest.consts", [
```
mortbopet, line 172: I've been bashed for excessive succinctness before, so i'm here going to push back on calling this `consts` instead of `constants` :).

mortbopet (approved):

PR #7491, opened 2024-08-09
[CreateSifiveMetadata] Update memory hierarchy paths to be pre-extrction

prithayan:
The `CreateSifiveMetadata` pass was creating hierarchy paths to the memory module pre-extraction.
The `om.path` was being updated as the memory was extracted in the following passes.
But recently we realized this was a bug, as the downstream tools consuming the metadata files, were actually expecting the pre-extraction hierarchy paths to the memory.
So, this PR fixes the path to terminate at the parent of the module that instantiates the memory and encodes the pre-extraction memory instance as a string.
The tool that will parse the final `mlir` must now construct the actual pre-extraction path from the two lists, by appending the pre-extraction instance name to the path.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
@@ -121,7 +121,7 @@ struct ObjectModelIR {
         buildSimpleClassOp(builderOM, unknownLoc, "ExtraPortsMemorySchema",
                            extraPortFields, extraPortsType);
 
-    mlir::Type classFieldTypes[12] = {
+    mlir::Type classFieldTypes[13] = {
         StringType::get(context),
         FIntegerType::get(context),
         FIntegerType::get(context),
@@ -133,9 +133,11 @@ struct ObjectModelIR {
         FIntegerType::get(context),
         ListType::get(context, cast<PropertyType>(PathType::get(context))),
         BoolType::get(context),
-        ListType::get(context,
-                      cast<PropertyType>(detail::getInstanceTypeForClassLike(
-                          extraPortsClass)))};
+        ListType::get(
+            context, cast<PropertyType>(
+                         detail::getInstanceTypeForClassLike(extraPortsClass))),
+        ListType::get(context, cast<PropertyType>(StringType::get(context))),
+    };
 
     memorySchemaClass =
         buildSimpleClassOp(builderOM, unknownLoc, "MemorySchema",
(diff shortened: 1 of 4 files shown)


lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -246,8 +248,13 @@ struct ObjectModelIR {
...
     auto memPaths = instancePathCache.getAbsolutePaths(mem);
     SmallVector<Value> memoryHierPaths;
+    SmallVector<Value> finalInstanceNames;
     for (auto memPath : memPaths) {
-      Operation *finalInst = memPath.leaf();
+      auto finalInst = memPath.leaf();
+      finalInstanceNames.emplace_back(
+          builderOM.create<StringConstantOp>(finalInst.getInstanceNameAttr()));
+      memPath.dropBack();
+      finalInst = memPath.leaf();
```
mikeurbach, line 257: Is this intending to update `finalInst` to the leaf after dropping the last element? I was under the impression `dropBack()` was not side-effecting, so maybe it would need to be `finalPath = memPath.dropBack().leaf()`? Or maybe I misunderstood.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -246,8 +248,13 @@ struct ObjectModelIR {
 
     auto memPaths = instancePathCache.getAbsolutePaths(mem);
     SmallVector<Value> memoryHierPaths;
+    SmallVector<Value> finalInstanceNames;
     for (auto memPath : memPaths) {
-      Operation *finalInst = memPath.leaf();
+      auto finalInst = memPath.leaf();
```
mikeurbach, line 253: nit, any reason to change this line? It might be a little smaller diff to keep the original `Operation *finalInst = ...`, and I find the type being spelled out helps personally.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
prithayan: You are right, fixed in latest version.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
prithayan: Need to get the instance name, updated the `Operation*` to `InstanceOpInterface`

mikeurbach: Ah I see, thanks.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -246,32 +246,52 @@ struct ObjectModelIR {
...
+    if (inDut) {
+      for (auto memPath : memPaths) {
+        {
+          igraph::InstanceOpInterface finalInst = memPath.leaf();
+          finalInstanceNames.emplace_back(builderOM.create<StringConstantOp>(
+              finalInst.getInstanceNameAttr()));
+        }
+        SmallVector<Attribute> namepath;
+        bool foundDut = dutMod == nullptr;
+        // The hiarpath will be created to the pre-extracted
```
mikeurbach, line 260: nit: hierpath.

mikeurbach (approved):
Looks good to me now, thanks!

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
prithayan: Thanks, fixed.

PR #7492, opened 2024-08-09
[ESI][Runtime] Parse and expose manifest constants

teqdruid:
Access per-module constants from the ModuleInfo class.


integration_test/Dialect/ESI/runtime/loopback.mlir
```diff
@@ -95,6 +96,7 @@ hw.module @CallableFunc1() {
 }
 
 esi.manifest.sym @Loopback name "LoopbackIP" version "v0.0" summary "IP which simply echos bytes" {foo=1}
+esi.manifest.consts @Loopback {depth=5:ui32}
```
mortbopet, line 99: I've been bashed for excessive succinctness before, so i'm here going to push back on calling this `consts` instead of `constants` :).

mortbopet (approved):

PR #7494, opened 2024-08-09
[Moore] Power operator folders and canonicalizers

maerhart:


lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -1101,6 +1101,78 @@ ReadOp::removeBlockingUses(const MemorySlot &slot,
...
 
+//===----------------------------------------------------------------------===//
+// PowSOp
+//===----------------------------------------------------------------------===//
+
+static OpFoldResult powCommonFolding(MLIRContext *ctxt, Attribute lhs,
+                                     Attribute rhs) {
+  auto lhsValue = dyn_cast_or_null<FVIntegerAttr>(lhs);
+  if (lhsValue && !lhsValue.getValue().hasUnknown() &&
+      lhsValue.getValue().toAPInt(false) == 1)
```
fabianschuiki, line 1112: I think `lhs.getValue() == 1` will do the same thing. FVInt has the `==` operator overloaded, which will check if the FVInt has no unknowns and whether the value matches.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -1101,6 +1101,78 @@ ReadOp::removeBlockingUses(const MemorySlot &slot,
...
+OpFoldResult PowSOp::fold(FoldAdaptor adaptor) {
+  return powCommonFolding(getContext(), adaptor.getLhs(), adaptor.getRhs());
+}
+
+LogicalResult PowSOp::canonicalize(PowSOp op, PatternRewriter &rewriter) {
+  Location loc = op.getLoc();
+  auto intType = cast<IntType>(op.getRhs().getType());
+  if (auto baseOp = op.getLhs().getDefiningOp<ConstantOp>()) {
+    if (!baseOp.getValue().hasUnknown() &&
+        baseOp.getValue().toAPInt(false) == 2) {
```
fabianschuiki, line 1133: `baseOp.getValue() == 2` should work

fabianschuiki (approved):
Very cool 😎

PR #7499, opened 2024-08-10
[MooreToCore] More extract op lowerings

maerhart:
We need to inset runtime checks for out-of-bound accesses because the semantics of the core dialects don't match the ones of SystemVerilog. I left this for a future PR since there are a few interesting cases to consider and it would increase the size of this PR considerably.

### DIFF
--- a/include/circt/Dialect/LLHD/IR/ExtractOps.td
+++ b/include/circt/Dialect/LLHD/IR/ExtractOps.td
@@ -18,8 +18,8 @@ class SigPtrIndexBitWidthConstraint<string index, string input>
   : TypesMatchWith<"Index width should be exactly clog2 (size of array)",
                    input, index,
                    [{
-                     IntegerType::get($_self.getContext(), std::max<uint64_t>(
-                       llvm::Log2_64_Ceil(llhd::getLLHDTypeWidth($_self)), 1))
+                      IntegerType::get($_self.getContext(), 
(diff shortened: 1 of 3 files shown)


hailongSun2000 (approved):
It makes sense! We ever discussed **bits slice**. @fabianschuiki thinks we can refactor `moore.extract` to `extract`, `array_get`, and `array_slice`. But I think `hw.array_get` cannot handle the unpacked array type, so maybe we can keep `moore.extract` to extract some bits from the vector, N-dimension packed/unpacked array at the moore level. It's also a great option that distinguishes them at the core level. LGTM 🥳 !

hailongSun2000:
I viewed the IEEE 1800-2017 § 7.4.6 "Indexing and slicing of arrays", if we access the index(out of bound), we can return a value based on Table 7-1.

maerhart:
Yes, it might be worth refactoring the extract operations a bit. I'm also wondering if it's worth having a static extract operation for arrays if we already have the dynamic one and could just pass a constant value. I also agree that using the same operation for `array_get` and `array_slice` is strange.

Thanks for looking it up in the standard! The problem we have here is that SV defines exactly that we need to return 0 or X, but CIRCT Core allows any value within the type constraint. So we need to insert a `comb.mux`that returns 0/X when out of bounds.

hailongSun2000:
I think that's worth it. It's a great idea to insert a `comb.mux`.

fabianschuiki:
Very nice :partying_face:!

PR #7501, opened 2024-08-10
[MooreToCore] Support ConditionalOp

fabianschuiki:
Add a lowering pattern from `moore.conditional` to `scf.if`. This currently relies on the condition being a two-valued integer after lowering. Once we support four-valued integers at the core dialect level, the lowering of `moore.conditional` will become a lot more complicated.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -17,6 +17,7 @@
 #include "circt/Dialect/Moore/MooreOps.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/BuiltinDialect.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Pass/Pass.h"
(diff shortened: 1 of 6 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -960,6 +961,39 @@ struct AssignOpConversion : public OpConversionPattern<OpTy> {
...
+    // evaluated and merged with the appropriate lookup table. See documentation
+    // for `ConditionalOp`.
+    auto type = typeConverter->convertType(op.getType());
+    auto ifOp =
+        rewriter.create<scf::IfOp>(op.getLoc(), type, adaptor.getCondition());
+    rewriter.inlineRegionBefore(op.getTrueRegion(), ifOp.getThenRegion(),
+                                ifOp.getThenRegion().end());
+    rewriter.inlineRegionBefore(op.getFalseRegion(), ifOp.getElseRegion(),
+                                ifOp.getElseRegion().end());
+    rewriter.eraseOp(op);
```
maerhart, line 981:
```suggestion
    rewriter.replaceOp(op, ifOp);
```
Maybe also add a use-site in the test-case to trigger this issue?

maerhart (approved):
This is really nice for a first version to get the compilation flow going! I agree with you that we don't want to use `scf.if` inside HW modules in the long run.

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Oops, good catch :see_no_evil: Thanks!

hailongSun2000:
Awesome!

PR #7502, opened 2024-08-10
[MooreToCore] Support four-valued VariableOp without init

fabianschuiki:
Initialize `VariableOp`s of a four-valued type with a zero. This is in line with the rest of `MooreToCore` which maps all X/Z to zero at the moment, either implicitly by mapping to `comb.*` ops, or explicitly by conjuring up zero constants.


maerhart (approved):
LGTM!

hailongSun2000:
Cool!

PR #7503, opened 2024-08-10
[MooreToCore] Support CaseZEq and CaseXZEq ops

fabianschuiki:
Add a conversion for `moore.casez_eq` and `moore.casexz_eq` operations. These only really make sense if the operands are four-valued integers, since the X and Z bits indicate which bits to ignore during the equality check. We don't have support for four-valued integers in the core dialects yet. However, the vast majority of uses of this op are comparing an SSA value against a `moore.constant`. This case we can handle properly by looking at the constant, identifying the unknown bits, and then masking them before performing a regular two-valued comparison between the two operands.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -718,13 +718,57 @@ struct ICmpOpConversion : public OpConversionPattern<SourceOp> {
   using OpAdaptor = typename SourceOp::Adaptor;
 
   LogicalResult
-  matchAndRewrite(SourceOp op, OpAdaptor adapter,
+  matchAndRewrite(SourceOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     Type resultType =
         ConversionPattern::typeConverter->convertType(op.getResult().getType());
 
     rewriter.replaceOpWithNewOp<comb::ICmpOp>(
-        op, resultType, pred, adapter.getLhs(), adapter.getRhs());
+        op, resultType, pred, adaptor.getLhs(), adaptor.getRhs());
+    return success();
+  }
+};
+
+template <typename SourceOp, bool withoutX>
+struct CaseXZEqOpConversion : public OpConversionPattern<SourceOp> {
+  using OpConversionPattern<SourceOp>::OpConversionPattern;
+  using OpAdaptor = typename SourceOp::Adaptor;
+
+  LogicalResult
+  matchAndRewrite(SourceOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Check each operand if it is a known constant and extract the X and/or Z
+    // bits to be ignored.
+    // TODO: Once the core dialects support four-valued integers, we will have
+    // to create ops that extract X and Z bits from the operands, since we also
+    // have to do the right casez/casex comparison on non-constant inputs.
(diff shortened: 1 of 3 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -718,13 +718,58 @@ struct ICmpOpConversion : public OpConversionPattern<SourceOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    // Check each operand if it is a known constant and extract the X and/or Z
+    // bits to be ignored.
+    // TODO: Once the core dialects support four-valued integers, we will have
+    // to create ops that extract X and Z bits from the operands, since we also
+    // have to do the right casez/casex comparison on non-constant inputs.
+    unsigned bitWidth = op.getLhs().getType().getWidth();
+    auto ignoredBits = APInt::getZero(bitWidth);
+    auto detectIgnoredBits = [&](Value value) {
+      auto constOp = value.template getDefiningOp<ConstantOp>();
```
maerhart, line 748: Isn't the `template` unnecessary?

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -718,13 +718,58 @@ struct ICmpOpConversion : public OpConversionPattern<SourceOp> {
...
+    Value rhs = adaptor.getRhs();
+    if (!ignoredBits.isZero()) {
+      ignoredBits.flipAllBits();
+      auto maskOp = rewriter.create<hw::ConstantOp>(op.getLoc(), ignoredBits);
+      lhs = rewriter.createOrFold<comb::AndOp>(op.getLoc(), lhs, maskOp);
+      rhs = rewriter.createOrFold<comb::AndOp>(op.getLoc(), rhs, maskOp);
+    }
+
+    rewriter.replaceOpWithNewOp<comb::ICmpOp>(op, ICmpPredicate::ceq, lhs, rhs);
+    // rewriter.replaceOpWithNewOp<comb::SubOp>(op, zero, adaptor.getInput());
```
maerhart, line 772: Remove this line?

maerhart (approved):
Quite hacky, but the best we can do for now.

Extracting the non-Z ranges and having multiple comparisons that are then ANDed would be more efficient in HW, but I like your more simulation focussed approach 😜

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Oops, thanks!

fabianschuiki:
> Extracting the non-Z ranges and having multiple comparisons that are then ANDed would be more efficient in HW, but I like your more simulation focussed approach 😜

Haha yeah, very true :grin:. Some form of known-bits analysis and using its results to canonicalize comparisons could be a very neat thing. I think @dtzSiFive was talking about some of that a while ago :heart_eyes:

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: It is! I don't really grok the rules about when you need a `template` in expressions embedded within other templates, and when you don't :grimacing:

maerhart: Interesting. I should lookup the rules about this at some point to understand it properly. If those exit.

Issue #6524, opened 2023-12-14
[Question] A way to getOrAddInnerSym from external programs?


SpriteOvO:
Last week, I implemented *probe* related stuff in Chisel binder (chipsalliance/chisel#3668). In CIRCT, `firrtl.ref.rwprobe` needs a `target` attribute with `InnerRefAttr` type. So I filled `inner_sym` attribute for all ops that have it. It worked fine then.

In recent days, I've noticed that CIRCT complains an error if an op's `inner_sym` is filled and not referenced later.

```
error: unable to lower due to symbol "_WIRE" with target not preserved by lowering
```

firtool doesn't have this problem, because firtool fills the `inner_sym` only when it needs to refer to an op result, via function [`getOrAddInnerSym`](https://github.com/llvm/circt/blob/96ddd2fed1a344979b8ffc3a1ca9343f2b5f7cb6/include/circt/Dialect/FIRRTL/FIRRTLUtils.h#L127-L146) (called from [`FIRStmtParser::parseRWProbe`](https://github.com/llvm/circt/blob/96ddd2fed1a344979b8ffc3a1ca9343f2b5f7cb6/lib/Dialect/FIRRTL/Import/FIRParser.cpp#L3397-L3400)).

But in the external program, there is no such information as `InnerSymbolNamespace` and `uint64_t fieldID` (because they are private). What would be the best way to handle a situation like the `target` attribute of `rwprobe`?

SpriteOvO:
Solved by set the `inner_sym` only when needed, by using `mlirOperationSetInherentAttributeByName` function.

Thanks @seldridge's help!

seldridge:
Nice! I'm glad you got this working.

PR #7504, opened 2024-08-10
[HW][CAPI] Add a function to get an empty `InnerSymAttr`

SpriteOvO:
This allows us to set sym for partial ports - something like this pseudo-code:

```c
mlirOperationSetInherentAttributeByName(
    moduleOp, // The module has 3 ports created
    "portSyms",
    mlirArrayAttrGet({ hwInnerSymAttrGetEmpty(ctx), hwInnerSymAttrGet("name"), hwInnerSymAttrGetEmpty(ctx) })
);
```

### DIFF
--- a/include/circt-c/Dialect/HW.h
+++ b/include/circt-c/Dialect/HW.h
@@ -161,6 +161,7 @@ MLIR_CAPI_EXPORTED MlirStringRef hwTypeAliasTypeGetScope(MlirType typeAlias);
 
 MLIR_CAPI_EXPORTED bool hwAttrIsAInnerSymAttr(MlirAttribute);
 MLIR_CAPI_EXPORTED MlirAttribute hwInnerSymAttrGet(MlirAttribute symName);
+MLIR_CAPI_EXPORTED MlirAttribute hwInnerSymAttrGetEmpty(MlirContext ctx);
 MLIR_CAPI_EXPORTED MlirAttribute hwInnerSymAttrGetSymName(MlirAttribute);
 
 MLIR_CAPI_EXPORTED bool hwAttrIsAInnerRefAttr(MlirAttribute);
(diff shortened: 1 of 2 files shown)


dobios (approved):
I can see how this can be useful, and I'm guessing it is to you if you're making this PR so LGTM!

dtzSiFive:
I don't have bandwidth so don't mean to interfere without helping fix things 😉  but probably this suggests a better API would be useful, so the whole "set empty if any are non-empty" stuff doesn't propagate across the C API too 😞 .

Although I think the philosophy of the C api might be to be super simple and direct, which the above might violate.

But regardless it is what it is presently.

On that -- is there any use for setting inner symbols on specific fields, or only the entire "thing" (fieldid=0)?
Sincerely curious, and in part asking because empty could be an inner symbol with no "fieldid/name" pairs (visibility is there too but welp) included or something.  Maybe? 😄 .

SpriteOvO:
> On that -- is there any use for setting inner symbols on specific fields, or only the entire "thing" (fieldid=0)? Sincerely curious, and in part asking because empty could be an inner symbol with no "fieldid/name" pairs (visibility is there too but welp) included or something. Maybe? 😄 .

So far, we only used `inner_sym` when emitting `firrtl.ref.rwprobe` from Chisel, and from my understanding, it's cleanest to set `inner_sym` only when needed (see #6524 for the reason). So when `firrtl.ref.rwprobe` references a port, we only set `inner_sym` for that port. And to align the `portSyms` array with the ports array, we need empty `InnerSymAttr`s to populate the unused/unreferenced ports in the `portSyms` array.

sequencer:
> Although I think the philosophy of the C api might be to be super simple and direct, which the above might violate.

We noticed it, basically the C-API we added here is only intended to Build the MLIR Module from Java Binding, in the future, it may be used by other bindings or even a new DSL construction flow, but for now, the only consumer is the Chisel panamabinding framework. 

the panamabinding framework aims to eliminate the serialization in Chisel and parse in `firtool` to speed up the elaboration time, and possibly remove the code duplication in the future: we only need to maintain the ser/deser in CIRCT, and LLVM/MLIR provide a good infrastructure for us.

PR #7505, opened 2024-08-10
[LLHD] Add `llhd.delay` operation

maerhart:
Add an operation that delays the propagation of value changes from its input to its output. This allows mem2reg of LLHD signals with a unique driver with a delay. For example
```
%sig = llhd.sig
llhd.drv %sig, %val after <1ns, 0d, 0e>
%res = llhd.prb %sig
```
can become
```
%res = llhd.delay %val by <1ns, 0d, 0e>
```


fabianschuiki (approved):
LGTM!

PR #7507, opened 2024-08-11
[LLHD][NFC] Clean up regression tests a bit

maerhart:
Mostly just moving tests to two files which should make it easier to find specific tests and reduce the number of circt-opt invocations when running check-circt. Also avoids specifying ops in the generic format where not necessary.


fabianschuiki (approved):
LGTM!

PR #7508, opened 2024-08-11
[LLHD] Remove RegOp

maerhart:
LLHD's `reg` operation dates back to when the `seq` dialect was not a thing yet. It has the disadvantage that
* it only works with `inout` values
* while it can model all kinds of registers and latches you can think of, more than is actually used/necessary in practice, it is very complicated to work with

Instead of having this operation, we can
* use the seq registers
* if we need to work with inout values, we can continuously drive the seq register output to the signal with an additional `llhd.drv` operation
* If some state element occurs frequently for which we don't have an operation in the seq dialect, we can add one there specifically for that kind of element (e.g., a latch?)
* the `lllhd.reg` operation can specify triggers with different clocks. The most common case are async resets which we also support in the seq dialect. Other, more weird constructs could also just be left in the `llhd.process` based representation. We need to support them in simulation and verilog export anyway. Alternatively, we could also decompose it to the bare logic gates with feedback loops directly, don't know why that would be necessary though.

Please let me know what you think about this. Maybe I'm missing something.


fabianschuiki (approved):
Love this! I totally agree that the original register wad way too generic to be of practical use. Using the Seq dialect instead is a great idea. It also gets us away from operating on signals, which should be a lot easier 🙂.

PR #7511, opened 2024-08-12
[OM] Support list concatenation in the Evaluator.

mikeurbach:
List concatenation is implemented by first ensuring all the sub-lists are evaluated, and then appending the elements of each sub-list into the final ListValue that represents the concatenation of the lists.


uenoku (approved):
LGTM!

PR #7512, opened 2024-08-12
[FIRRTL] Add list concatenation parser support.

mikeurbach:
This parses `list_concat(exp+)` into the ListConcatOp. The type of the list is inferred during parsing, and all expressions must be the same. It is a parser error to not concatenate at least one list.

### DIFF
--- a/lib/Dialect/FIRRTL/Import/FIRParser.cpp
+++ b/lib/Dialect/FIRRTL/Import/FIRParser.cpp
@@ -1720,6 +1720,7 @@ private:
   ParseResult parsePrimExp(Value &result);
   ParseResult parseIntegerLiteralExp(Value &result);
   ParseResult parseListExp(Value &result);
+  ParseResult parseListConcatExp(Value &result);
 
   std::optional<ParseResult> parseExpWithLeadingKeyword(FIRToken keyword);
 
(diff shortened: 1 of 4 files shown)


lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -2455,6 +2466,44 @@ ParseResult FIRStmtParser::parseListExp(Value &result) {
...
+                 << operand.getType() << " in List concat expression of type "
+                 << type;
+
+        operands.push_back(operand);
+        return success();
+      }))
+    return failure();
+
+  if (operands.empty())
+    return emitError(loc, "need at least one List to concatenate");
```
mikeurbach, line 2500: This is legal syntactically according to the spec, but for this specific operation, the IR requires at least one operand. It makes the parser simpler to implement by requiring at least one operand and inferring the type from the types of the operands, but if it would be preferable to not require this during parsing (and instead let the op's verifier enforce it), I can change that.

dtzSiFive (approved):
LGTM!

Should `list_concat` support inserting `AnyRef` casts like done for `propassign` and `List` constructor?

I think disallowing it for now is fine, just checking since thought of it.

mikeurbach:
Thanks for the review. Good question about inserting casts... I thought about that, and I can't think of a reason we'd want that today. The automatic casts during list create ops might still make sense in some cases, but really I think we should move away from that. By the time we have list concat ops, we should already be dealing with lists of the same types (and if they're lists of AnyRefs, the casts would be inserted if necessary when the lists were created). I can't think of a use case for concatenating lists of different object types and wanting that to automatically insert the necessary casts to lists of AnyRef type; the frontend should really just handle this.

dtzSiFive:
> Thanks for the review. Good question about inserting casts... I thought about that, and I can't think of a reason we'd want that today.

Great! That sounds solid to me (and ty for explanation not quoted above for brevity)!

PR #7514, opened 2024-08-13
[Comb][Fold] Fix idemp n^2 and bugfix + more optimization.

dtzSiFive:
* Fix n^2 behavior with canonicalizeIdempotentInputs. Add arbitrary "depth" check to bound the search.
* Don't allow flattening or idempotent to search into operations defined in other blocks.
* Support removing duplicates even when operands come from other blocks. (or(x, y, x) -> or(x, x) regardless of their origin).
* Don't walk into operations with different two-state-ness in the idempotent operand canonicalizer.
* When creating new operation in idempotent canonicalizer, create it with matching two-state-ness.

Add tests for functional changes above.

### DIFF
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -122,7 +122,7 @@ static inline ComplementMatcher<SubType> m_Complement(const SubType &subExpr) {
 }
 
 /// Return true if the op will be flattened afterwards. Op will be flattend if
-/// it has a single user which has a same op type.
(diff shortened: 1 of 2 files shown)


uenoku (approved):
It looks reasonable to me to set the limit. LGTM!

dtzSiFive:
Thanks!

dtzSiFive:
(This was tested on various internal designs, and looks good, FWIW!)

fabianschuiki:
Really nice fix! 🥳

PR #7515, opened 2024-08-13
[LLHD] Relax sig parent constraint

maerhart:
This is necessary because the moore dialect does not enforce such constraints for its variables and thus they can occur in functions


fabianschuiki (approved):
LGTM!

PR #7516, opened 2024-08-14
LowerLayers: fix how we emit layer bindfile headers

rwy7:
For each layer, we emit a bindfile. Each bindfile has a header consisting of an include guard, and an include for each parent layer's bindfile. To emit the bindfile's headers, we walk the layer declarations, and keep track of a stack of parent layers, which we use to emit the includes.

Unfortunately, a bug in the stack management code means that we forget to unwind the stack of parent layers, when the curent layer has no parents. This caused us to emit an include from one layer to the previous layer, when the previous layer had any children.

This PR fixes the stack unwind loop to correctly clear the stack, when the current layer has no parents.


seldridge (approved):

dtzSiFive (approved):
LGTM! 
Thanks!

PR #7517, opened 2024-08-14
[ESI][Runtime] Generate C++ header files for constants

teqdruid:
Start of static C++ header file generation. Just integer constants for now. Can work off of any manifest and will even connect to a live accelerator and read the manifest from there.

Generates a `types.h` file and one per module listed in the manifest. Inside the module header file, generates one class per module and adds the constants to that class. Puts _everything_ in a namespace specified by the user as `system_name`.

### DIFF
--- a/lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
+++ b/lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
@@ -1,0 +1,193 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See https://llvm.org/LICENSE.txt for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Code generation from ESI manifests to C++ headers. Intended to be extensible
+# for other languages.
+
+from typing import List, Type, Optional
+from .accelerator import AcceleratorConnection
+from .esiCppAccel import ModuleInfo
+from . import types
+
+import argparse
+from pathlib import Path
+import textwrap
+import sys
+
+_thisdir = Path(__file__).absolute().resolve().parent
+
+
+class Generator:
+  """Base class for all generators."""
+
+  language: Optional[str] = None
+
+  def __init__(self, conn: AcceleratorConnection):
+    self.manifest = conn.manifest()
+
+  def generate(self, output_dir: Path, system_name: str):
+    raise NotImplementedError("Generator.generate() must be overridden")
+
+
+class CppGenerator(Generator):
+  """Generate C++ headers from an ESI manifest."""
+
+  language = "C++"
+
+  # Supported bit widths for lone integer types.
+  int_width_support = set([8, 16, 32, 64])
+
+  def get_type_str(self, type: types.ESIType) -> str:
+    """Get the textual code for the storage class of a type.
+
+    Examples: uint32_t, int64_t, CustomStruct."""
+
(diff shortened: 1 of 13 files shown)


lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
```diff
@@ -0,0 +1,193 @@
...
+
+import argparse
+from pathlib import Path
+import textwrap
+import sys
+
+_thisdir = Path(__file__).absolute().resolve().parent
+
+
+class Generator:
```
mortbopet, line 21: move to other file?

lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
```diff
@@ -0,0 +1,193 @@
...
+    """Get the C++ code for a constant in a module."""
+    const_strs: List[str] = [
+        f"static constexpr {self.get_type_str(const.type)} "
+        f"{name} = 0x{const.value:x};"
+        for name, const in module_info.constants.items()
+    ]
+    return "\n".join(const_strs)
+
+  def write_modules(self, output_dir: Path, system_name: str):
+    """Write the C++ one header for each module in the manifest."""
```
mortbopet, line 64: Rephrase

lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
```diff
@@ -0,0 +1,193 @@
...
+
+    for module_info in self.manifest.module_infos:
+      s = f"""
+      /// Generated header for {system_name} module {module_info.name}.
+      #pragma once
+      #include "types.h"
```
mortbopet, line 70: Way too generic, and alias-prone. `ESITypes.h` at a minimum. Ideally, there'd be some include path here (`#include "ESI/types.h"`).

lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
```diff
@@ -0,0 +1,193 @@
...
+      """))
+
+  argparser.add_argument("--file",
+                         type=str,
+                         default=None,
+                         help="Path to the manifest file.")
+  argparser.add_argument(
+      "--platform",
+      type=str,
+      help="Name of backend for live accelerator connection.")
```
mortbopet, line 145: platform or backend? Seems like it's referring to backends, but maybe im misunderstanding the scope of backends/(platforms?).

mortbopet (commented):
Missing a test/check on the generated .h file. Would also make this PR a bit easier to review if there was something that one could inspect to see the output of the generator.

lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
teqdruid: The idea is that users could choose the name of the folder in which to output the files. (e.g. `ESI/` or `<system-specific-name>/`) so the output file would be something like `ESI/types.h`.

lib/Dialect/ESI/runtime/python/esiaccel/cppgen.py
teqdruid: Backend ~= platform. Technically, a backend is a particular runtime implementation of an FPGA platform. I do expect a 1:1 mapping in most cases. I tend to refer to platform for anything user facing.

mortbopet (approved):

Issue #7482, opened 2024-08-09
[Moore] Redesign `wait_event` operation


maerhart:
The current design of the operation is not able to capture all possible situations in SystemVerilog and can be confusing (e.g., someone might think execution blocks at the operation until the event is true). An initial discussion can be found in PR #7362

PR #7518, opened 2024-08-15
[Moore] Improve WaitEventOp, lower to LLHD

fabianschuiki:
Rework the `moore.wait_event` op to be able to accurately model the semantics of SystemVerilog's `@...` event control statements. The op now has a body region which is executed to detect a relevant change in one or more interesting values. A new `moore.detect_event` op serves as the mechanism to encode whether a posedge, negedge, both, or any change at all on a value should be detected as an event.

Based on this new `moore.wait_event` op we can properly convert most of the event control statements in `ImportVerilog` to a corresponding MLIR op. Delay control like `#1ns` is not handled yet.

In the MooreToCore conversion this new op allows us to properly generate `llhd.wait` operations at the right place that suspend process execution until an interesting event has occurred. This now also allows us to support almost all SystemVerilog processes in the lowering. The only missing ones are `always_comb` and `always_latch` which require an implicit `llhd.wait` to be inserted. @maerhart has a version of that lowering almost ready though.

This commit also adds an `llhd.final` op in order to be able to lower `final` procedures.

### DIFF
--- a/include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
+++ b/include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
@@ -12,6 +12,7 @@
 
 def ProcessOp : LLHDOp<"process", [
   NoRegionArguments,
+  RecursiveMemoryEffects,
   HasParent<"hw::HWModuleOp">
 ]> {
   let summary = "create a process";
@@ -48,6 +49,39 @@ def ProcessOp : LLHDOp<"process", [
 
   let regions = (region MinSizedRegion<1>: $body);
 
+  let assemblyFormat = "attr-dict-with-keyword $body";
+}
+
+def FinalOp : LLHDOp<"final", [
+  NoRegionArguments,
+  RecursiveMemoryEffects,
+  HasParent<"hw::HWModuleOp">,
+]> {
+  let summary = "A process that runs at the end of simulation";
+  let description = [{
+    An `llhd.final` op encapsulates a region of IR that is to be executed after
+    the last time step of a simulation has completed. This can be used to
+    implement various forms of state cleanup and tear-down. Some verifications
+    ops may also want to check that certain final conditions hold at the end of
+    a simulation run.
+
+    The `llhd.wait` terminator is not allowed in `llhd.final` processes since
+    there is no later time slot for the execution to resume. Control flow must
+    eventually end in an `llhd.halt` terminator.
+
+    Execution order between multiple `llhd.final` ops is undefined.
+
+    Example:
+    ```mlir
+    hw.module @Foo() {
+      llhd.final {
+        func.call @printSimulationStatistics() : () -> ()
+        llhd.halt
+      }
+    }
+    ```
+  }];
+  let regions = (region MinSizedRegion<1>: $body);
   let assemblyFormat = "attr-dict-with-keyword $body";
 }
 
@@ -108,12 +142,14 @@ def WaitOp : LLHDOp<"wait", [
   }];
 }
 
-def HaltOp : LLHDOp<"halt", [Terminator, HasParent<"ProcessOp">]> {
+def HaltOp : LLHDOp<"halt", [
+  Terminator,
+  ParentOneOf<["ProcessOp", "FinalOp"]>
+]> {
   let summary = "Terminates execution of a process.";
   let description = [{
     The `halt` instruction terminates execution of a process. All processes
     must halt eventually or consist of an infinite loop.
   }];
-
   let assemblyFormat = "attr-dict";
 }
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -358,58 +358,103 @@ def NonBlockingAssignOp : AssignOpBase<"nonblocking_assign"> {
 // Statements
 //===----------------------------------------------------------------------===//
 
-def None: I32EnumAttrCase<"None", 0, "none">;
-/// Transit from 0 to x/z/1, and from x/z to 1.
+// Any change on the input.
+def AnyChange: I32EnumAttrCase<"AnyChange", 0, "any">;
+// A transition from 0 to X/Z/1, or from X/Z to 1.
 def PosEdge: I32EnumAttrCase<"PosEdge", 1, "posedge">;
-/// Transit from 1 to x/z/0, and from x/z to 0.
+// A transition from 1 to X/Z/0, or from X/Z to 0.
 def NegEdge: I32EnumAttrCase<"NegEdge", 2, "negedge">;
-/// Include the negedge and posedge.
+// The combination of `PosEdge` and `NegEdge`.
 def BothEdges: I32EnumAttrCase<"BothEdges", 3, "edge">;
 
-def EdgeAtrr: I32EnumAttr<"Edge", "Edge kind",
-                          [None, PosEdge, NegEdge, BothEdges]>{
+def EdgeAttr: I32EnumAttr<"Edge", "Edge kind",
+                          [AnyChange, PosEdge, NegEdge, BothEdges]> {
   let cppNamespace = "circt::moore";
 }
 
-def EventOp : MooreOp<"wait_event", [
-  HasParent<"ProcedureOp">
+def WaitEventOp : MooreOp<"wait_event", [
+  RecursiveMemoryEffects,
+  NoRegionArguments,
+  SingleBlock,
+  NoTerminator
 ]> {
-  let summary = "Detecting posedge and negedge";
+  let summary = "Suspend execution until an event occurs";
   let description = [{
-    It is introduced by the symbol `@`, and it allows statement execution to
-    be delayed until the occurrence of some simulation event occurring in a
-    procedure executing concurrently with this procedure.
+    The `moore.wait_event` op suspends execution of the current process until
+    its body signals that an event has been the detected. Conceptually, the body
+    of this op is executed whenever any potentially relevant signal has changed.
+    If one of the contained `moore.detect_event` ops detect an event, execution
+    resumes after the `moore.wait_event` operation. If no event is detected, the
+    current process remains suspended.
 
-    For the implicit event control(`@(*)`), there are two situations that are
-    not automatically added to event expression:
-      1. Identifiers that only appear in wait or event expressions.
-         ```
-          always @(*) begin    // equivalent to @(b)
-            @(n) kid = b;      // n is not added to @(*)
-          end
-         ```
-      2. Identifiers that only appear as a hierarchical_variable_identifier
-         in the variable_lvalue of the left-hand side of assignments.
-         ```
-          always @(*) begin
-            a = b + c;         // a is not added to @(*)
-          end
-         ```
+    Example corresponding to the SystemVerilog `@(posedge x, negedge y iff z)`:
+    ```
+    moore.wait_event {
+      %0 = moore.read %x : <i1>
+      %1 = moore.read %y : <i1>
+      %2 = moore.read %z : <i1>
+      moore.detect_event posedge %0 : i1
+      moore.detect_event negedge %1 if %2 : i1
+    }
+    ```
 
-    Example:
+    The body may also contain any operations necessary to evaluate the event
+    conditions. For example, the SV `@(posedge ~x iff i == 42)`:
     ```
-      @(a, b, c)        // none
-      @(posedge clk)    // positive edge
-      @(negedge clk)    // negative edge
-      @(edge clk)       // both edge
-      @(*)              // implicit event control
+    moore.wait_event {
+      %0 = moore.read %x : <i1>
+      %1 = moore.not %0 : i1
+      %2 = moore.read %i : <i19>
+      %3 = moore.constant 42 : i19
+      %4 = moore.eq %2, %3 : i19
+      moore.detect_event posedge %0 if %4 : i1
+    }
     ```
+
     See IEEE 1800-2017 § 9.4.2 "Event control".
   }];
-  let arguments = (ins EdgeAtrr:$edge, UnpackedType:$input);
-  let results = (outs);
+  let regions = (region SizedRegion<1>:$body);
+  let assemblyFormat = [{ attr-dict-with-keyword $body }];
+}
+
+def DetectEventOp : MooreOp<"detect_event", [
+  HasParent<"WaitEventOp">
+]> {
+  let summary = "Check if an event occured within a `wait_event` op";
+  let description = [{
+    The `moore.detect_event` op is used inside the body of a `moore.wait_event`
+    to check if an interesting value change has occurred on its operand. The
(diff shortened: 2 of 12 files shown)


test/Conversion/MooreToCore/basic.mlir
```diff
@@ -425,134 +425,6 @@ moore.module @Struct(in %a : !moore.i32, in %b : !moore.i32, in %arg0 : !moore.s
...
-    moore.wait_event negedge %cr : i1
-    %0 = moore.add %br, %cr : i1
-    cf.cond_br %cond, ^bb1, ^bb2
-  ^bb1:
-    moore.nonblocking_assign %d, %0 : i1
-    moore.return
-  ^bb2:
-    moore.return
-  }
-}
```
maerhart, line 554: This removes tests for the assign operations without adding replacements.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -358,58 +358,103 @@ def NonBlockingAssignOp : AssignOpBase<"nonblocking_assign"> {
...
+    | 0     | -       | posedge | posedge | posedge |
+    | 1     | negedge | -       | negedge | negedge |
+    | X     | negedge | posedge | -       | -       |
+    | Z     | negedge | posedge | -       | -       |
+
+    See IEEE 1800-2017 § 9.4.2 "Event control".
+  }];
+  let arguments = (ins
+    EdgeAttr:$edge,
+    UnpackedType:$input,
```
maerhart, line 453: Should this be restricted to a one-bit type or should we add a test for multi-bit operands?

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -169,142 +170,234 @@ struct ProcedureOpConversion : public OpConversionPattern<ProcedureOp> {
...
-      rewriter.setInsertionPoint(returnOp);
-      rewriter.create<cf::BranchOp>(loc, wait);
-      rewriter.eraseOp(returnOp);
+    // If any `detect_event` op detected an event, branch to the "resume" block
+    // which contains any code after the `wait_event` op. If no events were
+    // detected, branch back to the "wait" block to wait for the next change on
+    // the interesting signals.
+    rewriter.setInsertionPointToEnd(checkBlock);
+    if (!triggers.empty()) {
+      auto triggered = rewriter.createOrFold<comb::OrOp>(loc, triggers, true);
```
maerhart, line 395: So, no short-circuiting?

maerhart (commented):
This is the first part of my review, I'll finish it later 😅

cepheus69:
It is excited to see its taking shape 👍.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -169,142 +170,234 @@ struct ProcedureOpConversion : public OpConversionPattern<ProcedureOp> {
...
+        newOp = rewriter.create<llhd::ProcessOp>(loc);
+      else
+        newOp = rewriter.create<llhd::FinalOp>(loc);
+      auto &body = newOp->getRegion(0);
+      rewriter.inlineRegionBefore(op.getBody(), body, body.end());
+      for (auto returnOp :
+           llvm::make_early_inc_range(body.getOps<ReturnOp>())) {
+        rewriter.setInsertionPoint(returnOp);
+        rewriter.replaceOpWithNewOp<llhd::HaltOp>(returnOp);
+      }
```
hailongSun2000, line 192: If I remember correctly, the **return** statement cannot be used in a(n) initial/final. Slang also detects it(`return statement is only valid inside task and function blocks`).

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
```diff
@@ -51,6 +51,32 @@ def ProcessOp : LLHDOp<"process", [
   let assemblyFormat = "attr-dict-with-keyword $body";
 }
 
+def FinalOp : LLHDOp<"final", [
+  NoRegionArguments,
```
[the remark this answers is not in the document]
fabianschuiki: Yeah good idea!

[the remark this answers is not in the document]
fabianschuiki: The ProcessOp probably also wants this, right?

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
```diff
@@ -51,6 +51,32 @@ def ProcessOp : LLHDOp<"process", [
...
+  HasParent<"hw::HWModuleOp">,
+]> {
+  let summary = "A process that runs at the end of simulation";
+  let description = [{
+    An `llhd.final` op encapsulates a region of IR that is to be executed after
+    the last time step of a simulation has completed. This corresponds to
+    Verilog's `final` process and can be used to implement various forms of
+    state cleanup and tear-down. Some verifications ops may also want to check
+    that certain final conditions hold at the end of a simulation run.
```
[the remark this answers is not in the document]
fabianschuiki: Added.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
```diff
@@ -51,6 +51,32 @@ def ProcessOp : LLHDOp<"process", [
...
 
+def FinalOp : LLHDOp<"final", [
+  NoRegionArguments,
+  HasParent<"hw::HWModuleOp">,
+]> {
+  let summary = "A process that runs at the end of simulation";
+  let description = [{
+    An `llhd.final` op encapsulates a region of IR that is to be executed after
+    the last time step of a simulation has completed. This corresponds to
+    Verilog's `final` process and can be used to implement various forms of
```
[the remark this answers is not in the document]
fabianschuiki: Done.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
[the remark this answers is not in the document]
fabianschuiki: Done

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -358,58 +358,103 @@ def NonBlockingAssignOp : AssignOpBase<"nonblocking_assign"> {
...
-                          [None, PosEdge, NegEdge, BothEdges]>{
+def EdgeAttr: I32EnumAttr<"Edge", "Edge kind",
+                          [AnyChange, PosEdge, NegEdge, BothEdges]> {
   let cppNamespace = "circt::moore";
 }
 
-def EventOp : MooreOp<"wait_event", [
-  HasParent<"ProcedureOp">
+def WaitEventOp : MooreOp<"wait_event", [
+  RecursiveMemoryEffects,
```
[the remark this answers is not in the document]
fabianschuiki: I don't think so, because you can write something like this:
```systemverilog
task foo(ref logic x);
  @(posedge x);
endtask

module Bar;
  logic y;
  always foo(y);
endmodule
```
which would probably lower to this:
```mlir
func.func @foo(%x: !moore.ref<l1>) {
  moore.wait_event {
    %0 = moore.read %x : <l1>
    moore.detect_event posedge %0 : l1
  }
}
hw.module @Bar() {
  %y = moore.variable : <l1>
  moore.procedure always {
    func.call @foo(%y) : (!moore.ref<l1>) -> ()
  }
}
```
I'm not entirely sure how we'd deal with this… probably some lowering pass on functions to make sure we can suspend in the middle of a function call 🤔

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -236,7 +236,11 @@ static void populateMooreTransforms(PassManager &pm) {
     // Perform module-specific transformations.
     auto &modulePM = pm.nest<moore::SVModuleOp>();
     modulePM.addPass(moore::createLowerConcatRefPass());
-    modulePM.addPass(moore::createSimplifyProceduresPass());
+    // TODO: Enable the following once it not longer interferes with @(...)
+    // event control checks. The introduced dummy variables make the event
+    // control observe a static local variable that never changes, instead of
+    // observing a module-wide signal.
```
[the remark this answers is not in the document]
fabianschuiki: Yes exactly. The current implementation of `SimplifyProcedures` introduces shadow variables for any global variable that it accesses. Since we now have regular Control Flow ops and basic blocks, and mem2reg is becoming more solid, we'll have to revisit this and figure out what exactly we want to do in this pass. Some form of mapping assignments of global variables to process outputs would be great.

test/Conversion/MooreToCore/basic.mlir
fabianschuiki: Ugh thanks for catching that! 👍

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: I think you can have multi-bit operands, and the edge detection truth table will apply per bit. If the result is non-zero you get an event. The current MooreToCore conversion might need a few extensions to work reliably with multi-bit operands. Maybe a follow-up PR?

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: I don't think there is short-circuiting between the events in SystemVerilog. Having it be an expression may make lowering a bit easier later 😃

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Yes you're right! The Moore dialect has its own `moore.return` operation, which is used to terminated `moore.procedure` ops. We should probably give it a better name… A `return` in Verilog is mapped to a `func.return` in MLIR, which is different from the `moore.return` inserted implicitly as a terminator for `moore.procedure`s 😬

hailongSun2000: Oh! I mistook `body.getOps<ReturnOp>` as the return of function. Thanks for correcting me!

tools/circt-verilog/circt-verilog.cpp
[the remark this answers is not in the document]
hailongSun2000: Because we originally planned to inline the body of `always_comb` into the top level/module(`hw.module`). Since we need to ensure the variable has a correct value(After running `SimplifyProcedures`, we will run `mem2reg` to eliminate the shadow variables). However, we would like to lower `moore.procedure` into `llhd.process` for the time being, I think maybe we needn't `SimplifyProcedures`, WDYT :grin:?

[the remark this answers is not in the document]
fabianschuiki: Yes that sounds fantastic! 😀 I'm pretty sure we can add SimplifyProcedures again in the future to optimize some procedures early on, but I totally agree with going to LLHD processes for now and doing optimizations there. 🥳

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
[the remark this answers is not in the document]
maerhart: Yes, good point!

include/circt/Dialect/Moore/MooreOps.td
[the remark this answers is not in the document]
maerhart: Ugh forgot about tasks 😞 
That's a good question indeed.

include/circt/Dialect/Moore/MooreOps.td
maerhart: Erroring out in MooreToCore for now sounds good to me.

lib/Conversion/MooreToCore/MooreToCore.cpp
maerhart: Ah, good to know that there's no short-circuiting in SV. 👍

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -169,142 +194,235 @@ struct ProcedureOpConversion : public OpConversionPattern<ProcedureOp> {
...
-    rewriter.create<llhd::WaitOp>(loc, toObserve, Value(), ValueRange{}, check);
-    rewriter.setInsertionPointToStart(check);
+    // Determine the values used during event detection that are defined outside
+    // the `wait_event`'s body region. We want to wait for a change on these
+    // signals before we check if any interesting event happened.
+    SmallVector<Value> observeValues;
+    SmallDenseSet<Value> alreadyObserved;
+    clonedOp.getBody().walk([&](Operation *operation) {
+      for (auto value : operation->getOperands()) {
+        if (value.getParentRegion() == &clonedOp.getBody())
```
maerhart, line 332: If there are nested regions within the wait_event op region, we'd have to check if `clonedOp.getBody()` is an ancestor region of the value defining region.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -169,142 +194,235 @@ struct ProcedureOpConversion : public OpConversionPattern<ProcedureOp> {
...
+    // If any `detect_event` op detected an event, branch to the "resume" block
+    // which contains any code after the `wait_event` op. If no events were
+    // detected, branch back to the "wait" block to wait for the next change on
+    // the interesting signals.
+    rewriter.setInsertionPointToEnd(checkBlock);
+    if (!triggers.empty()) {
+      auto triggered = rewriter.createOrFold<comb::OrOp>(loc, triggers, true);
+      rewriter.create<cf::CondBranchOp>(loc, triggered, resumeBlock, waitBlock);
+    } else {
+      rewriter.create<cf::BranchOp>(loc, waitBlock);
```
maerhart, line 423: In this case, the resume block would have no predecessors. Do you think it's save to leave this to the conversion framework? We don't have a test for this case to be sure.

maerhart (approved):
And the rest of my review. Thanks for all these great improvements! 🎉

maerhart:
Resolves #7482

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Good point, I have added the pattern I've seen failing before to the test with an empty `wait_event`. It looks like the dialect conversion uses the dominance information collected _before_ the conversion to iterate through ops -- looks like making a block unreachable doesn't exclude it from the conversion.

fabianschuiki: Good point, I have added the pattern I've seen failing before to the test with an empty `wait_event`. It looks like the dialect conversion uses the dominance information collected _before_ the conversion to iterate through ops -- looks like making a block unreachable doesn't exclude it from the conversion.

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Nice catch! I've changed this and added a test for it.

PR #7519, opened 2024-08-15
[LowerIntmodules] Fix EICG_wrapper intrinsic lowering, swap en/test_en.

dtzSiFive:
The extmodule and intrinsic disagree on port ordering, swap in LowerIntmodules.

Thanks to @prithayan for reporting this!

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerIntmodules.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerIntmodules.cpp
@@ -181,6 +181,20 @@ void LowerIntmodulesPass::runOnOperation() {
         };
 
         auto inputs = replaceResults(builder, inst.getResults().drop_back());
+        // en and test_en are swapped between extmodule and intrinsic.
+        if (inputs.size() > 2) {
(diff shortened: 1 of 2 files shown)


fabianschuiki (approved):
🥳 LGTM! I'm wondering whether the port order of `EICG_wrapper` is fixed, or if there may be different definitions that have totally different port order. Great to have the lowering complain about those cases 👍

prithayan (approved):
Thanks for the quick fix.

dtzSiFive:
> 🥳 LGTM! I'm wondering whether the port order of `EICG_wrapper` is fixed, or if there may be different definitions that have totally different port order. Great to have the lowering complain about those cases 👍

I'm leaning towards keeping it this way for now since grabbing and checking each by name appropriately is a pain and this is already something we're close to phasing out.

But you're right that's how it should work.

FWIW:
We never supported that and checked names in a particular order in the past, and blindly wired to it assuming the order 😓 .
Not that that's a great reason to keep that behavior...

We DO emit them in the other order in SeqToSV, so that's fun.

Issue #7418, opened 2024-07-31
[firtool] use FIRRTL layer syntax for specialization CL options


youngar:
The syntax for nested layers doesn't match the FIRRTL spec and instead matches the MLIR representation. It would seem to make more sense to use the FIRRTL syntax.
Namely, `firtool -enable-layers=Verification.Assert` instead of `firtool -enable-layers=Verification::Assert`.

youngar:
Fix would be to this code: https://github.com/llvm/circt/blob/3c12682a1aff12249d89fb01cfa0554c0ea28f1b/lib/Dialect/FIRRTL/Import/FIRParser.cpp#L5581-L5593

PR #7520, opened 2024-08-15
[firtool] Change layer specialization CLI interface

youngar:
This changes how nested layers are specified when using --enable-layers or --disable-layers.  Previously a nested layer was specified with `A::B`, but not it follows the syntax in the FIRRTL spec for layers, which is `A.B`.


seldridge (approved):
LGTM

dtzSiFive (approved):
LGTM.

Fixes https://github.com/llvm/circt/issues/7418 , right?

PR #7521, opened 2024-08-15
[NFC] Add circt copy of upstream's linkify.

dtzSiFive:
See: https://llvm.org/pr103496


dtzSiFive:
Suddenly I want a version that works for git repo's with github remotes as upstream or something! 😄 (for use globally haha)

fabianschuiki (approved):
🥳

PR #7522, opened 2024-08-15
[FIRRTL] Update LowerClasses to alter what PathInfo stores.

mikeurbach:
PathInfo stores a pointer to an Operation, which was problematic because that Operation may be deleted in updateInstanceInModule. The test case added in this patch would lead to a use-after-free.

This pointer was only really used for a few things, which can be handled differently to avoid needing to consider Operation lifetimes.

One use was the operator bool implementation, to check if a PathInfo is empty. In the one place this was used, an equivalent check is to query the PathInfoTable, and check if the key was not in the table.

Another use was adding the Operation's location in error messages and notes. We can safely store a Location directly for these messages.

The final use was to do an isa check while determining the target kind. This is where the use in the use-after-free would manifest. For this, we do the isa check early, and store the result in a bool.

In summary, we are able to simplify the data in PathInfo in order to avoid hanging on to an Operation pointer and needing to worry about its lifetime.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -81,17 +81,18 @@ static bool shouldCreateClassImpl(igraph::InstanceGraphNode *node) {
 /// to the targeted operation.
 struct PathInfo {
   PathInfo() = default;
-  PathInfo(Operation *op, FlatSymbolRefAttr symRef,
+  PathInfo(Location loc, bool canBeInstanceTarget, FlatSymbolRefAttr symRef,
            StringAttr altBasePathModule)
-      : op(op), symRef(symRef), altBasePathModule(altBasePathModule) {
-    assert(op && "op must not be null");
+      : loc(loc), canBeInstanceTarget(canBeInstanceTarget), symRef(symRef),
+        altBasePathModule(altBasePathModule) {
     assert(symRef && "symRef must not be null");
   }
 
-  operator bool() const { return op != nullptr; }
+  /// The Location of the hardware component targeted by this path.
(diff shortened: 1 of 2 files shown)


dtzSiFive (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -81,17 +81,18 @@ static bool shouldCreateClassImpl(igraph::InstanceGraphNode *node) {
...
-      : op(op), symRef(symRef), altBasePathModule(altBasePathModule) {
-    assert(op && "op must not be null");
+      : loc(loc), canBeInstanceTarget(canBeInstanceTarget), symRef(symRef),
+        altBasePathModule(altBasePathModule) {
     assert(symRef && "symRef must not be null");
   }
 
-  operator bool() const { return op != nullptr; }
+  /// The Location of the hardware component targeted by this path.
+  std::optional<Location> loc = std::nullopt;
```
youngar, line 92: Just double checking, since we're not using lookup anymore, do we need the PathInfo to be default constructible?

youngar (approved):
💯

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
mikeurbach: Hmm maybe not, I'm not sure if anything else still requires that.

mikeurbach: Yeah, default constructible is still required, at least for how we're using it when we add items with `try_emplace`. I will go ahead and merge this for now, and we could potentially keep reworking this.

PR #7523, opened 2024-08-15
[LLHD] Align signals with other wire/variable ops

fabianschuiki:
Make the `llhd.sig` op use the same naming pattern as HW wires, Moore variables, Seq registers, and a handful of other operations in CIRCT. These all use the `custom<ImplicitSSAName>` parser to provide uniform handling of optional names.

Make the signal name optional to align with other ops.

Rename the class to `SignalOp` for clarity.

### DIFF
--- a/lib/Dialect/LLHD/IR/LLHDOps.cpp
+++ b/lib/Dialect/LLHD/IR/LLHDOps.cpp
@@ -12,6 +12,7 @@
 
 #include "circt/Dialect/LLHD/IR/LLHDOps.h"
 #include "circt/Dialect/HW/HWOps.h"
+#include "circt/Support/CustomDirectiveImpl.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Matchers.h"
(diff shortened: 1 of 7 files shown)


lib/Dialect/LLHD/IR/LLHDOps.cpp
```diff
@@ -64,6 +66,15 @@ void llhd::ConstantTimeOp::build(OpBuilder &builder, OperationState &result,
   return build(builder, result, TimeType::get(ctx), attr);
 }
 
+//===----------------------------------------------------------------------===//
+// SignalOp
+//===----------------------------------------------------------------------===//
+
+void SignalOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
+  if (getName() && !getName()->empty())
```
maerhart, line 74: I'm wondering why we don't actually verify that the string is not empty?

maerhart (approved):
LGTM!

lib/Dialect/LLHD/IR/LLHDOps.cpp
fabianschuiki: I think there were some ops that used empty strings to encode an absent name... We should definitely change that to either null str attr or a non-empty string.

PR #7527, opened 2024-08-19
[CombFolds] Don't canonicalize extract(shl(1, x)) if shift is multiply used

uenoku:
There is a canonicalization for `exract(c, shl(1, x))` to `x == c` but this canonicalization introduces a bunch of comparision to constants. This harms PPA when bitwidth is large (e.g. 16 bit shift introduce 2^16 icmp op). To prevent such regressions this commit imposes restriction regarding the number of uses for shift.


seldridge (approved):
LGTM barring a diff of an internal design to see how much this changes.

uenoku:
There are substantial changes in output verilog but all of them are expected changes. The change should preserve a design intent of one-hot encoding in a much better way.

PR #7528, opened 2024-08-19
[LLHD] Let WaitOp observe plain values instead of signals

maerhart:
This is necessary to lower the moore dialect's always_comb and always_latch without introducing helper signals. It also allows for more mem2reg at the LLHD level.


fabianschuiki (approved):
This is really nice! :partying_face:

PR #7530, opened 2024-08-19
[FIRRTL][FIRParser] Cache property constants.

dtzSiFive:
Greatly reduces IR size generated in presence of many repeat constants as commonly occurs in practice due to the data-like nature of classes / properties.

Same tricky as done for FIRRTL integer constants,
use single cache/code for both.


mikeurbach (approved):
Nice, makes sense to me!

PR #7398, opened 2024-07-29
[FIRRTL] Inliner: Support for ops with regions.

dtzSiFive:
inlineInstances/flattenInstances:
* Walk entire body, not only top-level operations.
  Fixes missing instances and allows inlining them
  when conservatively legal.
* Reject inlining instances under when/match.

inlineInto/flattenInto:
  Walk entire body using new `inliningWalk` method
  that drives the per-operations handling but also
  handles cloning "structure" operations that have
  regions (when/match/layer) and managing what
  should be cloned where.

  This allows inlining modules that contain these
  operations.

Inliner now may produce errors, thread throughout.

This allows the inliner to run earlier in the pipeline,
particularly before LowerLayers.

PR #7534, opened 2024-08-19
[firtool] Run canonicalization before LowerLayers

### LINKED ISSUE #7533: [FIRRTL] design with layers not optimized as well as without
This was uncovered while trying to move verification statements automatically into layers in Chisel. I'm seeing a missed opportunity for eliminating a `circt_chisel_ifelsefatal` intrinsic.

Consider the following where the assertion predicate reduces to `0`:

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  layer A, bind:
  public module Foo:
    input clock: Clock
    input a: UInt<2>
    output c: UInt<2>

    node b = bits(a, 1, 0)
    layerblock A :
      intrinsic(circt_chisel_ifelsefatal<format = "bar">, clock, eq(a, b), UInt<1>(1))

    ; This user of b is important!
    connect c, b
```

When compiled `firtool Foo.fir` this leaves a trivially dead assertion in the design:

``` verilog
// Generated by CIRCT firtool-1.81.1-18-g6aa871500

// Users can define 'STOP_COND' to add an extra gate to stop conditions.
`ifndef STOP_COND_
  `ifdef STOP_COND
    `define STOP_COND_ (`STOP_COND)
  `else  // STOP_COND
    `define STOP_COND_ 1
  `endif // STOP_COND
`endif // not def STOP_COND_

// Users can define 'ASSERT_VERBOSE_COND' to add an extra gate to assert error printing.
`ifndef ASSERT_VERBOSE_COND_
  `ifdef ASSERT_VERBOSE_COND
    `define ASSERT_VERBOSE_COND_ (`ASSERT_VERBOSE_COND)
  `else  // ASSERT_VERBOSE_COND
    `define ASSERT_VERBOSE_COND_ 1
  `endif // ASSERT_VERBOSE_COND
`endif // not def ASSERT_VERBOSE_COND_
module Foo_A(
  input [1:0] a,
              b,
  input       clock
);

  `ifndef SYNTHESIS
    always @(posedge clock) begin
      if (a != b) begin
        if (`ASSERT_VERBOSE_COND_)
          $error("bar");
        if (`STOP_COND_)
          $fatal;
      end
    end // always @(posedge)
  `endif // not def SYNTHESIS
endmodule

module Foo(
  input        clock,
  input  [1:0] a,
  output [1:0] c
);

  assign c = a;
endmodule


// ----- 8< ----- FILE "layers_Foo_A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.81.1-18-g6aa871500
`ifndef layers_Foo_A
`define layers_Foo_A
bind Foo Foo_A a_0 (
  .a     (a),
  .b     (a),
  .clock (clock)
);
`endif // layers_Foo_A
```

However, if there is nothing to block `LayerSink` from moving the `node b` into the layerblock and optimization happens. This can be shown with the slightly modified circuit which removes the user of `b`:

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  layer A, bind:
  public module Foo:
    input clock: Clock
    input a: UInt<2>
    output c: UInt<2>

    node b = bits(a, 1, 0)
    layerblock A :
      intrinsic(circt_chisel_ifelsefatal<format = "bar">, clock, eq(a, b), UInt<1>(1))

    invalidate c
```

``` verilog
// Generated by CIRCT firtool-1.81.1-18-g6aa871500
module Foo(
  input        clock,
  input  [1:0] a,
  output [1:0] c
);

  assign c = 2'h0;
endmodule


// ----- 8< ----- FILE "layers_Foo_A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.81.1-18-g6aa871500
`ifndef layers_Foo_A
`define layers_Foo_A
`endif // layers_Foo_A
```

~I haven't tracked down the issue, yet. I expect it is something to do with optimizations or canonicalizations not working across layerblocks and this then relying on `LayerSink` to make the optimization available.~

The canonicalizer is responsible for removing this pattern. This runs _after_ `LowerLayers`. At that point, this analysis is no longer easy to do as it would require doing canonicalization across a module boundary or some notion of "inter module value numbering".

seldridge:
Run canonicalization before LowerLayers in order to optimize things which are difficult to do after layerblocks have been converted to modules.  The conversion to modules means that many trivial canonicalizers are now very difficult to do as they require inter-module analysis.

Ideally, it would be best if LowerLayers ran as late as possible.  (Generally any outlining should run as late as possible.) However, we aren't quite ready to make this harder move of pushing LowerLayers to the end of the FIRRTL pipeline.

Fixes #7533.

### DIFF
--- a/lib/Firtool/Firtool.cpp
+++ b/lib/Firtool/Firtool.cpp
@@ -140,6 +140,13 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
     modulePM.addPass(firrtl::createLayerSinkPass());
   }
 
+  // Run canonicalization before LowerLayers.  LowerLayers creates modules which
+  // makes certain optimizations much harderd.  E.g., many trivial
+  // canonicalizers will not work after LowerLayers outlines module bodies.
+  if (!opt.shouldDisableOptimization())
+    pm.nest<firrtl::CircuitOp>().nest<firrtl::FModuleOp>().addPass(
+        createSimpleCanonicalizerPass());
+
   pm.nest<firrtl::CircuitOp>().addPass(firrtl::createLowerLayersPass());
 
   pm.nest<firrtl::CircuitOp>().addPass(firrtl::createInlinerPass());
--- a/test/firtool/lower-layers.fir
+++ b/test/firtool/lower-layers.fir
@@ -177,3 +177,33 @@ circuit TestHarness:
 ; CHECK:   .reset     (reset)
 ; CHECK: );
 ; CHECK: `endif // layers_TestHarness_Verification
+
+; // -----
+
+; This is a test that the modules created from lowering layerblocks to modules
+; still have reasonable optimizations that are only possible with
+; canonicalization.  Creating modules from layers destroys a lot of optimization
+; opportunities and this test is ensuring that these are preserved.
+;
+; See: https://github.com/llvm/circt/issues/7533
+
+FIRRTL version 4.0.0
+
+circuit Foo: %[[
+  {"class": "firrtl.transforms.DontTouchAnnotation", "target": "~Foo|Foo>c"}
+]]
+  layer A, bind:
+  public module Foo:
+    input clock: Clock
+    input a: UInt<2>
+    output d: UInt<2>
+
+    node b = bits(a, 1, 0)
+    layerblock A :
+      node c = eq(a, b)
+
+    connect d, b
+
+; CHECK:     module Foo_A
+; CHECK-NOT: endmodule
+; CHECK:       wire c = 1'h1;


seldridge:
Discussed offline with @dtzSiFive. There appears to be at least one register canonicalizer (see `foldHiddenReset` in `FIRRTLFolds.cpp`) and this shouldn't be run _before_ `RandomizeRegisterInit` because that can cause changes in the randomization behavior due to optimizations.

It would be better to move `LowerLayers` later in the pipeline so that it runs after canonicalization. This would require landing https://github.com/llvm/circt/pull/7398 to move it after `ModuleInliner` and see about moving it after `RandomizeRegisterInit`. The latter change would be beneficial as this would ensure randomization stability involving specialization that enables layers (though not guarantee it for disabling layers).

uenoku:
IIRC we have wanted to perform register optimization only in `RegisterOptimization pass so removing foldHiddenReset  from canonicalizer looks fine solution to me.

Issue #7423, opened 2024-08-02
firtool misses `regreset` when generating .mlir from .fir with --preserve-aggregate=vec and array registers


Anillc:
command:

```bash
firtool -format=fir -ir-fir --preserve-aggregate=vec foo.fir

```

foo.fir:

```fir
FIRRTL version 4.0.0
circuit Top :%[[
  {
    "class":"firrtl.transforms.DontTouchAnnotation",
    "target":"~Top|Top>x[0][0]"
  }
]]


  public module Top :
    input clock : Clock
    input reset : UInt<1>


    wire foo : UInt<1>[1]
    connect foo[0], UInt<1>(0h1)
    wire bar : UInt<1>[1][1]
    connect bar[0], foo
    regreset x : UInt<1>[1][1], clock, reset, bar
```

result:

```mlir
module {
  firrtl.circuit "Top" {
    firrtl.module @Top(in %clock: !firrtl.clock, in %reset: !firrtl.uint<1>) attributes {convention = #firrtl<convention scalarized>} {
      %0 = firrtl.aggregateconstant [1 : ui1] : !firrtl.vector<uint<1>, 1>
      %foo = firrtl.wire : !firrtl.vector<uint<1>, 1>
      %x = firrtl.reg %clock {annotations = [{circt.fieldID = 2 : i32, class = "firrtl.transforms.DontTouchAnnotation"}], firrtl.random_init_start = 0 : ui64} : !firrtl.clock, !firrtl.vector<vector<uint<1>, 1>, 1>
      firrtl.matchingconnect %x, %x : !firrtl.vector<vector<uint<1>, 1>, 1>
      firrtl.matchingconnect %foo, %0 : !firrtl.vector<uint<1>, 1>
      %1 = firrtl.vectorcreate %foo : (!firrtl.vector<uint<1>, 1>) -> !firrtl.vector<vector<uint<1>, 1>, 1>
    }
  }
}
```

without `--preserve-aggregate=vec`:

```mlir
module {
  firrtl.circuit "Top" {
    firrtl.module @Top(in %clock: !firrtl.clock, in %reset: !firrtl.uint<1>) attributes {convention = #firrtl<convention scalarized>} {
      %c1_ui1 = firrtl.constant 1 : !firrtl.uint<1> {name = "foo_0"}
[...]

PR #7536, opened 2024-08-20
[FIRRTLUtils] Fix walkDrivers subfield id calculation

uenoku:
This fixes an overflow in walkDrivers when field ID is zero.

Fix https://github.com/chipsalliance/chisel/issues/4354 and https://github.com/llvm/circt/issues/7423.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLUtils.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLUtils.cpp
@@ -457,7 +457,13 @@ bool circt::firrtl::walkDrivers(FIRRTLBaseValue value, bool lookThroughWires,
         auto subRef = fieldRef.getSubField(subID);
         auto subOriginal = original.getSubField(subID);
         auto value = subfield.getResult();
-        workStack.emplace_back(subOriginal, subRef, value, fieldID - subID);
+        // If fieldID is zero, this points to entire subfields.
+        if (fieldID == 0)
+          workStack.emplace_back(subOriginal, subRef, value, 0);
+        else {
+          assert(fieldID >= subID);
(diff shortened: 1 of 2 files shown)


seldridge (approved):
LGTM

lib/Dialect/FIRRTL/FIRRTLUtils.cpp
```diff
@@ -457,7 +457,13 @@ bool circt::firrtl::walkDrivers(FIRRTLBaseValue value, bool lookThroughWires,
...
         auto subOriginal = original.getSubField(subID);
         auto value = subfield.getResult();
-        workStack.emplace_back(subOriginal, subRef, value, fieldID - subID);
+        // If fieldID is zero, this points to entire subfields.
+        if (fieldID == 0)
+          workStack.emplace_back(subOriginal, subRef, value, 0);
+        else {
+          assert(fieldID >= subID);
+          workStack.emplace_back(subOriginal, subRef, value, fieldID - subID);
+        }
```
uenoku, line 466: There might be the case that `fieldID < subID` so this fix may not work too.. :) Thank you @dtzWill for pointing it out!

youngar:
Trying to refresh my memory on how this code works - still pretty hazy but I think your fix makes sense.  Once the `fieldID` hits 0 we have found the field we are "interested in" and we should examine/include all further subfields.

dtzSiFive:
I was not sure and the test case from the issue IIRC didn't run into a case where either was true (the `>=` assertion).  Poking at it some more, I think this is right, especially if @youngar thinks so too.
The other direction is handled by chasing to the root of the source, AFAICT.

Thanks!

uenoku:
Thank you for review!

PR #7537, opened 2024-08-20
[OM] Rework ClassOp to use fields terminator

leonardt:
This changes the OM dialect Class/ExternClass to use a terminator operation similar to hw.output.

The ClassFields and ClassExternFields operations use a custom printer/parser to provide a syntax similar to hw.module where fields use the format `@name %val : !type` for non-extern fields and `@name : !type` for extern fields.

Internally, the fields operations use attributes to store required information. Both fields operations store a list of fields to preserve definition order.  Non-extern fields store a mapping from field name to operand index.  Extern fields store a mapping from field name to type.

The fields operations provide an API for getting the type of a field based on the name.  The extern fields looks this up in the attribute dictionary, while the non-extern fields operation looks up the operand index and retrieves the type from the appropriate operand.  This ensures that non-extern fields map to the latest type in cases where the operand type changes after construction.

The Class/ExternClass operations provide a convenience API that lifts the fields API to avoid having to frequently request the underlying fields operation.

The firrtl LowerClasses logic requires changes to use this API when constructing OM classes, as well as in the dialect conversion logic to handle the new operations.

The OM evaluator similarly is updated for the new fields API.

The OM linker is updated to skip constructing the field name to type map and instead using the getFieldType API provided by the classOp.

The OM object field verifier also no longer needs to construct the field name to type maps and instead can just use the new API.

Quite a few tests required updating for the new field syntax.

Two TODOs left:
* Is there a way to attach a note to a parser error? The old code used this when encountering duplicate fields (the note pointed to the previous definition).  Because this check is now done at parse time, we are emitting parser errors instead of an InFlightDiagnostic.
* Right now the getFieldType API uses a `std::optional` style for handling cases when the type of a field doesn't exist (i.e. the provided name is not field).  Another option would be to expose a `hasField` API.


leonardt:
Ref #7078

include/circt/Dialect/OM/OMOpInterfaces.td
```diff
@@ -15,6 +15,102 @@
...
+  let cppNamespace = "circt::om";
+
+  let description = [{
+    Marks the end of a class-like definition by providing a list of accessible
+    fields.
+  }];
+  let methods = [
+    InterfaceMethod<"Get the class-like field to type map",
+      "std::optional<mlir::Type>", "getFieldType", (ins "mlir::StringAttr":$name)>,
+    InterfaceMethod<"Print class fields like",
```
uenoku, line 28: I think  `printClassFieldsLike` is not suitable as OpInterfaceMethod since this is not useful other than MLIR prtiing. Could you move this as a static function in OMOps.cpp?

lib/Dialect/OM/OMOps.cpp
```diff
@@ -275,6 +448,56 @@ void circt::om::ClassExternOp::getAsmBlockArgumentNames(
...
+void circt::om::ClassExternOp::addFields(
+    mlir::OpBuilder &builder, mlir::Location loc,
+    llvm::ArrayRef<mlir::StringAttr> fieldNames,
+    llvm::ArrayRef<mlir::Type> fieldTypes) {
+  auto *op = builder.create<ClassExternFieldsOp>(loc).getOperation();
+
+  auto *ctx = builder.getContext();
+  llvm::SmallVector<NamedAttribute> typeAttrs;
+  llvm::SmallVector<NamedAttribute> fieldIdxs;
+  for (unsigned i = 0; i < fieldNames.size(); i++) {
```
uenoku, line 460: nit:
```suggestion
  for (unsigned i = 0, e = fieldNames.size(); i < e ; i++) {
```

include/circt/Dialect/OM/OMOps.td
```diff
@@ -82,46 +84,59 @@ def ClassOp : OMClassLike<"class"> {
...
 }
 
-def ClassFieldOp : OMClassFieldLike<"class.field",
-    [HasParent<"ClassOp">]> {
-  let arguments = (ins
-    SymbolNameAttr:$name,
-    AnyType:$value
-  );
+def ClassFieldsOp : OMClassFieldsLike<"class.fields", [HasParent<"ClassOp">]> {
+  let arguments = (ins Variadic<AnyType>:$fieldArgs);
```
uenoku, line 106: I think we need `fieldNames` as well?

include/circt/Dialect/OM/OMOpInterfaces.td
```diff
@@ -15,6 +15,102 @@
...
+        }
+        printer.decreaseIndent();
+        if (!fieldNames.empty())
+          printer.printNewline();
+        printer << ")";
+        printer.printOptionalAttrDict(op->getAttrs(),
+                                      /*elidedAttrs=*/{"fieldTypes", "fieldNames",
+                                                       "fieldIdxs", "perFieldAttrs",
+                                                       "perFieldLocs"});
+      }]>,
```
uenoku, line 95: i think we generally want to keep these attributes as inherent attributes not discardable. Could you add these attributes to op's arguments in ODS?

uenoku (commented):
I think this is great. Thank you for working on this! It's great improvement to use  terminator instead of individual field class op. I'm wondering if we can do further: can we encode field names into its class declaration? e.g.

```mlir
om.class @NestedField1() -> (baz: i1) {
  %0 = om.constant 1 : i1
  om.class.fields %0 : i1
}

om.class @NestedField2() -> (bar: !om.class.type<@NestedField1>) {
  %0 = om.object @NestedField1() : () -> !om.class.type<@NestedField1>
  om.class.fields %0 : !om.class.type<@NestedField1>
}
```

With that change I think we don't need ClassExternFieldsOp since extern class declaration has the same information (and interface OMClassFieldsLike would be unnecessary as well).

leonardt:
Thanks for the comments @uenoku , I'll work on them and let you know when it's ready for another review.

Moving fields to the class was something @mikeurbach mentioned at the outset. I think we thought that this approach seemed like a simpler incremental change, but now I agree that using the class would remove a layer of indirection (going from class op to fields op) and simplify things. Will look into that.

I don't know have experience with the issues with symbols so I'd trust your opinion on that, so happy to move to "raw" field names.

mikeurbach:
Sorry for the delay in giving a serious review. I can at least respond to a couple of the points you guys discussed...

I like the idea of putting the field names and types on the class... that is conceptually even closer to the design point that HW dialect settled on, and if it simplifies things, it might be worth going straight to that in this change. At this point @leonardt has iterated on different designs than I have, so I think it's up to you if it's worth doing more iterations or stopping at this point. Personally, I think we're not in a rush to get this out, so if it is worth looking into I vote go for it.

Regarding symbols... this was one of my original sins. Even after we made om.class.field no longer a Symbol op, i let them print and parse as symbol names, which is just plain misleading when you look at the textual IR. Regardless of how they're printed or parsed, symbol names are internally StringAttr, but using the `@foo` syntax is the printed form should probably be reserved for actual symbol names (the sym_name attribute of an op that declares the Symbol trait).

I'll take a pass through the PR for any low-level coding comments, which might be applicable regardless of whether we stay at this design point or move towards Hideto's suggestion.

include/circt/Dialect/OM/OMOpInterfaces.td
```diff
@@ -15,6 +15,102 @@
...
+      "void", "printClassFieldsLike", (ins "mlir::OpAsmPrinter &":$printer,
+                                           "llvm::SmallVector<mlir::Value>":$operands),
+      /*methodBody=*/[{}],
+      /*defaultImplementation=*/[{ 
+        // * assumes fieldNames dictionary attr exists
+        // * assumes each name in fieldNames has an entry in fieldTypes
+        //   dictionary attr mapping the name to type
+        // * perFieldAttrs, perFieldLocs optional attr/locs for each field
+        // * each field is emitted on a separate line
+        mlir::Operation *op = $_op.getOperation();
```
mikeurbach, line 38: IMHO this is too much C++ to write inline in a tablegen file. There aren't hard and fast rules, but if something gets to this level, and you still want to give a default implementation, I'd suggest defining a free function and calling it from the defaultImplementation block. As Hideto pointed out, maybe this doesn't need to be an interface method anyway, but just a thought about C++ in tablegen in general.

include/circt/Dialect/OM/OMOpInterfaces.td
mikeurbach: I agree, and if you haven't seen these terms, there is a brief description here: https://mlir.llvm.org/docs/LangRef/#attributes.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -1628,31 +1643,50 @@ struct ObjectSubfieldOpConversion
...
+    fieldTypes.push_back(mlir::NamedAttribute(
+        cast<StringAttr>(field),
+        mlir::TypeAttr::get(typeConverter->convertType(type.value()))));
+  }
+  auto *ctx = newOp.getContext();
+  newOp->setAttr("fieldNames", mlir::ArrayAttr::get(ctx, {fieldNames.begin(),
+                                                          fieldNames.end()}));
+  newOp->setAttr("fieldTypes", mlir::DictionaryAttr::get(ctx, fieldTypes));
+  newOp->setAttr("fieldIdxs", fieldIdxs);
+  return llvm::success();
```
mikeurbach, line 1666: super nit: just like you can use `failure()` above, you could use `success()` without the `llvm::` namespace.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -1628,31 +1643,50 @@ struct ObjectSubfieldOpConversion
...
+  for (auto field : fieldNames) {
+    std::optional<Type> type = oldOp.getFieldType(field);
+    if (!type.has_value())
+      return failure();
+    fieldTypes.push_back(mlir::NamedAttribute(
+        cast<StringAttr>(field),
+        mlir::TypeAttr::get(typeConverter->convertType(type.value()))));
+  }
+  auto *ctx = newOp.getContext();
+  newOp->setAttr("fieldNames", mlir::ArrayAttr::get(ctx, {fieldNames.begin(),
```
mikeurbach, line 1662: This is somewhat obscure, but MLIR has a somewhat nifty helper for generically replacing attributes/types in an operation: https://github.com/llvm/llvm-project/blob/1e70122cbc187c08de91a3fb42843efb1221e0e9/mlir/include/mlir/IR/AttrTypeSubElements.h#L122-L134

Here's an example in CIRCT: https://github.com/llvm/circt/blob/eef76ca7e302ccef144343926c4e653a0e43a558/lib/Dialect/OM/Transforms/FreezePaths.cpp#L92-L100

You could potentially build an AttrTypeReplacer in the ClassFieldsOpConversion, which delegates to the typeConverter to do the replacements, and then call it's `replace` method on the newOp in matchAndRewrite. This point may be moot if we end up following the suggestion to put the field names and types on the ClassLike op, but thought I'd bring it up since it's a neat tool.

include/circt/Dialect/OM/OMOpInterfaces.td
```diff
@@ -15,6 +15,102 @@
...
+
+        printer << "(";
+        printer.increaseIndent();
+
+        llvm::SmallVector<mlir::StringAttr> fieldNames = $_op.getFieldNames();
+
+        mlir::ArrayRef<mlir::Attribute> perFieldAttrs;
```
mikeurbach, line 45: Maybe I'm missing something, but do we use attributes on each field? I know the original `om.class.field` has an arbitrary attributes dictionary, but that's because all operations have that, to hold any discardable attributes. I don't think we've actually used per-field attributes for anything, so it might be fine to remove. We do have per-port attributes on HW modules, but I don't know if we need to follow that here unless we have a use case.

mikeurbach (commented):
Just some passing comments in addition to what Hideto left. Overall, I think this looks great for the current design point--the changes in the various passes look nice and straightforward, nice work!

include/circt/Dialect/OM/OMOpInterfaces.td
leonardt: Yea, I think I was just following the HW code.  Figured it might be useful but none of the test code actually needed it.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -966,12 +1024,14 @@ void LowerClassesPass::lowerClass(om::ClassOp classOp, FModuleLike moduleLike,
...
       continue;
 
-    // Get the original port name, create a Field, and erase the propassign.
-    auto name = moduleLike.getPortName(outputPort.getArgNumber());
-    builder.create<ClassFieldOp>(op.getLoc(), name, op.getSrc());
+    locs.push_back(op.getLoc());
+    fieldValues.push_back(op.getSrc());
+
     op.erase();
   }
```
leonardt, line 1031: We now have to traverse the class definition and collect PropAssignOps twice, once to make the field/type dict, another here to make the fields terminator.  If we wanted to avoid this, we could try to collect the values for the field terminator earlier (when creating the containing class op) and just forward that info here to create the terminator op rather than doing a second traversal.

leonardt:
Updated the syntax to use a return type style, added two review comments on code areas that we might be able to improve further

leonardt:
Will need to rebase on main to make sure there's no merge issues

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
mikeurbach: > we could try to collect the values for the field terminator earlier (when creating the containing class op) and just forward that info here to create the terminator op rather than doing a second traversal

This might be worth it... we have tried to optimize this pass, since it can be quite expensive, so if you already know you're doing two passes, it might be worth making one pass. I think in general, we should take a look at the performance before and after this PR regardless, and it make sense to spend some time on optimizations like this.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -821,6 +821,61 @@ bool LowerClassesPass::shouldCreateClass(StringAttr modName) {
...
+
+    StringAttr name = moduleLike.getPortNameAttr(outputPort.getArgNumber());
+
+    fieldNames.push_back(name);
+    fieldTypes.push_back(
+        NamedAttribute(name, TypeAttr::get(op.getSrc().getType())));
+  }
+  return buildClassLike<om::ClassOp>(builder, moduleLike.getLoc(), name,
+                                     formalParamNames, fieldNames, fieldTypes);
+}
```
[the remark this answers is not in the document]
mikeurbach: I will think about it but this looks like how we factor the code for HW module versus external module... the differences are handled at the start of the function, and as soon as possible, we call a shared, templated helper.

include/circt/Dialect/OM/OMOpInterfaces.td
```diff
@@ -31,28 +31,21 @@ def ClassLike : OpInterface<"ClassLike"> {
       "mlir::StringAttr", "getSymNameAttrName", (ins)>,
     InterfaceMethod<"Get the class-like formal parameter names attribute",
       "mlir::ArrayAttr", "getFormalParamNames", (ins)>,
-    InterfaceMethod<"Get the class-like formal parameter names attribute name",
+    InterfaceMethod<"Get the class-like dictionary mapping field names to type",
       "mlir::StringAttr", "getFormalParamNamesAttrName", (ins)>,
+    InterfaceMethod<"Get the class-like formal parameter names attribute name",
```
mikeurbach, line 36: nit: I think this description might be copy-paste

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -1667,28 +1694,41 @@ struct ObjectOpConversion : public OpConversionPattern<om::ObjectOp> {
...
+                                         &result)))
+    return failure();
+
+  rewriter.modifyOpInPlace(classOp, [&]() {
+    mlir::AttrTypeReplacer replacer;
+    replacer.addReplacement([&](TypeAttr typeAttr) {
+      return mlir::TypeAttr::get(
+          typeConverter.convertType(typeAttr.getValue()));
+    });
+    classOp.replaceFieldTypes(replacer);
```
mikeurbach, line 1719: Nice use of the AttrTypeReplacer!

mikeurbach (commented):
I took a pass through the latest revision and it's looking really good. I will go through and add a more detailed review. @uenoku please take a look as well.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
leonardt: Looking at `LowerClassesPass::runOnOperation` I think it should be straightforward to pass the information through so I'll take a look at it.

leonardt: I was able to "fuse" the loop with the previous loop, so I think this PR includes net the same amount of traversals now.  See https://github.com/llvm/circt/pull/7537/commits/945e497518526065748e17533dd1342b5ef3d791

We could further optimize this by traversing the ops once, collecting the relevant information but delaying construction of the class body until afterwards.  The main issue is that we need to traverse to construct the fieldTypes map before traversing to collect the class body.  I'd imagine the performance improvements here would be marginal if any (or worse) since we'd essentially need to do a similar complexity traversal of the stored information.  So I feel we can hold off on that until we do some profiling and decide it's a worthy place to improve

mikeurbach: Makes sense, that commit looks good.

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -945,28 +992,23 @@ void LowerClassesPass::lowerClass(om::ClassOp classOp, FModuleLike moduleLike,
     if (!needsClone && !propertyOperands && !propertyResults)
       continue;
 
-    // Actually clone the op over to the OM Class.
-    builder.clone(op, mapping);
+    if (isa<PropAssignOp>(op) &&
+        dyn_cast<BlockArgument>(cast<PropAssignOp>(op).getDest())) {
```
mikeurbach, line 996: nit: if you're just doing a boolean check, I think you could do `isa<BlockArgument>(...)`. `dyn_cast` is also useful for the first part of this predicate...

```
if (auto propAssign = dyn_cast<PropAssignOp>(op) &&
    isa<BlockArgument>(propAssign.getDest()))
    ...
```

I can't find the reference, but I think something in the [LLVM coding standards](https://llvm.org/docs/CodingStandards.html) discourages using the `propAssign` within the same if condition, so maybe this would be better:

```
if (auto propAssign = dyn_cast<PropAssignOp>(op))
    if (isa<BlockArgument>(propAssign.getDest()))
        ...
```

mikeurbach (approved):
I took another look through this. No major comments. Let's see how it looks on some large designs but after that I think it should be good to merge.

uenoku (approved):
Oh, sorry. I have missed the review request. Thank you for addressing comments! The new representation does look good to me!

leonardt:
CC @mikeurbach @uenoku  if I might get a final review on this change that was added after the initial approvals: https://github.com/llvm/circt/pull/7537/commits/85e6d6adc58812ee66b6b1e2a7fdc3f98bb219d6

Basically: FusedLoc will "merge" identical locations (so they are stored as a set).  This is problematic when we're trying to lookup per-field locations by using the fused location of the fields op.  This changes it so that the `metadata` field of `FusedLoc` is used to store the original location array so that locations are not merged and per field locations can be looked up by index.   I added some assertions in hopes that it might catch cases when this API/invariant is not obeyed (e.g. by someone constructing a fields op directly and adding a fused loc without the metadata attribute).  There likely might be a better way to enforce this invariant, or at least some more assertions we could include to catch cases when this invariant is broken.

mikeurbach:
Yep, I looked through that and it makes sense to me. I think the assertions are good--that's what they're for: https://llvm.org/docs/CodingStandards.html#assert-liberally. If user input could lead to the invariant being violated, that's an error, but if only compiler authors could lead to the invariant being violated, that's a prime candidate for assertions.

PR #7540, opened 2024-08-21
[MooreToCore] Ignore ConstantLike values in wait op

fabianschuiki:
Skip values defined by `ConstantLike` ops when collecting the list of values to observe in `llhd.wait` ops. Constants will never cause an `llhd.wait` to resume execution since they never change value.


maerhart (approved):
Thanks for cleaning this up! We could also add this as a folder of `llhd.wait`.

fabianschuiki:
Good idea!

PR #7541, opened 2024-08-22
[FIRRTL] Set output dirs on annotated blackboxes

### LINKED ISSUE #7538: [FIRRTL] Blackboxes, Directories, and Layers not working as expected
We may be missing some logic to move blackboxes into layer directories if they are only instantiated by that layer. I was seeing a blackbox not show up where I expected it to (in a layer) and golfed this down to the following example:

``` firrtl
FIRRTL version 4.0.0
circuit Foo: %[[
  {
    "class": "firrtl.transforms.BlackBoxInlineAnno",
    "target": "~Foo|Bar",
    "name": "Bar.sv",
    "text": "module Bar();\nendmodule"
  }
]]
  layer A, bind, "A":
  extmodule Bar:
    input a: UInt<1>

  public module Foo:
    input a: UInt<1>

    layerblock A:
      inst bar of Bar
      connect bar.a, a
```

This currently compiles to:

``` verilog
// Generated by CIRCT firtool-1.82.0-11-g89749efd8
// external module Bar

module Foo(
  input a
);

endmodule


// ----- 8< ----- FILE "A/layers_Foo_A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-11-g89749efd8
`ifndef layers_Foo_A
`define layers_Foo_A
bind Foo Foo_A a_0 (
  .a (a)
);
`endif // layers_Foo_A

// ----- 8< ----- FILE "A/Foo_A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-11-g89749efd8
module Foo_A(
  input a
);

  Bar bar (
    .a (a)
  );
endmodule


// ----- 8< ----- FILE "./Bar.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-11-g89749efd8
module Bar();
endmodule
// ----- 8< ----- FILE "firrtl_black_box_resource_files.f" ----- 8< -----

./Bar.sv

```

I see `Bar` in `./Bar.sv` and I'd expect to see it in `./A/Bar.sv`.

seldridge:
Change the `AssignOutputDirs` pass to add `hw::output_file` attributes on
external modules in the same way that these attributes are added to normal
modules.  While this has no effect on an external module which has no
implementation, this fixes a bug where the external module _does_ have an
implementation that the compiler will later resolve.

This is specifically done to make output directories compose correctly
with Chisel inline blackboxes where the body of the external module is in
an annotation and is supposed to be written to a file.

Reorder the FIRRTL pass pipeline so that Chisel blackboxes (represented by
a `firrtl.extmodule` and one of two blackbox annotations) will be assigned
an output directory based on their instantiation location.

Fixes #7538.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
@@ -54,47 +54,26 @@ using namespace firrtl;
 
 namespace {
 
-/// This is used to indicate the directory priority.  Multiple external modules
-/// with the same "defname" may have different output filenames.  This is used
-/// to choose the best filename.
-enum class Priority { TargetDir = 0, Verification, Explicit, TestBench, Unset };
-
 /// Data extracted from BlackBoxInlineAnno or BlackBoxPathAnno.
 struct AnnotationInfo {
   /// The name of the file that should be created for this BlackBox.
   StringAttr name;
-  /// The outputdirectory information for this extmodule.
+  /// The output directory information for this extmodule.
   hw::OutputFileAttr outputFileAttr;
   /// The body of the BlackBox.  (This should be Verilog text.)
   StringAttr inlineText;
-  /// The priority of this annotation.  In the even that multiple annotations
-  /// are provided for the same BlackBox, then use this as a tie-breaker if an
-  /// external module is instantiated multiple times and those multiple
-  /// instantiations disagree on where the module should go.
-  Priority priority = Priority::Unset;
 
 #if !defined(NDEBUG)
   /// Pretty print the AnnotationInfo in a YAML-esque format.
   void print(raw_ostream &os, unsigned indent = 0) const {
-    if (priority == Priority::Unset) {
-      os << "<null>\n";
-      return;
-    }
     os << llvm::formatv("name: {1}\n"
                         "{0}outputFile: {2}\n"
-                        "{0}priority: {3}\n"
-                        "{0}exclude: {4}\n",
+                        "{0}exclude: {3}\n",
                         llvm::fmt_pad("", indent, 0), name,
-                        outputFileAttr.getFilename(), (unsigned)priority,
+                        outputFileAttr.getFilename(),
                         outputFileAttr.getExcludeFromFilelist().getValue());
   };
 #endif
-};
-
-/// Collects the attributes representing an output file.
-struct OutputFileInfo {
-  hw::OutputFileAttr outputFileAttr;
-  Priority priority;
 };
 
 struct BlackBoxReaderPass
@@ -103,8 +82,8 @@ struct BlackBoxReaderPass
   bool runOnAnnotation(Operation *op, Annotation anno, OpBuilder &builder,
                        bool isCover, AnnotationInfo &annotationInfo);
   StringAttr loadFile(Operation *op, StringRef inputPath, OpBuilder &builder);
-  OutputFileInfo getOutputFile(Operation *origOp, StringAttr fileNameAttr,
-                               bool isCover = false);
+  hw::OutputFileAttr getOutputFile(Operation *origOp, StringAttr fileNameAttr,
+                                   bool isCover = false);
   // Check if module or any of its parents in the InstanceGraph is a DUT.
   bool isDut(Operation *module);
 
@@ -264,8 +243,9 @@ void BlackBoxReaderPass::runOnOperation() {
       // two blackbox paths.  This is the same logic used in `AssignOutputDirs`.
       // However, this needs to incorporate filenames that are only available
       // _after_ output directories are assigned.
-      auto *ptr = emittedFileMap.find(annotationInfo.name);
-      if (ptr == emittedFileMap.end()) {
+      auto [ptr, inserted] =
+          emittedFileMap.try_emplace(annotationInfo.name, annotationInfo);
+      if (inserted) {
         emittedFileMap[annotationInfo.name] = annotationInfo;
       } else {
         auto &fileAttr = ptr->second.outputFileAttr;
@@ -360,11 +340,9 @@ bool BlackBoxReaderPass::runOnAnnotation(Operation *op, Annotation anno,
       return true;
     }
 
-    auto outputFile = getOutputFile(op, name, isCover);
-    annotationInfo.outputFileAttr = outputFile.outputFileAttr;
+    annotationInfo.outputFileAttr = getOutputFile(op, name, isCover);
     annotationInfo.name = name;
     annotationInfo.inlineText = text;
-    annotationInfo.priority = outputFile.priority;
     return true;
   }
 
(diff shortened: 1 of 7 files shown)


dtzSiFive:
I noticed the generated file list for black boxes contains these:

```
// ----- 8< ----- FILE "firrtl_black_box_resource_files.f" ----- 8< -----

A/Bar.sv
```

This seems functionally correct, and I don't know what these are used for, but this does suggest challenges re:deleting `A/` and expecting the resulting design to be usable without modification.

These annotations and similar cannot go on parameterized modules, right (re:output dir attribute being different on different extmodule's that are really parameterizations of same (via defname))?

seldridge:
@dtzSiFive wrote:
> This seems functionally correct, and I don't know what these are used for, but this does suggest challenges re:deleting A/ and expecting the resulting design to be usable without modification.

A GitHub search acts like we don't have users of this internally. Externally, ChiselSim will use this file to know what extra files to copy to an input source area and will include these in compilation: https://github.com/chipsalliance/chisel/blob/5c129120ba06229e6c5d28c8120b35166770fa09/src/main/scala/chisel3/simulator/package.scala#L164

Practically, I think if the blackboxes move, so long as the blackbox filelist is updated to reflect these paths relative to the same directory as before, things will continue to work.

That said, the whole filelist approach is likely bad as it is, pedantically, undefined as to what should be included and doesn't work well with the FIRRTL ABI. It is likely saner to tell backend tooling to not use either of these filelists and instead rely on a glob include from the output directory. I.e., most tools support running them like: `-I build/ --top-module Foo`. (This will read all files and then elaborate from the specified top module.)

@dtzSiFive wrote:
> These annotations and similar cannot go on parameterized modules, right (re:output dir attribute being different on different extmodule's that are really parameterizations of same (via defname))?

They can, yes. ExtModules cannot capture parameterization, only the arguments used to parameterize. Therefore, if you want to collapse multiple ExtModules down to a single parametric Verilog blackbox, you need to put these annotations on all of the ExtModules.

dtzSiFive:
> They can, yes. ExtModules cannot capture parameterization, only the arguments used to parameterize. Therefore, if you want to collapse multiple ExtModules down to a single parametric Verilog blackbox, you need to put these annotations on all of the ExtModules.

Okay, thanks.  Probably you can see where I was going with that -- I was concerned about different parameterizations getting assigned different directories (and what happens if that's true) which seems like a problem as the following demonstrates:

```firrtl
FIRRTL version 4.0.0

circuit Foo: %[[
  {
    "class": "firrtl.transforms.BlackBoxInlineAnno",
    "target": "~Foo|Bar",
    "name": "Bar.sv",
    "text": "Bar definition that takes parameters"
  },
  {
    "class": "firrtl.transforms.BlackBoxInlineAnno",
    "target": "~Foo|Baz",
    "name": "Bar.sv",
    "text": "Bar definition that takes parameters"
  }
]]
  layer A, bind, "A":
  layer B, bind, "B":

  extmodule Baz:
    input a: UInt<1>
    defname = Bar
    parameter x = 1

  extmodule Bar:
    input a: UInt<1>
    defname = Bar
    parameter x = 0


  public module Foo:
    input a: UInt<1>

    layerblock B:
      inst baz of Baz
      connect baz.a, a

    layerblock A:
      inst bar of Bar
      connect bar.a, a
```

Which with this PR yields:

```systemverilog
// Generated by CIRCT firtool-1.82.0-18-g1ceee216b
// external module Bar

// external module Bar

module Foo(
  input a
);

endmodule


// ----- 8< ----- FILE "B/layers_Foo_B.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-18-g1ceee216b
`ifndef layers_Foo_B
`define layers_Foo_B
bind Foo Foo_B b (
  .a (a)
);
`endif // layers_Foo_B

// ----- 8< ----- FILE "A/layers_Foo_A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-18-g1ceee216b
`ifndef layers_Foo_A
`define layers_Foo_A
bind Foo Foo_A a_0 (
  .a (a)
);
`endif // layers_Foo_A

// ----- 8< ----- FILE "B/Foo_B.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-18-g1ceee216b
module Foo_B(
  input a
);

  Bar #(
    .x(1)
  ) baz (
    .a (a)
  );
endmodule


// ----- 8< ----- FILE "A/Foo_A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-18-g1ceee216b
module Foo_A(
  input a
);

  Bar #(
    .x(0)
  ) bar (
    .a (a)
  );
endmodule


// ----- 8< ----- FILE "B/Bar.sv" ----- 8< -----

// Generated by CIRCT firtool-1.82.0-18-g1ceee216b
Bar definition that takes parameters
// ----- 8< ----- FILE "firrtl_black_box_resource_files.f" ----- 8< -----

B/Bar.sv
```

seldridge:
Nice test case. Yes, this doesn't look right. It seems like this should go into the common parent, i.e., this should create just `./Bar.sv` as that is the LCA of `A/Bar.sv` and `B/Bar.sv`.

lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
```diff
@@ -53,59 +54,36 @@ using namespace firrtl;
...
-/// to choose the best filename.
-enum class Priority { TargetDir = 0, Verification, Explicit, TestBench, Unset };
-
 /// Data extracted from BlackBoxInlineAnno or BlackBoxPathAnno.
 struct AnnotationInfo {
   /// The name of the file that should be created for this BlackBox.
   StringAttr name;
-  /// The output directory where this annotation should be written.
-  StringAttr outputFile;
+  /// The outputdirectory information for this extmodule.
```
dtzSiFive, line 61:
```suggestion
  /// The output directory information for this extmodule.
```

lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
```diff
@@ -260,10 +238,23 @@ void BlackBoxReaderPass::runOnOperation() {
...
-      auto &bestAnnotationInfo = emittedFileMap[annotationInfo.name];
-      if (annotationInfo.priority < bestAnnotationInfo.priority) {
-        bestAnnotationInfo = annotationInfo;
-
+      // If we have seen a black box trying to create a blackbox with this
+      // filename before, then compute the lowest commmon ancestor between the
+      // two blackbox paths.  This is the same logic used in `AssignOutputDirs`.
+      // However, this needs to incorporate filenames that are only available
+      // _after_ output directories are assigned.
+      auto *ptr = emittedFileMap.find(annotationInfo.name);
```
dtzSiFive, line 246: nit: doing a find followed by insert-if-not-there can be collapsed into something like `try_emplace`?
Probably not a big deal here but good habit to get into avoiding the double-lookup (hash or tree walk or however that's done).

dtzSiFive (commented):
Generally LGTM, some nits, and some pre-existing details I might want to chase down but nothing blocking.

Thanks for handling the extmodule prefix issue discussed!

seldridge:
The @dtzSiFive test does now work. I've tested this complete PR on a number of internal designs and observed no change in output directories.

lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
seldridge: Fixed in: https://github.com/llvm/circt/pull/7541/commits/836291f8d0bb35068f1719634578a26b962a9902

lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
seldridge: :+1: Fixed in: https://github.com/llvm/circt/pull/7541/commits/61a594cbcd44eed7df834d0f9d3dc86bcaf62486

PR #7547, opened 2024-08-23
[firrtl] Add CheckLayers diagnostics pass

youngar:
This adds a pass to check for illegal instantiation of a module with
layers underneath a layer.  This is a situation that leads to
bind-unde-bind when emitted to Verilog, which is illegal.  This does not
differentiate between inline layers and bind layers, although
theoretically this is only a problem for bind layers. This does not
create errors on extmodules, as we have no way of knowing whether they
contain a layer or not, and we don't want false positives.


dtzSiFive (approved):
Nice work!! LGTM! 🎉

PR #7548, opened 2024-08-23
Advanced LayerSink

rwy7:
Add advanced layer sink, a new FIRRTL-dialect transform pass. Advanced layer sink serves the same purpose as the original layer sink pass: move ops into layers if their uses permits it. Advanced layer sink works by finding the shallowest block a value is used by an effectful op, and then sinking the defining op to that block. Compared to the original layer sink pass, which only relied on dominance info, this pass can sink cyclical structures, such as registers.

### DIFF
--- a/test/Dialect/FIRRTL/ref.mlir
+++ b/test/Dialect/FIRRTL/ref.mlir
@@ -1,5 +1,5 @@
 // RUN: circt-opt %s -split-input-file | circt-opt -split-input-file
-// RUN: firtool %s -split-input-file
+// RUN: firtool %s -preserve-values=named -split-input-file
 // These tests are just for demonstrating RefOps, and expected to not error.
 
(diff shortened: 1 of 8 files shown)


test/Dialect/FIRRTL/ref.mlir
```diff
@@ -1,5 +1,5 @@
 // RUN: circt-opt %s -split-input-file | circt-opt -split-input-file
-// RUN: firtool %s -split-input-file
+// RUN: firtool %s -preserve-values=named -split-input-file
```
rwy7, line 2: Using `interesting_name` and `preserve-values=named` together to prevent layer-sink from deleting instances or wires that are necessary for the test.

youngar (approved):

dtzSiFive:
🥳 🎉

PR #7549, opened 2024-08-23
[FreezePaths] Add an optional argument, to get the operation name

prithayan:
`FreezePaths` pass may sometimes need to be invoked before `ExportVerilog`. 
This change removes the dependency of the pass on the `hw.verilogName` attribute, and adds an optional get operation name function, that can be invoked if the verilog name is absent. 
This enables any tool to invoke it earlier by providing an appropriate name-getter function. 
But as is expected, the names used to freeze the paths, may not match the verilog names. This should be used with caution, only when the path will not be used and its okay to be incorrect.

### DIFF
--- a/lib/Dialect/OM/Transforms/FreezePaths.cpp
+++ b/lib/Dialect/OM/Transforms/FreezePaths.cpp
@@ -32,7 +32,7 @@ namespace {
 struct PathVisitor {
   PathVisitor(hw::InstanceGraph &instanceGraph, hw::InnerRefNamespace &irn,
               std::function<StringAttr(Operation *)> &getOpName)
-      : instanceGraph(instanceGraph), irn(irn), getOpName(getOpName) {}
+      : instanceGraph(instanceGraph), irn(irn), getOpNameFallback(getOpName) {}
 
   StringAttr field;
   LogicalResult processPath(Location loc, hw::HierPathOp hierPathOp,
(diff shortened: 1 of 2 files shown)


lib/Dialect/OM/Transforms/FreezePaths.cpp
```diff
@@ -118,6 +120,8 @@ LogicalResult PathVisitor::processPath(Location loc, hw::HierPathOp hierPathOp,
     auto *op = target.getOp();
     // Get the verilog name of the target.
     auto verilogName = op->getAttrOfType<StringAttr>("hw.verilogName");
+    if (!verilogName && getOpName)
+      verilogName = getOpName(op);
```
mikeurbach, line 124: My only nitpick is this is only invoked as a fallback when hw.verilogName is not set, and calling it getOpName seems to me like this is the main name-getting function, not a fallback. Do with this feedback what you will... I think this is fine, but I might call it something like `getOpNameFallback` or something like that.

mikeurbach (approved):
This extension point makes sense to me for tools that want to run this pass but might not export verilog. Thanks!

prithayan:
> I think FreezePaths relied on `hw.verilogName` because it needed correct verilog names. How does your callback ensure names are same to verilog names?

You are right, the verilog name might not match, but this is required for certain flows when we are certain that the path and the names do not matter/would not be used. But other (non-path)parts of the OMIR are needed and the freeze-path pass is required to implement the OFFRITL ABI.
So, in our case, the MLIR is generated by a tool that does not run the ExportVerilog but expects the FreezePaths to run, so we wanted remove this dependency by providing a fallback operation name getter, without which FreezePaths would fail.

uenoku (approved):

PR #7550, opened 2024-08-24
Bump LLVM to 10407be542aeb2b59477b167bbba3716538dc722.

mikeurbach:
Standard LLVM version update.


mikeurbach:
Opening this now, because I want to get some eyes on it. This is mostly a benign bump, but one thing has merged upstream that affects SystemC: https://github.com/llvm/llvm-project/pull/91475. More discussion here: https://discourse.llvm.org/t/psa-modelling-memory-of-emitc-variables/80556.

Concretely, the only thing I saw that effects our tests is this: https://github.com/llvm/circt/blob/6743073c377b8b2467ba7fa50cb7920a8404820b/test/Target/ExportSystemC/basic.mlir#L153

emitc.apply "&" now requires an lvalue type for its operand.

I'm curious what @maerhart thinks... it's been a while since I looked at the SystemC dialect, but it looks like we have our own ops for modeling lvalues (systemc.cpp.variable, systemc.cpp.assign, etc.). I'm not sure how far we want to go to down the road of compatibility with EmitC dialect's lvalues. Perhaps systemc.cpp.variable's result type could be an lvalue type, and the destination of systemc.cpp.assign could be an lvalue type. Or we could start using the upstream EmitC ops instead. I took a brief look at the former, but decided it would be best to let the experts weigh in so I've sent this PR as-is. The in-tree CI should pass, but I'm not sure if there are deeper things to consider.

FYI I will now be on vacation for a week, so please feel free to pick this up while I'm out or chat with @dtzSiFive who has offered to help with LLVM bumps.

maerhart:
Thanks @mikeurbach
I think making the result of `systemc.cpp.variable` an lvalue is a good solution.
We might want to revisit the systemc dialect at some point because a lot has changed over the last year in the upstream emitc dialect and we can probably reuse a lot more of their ops, but that's a topic for another PR. E.g., the variable op upstream is still slightly different in that it only accepts an attribute as the init value, so replacing it might be a bit tricky.

I can take care of this on Friday if nobody else has picked it up until then.

dtzSiFive:
> I can take care of this on Friday if nobody else has picked it up until then.

That would be great, thank you!

uenoku (approved):
LGTM, thank you for updating initial op lowering. Interesting that unrealized_conversion cast is removed even whey cast is not introduced by dialect conversion.

Issue #7551, opened 2024-08-24
[FIRRTL] Poor Asymptotic Performance of Layer Merge and Layer Sink


seldridge:
There are scaling problems with both `LayerMerge` and `LayerSink`. You can tickle the bad behavior here by generating circuits which have large numbers of layerblocks which need to be merged and have lots of operations that need to be sunk.

You can generate such circuits with the following scala-cli snippet:

``` scala
//> using scala "2.13.12"
//> using repository sonatype-s01:snapshots
//> using lib "org.chipsalliance::chisel::7.0.0-M2+59-5c129120-SNAPSHOT"
//> using plugin "org.chipsalliance:::chisel-plugin::7.0.0-M2+59-5c129120-SNAPSHOT"
//> using options "-unchecked", "-deprecation", "-language:reflectiveCalls", "-feature", "-Xcheckinit", "-Xfatal-warnings", "-Ywarn-dead-code", "-Ywarn-unused", "-Ymacro-annotations"

import chisel3._
import circt.stage.ChiselStage

class Foo(size: Int) extends Module {
  val a = Seq.fill(size)(IO(Input(Bool())))
  val b = Seq.fill(size)(IO(Input(Bool())))

  a.zip(b).foreach{ case (a, b) => assert(a ^ b) }
}

object Main extends App {
  val gen = () => new Foo(1024 * 16)

  println(
    ChiselStage.emitCHIRRTL(
      gen()
    )
  )

}
```

I'm seeing the following runtimes in a release build. Times are for layer merge followed by layer sink:

- 16K assertions: 1s, 0.5s
- 32K assertions: 3s, 1s
- 64K assertions: 15s, 7s
- 128k assertions: 148s, 40s

PR #7552, opened 2024-08-24
[FIRRTL] Improve Layer Merge Performance

seldridge:
Fix a performance issue in the `LayerMerge` pass [[1]].  Apparently, repeated use of `inlineBlockBefore` can be a performance issue.  Fix this, by using `mergeBlocks` which will append at the end of a block as opposed to prepend.  This requires an additional move, however, this empirically doesn't matter.

For details of the algorithm, see the updated comments in the pass.

h/t @youngar for algorithmic suggestions.

[1]: https://github.com/llvm/circt/issues/7551

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LayerMerge.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LayerMerge.cpp
@@ -44,7 +44,7 @@ void LayerMerge::runOnOperation() {
                    << "Module: '" << getOperation().getName() << "'\n";);
 
   // Track the last layer block in a module associated with a specific layer.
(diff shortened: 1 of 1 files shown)


youngar:
To me it might be a bit more natural to do this as a backwards walk and shift everything downward into the first found block, slightly because you never violate the use before def when modifying the code

youngar (approved):
🚀

seldridge:
> To me it might be a bit more natural to do this as a backwards walk and shift everything downward into the first found block, slightly because you never violate the use before def when modifying the code

I switched to this. Interestingly, this still uses `inlineBlockBefore`. However, it doesn't have performance issues in my local stress tests. I'm going to make sure that this fixes things on the internal design.

Algorithm is much cleaner now. Nice suggestion.

Issue #7213, opened 2024-06-19
[Seq] FirReg is not appropriate for the seq dialect


teqdruid:
Having an op called "FirReg" in the seq dialect screams abstraction leakage. Figure out what to do about this. If it's appropriate for a general FF, it should get renamed.

PR #7553, opened 2024-08-26
[Seq] Introduce seq.initial, !seq.immutable and replace powerOn value with initial value in compreg

uenoku:
This PR extends compreg's powerOnValue operand to be able to capture more complicated initialization such as firreg's randomized initialization or DPI calls. This change should make register initialization more modular and a step forward towards https://github.com/llvm/circt/issues/7213. 

While ASICs might not have explicit initial values, they are crucial for simulation
and FPGA implementation. FPGA designs often require constant initial values, while
simulation allows for more complex initialization using expressions like function calls
`$random`, `$readmem`, and `$fopen`.

seq.compreg has a (optional) powerOn operand that is lowered into inlined assignment in SV which allows users to initialize registers with user-specified values. However this representation is not sufficient for initialization with function calls. 

In order to represent various kinds of initialization, `seq.initial` op and `!seq.immutable` type
are introduced. The `seq.initial` operation produces values with types wrapped in `!seq.immutable`.
The `!seq.immutable` type wrapper prevents initial values from depending on time-variant values.
Stateful operations typically require corresponding initial values with the `!seq.immutable` type.
This ensures that the initial state of the operation is well-defined and independent of time-variant factors.

Example Input:
```mlir
%r_init, %u_init = seq.initial {
  %rand = sv.macro.ref.se @RANDOM() : () -> i32
  %c0_i32 = hw.constant 0 : i32
  seq.yield %rand, %c0_i32 : i32, i32
} : !seq.immutable<i32>, !seq.immutable<i32>
%r = seq.compreg %i, %clk initial %r_init : i32
%u = seq.compreg %i, %clk initial %u_init : i32
```

Output Verilog:
```verilog
reg [31:0] r;
initial
  r = `RANDOM;
reg [31:0] u = 32'h0;
```

### DIFF
--- a/lib/Dialect/Seq/SeqOps.cpp
+++ b/lib/Dialect/Seq/SeqOps.cpp
@@ -12,6 +12,7 @@
 
 #include "circt/Dialect/Seq/SeqOps.h"
 #include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/Sim/SimTypes.h"
 #include "circt/Support/CustomDirectiveImpl.h"
 #include "circt/Support/FoldUtils.h"
 #include "mlir/IR/Builders.h"
@@ -81,6 +82,20 @@ parseOptionalTypeMatch(OpAsmParser &parser, Type refType,
 
 static void printOptionalTypeMatch(OpAsmPrinter &p, Operation *op, Type refType,
                                    Value operand, Type type) {
+  // Nothing to do - this is strictly an implicit parsing helper.
+}
+
+static ParseResult parseOptionalImmutableTypeMatch(
+    OpAsmParser &parser, Type refType,
+    std::optional<OpAsmParser::UnresolvedOperand> operand, Type &type) {
+  if (operand)
+    type = seq::ImmutableType::get(refType);
+  return success();
+}
+
+static void printOptionalImmutableTypeMatch(OpAsmPrinter &p, Operation *op,
+                                            Type refType, Value operand,
+                                            Type type) {
   // Nothing to do - this is strictly an implicit parsing helper.
 }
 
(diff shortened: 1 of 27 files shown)


lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -1004,6 +1019,63 @@ FirMemory::FirMemory(hw::HWModuleGeneratedOp op) {
   initIsInline = op->getAttrOfType<BoolAttr>("initIsInline").getValue();
 }
 
+LogicalResult InitialOp::verify() {
+  auto *terminator = this->getBody().front().getTerminator();
+  if (terminator->getOperands().size() != getNumResults())
+    return emitError() << "result type doesn't match with the terminator";
```
mortbopet, line 1025: Missing test

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -1004,6 +1019,63 @@ FirMemory::FirMemory(hw::HWModuleGeneratedOp op) {
...
+  for (auto [lhs, rhs] :
+       llvm::zip(terminator->getOperands().getTypes(), getResultTypes())) {
+    if (cast<seq::ImmutableType>(rhs).getInnerType() != lhs)
+      return emitError() << cast<seq::ImmutableType>(rhs).getInnerType()
+                         << " is expected but got " << lhs;
```
mortbopet, line 1030: Missing test

mortbopet (approved):

lib/Dialect/Seq/SeqOps.cpp
uenoku: Added, thanks!

fabianschuiki (approved):
I like the idea of having a separate set of operations define the initial value for registers, instead of trying to add regions to the registers to compute initial values. The separate `seq.initial` can have its results used multiple times, and there is a potential extension where `seq.initial` could accept `!seq.immutable<T>` operands that are available inside the body as `T` block arguments.

Will it be possible to represent the `FirReg` randomization using `seq.initial` in the future, and possibly to combine `FirReg` and `CompReg` into a `Reg`?

uenoku:
> there is a potential extension where seq.initial could accept !seq.immutable<T> operands

Yes. This requires shceduling of initial operations while lowering but it's feasible to do that. 

> Will it be possible to represent the FirReg randomization using seq.initial in the future, and possibly to combine FirReg and CompReg into a Reg?

Yes, that's definitely the goal.

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -1004,6 +1019,63 @@ FirMemory::FirMemory(hw::HWModuleGeneratedOp op) {
   initIsInline = op->getAttrOfType<BoolAttr>("initIsInline").getValue();
 }
 
+LogicalResult InitialOp::verify() {
+  auto *terminator = this->getBody().front().getTerminator();
```
prithayan, line 1023: Do we need to verify, this is a `YieldOp` ?

prithayan (approved):
I think this is a great idea, looks good to me.

lib/Dialect/Seq/SeqOps.cpp
uenoku: It's verified/ensured by `SingleBlockImplicitTerminator<YieldOp>`

PR #7555, opened 2024-08-26
[ESI] Add optional non-blocking write API to `WriteChannelPort`

mortbopet:


teqdruid:
Could you implement this for cosim and trace (and add tests as well)?

Also, is there a reason this should be optional?

mortbopet:
> Could you implement this for cosim and trace (and add tests as well)?
> Also, is there a reason this should be optional?

If we want everything to implement this, I don't think we can call it "non-blocking".
Since neither cosim nor trace seems to have an analogue for a non-blocking write, we'd have to change the wording to indicate that it's a failable write... where the default implementation would just call `write`.
```c++
  /// A basic failable write API. Returns true if the data was written.
  virtual bool tryWrite(const MessageData &data) {
    write(data);
    return true;
  }
```

teqdruid:
> > Could you implement this for cosim and trace (and add tests as well)?
> > Also, is there a reason this should be optional?
> 
> If we want everything to implement this, I don't think we can call it "non-blocking". Since neither cosim nor trace seems to have an analogue for a non-blocking write, we'd have to change the wording to indicate that it's a failable write... 

"failable" isn't specific enough. The only reason that this method should "fail" is if it "fails" to write or write to a buffer instantaneously and the caller should try again later. So I think 'non-blocking' is the correct verbiage here. We should, however, specify that it is not valid for a backend to always return false.

As for cosim/trace having no analog... they have no notion of backpressure, which is probably a bad thing since somewhere there's gotta be an infinitely large buffer, which is clearly not realistic. We should probably implement a cosim backpressure mechanism to indicate that the simulated hardware is backpressuring. So yeah, they don't have an analog *yet*.

> where the default implementation would just call `write`.
> 
> ```c++
>   /// A basic failable write API. Returns true if the data was written.
>   virtual bool tryWrite(const MessageData &data) {
>     write(data);
>     return true;
>   }
> ```

`write()` can block, so that would be an invalid default implementation. It's fine for cosim/trace, but not as a default. Defaults shouldn't do something which is only valid for a subset of implementations. Since -- as I mentioned above -- we should implement so backpressure mechanism in the (near/midterm) future, I would just copy that implementation into both backends. For extra credit, implement a pseudorandom backpressure on the trace backend.

teqdruid (approved):
Perhaps add some tests calling 'tryWrite' so if/when we add some backpressure in the future we'll know if we break anything. May also be desired to add to the python binding as well if you want extra credit.

teqdruid:
Thanks!

Issue #7447, opened 2024-08-06
[FIRRTL] extmodule's supporting enablelayer?


dtzSiFive:
FIRRTL spec only covers this for modules, and use on extmodule's in untested and unverified.

The following parses and compiles fine:

```firrtl
FIRRTL version 4.0.0

circuit EL:
  extmodule X enablelayer A.B.C :
  public module EL :
```

Supporting this allows extmodule to describe a module with `enablelayers` such that it's instantiated appropriately.
Assuming we want this, spec should be updated and verifier added accordingly.

dtzSiFive:
Looks like other FModuleLike's support the `layers` attribute too, FWIW.

PR #7556, opened 2024-08-26
[FIRRTL] Verify enable/disable layer lists of circuit.

dtzSiFive:
* Verify specified symbols resolve
* Verify specified symbols resolve to LayerOp's
* Verify main module doesn't enable circuit-disabled layers. (SpecializeLayers will delete main if so)

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLOps.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOps.cpp
@@ -532,6 +532,37 @@ LogicalResult CircuitOp::verifyRegions() {
       mlir::SymbolTable::Visibility::Public)
     return mainModule->emitError("main module must be public");
 
+  // Check enable/disable layers arrays.
+  if (auto enabledLayers = getEnableLayersAttr())
+    for (auto enabledLayer : enabledLayers.getAsRange<SymbolRefAttr>()) {
+      auto *op = symtbl.lookupSymbolIn(*this, enabledLayer);
+      if (!op)
+        return emitOpError("unknown layer ")
+               << enabledLayer << " in enable layers list";
+      if (!isa<LayerOp>(op))
+        return emitOpError("expected layer for ")
+            .append(enabledLayer, " in enable layers list")
+            .attachNote(op->getLoc())
+            .append("not a layer");
+    }
+  if (auto disabledLayers = getDisableLayersAttr())
+    for (auto disabledLayer : disabledLayers.getAsRange<SymbolRefAttr>()) {
+      auto *op = symtbl.lookupSymbolIn(*this, disabledLayer);
+      if (!op)
+        return emitOpError("unknown layer ")
+               << disabledLayer << " in disable layers list";
+      if (!isa<LayerOp>(op))
+        return emitOpError("expected layer for ")
+            .append(disabledLayer, " in disable layers list")
+            .attachNote(op->getLoc())
+            .append("not a layer");
+      if (llvm::is_contained(cast<FModuleLike>(mainModule).getLayers(),
+                             disabledLayer))
+        return mainModule->emitOpError(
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -532,6 +532,37 @@ LogicalResult CircuitOp::verifyRegions() {
...
+        return emitOpError("expected layer for ")
+            .append(disabledLayer, " in disable layers list")
+            .attachNote(op->getLoc())
+            .append("not a layer");
+      if (llvm::is_contained(cast<FModuleLike>(mainModule).getLayers(),
+                             disabledLayer))
+        return mainModule->emitOpError(
+                   "main module has circuit-disabled layer ")
+               << disabledLayer << " enabled";
+    }
```
seldridge, line 564: Two questions (I'm not sure of the answers):

1. Is there a performance issue with the use of `lookupSymbolIn` here where this would compute the symbol table multiple times?
2. Are the enable/disable layer circuit options and enable layer module level option tracked as symbol users? If they are, then would this be more suited to symbol user verifier on the layer operations?

dtzSiFive: Thanks!

Regarding (1) -- looks like maybe yes, whoops.  Boooo (but thanks!).
Forgot that the efficient version is part of SymbolTableCollection via verifySymbolUses, I'll check if can leverage the SymbolUserOpInterface on the circuitOp itself or not (I'd guess not, but let's find out -- not w.r.t. symbols defined within itself anyway, but I'll check (for example I don't think a SymbolTable operation is itself considered a user of symbols within itself -- symbol references it has should resolve in parent SymbolTable).

If it's lazily built, maybe can be used here, since we already built the SymbolTable above so the extra steps are unlikely to be a problem re:nested.

Regarding (2), yes and we do that partially today.  I have PR in-flight for that (cc #7447), I'll poke at this a bit more.
That doesn't get us (1) however.

I'll submit that PR and update this one in a few, thanks!

youngar:
Do we have to worry about this having a good user-facing error messages? Will this dump the entire circuit?  We might be able to verify this during IR generation in the parser or something.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
youngar: Just making sure we're on the same page, switching `symtbl.lookupSymbolIn(*this, disabledLayer)` to `symtbl.lookup(disabledLayer)` should fix 1 if the other stuff doesn't work out.

youngar: I forgot that this is a nested symbol reference - so that doesn't actually work - woops

dtzSiFive:
> Do we have to worry about this having a good user-facing error messages? Will this dump the entire circuit? We might be able to verify this during IR generation in the parser or something.

We use `emitOpError` for the other circuit-level checks, and for consistency this does the same.  Determining a policy and implementing it consistently here or elsewhere seems like a good follow-up/cleanup task for anyone interested.

We could verify this during parsing, but we should also verify (as reasonable anyway) as part of IR sanity checking (IMO).

lib/Dialect/FIRRTL/FIRRTLOps.cpp
dtzSiFive: Verifying enable lists on modules, including checking for enable of circuit-disabled layers here: https://github.com/llvm/circt/pull/7557 .

dtzSiFive:
Thanks for the feedback, all!

Closing this for now, as I think this is a task for another day and the other PR covers the main check(s) I wanted to get online.

We can discuss/sort out how/if we want additional checking of these in the future, as needed.

PR #7557, opened 2024-08-26
[FIRRTL] Verify enabled layers of module, extmodule.

dtzSiFive:
* Verify specified symbols resolve for extmodule
* Verify specified symbols resolve to LayerOp's
* Verify public modules don't enable circuit-disabled layers. (SpecializeLayers will delete if so)


lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -1694,23 +1694,51 @@ static LogicalResult verifyPortSymbolUses(FModuleLike module,
...
 
+static LogicalResult verifyEnabledLayers(FModuleLike module,
+                                         SymbolTableCollection &symbolTable) {
+  auto circuitOp = module->getParentOfType<CircuitOp>();
+  for (auto layer : module.getLayers()) {
+    auto *op =
+        symbolTable.lookupSymbolIn(circuitOp, cast<SymbolRefAttr>(layer));
+    if (!op)
+      return module->emitOpError() << "enables unknown layer '" << layer << "'";
+    if (!isa<LayerOp>(op))
```
dtzSiFive, line 1705: This only checks the final reference is a layer, but structural requirements on layers ensure all parents will also be layerop's.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -1694,23 +1694,51 @@ static LogicalResult verifyPortSymbolUses(FModuleLike module,
...
+  }
+
+  // Check public modules don't enable circuit-disabled layers.
+  if (auto disabledLayers = circuitOp.getDisableLayersAttr()) {
+    auto disabledLayersSymRefs = disabledLayers.getAsRange<SymbolRefAttr>();
+    if (module.isPublic()) {
+      for (auto layer : module.getLayers())
+        if (llvm::is_contained(disabledLayersSymRefs, layer))
+          return module->emitOpError(
+                     "public module has circuit-disabled layer ")
```
dtzSiFive, line 1719: Is set membership sufficient, or do we need to walk and check if parent layer disabled....?

youngar: You're going to have to check if a parent layer is disabled

dtzSiFive: Great, thanks!

PR #7561, opened 2024-08-30
[ESI] Manifest: change the schema to be more rational

teqdruid:
Changes falling out of an audit conducted by Morten and myself.

### DIFF
--- a/include/circt/Dialect/ESI/ESIManifest.td
+++ b/include/circt/Dialect/ESI/ESIManifest.td
@@ -70,11 +70,11 @@ def ServiceRequestRecordOp : ESI_Op<"manifest.req", [
   let arguments = (ins AppIDAttr:$requestor,
                        InnerRefAttr:$servicePort,
                        OptionalAttr<StrAttr>:$stdService,
-                       TypeAttrOf<ChannelBundleType>:$bundleType);
+                       TypeAttrOf<ChannelBundleType>:$typeID);
 
   let assemblyFormat = [{
     qualified($requestor) `,` $servicePort (`std` $stdService^)?
-    `,` $bundleType attr-dict
+    `,` $typeID attr-dict
   }];
 
   let extraClassDeclaration = [{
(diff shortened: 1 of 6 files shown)


teqdruid:
@mortbopet I think these are all the changes we discussed with the exception of the `types` section. Changing it to make all of the "type" fields to field references would have introduced complexity to the runtime parser and (IIRC) the manifest output code. Yeah, it's still redundant but it's easier to parse and that's the primary goal.

mortbopet:
> Changing it to make all of the "type" fields to field references would have introduced complexity to the runtime parser and (IIRC) the manifest output code. Yeah, it's still redundant but it's easier to parse and that's the primary goal.

I still heavily disagree with this. One can just choose to parse the types section before anything else in the manifest, or one can resolve type references as a secondary step after the manifest has been parsed. Those are both easy steps to implement, so i don't think having inline types does that much, if anything, for making the manifest easier to parse.

With the way my brain works, having inline type definitions creates a large amount of visual clutter (i wouldn't be surprised if, for some manifests, *redundant) types take up more than half of the entire manifest). And this (for a human being, maybe also for a computer) makes the manifest **harder** to parse.

teqdruid:
> > Changing it to make all of the "type" fields to field references would have introduced complexity to the runtime parser and (IIRC) the manifest output code. Yeah, it's still redundant but it's easier to parse and that's the primary goal.
> 
> I still heavily disagree with this. One can just choose to parse the types section before anything else in the manifest, or one can resolve type references as a secondary step after the manifest has been parsed. Those are both easy steps to implement, so i don't think having inline types does that much, if anything, for making the manifest easier to parse.
> 
> With the way my brain works, having inline type definitions creates a large amount of visual clutter (i wouldn't be surprised if, for some manifests, *redundant) types take up more than half of the entire manifest). And this (for a human being, maybe also for a computer) makes the manifest **harder** to parse.

I already do parse the types section first. The problem is that that during parsing, a particular typeID many not (probably hasn't) been resolved to an actual type when it is encountered as used by another type. If it's inline, that's not a problem. If one has to wait until later, that's a problem and it increases the complexity of the parsing code.

Either way, this is strictly better than what we've got now.

Issue #7535, opened 2024-08-20
[MooreToCore] VariableOp lowered failed


mingzheTerapines:
Dear @maerhart @fabianschuiki ,
 When lowering `SV` to `Hw` Dialect, there is a stack dump. 
Driver: circt-verilog %s
```
module top();
  typedef struct {
    int a;
    int b;
  } ms_t;

  ms_t ms;

  initial begin
    ms = '{ 0, 1};

    ms = '{ default:1, int:1};

    ms = '{ int:0, int:1};
  end

endmodule
```
It can be converted to `moore` Dialect like this
Driver:  circt-verilog --ir-moore %s
```
module {
  moore.module @top() {
    %0 = moore.constant 1 : i32
    %1 = moore.constant 0 : i32
    %ms = moore.variable : <ustruct<{a: i32, b: i32}>>
    moore.procedure initial {
      %2 = moore.struct_create %1, %0 : !moore.i32, !moore.i32 -> ustruct<{a: i32, b: i32}>
      moore.blocking_assign %ms, %2 : ustruct<{a: i32, b: i32}>
      %3 = moore.struct_create %0, %0 : !moore.i32, !moore.i32 -> ustruct<{a: i32, b: i32}>
      moore.blocking_assign %ms, %3 : ustruct<{a: i32, b: i32}>
      moore.blocking_assign %ms, %3 : ustruct<{a: i32, b: i32}>
      moore.return
    }
    moore.output
  }
}
```
But it got stack dump when casting `hw::InOutType`. Maybe `structType` should be converted somehow.
This is part of error codes.
```
#19 0x00005f9660f42af4 (anonymous namespace)::OperationLegalizer::legalizeWithPattern(mlir::Operation*, mlir::ConversionPatternRewriter&) /home/pluto/Documents/circt/circt/llvm/mlir/lib/Transforms/Utils/DialectConversion.cpp:1958:21
#20 0x00005f9660f3b100 (anonymous namespace)::OperationLegalizer::legalize(mlir::Operation*, mlir::ConversionPatternRewriter&) /home/pluto/Documents/circt/circt/llvm/mlir/lib/Transforms/Utils/DialectConversion.cpp:1850:17
#21 0x00005f9660f3aa73 mlir::OperationConverter::convert(mlir::ConversionPatternRewriter&, mlir::Operation*) /home/pluto/Documents/circt/circt/llvm/mlir/lib/Transforms/Utils/DialectConversion.cpp:2384:26
[...]

PR #7565, opened 2024-08-31
[MooreToCore] Support to lower unpacked struct type

cepheus69:
This is a simple way to map unpackedStructType (Moore type) to structType in hw dialect directly.


cepheus69:
This pull request is an attempt to fix the problem: https://github.com/llvm/circt/issues/7535

mingzheTerapines:
Glad to see this resolved

hailongSun2000 (approved):
I agree with this solution.

PR #7567, opened 2024-08-31
[ESI][Runtime] Pretty printing of service ports

teqdruid:
esiquery now displays the function signatures instead of the consistuent channels. More representative of the API which is exposed.


teqdruid:
The runtime isn't checked by the CI gates.

PR #7568, opened 2024-08-31
[ESI][Runtime] Address MMIO regions symbolically

teqdruid:
Access MMIO address regions by the AppID of the requestor. Also provide access to the MMIO space descriptors.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
+++ b/lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
@@ -21,6 +21,7 @@
 #define ESI_RUNTIME_SERVICES_H
 
 #include "esi/Common.h"
+#include "esi/Context.h"
 #include "esi/Ports.h"
 
 #include <cstdint>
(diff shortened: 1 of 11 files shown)


lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
```diff
@@ -48,6 +49,18 @@ class Service {
 
   virtual std::string getServiceSymbol() const = 0;
 
+  /// Create a "child" service of this service. Used when a service already
+  /// exists in the active services table, but a new one wants to replace it.
```
mortbopet, line 53: I think this is a bit unclear... does 'a service' here imply that a 'child service' _of the same type_ wants to replace it? but then, below, it says that the child service needs to _use_ the parent service...

lib/Dialect/ESI/runtime/cpp/lib/Services.cpp
```diff
@@ -41,7 +49,89 @@ std::string SysInfo::getJsonManifest() const {
...
+    return nullptr;
+  return new MMIORegion(id.back(), conn.getService<MMIO>(), regionIter->second);
+}
+
+namespace {
+class MMIOPassThrough : public MMIO {
+public:
+  MMIOPassThrough(Context &ctxt, AppIDPath idPath, std::string implName,
+                  const ServiceImplDetails &details,
+                  const HWClientDetails &clients, MMIO *parent)
```
mortbopet, line 96: `parent` argument unused?

mortbopet (approved):

lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
teqdruid: It doesn't necessarily have to be -- it's entirely up to the `getChildService` method. I've clarified the code doc.

PR #7569, opened 2024-09-03
[ESI][Runtime] Logging API

teqdruid:
The ESI logging API is intended to be agnostic of the application's logging framework. In other words, it can be adapted to the application's logging method with ease -- users must only extend one class. It also comes with a simple text logger.

It is also designed to enable debug logging in Release builds with (very) minimal performance overhead if the user does not set the log level to debug.

This PR introduces the API, but doesn't use it much. More plumbing will be required to actually use it. In the interest of keeping this PR small, this is left for future work.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/include/esi/Logging.h
+++ b/lib/Dialect/ESI/runtime/cpp/include/esi/Logging.h
@@ -1,0 +1,160 @@
+//===- Logging.h - ESI Runtime logging --------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// ESI runtime logging is very simple but very flexible. Rather than mandating a
+// particular logging library, it allows users to hook into their existing
+// logging system by implementing a simple interface.
+//
+//===----------------------------------------------------------------------===//
+//
+// DO NOT EDIT!
+// This file is distributed as part of an ESI package. The source for this file
+// should always be modified within CIRCT.
+//
+//===----------------------------------------------------------------------===//
+
+// NOLINTNEXTLINE(llvm-header-guard)
+#ifndef ESI_LOGGING_H
+#define ESI_LOGGING_H
+
+#include <any>
+#include <functional>
+#include <iosfwd>
+#include <map>
+#include <memory>
+#include <string>
+
+namespace esi {
+
+class Logger {
+public:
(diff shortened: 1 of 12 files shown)


lib/Dialect/ESI/runtime/cpp/lib/Logging.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include "esi/Logging.h"
+#include "esi/Common.h"
+
+#include <iostream>
+#include <mutex>
+
+using namespace esi;
+
+// Necessary in some versions of the c++ standard library to avoid warnings
```
teqdruid, line 23: @mortbopet I think this is true, but I find it quite unusual... am I doing something wrong?

mortbopet: This is the first time i've seen this pattern, so 🤷.

lib/Dialect/ESI/runtime/cpp/include/esi/Logging.h
```diff
@@ -0,0 +1,160 @@
...
+  StreamLogger(Level minLevel);
+  void log(Level level, const std::string &subsystem, const std::string &msg,
+           const std::map<std::string, std::any> *details) override;
+
+private:
+  /// The minimum log level to emit.
+  Level minLevel;
+
+  /// Mutex to protect the stream from interleaved logging writes.
+  std::mutex mutex;
```
mortbopet, line 140: I think _most_ loggers will want to have mutually exclusive access to logging. How about providing a thread-safe `TSLogger` base class, which has all of the lock grabbing logic, before calling a logging implementation's `logImpl` function? (`StreamLogger::log -> StreamLogger->logImpl`).

mortbopet (commented):
I generally like the structure. However, I think it would be a nice-to-have (and possibly need-to-have) if we can register multiple loggers (somewhat as what one can do in spdlog). This PR wouldn't change that much, you'd just have to:
1. iterate through all registered loggers in the `log(...)` function
2. Not create the null logger by default

This is super useful in situations where a runtime wants to be in control of it's own (internal) logging implementation, but also need to expose a logging interface externally, to users of the runtime.

teqdruid:
> I generally like the structure. However, I think it would be a nice-to-have (and possibly need-to-have) if we can register multiple loggers (somewhat as what one can do in spdlog).

This would introduce some performance issues, which I'm ultra concerned about given that I want debug level logging to have the capability to log individual messages. Checking the `debugEnable` flag in an inline method is relatively cheap. Iterating through loggers isn't. So I think either way, I'd implement it as a "ListLogger" which iterates through subloggers. So can be future work.

teqdruid:
Merged (& will rely on post-merge comments) so I can cut releases for both PyCDE and ESIRuntime.

PR #7572, opened 2024-09-03
Bump LLVM

maerhart:
Currently fails because of https://github.com/llvm/llvm-project/pull/106778
@mikeurbach previously pointed out the issue in https://github.com/llvm/llvm-project/pull/101514 which wasn't reverted, thus exposing the issue again.


maerhart:
Waiting for https://github.com/llvm/llvm-project/pull/107109 to land

mikeurbach (approved):
Looks good to me, thanks for staying on top of this!

PR #7573, opened 2024-09-03
[MooreToCore] Separate conversion pattern for moore.output

maerhart:
Don't convert the output op in the module pattern because the operands will be of the wrong type since the body has not been converted yet, so it need to apply the hw output pattern afterwards anyway. Instead of relying on that pattern we should have a separate one for moore.output to hw.output that is applied once the rest of the body was converted. The HW output pattern is unnecessary because in the IR before conversion, no hw.output operation should be present (at least none that has a moore typed operand), thus its presence is only the consequence of a bad conversion process (if a target operation is inserted of which the operands are still of the source type, conversion casts should be inserted).


fabianschuiki (approved):
LGTM!

hailongSun2000:
Sounds Greate :partying_face:!

cepheus69:
It is a leftover problem since we start to support materialize value with different type. The original Moore module operation limits its implementation way (module without port definition and need net or variable op to denote port, and there is no output op). So it is great to see this remaining problem solved!

PR #7574, opened 2024-09-03
[FIRRTL][FIRParser] Check force source more.

dtzSiFive:
* Must be passive
* Explicitly diagnose if source/dest diff cannot be cast away.

Add tests.  Second crashed attempting to construct invalid RefType before this change (non-passive base type).

While visiting, preserve layer "color" when inserting ref.cast.

Found via fuzzing.


dtzSiFive:
Since fixes a crash, landing.  Post-merge review welcome!

PR #7575, opened 2024-09-03
[HW][HWTypes] Fix use of OptionalParseResult.

dtzSiFive:
Don't return success if there's a value, that value may be failure.  Fixes crash.

Emit diagnostic if no value found, to ensure something is produced.

Found via fuzzing.


uenoku (approved):
Nice find!

PR #7577, opened 2024-09-03
[FIRRTL][GrandCentral] Don't crash on missing keys, getAs.

dtzSiFive:
Add missing getAs return value checking, emit diagnostics.

Found via fuzzing.


seldridge (approved):
Nice catch. LGTM

PR #7578, opened 2024-09-03
[HW] Fix crash when error encountered parsing hw.array.

dtzSiFive:
Fix checking of OptionalParseResult, having value does not mean that the parsing succeeded.

Add test.

Found via fuzzing.


uenoku (approved):
LGTM

PR #7579, opened 2024-09-03
[FIRRTL][Lexer] Don't crash on trailing slash in inline anno.

dtzSiFive:
Add test.

Found via fuzzing.


seldridge (approved):
LGTM

Originally the fallthrough seemed a little too cute (incrementing then comparing with the decremented version)... however, this should be fine.

dtzSiFive:
> Originally the fallthrough seemed a little too cute (incrementing then comparing with the decremented version)... however, this should be fine.

Agreed, this is a bit too fiddly.  I'll poke at it.  Thanks!

youngar (approved):
Thanks!

PR #7582, opened 2024-09-04
[ImportVerilog] Skip defparams which have been handled by slang.

hailongSun2000:
For `defparam`, slang has propagated its value for the corresponding parameter. Such as:
```
module Top;
  Foo #(1) foo();
endmodule

module Foo #(
  parameter P = 32
);
  int a = P;
endmodule

module Bar;
  defparam Top.foo.P  = 4;
endmodule
```
On the AST, `P` and `a` both are 4. So we can skip `defparam`.


fabianschuiki:
Nice :partying_face:!

PR #7584, opened 2024-09-04
[FIRRTL][FIRParser] Defer inner symbols to post-processing, fix race.

dtzSiFive:
Adding inner symbols, specifically to module ports, mutates the attribute dictionary and that races / breaks concurrent access to properties (e.g., port names and types) commonly inspected during parsing other module bodies (e.g., InstanceOp's).

Avoid this by instead gathering a list of operations and their intended target as "fixups" to be processed after the original (possibly parallel) parsing has completed and it is safe to do so.

This could be parallellized (per-module fixups)
but is not expected to have sufficient work to justify the overhead.


youngar (approved):
LGTM

PR #7587, opened 2024-09-05
[HW] Add reduction patterns to trim port list of top-level module

maerhart:


fabianschuiki (approved):
This is awesome! I've wanted something like this every single time I've used the reducer :grin:

PR #7588, opened 2024-09-05
[FIRRTL] Add back unambiguous path requirement.

mikeurbach:
This was made a warning in https://github.com/llvm/circt/pull/7129. In reality, the flows can fail during LowerClasses if we do allow this through, so this adds back the previous restrictions so we ensure from the start that we have unambiguous paths as originally intended.

Closes https://github.com/llvm/circt/issues/7128.


dtzSiFive (approved):
LGTM!

Is it possible these `node`s are the top-levels such that they don't have one use but also aren't problematically multiply-instantiated?

I think that's not a thing but just asking the defensive question in case it's clear it might 👍 .

mikeurbach:
> Is it possible these nodes are the top-levels such that they don't have one use but also aren't problematically multiply-instantiated?

Good question, I think in this case, we wouldn't hit the error, because such modules would be the "owning module" of themselves, and we'd immediately stop here: https://github.com/llvm/circt/blob/75a192e2bc4d1eb1e7b48cd228c81721daafbedb/lib/Dialect/FIRRTL/Transforms/ResolvePaths.cpp#L52

I can add a small test case to be sure this works.

mikeurbach:
This is the exact test case I was going to write: https://github.com/llvm/circt/blob/75a192e2bc4d1eb1e7b48cd228c81721daafbedb/test/Dialect/FIRRTL/resolve-paths.mlir#L25-L32

Does that cover your concern @dtzSiFive? Or were you thinking of another scenario?

mikeurbach:
I'm going to go ahead and merge this, but happy to add anything else in a follow up.

dtzSiFive:
> Does that cover your concern @dtzSiFive? Or were you thinking of another scenario?

Nope, that covers it! Thanks! 💯

PR #7589, opened 2024-09-06
[Moore] Add assert, assume, and cover ops.

hailongSun2000:
These immediate assertion ops borrow from the `SV` dialect.

### DIFF
--- a/lib/Conversion/ImportVerilog/Statements.cpp
+++ b/lib/Conversion/ImportVerilog/Statements.cpp
@@ -463,6 +463,76 @@ struct StmtVisitor {
     return success();
   }
 
+  // Handle immediate assertion statements.
+  LogicalResult visit(const slang::ast::ImmediateAssertionStatement &stmt) {
+    auto cond = context.convertRvalueExpression(stmt.cond);
+    cond = context.convertToBool(cond);
+    if (!cond)
+      return failure();
+
+    // Handle assertion statements that don't have an action block.
+    if (stmt.ifTrue && stmt.ifTrue->as_if<slang::ast::EmptyStatement>()) {
+      auto defer = moore::DeferAssert::Immediate;
+      if (stmt.isFinal)
+        defer = moore::DeferAssert::Final;
+      else if (stmt.isDeferred)
+        defer = moore::DeferAssert::Observed;
+
+      switch (stmt.assertionKind) {
+      case slang::ast::AssertionKind::Assert:
+        builder.create<moore::AssertOp>(loc, defer, cond, StringAttr{});
+        return success();
+      case slang::ast::AssertionKind::Assume:
+        builder.create<moore::AssumeOp>(loc, defer, cond, StringAttr{});
+        return success();
+      case slang::ast::AssertionKind::CoverProperty:
+        builder.create<moore::CoverOp>(loc, defer, cond, StringAttr{});
+        return success();
+      default:
+        break;
+      }
+      mlir::emitError(loc) << "unsupported immediate assertion kind: "
+                           << slang::ast::toString(stmt.assertionKind);
+      return failure();
+    }
+
+    // Regard assertion statements with an action block as the "if-else".
+    cond = builder.create<moore::ConversionOp>(loc, builder.getI1Type(), cond);
+
+    // Create the blocks for the true and false branches, and the exit block.
+    Block &exitBlock = createBlock();
+    Block *falseBlock = stmt.ifFalse ? &createBlock() : nullptr;
+    Block &trueBlock = createBlock();
+    builder.create<cf::CondBranchOp>(loc, cond, &trueBlock,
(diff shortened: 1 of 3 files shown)


hailongSun2000:
I'm not sure what the `label` does :sweat_smile:.

hailongSun2000:
I know what `label` is. Like: **CHECK_THE_X_VALUE**
```
CHECK_THE_X_VALUE:
  assert...
```

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -463,6 +473,31 @@ struct StmtVisitor {
...
   }
 
+  // Handle immediate assertion statements.
+  LogicalResult visit(const slang::ast::ImmediateAssertionStatement &stmt) {
+    auto cond = context.convertRvalueExpression(stmt.cond);
+    cond = context.convertToBool(cond);
+    if (!cond)
+      return failure();
+
+    auto defer = convertDeferAssert(stmt.isDeferred, stmt.isFinal);
```
fabianschuiki, line 483: You could inline `convertDeferAssert` here and make it switch to different assert flavors conservatively:
```c++
auto defer = moore::DeferAssert::Immediate;
if (stmt.isFinal)
  defer = moore::DeferAssert::Final;
else if (stmt.isDeferred)
  defer = moore::DeferAssert::Observed;
```

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -463,6 +473,31 @@ struct StmtVisitor {
...
+    case slang::ast::AssertionKind::Assume:
+      builder.create<moore::AssumeOp>(loc, defer, cond, StringAttr{});
+      return success();
+    case slang::ast::AssertionKind::CoverProperty:
+      builder.create<moore::CoverOp>(loc, defer, cond, StringAttr{});
+      return success();
+    default:
+      break;
+    }
+    mlir::emitError(loc, "all immediate assertions handled");
```
fabianschuiki, line 497: Can you make this complain about unsupported assertions kinds? I assume there are more than just the three listed above. You could print something like:
```suggestion
    mlir::emitError(loc) << "unsupported immediate assertion kind: " << slang::ast::toString(stmt.assertionKind);
```

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -463,6 +473,31 @@ struct StmtVisitor {
...
 
+  // Handle immediate assertion statements.
+  LogicalResult visit(const slang::ast::ImmediateAssertionStatement &stmt) {
+    auto cond = context.convertRvalueExpression(stmt.cond);
+    cond = context.convertToBool(cond);
+    if (!cond)
+      return failure();
+
+    auto defer = convertDeferAssert(stmt.isDeferred, stmt.isFinal);
+    switch (stmt.assertionKind) {
```
fabianschuiki, line 484: Immediate assertions can have an action block associated with them, for example:
```
assert (x) $display("success");
assert (x) else $error("failure");
assert (x) $display("yay") else $error("nay");
```
It's not entirely clear to me if the presence of an action block simply turns the assert into an "if" statement, or whether the fact that it's an assert allows for additional global enabling/disabling controls to take effect. Do you know how this is supposed to work?

In any case, we can definitely deal with all of this later. If we postpone handling this, I'd suggest to add a check above the switch statement that ensures there is no action block associated with the assertion. That should produce an error like "unsupported immediate assertion action block" or similar.

hailongSun2000: Yeah, I also noticed that. It's so similar to `if-else` in the slang AST. Therefore, I cannot make up my mind on how to deal with these different situations. It would be great if we could lower the situations you mentioned to `if-else`.

lib/Conversion/ImportVerilog/Statements.cpp
hailongSun2000: Yeah, I need to inline `convertDeferAssert`. I did it.

lib/Conversion/ImportVerilog/Statements.cpp
hailongSun2000: This new error message is obviously better.

lib/Conversion/ImportVerilog/Statements.cpp
fabianschuiki: That sounds great! 🥳

fabianschuiki (approved):
This looks great! Can you add your
```
  assert (x) a = 1;  
  assert (x) else a = 0;
  assert (x) a = 1; else a = 0;
```
examples as a test, to ensure things work as expected before merging?

hailongSun2000:
I have added these.

fabianschuiki (approved):

PR #7592, opened 2024-09-10
[FIRRTL][GrandCentral] Remove legacy Augmented types.

dtzSiFive:
Remove String, Boolean, Integer, Double, Literal, Deleted.

Instead of ignoring these, they will now be errors.

This reduces supported types to: Ground, Vector, Bundle.


seldridge (approved):
LGTM

Amazingly I don't think this was ever load bearing. I can look at changes internally to make this impossible to emit.

PR #7597, opened 2024-09-12
[FIRRTL] Change InferReadWrite to use a walk, error if it sees WhenOps

seldridge:
Fix the InferReadWrite pass to walk into nested regions instead of only
visiting the top-level ops immediately in an FModuleOp's body.  This has
the effect of allowing it to work on memories that are declared inside
layers.

Change the InferReadWrite pass to fail if it ever sees a WhenOp.  If this
happens, the pass can silently do the wrong thing when trying to determine
signal drivers.  This is a conservative check (with false positives), but
will have no false negatives.


rwy7 (approved):

prithayan (approved):
Looks good.

dtzSiFive (approved):
LGTM, thanks!

PR #7598, opened 2024-09-13
[FIRRTL] Make IMCP work with Layers

seldridge:
Change IMCP's module update to use a walk instead of only visiting
top-level ops.  This is intended to be an entirely mechanical change and
is broken out in a separate commit because of its mechanical nature.

Change IMCP to recurse into layers.  This has the effect of allowing IMCP
to sink constants into layers and to properly visit instances which are
instantiated under layers.


dtzSiFive (approved):
Awesome! This looks right to me!

PR #7599, opened 2024-09-14
[FIRRTL] Use walk, InstanceInfo in AddSeqMemPorts

seldridge:
Change the `AddSeqMemPorts` pass (whic adds MBIST ports to memories) to
use a walk when adding ports to memories.  Practically, this causes the
pass to now be capable of adding MBIST to memories under `WhenOp`s or
`LayerBlockOp`s.  However, because this pass is intentionally _not_
supposed to add MBIST ports to anything that is not in the "design", this
will skip over any layers or modules that are instantiated wholly under
layers.

To do this, this pass was changed to use the new `InstanceInfo` analysis
instead of computing information about what modules are under the
design-under-test (DUT) manually.  Any checking related to layers is
similarly done using this analysis.

This change is prerequisite work to move the `LowerLayers` after
`AddSeqMemPorts` while still preserving the functionality of the original
pass order.  Unfortunately, layers do not compose sanely with
`AddSeqMemPorts`.  This is a failing of the original pass and its
assumptions about what is "design" vs. "verification" as opposed to any
fundamental issue with layers.  Generally, layers are intended as
verification features.  However, there is no actual requirement or
expectation that they will be used this way.  Going forward,
`AddSeqMemPorts` is intended to be removed and replaced with a feature in
Chisel.

Additionally, this this commit includes a small change to the pass output
where the connects created are placed immediately following the memory
instance.  This is necessary to create the connects inside the layer
blocks.  However, it is also slightly better output than before.  (This
changed one test superficially.)

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "circt/Analysis/FIRRTLInstanceInfo.h"
 #include "circt/Dialect/Emit/EmitOps.h"
 #include "circt/Dialect/FIRRTL/AnnotationDetails.h"
 #include "circt/Dialect/FIRRTL/FIRRTLAnnotationHelper.h"
@@ -49,7 +50,7 @@ struct AddSeqMemPortsPass
   void createOutputFile(igraph::ModuleOpInterface moduleOp);
   InstanceGraphNode *findDUT();
   void processMemModule(FMemModuleOp mem);
-  LogicalResult processModule(FModuleOp moduleOp, bool isDUT);
+  LogicalResult processModule(FModuleOp moduleOp);
 
   /// Get the cached namespace for a module.
   hw::InnerSymbolNamespace &getModuleNamespace(FModuleLike moduleOp) {
@@ -83,6 +84,7 @@ struct AddSeqMemPortsPass
   DenseMap<Attribute, Operation *> innerRefToInstanceMap;
 
   InstanceGraph *instanceGraph;
+  InstanceInfo *instanceInfo;
 
   /// If the metadata output file was specified in an annotation.
   StringAttr outputFile;
@@ -187,18 +189,16 @@ void AddSeqMemPortsPass::processMemModule(FMemModuleOp mem) {
   size_t portIndex = mem.getNumPorts();
   auto &memInfo = memInfoMap[mem];
   auto &extraPorts = memInfo.extraPorts;
-  for (auto &p : userPorts)
+  for (auto const &p : userPorts)
     extraPorts.emplace_back(portIndex, p);
   mem.insertPorts(extraPorts);
   // Attach the extraPorts metadata.
   mem.setExtraPortsAttr(extraPortsAttr);
 }
 
-LogicalResult AddSeqMemPortsPass::processModule(FModuleOp moduleOp,
-                                                bool isDUT) {
+LogicalResult AddSeqMemPortsPass::processModule(FModuleOp moduleOp) {
   auto *context = &getContext();
-  // Insert the new port connections at the end of the module.
-  auto builder = OpBuilder::atBlockEnd(moduleOp.getBodyBlock());
+  auto builder = OpBuilder(moduleOp.getContext());
   auto &memInfo = memInfoMap[moduleOp];
   auto &extraPorts = memInfo.extraPorts;
   // List of ports added to submodules which must be connected to this module's
@@ -208,7 +208,7 @@ LogicalResult AddSeqMemPortsPass::processModule(FModuleOp moduleOp,
   // The base index to use when adding ports to the current module.
   unsigned firstPortIndex = moduleOp.getNumPorts();
 
-  for (auto &op : llvm::make_early_inc_range(*moduleOp.getBodyBlock())) {
+  auto result = moduleOp.walk([&](Operation *op) {
     if (auto inst = dyn_cast<InstanceOp>(op)) {
       auto submodule = inst.getReferencedModule(*instanceGraph);
 
@@ -216,7 +216,7 @@ LogicalResult AddSeqMemPortsPass::processModule(FModuleOp moduleOp,
       // If there are no extra ports, we don't have to do anything.
       if (subMemInfoIt == memInfoMap.end() ||
           subMemInfoIt->second.extraPorts.empty())
-        continue;
+        return WalkResult::advance();
       auto &subMemInfo = subMemInfoIt->second;
       // Find out how many memory ports we have to add.
       auto &subExtraPorts = subMemInfo.extraPorts;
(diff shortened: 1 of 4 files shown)


seldridge:
This does not handle some of the weirder things that could be going on here related to the memory metadata file. This file is intended to show where these extra ports are on which memory. However, the memory may be moved under a layer block bind file.

It may be better to instead completely disallow this pass for memories under layer blocks. It just really doesn't make sense to be adding MBIST ports to memories which are basically design verification collateral. Alternative way of arguing this, because layers are new, nobody is using this behavior, thereby we can punt on what to do here until there is an actual user.

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
```diff
@@ -403,27 +417,75 @@ void AddSeqMemPortsPass::runOnOperation() {
...
+              continue;
+            }
+            if (instanceInfo->anyInstanceUnderLayer(
+                    instNode->getParent()->getModule())) {
+              diag.attachNote(instanceOp.getLoc())
+                  << "this instance is inside a module that is instantiated "
+                     "under a layer block";
+            }
+          }
+          return signalPassFailure();
```
rwy7, line 478: I think this check belongs in processMemModule, which should be return a LogicalResult.

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
```diff
@@ -403,27 +417,75 @@ void AddSeqMemPortsPass::runOnOperation() {
...
+    // If any memories are partially instantiated under a layer then error.
+    for (auto *node : llvm::post_order(instanceGraph)) {
       auto op = node->getModule();
+      // Skip the module if there is a DUT and no instances are under it.
+      if (instanceInfo->hasDut() && !instanceInfo->anyInstanceUnderDut(op))
+        continue;
+
+      // Skip anything wholly under a layer.
+      if (instanceInfo->allInstancesUnderLayer(op))
+        continue;
```
rwy7, line 442: If the DUT is wholly under a layer, then we won't add any MBIST ports. Seems OK.

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
```diff
@@ -403,27 +417,75 @@ void AddSeqMemPortsPass::runOnOperation() {
...
+    //
+    //   1. If no design-under-test is specified, visit all modules.
+    //   2. If the design-under-test is specified, only visit modules under it.
+    //
+    // Additionally, skip any modules that are wholly instantiated under layers.
+    // If any memories are partially instantiated under a layer then error.
+    for (auto *node : llvm::post_order(instanceGraph)) {
       auto op = node->getModule();
+      // Skip the module if there is a DUT and no instances are under it.
+      if (instanceInfo->hasDut() && !instanceInfo->anyInstanceUnderDut(op))
```
rwy7, line 437: This could be, "not anyInstanceUnderEffectiveDut", but I'm not sure you need it, if you start the walk at the effective dut node.

rwy7 (approved):

lib/Dialect/FIRRTL/Transforms/CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ add_circt_dialect_library(CIRCTFIRRTLTransforms
   LINK_LIBS PUBLIC
   CIRCTFIRRTLAnalysis
   CIRCTFIRRTL
+  CIRCTFIRRTLAnalysis
```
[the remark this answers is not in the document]
seldridge: 👍 Fixed in a430b6b94.

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
```diff
@@ -403,27 +417,75 @@ void AddSeqMemPortsPass::runOnOperation() {
...
-    // Visit the nodes in post-order.
-    for (auto *node : llvm::post_order(dutNode)) {
+    // Visit the modules in post-order in one of two ways:
+    //
+    //   1. If no design-under-test is specified, visit all modules.
+    //   2. If the design-under-test is specified, only visit modules under it.
+    //
+    // Additionally, skip any modules that are wholly instantiated under layers.
+    // If any memories are partially instantiated under a layer then error.
+    for (auto *node : llvm::post_order(instanceGraph)) {
```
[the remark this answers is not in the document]
seldridge: Good catch. Changed in c76b4deac.

Part of what was going on here is that this PR was written _before_ `InstanceInfo.effectiveDut` was added. That then makes this trivial. 👍

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
seldridge: Changed to start from the effective DUT in c76b4deac.

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
seldridge: That seems reasonable, though I suppose it could work? E.g., there's no problem wiring to the DUT as long as you don't have to pass through any intermediary layers to do it. (InstanceInfo doesn't track that, though!)

This smells like a missing check that should be in `LowerAnnotations` (as I don't think we can do it with a verifier).

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
seldridge: Nice idea. This also makes this have a consistent signature with `processModule`. Changed in 2e5dd7fe8.

lib/Dialect/FIRRTL/Transforms/AddSeqMemPorts.cpp
```diff
@@ -273,20 +273,35 @@ LogicalResult AddSeqMemPortsPass::processModule(FModuleOp moduleOp,
...
+  });
+
+  if (result.wasInterrupted())
+    return failure();
 
   // Add the extra ports to this module.
   moduleOp.insertPorts(extraPorts);
 
   // Connect the submodule ports to the parent module ports.
+  DenseMap<Operation *, OpBuilder::InsertPoint> instToInsertionPoint;
```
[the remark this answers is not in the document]
seldridge: I like this idea. I realized that this change was now very out-of-place for this PR. This PR doesn't need this anymore (as it is not creating connects under layer blocks). I opted instead to just back this out of the patch in b11b0723a.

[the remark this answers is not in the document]
seldridge: I realized that this code is still necessary as this pass _should_ work for `WhenOp`s and it was still busted. I reverted the removal of this in f83e453bf. I then added code to handle `WhenOp`s in c43c02c93. This uses invalidation to handle the case of output ports.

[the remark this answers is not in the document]
seldridge: I'm going to leave this for a follow-up as this would permute the code a bit too much in this PR. I like the idea.

PR #5358, opened 2023-06-11
[circt-lec] Fixed equivalence check for multiple outputs

fzi-hielscher:
Currently the solver of the equivalence check creates a separate _not equal_ constraint for each output of the processed module(s). Consequently, the constraint system becomes unsatisfiable as soon as at least one of the outputs is equivalent, causing the equivalence check for the entire module to pass. This PR changes the constraints to make them unsatisfiable iff all outputs are equivalent, by solving for the disjunction of the _not equal_ expressions.

This also adds a test case, checking that two non-equivalent circuits do not pass the comparison. This turned out to be more complicated than I was expecting. When checking the output of `circt-opt` for `c1 != c2` the test will still fail, because it produces a non-zero exit code for non-equivalent circuits. The best thing I could come up with was checking for equivalence and marking the test as expected to fail.

PR #7603, opened 2024-09-16
Add `verif.bmc` VerifToSMT lowering

TaoBi22:
This PR adds a pattern to lower the `verif.bmc` op into the program to actually perform the check (from the splitting up of #7259 - only the actual tool left after this!), similarly to what circt-lec currently does.

### DIFF
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -9,12 +9,17 @@
 #include "circt/Conversion/VerifToSMT.h"
 #include "circt/Conversion/HWToSMT.h"
 #include "circt/Dialect/SMT/SMTOps.h"
+#include "circt/Dialect/Seq/SeqTypes.h"
 #include "circt/Dialect/Verif/VerifOps.h"
+#include "circt/Support/Namespace.h"
 #include "mlir/Conversion/ReconcileUnrealizedCasts/ReconcileUnrealizedCasts.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/IR/ValueRange.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/DialectConversion.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/ADT/SmallVector.h"
 
 namespace circt {
 #define GEN_PASS_DEF_CONVERTVERIFTOSMT
@@ -30,7 +35,8 @@ using namespace hw;
 //===----------------------------------------------------------------------===//
 
 namespace {
-/// Lower a verif::AssertOp operation with an i1 operand to a smt::AssertOp.
+/// Lower a verif::AssertOp operation with an i1 operand to a smt::AssertOp,
+/// negated to check for unsatisfiability.
 struct VerifAssertOpConversion : OpConversionPattern<verif::AssertOp> {
   using OpConversionPattern<verif::AssertOp>::OpConversionPattern;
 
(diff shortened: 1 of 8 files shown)


lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -30,7 +35,8 @@ using namespace hw;
 //===----------------------------------------------------------------------===//
 
 namespace {
-/// Lower a verif::AssertOp operation with an i1 operand to a smt::AssertOp.
+/// Lower a verif::AssertOp operation with an i1 operand to a smt::AssertOp,
+/// negated to check for unsatisfiability.
```
TaoBi22, line 39: Currently `verif.assert` is lowered directly to `smt.assert`, whereas we want it negated in this context to check for unsatisfiability. I'm not aware of any verif dialect use cases that mean this negation is undesirable, but let me know if any exist

fzi-hielscher: Sorry for the drive-by comment, I haven't looked into this properly. But won't this change inevitably lead to effectively the same problem as in #5358 ? I.e., any `verif.assert(true)` will now become `smt.assert(false)`, thus rendering the model unsatisfiable regardless of all other (failing) assertions?

dobios: Yeah I ran into a similar issue in btor and opted simply the prune those assertions as they're trivially true, maybe the same should be done here?

maerhart: Removing such trivial assertions is nice, but I'd probably not do it in a canonicalizer because it makes it harder to keep track of a list of passed assertions to report back to the engineer, etc.
Also, just pruning these trivial ones doesn't make the model checker sound.

Given
```
assert P1
assert P2
assert P3
// equivalent to
assert P1 & P2 & P3
```
we currently emit
```
assert !P1
assert !P2
assert !P3
// equivalent to
assert !(P1 || P2 || P3)
```
but we want
```
assert !(P1 & P2 & P3)
```
Not sure how other open-source or commercial model checkers deal with that. I see two options
* add assertion by assertion "interactively" by checking satisfiability after each added assertion (bad performance-wise unless the solver is really good in caching things, which I don't know enough about)
* Combine all assertions into a single one (returning good counter-examples may need some post processing/running additional checks)

I like the approach here of keeping the assert lowering simple. Maybe we can add a separate pass earlier in the pipeline to restructure the circuit for one of the two approaches later on.

TaoBi22: Thanks everyone for taking a look, and thanks for catching that @fzi-hielscher!

This is a great point - thanks for the suggestions @maerhart, those sound good to me! I guess the easy interrim fix is to add on an error in this lowering (or a canonicalizer to the BMC op, though to me this seems more of a point in the lowering than the op itself) that requires the BMC op to have only one assertion - that shouldn't really introduce any compromises in functionality as what would have previously been individual assertions can all just be ANDed, but means that we won't unsoundly produce an unsat when there's actually an assertion that can be violated.

test/Conversion/VerifToSMT/verif-to-smt.mlir
```diff
@@ -60,3 +61,89 @@ func.func @test(%arg0: !smt.bv<1>) -> (i1, i1, i1) {
...
+  circuit {
+  ^bb0(%clk: !seq.clock, %arg0: i32, %state0: i32):
+    %c-1_i32 = hw.constant -1 : i32
+    %0 = comb.add %arg0, %state0 : i32
+    // %state0 is the result of a seq.compreg taking %0 as input
+    %2 = comb.xor %state0, %c-1_i32 : i32
+    verif.yield %2, %0 : i32, i32
+  }
+  func.return %bmc : i1
+}
```
maerhart, line 149: A test for `verif.assume` seems to be missing

TaoBi22: Good catch, thanks!

TaoBi22:
@maerhart I think that should be all comments addressed - if you could take a quick scan over the added assertion count checker that would be great! I

test/Conversion/VerifToSMT/verif-to-smt-errors.mlir
```diff
@@ -15,3 +15,68 @@ func.func @assert_with_unsupported_property_type(%arg0: !smt.bv<1>) {
...
+    %sum = comb.add %arg0, %arg0 : i32
+    verif.yield %sum : i32
+  }
+  func.return %bmc : i1
+}
+
+hw.module @TwoAssertions(in %x: i1, in %y: i1) {
+  verif.assert %x : i1
+  verif.assert %y : i1
+}
```
maerhart, line 82: Really nice tests! 💯

lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -146,6 +168,230 @@ struct LogicEquivalenceCheckingOpConversion
...
+        rewriter.create<func::CallOp>(loc, initFuncOp)->getResults();
+
+    // InputDecls order should be <circuit arguments> <state arguments>
+    // <wasViolated>
+    // Get list of clock indexes in circuit args
+    size_t initIndex = 0;
+    SmallVector<Value> inputDecls;
+    SmallVector<int> clockIndexes;
+    for (auto [curIndex, types] :
+         llvm::enumerate(llvm::zip(oldCircuitInputTy, circuitInputTy))) {
```
maerhart, line 295: Nit: Enumerate can zip at the same time
```suggestion
         llvm::enumerate(oldCircuitInputTy, circuitInputTy)) {
```

maerhart (approved):
Thanks!

lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -146,6 +168,230 @@ struct LogicEquivalenceCheckingOpConversion
...
+                  ConversionPatternRewriter &rewriter) const override {
+    // Check there is exactly one assertion
+    auto topAssertions = op.getCircuit().getOps<verif::AssertOp>();
+    int numAssertions =
+        std::distance(topAssertions.begin(), topAssertions.end());
+    SymbolTable symbolTable(op->getParentOfType<ModuleOp>());
+    SmallVector<mlir::Operation *> worklist;
+    for (auto instance : op.getCircuit().getOps<InstanceOp>()) {
+      worklist.push_back(symbolTable.lookup(instance.getModuleName()));
+    }
```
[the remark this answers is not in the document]
TaoBi22: :woman_facepalming: thanks!

lib/Conversion/VerifToSMT/VerifToSMT.cpp
TaoBi22: :star_struck:

lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -160,21 +378,70 @@ struct ConvertVerifToSMTPass
...
+            if (numAssertions > 1)
+              break;
+          }
+          if (numAssertions > 1) {
+            op->emitError(
+                "bounded model checking problems with multiple assertions are "
+                "not yet "
+                "correctly handled - instead, you can assert the "
+                "conjunction of your assertions");
+            signalPassFailure();
```
maerhart, line 430: We could early exit the pass here. Otherwise fine to merge from my side. Thanks for moving it out!

PR #7604, opened 2024-09-18
Fix links

ju-sh:


mikeurbach:
Thanks for sending this PR, but I don't think we can merge this change as-is.

It has come up different times, for example: https://github.com/llvm/circt/issues/4343#issuecomment-1583330754. Basically, we are using these files from the github repository to build the website (circt.llvm.org), and we want the links to work on the website. If we merged this change, it would fix the links in the github rendering of the markdown, but would then break the website.

No one has taken the time to figure out the right way to encode these links in a way that works for both github and the website, and we are prioritizing keeping the website links healthy at the expense of the github rendering.

ju-sh:
Ah.. right. Never mind then.

PR #7605, opened 2024-09-18
[Arc] Fully support initialization through seq.initial

uenoku:
This PR implements state initialization using `seq.initial` not limited to constants. Combined with https://github.com/llvm/circt/pull/7606 we can perform random initialization in arcilator

### DIFF
--- a/integration_test/arcilator/JIT/reg.mlir
+++ b/integration_test/arcilator/JIT/reg.mlir
@@ -1,0 +1,66 @@
+// RUN: arcilator %s --run --jit-entry=main | FileCheck %s
+// REQUIRES: arcilator-jit
+
+// CHECK: o1 = 10
+// CHECK-NEXT: o2 = 5
+// CHECK-NEXT: o1 = 11
+// CHECK-NEXT: o2 = 6
+// CHECK-NEXT: o1 = 12
+// CHECK-NEXT: o2 = 7
+
+func.func private @random() -> i32 {
+  %0 = arith.constant 16 : i32
+  func.return %0: i32
+}
+
+hw.module @counter(in %clk: i1, out o1: i8, out o2: i8) {
+  %seq_clk = seq.to_clock %clk
+  %c0_i8 = hw.constant 0 : i8
+
+  %r0 = seq.compreg %added1, %seq_clk initial %0#0 : i8
+  %r1 = seq.compreg %added2, %seq_clk initial %0#1 : i8
+  %0:2 = seq.initial {
+    %1 = func.call @random() : () -> i32
+    %2 = comb.extract %1 from 0 : (i32) -> i8
+    %3 = hw.constant 5 : i8
+    seq.yield %2, %3: i8, i8
+  } : !seq.immutable<i8>, !seq.immutable<i8>
+
+  %one = hw.constant 1 : i8
+  %added1 = comb.add %r0, %one : i8
+  %added2 = comb.add %r1, %one : i8
+
+  hw.output %r0, %r1 : i8, i8
+}
+
+func.func @main() {
+  %zero = arith.constant 0 : i1
+  %one = arith.constant 1 : i1
+  %lb = arith.constant 0 : index
+  %ub = arith.constant 2 : index
+  %step = arith.constant 1 : index
+
+  arc.sim.instantiate @counter as %model {
+    %init_val1 = arc.sim.get_port %model, "o1" : i8, !arc.sim.instance<@counter>
+    %init_val2 = arc.sim.get_port %model, "o2" : i8, !arc.sim.instance<@counter>
+
+    arc.sim.emit "o1", %init_val1 : i8
+    arc.sim.emit "o2", %init_val2 : i8
+
+    scf.for %i = %lb to %ub step %step {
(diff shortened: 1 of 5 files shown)


lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
```diff
@@ -35,15 +35,14 @@ static LogicalResult convertInitialValue(seq::CompRegOp reg,
...
-  auto *op = init.getDefiningOp()->clone();
-  reg->getBlock()->getOperations().insert(Block::iterator(reg), op);
-  auto result = op->getResult(0);
-  values.push_back(result);
+  // unrealized_conversion_cast to normal type
+  OpBuilder builder(reg);
+  auto init = builder
+                  .create<mlir::UnrealizedConversionCastOp>(
+                      reg.getLoc(), reg.getType(), reg.getInitialValue())
+                  .getResult(0);
```
fabianschuiki, line 43: I'm wondering if it might be worthwhile having an `immutable<T> -> T` cast operation, something like `seq.unwrap_immutable`. That could also enable some canonicalizations where a `seq.compreg` could lower to its initial value if it provably can never change.

Unrealized conversion cast also works though :smiley:

fabianschuiki (approved):
This is really really cool! Thanks for adding support for this to Arc :partying_face: :pray:

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
```diff
@@ -100,10 +99,8 @@ LogicalResult Converter::runOnModule(HWModuleOp module) {
   // Find all arc-breaking operations in this module and assign them an index.
   arcBreakers.clear();
   arcBreakerIndices.clear();
-  SmallVector<seq::InitialOp> initialOps;
   for (Operation &op : *module.getBodyBlock()) {
     if (isa<seq::InitialOp>(&op)) {
-      initialOps.push_back(cast<seq::InitialOp>(&op));
       continue;
     }
```
maerhart, line 105: Nit: can remove braces for consistency

test/Conversion/ConvertToArcs/convert-to-arcs.mlir
```diff
@@ -236,8 +242,9 @@ hw.module @Trivial(in %clock: !seq.clock, in %i0: i4, in %reset: i1, out out: i4
 
 // CHECK-LABEL: hw.module @TrivialWithInit(
 hw.module @TrivialWithInit(in %clock: !seq.clock, in %i0: i4, in %reset: i1, out out: i4) {
-  // CHECK: [[CST2:%.+]] = hw.constant 2 : i4
-  // CHECK: [[RES0:%.+]] = arc.state @[[TRIVIALINIT_ARC]](%i0) clock %clock reset %reset initial ([[CST2]] : i4) latency 1 {names = ["foo"]
+  // CHECK:      %[[INIT:.+]] = seq.initial {
```
maerhart, line 245: Nit:
```suggestion
  // CHECK: %[[INIT:.+]] = seq.initial {
```

integration_test/arcilator/JIT/reg.mlir
```diff
@@ -0,0 +1,66 @@
+// RUN: arcilator %s --run --jit-entry=main | FileCheck %s
+// REQUIRES: arcilator-jit
+
+// CHECK: o1 = 10
```
maerhart, line 4: Shouldn't this be 16, or am I missing something?

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
maerhart: Or we might reconsider supporting `arc.initial` before lower state and do the conversion to it in ConvertToArcs?

maerhart (commented):
Really cool work! Just a bit puzzled about the integration test results 🤔

integration_test/arcilator/JIT/reg.mlir
uenoku: https://github.com/llvm/circt/blob/bba8cfb872ae7e1e3263ce77e87ec65305b49b7c/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp#L539-L540

It is printed in hex but yeah, it's a bit confusing so I'll change to use a smaller number.

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
uenoku: `seq.unwrap_immutable` sounds reasonable and will follow-up. `arc.initial` might work but currently `arc.initial` doesn't produce any value and we need to allocate states for them. So I guess there would be a big overlap with LowerState. Another approach might be to change `arc.state`'s initial value operand to take `seq.immutable` type instead of a normal integer.

integration_test/arcilator/JIT/reg.mlir
maerhart: Oh, totally forgot about that. Maybe we should prepend a `0x` to this format string. Or a smaller number, a comment, or using `0x10` at the constant op would be nice 🙂

maerhart (approved):

uenoku:
Thank you for the review! I'm going to merge this and will follow up how to deal with unrealized conversion cast.

PR #7607, opened 2024-09-18
[SV] Add MacroRefOp to represet macro statement

uenoku:
There has been MacroRefExprOp and MacroRefExprSEOp to use macro symbols as expressions but it was not possible to use macro as a statement. This commit adds `sv.macro.ref` op to represent a statement. Since `sv.macro.ref` mnemonic was used by MacroRefExprOp this commit also renames `sv.macro.ref` to `sv.macro.ref.expr` at the same time.

### DIFF
--- a/include/circt/Dialect/SV/SVStatements.td
+++ b/include/circt/Dialect/SV/SVStatements.td
@@ -550,7 +550,7 @@ def MacroRefOp : SVOp<"macro.ref", [
   let summary = "Statement to refer to a SystemVerilog macro";
   let description = [{
     This operation represent a statement by referencing a named macro.
-    }];
+  }];
 
   let arguments = (ins FlatSymbolRefAttr:$macroName, Variadic<AnyType>:$inputs);
   let results = (outs);
--- a/include/circt/Dialect/SV/SVStatements.td
+++ b/include/circt/Dialect/SV/SVStatements.td
@@ -545,6 +545,29 @@ def VerbatimOp : SVOp<"verbatim"> {
   ];
 }
 
+def MacroRefOp : SVOp<"macro.ref", [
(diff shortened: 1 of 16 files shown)


include/circt/Dialect/SV/SVStatements.td
```diff
@@ -545,6 +545,29 @@ def VerbatimOp : SVOp<"verbatim"> {
   ];
 }
 
+def MacroRefOp : SVOp<"macro.ref", [
+        DeclareOpInterfaceMethods<SymbolUserOpInterface>]> {
+  let summary = "Statement to refer to a SystemVerilog macro";
+  let description = [{
+    This operation represent a statement by referencing a named macro.
+    }];
```
seldridge, line 553: Mega nit:
```suggestion
  }];
```

test/Conversion/ExportVerilog/sv-dialect.mlir
```diff
@@ -265,6 +265,9 @@ hw.module @M1<param1: i42>(in %clock : i1, in %cond : i1, in %val : i8) {
       // CHECK-NEXT: $fwrite(32'h80000002, "M: %x\n", `MACRO(8'(val + 8'h2A), val ^ 8'h2A));
       sv.fwrite %fd, "M: %x\n"(%text) : i8
 
+      // CHECK-NEXT: `INIT_RANDOM(val)
+      sv.macro.ref @INIT_RANDOM (%val) : i8
```
seldridge, line 269: This test doesn't hit the interleave comma path nor the path without parentheses. Could the PR include a test of these if it is not there already?

seldridge (approved):
Nice. This was an obviously missing feature. Thanks!

PR #7610, opened 2024-09-19
[CMake] Consistently declare conversion libraries and simplify circt-opt link target list

maerhart:
* Use `add_circt_conversion_library` to define the libraries in `lib/Conversion`. Currently some already do that and others just declare them as regular libraries. `ExportChiselInterface`, `ExportVerilog`, and `ImportVerilog` are registered as translation libraries which makes sense, but makes me wonder: shouldn't they exist in another directory (e.g., `lib/Target` for the export passes)?
* Use the CMake property to add them as link targets to circt-opt. When adding a new dialect, it will then be linked against it automatically as with the other non-dialect-specific tools.


Dinistro (approved):
LGTM, nice cleanup 🙂

mikeurbach (approved):
Nice!

rachitnigam (approved):
LGTM!

dtzSiFive (approved):
LGTM, thanks!

PR #7612, opened 2024-09-20
[FIRRTL] Add InstanceInfo Analysis

seldridge:
Add a new InstanceInfo analysis.  This analysis provides common information which can be computed from a single walk of the InstanceGraph. The idea behind this is that there are a large amount of common information that different passes want to compute which all follows this same walk (and therefore has the same algorithmic complexity to compute it).  Move all this information into a single, common analysis.

This analysis currently enables O(1) queries of if something is the DUT, is instantiated under the DUT, or is instantiated under a layer.  More queries can be added as they are identified.

This new analysis naturally depends on the InstanceGraph analysis.

### DIFF
--- a/include/circt/Analysis/FIRRTLInstanceInfo.h
+++ b/include/circt/Analysis/FIRRTLInstanceInfo.h
@@ -28,29 +28,50 @@ class InstanceInfo {
 public:
   explicit InstanceInfo(Operation *op, mlir::AnalysisManager &am);
 
-  /// A lattice value recording whether or not a property is true for a set of
-  /// operations.
-  struct LatticeValue {
+  /// A lattice value to record the value of a property.
+  class LatticeValue {
+
+  public:
     enum Kind {
-      // Indicates that we have no information about this instance.  If the
-      // circuit has been completely analyzed, then this is equivalent to the
-      // property holds for zero instances.
+      // Indicates that we have no information about this property.  Properties
+      // start in this state and then are changed to Constant or Mixed.
       Unknown = 0,
-      // Indicates that the property has a true or false value for all instances
-      // observed.
+
+      // Indicates that the property has a true or false value.
       Constant,
-      // Indicates that the property holds for some, but not all instances.
+
+      // Indicates that the property was found to be both true and false.
       Mixed
     };
 
+  private:
     /// Whether or not the property holds.
     Kind kind = Kind::Unknown;
 
     /// The value of the property if `kind` is `Constant`.
-    bool constant = false;
+    bool value = false;
+
+  public:
+    /// Return true if the kind is Unknown.
+    bool isUnknown() const;
+
+    /// Return true if the kind is Constant.
+    bool isConstant() const;
+
+    /// Return true if the kind is Mixed.
+    bool isMixed() const;
+
+    /// Return the value.  This should only be used if the kind is Constant.
+    bool getConstant() const;
+
+    /// Set this LatticeValue to a constant.
+    void markConstant(bool constant);
 
     /// Merge attributes from another LatticeValue into this one.
     void mergeIn(LatticeValue that);
+
+    /// Merge a constant value into this one.
+    void mergeIn(bool value);
   };
 
   struct ModuleAttributes {
@@ -59,10 +80,10 @@ public:
     bool isDut;
 
     /// Indicates if this module is instantiated under the design-under-test.
-    InstanceInfo::LatticeValue underDut = {LatticeValue::Kind::Unknown};
+    InstanceInfo::LatticeValue underDut;
 
     /// Indicates if this module is instantiated under a layer.
-    InstanceInfo::LatticeValue underLayer = {LatticeValue::Kind::Unknown};
+    InstanceInfo::LatticeValue underLayer;
   };
 
   /// Return true if this module is the design-under-test.
@@ -97,15 +118,11 @@ private:
 #ifndef NDEBUG
 inline llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
                                      const InstanceInfo::LatticeValue &value) {
-  switch (value.kind) {
-  case InstanceInfo::LatticeValue::Kind::Unknown:
+  if (value.isUnknown())
     return os << "unknown";
-  case InstanceInfo::LatticeValue::Kind::Constant:
-    return os << "constant<" << (value.constant ? "true" : "false") << ">";
-  case InstanceInfo::LatticeValue::Kind::Mixed:
-    return os << "mixed";
-  }
-  return os;
+  if (value.isConstant())
+    return os << "constant<" << (value.getConstant() ? "true" : "false") << ">";
+  return os << "mixed";
 }
 #endif
 
--- a/include/circt/Analysis/FIRRTLInstanceInfo.h
+++ b/include/circt/Analysis/FIRRTLInstanceInfo.h
@@ -23,8 +23,6 @@
 namespace circt {
 namespace firrtl {
 
-using namespace detail;
-
 class InstanceInfo {
 
 public:
@@ -46,16 +44,13 @@ public:
     };
 
     /// Whether or not the property holds.
-    Kind kind = Unknown;
+    Kind kind = Kind::Unknown;
 
     /// The value of the property if `kind` is `Constant`.
     bool constant = false;
 
-    // Combine this lattice value with another lattice value.
-    void merge(LatticeValue that);
-
-    // Merge evidence that a property is true.
-    void merge(bool property);
+    /// Merge attributes from another LatticeValue into this one.
+    void mergeIn(LatticeValue that);
   };
 
   struct ModuleAttributes {
--- a/include/circt/Analysis/FIRRTLInstanceInfo.h
+++ b/include/circt/Analysis/FIRRTLInstanceInfo.h
@@ -68,18 +68,21 @@ public:
   /// Return true if this module is the design-under-test.
   bool isDut(FModuleOp op);
 
-  /// Return true if this module is instantiated under the design-under-test.
-  bool isUnderDut(FModuleOp op);
+  /// Return true if at least one instance of this module is under (or
+  /// transitively under) the design-under-test.
+  bool atLeastOneInstanceUnderDut(FModuleOp op);
 
-  /// Return true if all instantiations of this module are under the
-  /// design-under-test.
-  bool isFullyUnderDut(FModuleOp op);
+  /// Return true if all instances of this module are under (or transitively
+  /// under) the design-under-test.
+  bool allInstancesUnderDut(FModuleOp op);
 
-  /// Return true if this module is instantiated under a layer block.
-  bool isUnderLayer(FModuleOp op);
+  /// Return true if at least one instance of this module is under (or
+  /// transitively under) a layer.
+  bool atLeastOneInstanceUnderLayer(FModuleOp op);
 
-  /// Return true if all instantiations of this module are under layer blocks.
-  bool isFullyUnderLayer(FModuleOp op);
+  /// Return true if all instances of this module are under (or transitively
+  /// under) layer blocks.
+  bool allInstancesUnderLayer(FModuleOp op);
 
 private:
   /// Internal mapping of operations to module attributes.
--- a/include/circt/Analysis/FIRRTLInstanceInfo.h
+++ b/include/circt/Analysis/FIRRTLInstanceInfo.h
@@ -59,7 +59,7 @@ public:
   };
 
   struct ModuleAttributes {
-    /// Indicates that this module is the desgin-under-test.  This is indicated
+    /// Indicates that this module is the design-under-test.  This is indicated
     /// with a `sifive.enterprise.firrtl.MarkDUTAnnotation`.
     bool isDut;
 
--- a/include/circt/Analysis/FIRRTLInstanceInfo.h
+++ b/include/circt/Analysis/FIRRTLInstanceInfo.h
@@ -69,11 +69,13 @@ public:
   bool isDut(FModuleOp op);
 
   /// Return true if at least one instance of this module is under (or
-  /// transitively under) the design-under-test.
+  /// transitively under) the design-under-test.  This is true if the module is
+  /// the design-under-test.
   bool atLeastOneInstanceUnderDut(FModuleOp op);
 
   /// Return true if all instances of this module are under (or transitively
-  /// under) the design-under-test.
+  /// under) the design-under-test.  This is true if the module is the
+  /// design-under-test.
   bool allInstancesUnderDut(FModuleOp op);
 
   /// Return true if at least one instance of this module is under (or
--- a/include/circt/Analysis/FIRRTLInstanceInfo.h
+++ b/include/circt/Analysis/FIRRTLInstanceInfo.h
@@ -1,0 +1,115 @@
+//===- FIRRTLInstanceInfo.h - Instance info analysis ------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file defines the InstanceInfo analysis.  This is an analysis that
+// depends on the InstanceGraph analysis, but provides additional information
+// about FIRRTL operations.  This is useful if you find yourself needing to
+// selectively iterate over parts of the design.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_DIALECT_FIRRTL_FIRRTLINSTANCEINFO_H
+#define CIRCT_DIALECT_FIRRTL_FIRRTLINSTANCEINFO_H
+
+#include "circt/Dialect/FIRRTL/FIRRTLOps.h"
+#include "circt/Support/LLVM.h"
+#include "mlir/Pass/AnalysisManager.h"
+
+namespace circt {
+namespace firrtl {
+
+using namespace detail;
+
+class InstanceInfo {
+
+public:
+  explicit InstanceInfo(Operation *op, mlir::AnalysisManager &am);
+
+  /// A lattice value recording whether or not a property is true for a set of
+  /// operations.
+  struct LatticeValue {
+    enum Kind {
+      // Indicates that we have no information about this instance.  If the
+      // circuit has been completely analyzed, then this is equivalent to the
+      // property holds for zero instances.
+      Unknown = 0,
+      // Indicates that the property has a true or false value for all instances
+      // observed.
+      Constant,
+      // Indicates that the property holds for some, but not all instances.
+      Mixed
+    };
+
+    /// Whether or not the property holds.
+    Kind kind = Unknown;
+
+    /// The value of the property if `kind` is `Constant`.
+    bool constant = false;
+
+    // Combine this lattice value with another lattice value.
+    void merge(LatticeValue that);
+
+    // Merge evidence that a property is true.
+    void merge(bool property);
+  };
+
+  struct ModuleAttributes {
+    /// Indicates that this module is the desgin-under-test.  This is indicated
+    /// with a `sifive.enterprise.firrtl.MarkDUTAnnotation`.
+    bool isDut;
+
+    /// Indicates if this module is instantiated under the design-under-test.
+    InstanceInfo::LatticeValue underDut = {LatticeValue::Kind::Unknown};
+
+    /// Indicates if this module is instantiated under a layer.
+    InstanceInfo::LatticeValue underLayer = {LatticeValue::Kind::Unknown};
+  };
+
+  /// Return true if this module is the design-under-test.
+  bool isDut(FModuleOp op);
+
+  /// Return true if this module is instantiated under the design-under-test.
+  bool isUnderDut(FModuleOp op);
+
+  /// Return true if all instantiations of this module are under the
+  /// design-under-test.
+  bool isFullyUnderDut(FModuleOp op);
+
+  /// Return true if this module is instantiated under a layer block.
+  bool isUnderLayer(FModuleOp op);
+
+  /// Return true if all instantiations of this module are under layer blocks.
+  bool isFullyUnderLayer(FModuleOp op);
+
+private:
+  /// Internal mapping of operations to module attributes.
+  DenseMap<Operation *, ModuleAttributes> moduleAttributes;
+
+  /// Return the module attributes associated with a module.
+  const ModuleAttributes &getModuleAttributes(FModuleOp op);
+};
+
+#ifndef NDEBUG
+inline llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                     const InstanceInfo::LatticeValue &value) {
+  switch (value.kind) {
+  case InstanceInfo::LatticeValue::Kind::Unknown:
+    return os << "unknown";
+  case InstanceInfo::LatticeValue::Kind::Constant:
+    return os << "constant<" << (value.constant ? "true" : "false") << ">";
+  case InstanceInfo::LatticeValue::Kind::Mixed:
+    return os << "mixed";
+  }
+  return os;
+}
+#endif
+
+} // namespace firrtl
+} // namespace circt
+
+#endif // CIRCT_DIALECT_FIRRTL_FIRRTLINSTANCEINFO_H
--- a/lib/Analysis/FIRRTLInstanceInfo.cpp
+++ b/lib/Analysis/FIRRTLInstanceInfo.cpp
@@ -106,13 +106,15 @@ InstanceInfo::getModuleAttributes(FModuleOp op) {
 bool InstanceInfo::isDut(FModuleOp op) { return getModuleAttributes(op).isDut; }
 
 bool InstanceInfo::atLeastOneInstanceUnderDut(FModuleOp op) {
-  return getModuleAttributes(op).underDut.kind == LatticeValue::Mixed ||
+  return isDut(op) ||
+         getModuleAttributes(op).underDut.kind == LatticeValue::Mixed ||
          allInstancesUnderDut(op);
 }
 
 bool InstanceInfo::allInstancesUnderDut(FModuleOp op) {
   auto underDut = getModuleAttributes(op).underDut;
-  return underDut.kind == LatticeValue::Constant && underDut.constant;
+  return isDut(op) ||
+         (underDut.kind == LatticeValue::Constant && underDut.constant);
 }
 
 bool InstanceInfo::atLeastOneInstanceUnderLayer(FModuleOp op) {
--- a/lib/Analysis/FIRRTLInstanceInfo.cpp
+++ b/lib/Analysis/FIRRTLInstanceInfo.cpp
@@ -16,16 +16,19 @@
 #include "circt/Analysis/FIRRTLInstanceInfo.h"
 #include "circt/Dialect/FIRRTL/AnnotationDetails.h"
 #include "circt/Dialect/FIRRTL/FIRRTLInstanceGraph.h"
-#include "llvm/ADT/DepthFirstIterator.h"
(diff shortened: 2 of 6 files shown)


test/Analysis/firrtl-test-instance-info.mlir
```diff
@@ -0,0 +1,59 @@
...
+firrtl.circuit "Foo" {
+  firrtl.layer @A bind {
+  }
+  // CHECK:      firrtl.module @Corge
+  // CHECK-NEXT:   isDut: false
+  // CHECK-NEXT:   isUnderDut: false
+  // CHECK-NEXT:   isFullyUnderDut: false
+  // CHECK-NEXT:   isUnderLayer: true
+  // CHECK-NEXT:   isFullyUnderLayer: false
+  firrtl.module @Corge() {}
```
dtzSiFive, line 12: Consider making most of these modules private.  Once we land/start using multiple public modules, I'd guess this pass will want to walk down from all entry points not only a single one?

That said I know you prefer minimal tests so if you'd rather update it later (when adding support) that works for me too 👍 .

test/Analysis/firrtl-test-instance-info.mlir
```diff
@@ -0,0 +1,59 @@
...
+  firrtl.module @Baz() {}
+  // CHECK:      firrtl.module @Bar
+  // CHECK-NEXT:   isDut: true
+  // CHECK-NEXT:   isUnderDut: false
+  // CHECK-NEXT:   isFullyUnderDut: false
+  // CHECK-NEXT:   isUnderLayer: false
+  // CHECK-NEXT:   isFullyUnderLayer: false
+  firrtl.module @Bar() attributes {annotations = [{class = "sifive.enterprise.firrtl.MarkDUTAnnotation"}]} {
+    firrtl.instance baz interesting_name @Baz()
+    firrtl.instance qux interesting_name @Qux()
```
dtzSiFive, line 42: Love these extra names :grin:.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    // Set the baseline attributes.  This is what the attributes will be if
+    // there is no other, existing attributes.
+    ModuleAttributes attributes({/*isDut=*/isDut});
+    attributes.underDut.merge(frame.underDut);
+    attributes.underLayer.merge(frame.underLayer);
+
+    // Merge the baseline attributes with the existing attributes, if they
+    // exist.
+    auto it = moduleAttributes.find(moduleOp);
```
dtzSiFive, line 83: Consider something like `insert` to set them if they don't have values already, and if they do you can do the merge logic already having a handle to them from the first search.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+    if (it == moduleAttributes.end())
+      moduleAttributes[moduleOp] = attributes;
+    else {
+      auto &oldAttributes = it->getSecond();
+      oldAttributes.underDut.merge(attributes.underDut);
+      oldAttributes.underLayer.merge(attributes.underLayer);
+      attributes = oldAttributes;
```
dtzSiFive, line 90: This appears unused after the copy?

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+    for (auto *inst : *frame.node) {
+      auto underDut = frame.underDut || isDut;
+      auto underLayer = frame.underLayer ||
+                        inst->getInstance()->getParentOfType<LayerBlockOp>();
+      worklist.push_back({inst->getTarget(),
+                          /*underDut=*/underDut,
+                          /*underLayer=*/underLayer});
+    }
+  }
+  for (auto *node : llvm::depth_first(iGraph.getTopLevelNode())) {
```
dtzSiFive, line 102: This walk seems redundant / from an earlier iteration, can it now be deleted?

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+  /// Indicates that the current hierarchy is under a layer.
+  bool underLayer = false;
+};
+} // namespace
+
+InstanceInfo::InstanceInfo(Operation *op, mlir::AnalysisManager &am) {
+  auto &iGraph = am.getAnalysis<InstanceGraph>();
+
+  SmallVector<Frame> worklist({{iGraph.getTopLevelNode()}});
+  while (!worklist.empty()) {
```
prithayan, line 67: Isn't this also a DFS from the top level node ?

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+      auto &oldAttributes = it->getSecond();
+      oldAttributes.underDut.merge(attributes.underDut);
+      oldAttributes.underLayer.merge(attributes.underLayer);
+      attributes = oldAttributes;
+    }
+
+    for (auto *inst : *frame.node) {
+      auto underDut = frame.underDut || isDut;
+      auto underLayer = frame.underLayer ||
+                        inst->getInstance()->getParentOfType<LayerBlockOp>();
```
dtzSiFive, line 96: Modules are visited once per occurrence in the unfolded representation, if I'm getting this right.
And so modules with many instances will re-walk all of those instances looking for layer blocks.

Instead, it'd be better to compute non-trivial (if not wildly expensive) details only once per module (and its instances).

Perhaps when the worklist has a module you haven't visited yet, at that point you 1)check its annotations and 2)walk its instances (and combine with information it's being fed -- you could skip either if you felt like based on forwarded `underDut`/`underLayer`)), and otherwise just happily propagate.

lib/Analysis/FIRRTLInstanceInfo.cpp
prithayan, line 102: Why do we need a second traversal over the graph, can't we set this in the above traversal ?

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+  /// Indicates that the current hierarchy is under the DUT.
+  bool underDut = false;
+
+  /// Indicates that the current hierarchy is under a layer.
+  bool underLayer = false;
+};
+} // namespace
+
+InstanceInfo::InstanceInfo(Operation *op, mlir::AnalysisManager &am) {
```
prithayan, line 63: Is the `op` required here ?

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+      attributes.isDut = true;
+    }
+  }
+
+  LLVM_DEBUG({
+    llvm::dbgs() << "InstanceInfo Analysis Results:\n";
+    DenseSet<Operation *> visited;
+    for (auto *node : llvm::depth_first(iGraph.getTopLevelNode())) {
+      auto moduleOp = node->getModule();
+      if (!visited.insert(moduleOp.getOperation()).second)
```
dtzSiFive, line 117: Checking if already visited shouldn't be necessary, although using the variant of the iterator that stores the set externally is useful sometimes but this is debug code so whatever you like.

prithayan (commented):
Initial comments, will take a deeper look in a while.

dtzSiFive (commented):
Comments, almost done...

lib/Analysis/FIRRTLInstanceInfo.cpp
dtzSiFive: (it's DFS without tracking what's been visited already)

Consider RPO (inverse_post_order) since this is a forward dataflow over the instance graph (right?) so this visits a module only after visiting all the places it's instantiated, which might combine well with avoiding re-computing same things for a given module (since you'd only visit it once).

Might be missing something, but that seems the way to go?

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+      moduleAttributes[moduleOp] = attributes;
+    else {
+      auto &oldAttributes = it->getSecond();
+      oldAttributes.underDut.merge(attributes.underDut);
+      oldAttributes.underLayer.merge(attributes.underLayer);
+      attributes = oldAttributes;
+    }
+
+    for (auto *inst : *frame.node) {
+      auto underDut = frame.underDut || isDut;
```
prithayan, line 94: Seems like this should be `attributes.underDut` and same for the following `underLayer`.
```suggestion
      auto underDut = attributes.underDut || isDut;
```

prithayan (commented):
Another round of comments.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: This is entirely vestigial... Removed in: b120d1058

include/circt/Analysis/FIRRTLInstanceInfo.h
```diff
@@ -0,0 +1,115 @@
...
+
+    // Combine this lattice value with another lattice value.
+    void merge(LatticeValue that);
+
+    // Merge evidence that a property is true.
+    void merge(bool property);
+  };
+
+  struct ModuleAttributes {
+    /// Indicates that this module is the desgin-under-test.  This is indicated
```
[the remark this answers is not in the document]
seldridge: Fixed in: 3a4f7fd52

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: Let me see if I can rework this. Yes, this is a forward dataflow over the instance graph. It's pushing information from the parents to the children and then merging lattice values in the children based on this information. If the children are visited after all their parents, then the children can query the attributes of the parents and this will work the same way, but be way more efficient.

seldridge: This is reworked to use `inverse_post_order` in 6abbeb041. This is _dramatically_ simpler. Nice comments in this direction.

This also removes the need for the `struct Frame` then by querying the attributes of the parent.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: According to the docs [[1]], yes, it needs to have a constructor of `(Operation *)` or `(Operation *, AnalysisManager &)`. I tested this just now and the compiler will reject it without it.

Good eye, though. This is clearly unused in the body.

[1]: https://mlir.llvm.org/docs/PassManagement/#analysis-management

include/circt/Analysis/FIRRTLInstanceInfo.h
```diff
@@ -0,0 +1,115 @@
...
+
+namespace circt {
+namespace firrtl {
+
+using namespace detail;
+
+class InstanceInfo {
+
+public:
+  explicit InstanceInfo(Operation *op, mlir::AnalysisManager &am);
```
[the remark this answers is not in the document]
seldridge: I gave the docs a read before writing the pass. 😉

include/circt/Analysis/FIRRTLInstanceInfo.h
```diff
@@ -0,0 +1,115 @@
...
+#define CIRCT_DIALECT_FIRRTL_FIRRTLINSTANCEINFO_H
+
+#include "circt/Dialect/FIRRTL/FIRRTLOps.h"
+#include "circt/Support/LLVM.h"
+#include "mlir/Pass/AnalysisManager.h"
+
+namespace circt {
+namespace firrtl {
+
+using namespace detail;
```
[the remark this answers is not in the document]
seldridge: This was trying to get some things out of the `namespace detail` inside `FIRRTLInstanceInfo.h`. With the refactoring in 6abbeb041, this is now gone.

include/circt/Analysis/FIRRTLInstanceInfo.h
```diff
@@ -0,0 +1,115 @@
...
+    void merge(bool property);
+  };
+
+  struct ModuleAttributes {
+    /// Indicates that this module is the desgin-under-test.  This is indicated
+    /// with a `sifive.enterprise.firrtl.MarkDUTAnnotation`.
+    bool isDut;
+
+    /// Indicates if this module is instantiated under the design-under-test.
+    InstanceInfo::LatticeValue underDut = {LatticeValue::Kind::Unknown};
```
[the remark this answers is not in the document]
seldridge: I think this interpretation of `underDut` makes more sense.

lib/Analysis/TestPasses.cpp
```diff
@@ -176,6 +179,51 @@ void InferTopModulePass::runOnOperation() {
...
+               << "    isFullyUnderDut: " << iInfo.isFullyUnderDut(op) << "\n"
+               << "    isUnderLayer: " << iInfo.isUnderLayer(op) << "\n"
+               << "    isFullyUnderLayer: " << iInfo.isFullyUnderLayer(op)
+               << "\n";
+}
+
+void FIRRTLInstanceInfoPass::runOnOperation() {
+  auto &iInfo = getAnalysis<firrtl::InstanceInfo>();
+
+  getOperation()->walk([&](firrtl::FModuleOp op) { printInfo(op, iInfo); });
```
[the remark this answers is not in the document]
seldridge: I can't argue with this. Fixed in: 3538b70f8

test/Analysis/firrtl-test-instance-info.mlir
seldridge: Good suggestion. Changed in 706c9752b. I made the DUT private, too, but this may be wrong.

The lack of `private` was more motivated by simplifying the FileCheck check lines. I'll switch those to be slightly terser and match on only the symbol and avoid the regex the ignore private.

test/Analysis/firrtl-test-instance-info.mlir
seldridge: Will remove the `interesting_name` part of this. Gone in: 5e9cba729

lib/Analysis/TestPasses.cpp
```diff
@@ -176,6 +179,51 @@ void InferTopModulePass::runOnOperation() {
...
+  if (a)
+    return os << "true";
+  return os << "false";
+}
+
+static void printInfo(firrtl::FModuleOp op, firrtl::InstanceInfo &iInfo) {
+  OpPrintingFlags flags;
+  flags.skipRegions();
+  llvm::errs() << "  - operation: ";
+  op->print(llvm::errs(), flags);
```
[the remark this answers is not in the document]
seldridge: I liked this, too. I couldn't find a way to make it more terse like `llvm::errs() << flags << /* .. */` like I wanted, though.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: This is fixed after 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda and the module attributes are now merged per-module and then recorded in the `DenseMap moduleAttributes` exactly once.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: Refactored to avoid this in 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: It was totally redundant. 🙃 

This was removed in b120d10586222bc81c064b82cdd6dc8ee401dedc.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: Reworked to visit once-per-module in 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: 👍 Fixed in: d3e59a29b

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+const InstanceInfo::ModuleAttributes &
+InstanceInfo::getModuleAttributes(FModuleOp op) {
+  return moduleAttributes.find(op)->getSecond();
+}
+
+bool InstanceInfo::isDut(FModuleOp op) { return getModuleAttributes(op).isDut; }
+
+bool InstanceInfo::isUnderDut(FModuleOp op) {
```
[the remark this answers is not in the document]
seldridge: I'd opt against "maybe/may be" only because it implies the possibility of something being true. I.e., this seems closer to "unknown". Since I couldn't think of something pithy, I opted for more verbose names in 09b03f18a. This uses the verbose convention `atLeastOneInstance<property>` and `allInstances<property>`.

I'm open to alternatives, though.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+  auto underDut = getModuleAttributes(op).underDut;
+  return underDut.kind == LatticeValue::Mixed ||
+         (underDut.kind == LatticeValue::Constant && underDut.constant);
```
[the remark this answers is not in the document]
seldridge: I like the reuse of existing functions. 💯 Switched to this in: 632b419cd

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+InstanceInfo::InstanceInfo(Operation *op, mlir::AnalysisManager &am) {
+  auto &iGraph = am.getAnalysis<InstanceGraph>();
+
+  SmallVector<Frame> worklist({{iGraph.getTopLevelNode()}});
+  while (!worklist.empty()) {
+    auto frame = worklist.pop_back_val();
+    auto *node = frame.node;
+    auto moduleOp = node->getModule();
+
+    // Compute information about this instance.
```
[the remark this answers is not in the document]
seldridge: Good catch. This was fixed in: 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+    bool isDut = AnnotationSet(moduleOp).hasAnnotation(dutAnnoClass);
+
+    // Set the baseline attributes.  This is what the attributes will be if
+    // there is no other, existing attributes.
+    ModuleAttributes attributes({/*isDut=*/isDut});
```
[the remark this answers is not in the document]
seldridge: 👍 This was cleaned up by removing `struct Frame` in 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+  auto &iGraph = am.getAnalysis<InstanceGraph>();
+
+  SmallVector<Frame> worklist({{iGraph.getTopLevelNode()}});
+  while (!worklist.empty()) {
+    auto frame = worklist.pop_back_val();
+    auto *node = frame.node;
+    auto moduleOp = node->getModule();
+
+    // Compute information about this instance.
+    bool isDut = AnnotationSet(moduleOp).hasAnnotation(dutAnnoClass);
```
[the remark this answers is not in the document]
seldridge: The `inverse_post_order` version in 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda now avoids this. Good catch.

lib/Analysis/FIRRTLInstanceInfo.cpp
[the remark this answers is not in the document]
seldridge: Yep, is unused. Removed in the 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda refactor.

lib/Analysis/FIRRTLInstanceInfo.cpp
seldridge: Yes, this looks like a mistake. I'm surprised the tests didn't uncover this. 😕 

This code was rewritten in 6abbeb0410a5e5bfb972f1290218d6e2da7d0bda.

include/circt/Analysis/FIRRTLInstanceInfo.h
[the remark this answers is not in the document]
seldridge: I switched to some very exact language in 09b03f18a89cab75f3b553652b5ff85ed7b9d0a4. This language could hint the new interpretation of "under" isn't quite right. Specifically, is an instance of the DUT under the DUT? I went back and forth here and settled on it is under the DUT, i.e., the new interpretation makes sense.

Changed in 0d28c7aec.

WDYT?

seldridge:
Following up on an offhand remark from @rwy7, I do think that this can grow into computing edge-centric properties as well. I.e., this can get information about the instances and not just the modules/aggregate properties of all instances of modules. I'll leave this to future work as I don't want to overload this with too many things that don't have users.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,156 @@
...
+const InstanceInfo::ModuleAttributes &
+InstanceInfo::getModuleAttributes(FModuleOp op) {
+  return moduleAttributes.find(op)->getSecond();
+}
+
+bool InstanceInfo::isDut(FModuleOp op) { return getModuleAttributes(op).isDut; }
+
+bool InstanceInfo::isUnderDut(FModuleOp op) {
+  auto underDut = getModuleAttributes(op).underDut;
+  return underDut.kind == LatticeValue::Mixed ||
```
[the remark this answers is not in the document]
seldridge: Yeah, this should grow some of these helpers. After applying a comment from Prithayan to reuse the `allInstances*` functions, the need for these helpers is lessened in this PR.

I'll leave this to future work.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+      // Merge in attributes from instantiations one-by-one.
+      for (auto *useIt : modIt->uses()) {
+        auto parentAttrs =
+            moduleAttributes.find(useIt->getParent()->getModule())->getSecond();
+        // Merge underDut.
+        if (parentAttrs.isDut)
+          attributes.underDut.mergeIn({LatticeValue::Constant, true});
+        else
+          attributes.underDut.mergeIn(parentAttrs.underDut);
```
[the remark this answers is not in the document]
seldridge: Good catch. I had "fixed" this in https://github.com/llvm/circt/pull/7612/commits/0d28c7aec219522c3a02fea0ec190a5e241d751b by including the `isDut` value in the member functions that query the `moduleAttributes` (`atLeastOneInstanceUnderDut` and `allInstancesUnderDut`) However, you are right in that with the new interpretation of this that Will suggested, this is better to just set directly.

Changed in: d12fed29b.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+InstanceInfo::InstanceInfo(Operation *op, mlir::AnalysisManager &am) {
+  auto &iGraph = am.getAnalysis<InstanceGraph>();
+
+  // Visit modules in reverse post-order (visit parents before children) because
+  // information flows in this direction---the attributes of modules are
+  // determinend by their instantiations.
+  DenseSet<InstanceGraphNode *> visited;
+  for (auto *root : iGraph) {
+    for (auto *modIt : llvm::inverse_post_order_ext(root, visited)) {
```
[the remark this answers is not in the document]
dtzSiFive: Presumably reverse of post order VS post-order on the inverse graph.

Good question! RPO is classic/ideal here.

Without cycles I'm not sure it matters (but would want to stare at it more...), and with them we can't guarantee convergence in a single walk (FWIW).

Using RPO, might be simplest conceptually as it's what we all expect here anyway and is more familiar for this sort of thing.

I was reaching for RPO in my suggestion, tbh.

FWIW we use inv post order elsewhere so hopefully thats correct/intentional for the same reason mentioned above, but might be good to check or ping those authors :).

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -0,0 +1,128 @@
...
+        else
+          attributes.underDut.mergeIn(parentAttrs.underDut);
+        // Merge underLayer.
+        if (useIt->getInstance()->getParentOfType<LayerBlockOp>())
+          attributes.underLayer.mergeIn({LatticeValue::Constant, true});
+        else
+          attributes.underLayer.mergeIn(parentAttrs.underLayer);
+      }
+
+      // If attributes are unknown at this point, set default values.
```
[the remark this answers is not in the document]
seldridge: Good catch. Yes, this is really what happens if the module is uninstantiated. I reordered this logic to use an early exit for this case (and setting the attributes via reference) in 96652d024.

lib/Analysis/FIRRTLInstanceInfo.cpp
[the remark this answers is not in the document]
seldridge: Empirically, I think it's the same in how it's being used here. Possibly, `ReversePostOrder` is slightly cleaner as it avoids having to do the pattern of nesting an `inverse_post_order_ext` under iteration over the nodes in the graph. I.e., the below can be replaced with what follows it:

``` c++
DenseSet<InstanceGraphNode *> visited;
for (auto *root : iGraph) {
  for (auto *modIt : llvm::inverse_post_order_ext(root, visited)) {
    /* ... */
  }
}
```

``` c++
ReversePostOrder<InstanceGraph *> rpo(iGraph);
for (auto *modIt : rpo) {
  /* ... */
}
```

I switched over to this in 8f1cedfec.

lib/Analysis/FIRRTLInstanceInfo.cpp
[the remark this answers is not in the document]
seldridge: There are some helpers inspired from what IMCP's lattice value provides in https://github.com/llvm/circt/pull/7612/commits/7efde85842c240778618c89053c7f5b6e9fdadbb.

lib/Analysis/FIRRTLInstanceInfo.cpp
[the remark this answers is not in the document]
youngar: Chiming in a little late to this thread,  but yes RPO and IPO should both visit the "parent" modules before the "children" modules, so it doesn't really matter which you use, but we usually use IPO for two reasons:
1. RPO does a post-order walk, storing the elements to a SmallVector, and then walks it in reverse.  IPO might be slightly faster because it doesn't need to build a vector of all modules.
2. RPO will only visit modules which are reachable from the top level module (i.e. our instance graph trait implementation does not work well with multiple public modules), while the double loop used for IPO side steps this issue and visits all modules.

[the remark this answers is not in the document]
prithayan: Okay, thanks youngar , this clarifies my confusion of RPO vs IPO, sticking with IPO makes sense.

include/circt/Analysis/FIRRTLInstanceInfo.h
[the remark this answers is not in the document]
seldridge: After further thought on this, I changed to use "any" instead of "atLeastOne" in fa3de1e7d2456d9f01cbd9023d00a0ba985709c3. This tends to better match, for lack of better inspiration, functional programming style "any of" or "all of" methods. "any" is also much shorter.

lib/Analysis/FIRRTLInstanceInfo.cpp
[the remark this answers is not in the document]
seldridge: SGTM. Changed back in 3fbec15b7.

PR #7613, opened 2024-09-20
[FIRRTL] Allow local targets to be multiply-instantiated.

mikeurbach:
The existing path support was built up based on the assumption that
every target is unique. That is true for FIRRTL produced by standard
Chisel code, which elaborates unique modules for each instance. We
definitely don't want to limit ourselves to this world, and we should
support targeting things that are multiply instantiated when it is not
ambiguous what we refer to.

This patch relaxes the single-instantiation constraints for local
targets, which refer to a module or something inside a module,
regardless of how many times or at what paths that particular module
was instantiated.

This required a couple changes through the pipeline:

ResolvePaths already had an early exit for the local path case, but
this needed to come before the single-instantiation check.

LowerClasses needed a couple small changes to not enforce the
single-instantiation check in the local path case, and to build a
hierpath that just has a single element.

While this is not a new requirement, we can still get ambiguous local
targets, for instance from nested module prefixing. The error message
in LowerClasses for this case was made a little more clear.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -385,8 +385,8 @@ PathTracker::getOrComputeNeedsAltBasePath(Location loc, StringAttr moduleName,
       needsAltBasePath = true;
       break;
     }
-    // If there is more than one instance of this module, then the path
-    // operation is ambiguous, which is an error.
+    // If there is more than one instance of this module, and the target is
+    // non-local, then the path operation is ambiguous, which is an error.
     if (isNonLocal && !node->hasOneUse()) {
       auto diag = mlir::emitError(loc)
                   << "unable to uniquely resolve target due "
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -275,7 +275,8 @@ private:
   // and `owningModule`.
   FailureOr<bool> getOrComputeNeedsAltBasePath(Location loc,
                                                StringAttr moduleName,
-                                               FModuleOp owningModule);
+                                               FModuleOp owningModule,
+                                               bool isNonLocal);
   FModuleLike module;
 
   // Local data structures.
@@ -364,7 +365,8 @@ LogicalResult PathTracker::runOnModule() {
 
 FailureOr<bool>
 PathTracker::getOrComputeNeedsAltBasePath(Location loc, StringAttr moduleName,
-                                          FModuleOp owningModule) {
+                                          FModuleOp owningModule,
+                                          bool isNonLocal) {
 
   auto it = needsAltBasePathCache.find({moduleName, owningModule});
   if (it != needsAltBasePathCache.end())
@@ -385,7 +387,7 @@ PathTracker::getOrComputeNeedsAltBasePath(Location loc, StringAttr moduleName,
     }
     // If there is more than one instance of this module, then the path
     // operation is ambiguous, which is an error.
-    if (!node->hasOneUse()) {
+    if (isNonLocal && !node->hasOneUse()) {
       auto diag = mlir::emitError(loc)
                   << "unable to uniquely resolve target due "
                      "to multiple instantiation";
@@ -517,8 +519,8 @@ PathTracker::processPathTrackers(const AnnoTarget &target) {
     }
 
     // Check if we need an alternative base path.
-    auto needsAltBasePath =
-        getOrComputeNeedsAltBasePath(op->getLoc(), moduleName, owningModule);
+    auto needsAltBasePath = getOrComputeNeedsAltBasePath(
+        op->getLoc(), moduleName, owningModule, hierPathOp);
     if (failed(needsAltBasePath)) {
       error = true;
       return false;
@@ -528,6 +530,10 @@ PathTracker::processPathTrackers(const AnnoTarget &target) {
     // to the start of the annotation's NLA.
     InstanceGraphNode *node = instanceGraph.lookup(moduleName);
     while (true) {
+      // If it's not a non-local target, we don't have to append anything.
+      if (!hierPathOp)
+        break;
+
       // If we get to the owning module or the top, we're done.
       if (node->getModule() == owningModule || node->noUses())
         break;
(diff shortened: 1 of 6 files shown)


lib/Dialect/FIRRTL/Transforms/ResolvePaths.cpp
```diff
@@ -41,6 +41,12 @@ struct PathResolver {
                                 const AnnoPathValue &target,
                                 FlatSymbolRefAttr &result) {
 
+    // If the path is empty then this is a local reference and we should not
+    // construct a HierPathOp.
+    if (target.instances.empty()) {
+      return success();
+    }
```
uenoku, line 48: nit:
```suggestion
    if (target.instances.empty())
      return success();
```

uenoku (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -385,7 +387,7 @@ PathTracker::getOrComputeNeedsAltBasePath(Location loc, StringAttr moduleName,
     }
     // If there is more than one instance of this module, then the path
     // operation is ambiguous, which is an error.
```
seldridge, line 389: Can the PR also update this comment which is now stale?

mikeurbach: Oops, yep will update.

mikeurbach:
> This does mean that this approach can work with pre-deduplicated circuits. However, it cannot, and should not, work with circuits that go through deduplication in CIRCT (which will necessarily create non-local annotations).

Hmm, I think I need to understand more about what won't work here, since my hope is to start supporting pre-deduplicated circuits in some cases.

Here's how I would represent what CIRCT main supports:

* Local targets: check the target is singly instantiated and uniquely identifiable, make a non-local annotation, and the rest of the pipeline handles it
* Non-local targets: check the target is singly instantiated and uniquely identifiable, make a non-local annotation, and the rest of the pipeline handles it

This works for both cases in practice because standard Chisel produces duplicated circuits in the FIRRTL. If the FIRRTL is pre-deduplicated, the local target case can fail immediately in ResolvePaths.

What I'm trying to change here is the Local targets case--rather than checking the target is singly instantiated and making a non-local annotation, it makes a local annotation. This is intended to be similar to how references worked with the original EmitOMIR flow--if something like dedup needs to turn a local annotation into a non-local annotation, the pass is free to do that, and the rest of the pipeline should handle it or error out in the face of ambiguity, similarly to what EmitOMIR does, but using a slightly different mechanism with distinct attributes.

This is intended to make _some_ local targets now possible, for the case where the FIRRTL comes pre-deduplicated, and those targets don't become ambiguous at some point in the pipeline. There are still cases where we wouldn't fail at ResolvePaths, but transformations would lead to a failure later (with a decent error message).

Maybe I'm missing something, but can you share more about what can't work? The existing path support was built up based on the same OMIRTrackerAnnotations we previously used with EmitOMIR, and can handle both local and non-local annotations. This patch is just allowing local annotations to be used directly when requested from the input, and letting the later stages of the pipeline ensure they are unambiguous after transforms like PrefixModules and Dedup.

mikeurbach:
> we would need to move to a representation that does not use annotations, e.g., an intrinsic.

And yes, I think we do need to move to a different representation in the long term, but we went the route of reusing the OMIRTrackerAnnotations when we introduced paths to FIRRTL so we could resolve them into something the rest of the pipeline already works with in ResolvePaths. So this PR is about allowing us to resolve local annotations, which the existing flow should support (or give an error if it can't). I do want to come up with a better representation, but we'll need to teach lots of passes that currently work with annotations to keep that representation up to date.

test/Dialect/FIRRTL/lower-classes.mlir
```diff
@@ -471,3 +472,25 @@ firrtl.circuit "PathTargetReplaced" {
...
+    // CHECK: om.path_create instance %basepath [[MODULE_NLA]]
+    %0 = firrtl.path instance distinct[0]<>
+
+    // CHECK: om.path_create reference %basepath [[WIRE_NLA]]
+    %1 = firrtl.path reference distinct[1]<>
+
+    firrtl.instance child0 @Child()
+    firrtl.instance child1 @Child()
+  }
+}
```
seldridge, line 496: To elaborate on my comment... What I'm concerned about is if there is a way to rewrite this circuit such that `@Child` is duplicated into `@Child0` and `@Child1`. What are the trackers used then? After deduplication, it would seem like this would result in non-local annotation trackers which would then trip the error checking logic.

mikeurbach: I see. The main way I have noticed something like this happening is when PrefixModules clones modules. You can set up an example that would cause `@Child` to `@Child0` and `@Child1`, each with a local annotation with the same `distinct` ID. This would then error out in LowerClasses here: https://github.com/llvm/circt/blob/afd61f26eff984daf8807e3a57d79a4a93705f31/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp#L585-L586

Which is equivalent to this existing check: https://github.com/llvm/circt/blob/afd61f26eff984daf8807e3a57d79a4a93705f31/lib/Dialect/FIRRTL/Transforms/EmitOMIR.cpp#L650-L654

Note that PrefixModules happens after Dedup, but it might be possible to hit a scenario where Dedup makes local annotations non-local, and that could potentially fail the single-instantiation check, which I think is what you're referring to.

I guess my thought is with this PR, we can support _some_ local annotations end-to-end, and in the cases where it doesn't work, we'd hit a well defined error case. So, this should support some cases, but it shouldn't allow any ambiguous cases through; those should still error in the expected ways.

mikeurbach: Does this help assuage your concerns? If the current state is fine, I'll go ahead and merge this PR.

seldridge: I have zero problems with this PR! Good to go!

I'm bringing up this problem with dedup because it's a general problem with any annotation representation. If this fix gets you further, please go for it. 👍 

Annotations are generally problematic and we're slowly tackling getting off them. You're aware of the problems.

mikeurbach: Yep makes sense, thanks. I think with this we'll at least be at parity with what we could do with the tracker annotations and EmitOMIR. Excited to design a more dataflow-y representation for paths.

PR #7614, opened 2024-09-20
[MooreToCore] Fix the stack error caused by the uarray reftype.

hailongSun2000:
I attempted to convert `moore.uarray` into `hw.array`. However, there is an example that will trigger an error: 
```
error: 'llhd.sig.array_slice' op failed to verify that width of result type has to be smaller than or equal to the input type
  assign uarr[0][2] = 'b0;
         ^
/home/phoenix/work/HDLBits/uarray.sv:3:10: note: see current operation: %5 = "llhd.sig.array_slice"(%3, %4) : (!hw.inout<array<2xarray<4xarray<4xi2>>>>, i1) -> !hw.inout<array<4xarray<4xi2>>>
```
Case:
```
module UnpackedArray;
  logic [3:0][1:0] uarr [0:1][0:3];
  assign uarr[0][2] = 'b0;
endmodule
```

And due to the uarray type like `reg [31:0] Memory[0:31];` will lead to **stack error,** so I decide to convert `moore.uarray` into `hw.uarray`.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -1247,6 +1247,14 @@ static void populateTypeConversion(TypeConverter &typeConverter) {
                               type.getSize());
   });
 
+  // FIXME: In order to avoid the stack error caused by the conversion of
+  // "RefType" which can return "{}", such as "ref<uarray<4xi4>>", we can
+  // convert "moore::uarray" into "hw::array" temporarily.
+  typeConverter.addConversion([&](UnpackedArrayType type) {
+    return hw::UnpackedArrayType::get(
+        typeConverter.convertType(type.getElementType()), type.getSize());
+  });
+
   typeConverter.addConversion([&](StructType type) {
     SmallVector<hw::StructType::FieldInfo> fields;
     for (auto field : type.getMembers()) {


fabianschuiki:
Hey @hailongSun2000, how do the changes in #7631 affect this PR? The comment in the code makes it sound like this tackles a similar problem.

maerhart:
The `moore.variable` op does not contraint the type (as per it's verifier or the ref type verifier). Any unexpected variable type (i.e., any type not having a registered type conversion in this pass) would yield a null-type in the variable op conversion which causes the stack trace. My PR added a check for that to not apply the pattern in such a case rather than crashing. This PR, on the other hand, adds lowering support for the `uarray` type. So, if we want to lower unpacked arrays to `hw.array`, this PR still applies (I'd change the title/commit message though).

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1247,6 +1247,14 @@ static void populateTypeConversion(TypeConverter &typeConverter) {
                               type.getSize());
   });
 
+  // FIXME: In order to avoid the stack error caused by the conversion of
+  // "RefType" which can return "{}", such as "ref<uarray<4xi4>>", we can
+  // convert "moore::uarray" into "hw::array" temporarily.
+  typeConverter.addConversion([&](UnpackedArrayType type) {
+    return hw::UnpackedArrayType::get(
+        typeConverter.convertType(type.getElementType()), type.getSize());
```
maerhart, line 1255: We should check if the type returned by `convertType` is NULL and in that case forward the NULL rather than calling `UnpackedArrayType::get` with it (which could cause another segfault).

hailongSun2000:
> Hey @hailongSun2000, how do the changes in #7631 affect this PR? The comment in the code makes it sound like this tackles a similar problem.

This PR (#7631) is more thoughtful. I'll close my PR.

lib/Conversion/MooreToCore/MooreToCore.cpp
hailongSun2000: I quite agree with you. I checked the location of the segfault, and all point to `moore.variable`. I think I can close this PR, because no operation can handle `hw::UnpackedArrayType` now.

PR #7615, opened 2024-09-23
[LLHD][TCM] Simplify region directly

maerhart:
The simple canonicalizer does not do this simplification for us if too many blocks have to be merged.


fabianschuiki (approved):
LGTM

PR #7616, opened 2024-09-23
[LLHD] Add DesequentializationPass

maerhart:
There are still a lot of cases that are not supported because SV can represent all kind of weird things, but this should be enough to lower Snitch (there are 2/22 processes that have CFG loops that we need to unroll before I can test deseq on them, but from just looking at them it should work for them as well).
The two processes with async resets might not be lowered correctly for arcilator simulation because `seq.compreg` and arcilator do not support async resets yet.


lib/Dialect/LLHD/Transforms/DesequentializationPass.cpp
```diff
@@ -0,0 +1,721 @@
...
+    if (triggers.size() > 2 || triggers.empty())
+      return WalkResult::interrupt();
+
+    // TODO: add support
+    if (triggers[0].clocks.size() != 1 || triggers[0].clocks.size() != 1)
+      return WalkResult::interrupt();
+
+    // TODO: add support
+    if (triggers[0].kinds[0] == Trigger::Kind::Edge)
+      return WalkResult::interrupt();
```
fabianschuiki, line 647: Nice, love the defensive approach :+1:

lib/Dialect/LLHD/Transforms/DesequentializationPass.cpp
```diff
@@ -0,0 +1,721 @@
...
+
+    Value clock = builder.create<seq::ToClockOp>(loc, triggers[0].clocks[0]);
+    Value reset, resetValue;
+
+    if (triggers[0].kinds[0] == Trigger::Kind::NegEdge)
+      clock = builder.create<seq::ClockInverterOp>(loc, clock);
+
+    if (triggers[0].enable)
+      clock = builder.create<seq::ClockGateOp>(loc, clock, triggers[0].enable);
```
fabianschuiki, line 656: Nice idea to use clock inverters and clock gates to materialize the different trigger conditions :+1:

lib/Dialect/LLHD/Transforms/DesequentializationPass.cpp
```diff
@@ -0,0 +1,721 @@
...
+      // TODO: add support
+      if (triggers[1].kinds[0] == Trigger::Kind::Edge)
+        return WalkResult::interrupt();
+
+      // TODO: add support
+      if (triggers[1].enable)
+        return WalkResult::interrupt();
+
+      reset = triggers[1].clocks[0];
+      resetValue = op.getValue();
```
fabianschuiki, line 672: Do the reset and next value get merged into a single drive value, with a multiplexer distinguishing between the two?

test/Dialect/LLHD/Transforms/desequentialization.mlir
```diff
@@ -0,0 +1,111 @@
...
+  // CHECK: [[V2:%.+]] = seq.compreg %false{{.*}}, [[V1]]
+  // CHECK: llhd.drv [[SIG]], [[V2]] after
+  %time = llhd.constant_time <0ns, 1d, 0e>
+  %false = hw.constant false
+  %true = hw.constant true
+  llhd.process {
+    cf.br ^bb1
+  ^bb1:
+    %p1 = llhd.prb %clk : !hw.inout<i1>
+    llhd.wait (%p1 : i1), ^bb2
```
fabianschuiki, line 21: Technically this wait will block indefinitely, since `%p1` never changes its value. Do we need to consider this in the deseq pass? The Verilog frontend would insert an `llhd.prb` outside the process and wait on that value.

test/Dialect/LLHD/Transforms/desequentialization.mlir
```diff
@@ -0,0 +1,111 @@
...
+  // CHECK: llhd.drv [[SIG]], [[V4]] after
+  %time = llhd.constant_time <0ns, 1d, 0e>
+  %false = hw.constant false
+  %true = hw.constant true
+  llhd.process {
+    cf.br ^bb1
+  ^bb1:
+    %p1 = llhd.prb %clk : !hw.inout<i1>
+    %r1 = llhd.prb %rst : !hw.inout<i1>
+    llhd.wait (%p1 : i1), ^bb2
```
fabianschuiki, line 99: I think this wait needs to also observe the reset in order to make the reset async. Otherwise this would be more like a sync reset.

fabianschuiki (approved):
This is absolutely awesome! :partying_face: The deseq pass does some really heavy lifting, which is fantastic. I'm wondering if in the future we can move some of the work done in the `DnfAnalyzer` into dedicated ops and canonicalizers.

lib/Dialect/LLHD/Transforms/DesequentializationPass.cpp
```diff
@@ -0,0 +1,721 @@
...
+            if (truthTable[l][i] != truthTable[l][k])
+              dontCare[l].setBit(i);
+          }
+        }
+      }
+    }
+  }
+
+  void computeTruthTable() {
+    uint64_t numEntries = 1 << primitives.size();
```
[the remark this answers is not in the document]
maerhart: That's a good idea. I could also add it as a pass option with 8 as default value.

lib/Dialect/LLHD/Transforms/DesequentializationPass.cpp
maerhart: Yes, the TemporalCodeMotion pass does that. There is usually one `llhd.drv` operation in some BB doing the reset and another `llhd.drv` in another BB doing the regular update. TCM adds the condition under which each of those is reached as the drive condition and then also merges the two drives by merging the drive conditions and muxing the data value.

PR #7617, opened 2024-09-23
[LLHD] Make process lowering best-effort and allow constants from outside the region

maerhart:


lib/Dialect/LLHD/Transforms/ProcessLoweringPass.cpp
```diff
@@ -38,31 +41,46 @@ static LogicalResult isProcValidToLower(llhd::ProcessOp op) {
...
-    if (!isa<llhd::HaltOp>(op.getBody().back().getTerminator()))
-      return op.emitOpError("during process-lowering: entry block is required "
-                            "to be terminated by llhd.halt");
+    if (!isa<llhd::HaltOp>(op.getBody().back().getTerminator())) {
+      LLVM_DEBUG({
+        llvm::dbgs()
+            << "entry block is required to be terminated by llhd.halt\n";
+      });
+      return failure();
+    }
```
fabianschuiki, line 50: Is it legal to inline a process with a `halt` terminator into the parent module? Wouldn't the process only execute once, but the inlined version would be continuously executed?

fabianschuiki (approved):
Really cool :sunglasses:

lib/Dialect/LLHD/Transforms/ProcessLoweringPass.cpp
maerhart: That's right, let me remove this case. Not sure why it was there to begin with.

PR #7619, opened 2024-09-24
[DC] Adding `dc-print-dot` pass

luisacicolini:
This PR adds a new `dc-print-dot` pass that emulates `handshake-print-dot`. The main difference is that it produces dataflow graphs highlighting the separation between data and control, namely [DC values and tokens](https://circt.llvm.org/docs/Dialects/DC/).


include/circt/Dialect/DC/DCPasses.h
```diff
@@ -28,4 +33,4 @@ std::unique_ptr<mlir::Pass> createDCDematerializeForksSinksPass();
 } // namespace dc
 } // namespace circt
 
-#endif // CIRCT_DIALECT_DC_DCPASSES_H
+#endif // CIRCT_DIALECT_DC_DCPASSES_H
```
TaoBi22, line 36: Nit: newline at end of files

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,283 @@
...
+    return DotNode{.nodeType = "pack",
+                   .incoming = valueToName(op.getOperands(), valuesMap, false),
+                   .outgoing = valueToName(op.getResults(), valuesMap, false)};
+  if (auto castOp = llvm::dyn_cast<dc::ToESIOp>(op))
+    return DotNode{.nodeType = "toESI",
+                   .incoming = valueToName(op.getOperands(), valuesMap, false),
+                   .outgoing = valueToName(op.getResults(), valuesMap, false)};
+  if (auto castOp = llvm::dyn_cast<dc::UnpackOp>(op))
+    return DotNode{.nodeType = "unpack",
+                   .incoming = valueToName(op.getOperands(), valuesMap, false),
```
TaoBi22, line 122: Am I correct that the only thing that varies here between the DotNodes that are being returned is nodeType? It might be worth just letting these if statements define what nodeType is, and then feed that to just one call to the DotNode constructor which is probably a lot more concise

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,283 @@
...
+      currentMap.push_back({v, name});
+    }
+  }
+  return res;
+}
+
+DotNode createDCNode(
+    Operation &op,
+    llvm::SmallVector<std::pair<mlir::Value, std::string>> &valuesMap) {
+  if (auto castOp = llvm::dyn_cast<dc::ForkOp>(op))
```
TaoBi22, line 76: Nit: castOp never seems to be used so you can just do
```suggestion
  if (isa<dc::ForkOp>(op))
```

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,283 @@
...
+    os << "}";
+    outfile.close();
+  }
+};
+} // namespace
+
+std::unique_ptr<mlir::OperationPass<mlir::ModuleOp>>
+circt::dc::createDCDotPrintPass() {
+  return std::make_unique<DCDotPrintPass>();
+}
```
TaoBi22, line 283: nit: newline at EOF

TaoBi22 (commented):
Not super knowledgeable on DC so haven't looked in too much depth but just a few nits!

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+  }
+  return res;
+}
+
+DotNode createDCNode(
+    Operation &op,
+    llvm::SmallVector<std::pair<mlir::Value, std::string>> &valuesMap) {
+  if (auto castOp = llvm::dyn_cast<dc::ForkOp>(op))
+    return DotNode{.nodeType = "fork",
```
uenoku, line 77: drive-by-comment: This designated initialization (https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers) is c++ 20 feature so please avoid that. Also you might want to use `op.getName().stripDialect()` to get `nodeType` without hardcoding operation names here.

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+namespace {
+/// Emit the dot nodes
+struct DCDotPrintPass : public circt::dc::impl::DCDotPrintBase<DCDotPrintPass> {
+  void runOnOperation() override {
+
+    ModuleOp op = getOperation();
+
+    std::error_code ec;
+    llvm::raw_fd_ostream outfile("dc-out.dot", ec);
```
Dinistro, line 173: It might be beneficial when this file name could be provided as a pass parameter

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,238 @@
...
+    mlir::raw_indented_ostream os(outfile);
+
+    llvm::SmallVector<std::pair<mlir::Value, std::string>> valuesMap;
+    llvm::SmallVector<DotNode> nodes;
+
+    auto &moduleOps = op->getRegion(0).getBlocks();
+    // getBlock()->getOps<hw::HWModuleOp>();
```
Dinistro, line 180: Nit: Debug remnant?

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,238 @@
...
+    tokenFlag = true;
+
+  DotNode n = {op.getName().stripDialect().str(),
+               valueToName(op.getOperands(), valuesMap, false),
+               valueToName(op.getOperands(), valuesMap, tokenFlag)};
+
+  return n;
+}
+
+DotNode createCombNode(
```
Dinistro, line 88: Nit: A comment for this function would be nice

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+using namespace circt;
+using namespace dc;
+using namespace mlir;
+
+using ValueMap = llvm::ScopedHashTable<mlir::Value, std::string>;
+
+// stores operand and results for each node in the dot graph
+struct DotNode {
+  std::string nodeType;
```
Dinistro, line 43: Nit: It might be a bit nicer and more efficient to use an enum for this instead. This then ofc. requires some stringify function.

lib/Dialect/DC/Transforms/Analysis.cpp
Dinistro: This was not resolved. You might have some auto format setting that reverts this change. IIRC, VS code has some odd behavior sometimes.

lib/Dialect/DC/Transforms/Analysis.cpp
Dinistro, line 1: I would suggest to rename this to `DCPrintDot.cpp` or something along these lines.

Dinistro (commented):
Nice work :) I dropped a few nit comments mainly regarding style. I'm aware that the handshake pendant also violates many of the comments I've added, due to it being quite a relic.

I think this should probably be reviewed by @mortbopet .

luisacicolini:
thanks a lot for your review! I fixed most minor comments and will fix the remaining ones as soon as possible :)

lib/Dialect/DC/Transforms/Analysis.cpp
luisacicolini: Thank you! Would it be better to pass it as a parameter or just print to std out (as for example [HW](https://github.com/llvm/circt/blob/main/lib/Dialect/HW/Transforms/HWPrintInstanceGraph.cpp) does)?

lib/Dialect/DC/Transforms/Analysis.cpp
Dinistro: That's what I had in mind, yes.

llvm
Dinistro, line 1: You commited a submodule bump. This is probably not intended and might explain the CI failures.

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,338 @@
...
+using namespace circt;
+using namespace dc;
+using namespace mlir;
+
+using ValueMap = llvm::ScopedHashTable<mlir::Value, std::string>;
+
+namespace {
+
+/// all Comb and DC nodetypes
+enum NodeType {
```
Dinistro, line 41:
```suggestion
enum class NodeType {
```
C enums should be avoided as they allow some weird conversions and do pollute namespaces. Note that changing this to `enum class` will force you to fully qualify later usages with `NodeType::`.

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,338 @@
...
+  addOp,
```
Dinistro, line 42: Nit: We usually start enum members with a capital letter, similar to classes.
```suggestion
  AddOp,
```

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,338 @@
...
+  case sourceOp:
+    return "source";
+  case toESIOp:
+    return "toESI";
+  case unpackOp:
+    return "unpack";
+  case null:
+    return "null";
+  default:
+    return "Unknown NodeType";
```
Dinistro, line 122:
```suggestion
```
Not adding a default type is preferred, as this ensures that you get a compiler warning when an enum member is not covered.

lib/Dialect/DC/Transforms/Analysis.cpp
Dinistro: Probably pass it as a parameter.

lib/Dialect/DC/Transforms/Analysis.cpp
```diff
@@ -0,0 +1,238 @@
...
+  llvm::SmallVector<std::pair<mlir::Value, std::string>> incoming;
+  llvm::SmallVector<std::pair<mlir::Value, std::string>> outgoing;
+};
+
+// gives unique name to each value in the graph
+llvm::SmallVector<std::pair<mlir::Value, std::string>>
```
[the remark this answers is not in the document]
Dinistro: Make sure to cover all such occurrences by grepping for them.

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+/// gives a unique name to each value in the graph
+llvm::SmallVector<std::pair<mlir::Value, std::string>> static valueToName(
+    const llvm::SmallVector<mlir::Value> &values,
+    llvm::SmallVector<std::pair<mlir::Value, std::string>> &currentMap,
+    bool tokenFlag) {
+  SmallVector<std::pair<mlir::Value, std::string>> res;
+  for (auto [i, v] : llvm::enumerate(values)) {
+    auto found = false;
+    for (const auto &cm : currentMap) {
```
Dinistro, line 142: Nit: you can probably use structured bindings as well here.
```suggestion
    for (const [key, value] &cm : currentMap) {
```

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,338 @@
...
+        }
+        }
+      });
+  return n;
+}
+
+namespace {
+/// Emit the dot nodes
+struct DCDotPrintPass : public circt::dc::impl::DCDotPrintBase<DCDotPrintPass> {
+  DCDotPrintPass(llvm::raw_ostream &os) : os(os) {}
```
Dinistro, line 269:
```suggestion
  explicit DCDotPrintPass(llvm::raw_ostream &os) : os(os) {}
```
Nit: Without this, C++ allow to use this constructor for implicit type conversion. Fantastic language design, btw ;P

Dinistro (commented):
Thanks for addressing the comments, dropped a few more nits, but that should be ready soonish 🙂

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,346 @@
+//===- DCPrintDot.cpp - Analysis Pass -----------------------------*- C++
+//-*-===//
```
Dinistro, line 2: Nit: Make sure that this doesn't get wrapped. Same for other banners like this.

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,346 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Contains the definitions of the Analysis pass.
```
Dinistro, line 10: Nit: Outdated comment.

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,346 @@
...
+using namespace dc;
+using namespace mlir;
+
+using ValueMap = llvm::ScopedHashTable<mlir::Value, std::string>;
+
+namespace {
+
+/// all Comb and DC nodetypes
+class NodeType {
+  public:
```
Dinistro, line 42: Sorry for the confusion, I was referring to an enum class like this:

```
enum class NodeType {
  AddOp,
  AndOp,
  ...
}
```

Dinistro (commented):
Few more comments, then this will be ready to be shipped. Morten isn't around so I can accept it afterwards.

Dinistro:
something with the submidule is still broken, btw

lib/Dialect/DC/Transforms/DCPrintDot.cpp
luisacicolini: oh sorry! got it now, thanks

luisacicolini:
thanks! llvm is now back - now fixing builod issues

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,335 @@
...
+        }
+        }
+      });
+  return n;
+}
+
+namespace {
+/// Emit the dot nodes
+struct DCDotPrintPass : public circt::dc::impl::DCDotPrintBase<DCDotPrintPass> {
+  DCDotPrintPass(raw_ostream &os) : os(os) {}
```
Dinistro, line 266:
```suggestion
  explicit DCDotPrintPass(raw_ostream &os) : os(os) {}
```

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,335 @@
...
+/// stores operand and results for each node in the dot graph
+struct DotNode {
+  NodeType nodeType;
+  SmallVector<std::pair<mlir::Value, std::string>> incoming;
+  SmallVector<std::pair<mlir::Value, std::string>> outgoing;
+};
+} // namespace
+
+/// gives a unique name to each value in the graph
+SmallVector<std::pair<mlir::Value, std::string>> static valueToName(
```
Dinistro, line 132:
```suggestion
static SmallVector<std::pair<mlir::Value, std::string>> static valueToName(
```

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,335 @@
...
+        name = "token_" + std::to_string(currentMap.size());
+      res.push_back({v, name});
+      currentMap.push_back({v, name});
+    }
+  }
+  return res;
+}
+
+/// creates node in the dataflow graph for DC operations
+DotNode
```
Dinistro, line 157:
```suggestion
static DotNode
```

lib/Dialect/DC/Transforms/DCPrintDot.cpp
```diff
@@ -0,0 +1,335 @@
...
+      .Case([&](dc::SinkOp) { n.nodeType = NodeType::SinkOp; })
+      .Case([&](dc::SourceOp) { n.nodeType = NodeType::SourceOp; })
+      .Case([&](dc::ToESIOp) { n.nodeType = NodeType::ToESIOp; })
+      .Case([&](dc::UnpackOp) { n.nodeType = NodeType::UnpackOp; });
+
+  return n;
+}
+
+/// creates node in the dataflow graph for Comb operations
+DotNode
```
Dinistro, line 186:
```suggestion
static DotNode
```

Dinistro (approved):
Thanks for addressing all the comment. 
LGTM % final nits.

luisacicolini:
Thanks again for your reviews! Everything should be fixed and working now: ready to merge from my side (I have no write permission) :) 
CC: @Dinistro @TaoBi22

TaoBi22:
Cool, I can hit merge for you!

PR #7624, opened 2024-09-24
[Moore] Drop named_constant op in favor of dbg.variable

fabianschuiki:
Remove the `NamedConstantOp` and replace its uses with `VariableOp` from the debug dialect.

The op was originally added to track the value of constant parameters, localparams, and specparams in the IR. In ImportVerilog, such parameters would generate a corresponding `named_constant` op and all references to the parameter by name would be replaced with the `named_constant`'s result.

This doesn't really work well for parameters defined outside a module, such as in packages or at the root of the Verilog source file. (Modules are isolated from above, preventing the use of `named_constant`s from outside the module.) Therefore expressions would generally fall back to materializing constants directly where they were used.

Since the named constant ops are only there to track a constant value in the IR for the user's debugging convenience, using the debug dialect directly feels a lot more appropriate.


hailongSun2000 (approved):
LGTM :tada:! It's so perfect if we can implement a case like:
```
parameter P = 32;
module Foo;
  localparam  LP = P;
endmodule
```

fabianschuiki:
@hailongSun2000 `circt-verilog --parse-only -g` produces this:
```mlir
module {
  %0 = moore.constant 32 : l32
  dbg.variable "P", %0 : !moore.l32
  moore.module @Foo() {
    %1 = moore.constant 32 : l32
    dbg.variable "LP", %1 : !moore.l32
    moore.output
  }
}
```
:smiley:

hailongSun2000:
So nice!

hailongSun2000 (approved):

uenoku (approved):
Awesome that Debug works well here!

PR #7625, opened 2024-09-24
[MooreToCore] Alternative conditional lowering and type conversion fixes

maerhart:
* Alternative conditional op lowering to a mux if there are no relevant side-effects in the regions
* Add missing type conversions (mostly target types not marked legal and target types with nested source types not supported). This is necessary to make the conditional lowering work because the dialect conversion framework pushes an InlineBlockRewrite to the queue with triggers a signature conversion call to the function body the conditional op is in, which failed because the target types were not marked legal.
* Use the type converters isLegal helpers to determine dynamic legality instead of a custom solution.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -1156,6 +1156,40 @@ struct ConditionalOpConversion : public OpConversionPattern<ConditionalOp> {
     // evaluated and merged with the appropriate lookup table. See documentation
     // for `ConditionalOp`.
     auto type = typeConverter->convertType(op.getType());
+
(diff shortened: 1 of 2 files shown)


fabianschuiki (approved):
LGTM! The `isLegal` query is really neat and cleans stuff up very nicely.

hailongSun2000 (commented):
I have a personal question regarding `typeConverter.addConversion([&](hw::InOut/Array/StructType type)...`. If we convert `moore.ref/array/struct` to `hw.inout/array/struct`, would we get an error type like `hw.inout<moore.i32>` :thinking:?

hailongSun2000:
I understand these callbacks but sorry for not realizing which special case will trigger this error :sweat_smile:.

PR #7631, opened 2024-09-25
[MooreToCore] Return early on non-convertable type

### LINKED ISSUE #7627: [MooreToCore] Unpacked array causes crash
The following input crashes the MooreToCore conversion. Might be fixed by #7625.
```mlir
module {
  moore.module @top() {
    %mem = moore.variable : <uarray<1024 x l8>>
    %a = moore.variable : <l8>
    moore.procedure initial {
      %0 = moore.read %mem : <uarray<1024 x l8>>
      %1 = moore.extract %0 from 123 : uarray<1024 x l8> -> l8
      moore.blocking_assign %a, %1 : l8
      moore.return
    }
    moore.output
  }
}
```

maerhart:
Fixes #7627

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -472,6 +472,8 @@ struct VariableOpConversion : public OpConversionPattern<VariableOp> {
                   ConversionPatternRewriter &rewriter) const override {
     Location loc = op.getLoc();
     Type resultType = typeConverter->convertType(op.getResult().getType());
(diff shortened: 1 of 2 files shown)


hailongSun2000 (approved):
LGTM!
I do the same thing on this PR(https://github.com/llvm/circt/pull/7614).

maerhart:
Thanks for the review @hailongSun2000!
It's not the same thing because the variable op might have other return types as well. But in that case I should probably change the regression test here because that'd otherwise break your PR.

PR #7632, opened 2024-09-25
[ImportVerilog] Add support for elaboration system tasks

fabianschuiki:
Simply skip the elaboration system tasks `$info`, `$warning`, `$error`, and `$fatal` outside procedural code. Slang already processes these tasks and generates the corresponding diagnostics for us.


maerhart (approved):
LGTM! Nice that Slang already distinguishes between the elaboration and runtime versions of those!

hailongSun2000:
Cool!

PR #7633, opened 2024-09-25
[FIRRTL] Add verifier of single MarkDUTAnnotation

seldridge:
Add a CircuitOp verifier that guarantees that there is zero or one modules annotated with a `MarkDUTAnnotation`.  This is important as many passes assume that this is the case.  Moving this check into a verifier means that logic checking this behavior in each pass can be removed.  I am slightly concerned of the performance cost of doing this as examining the annotations is non-trivial and this will cause that examination to now happen more often.


fabianschuiki (approved):
LGTM!

PR #7634, opened 2024-09-25
[ImportVerilog] Add support for $stop/$finish/$exit

fabianschuiki:
Add support for the simulation control system tasks `$stop`, `$finish`, and `$exit`. Also add corresponding ops to the Moore dialect that handle the orthogonal pieces of functionality represented by these tasks.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -187,6 +187,18 @@ def ReturnOp : MooreOp<"return", [
   let assemblyFormat = [{ attr-dict }];
 }
 
+def UnreachableOp : MooreOp<"unreachable", [Terminator]> {
+  let summary = "Terminates a block as unreachable";
+  let description = [{
+    The `moore.unreachable` op is used to indicate that control flow never
+    reaches the end of a block. This is useful for operations such as `$fatal`
+    which never return as they cause the simulator to shut down. Behavior is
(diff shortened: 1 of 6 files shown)


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1345,4 +1357,67 @@ def CoverOp : ImmediateAssertOp<"cover">{
...
+def FinishBIOp : Builtin<"finish"> {
+  let summary = "Exit simulation";
+  let description = [{
+    Corresponds to the `$finish` system task. Causes the simulator to exit and
+    pass control back to the host operating system. Printing of the optional
+    diagnostic message is handled by the `finish_message` op.
+
+    The exit code argument of this op is not directly accessible from Verilog,
+    but is used to distinguish between the implicit `$finish` call in `$fatal`
+    and an explicit `$finish` called by the user.
```
hailongSun2000, line 1392: I don't have any questions.
nit: Maybe we could explicitly notify `moore.builtin.finish 0` representing **an explicit $finish called**, `1` is **implicit $finish**.

hailongSun2000 (approved):
LGTM :sunglasses:!

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: Yeah we could add a marker for that. For example a `implicit` unit attribute to distinguish it from the exit code.

PR #7639, opened 2024-09-26
[firtool] Move LowerLayers later in pipeline.

dtzSiFive:
LowerLayers belongs with the other passes that lower out of FIRRTL.

Start moving it, now that first batch of passes have been fixed to handle enough FIRRTL IR to run in presence of layer operations.

### DIFF
--- a/lib/Firtool/Firtool.cpp
+++ b/lib/Firtool/Firtool.cpp
@@ -141,8 +141,6 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
     modulePM.addPass(firrtl::createLayerSinkPass());
   }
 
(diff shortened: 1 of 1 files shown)


seldridge (approved):
Nice.

I think this also fixes https://github.com/llvm/circt/issues/7533 ?

dtzSiFive:
> I think this also fixes #7533 ?

It does appear to fix the result of the example posted.  If that resolves the issue sufficiently, then yes.  Looks like canonicalization now runs before this as well 🎉 .

FWIW I wasn't sure if we wanted to move LayerSink later in the pipeline or not, but figured we can sort that out later (and pipeline placement re:LayerMerge which has other considerations).

fabianschuiki (approved):
🥳

PR #7640, opened 2024-09-26
[FIRRTL][LowerLayers] Remove handling of some ref ops

rwy7:
Remove code that handles the capturing of ref cast and ref send operands specially, since this code is doing only what the generic capture-handling code does just a few lines [below](https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp#L611-L616).

There is a bug in the handling of ref.cast which is removed by this change: We are accidentally checking that the cast op itself is within the layerblock, when we should actually be checking if its operand is defined within the layerblock. This causes us to skip converting the captured operand of a ref.cast op into a port. The generic codepath works fine.


dtzSiFive (commented):
LGTM, thanks!

PR #7642, opened 2024-09-27
[ImportVerilog] Add $display/$write/$info/$warning/$error/$fatal

fabianschuiki:
Add support for the `$display`, `$write`, `$info`, `$warning`, `$error`, and `$fatal` system tasks.

These tasks are pretty complicated since they involve string formatting. Verilog has quite a few format specifiers and very strange rules for them. Luckily, Slang contains a handy utility that parses format strings and delegates handling of arguments and intermittent pieces of text to callbacks.

I've decided to follow the same IR design as the printing op in the Sim dialect: each format specifier is turned into a dedicated op that captures all relevant format options, plus additional literals for the text in between interpolated values, and concatenate everything into a single `!moore.format_string`. (Shoutout to @fzi-hielscher for trailblazing the nice design for `sim.print`!) This is handled in a new `FormatStrings.cpp` file inside of ImportVerilog.

The actual system tasks are mapped to new `moore.builtin.display` and `moore.builtin.severity` ops. These handle only the printing of the message in question, plus potential error bookkeeping. The `$fatal` system task creates additional `moore.builtin.finish_message` and `moore.builtin.finish` ops to represent its implicit call to `$finish`.

The implementation also handles the strange `$displayb`, `$displayo`, `$displayh`, `$writeb`, `$writeo`, and `$writeh` flavors of the tasks, where the suffix indicates the default format to use for arguments that are not covered by a format string literal. SystemVerilog is weird.

Thanks @hailongSun2000 for your prior work on this!

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -1358,6 +1358,84 @@ def CoverOp : ImmediateAssertOp<"cover">{
 }
 
 //===----------------------------------------------------------------------===//
+// Format Strings
+//===----------------------------------------------------------------------===//
+
+def FormatLiteralOp : MooreOp<"fmt.literal", [Pure]> {
+  let summary = "A constant string fragment";
+  let description = [{
+    Creates a constant string fragment to be used as a format string. The
+    literal is printed as is, without any further escaping or processing of its
+    characters.
+  }];
+  let arguments = (ins StrAttr:$literal);
+  let results = (outs FormatStringType:$result);
+  let assemblyFormat = "$literal attr-dict";
+}
+
+def FormatConcatOp : MooreOp<"fmt.concat", [Pure]> {
+  let summary = "Concatenate string fragments";
+  let description = [{
+    Concatenates an arbitrary number of format string into one larger format
+    string. The strings are concatenated from left to right, with the first
+    operand appearing at the left start of the result string, and the last
+    operand appearing at the right end. Produces an empty string if no inputs
+    are provided.
+  }];
+  let arguments = (ins Variadic<FormatStringType>:$inputs);
+  let results = (outs FormatStringType:$result);
+  let assemblyFormat = "` ` `(` $inputs `)` attr-dict";
+}
+
+def FmtDec : I32EnumAttrCase<"Decimal", 0, "decimal">;
+def FmtBin : I32EnumAttrCase<"Binary", 1, "binary">;
+def FmtOct : I32EnumAttrCase<"Octal", 2, "octal">;
+def FmtHexL : I32EnumAttrCase<"HexLower", 3, "hex_lower">;
+def FmtHexU : I32EnumAttrCase<"HexUpper", 4, "hex_upper">;
+def IntFormatAttr : I32EnumAttr<"IntFormat", "Integer format",
+                               [FmtDec, FmtBin, FmtOct, FmtHexL, FmtHexU]> {
+  let cppNamespace = "circt::moore";
+}
+
+def AlignRight : I32EnumAttrCase<"Right", 0, "right">;
+def AlignLeft : I32EnumAttrCase<"Left", 1, "left">;
+def IntAlignAttr : I32EnumAttr<"IntAlign", "Integer alignment",
+                              [AlignRight, AlignLeft]> {
+  let cppNamespace = "circt::moore";
+}
+
+def PadSpace : I32EnumAttrCase<"Space", 0, "space">;
+def PadZero : I32EnumAttrCase<"Zero", 1, "zero">;
+def IntPaddingAttr : I32EnumAttr<"IntPadding", "Integer alignment",
+                                 [PadSpace, PadZero]> {
+  let cppNamespace = "circt::moore";
+}
+
+def FormatIntOp : MooreOp<"fmt.int", [Pure]> {
+  let summary = "Format an integer value";
+  let description = [{
+    Format an integer value as a string according to the specified format.
+
+    See IEEE 1800-2017 § 21.2.1.2 "Format specifications".
+  }];
+  let arguments = (ins
+    IntType:$value,
+    IntFormatAttr:$format,
+    I32Attr:$width,
+    IntAlignAttr:$alignment,
+    IntPaddingAttr:$padding
+  );
+  let results = (outs FormatStringType:$result);
+  let assemblyFormat = [{
+    $format $value `,`
+    `width` $width `,`
+    `align` $alignment `,`
+    `pad` $padding
+    attr-dict `:` type($value)
+  }];
+}
+
+//===----------------------------------------------------------------------===//
 // Builtin System Tasks and Functions
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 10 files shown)


fabianschuiki:
@hailongSun2000 I've based this on your #7600 PR :smiley:

hailongSun2000 (approved):
I have reviewed the whole PR and understood the logical relationship. In terms of the functional logic, I approve. However, it's difficult that if I were to implement it. Therefore, I think maybe we can wait for another person to review it again. WDYT @fabianschuiki :smiley:?

lib/Conversion/ImportVerilog/FormatStrings.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    case 's':
+      // Simplified handling for literals.
+      if (auto *lit = arg.as_if<slang::ast::StringLiteral>()) {
+        if (options.width)
+          return mlir::emitError(loc)
+                 << "string format specifier with width not supported";
+        emitLiteral(lit->getValue());
+        return success();
+      }
```
uenoku, line 139: It only handles literal passed to format string arguments and doesn't accept a string variable like this, correct?
```
string s = "foo";
initial begin
  display("%s", s);
end
```
Maybe it's more flexible if StringLiteralOp takes `moore.string` type SSA value and produces format string value but I think we can revisit this later :+1:

lib/Conversion/ImportVerilog/FormatStrings.cpp
```diff
@@ -0,0 +1,200 @@
...
+  /// format specifier.
+  LogicalResult emitArgument(char specifier, StringRef fullSpecifier,
+                             const FormatOptions &options) {
+    auto specifierLower = std::tolower(specifier);
+
+    // Special handling for format specifiers that consume no argument.
+    if (specifierLower == 'm' || specifierLower == 'l') {
+      return mlir::emitError(loc)
+             << "unsupported format specifier `" << fullSpecifier << "`";
+    }
```
uenoku, line 107: super nit:
```suggestion
    if (specifierLower == 'm' || specifierLower == 'l') 
      return mlir::emitError(loc)
             << "unsupported format specifier `" << fullSpecifier << "`";
    
```

lib/Conversion/ImportVerilog/FormatStrings.cpp
```diff
@@ -0,0 +1,200 @@
...
+      emitLiteral(text);
+    };
+    auto onArg = [&](auto specifier, auto offset, auto len,
+                     const auto &options) {
+      if (anyFailure)
+        return;
+      if (failed(emitArgument(specifier, format.substr(offset, len), options)))
+        anyFailure = true;
+    };
+    auto onError = [&](auto, auto, auto, auto) {};
```
uenoku, line 87: Does `onError` need to set `anyFailure` when invoked?

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -578,6 +580,83 @@ struct StmtVisitor {
       return true;
     }
 
+    // Display and Write Tasks (`$display[boh]?` or `$write[boh]?`)
+
+    // Check for a `$display` or `$write` prefix.
+    bool isDisplay = false;     // display or write
+    bool appendNewline = false; // display
+    StringRef name = subroutine.name;
```
uenoku, line 588: nit: this `name` variable looks like generic but it cannot be used other than handling `display/write` due to mutation. To avoid confusion can we change the variable name or create a local scope for handling `display/write`?

uenoku (approved):
LGTM! Not blocking but it would be great if in the future we can refactor enum attributes (format, padding, align) to share with sim.fmt.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1357,6 +1357,84 @@ def CoverOp : ImmediateAssertOp<"cover">{
...
+  let assemblyFormat = "` ` `(` $inputs `)` attr-dict";
+}
+
+def FmtDec : I32EnumAttrCase<"Decimal", 0, "decimal">;
+def FmtBin : I32EnumAttrCase<"Binary", 1, "binary">;
+def FmtOct : I32EnumAttrCase<"Octal", 2, "octal">;
+def FmtHexL : I32EnumAttrCase<"HexLower", 3, "hex_lower">;
+def FmtHexU : I32EnumAttrCase<"HexUpper", 4, "hex_upper">;
+def IntFormatAttr : I32EnumAttr<"IntFormat", "Integer format",
+                               [FmtDec, FmtBin, FmtOct, FmtHexL, FmtHexU]> {
```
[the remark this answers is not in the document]
fabianschuiki: The exact opposite :grin:. `hex_lower` gives you lowercase letters, `fff`, and `hex_upper` gives you uppercase letters, `FFF`.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -40,26 +40,7 @@ struct RvalueExprVisitor {
...
-        auto sbvType =
-            moore::IntType::get(value.getContext(), *bits, packed.getDomain());
-        return builder.create<moore::ConversionOp>(loc, sbvType, value);
-      }
-    }
-
-    mlir::emitError(loc, "expression of type ")
-        << value.getType() << " cannot be cast to a simple bit vector";
-    return {};
+    return context.convertToSimpleBitVector(value);
```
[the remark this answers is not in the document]
fabianschuiki: Yeah that would be cleaner...

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -2055,3 +2055,150 @@ function void SimulationControlBuiltins(bit x);
   // CHECK-NOT: moore.builtin.finish
   $exit;
 endfunction
+
+// CHECK-LABEL: func.func private @DisplayAndSeverityBuiltins(
+// CHECK-SAME: [[X:%.+]]: !moore.i32
+function void DisplayAndSeverityBuiltins(int x);
+  // CHECK: [[TMP:%.+]] = moore.fmt.literal "\0A"
+  // CHECK: moore.builtin.display [[TMP]]
```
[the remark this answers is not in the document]
fabianschuiki: The `\n` is parsed by Slang into the actual newline character (ASCII code 10, or 0x0A). This is then escaped again by MLIR's printer, which simply inserts the hexcodes for unprintable characters.

lib/Conversion/ImportVerilog/FormatStrings.cpp
fabianschuiki: Yeah totally! I started out with supporting only literals because they have a simpler special handling. Other cases we'd have to handle with something like `moore.fmt.string` which could take the different flavors of string that can be formatted. SystemVerilog is pretty weird: you can have a `string` type which we need to support, but _unpacked byte arrays_ also count as strings, and string literals like `"foo"` themselves don't have `string` type, but rather generate a `bit [N-1:0]` array of appropriate length. There's also a whole bunch of craziness about taking arbtirary integers and interpreting them as right-aligned ASCII characters. :exploding_head:

lib/Conversion/ImportVerilog/FormatStrings.cpp
fabianschuiki: The Slang frontend already runs this `parse` function and reports any errors it encounters. When ImportVerilog gets to this point, all errors have already reported and compilation would have already been aborted. I'll add a comment, and an assertion inside `onError` to check that this is correct. Thanks for the pointer!

lib/Conversion/ImportVerilog/Statements.cpp
fabianschuiki: Good point. It might be worthwhile factoring the different flavors of builtins out into separate functions, and calling them in sequence to try and match the subroutine :thinking:. I'll just rename the variable for now.

fabianschuiki:
> @uenoku [...] it would be great if in the future we can refactor enum attributes (format, padding, align) to share with sim.fmt.

I totally agree! @maerhart has suggested that we could try to factor out the formatting into a separate `fmt` dialect, and use that from Moore, FIRRTL, Sim, and others.

include/circt/Dialect/Moore/MooreOps.td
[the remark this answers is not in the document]
fzi-hielscher: Can you convince me that uppercase hex formatting actually exists in SV? Table 21-2 in the spec sounds like %x, %h, %X, %H are interchangeable. Testing it on some simulators, they all print lowercase only. [And if Dave Rich hasn't seen it](https://verificationacademy.com/forums/t/how-to-print-hexadecimal-numbers-in-upper-case/40264), I'm more inclined to believe in the existence of unicorns. 😉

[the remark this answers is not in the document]
fabianschuiki: :rofl: Huh interesting, I could have sworn I've seen this print uppercase hexadecimal in VCS and Questa a while back :thinking: I might be wrong though.

test/Conversion/ImportVerilog/basic.sv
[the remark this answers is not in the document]
hailongSun2000: Cool :sunglasses:!

PR #7643, opened 2024-09-27
[FIRRTL] Avoid InstanceGraphNode* in InstanceInfo

seldridge:
Fix a problem with the InstanceInfo analysis stemming from the use of `InstanceGraphNode *` to track the design-under-test (DUT) and "effective" DUT.  If the instance graph changed, then these would be invalidated. Instead use a `igraph::ModuleOpInterface` which is going to be stable.

### DIFF
--- a/include/circt/Analysis/FIRRTLInstanceInfo.h
+++ b/include/circt/Analysis/FIRRTLInstanceInfo.h
@@ -77,10 +77,10 @@ public:
   /// Information about a circuit
   struct CircuitAttributes {
     /// The design-under-test if one is defined.
-    igraph::ModuleOpInterface dutNode;
+    igraph::ModuleOpInterface dut;
 
     /// The design-under-test if one is defined or the top module.
-    igraph::ModuleOpInterface effectiveDutNode;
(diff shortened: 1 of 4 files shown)


include/circt/Analysis/FIRRTLInstanceInfo.h
```diff
@@ -77,10 +77,10 @@ class InstanceInfo {
   /// Information about a circuit
   struct CircuitAttributes {
     /// The design-under-test if one is defined.
-    igraph::InstanceGraphNode *dutNode;
+    igraph::ModuleOpInterface dutNode;
```
dtzSiFive, line 80: nit: rename to drop "Node" suffix?

dtzSiFive (approved):
Looks reasonable to me, thanks!

include/circt/Analysis/FIRRTLInstanceInfo.h
seldridge: SGTM. Changed (and changed all the debug code) in fc50801c3

PR #7644, opened 2024-09-27
[Analysis] Add OpCount Analysis

TaoBi22:
Adds a very simple analysis pass to collect information about the kinds of operations (and operand counts) found in a `hw.module` so we can look at operation usage across big designs. Once this is upstream I'll add a pass to export this info as JSON.

### DIFF
--- a/include/circt/Analysis/OpCountAnalysis.h
+++ b/include/circt/Analysis/OpCountAnalysis.h
@@ -25,11 +25,18 @@ class AnalysisManager;
 namespace circt {
 namespace analysis {
 
-struct OpCountAnalysis {
+class OpCountAnalysis {
+public:
   OpCountAnalysis(Operation *moduleOp, mlir::AnalysisManager &am);
 
+  /// Get the frequency of operations of a specific name
   size_t getOpCount(OperationName opName);
+
+  /// Get the names of all distinct operations found by the analysis
   SmallVector<OperationName> getFoundOpNames();
+
+  /// Get a map from number of operands to corresponding frequency for the given
+  /// operation
   DenseMap<size_t, size_t> getOperandCountMap(OperationName opName);
 
 private:
--- a/include/circt/Analysis/OpCountAnalysis.h
+++ b/include/circt/Analysis/OpCountAnalysis.h
@@ -1,0 +1,40 @@
+//===- OpCountAnalysis.h - operation count analyses -----------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This header file defines prototypes for methods that perform analysis
+// involving the quantity of different kinds of operation.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_ANALYSIS_OPCOUNT_ANALYSIS_H
+#define CIRCT_ANALYSIS_OPCOUNT_ANALYSIS_H
+
+#include "circt/Support/LLVM.h"
+#include "llvm/ADT/DenseMap.h"
+
+namespace mlir {
+class AnalysisManager;
+} // namespace mlir
+namespace circt {
+namespace analysis {
+
+struct OpCountAnalysis {
+  OpCountAnalysis(Operation *moduleOp, mlir::AnalysisManager &am);
+
+  size_t getOpCount(OperationName opName);
+  DenseMap<size_t, size_t> getOperandCountMap(OperationName opName);
+
+private:
+  DenseMap<OperationName, size_t> opCounts;
+  DenseMap<OperationName, DenseMap<size_t, size_t>> operandCounts;
+};
+
+} // namespace analysis
+} // namespace circt
+
+#endif // CIRCT_ANALYSIS_SCHEDULING_ANALYSIS_H
--- a/include/circt/Analysis/OpCountAnalysis.h
+++ b/include/circt/Analysis/OpCountAnalysis.h
@@ -7,7 +7,8 @@
 //===----------------------------------------------------------------------===//
 //
 // This header file defines prototypes for methods that perform analysis
-// involving the quantity of different kinds of operation.
+// involving the frequency of different kinds of operations found in a
+// builtin.module.
 //
 //===----------------------------------------------------------------------===//
 
@@ -15,6 +16,7 @@
 #define CIRCT_ANALYSIS_OPCOUNT_ANALYSIS_H
 
 #include "circt/Support/LLVM.h"
+#include "mlir/IR/Operation.h"
 #include "llvm/ADT/DenseMap.h"
 
 namespace mlir {
@@ -27,6 +29,7 @@ struct OpCountAnalysis {
   OpCountAnalysis(Operation *moduleOp, mlir::AnalysisManager &am);
 
   size_t getOpCount(OperationName opName);
+  SmallVector<OperationName> getFoundOpNames();
   DenseMap<size_t, size_t> getOperandCountMap(OperationName opName);
 
 private:
--- a/include/circt/Analysis/OpCountAnalysis.h
+++ b/include/circt/Analysis/OpCountAnalysis.h
@@ -47,4 +47,4 @@ private:
 } // namespace analysis
 } // namespace circt
 
-#endif // CIRCT_ANALYSIS_SCHEDULING_ANALYSIS_H
+#endif // CIRCT_ANALYSIS_OPCOUNT_ANALYSIS_H
--- a/lib/Analysis/OpCountAnalysis.cpp
+++ b/lib/Analysis/OpCountAnalysis.cpp
@@ -6,20 +6,20 @@
 //
 //===----------------------------------------------------------------------===//
 //
-// This file defines the InstanceInfo analysis.  This is an analysis that
-// depends on the InstanceGraph analysis, but provides additional information
-// about FIRRTL operations.  This is useful if you find yourself needing to
-// selectively iterate over parts of the design.
+// This file defines the op count analysis.  This is an analysis that
+// provides information about the frequency of different kinds of operations
+// found in a builtin.module.
 //
 //===----------------------------------------------------------------------===//
 
 #include "circt/Analysis/OpCountAnalysis.h"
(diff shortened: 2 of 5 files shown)


include/circt/Analysis/OpCountAnalysis.h
```diff
@@ -0,0 +1,43 @@
...
+struct OpCountAnalysis {
+  OpCountAnalysis(Operation *moduleOp, mlir::AnalysisManager &am);
+
+  size_t getOpCount(OperationName opName);
+  SmallVector<OperationName> getFoundOpNames();
+  DenseMap<size_t, size_t> getOperandCountMap(OperationName opName);
+
+private:
+  DenseMap<OperationName, size_t> opCounts;
+  DenseMap<OperationName, DenseMap<size_t, size_t>> operandCounts;
```
TaoBi22, line 37: `opCounts` could be calculated from `operandCounts` here, but it would require a lot of extra iterating over `DenseMap`s so went for the faster option since we probably don't want this to be slower than necessary if it's running on huge blocks of IR.

dtzSiFive:
Awesome, thanks for putting this together!

Looks like this is similar to upstream MLIR's `print-op-stats` pass -- not currently registered re:`circt-opt` but can be added, FWIW.

Have you seen this, have thoughts on how this compares?

At a glance, this seems to additionally bin by operand count, not just operation name...?

TaoBi22:
> Looks like this is similar to upstream MLIR's `print-op-stats` pass -- not currently registered re:`circt-opt` but can be added, FWIW.

Good point, I wasn't actually aware of that one! They definitely look similar - I think the big difference is that this is an analysis rather than just a pass. This might be handy as this is being upstreamed so we can get various pieces of data on how different operations are actually used in CIRCT (e.g. the spread of operand counts on variadic operations, hence the binning by operand count), and it would be interesting to be able to directly query the analysis as different transformations are done on the IR. Can drop the operation counting if this smells like duplication of functionality though.

include/circt/Analysis/OpCountAnalysis.h
```diff
@@ -0,0 +1,43 @@
...
+#include "mlir/IR/Operation.h"
+#include "llvm/ADT/DenseMap.h"
+
+namespace mlir {
+class AnalysisManager;
+} // namespace mlir
+namespace circt {
+namespace analysis {
+
+struct OpCountAnalysis {
```
seldridge, line 28: Nit: I think LLVM/MLIR coding guidelines say to only use a `struct` if all members are public. This would be better as a `class`.

include/circt/Analysis/OpCountAnalysis.h
```diff
@@ -0,0 +1,43 @@
...
+  OpCountAnalysis(Operation *moduleOp, mlir::AnalysisManager &am);
+
+  size_t getOpCount(OperationName opName);
+  SmallVector<OperationName> getFoundOpNames();
+  DenseMap<size_t, size_t> getOperandCountMap(OperationName opName);
```
seldridge, line 33: Nit: Doxygen-compatible comments on all public members would be great!

lib/Analysis/OpCountAnalysis.cpp
```diff
@@ -0,0 +1,62 @@
...
+    auto opName = op->getName();
+    // Update opCounts
+    if (opCounts.find(opName) == opCounts.end())
+      opCounts[opName] = 1;
+    else
+      opCounts[op->getName()]++;
+
+    // Update operandCounts
+    if (operandCounts.find(opName) == operandCounts.end())
+      operandCounts[opName] = DenseMap<size_t, size_t>();
```
seldridge, line 33: This can probably be dropped as `operandCounts[opName]` will default construct a `DenseMap` already.

lib/Analysis/OpCountAnalysis.cpp
```diff
@@ -0,0 +1,62 @@
...
+    if (operandCounts[opName].find(op->getNumOperands()) ==
+        operandCounts[opName].end())
+      operandCounts[opName][op->getNumOperands()] = 1;
+    else
+      operandCounts[opName][op->getNumOperands()]++;
```
seldridge, line 38: Nit: `operandCounts[opName]` shows up a bunch. Switching to a shared reference, i.e., `auto &counts = operandCounts[opName];` would clean this up a bit.

lib/Analysis/OpCountAnalysis.cpp
```diff
@@ -0,0 +1,62 @@
...
+  for (auto pair : opCounts)
+    opNames.push_back(pair.first);
+  return opNames;
+}
+
+size_t OpCountAnalysis::getOpCount(OperationName opName) {
+  size_t count = 0;
+  if (opCounts.find(opName) != opCounts.end())
+    count = opCounts[opName];
+  return count;
```
seldridge, line 53: Mega-nit:

```suggestion
  if (opCounts.find(opName) != opCounts.end())
    return opCounts[opName];
  return 0;
```

lib/Analysis/OpCountAnalysis.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+DenseMap<size_t, size_t>
+OpCountAnalysis::getOperandCountMap(OperationName opName) {
+  auto map = DenseMap<size_t, size_t>();
+  if (operandCounts.find(opName) != operandCounts.end())
+    map = operandCounts[opName];
+  return map;
```
seldridge, line 61: Default construction should allow for:

```suggestion
  return operandCounts[opName];
```

lib/Analysis/TestPasses.cpp
```diff
@@ -250,6 +251,49 @@ void FIRRTLInstanceInfoPass::runOnOperation() {
...
+    auto operandMap = opCount.getOperandCountMap(opName);
+    // Sort for determinism again
+    llvm::SmallVector<size_t> keys;
+    for (auto pair : operandMap)
+      keys.push_back(pair.first);
+    llvm::sort(keys);
+    for (auto num : keys)
+      llvm::errs() << " with " << num << " operands: " << operandMap[num]
+                   << "\n";
+  }
```
seldridge, line 289: I'll leave this format to you, particularly since you mentioned bridging this to JSON or something like that.

I tend to actually export this information directly as YAML (manually, not using the internal library). You may be able to go to JSON directly here if you want.

Regardless, I think the format should be less wordy eventually, e.g.:

```
operations:
  - name: builtin.module
    count: 1
  - name: comb.add
    count: 4
    operandCounts:
      - operands: 1
        count: 2
      - operands: 2
        count: 4
```

seldridge (approved):
It would be interesting if this could unify with upstream. That said, this is fine to go in as we need it for running some experiments. (If we completely throw this away in the future after it's served it's purpose, that's fine, too!)

lib/Analysis/OpCountAnalysis.cpp
TaoBi22: Ah cool, didn't actually know that!

lib/Analysis/TestPasses.cpp
TaoBi22: Yeah, I just put this in for testing so figured something readable in case of debugging failure wouldn't hurt. Happy to trim it down if preferred though.

lib/Analysis/OpCountAnalysis.cpp
```diff
@@ -0,0 +1,56 @@
...
+  SmallVector<OperationName> opNames;
+  for (auto pair : opCounts)
+    opNames.push_back(pair.first);
+  return opNames;
+}
+
+size_t OpCountAnalysis::getOpCount(OperationName opName) {
+  if (opCounts.find(opName) != opCounts.end())
+    return opCounts[opName];
+  return 0;
```
maerhart, line 50: Nit:
```suggestion
  return opCounts.lookup(opName);
```

lib/Analysis/TestPasses.cpp
```diff
@@ -250,6 +251,49 @@ void FIRRTLInstanceInfoPass::runOnOperation() {
...
 }
 
+//===----------------------------------------------------------------------===//
+// Op Count
+//===----------------------------------------------------------------------===//
+
+namespace {
+struct TestOpCountAnalysisPass
+    : public PassWrapper<TestOpCountAnalysisPass, OperationPass<ModuleOp>> {
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(TestOpCountAnalysisPass)
```
maerhart, line 261: If you plan to add a proper pass to print it as JSON, why do we even need a separate test pass? That proper pass could support two formats (a human-readable one and an easy-to-parse one) like the upstream variant.

maerhart (approved):
Thanks for putting this together!

include/circt/Analysis/OpCountAnalysis.h
```diff
@@ -0,0 +1,50 @@
...
+
+private:
+  DenseMap<OperationName, size_t> opCounts;
+  DenseMap<OperationName, DenseMap<size_t, size_t>> operandCounts;
+};
+
+} // namespace analysis
+} // namespace circt
+
+#endif // CIRCT_ANALYSIS_SCHEDULING_ANALYSIS_H
```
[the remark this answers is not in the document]
TaoBi22: Oops, thanks!

lib/Analysis/OpCountAnalysis.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+OpCountAnalysis::OpCountAnalysis(Operation *moduleOp,
+                                 mlir::AnalysisManager &am) {
+  moduleOp->walk([&](Operation *op) {
+    auto opName = op->getName();
+    // Update opCounts
+    if (opCounts.find(opName) == opCounts.end())
+      opCounts[opName] = 1;
+    else
+      opCounts[opName]++;
```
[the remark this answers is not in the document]
TaoBi22: :star_struck: wondered if I could do this but didn't think it would initialise a size_t properly, good to know! Thanks!

lib/Analysis/TestPasses.cpp
TaoBi22: Good point - I just put this in so we had something upstream quickly to start getting data while I did the JSON pass, but definitely agree it would make sense to have the JSON pass also do a human readable form.

TaoBi22:
Merging this now, I'll get another PR in shortly to do JSON export (and move the readable export into that pass too). I'll make the format less wordy as requested by @seldridge too

PR #7645, opened 2024-09-27
[ImportVerilog] Add support for $clog2

fabianschuiki:
Add a `moore.builtin.clog2` op and use it to convert `$clog2` calls in ImportVerilog.


maerhart (approved):
LGTM!

PR #7646, opened 2024-09-27
[ImportVerilog] Support assignment patterns with integer type

fabianschuiki:
Allow `'{...}` assignment patterns to be used to assign integer values. Slang already computes the value for each element in the resulting type. All that's needed is a concat with the right operand order.

### DIFF
--- a/test/Conversion/ImportVerilog/basic.sv
+++ b/test/Conversion/ImportVerilog/basic.sv
@@ -655,6 +655,8 @@ module Expressions;
   bit [1:0][31:0] arrayInt;
   // CHECK: %uarrayInt = moore.variable : <uarray<2 x i32>>
   bit [31:0] uarrayInt [2];
+  // CHECK: %m = moore.variable : <l4>
+  logic [3:0] m;
 
   initial begin
     // CHECK: moore.constant 0 : i32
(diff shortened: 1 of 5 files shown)


test/Conversion/ImportVerilog/basic.sv
```diff
@@ -655,6 +655,8 @@ module Expressions;
   bit [1:0][31:0] arrayInt;
   // CHECK: %uarrayInt = moore.variable : <uarray<2 x i32>>
   bit [31:0] uarrayInt [2];
+  // CHECK: %m = moore.variable : <l4>
+  logic [3:0] m;
```
maerhart, line 659: If you change this to `logic [0:3]` the concat element order reverses as well, right?

maerhart (approved):
LGTM!

test/Conversion/ImportVerilog/basic.sv
fabianschuiki: Hmmm good point. I'm not sure what the expected bit order would be in that case :thinking:. Intuitively that would make sense. But I also remember the standard mentioning that the array direction doesn't really matter -- e.g. you can't reverse bits by assigning a `[0:3]` array to a `[3:0]` array, people usually resort to the streaming operators for that. Time to look through the standard some more :wink:

fabianschuiki: Oh I just realized that the assignment patterns talk about element indices explicitly: `'{3: 'x}` will assign bit index 3, regardless of whether the array was declared as `[3:0]` or `[0:3]`. And since the Moore dialect removes array directionality (if it becomes important we'll just emit flip ops in the right places), the concat should work as it is now (bit index 3 is always in the same location in the Moore dialect).

PR #7647, opened 2024-09-27
[ImportVerilog] Insert missing conversions around instance ports

fabianschuiki:
In certain corner cases, Slang does not insert the necessary conversion AST nodes around instance ports. This can be problematic if an input or output port is connected to a cast-compatible type (e.g. `i32` and `array<1 x i32>`). To fix this, always insert conversions if the types of a port and the connected value differ.


maerhart (approved):
Great catch! Thanks for the fix!

PR #7652, opened 2024-10-01
[MooreToCore] Lower moore.net to llhd.sig

fabianschuiki:
Lower `moore.net` operations to the corresponding `llhd.sig`, similar to how `moore.variable` is lowered. This currently only makes sense for "wire" nets, since it's not yet clear how all the other net types map to LLHD.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -496,6 +496,44 @@ struct VariableOpConversion : public OpConversionPattern<VariableOp> {
   }
 };
 
+struct NetOpConversion : public OpConversionPattern<NetOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(NetOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto loc = op.getLoc();
+    if (op.getKind() != NetKind::Wire)
+      return rewriter.notifyMatchFailure(loc, "only wire nets supported");
+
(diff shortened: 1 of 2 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -496,6 +496,44 @@ struct VariableOpConversion : public OpConversionPattern<VariableOp> {
...
+    auto loc = op.getLoc();
+    if (op.getKind() != NetKind::Wire)
+      return rewriter.notifyMatchFailure(loc, "only wire nets supported");
+
+    auto resultType = typeConverter->convertType(op.getResult().getType());
+    if (!resultType)
+      return rewriter.notifyMatchFailure(loc, "invalid net type");
+
+    // TODO: Once the core dialects support four-valued integers, this code
+    // will additionally need to generate an all-X value for four-valued nets.
```
maerhart, line 514: Shouldn't it be initialized to all Z once we support four-valued types? I think the standard mentions that as the default value in case there are no drivers.

maerhart (approved):
Nice! 🎉

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: That is a good point! I'm also starting to wonder how we'll deal with multiple concurrent drivers on a net. That doesn't easily map to `llhd.sig` at the moment. It's more like you'd have a list of multiple `llhd.sig`, one for each driver, and you'd then have a conflict resolution step that merges all the drivers into a final net value. We might need a slightly different lowering in the future :thinking:

PR #7653, opened 2024-10-02
[Arc] Fix folding of initialized StateOp

fzi-hielscher:
When adding the initial value parameter to the `StateOp` I forgot to adapt its folder.  This PR should fix that.
 - Fold a permanently disabled state to its initial value (if present) instead of zero.
 - Don't fold a state that is permanently reset if it has an explicit initial value.
 - Don't attempt to constant fold a result-less `StateOp` to avoid an infinite loop.


maerhart (approved):
Thanks for the fix! And great catch with the 0 result values.

fzi-hielscher:
Sorry, it just occurred to me that this is also incorrect if the initial value is  time-variant (you'd think I knew better after the discussion in #7638  😅 ). We need to make sure that a 'dynamic' initial value is materialized in the `arc.initial` op and not in the actual clock tree.
I'll restrict the folding to constants for now...

PR #7654, opened 2024-10-02
[Transforms][OpCountAnalysis] Add PrintOpCountPass

TaoBi22:
This adds a proper pass to export the data generated by the analysis upstreamed in #7644 and deprecates the test pass that existed before. It's in Transform to align where the similar upstream MLIR pass is found, but I'd be easily swayed on putting it elsewhere.

Unfortunately couldn't find a nice way to FileCheck test the JSON output as you don't seem to be able to have a CHECK-DAG node span multiple lines (although I could add a sorting flag if we're very set on testing it, just didn't bother here as it seems like a slightly redundant flag for output that's realistically always going to be imported elsewhere for processing).


lib/Transforms/PrintOpCount.cpp
```diff
@@ -0,0 +1,111 @@
...
+        os << ",";
+      os << "\n";
+    }
+    os << " }";
+    if (i != opNames.size() - 1)
+      os << ",";
+    os << "\n";
+  }
+  os << "}\n";
+}
```
seldridge, line 74: This would benefit from use of `llvm::json` to avoid having to hand-roll JSON emission.

An example of a very compact usage of that API: https://github.com/llvm/llvm-project/blob/09ba83be0ac178851e3c9c9c8fefddbdd4d8353f/llvm/lib/Analysis/TensorSpec.cpp#L50

lib/Transforms/PrintOpCount.cpp
```diff
@@ -0,0 +1,111 @@
...
+      keys.push_back(pair.first);
+    // Sort for determinism if required again
+    if (sorted)
+      llvm::sort(keys);
+    for (auto num : keys) {
+      os << "    - operands: " << num << "\n";
+      os << "      count: " << operandMap[num] << "\n";
+    }
+  }
+}
```
seldridge, line 53: Nice! Thanks for the simpler emission.

seldridge (approved):
LGTM

Consider switching to the `llvm::json` library at some point.

lib/Transforms/PrintOpCount.cpp
TaoBi22: Good point, this was a pretty easy change so I've updated this PR to use it

lib/Transforms/PrintOpCount.cpp
```diff
@@ -0,0 +1,102 @@
...
+void printOpAndOperandJSON(analysis::OpCountAnalysis &opCount,
+                           raw_ostream &os) {
+  auto jos = llvm::json::OStream(os);
+  jos.object([&] {
+    for (auto opName : opCount.getFoundOpNames())
+      jos.attributeObject(opName.getStringRef(), [&] {
+        for (auto pair : opCount.getOperandCountMap(opName))
+          jos.attribute(std::to_string(pair.first), pair.second);
+      });
+  });
```
seldridge, line 64: This is super clean and compact now. 👍

PR #7657, opened 2024-10-03
[FIRRTL] Add integer shift left property op

maerhart:


mikeurbach (approved):
Looks good to me, thanks for adding this and including some folds!

BTW, we should also define this in the FIRRTL spec. We haven't been great about whether the spec or implementation should be defined first, but they should eventually converge :) Maybe you could add another section along the lines of https://github.com/chipsalliance/firrtl-spec/pull/176?

maerhart:
Here it is: https://github.com/chipsalliance/firrtl-spec/pull/238

PR #7658, opened 2024-10-03
[OM] Add integer shift left op

maerhart:


mikeurbach (approved):
Awesome, looks great!

PR #7660, opened 2024-10-03
[FIRRTL] Add integer shift left conversion to LowerClasses

maerhart:


mikeurbach (approved):
Looks good to me!

maerhart:
Changed constant operands to class inputs to avoid constant folding (for `shl` that's already a problem, for the others it will become one once folders are added for them)

PR #7663, opened 2024-10-04
[Not-for-commit] Demonstrate Yosys integration

uenoku:
This PR demonstrates Yosys integration which interacts with Yosys's IR (RTLIL). Yosys is de-facto standard as an opensource RTL synthesis and verification tool. The purpose of the yosys integration is to serve as a baseline for circt-based synthesis flow. It's super experimental and this PR needs to be separated into 5~7 PRs for review. This integration work is inspired by HEIR's [Yosys optimizer pass](https://github.com/google/heir/tree/116c4577a5cb2ef9bcd16a3e71423872f3ed6257/lib/Transforms/YosysOptimizer) but in this PR we directly translate CIRCT IR to RTLIL instead of using SystemVerilog. This integration makes it possible to import designs from RTLIL so technically we can use the circuits generated by framework based on Yosys such as amaranth-lang/amaranth.

## Build
Yosys integration is disabled by default. To enable Yosys integration, you need to build CIRCT with `-DCIRCT_YOSYS_INTEGRATION_ENABLED=ON`. This will download Yosys release and generate a shared lib (as you can see the CMakeFile.txt in CIRCT

## Tranlation between RTLIL

`circt-translate` provides subcommand for translation between CIRCT IR and RTLIL.

* CIRCT IR -> RTLIL

```bash
circt-tranlsate --export-rtlil
```
Currently only subest of core dialects are translated during translation to RTLIL.

* RTLIL -> CIRCT IR

```bash
circt-tranlsate --import-rtlil
```

## Run Yosys passes on CIRCT IR

To run Yosys passes on CIRCT there are two passes `yosys-optimizer` and `yosys-optimizer-parallel`.

```bash
# Run synth (canonicalizer is currently required to get "clean" core dialect operations)
circt-opt  -pass-pipeline='builtin.module(yosys-optimizer{passes=synth},canonicalize)'

# Pipe Yosys log to the stderr via `redirect-log=true`.
circt-opt  -pass-pipeline='builtin.module(yosys-optimizer{passes=synth redirect-log=true})'

# Run synth_xilinx (symbol-dce is recommended to remove unused external modules)
circt-opt  -pass-pipeline='builtin.module(yosys-optimizer{passes=synth_xilinx},canonicalize,symbol-dce)'

# Run multiple yosys pass (comma separated).
circt-opt  -pass-pipeline='builtin.module(yosys-optimizer{passes=write_rtlil,write_verilog,synth},canonicalize)'

# Emit Verilog using yosys backend. It's recommended to run `opt` for the fair comparison
circt-opt  -pass-pipeline='builtin.module(yosys-optimizer{passes=opt,write_verilog})'
```

### CIRCT as a parallel Yosys driver

Yosys has a globally context which is a not thread-safe so we cannot parallelly run `yosys-optimizer` on each HW module. As a workaround CIRCT provides a `yosys-optimizer-parallel` pass that parallelly invokes yosys in child processes. `yosys-optimizer-parallel` cannot be used for transformation that requires module hierarchly (e.g. inlining/flattening etc)

## Testing

Testing Yosys integration is tricky since RTLIL textual format could differ between Yosys versions. Currently we test the correctness of RTLIL translation with a few round-trip tests and LEC on pre-and-post synthesis. 

## License 
Yosys has [ISC License](https://opensource.org/license/isc-license-txt) so it should be fine to use it through shared lib.


programmerjake:
I think this (or at the very least the rtlil export functionality) would be useful to include in circt, the current workflow we're using in [fayalite](https://git.libre-chip.org/libre-chip/fayalite) is using firtool to translate firrtl to verilog which is then read by yosys, which is a lossy process (e.g. source location information isn't properly translated, assert/assume messages must be empty, etc.). I then use yosys to do formal verification using symbiyosys and am planning on using yosys and nextpnr to generate fpga bitstreams.

Issue #7664, opened 2024-10-04
[HWToSMT] Support hw.array_create and hw.array_get


uenoku:
Currently we cannot verify LEC between following two modules:
```
  hw.module @MultibitMux(in %a_0 : i1, in %a_1 : i1, in %sel : i1, out b : i1) {
    %0 = hw.array_create %a_1, %a_0 : i1
    %1 = hw.array_get %0[%sel] : !hw.array<2xi1>, i1
    hw.output %1 : i1
  }

  hw.module @MultibitMux2(in %a_0 : i1, in %a_1 : i1, in %sel : i1, out b : i1) {
    %0 = comb.mux bin %sel, %a_1, %a_0 : i1
    hw.output %0 : i1
  }
```

```
$  circt-lec bar.mlir --c1 MultibitMux --c2 MultibitMux2
bar.mlir:2:10: error: failed to legalize operation 'hw.array_create' that was explicitly marked illegal
    %0 = hw.array_create %a_1, %a_0 : i1
```

PR #7666, opened 2024-10-04
[HWToSMT] ArrayCreateOp and ArrayGetOp support

maerhart:
Resolves https://github.com/llvm/circt/issues/7664

### DIFF
--- a/integration_test/circt-lec/hw.mlir
+++ b/integration_test/circt-lec/hw.mlir
@@ -76,3 +76,27 @@ hw.module @onePlusTwoNonSSA(out out: i2) {
 //  RUN: circt-lec %s -c1=onePlusTwo -c2=onePlusTwoNonSSA --shared-libs=%libz3 | FileCheck %s --check-prefix=HW_MODULE_GRAPH
 //  HW_MODULE_GRAPH: c1 == c2
 
+// array_create + array_get test
+//  RUN: circt-lec %s -c1=MultibitMux -c2=MultibitMux2 --shared-libs=%libz3 | FileCheck %s --check-prefix=ARRAY_GET
+//  ARRAY_GET: c1 == c2
+
+hw.module @MultibitMux(in %a_0 : i1, in %a_1 : i1, in %sel : i1, out b : i1) {
+  %0 = hw.array_create %a_1, %a_0 : i1
+  %1 = hw.array_get %0[%sel] : !hw.array<2xi1>, i1
+  hw.output %1 : i1
+}
+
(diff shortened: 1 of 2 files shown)


uenoku:
Thank you for quickly working on adding this! Using array theory sounds interesting but IIRC array theory is fairly expensive compared to bitvector.  I think there are several representations and I'm curious if other lightweight representation might make sense.

1. Lower to smt::ArrayStore/ArraySelect
2. Lower it as `comb.concat` + `comb.shru` + `comb.extract`. 
3. Lower it as mux chains. 

I created larger test case for (1) and (2) representation (https://gist.github.com/uenoku/7fcf89f676ff3d586204669c044ea924) and ran LEC against the module itself. LEC didn't finish for (1) within 2 minutes whereas it finished less than a second for (2).

maerhart:
Thanks for this very cool performance analysis! There are definitely a lot of representations we can try and benchmark against each other. Staying within BV theory also has the advantage that we can use SMT solvers optimized for BV only and don't support array theory.

There are two things about your concat+shru+extract version I'd like to point out:
* `i1` element types are the most convenient for this representation. For wider integers we'd additionally need a bitvector multiplication. Those are very problematic, but the usual optimizations can be done here, such as replacing it with a shift or concat if the integer width is a power of 2. Alternatively, we can pad the array elements to the a power of 2 when concatenating them to the big BV, which allows us to always use a concat instead of a multiplication.
* Out-of-bounds accesses are not handled correctly. An OOB access is just defined to be zero here, but we'd need to return a symbolic value (I'm not sure we properly defined this, I'm just assuming we use the same semantics as division by 0)

Furthermore, we should be careful with benchmarking the approaches by just comparing the same circuit with itself.
I changed the array lowering to use your BV approach and incorporated above two points (not optimizing away the SMT multiplication and using `i128` as array element types). Proving it equivalent with itself also takes ~1 sec, but proving it equivalent with the Comb implementation using a concat instead of the multiplication didn't terminate within 10 min.

uenoku (commented):
> i1 element types are the most convenient for this representation. For wider integers we'd additionally need a bitvector multiplication.

Yeah, that's a great point. I agree that using array theory is good for non-i1 integer arrays. 

> Out-of-bounds accesses are not handled correctly. 

Yeah but can't we insert `smt.ite` for this representation as well while lowering `hw.array_get`?

uenoku (approved):
I think starting from simple solution is always great so lowering it to ArrayStore and ArraySelect look good to me! Maybe we can do something complicated in the future when array-theory does not scale well 😃

TaoBi22 (approved):
LGTM! Very cool performance analysis @uenoku, agreed that it's worth getting this in for now and potentially moving to something more performant later!

PR #7667, opened 2024-10-04
 [ExtractInstances] Add the extract instances metadata to OM Classes

prithayan:
This change adds the `ExtractInstances` metadata to the OM classes. The OM dialect can then be parsed to generate the respective metadata files.
This change enables to move all the metadata to the OM dialect, instead of existing as verbatim ops.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
@@ -356,13 +356,18 @@ def ClassOp : FIRRTLModuleLike<"class", [
   let builders = [
     OpBuilder<(ins
       "StringAttr":$name,
-      "ArrayRef<PortInfo>":$ports)>];
+      "ArrayRef<PortInfo>":$ports)>,
+    // This builds a ClassOp, with the specified fieldNames and fieldTypes as
+    // ports. The output property is set from the input property port.
+    OpBuilder<(ins
+      "Twine":$name,
+      "mlir::ArrayRef<mlir::StringRef>":$fieldNames,
+      "mlir::ArrayRef<mlir::Type>":$fieldTypes)>
+    ];
 
   let extraModuleClassDeclaration = [{
     Block *getBodyBlock() { return &getBody().front(); }
 
-    // This builds a ClassOp, with the specified fieldNames and fieldTypes as
-    // ports. The output property is set from the input property port.
     ClassOp static buildSimpleClassOp(
       mlir::OpBuilder &odsBuilder, mlir::Location loc, mlir::Twine name,
       mlir::ArrayRef<mlir::StringRef> fieldNames,
--- a/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
@@ -361,6 +361,13 @@ def ClassOp : FIRRTLModuleLike<"class", [
   let extraModuleClassDeclaration = [{
     Block *getBodyBlock() { return &getBody().front(); }
 
+    // This builds a ClassOp, with the specified fieldNames and fieldTypes as
+    // ports. The output property is set from the input property port.
+    ClassOp static buildSimpleClassOp(
+    mlir::OpBuilder &odsBuilder, mlir::Location loc, mlir::Twine name,
+    mlir::ArrayRef<mlir::StringRef> fieldNames,
+    mlir::ArrayRef<mlir::Type> fieldTypes);
+
     using iterator = Block::iterator;
     iterator begin() { return getBodyBlock()->begin(); }
     iterator end() { return getBodyBlock()->end(); }
--- a/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
@@ -364,9 +364,9 @@ def ClassOp : FIRRTLModuleLike<"class", [
     // This builds a ClassOp, with the specified fieldNames and fieldTypes as
     // ports. The output property is set from the input property port.
     ClassOp static buildSimpleClassOp(
-    mlir::OpBuilder &odsBuilder, mlir::Location loc, mlir::Twine name,
-    mlir::ArrayRef<mlir::StringRef> fieldNames,
-    mlir::ArrayRef<mlir::Type> fieldTypes);
+      mlir::OpBuilder &odsBuilder, mlir::Location loc, mlir::Twine name,
+      mlir::ArrayRef<mlir::StringRef> fieldNames,
+      mlir::ArrayRef<mlir::Type> fieldTypes);
 
     using iterator = Block::iterator;
     iterator begin() { return getBodyBlock()->begin(); }
--- a/include/circt/Dialect/FIRRTL/FIRRTLUtils.h
+++ b/include/circt/Dialect/FIRRTL/FIRRTLUtils.h
@@ -228,7 +228,8 @@ inline FIRRTLBaseType getBaseType(Type type) {
 }
 
 /// Get base type if isa<> the requested type, else null.
-template <typename T> inline T getBaseOfType(Type type) {
+template <typename T>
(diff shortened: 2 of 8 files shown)


lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -199,8 +209,16 @@ void ExtractInstancesPass::runOnOperation() {
   if (anyFailures)
     return signalPassFailure();
 
+  ClassOp sifiveMetadata;
+  for (ClassOp classOp : circtOp.getOps<ClassOp>()) {
+    if (classOp.getSymNameAttr().getValue().compare("SiFive_Metadata"))
```
uenoku, line 214: `compare` returns non-zero when they are different so probably this is not intended. Could you use symbol table look up instead of manual string comparison?

include/circt/Dialect/FIRRTL/FIRRTLUtils.h
```diff
@@ -322,6 +322,20 @@ static ResultTy transformReduce(MLIRContext *context, RangeTy &&r,
 /// Truncate `a` to the common prefix of `a` and `b`.
 void makeCommonPrefix(SmallString<64> &a, StringRef b);
 
+//===----------------------------------------------------------------------===//
+// OM dialect utilities
+//===----------------------------------------------------------------------===//
+
+/// Create a ClassOp, with the specified fieldNames and fieldTypes as ports. The
+/// output property is set from the input property port.
+ClassOp buildSimpleClassOp(OpBuilder &odsBuilder, Location loc, Twine name,
```
uenoku, line 331: Could you define this as a class op builder method?

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -1011,7 +1034,10 @@ void ExtractInstancesPass::createTraceFiles() {
   // Generate the trace files.
   SmallVector<Attribute> symbols;
   SmallDenseMap<Attribute, unsigned> symbolIndices;
+  if (sifiveMetadataClass && !extractMetadataClass)
+    createSchema();
```
mikeurbach, line 1038: Just want to be sure--createSchema and createPathRef might modify the top-level builtin.module, so is this happening in a single thread?

mikeurbach (commented):
The gist of this looks good, just some minor suggestions and questions.

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
prithayan: Right, this is all in a single thread.

include/circt/Dialect/FIRRTL/FIRRTLUtils.h
prithayan: Thanks for the suggestion, moved it to a class op builder.

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
prithayan: Thanks Hideto, The `compare` was used as intended here, but using the symboltable makes sense. Updated to use it.

prithayan:
@mikeurbach , @uenoku, this PR is ready for another round of review, please take a look.

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -361,6 +361,13 @@ def ClassOp : FIRRTLModuleLike<"class", [
   let extraModuleClassDeclaration = [{
     Block *getBodyBlock() { return &getBody().front(); }
 
+    // This builds a ClassOp, with the specified fieldNames and fieldTypes as
+    // ports. The output property is set from the input property port.
+    ClassOp static buildSimpleClassOp(
+    mlir::OpBuilder &odsBuilder, mlir::Location loc, mlir::Twine name,
```
mikeurbach, line 367: uber nit: maybe indent this and the following lines?

include/circt/Dialect/FIRRTL/FIRRTLUtils.h
```diff
@@ -322,6 +322,14 @@ static ResultTy transformReduce(MLIRContext *context, RangeTy &&r,
 /// Truncate `a` to the common prefix of `a` and `b`.
 void makeCommonPrefix(SmallString<64> &a, StringRef b);
 
+//===----------------------------------------------------------------------===//
+// OM dialect utilities
```
mikeurbach, line 326: nit: it's the FIRRTL dialect, but "object" utilities. Maybe call it something like "FIRRTL dialect object utilities" or something like that?

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -361,6 +361,13 @@ def ClassOp : FIRRTLModuleLike<"class", [
...
+    mlir::ArrayRef<mlir::StringRef> fieldNames,
+    mlir::ArrayRef<mlir::Type> fieldTypes);
```
uenoku, line 369: Can it be moved to the section `let builders = [..]`?

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -159,12 +163,18 @@ struct ExtractInstancesPass
...
+  /// Cache of the inner ref to the new instances created. Will be used to
+  /// create a path to the instance
+  DenseMap<InnerRefAttr, InstanceOp> innerRefToInstances;
 };
 } // end anonymous namespace
 
 /// Emit the annotated source code for black boxes in a circuit.
 void ExtractInstancesPass::runOnOperation() {
-  CircuitOp circuitOp = getOperation();
+  circtOp = getOperation();
```
mikeurbach, line 177: nit: I don't think this change is needed in this PR?

mikeurbach: specifically, the name could still be `circuitOp`.

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -159,12 +163,18 @@ struct ExtractInstancesPass
   CircuitNamespace circuitNamespace;
   /// Cached module namespaces.
   DenseMap<Operation *, hw::InnerSymbolNamespace> moduleNamespaces;
+  /// The metadata class ops.
+  ClassOp extractMetadataClass, schemaClass;
+  const unsigned prefixNameFieldId = 0, pathFieldId = 2, fileNameFieldId = 4;
+  /// Cache of the inner ref to the new instances created. Will be used to
+  /// create a path to the instance
+  DenseMap<InnerRefAttr, InstanceOp> innerRefToInstances;
```
mikeurbach, line 171: this, and the `extractMetadataClass` and `schemaClass` variables should be reset in `runOnOperation` for good hygiene, even if we only ever run this pass once on the whole IR.

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -1080,8 +1151,54 @@ void ExtractInstancesPass::createTraceFiles() {
...
+             extractMetadataClass,
+             builderOM.getStringAttr("extract_instances_metadata")),
+         "extractedInstances_field"}};
+
+    addPortsToClass(classFields, sifiveMetadataClass);
+    auto *node = instanceGraph->lookup(sifiveMetadataClass);
+    assert(node && node->hasOneUse());
+    ObjectOp metadataObj =
+        dyn_cast_or_null<ObjectOp>((*node->usesBegin())->getInstance());
+    assert(metadataObj);
```
mikeurbach, line 1172: If it's just going to be asserted to exist with no error message, maybe you can just use `cast` above? Or, do something like `assert(metadataObj && "expected to find a metadataObj")`

mikeurbach (approved):
A few more minor comments from me, but I think this is looking good.

PR #7668, opened 2024-10-05
[FIRRTL][CAPI] Add documentation

SpriteOvO:
Add documentation for all FIRRTL C-API public items.

### DIFF
--- a/include/circt-c/Dialect/FIRRTL.h
+++ b/include/circt-c/Dialect/FIRRTL.h
@@ -5,6 +5,13 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 //===----------------------------------------------------------------------===//
+//
+// This file declares the C-API for FIRRTL dialect.
+//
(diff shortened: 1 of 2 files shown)


sequencer:
I think we can add more documentation for the unittest for [`firrtl.c`](https://github.com/llvm/circt/blob/09f07f5a601a264d69bf6d1c65fcd9a6b067497e/test/CAPI/firrtl.c#L86)? and provide a link to the panamaconverter usage from chisel.

SpriteOvO:
> I think we can add more documentation for the unittest for [`firrtl.c`](https://github.com/llvm/circt/blob/09f07f5a601a264d69bf6d1c65fcd9a6b067497e/test/CAPI/firrtl.c#L86)? and provide a link to the panamaconverter usage from chisel.

IMO, the test doesn't really need to be doc, they are all MLIR stuff, and these API are all already well documented.

I have added a link to Chisel example in the comment.

PR #7669, opened 2024-10-06
[SCFToCalyx] Build control for nested if operation

jiahanxie353:
This patch supports `BuildControl` for nested `scf::If` operations

### DIFF
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -1595,8 +1595,13 @@ private:
             rewriter.create<calyx::SeqOp>(ifOp.getThenRegion().getLoc());
         auto *thenSeqOpBlock = thenSeqOp.getBodyBlock();
 
-        rewriter.setInsertionPointToEnd(thenSeqOpBlock);
+        auto *thenBlock = &ifOp.getThenRegion().front();
+        LogicalResult res = buildCFGControl(path, rewriter, thenSeqOpBlock,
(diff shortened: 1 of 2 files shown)


rachitnigam:
PR needs description. Please add one and then tag reviewers!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1551,8 +1551,13 @@ class BuildControl : public calyx::FuncOpPartialLoweringPattern {
             rewriter.create<calyx::SeqOp>(ifOp.getThenRegion().getLoc());
         auto *thenSeqOpBlock = thenSeqOp.getBodyBlock();
 
-        rewriter.setInsertionPointToEnd(thenSeqOpBlock);
+        auto *thenBlock = &ifOp.getThenRegion().front();
+        LogicalResult res = buildCFGControl(path, rewriter, thenSeqOpBlock,
```
cgyurgyik, line 1555: Just to make sure I'm understanding correctly: your code is only supporting one level of nesting of `if` statements?

jiahanxie353: Arbitrary level because `buildCFGControl` is a recursive function

cgyurgyik: Ah got it, so initially it was just ignoring the `then` and `else` blocks of the `if` control.

cgyurgyik (approved):
LGTM! Thanks!

PR #7671, opened 2024-10-06
[SCFToCalyx] Memory banking for Calyx

jiahanxie353:
Breaking down #7409 

This patch will support memory banking in Calyx by user specifying the number of available banks for each memory (in the order they are declared in the source `.mlir` file) in a `json` file

### DIFF
--- a/include/circt/Conversion/Passes.td
+++ b/include/circt/Conversion/Passes.td
@@ -182,7 +182,9 @@ def SCFToCalyx : Pass<"lower-scf-to-calyx", "mlir::ModuleOp"> {
             "Identifier of top-level function to be the entry-point component"
             " of the Calyx program.">,
     Option<"ciderSourceLocationMetadata", "cider-source-location-metadata", "bool", "",
-            "Whether to track source location for the Cider debugger.">
+            "Whether to track source location for the Cider debugger.">,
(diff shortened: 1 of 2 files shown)


rachitnigam:
This should be done using attributes on memories.

jiahanxie353:
> This should be done using attributes on memories. An extra JSON file that lives separate from the source program seems unnecessary

I agree, and I have thought of that, too. The cons are users have to go to the source `.mlir` file and manually insert attributes like `{calyx.num_banks = n}` to each `memref::alloc`, `memref::alloca`, `memref::getglobal`, etc. I thought JSON could be a automated way, but using attributes does make more sense.

And I'd appreciate it if anyone could let me know that attribute insertion can be automated.

rachitnigam:
> And I'd appreciate it if anyone could let me know that attribute insertion can be automated.

Attributes can be automatically inserted by frontends.

jiahanxie353:
> Attributes can be automatically inserted by frontends.

That's great!

And just removed the needs for passing JSON and now only relies on attributes on memory allocations

rachitnigam:
Cool! Can you add tests and fix the failures? I can review it then and we can merge after the review.

PR #7672, opened 2024-10-07
[LLHD][HW] Implement SROA interfaces

maerhart:
Implement SROA to destructure signals used by `llhd.sig.array_get` (with constant index) and `llhd.sig.struct_extract`.

Let me know if you'd prefer the interfaces implemented via external models.


fabianschuiki (approved):
Very nice! LGTM.

I wasn't aware of `SafeMemorySlotAccessOpInterface`. Pretty neat!

Issue #7675, opened 2024-10-08
[FIRRTL] InferResets FullReset (formerly FART) does not add resets to not reset fields of partially reset registers


jackkoenig:
The original iteration of this bug was fixed in https://github.com/llvm/circt/pull/6912, but it appears that fix only works for registers in modules contain the wire or port marked with a `circt.FullResetAnnotation`, but does not work for children of the module with the annotated port or wire.

Consider the following FIRRTL:
```firrtl
FIRRTL version 3.3.0
; CHECK-LABEL: module test(
circuit top :%[[{
    "class":"circt.FullResetAnnotation",
    "target":"~top|top>reset",
    "resetType":"async"
  }]]
  module test :
    input clock : Clock
    input reset : AsyncReset
    input in : { foo : UInt<8>, bar : UInt<8>}
    output out : { foo : UInt<8>, bar : UInt<8>}

    wire reg1_w : { foo : UInt<8>, bar : UInt<8>}
    invalidate reg1_w
    ; CHECK: reg1_foo <= 8'hC;
    ; CHECK: reg1_bar <= 8'h0;
    connect reg1_w.foo, UInt<8>(0hc)
    invalidate reg1_w.bar
    ; CHECK: reg1_foo = 8'hC;
    ; CHECK: reg1_bar = 8'h0;
    regreset reg1 : { foo : UInt<8>, bar : UInt<8>}, clock, reset, reg1_w
    wire reg2 : { foo : UInt<8>, bar : UInt<8>}
    connect reg1, in
    connect reg2, reg1
    connect out, reg2

  module top :
    input clock : Clock
    input reset : AsyncReset
    input in : { foo : UInt<8>, bar : UInt<8>}
    output out : { foo : UInt<8>, bar : UInt<8>}

    inst child of test
    connect child.clock, clock
    connect child.reset, reset
    connect child.in, in
    connect out, child.out
```

Note that this is exactly the same as the first test in https://github.com/llvm/circt/blob/f75bbd7986d4d5ab254ce428492f0e2366c16055/test/firtool/async-reset.fir#L5, except that there is a new top module wrapping the module with registers, and the annotated port is in the top module.

Currently, `reg1.bar` does not properly get the expected async reset to `0`, but it should. If you change the annotation to target `~top|test>reset`, you will see the correct output.

PR #7680, opened 2024-10-08
[FIRRTL] InferResets: properly lower FART'd registers

youngar:
This is quick fix for https://github.com/llvm/circt/issues/7675.

The way this is supposed to work is, FRT is not supposed to add a reset if a register already has one, but it isn't smart enough to understand that an invalid value corresponds to a register without a reset. As a result, it will not add a reset to this register. Later on, in SFCCompat, when choosing how to lower the Invalid value, it should "see" that the registers should all have a reset, by looking for the FRT annotation, and choose to lower the reset value to 0 or remove the reset accordingly. The "bug" is that the FRT annotation on a parent module won't be discovered on the child module.

This attaches the annotation on to every module in a reset domain, so that it can be found.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
@@ -1685,6 +1685,14 @@ LogicalResult InferResetsPass::implementFullReset(FModuleOp module,
     return success();
   }
 
+  // Add an annotation indicating that this module belongs to a reset domain.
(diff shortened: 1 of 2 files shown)


jackkoenig (approved):
LGTM! Thanks Andrew!

saikmaddali:
@jackkoenig ,
Could u pls work with Bunnaroath to get this backported to New Reindeer Branch ?

Issue #7630, opened 2024-09-25
[MooreToCore] Support assert, assume, cover ops


fabianschuiki:
Add conversion patterns from `moore.{assert,assume,cover}` to the corresponding `verif.*` op in MooreToConv. Consider rejecting deferred assertions for now. https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+16.2+assert_test

mingzheTerapines:
@fabianschuiki  
https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+16.2+assert_test
and
https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+16.4+assert0_test
all cleared, happy to see this issue could be closed =)

fabianschuiki:
Very exciting :partying_face:! Thanks for adding support for this in #7681 :smile_cat:

fabianschuiki:
Also looks like we're up to 42% of tests passing now: https://chipsalliance.github.io/sv-tests-results/history :tada:

PR #7681, opened 2024-10-08
[MooreToCore] Support assert, assume, cover ops

mingzheTerapines:
Add conversion patterns from moore.{assert,assume,cover} to the corresponding verif.* op in MooreToConv. 
Consider rejecting deferred assertions for now. https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+16.2+assert_test
Fix issue https://github.com/llvm/circt/issues/7630


fabianschuiki:
Hey @mingzheTerapines, thanks for working on this 🙌! Looking at the SV standard, IEEE 1800-2017 § 16.4 "Deferred assertions" says:
> As with all immediate assertions, a deferred assertion’s expression is evaluated *at the time the deferred assertion statement is processed*. However, in order to facilitate glitch avoidance, the reporting or action blocks are scheduled at a later point in the current time step.

So the condition being checked is handled the same way for `assert`, `assert #0`, and `assert final`. You don't even need to put it into an `llhd.final`. Instead, you can just generate the code for the checking in the same way as you do for regular `assert`. The only difference is when the action block will be executed:
> - For an observed deferred assertion, the subroutine shall be scheduled in the *Reactive* region. Actual argument expressions that are passed by reference use or assign the current values of the underlying variables in the Reactive region.
> - For a final deferred assertion, the subroutine shall be scheduled in the *Postponed* region. Actual argument expressions that are passed by reference use the current values of the underlying variables in the Postponed region.

We don't currently differentiate between the different scheduling regions in SV. In the future we'll definitely want to do that, though. Doing so requires us to figure out how we can describe the regions in the LLHD dialect and how we can make them work in a simulator like Arcilator.

Maybe for the time being you can just treat `#0` and `final` assertions the exact same way as regular assertions? Once we have figure out how to deal with the subtleties of scheduling regions, we can revisit this lowering and improve it.

mingzheTerapines:
@fabianschuiki Well, that will make coding easier. 
But in verilator, `assert` may trigger `$ERROR` or something, do we need this trigger explictly in operations?

test/Conversion/MooreToCore/basic.mlir
```diff
@@ -836,3 +836,16 @@ dbg.variable "b", %dbg0 scope %dbg1 : !moore.l32
...
+
+// CHECK-LABEL: hw.module @Assert
+moore.module @Assert(in %cond : !moore.l1)  {
+  moore.procedure always {
+  // CHECK: verif.assert %cond label "" : i1
+  moore.assert immediate %cond : l1
+  // CHECK: verif.assume %cond label "" : i1
+  moore.assume observed %cond  : l1
+  // CHECK: verif.cover %cond label "" : i1
+  moore.cover final %cond : l1
```
fabianschuiki, line 848: Can you also add a check that has a non-empty label?

fabianschuiki (approved):
LGTM, thanks!

fabianschuiki:
I suggest we leave the termination on asserts up to how the simulator, e.g. arcilator, interprets the `verif.assert` op you create :smiley:

test/Conversion/MooreToCore/basic.mlir
mingzheTerapines: Sure, sorry for missing this check

fabianschuiki (approved):

PR #7684, opened 2024-10-08
[JSON] Leave json::OStream in valid state on failure

fabianschuiki:
Make sure that calls to begin/end functions are balanced even when an error occurs. This leaves the `json::OStream` object in a valid state and allows surrounding code to handle the failure gracefully. Otherwise an assertion in the destructor of `json::OStream` would fire because of unclosed JSON scopes.


uenoku (approved):
LGTM

PR #7686, opened 2024-10-09
[FIRRTL] collection of changes to InferWidths

youngar:
This PR has a collection of mostly non-functional changes to InferWidths, with the aim of cleaning up some of the code and improving performance.


lib/Dialect/FIRRTL/Transforms/InferWidths.cpp
```diff
@@ -2025,14 +1992,24 @@ Expr *InferenceMapping::getExpr(Value value) const {
...
+  if (it != opExprs.end())
+    return it->second;
+  // If we don't have an expression for this fieldRef, it should have a
+  // constant width.
+  auto baseType = getBaseType(fieldRef.getValue().getType());
+  auto type =
+      hw::FieldIdImpl::getFinalTypeByFieldID(baseType, fieldRef.getFieldID());
+  auto width = cast<FIRRTLBaseType>(type).getBitWidthOrSentinel();
+  if (width < 0)
+    return nullptr;
```
uenoku, line 2011: This is really smart!

uenoku (approved):
LGTM, thank you for cleaning up a lot of stuffs. I think lazy creation of KnownExprOp is definitely safe and great improvement. Removing duplication of InvalidValueOp change would be a breaking change but seems more reasonable. 

I checked a diff on one of largest cores and there was no change.

dtzSiFive:
Awesome work, thanks!!!

Issue #7682, opened 2024-10-08
[Calyx] IfOp's group has to be a CombGroupOp?


jiahanxie353:
Does `IfOp`'s `Group` has to be a `CombGroupOp`, as required [here](https://github.com/llvm/circt/blob/e76d2644a7782f87497668efbc57e76365b887c1/lib/Dialect/Calyx/CalyxOps.cpp#L2228)?

What if we have some examples like:
```mlir
  func.func @main(%arg0 : i32) {
    %two = arith.constant 2: i32
    %rem = arith.remui %arg0, %two : i32
    %cond0 = arith.cmpi eq, %arg0, %rem : i32

    scf.if %cond0 {
      some operations
    } else {
      some operations
    }
```
where `%cond0` involves comparing the `out` of `remui_pipe_op`, which is [not combinational](https://github.com/llvm/circt/blob/e76d2644a7782f87497668efbc57e76365b887c1/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp#L651)?

What was the original design criteria? I believe @mortbopet is the original author. Also tagging  @rachitnigam @cgyurgyik

cgyurgyik:
This was probably a requirement that existed when this was initially implemented: https://docs.calyxir.org/lang/ref.html#if

jiahanxie353:
> This was probably a requirement that existed when this was initially implemented: https://docs.calyxir.org/lang/ref.html#if

Ahh indeed, thanks for pointing that out!

Is it true that `if` statements with `with` `group`s  is still an ongoing discussion: https://github.com/calyxir/calyx/issues/1699?

Semantically, I believe my example makes sense, where one of the `lhs`/`rhs` of the comparison is a result of some operation, which can be either a combinational circuit or sequential ones (like `multi_pipe_op`/`divi_pipe_op`/`remui_pipe_op`, since the upstream tools should be unaware if an operation is implemented combinationally or sequentially in hardware?).

As separate question, is there any workaround over this Well-formedness check, so that we can still have `if %cond with comb_group`, and that `comb_group` somehow compares `remui_pipe_op.out` with `%arg0` combinationally?

rachitnigam:
No, for non-combinational computations, you need to manually perform the computation in a group, store the value in a register, and then use the value of the register to perform the computation. For example:
```
compute_cond; // computes rem and store in reg
if reg.out { ... } else { ... }
```

cgyurgyik:
This is good to close @jiahanxie353 ?

[...]

PR #7687, opened 2024-10-09
[SCFToCalyx] If op with sequential condition

jiahanxie353:
This patch addresses #7682. It supports IfOp whose condition check involves sequential operations by enabling `buildLibraryOp` to create `GroupOp`.


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -240,13 +255,36 @@ class ForLoopLoweringStateInterface
   }
 };
 
+class PipeOpLoweringStateInterface {
```
jiahanxie353, line 258: I'm not very satisfied with this approach though, having to introduce another interface to "remember" the register set for the sequential/pipelined library operator.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -125,6 +125,20 @@ using Scheduleable =
 
 class IfLoweringStateInterface {
 public:
+  void setCondReg(scf::IfOp op, calyx::RegisterOp regOp) {
+    Operation *operation = op.getOperation();
+    assert(condReg.count(operation) == 0 &&
+           "A condition register was already set for this scf::IfOp!\n");
+    condReg[operation] = regOp;
```
cgyurgyik, line 132: To avoid the double lookup here (once for `count` and once for the `[]` operator, we can do:

```cpp
auto [it, succeeded] = condReg.insert(std::make_pair(operation, regOp));
assert(succeeded && "...");
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -172,6 +186,7 @@ class IfLoweringStateInterface {
   }
 
 private:
+  DenseMap<Operation *, calyx::RegisterOp> condReg;
```
cgyurgyik, line 189: I know this hasn't been the norm, but it would sure be nice to describe why this data structure exists for future readers :-)

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -240,13 +255,36 @@ class ForLoopLoweringStateInterface
   }
 };
 
+class PipeOpLoweringStateInterface {
+public:
+  void setPipeResReg(Operation *op, calyx::RegisterOp reg) {
+    assert(isa<calyx::MultPipeLibOp>(op) || isa<calyx::DivUPipeLibOp>(op) ||
```
cgyurgyik, line 261: prefer: `isa<calyx::MultPipeLibOp, calyx::DivUPipelibOp, ...>(op)`

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -240,13 +255,36 @@ class ForLoopLoweringStateInterface
...
+           isa<calyx::DivSPipeLibOp>(op) || isa<calyx::RemUPipeLibOp>(op) ||
+           isa<calyx::RemSPipeLibOp>(op));
+    assert(resultRegs.count(op) == 0 &&
```
cgyurgyik, line 264: Similar comment as above.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -365,26 +408,54 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
...
-      rewriter.create<calyx::AssignOp>(op.getLoc(), dstOp.value(),
-                                       op->getOperand(dstOp.index()));
+
+    for (auto dstOp : enumerate(opInputPorts)) {
+      if (isSequential)
+        rewriter.create<calyx::AssignOp>(op.getLoc(), dstOp.value(),
+                                         srcReg.getOut());
+      else
+        rewriter.create<calyx::AssignOp>(op.getLoc(), dstOp.value(),
+                                         op->getOperand(dstOp.index()));
```
cgyurgyik, line 426:
```suggestion
auto srcOp = isSequential ? srcReg.getOut() : op->getOperand(dstOp.index());
rewriter.create<calyx::AssignOp>(op.getLoc(), dstOp.value(), srcOp);
```

Maybe this doesn't work, but I would try to limit the scope of your conditional.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -365,26 +408,54 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
...
 
     /// Replace the result values of the source operator with the new operator.
     for (auto res : enumerate(opOutputPorts)) {
       getState<ComponentLoweringState>().registerEvaluatingGroup(res.value(),
                                                                  group);
-      op->getResult(res.index()).replaceAllUsesWith(res.value());
+      if (isSequential)
+        op->getResult(res.index()).replaceAllUsesWith(dstReg.getOut());
+      else
+        op->getResult(res.index()).replaceAllUsesWith(res.value());
```
cgyurgyik, line 436: same here.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -939,9 +1013,43 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+          getState<ComponentLoweringState>().getUniqueName(opName));
+
+      for (auto *user : op->getUsers()) {
+        if (auto ifOp = dyn_cast<scf::IfOp>(user))
+          getState<ComponentLoweringState>().setCondReg(ifOp, condReg);
+      }
+
+      calyx::RegisterOp pipeResReg;
+      if (isPipeLibOp(op.getLhs()))
+        pipeResReg = getState<ComponentLoweringState>().getPipeResReg(
```
cgyurgyik, line 1042: Please use curly braces if doing `if/else` / multi-line conditionals.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -125,6 +125,20 @@ using Scheduleable =
...
 class IfLoweringStateInterface {
 public:
+  void setCondReg(scf::IfOp op, calyx::RegisterOp regOp) {
+    Operation *operation = op.getOperation();
+    assert(condReg.count(operation) == 0 &&
+           "A condition register was already set for this scf::IfOp!\n");
+    condReg[operation] = regOp;
+  }
+
+  calyx::RegisterOp getCondReg(scf::IfOp op) {
```
cgyurgyik, line 135: Is this needed if its only used once?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -939,9 +1013,43 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+      return isa<calyx::MultPipeLibOp, calyx::DivUPipeLibOp,
+                 calyx::DivSPipeLibOp, calyx::RemUPipeLibOp,
+                 calyx::RemSPipeLibOp>(defOp);
+    }
+    return false;
+  };
+
   switch (op.getPredicate()) {
-  case CmpIPredicate::eq:
+  case CmpIPredicate::eq: {
```
cgyurgyik, line 1026: Isn't this the case for any conditional predicate used in a Calyx `if`? Just trying to make sure I understand correctly what's going on here. If this is true, don't feel pressured to make those changes now but you should add some assertions so that there is some sort of failure.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -240,13 +255,36 @@ class ForLoopLoweringStateInterface
...
+  // Get the register for a specific pipe operation
+  calyx::RegisterOp getPipeResReg(Operation *op) {
+    auto it = resultRegs.find(op);
+    assert(it != resultRegs.end() &&
+           "No register was set for this pipe operation!\n");
+    return it->second;
+  }
+
+private:
+  DenseMap<Operation *, calyx::RegisterOp> resultRegs;
```
cgyurgyik, line 277: A comment would be nice here!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -240,13 +255,36 @@ class ForLoopLoweringStateInterface
...
+  void setPipeResReg(Operation *op, calyx::RegisterOp reg) {
+    assert(isa<calyx::MultPipeLibOp>(op) || isa<calyx::DivUPipeLibOp>(op) ||
+           isa<calyx::DivSPipeLibOp>(op) || isa<calyx::RemUPipeLibOp>(op) ||
+           isa<calyx::RemSPipeLibOp>(op));
+    assert(resultRegs.count(op) == 0 &&
+           "A register was already set for this pipe operation!\n");
+    resultRegs[op] = reg;
+  }
+  // Get the register for a specific pipe operation
+  calyx::RegisterOp getPipeResReg(Operation *op) {
```
cgyurgyik, line 269: Similarly for this function, it is used twice but only because you condition on an entire function call rather than the one element that is different: the function argument. I think we should avoid introducing this unless you think it improves readability.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
cgyurgyik: Well, we'll definitely need some way to recall this mapping, since it isn't explicitly stated in the IR. If this is only necessary for lowering conditionals in `if` statements, then I think it would make more sense in the `IfLoweringStateInterface`.

With that said, I don't think this is the only Calyx control flow that will require this special treatment, i.e., `while` should be similar. What other approaches are you thinking about?

In any case, it would be nice to add documentation on why this interface exists.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -339,7 +377,12 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
   /// source operation TSrcOp.
   template <typename TGroupOp, typename TCalyxLibOp, typename TSrcOp>
   LogicalResult buildLibraryOp(PatternRewriter &rewriter, TSrcOp op,
-                               TypeRange srcTypes, TypeRange dstTypes) const {
+                               TypeRange srcTypes, TypeRange dstTypes,
+                               calyx::RegisterOp srcReg = nullptr,
```
cgyurgyik, line 381: Please add comments about these.

test/Conversion/SCFToCalyx/convert_controlflow.mlir
```diff
@@ -641,3 +641,122 @@ module {
...
+// CHECK:           }
+// CHECK:         } {toplevel}
+  func.func @main(%arg0 : i32) -> i32 {
+    %1 = memref.alloc() : memref<120xi32>
+    %idx_one = arith.constant 1 : index
+    %two = arith.constant 2: i32
+    %rem = arith.remui %arg0, %two : i32
+    %cond = arith.cmpi eq, %arg0, %rem : i32
+
+    %res = scf.if %cond -> i32 {
```
cgyurgyik, line 750: Nit: could we minimize what you're trying to test even further, e.g., do we need the `else` case, mem alloc, ...? It seems the critical piece is an if statement that uses the output of a sequential operation.

cgyurgyik (commented):
Just to check my understanding, this PR is ensuring that if the condition of a SCF `if` op depends on some sequential operation, we create a separate group and save its value in a register to be used in the Calyx `if` op. Is that correct?

jiahanxie353:
> Just to check my understanding, this PR is ensuring that if the condition of a SCF if op depends on some sequential operation, we create a separate group and save its value in a register to be used in the Calyx if op. Is that correct?

That's exactly right! :-)

And to reply the following comment as a whole:
> With that said, I don't think this is the only Calyx control flow that will require this special treatment, i.e., while should be similar. What other approaches are you thinking about?

> In any case, it would be nice to add documentation on why this interface exists.

Let me shamelessly copy your wording to describe the feature I'm trying to support in general:
This (and future) PR  is ensuring that if the condition of a SCF if/while op depends on some sequential operation, we create a separate group and save its value in a register to be used in the Calyx if/while op.

> I don't think this is the only Calyx control flow that will require this special treatment

So you are right, both if and while need it.

> What other approaches are you thinking about?

Currently, my implementation has the following lowered result:
```mlir
      calyx.group @bb0_0 {
        calyx.assign %std_remu_pipe_0.left = %in0 : i32
        calyx.assign %std_remu_pipe_0.right = %c2_i32 : i32
        calyx.assign %remui_0_reg.in = %std_remu_pipe_0.out_remainder : i32
        calyx.assign %remui_0_reg.write_en = %std_remu_pipe_0.done : i1
        %0 = comb.xor %std_remu_pipe_0.done, %true : i1
        calyx.assign %std_remu_pipe_0.go = %0 ? %true : i1
        calyx.group_done %remui_0_reg.done : i1
      }
      calyx.group @bb0_1 {
        calyx.assign %std_eq_0.left = %remui_0_reg.out : i32
        calyx.assign %std_eq_0.right = %remui_0_reg.out : i32
        calyx.assign %cmpi_0_reg.in = %std_eq_0.out : i1
        calyx.assign %cmpi_0_reg.write_en = %true : i1
        calyx.group_done %cmpi_0_reg.done : i1
      }
    calyx.control {
        calyx.seq {
          calyx.enable @bb0_0
          calyx.enable @bb0_1
          calyx.if %cmpi_0_reg.out {
              some operation
```
I wonder, is it possible to do:
```mlir
      calyx.group @bb0_0 {
        calyx.assign %std_remu_pipe_0.left = %in0 : i32
        calyx.assign %std_remu_pipe_0.right = %c2_i32 : i32
        calyx.assign %remui_0_reg.in = %std_remu_pipe_0.out_remainder : i32
        calyx.assign %remui_0_reg.write_en = %std_remu_pipe_0.done : i1
        %0 = comb.xor %std_remu_pipe_0.done, %true : i1
        calyx.assign %std_remu_pipe_0.go = %0 ? %true : i1
        calyx.group_done %remui_0_reg.done : i1
      }
      calyx.comb_group @bb0_1 {
        calyx.assign %std_eq_0.left = %remui_0_reg.out : i32
        calyx.assign %std_eq_0.right = %remui_0_reg.out : i32
      }

    calyx.control {
        calyx.seq {
          calyx.enable @bb0_0
          calyx.if %std_eq_0.out with @bb0_1 {
              some operation

```
So that we can save a cycle, but I'm not sure what needs to be check with respect to the stability of `%std_eq_0.out with @bb0_1` 

@rachitnigam recommends the first approach though, as he mentioned in [the issue](https://github.com/llvm/circt/issues/7682):
> compute_cond; // computes rem and store in reg
> if reg.out { ... } else { ... }

cgyurgyik:
> I wonder, is it possible to do:
> 
> ```mlir
>       calyx.group @bb0_0 {
>         calyx.assign %std_remu_pipe_0.left = %in0 : i32
>         calyx.assign %std_remu_pipe_0.right = %c2_i32 : i32
>         calyx.assign %remui_0_reg.in = %std_remu_pipe_0.out_remainder : i32
>         calyx.assign %remui_0_reg.write_en = %std_remu_pipe_0.done : i1
>         %0 = comb.xor %std_remu_pipe_0.done, %true : i1
>         calyx.assign %std_remu_pipe_0.go = %0 ? %true : i1
>         calyx.group_done %remui_0_reg.done : i1
>       }
>       calyx.comb_group @bb0_1 {
>         calyx.assign %std_eq_0.left = %remui_0_reg.out : i32
>         calyx.assign %std_eq_0.right = %remui_0_reg.out : i32
>       }
> 
>     calyx.control {
>         calyx.seq {
>           calyx.enable @bb0_0
>           calyx.if %std_eq_0.out with @bb0_1 {
>               some operation
> ```
> 
> So that we can save a cycle, but I'm not sure what needs to be check with respect to the stability of `%std_eq_0.out with @bb0_1`

My first thought is this additional step could be enabled as a peephole optimization starting from the `calyx.if`, i.e., 


```mlir
calyx.enable @bb0_1
calyx.if %cmpi_0_reg.out {
```
-->
```mlir
calyx.enable @bb0_1
calyx.if %std_eq_0.out with @bb0_1 {
```

where we get rid of an intermediate register.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -365,26 +408,54 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
...
+                                                              groupOp);
+    }
+
     rewriter.setInsertionPointToEnd(group.getBodyBlock());
-    for (auto dstOp : enumerate(opInputPorts))
-      rewriter.create<calyx::AssignOp>(op.getLoc(), dstOp.value(),
-                                       op->getOperand(dstOp.index()));
+
+    for (auto dstOp : enumerate(opInputPorts)) {
+      if (isSequential)
```
jiahanxie353, line 421: Bug, should be:
```suggestion
      if (isPipeLibOp(dstOp.value()))
        rewriter.create<calyx::AssignOp>(op.getLoc(), dstOp.value(),
                                         srcReg.getOut());
```

Then seems like we are checking `isPipeLibOp` twice, once in a caller function, another in a callee..

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: You are right, any comparison operator can be a condition check.
I'm only adding `eq` for now to use it as a prototype since some approaches need to be discussed (like the `PipeLoweringState`). Once we reach agreement, I will add support for all `CmpIPredicate`, which will just be cloning the same code from `eq`.

jiahanxie353:
Sorry for the late late reply. Just did some changes and refactor based on the review; and it's ready for another round of iteration!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -444,6 +444,41 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
                         scf::ExecuteRegionOp executeRegionOp) const;
   LogicalResult buildOp(PatternRewriter &rewriter, CallOp callOp) const;
 
+  void handleCmpIOp(PatternRewriter &rewriter, CmpIOp cmpIOp, Operation *group,
+                    calyx::RegisterOp &condReg, calyx::RegisterOp &resReg,
+                    bool &isSequential) const {
+    isSequential = calyx::isPipeLibOpRes(cmpIOp.getLhs()) ||
+                   calyx::isPipeLibOpRes(cmpIOp.getRhs());
+    if (isSequential) {
```
cgyurgyik, line 452: Nit: can we invert this to reduce indentation?

if (!isSequential) { return; }

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -881,6 +881,8 @@ Type toBitVector(T type) {
   return type;
 }
 
+bool isPipeLibOpRes(Value value);
```
cgyurgyik, line 884: Document this please.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: sounds good, we should!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1461,26 +1556,86 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
-  case CmpIPredicate::sle:
-    return buildLibraryOp<calyx::CombGroupOp, calyx::SleLibOp>(rewriter, op);
+  case CmpIPredicate::eq: {
+    bool isSequential = calyx::isPipeLibOpRes(op.getLhs()) ||
+                        calyx::isPipeLibOpRes(op.getRhs());
+    return isSequential
+               ? buildLibraryOp<calyx::GroupOp, calyx::EqLibOp>(rewriter, op)
+               : buildLibraryOp<calyx::CombGroupOp, calyx::EqLibOp>(rewriter,
+                                                                    op);
+  }
```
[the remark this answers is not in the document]
jiahanxie353: sure!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -255,13 +271,40 @@ class ForLoopLoweringStateInterface
...
 };
 
+class PipeOpLoweringStateInterface {
+public:
+  void setPipeResReg(Operation *op, calyx::RegisterOp reg) {
+    assert(
+        (isa<calyx::MultPipeLibOp, calyx::DivUPipeLibOp, calyx::DivSPipeLibOp,
+             calyx::RemUPipeLibOp, calyx::RemSPipeLibOp, calyx::AddFOpIEEE754,
+             calyx::MulFOpIEEE754, calyx::FpToIntOpIEEE754,
+             calyx::IntToFpOpIEEE754, calyx::DivSqrtOpIEEE754>(op)));
```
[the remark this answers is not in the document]
jiahanxie353: Actually it's very hard:
1. Because what we have right now is `CalyxCellInterface`, which exposes `isCombinational`;
but there are so many cells like `RegisterOp`, `GroupOp`, all the unary/binary ops so if we were to use `CalyxCellInterface` to tell if an operation one of the binary pipelined operation, is very hard.
2. Unless we come up with a new interface `BinaryPipeOpInterface` to attach every `CalyxPrimitive`. This solution to me, is no different from enumerating them using `isa<calyx::MultPipeLibOp, calyx::DivUPipeLibOp,`

I'm curious about what you think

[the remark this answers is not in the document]
jiahanxie353: Simplified the logic to using `isCombinational`, but also involves checking if the parent operation is a `scf.if` - probably we can non-`scf.if` cases afterwards

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -401,6 +443,58 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
...
+        width.getIntOrFloatBitWidth(),
+        getState<ComponentLoweringState>().getUniqueName(opName));
+
+    for (auto *user : cmpIOp->getUsers()) {
+      if (auto ifOp = dyn_cast<scf::IfOp>(user))
+        getState<ComponentLoweringState>().setCondReg(ifOp, condReg);
+    }
+
+    assert(calyx::isSeqLibOpRes(cmpIOp.getLhs()) !=
+           calyx::isSeqLibOpRes(cmpIOp.getRhs()));
```
cgyurgyik, line 472: I don't see why this would be illegal, so let's add a clear error message.

```suggestion
    assert(calyx::isSeqLibOpRes(cmpIOp.getLhs()) !=
           calyx::isSeqLibOpRes(cmpIOp.getRhs()) && "unexpected sequential operation on both sides; please open an issue");
```

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -881,6 +881,10 @@ Type toBitVector(T type) {
   return type;
 }
 
+// Returns `true` if `value` is the result of a sequential Calyx library
+// operation.
+bool isSeqLibOpRes(Value value);
```
cgyurgyik, line 886: Nit: maybe just `isSequentialCell` or `isSeqCell`?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -401,6 +443,58 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
                         scf::ExecuteRegionOp executeRegionOp) const;
   LogicalResult buildOp(PatternRewriter &rewriter, CallOp callOp) const;
 
+  // Sets up the necessary state and resources for a `CmpIOp` in
+  // `buildLibraryBinaryPipeOp` if `cmpIOp` has sequential logic based on its
+  // operands.
+  void setupCmpIOp(PatternRewriter &rewriter, CmpIOp cmpIOp, Operation *group,
+                   calyx::RegisterOp &condReg, calyx::RegisterOp &resReg,
+                   bool &isSeqCondCheck) const {
+    isSeqCondCheck = (calyx::isSeqLibOpRes(cmpIOp.getLhs()) ||
```
cgyurgyik, line 452: You call this at least twice. Does it make sense to store this in a local variable, e.g.,

```cpp
bool lhsIsSeqOp = calyx::isSeqLibOpRes(cmpIOp.getLhs());
bool rhsIsSeqOp = calyx::isSeqLibOpRes(cmpIOp.getRhs());

...

assert(lhsIsSeqOp != rhsIsSeqOp && "...");
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -433,17 +527,31 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
 
     /// Create assignments to the inputs of the library op.
     auto group = createGroupForOp<TGroupOp>(rewriter, op);
+
+    bool isSeqCondCheck = false;
+    calyx::RegisterOp condReg = nullptr, resReg = nullptr;
+    if (isa<CmpIOp>(op)) {
+      auto cmpIOp = cast<CmpIOp>(op);
```
cgyurgyik, line 534:
```suggestion
    if (auto cmpIOp = dyn_cast<CmpIOp>(op)) {
```

cgyurgyik (commented):
This looks much better, thanks! I have a few smaller code style comments.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: sure!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: This one requires static check so `dyn_cast` didn't work when I tried to compile it.
Other than that, I have fixed everything based on your review!

cgyurgyik (approved):
LGTM!

Issue #7224, opened 2024-06-21
[FIRRTL] LowerLayers creates empty port names, preventing round-trip


fzi-hielscher:
When we have anonymous signals cross layer boundaries (e.g, due to CSE, [like so](https://scastie.scala-lang.org/7m2Ea1gUTVKlNdwrHLIn6g)), `LowerLayers` will create an `FModuleOp` where the port name attribute for these signals is the empty string. For the most part, this does not appear to cause any issues. 
But when we serialize the IR, the name is handled differently by the printers of the `firrtl.module` and the `firrtl.instance` op respectively.

Since the module argument's SSA value name cannot be empty, the port is called "arg0" in the module declaration:
`firrtl.module private @Example_A(in %arg0: !firrtl.uint<9>, in %reset: !firrtl.uint<1>, in %clock: !firrtl.clock)`
The instantiation on the other hand will just use the empty string attribute:
`firrtl.instance a { ... } @Example_A(in "": !firrtl.uint<9>, in reset: !firrtl.uint<1>, in clock: !firrtl.clock)`
Which prevents the circuit from being parsed again:
`error: 'firrtl.instance' op name for port 0 must be "arg0", but got ""`

I don't know if it is by design or by accident that empty port names don't seem to break anything else. Are they considered legal?

(Ping @fzi-antonpaule, who found this problem)

PR #7688, opened 2024-10-09
[FIRRTL] Print port names attributes if a port has an empty name

rwy7:
If a port name is empty, then the SSA-name cannot match, and we must print the port names as an attribute instead. This PR removes a bogus check in the module printer, which prevents us from printing the port names attribute,  if the port name is empty.

With this PR, modules with empty port names should round trip correctly.

When lower-layers converts layerblocks into modules, a lot of the time there is no good name for the port, so it generates an empty port name. With this patch, these modules will now correctly print the port names attribute, so the tests had to be updated to match against the attribute dictionary.

Fixes: https://github.com/llvm/circt/issues/7224


youngar (approved):
LGTM

mikeurbach (approved):
Nice!

dtzSiFive:
Awesome, thanks!

PR #7690, opened 2024-10-10
[FIRRTL] Ensure LowerClasses respects alt base path for local paths.

mikeurbach:
We recently started supporting local-only paths in LowerClasses, but this logic erroneously disabled the alternative base path feature for paths whose targets were relocated outside the owning module. In the case that we need an alternative base path for a local target, convert it to a non-local target, so the alternative base path can be correctly prefixed.


mikeurbach:
Needs a unit test or two, but I have been testing this downstream and I think it's correct.

PR #7694, opened 2024-10-10
[MooreToCore] Add support for format strings and display task

fabianschuiki:
Lower the format string ops and the `moore.builtin.display` task to the corresponding ops in the Sim dialect. This is not perfect yet, since Sim does not support some of the formatting options. They will be easy to add in the future though.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -16,6 +16,7 @@
 #include "circt/Dialect/HW/HWOps.h"
 #include "circt/Dialect/LLHD/IR/LLHDOps.h"
 #include "circt/Dialect/Moore/MooreOps.h"
+#include "circt/Dialect/Sim/SimOps.h"
 #include "circt/Dialect/Verif/VerifOps.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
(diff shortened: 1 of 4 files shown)


hailongSun2000 (approved):
LGTM :partying_face:!

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1256,6 +1257,69 @@ struct AssertLikeOpConversion : public OpConversionPattern<MooreOpTy> {
...
+struct FormatIntOpConversion : public OpConversionPattern<FormatIntOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(FormatIntOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // TODO: These should honor the width, alignment, and padding.
+    switch (op.getFormat()) {
+    case IntFormat::Decimal:
+      rewriter.replaceOpWithNewOp<sim::FormatDecOp>(op, adaptor.getValue());
```
fzi-hielscher, line 1296: Do we have a way of handling signedness here? By not setting the `signed` attribute the value will always get formatted as unsigned. How would we deal with the difference in this example:
```SV
module tb;
  byte foo = 8'h80;
  initial $display("Signed:   %d", $signed(foo));
  initial $display("Unsigned: %d", $unsigned(foo));
endmodule
```

fabianschuiki: That's a great point. Do you know what the standard says how your example would be printed? Does `%d` check the type's signedness?

fzi-hielscher: I don't see any mention of the sign in the LRM. But for once all simulators on edaplayground agree: It depends on the type's signedness. The above example is the same as
```SV
module tb;
  logic signed [7:0] fooS = 8'h80;
  logic [7:0] fooU = 8'h80;
  initial $display("Signed:   %d", fooS); // "Signed:   -128"
  initial $display("Unsigned: %d", fooU); // "Unsigned: 128"
endmodule
```
I wish they'd also agree on the padding...

fabianschuiki: 🎉 Nice! I also love how the `%d` field expands to the maximum width a decimal number can have, 3, but then adds a random `-` in front that throws everything off 🤣. I love how SV consistently picks the wrong default for a lot of things.

PR #7697, opened 2024-10-12
[FIRRTL] InferWidths: fix invalid frame reference

youngar:
The frame.indent reference can be invalidated when we push to the worklist. This change caches the indentation level locally to fix the issue. Bug was introduced in b836e24.


seldridge (approved):
LGTM

PR #7699, opened 2024-10-14
[ArcToLLVM] Add support for index dialect

fabianschuiki:
Support the index dialect when lowering from Arc to LLVM. Also fix a minor issue in the lowering `SimGetPortOp`.


lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -440,20 +441,19 @@ struct SimGetPortOpLowering : public ModelAwarePattern<arc::SimGetPortOp> {
                            .getValue());
     ModelInfoMap &model = modelIt->second;
 
+    auto type = typeConverter->convertType(op.getValue().getType());
```
maerhart, line 444: Nit: We don't specifically restrict the type of arc model ports, right? So maybe we should check for null type here and return failure? Or at least assert?

maerhart (approved):
LGTM!

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
fabianschuiki: Good point! Fixed

PR #7700, opened 2024-10-14
[Arc] Add arc.final op

fabianschuiki:
Add the `arc.final` op to complement the existing `arc.initial`. The body of `arc.final` is executed after the last time step, when the simulation is shut down. This corresponds to SystemVerilog's `final` procedure, and can also be used to report statistics about which `cover` statements were hit, or to check if an `eventually` property has been satisfied.

The `arc.final` op is extracted into a `*_final` function. The contract with the user is that the `*_final` function must be called exactly once after the last call to `*_eval`. The simulation lifetime now looks like this:

```
design_initial()
design_eval()
design_eval()
design_eval()
...
design_eval()
design_final()
```

The `arc.sim.instantiate` op inserts a corresponding call to `*_final` when lowering to LLVM.


include/circt/Dialect/Arc/ArcOps.td
```diff
@@ -672,11 +679,15 @@ def ModelOp : ArcOp<"model", [
   }];
   let arguments = (ins SymbolNameAttr:$sym_name,
                        TypeAttrOf<ModuleType>:$io,
-                       OptionalAttr<FlatSymbolRefAttr>:$initialFn);
+                       OptionalAttr<FlatSymbolRefAttr>:$initialFn,
+                       OptionalAttr<FlatSymbolRefAttr>:$finalFn);
```
maerhart, line 683: The initial symbol has a verifier. Should we do the same for the final symbol? (and also add a test for it)

lib/Dialect/Arc/ModelInfo.cpp
```diff
@@ -116,7 +116,8 @@ LogicalResult circt::arc::collectModels(mlir::ModuleOp module,
     llvm::sort(states, [](auto &a, auto &b) { return a.offset < b.offset; });
 
     models.emplace_back(std::string(modelOp.getName()), storageType.getSize(),
-                        std::move(states), modelOp.getInitialFnAttr());
+                        std::move(states), modelOp.getInitialFnAttr(),
+                        modelOp.getFinalFnAttr());
```
maerhart, line 120: Do we want to add the finalizer function name to the JSON such that an external test bench can pick it up rather than guessing or hardcoding the function name?

maerhart (approved):
Nice! Should we also add an integration test for the final block?

lib/Dialect/Arc/ModelInfo.cpp
fabianschuiki: Great idea! I've added that :+1:

PR #7702, opened 2024-10-14
[Arc] Add dominance-aware pass to sink ops and merge scf.if ops

fabianschuiki:
Add the `MergeIfs` pass to the Arc dialect. This pass covers a handful of control flow optimizations that are valuable to pick up after `hw.module`s have been linearized and lowered into `arc.model` ops:

- It moves operations closer to their earliest user, if possible sinking them into blocks if all uses are nested in the same block.
- It merges adjacent `scf.if` operations with the same condition.
- It moves operations in between two `scf.if` operations ahead of the first if op to allow them to be merged.

The `MergeIfs` pass can operate on SSACFG regions. It assigns an integer order to each operation and considers that order to determine up to which point operations can be moved without moving beyond their first use and without crossing interfering side-effecting ops. The pass is aware of side-effects, and in particular uses the non-aliasing between `arc.state` and `arc.memory` to track read/write side-effects at a per-state level, which allows for fairly aggressive optimization. Other side-effecting ops act as a hard barrier and will not be moved or moved over.

This pass supersedes the very effective `GroupResetsAndEnables` pass we have been using until now. The latter relies on the `LegalizeStateUpdate` pass to run at a later point however, which will be removed in a future PR, thus making this new pass necessary.

This is a preparatory step for a later PR that overhauls the LowerState pass. That rewrite will make the `arc.clock_tree` and `arc.passthrough` ops obsolete, which is why they are not present in the tests.


lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,367 @@
...
+    if (ifOp.getCondition() != prevIfOp.getCondition())
+      continue;
+    if (ifOp.getNumResults() != 0 || prevIfOp.getNumResults() != 0)
+      continue;
+    if (!ifOp.getThenRegion().hasOneBlock() ||
+        !prevIfOp.getThenRegion().hasOneBlock())
+      continue;
+    if (!ifOp.getElseRegion().empty() &&
+        (!ifOp.getElseRegion().hasOneBlock() ||
+         !prevIfOp.getElseRegion().hasOneBlock()))
```
TaoBi22, line 286: What's the purpose of the `empty()` check here? If `hasOneBlock()` will error if the region is empty, can this not also happen here if `ifOp` doesn't have an empty else region but `prevIfOp` does?

TaoBi22 (approved):
This is super cool! Very cool that this detects and works with non-overlapping side effects too!

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+/// Iteratively sink ops into block and move them closer to their uses, and
+/// merge adjacent `scf.if` operations.
+LogicalResult Sinker::run() {
+  LLVM_DEBUG(llvm::dbgs() << "Running on block in "
+                          << rootBlock.getParentOp()->getName() << "\n");
+  unsigned iteration = 0;
+  do {
+    if (++iteration > 10)
+      return rootBlock.getParentOp()->emitOpError()
```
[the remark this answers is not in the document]
fabianschuiki: Under normal circumstances this should settled very quickly. Sinking operations and merging adjacent ifs can create additional optimization opportunities for each other. In the Rocket core this happens after 2-3 iterations, and when this hangs for too long it's usually an indication that there's a bug in the pass. But maybe the iteration count should rather be 100 or so.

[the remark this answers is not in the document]
fabianschuiki: Actually, mabye this should just silently stop after hitting a maximum. You're right that this is an optimization and not essential to make the model work, so you probably don't want to abort the whole build because of some arbitrary limit.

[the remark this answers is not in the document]
fabianschuiki: Or, thinking about this some more, maybe this shouldn't check an iteration count at all. If the pass works correctly it always converges. Otherwise it's a bug in the pass and you probably want this to just blow up.

lib/Dialect/Arc/Transforms/MergeIfs.cpp
fabianschuiki: Oh nice catch! This was missing a check above that ensures that either both ops have an else region, or both ops don't have an else region. Once that's done, this check makes sure that _if_ they have a non-empty else region, the two regions must have exactly one block (a limitation of my implementation).

lib/Dialect/Arc/Transforms/MergeIfs.cpp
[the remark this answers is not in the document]
TaoBi22: > you probably don't want to abort the whole build because of some arbitrary limit.

Agreed, it seems like a tricky thing to understand for a user who runs into it trying to get started with Arcilator. I think going without an iteration check makes sense personally.

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+               !subOp->hasTrait<OpTrait::HasRecursiveMemoryEffects>() &&
+               !mlir::isMemoryEffectFree(subOp))
+        nextSideEffect = &op;
+    });
+
+    // Determine how much the op can be moved.
+    OpAndOrder moveLimit;
+    if (auto ptr = getPointerReadByOp(&op)) {
+      // Don't move across writes to the same state/memory.
+      if (auto *write = nextWrite.lookup(ptr))
```
maerhart, line 158: This assumes no aliasing can happen. It's probably fine in the current arcilator pipeline, but we should maybe make that more explicit in the pass documentation.

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+  for (auto &op : llvm::make_early_inc_range(llvm::reverse(rootBlock))) {
+    // Assign an order to this op.
+    auto order = OpOrder{opOrder.size() + 1, 0};
+    opOrder[&op] = order;
+
+    // Analyze the side effects in the op.
+    op.walk([&](Operation *subOp) {
+      if (auto ptr = getPointerWrittenByOp(subOp))
+        nextWrite[ptr] = &op;
```
maerhart, line 147: This means that write operations are not marked as general side-effect ops and operations that contain reads may be moved over them. I don't think this is a problem in practice, though, because calls aren't moved anyway.

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+      continue;
+
+    // Only handle simple cases for now. (Same condition, no results, only
+    // single blocks, and both ifs either have or don't have an else region.)
+    if (ifOp.getCondition() != prevIfOp.getCondition())
+      continue;
+    if (ifOp.getNumResults() != 0 || prevIfOp.getNumResults() != 0)
+      continue;
+    if (!ifOp.getThenRegion().hasOneBlock() ||
+        !prevIfOp.getThenRegion().hasOneBlock())
```
maerhart, line 278: Isn't this guaranteed by the `SizedRegion<1>` constraint?

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+      continue;
+    if (ifOp.getElseRegion().empty() != prevIfOp.getElseRegion().empty())
+      continue;
+    if (!ifOp.getElseRegion().empty() &&
+        (!ifOp.getElseRegion().hasOneBlock() ||
+         !prevIfOp.getElseRegion().hasOneBlock()))
+      continue;
+
+    // Try to move ops in between the if ops above the previous if in order to
```
maerhart, line 287:
```suggestion
    // Try to move ops in between the 'if' ops above the previous 'if' in order to
```

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+      bool hasSideEffects = false;
+      writes.clear();
+      reads.clear();
+      prevIfOp.walk([&](Operation *op) {
+        if (auto ptr = getPointerWrittenByOp(op))
+          writes.insert(ptr);
+        else if (auto ptr = getPointerReadByOp(op))
+          reads.insert(ptr);
+        else if (!hasSideEffects &&
+                 !op->hasTrait<OpTrait::HasRecursiveMemoryEffects>() &&
```
maerhart, line 300: Same as above, why do we need the check for the trait if we already have `isMemoryEffectFree`?

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+                 !mlir::isMemoryEffectFree(op))
+          hasSideEffects = true;
+      });
+
+      // Check if it is legal to throw all ops over the previous if op, given
```
maerhart, line 305:
```suggestion
      // Check if it is legal to throw all ops over the previous 'if' op, given
```

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+        auto result = op.walk([&](Operation *subOp) {
+          if (auto ptr = getPointerWrittenByOp(subOp)) {
+            // We can't move writes over writes or reads of the same state.
+            if (writes.contains(ptr) || reads.contains(ptr))
+              return WalkResult::interrupt();
+          } else if (auto ptr = getPointerReadByOp(subOp)) {
+            // We can't move reads over writes to the same state.
+            if (writes.contains(ptr))
+              return WalkResult::interrupt();
+          } else if (!subOp->hasTrait<OpTrait::HasRecursiveMemoryEffects>() &&
```
maerhart, line 321: Same as above

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+/// Iteratively sink ops into block and move them closer to their uses, and
+/// merge adjacent `scf.if` operations.
+LogicalResult Sinker::run() {
+  LLVM_DEBUG(llvm::dbgs() << "Running on block in "
+                          << rootBlock.getParentOp()->getName() << "\n");
+  do {
+    anyChanges = false;
+    sinkOps();
+    mergeIfs();
+  } while (anyChanges);
```
maerhart, line 73: Nothing important, but the number of iterations here might make for another nice pass statistic 🙂

maerhart (approved):
Awesome work! 🎉

lib/Dialect/Arc/Transforms/MergeIfs.cpp
fabianschuiki: You are absolutely right! That simplifies things.

lib/Dialect/Arc/Transforms/MergeIfs.cpp
fabianschuiki: Yeah that's a great point. Maybe we should make that a feature of the dialect itself somewhere? As in, document that states and memories are assumed to never alias. Not sure how well that will work with a more struct-like memory layout though, where we'll need something like `arc.state_ptr %storage, @DataLayout::someField` instead of `arc.alloc_state`. Those storage-to-state-pointer ops probably will get duplicated and it'll be harder to know about the aliasing. Might need to hook into MLIR's alias analysis at that point.

I'll add something to the pass' description for now.

lib/Dialect/Arc/Transforms/MergeIfs.cpp
fabianschuiki: It is 😄

lib/Dialect/Arc/Transforms/MergeIfs.cpp
fabianschuiki: Haha I had that but removed it again 😅. Let me add it back in -- it's good to have around!

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,365 @@
...
+    // Assign an order to this op.
+    auto order = OpOrder{opOrder.size() + 1, 0};
+    opOrder[&op] = order;
+
+    // Analyze the side effects in the op.
+    op.walk([&](Operation *subOp) {
+      if (auto ptr = getPointerWrittenByOp(subOp))
+        nextWrite[ptr] = &op;
+      else if (!isa<StateReadOp, MemoryReadOp>(subOp) &&
+               !subOp->hasTrait<OpTrait::HasRecursiveMemoryEffects>() &&
```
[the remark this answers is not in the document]
fabianschuiki: My thinking here was that the outer `op.walk(...)` traverses nested ops, but the call to `isMemoryEffectFree` _also_ traverses nested ops. To avoid that, I only call `isMemoryEffectFree` on ops without `HasRecursiveMemoryEffects`, since we're already doing the recursion in `op.walk(...)`. This kind of assumes that an op with `HasRecursiveMemoryEffects` has no side-effects of its own… not sure how valid that assumption is 🤔

lib/Dialect/Arc/Transforms/MergeIfs.cpp
```diff
@@ -0,0 +1,367 @@
...
+  using MergeIfsPassBase::numOpsSunk;
+};
+
+/// A helper to perform the op sinking within a specific block.
+struct Sinker {
+  MergeIfsPass &pass;
+  Block &rootBlock;
+  bool anyChanges;
+
+  Sinker(MergeIfsPass &pass, Block &rootBlock)
```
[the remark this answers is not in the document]
fabianschuiki: That's a good point. The `Sinker` didn't end up carrying any real state. Might as well merge that into the pass itself!

lib/Dialect/Arc/Transforms/MergeIfs.cpp
maerhart: Yes, good point. That would be a nice property for the dialect, but not sure if it would restrict us too much in the struct case you mentioned. I guess we'll see.

Your storage-to-state-pointer duplication is more like what I thought of here. It could also be an `scf.if` returning a state or a call or any other op that could forward a state value in some way. MLIR's alias analysis should make this easy to support properly once we need it, though.

lib/Dialect/Arc/Transforms/MergeIfs.cpp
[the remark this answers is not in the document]
fabianschuiki: Turns out the `isMemoryEffectFree` function is pretty simple. I've created a new `hasSideEffects` function that does the necessary non-recursive check.

PR #7703, opened 2024-10-14
[Arc] Improve LowerState to never produce read-after-write conflicts

### LINKED ISSUE #6390: [Arc] Improve how memory writes are legalized
Memory write operations are currently legalized by just moving them after the last read from the same memory. This approach does not always work nicely because multiple memory reads and writes might already be grouped in a nested region (e.g., a `scf.if`). If the last read from memory happens outside of this nested region, but a write is inside that region, either the whole region has to be moved (possibly creating new conflicts due to the reads in the region also being moved) or the region has to be split up (currently not implemented).
Also, the implementation uses the following bit of code which leads to the pass running for ~17 sec (or 25% of the entire arcilator pipeline) on a dual-core MegaBoom config. Improving this would improve the runtime to only ~0.6 sec. Getting rid of this pattern would require first collecting all the op movements that have to be performed and then applying them all in the end. This can be difficult since the movement of read operations in the same nested region would also have to be considered.

https://github.com/llvm/circt/blob/d7088aa368911fd1b47770a6c3f7ff9bfdc1b2c9/lib/Dialect/Arc/Transforms/LegalizeStateUpdate.cpp#L521-L524

Alternatively, we could cache the memory write transactions somewhere (e.g., some space in the simulation state) until all reads are performed and only apply them afterward.

fabianschuiki:
This is a complete rewrite of the `LowerState` pass that makes the `LegalizeStateUpdate` pass obsolete.

The old implementation of `LowerState` produces `arc.model`s that still contain read-after-write conflicts. This primarily happens because the pass simply emits `arc.state_write` operations that write updated values to simulation memory for each `arc.state`, and any user of `arc.state` would use an `arc.state_read` operation to retrieve the original value of the state before any writes occurred. Memories are similar. The Arc dialect considers `arc.state_write` and `arc.memory_write` operations to be _deferred_ writes until the `LegalizeStateUpdate` pass runs, at which point they become _immediate_ since the legalization inserts the necessary temporaries to resolve the read-after-write conflicts.

The previous implementation would also not handle state-to-output and state-to-side-effecting-op propagation paths correctly. When a model's eval function is called, registers are updated to their new value, and any outputs that combinatorially depend on those new values should also immediately update. Similarly, operations such as asserts or debug trackers should observe new values for states immediately after they have been written. However, since all writes are considered deferred, there is no way for `LowerState` to produce a mixture of operations that depend on a register's _old_ state (because they are used to compute a register's new state), and on a _new_ state because they are combinatorially derived values.

This new implementation of `LowerState` completely avoids read-after-write conflicts. It does this by changing the way modules are lowered in two ways:

**Phases:** The pass tracks in which _phase_ of the simulation lifecycle a value is needed and allows for operations to have different lowerings in different phases. An `arc.state` operation for example requires its inputs, enable, and reset to be computed based on the _old_ value they had, i.e. the value the end of the previous call to the model's eval function. The clock however has to be computed based on the _new_ value it has in the current call to eval. Therefore, the ops defining the inputs, enable, and reset are lowered in the _old_ phase, while the ops defining the clock are lowered in the _new_ phase. The `arc.state` op lowering will then write its _new_ value to simulation storage.

This phase tracking allows registers to be used as the clock for other registers: since the clocks require _new_ values, registers serving as clock to others are lowered first, such that the dependent registers can immediately react to the updated clock. It also allows for module outputs and side-effecting ops based on `arc.state`s to be scheduled after the states have been updated, since they depend on the state's _new_ value.

The pass also covers the situation where an operation depends on a module input and a state, and feeds into a module output as well as another state. In this situation that operation has to be lowered twice: once for the _old_ phase to serve as input to the subsequent state, and once for the _new_ phase to compute the new module output.

In addition to the _old_ and _new_ phases representing the previous and current call to eval, the pass also models an _initial_ and _final_ phase. These are used for `seq.initial` and `llhd.final` ops, and in order to compute the initial values for states. If an `arc.state` op has an initial value operand it is lowered in the _initial_ phase. Similarly for the ops in `llhd.final`. The pass places all ops lowered in the initial and final phases into corresponding `arc.initial` and `arc.final` ops. At a later point we may want to generate the `*_initial`, `*_eval`, and `*_final` functions directly.

**No more clock trees:** The new implementation also no longer generates `arc.clock_tree` and `arc.passthrough` operations. These were a holdover from the early days of the Arc dialect, where no eval function would be generated. Instead, the user was required to directly call clock functions. This was never able to model clocks changing at the exact same moment, or having clocks derived from registers and other combinatorial operations. Since Arc has since switched to generating an eval function that can accurately interleave the effects of different clocks, grouping ops by clock tree is no longer needed. In fact, removing the clock tree ops allows for the pass to more efficiently interleave the operations from different clock domains.

The Rocket core in the circt/arc-tests repository still works with this new implementation of LowerState. In combination with the MergeIfs pass the performance stays the same. (Actually 10% speedup on my workstation.)

I have renamed the implementation and test files to make the git diffs easier to read. The names will be changed back in a follow-up commit.

Fixes #6390.


integration_test/arcilator/JIT/reg.mlir
```diff
@@ -41,6 +41,7 @@ func.func @main() {
   %step = arith.constant 1 : index
 
   arc.sim.instantiate @counter as %model {
+    arc.sim.step %model : !arc.sim.instance<@counter>
```
uenoku, line 44: Is this change required to run initialization?

fabianschuiki: Yeah. In its current form, a call to the `initial` function only executes `seq.initial` blocks and initializes the storage for states. It does not propagate those changes to outputs and side-effecting ops though. I wasn't sure if the `initial` function should also call `eval` once after initialization (potentially triggering first state updates and other things to run immediately), or if the user might want more fine grained control over when the first eval happens. WDYT?

uenoku: Yeah I see that makes sense. I agree that we want to avoid implicitly calling `eval` from `initial`

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+
+  /// A mapping from unlowered clocks to a value indicating a posedge. This is
+  /// used to not create an excessive number of posedge detectors.
+  DenseMap<Value, Value> loweredPosedges;
+  /// The previous enable and the value it was lowered to. This is used to reuse
+  /// previous if ops for the same enable value.
+  std::pair<Value, Value> prevEnable;
+  /// The previous reset and the value it was lowered to. This is used to reuse
+  /// previous uf ops for the same reset value.
+  std::pair<Value, Value> prevReset;
```
maerhart, line 169: If there is more than one reset in a module and we process operations in an order where op1 uses resetA, op2 uses resetB, op3 uses resetA, etc., it would clone the reset fan-in each time?

test/Dialect/Arc/lower-state.mlir
```diff
@@ -1,296 +1,611 @@
...
+  // CHECK:     [[MASK_INV:%.+]] = comb.xor bin [[TMP]]#3, [[ALL_ONES]]
+  // CHECK:     [[DATA_OLD:%.+]] = arc.memory_read [[MEM]][[[TMP]]#0]
+  // CHECK:     [[MASKED_OLD:%.+]] = comb.and bin [[MASK_INV]], [[DATA_OLD]]
+  // CHECK:     [[MASKED_NEW:%.+]] = comb.and bin [[TMP]]#3, [[TMP]]#1
+  // CHECK:     [[DATA_NEW:%.+]] = comb.or bin [[MASKED_OLD]], [[MASKED_NEW]]
+  // CHECK:     arc.memory_write [[MEM]][[[TMP]]#0], [[DATA_NEW]]
+  // CHECK:   }
+  // CHECK: }
+  %mem = arc.memory <4 x i42, i2>
+  arc.memory_write_port %mem, @IdI2AndI42AndI1AndI42Arc(%a, %b, %en, %mask) clock %clock enable mask latency 1 : <4 x i42, i2>, i2, i42, i1, i42
```
maerhart, line 356: A second masked write to the same location, but inverted mask could be interesting to check the correct 'old' value is read.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+    mapping.map(operand, lowered);
+  }
+  if (initial)
+    return success();
+
+  // Handle the simple case where the final op contains only one block, which we
+  // can inline directly.
+  if (op.getBody().hasOneBlock()) {
+    for (auto &bodyOp : op.getBody().front().without_terminator())
+      module.finalBuilder.clone(bodyOp, mapping);
```
maerhart, line 954: If we make this pass operate on OpInterfaces in the future, we probably also want to call `lower` on these nested operations.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+  if (auto stateOp = dyn_cast<StateOp>(op))
+    return lowerValue(stateOp, result, phase);
+  if (auto dpiOp = dyn_cast<sim::DPICallOp>(op); dpiOp && dpiOp.getClock())
+    return lowerValue(dpiOp, result, phase);
+  if (auto readOp = dyn_cast<MemoryReadPortOp>(op))
+    return lowerValue(readOp, result, phase);
+  if (auto initialOp = dyn_cast<seq::InitialOp>(op))
+    return lowerValue(initialOp, result, phase);
+  if (auto castOp = dyn_cast<seq::FromImmutableOp>(op))
+    return lowerValue(castOp, result, phase);
```
maerhart, line 1025: Nit: maybe use a TypeSwitch?

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+  // If we want to read the old value, no writes must have been lowered yet.
+  if (phase == Phase::Old)
+    assert(!module.loweredOps.contains({op, Phase::New}) &&
+           "need old value but new value already written");
+
+  auto state = module.getAllocatedState(result);
+  return module.getBuilder(phase).create<StateReadOp>(result.getLoc(), state);
+}
+
+/// Handle uses of a state. This creates a `arc.state_read` op to read from the
```
maerhart, line 1069:
```suggestion
/// Handle uses of a DPI call. This creates an `arc.state_read` op to read from the
```

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+/// Handle instance outputs. They behave essentially like a top-level module
+/// input, and read the same in all phases.
+Value OpLowering::lowerValue(InstanceOp op, OpResult result, Phase phase) {
+  if (initial)
+    return {};
+  auto state = module.getAllocatedState(result);
+  return module.getBuilder(phase).create<StateReadOp>(result.getLoc(), state);
+}
+
+/// Handle uses of a state. This creates a `arc.state_read` op to read from the
```
maerhart, line 1047:
```suggestion
/// Handle uses of a state. This creates an `arc.state_read` op to read from the
```

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+
+  return success();
+}
+
+/// Lower a memory and its read and write ports to corresponding
+/// `arc.memory_write` operations. Reads are also executed at this point and
+/// stored in `loweredValues` for later operations to pick up.
+LogicalResult OpLowering::lower(MemoryOp op) {
+  if (phase != Phase::New)
+    return op.emitOpError() << "cannot be lowered in the " << phase << " phase";
```
maerhart, line 675: Can these errors actually be triggered from some input IR or only if this function is used incorrectly? If the latter, we could make it an assert. If the earlier, maybe some error test?

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+      opLowering.initial = false;
+    }
+
+    // Push operands onto the worklist.
+    if (!opLowering.pending.empty()) {
+      auto [defOp, phase] = opLowering.pending.pop_back_val();
+      if (loweredOps.contains({defOp, phase}))
+        continue;
+      if (!opsSeen.insert({defOp, phase}).second) {
+        defOp->emitOpError("is on a combinational loop");
```
maerhart, line 286: An error test for that might be nice.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+LogicalResult OpLowering::lower() {
+  return TypeSwitch<Operation *, LogicalResult>(op)
+      // Operations with special lowering.
+      .Case<StateOp, sim::DPICallOp, MemoryOp, TapOp, InstanceOp, hw::OutputOp,
+            seq::InitialOp, llhd::FinalOp>([&](auto op) { return lower(op); })
+
+      // Operations that should be skipped entirely and never land on the
+      // worklist to be lowered.
+      .Case<MemoryWritePortOp, MemoryReadPortOp>([&](auto op) {
+        op.emitOpError() << "is handled by memory op and must be skipped";
```
maerhart, line 423: Can this actually be triggered by some input or should this be an assertion?

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+
+  return alloc;
+}
+
+/// Allocate the necessary storage, reads, writes, and comparisons to detect a
+/// rising edge on a clock value.
+Value ModuleLowering::detectPosedge(Value clock) {
+  auto loc = clock.getLoc();
+  if (isa<seq::ClockType>(clock.getType()))
+    clock = builder.createOrFold<seq::FromClockOp>(loc, clock);
```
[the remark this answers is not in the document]
fabianschuiki: I guess very rarely when the clock is derived from an `i1` through `seq.to_clock` :grin:. Not very common I assume.

lib/Dialect/Arc/ArcTypes.cpp
```diff
@@ -21,11 +21,17 @@ using namespace mlir;
...
+  if (llvm::isa<seq::ClockType>(getType()))
+    return 1;
+  return hw::getBitWidth(getType());
+}
 
 LogicalResult
 StateType::verify(llvm::function_ref<InFlightDiagnostic()> emitError,
                   Type innerType) {
+  if (llvm::isa<seq::ClockType>(innerType))
+    return success();
```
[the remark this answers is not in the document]
fabianschuiki: Not in `hw::getBitWidth` though :cry:

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+  if (auto instOp = dyn_cast<InstanceOp>(result.getOwner()))
+    alloc->setAttr(
+        "name", builder.getStringAttr(
+                    instOp.getInstanceName() + "/" +
+                    instOp.getResultName(result.getResultNumber()).getValue()));
+
+  // HACK: If the result comes from an op that has a "names" attribute, use that
+  // as a name for the allocation. This should no longer be necessary once we
+  // properly support the Debug dialect.
+  if (isa<StateOp, sim::DPICallOp>(result.getOwner()))
```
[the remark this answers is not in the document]
fabianschuiki: The `TapOp` creates an `AllocStateOp` directly further down, where it also sets the name properly. This code up here is only for state-like ops that have to create an allocation when their def or a use is encountered for the first time.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+
+/// Called for all operations for which there is no special lowering. Simply
+/// clones the operation.
+LogicalResult OpLowering::lowerDefault() {
+  // Make sure that all operand values are lowered first.
+  IRMapping mapping;
+  auto anyFailed = false;
+  op->walk([&](Operation *nestedOp) {
+    for (auto operand : nestedOp->getOperands()) {
+      if (op->isAncestor(operand.getParentBlock()->getParentOp()))
```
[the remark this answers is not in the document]
fabianschuiki: Yes. We'll need to figure out how exactly nested stateful ops can be handled, for example when there is an `arc.clock_domain` op present, or something like an `arc.vectorize` or `arc.task` or `arc.group`. I'm not entirely sure how that will work, since the container op's semantics likely affects how the contained ops are lowered.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+      inputs.push_back(lowerValue(operand, phase));
+    if (initial)
+      return success();
+    if (llvm::is_contained(inputs, Value{}))
+      return failure();
+    if (op.getEnable())
+      return op.emitOpError() << "without clock cannot have an enable";
+
+    // Lower the op to a regular function call.
+    auto callOp = module.getBuilder(phase).create<func::CallOp>(
```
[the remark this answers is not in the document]
fabianschuiki: Hmmm I think you're right. An `arc.cached` operation would be useful in general, to memoize computation results :grin:. We'd have to figure out how to compare something for equality though. It's easy for integers, but I don't think we have equality ops for clocks, structs, arrays, and some of the other types. We might have to create an `arc.eq` op that can check arbitrary values for equality. And then also go and fix up the other dialects to provide equality checks on their types.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+    // Make sure that all operands have been lowered.
+    SmallVector<Value> inputs;
+    for (auto operand : op.getInputs())
+      inputs.push_back(lowerValue(operand, phase));
+    if (initial)
+      return success();
+    if (llvm::is_contained(inputs, Value{}))
+      return failure();
+    if (op.getEnable())
+      return op.emitOpError() << "without clock cannot have an enable";
```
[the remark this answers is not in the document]
fabianschuiki: I don't remember where we landed on this. @uenoku Does the DPI op just say the outputs are undefined if it has an enable but is unclocked?

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
```diff
@@ -0,0 +1,1206 @@
...
+                                             op.getArc(), inputs)
+                             .getResults();
+                       });
+}
+
+/// Lower a state to a corresponding storage allocation and write of the state's
+/// new value to it. This function uses the `Old` phase to get the values at the
+/// state input before the current update, and then uses them to compute the
+/// `New` value.
+LogicalResult OpLowering::lower(sim::DPICallOp op) {
```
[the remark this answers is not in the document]
fabianschuiki: Yeah that would definitely be nice. I remember that there was some discussion in MLIR about having op interfaces specify lowerings, and that MLIR was very adamantly advocating _against_ defining lowerings in op interfaces. Don't remember where exactly that was -- maybe @darthscsi still knows.

An alternative might be to make `arc.state` ops have a region, pre-lower clocked DPI calls into unclocked calls within such a state up. More composition feels like a good thing to try.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
fabianschuiki: This just opportunistically remembers the last enable and reset that was lowered, in order to get a chance to fuse `scf.if` ops on the fly. You could delete all this and would get correct output as well, just with more if ops. The "lowering" here is also pretty trivial: it's mostly just `arc.state_read` operations that are reused. There is no more copying of fan-ins happening.

test/Dialect/Arc/lower-state.mlir
fabianschuiki: Yeah we should definitely have that as an integration test somewhere. I don't think the correct behavior would be apparent here in the lowering test :thinking:

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
fabianschuiki: Yeah this needs some careful thought to figure out how exactly ops with nested regions would be lowered.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
fabianschuiki: Hmmm I'm not sure it would make this more readable. It generates the exact same code in the background (unfortunately it doesn't optimize anything really). The early returns are annoying to do in type switches :cry:.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
fabianschuiki: These should never be triggered by user input. This was just a sanity check while developing the pass, where it was useful to get a meaningful error message. I can definitely turn these into asserts.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
fabianschuiki: Great idea! Added one :+1:

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
fabianschuiki: This should definitely be an assertion.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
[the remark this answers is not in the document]
maerhart: For structs and arrays, `hw.bitcast` to an integer could work. If we factor out op specific things into an OpInterface, we could also have some type interface for value equality. 🤔

lib/Dialect/Arc/ArcTypes.cpp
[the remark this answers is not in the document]
maerhart: Ah right, you're calling `hw::getBitWidth`. Why not `StateType::getBitWidth` though?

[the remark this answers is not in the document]
fabianschuiki: Because that is not available at this point :cry: (it takes the type from the `this` parameter instead of a function argument). I should just factor this out into a static helper.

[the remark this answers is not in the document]
fabianschuiki: Let me fix that post-merge.

lib/Dialect/Arc/Transforms/LowerStateRewrite.cpp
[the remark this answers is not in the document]
maerhart: Interesting. Upstream MLIR heavily uses OpInterfaces to specify custom op behavior for generic optimization and lowering passes.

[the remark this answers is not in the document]
fabianschuiki: Huh, in that case we should probably be free to follow that pattern as well. :+1:

PR #7704, opened 2024-10-14
[Arc] Remove obsolete arc.clock_tree and arc.passthrough ops

fabianschuiki:
The new LowerState pass does not produce `arc.clock_tree` and `arc.passthrough` ops anymore. Remove them from the dialect entirely.


TaoBi22 (approved):
LGTM (once it builds :wink:)!

lib/Dialect/Arc/Transforms/LowerClocksToFuncs.cpp
```diff
@@ -197,16 +166,6 @@ LogicalResult LowerClocksToFuncsPass::lowerClock(Operation *clockOp,
...
 
-  if (isa<InitialOp>(clockOp) && hasPassthroughOp) {
-    // Call PassThroughOp after init
-    builder.setInsertionPoint(funcOp.getBlocks().front().getTerminator());
-    funcName.clear();
-    funcName.append(modelOp.getName());
-    funcName.append("_passthrough");
-    builder.create<func::CallOp>(clockOp->getLoc(), funcName, TypeRange{},
-                                 ValueRange{funcOp.getBody().getArgument(0)});
-  }
```
maerhart, line 208: If the initial does affect an output port value directly, do we not need to guarantee that this change is propagated to the output port?

fabianschuiki: Yeah that's a good question. Since passthrough is now gone, this boils down to whether the initial function should also call the eval function to propagate the effects of the initial assignments to outputs and side-effecting ops. I'm trying to think whether there is a use case where the user wants to run `initial`, but then do some other tweaking (or maybe dump a first timestep to VCD?) before calling the first `eval` and potentially triggering output or state changes.

@uenoku Has observed that some simulators have fairly intricate bring-up sequences. Generally it seems like there's an initial value assignment to variables, often X, then an execution of all variable initializers, then executing all `initial` blocks, and then triggering any `posedge` and the like based on those variables. We might need to do something similar at some point? Maybe some form of tiered `arc.initial` to have containers for the different rounds of initialization?

fzi-hielscher: That is what I've been desperately trying to avoid so far. But as the SV LRM so nonchalantly states on initialization at declaration:

> This may require a special pre-initial pass at run time.

☹️ 

Considering that commercial simulators also tend to take their freedoms here [citation needed], I was hoping we could avoid that madness. If feel when we have to debate whether a `posedge` has occurred when a `bit` is default initialized to 0, then initialized to 1 at declaration, and then initialized back to 0 in an `initial` process, we have already lost.

fabianschuiki: Yeah I totally agree. We should probably imitate the basic fixing of Verilog semantics that simulators do with their observable X-to-initial-value transitions, because otherwise Verilog's `always @(...)` combinational processes and async resets are virtually unusable. But we should do as little as possible/necessary.

fzi-hielscher: For the triggered processes over in #7676 I currently require a tieoff constant for all processes, which also servers as pre-initial value. I've thought about changing this to take `!seq.immutable` values.  Considering that neither me nor @uenoku have intended it that way, I think this would mesh surprisingly well. It would de facto push `seq.initial` into the pre-initial phase and we wouldn't be able to use non-`!seq.immutable` values in that phase. But I guess this is what we already arrived at in #7638 ?!?

But to go back to Martin's original question here: I think it is a fair choice to not force propagation after initialization. In the current implementation I've ended up introducing an inconsistency between models without an initializer (which won't call `passthrough` at startup) vs. the same model with an empty initializer  (which _will_ call `passthrough`). It's a good thing to avoid that.

PR #7705, opened 2024-10-14
[arcilator] Add clock divider integration test

fabianschuiki:
Add a test to check that arcilator can simulate a simple clock divider. This exercises a corner case of arcilator's simulation model scheduling, where a state updating its value can trigger other states and module outptus to update their values. In this case, a cascade of clock edges is generated by feeding one state's output into the clock input of the next state.


fzi-hielscher (commented):
This needs #7703 to work correctly, doesn't it?

Independently of that, it is currently blasting through the C calling conventions, giving me this output locally:
```
0 0 2191024128 1808528296
0 0 2191024128 1808528296
0 0 2191024128 1808528296
[...]
```
The changes below should fix that.

fabianschuiki:
Uh oh, damn calling conventions :see_no_evil:. Thanks for the fixes @fzi-hielscher :partying_face:.

integration_test/arcilator/JIT/clock-divider.mlir
```diff
@@ -0,0 +1,64 @@
...
+
+      %0 = arc.sim.get_port %dut, "div1" : !seq.clock, !arc.sim.instance<@ClockDivBy4>
+      %1 = arc.sim.get_port %dut, "div2" : !seq.clock, !arc.sim.instance<@ClockDivBy4>
+      %2 = arc.sim.get_port %dut, "div4" : !seq.clock, !arc.sim.instance<@ClockDivBy4>
+
+      %3 = llvm.mlir.addressof @string : !llvm.ptr
+      %4 = seq.from_clock %0
+      %5 = seq.from_clock %1
+      %6 = seq.from_clock %2
+      func.call @printf(%3, %i0, %4, %5, %6) : (!llvm.ptr, i1, i1, i1, i1) -> ()
```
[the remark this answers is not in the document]
fabianschuiki: Oh interesting, thanks for providing that link with further details. This smells a lot like "we couldn't make printf safe to use for the average user, so let's fix the language instead" :grin:

uenoku (approved):
Cool, it's impressive that ClockDiv just works!

integration_test/arcilator/JIT/clock-divider.mlir
```diff
@@ -0,0 +1,68 @@
...
+  %q1 = arc.state @Not(%q1) clock %0 latency 1 {names = ["q1"]} : (i1) -> i1
+  %1 = seq.to_clock %q1
+  hw.output %clock, %0, %1 : !seq.clock, !seq.clock, !seq.clock
+}
+
+func.func @entry() {
+  %true = hw.constant true
+  %false = hw.constant false
+  %clk1 = seq.to_clock %true
+  %clk0 = seq.to_clock %false
```
maerhart, line 33: Unused and could thus be removed?

maerhart (approved):
Really nice that this just works!

integration_test/arcilator/JIT/clock-divider.mlir
fabianschuiki: Good point

PR #7706, opened 2024-10-14
[FIRRTL][ProbesToSignals] RWProbe support

dtzSiFive:
* Define type conversion mapping rwprobe<T> -> passive(T)
* Forceable support (passive read of data result)
* RWProbeOp support (materialize target, to passive)
* Reject {force,release}{,_initial}

Add some basic tests.


prithayan (approved):
Thanks for adding the RWProbe support, LGTM.

PR #7707, opened 2024-10-15
[Verif] Add LowerFormalToHW pass

leonardt:
* Converts `verif.formal` to `hw.module`
* Lifts `verif.symbolic_value` to be an input to the module

### DIFF
--- a/include/circt/Dialect/Verif/Passes.td
+++ b/include/circt/Dialect/Verif/Passes.td
@@ -43,14 +43,13 @@ def VerifyClockedAssertLike : Pass<"verify-clocked-assert-like", "hw::HWModuleOp
 }
 
 /**
- * Converts verif.formal ops into hw.module to prepare for export verilog.
+ * Converts verif.formal ops into hw.module.
  */
-def PrepareVerifForExportVerilog : 
-    Pass<"prepare-verif-for-export-verilog", "mlir::ModuleOp"> {
-  let summary = "Prepare verif for verilog export by lowering to hw dialect.";
+def LowerFormalToHW : 
+    Pass<"lower-formal-to-hw", "mlir::ModuleOp"> {
+  let summary = "Lower verif.formal ops to hw.module ops.";
   let description = [{
-    Converts verif.formal ops into hw.module to reuse the export verilog module
-    logic.
+    Converts verif.formal ops into hw.module ops.
   }];
 }
 
--- a/include/circt/Dialect/Verif/Passes.td
+++ b/include/circt/Dialect/Verif/Passes.td
@@ -45,8 +45,7 @@ def VerifyClockedAssertLike : Pass<"verify-clocked-assert-like", "hw::HWModuleOp
 /**
  * Converts verif.formal ops into hw.module.
  */
-def LowerFormalToHW : 
-    Pass<"lower-formal-to-hw", "mlir::ModuleOp"> {
+def LowerFormalToHW : Pass<"lower-formal-to-hw", "mlir::ModuleOp"> {
   let summary = "Lower verif.formal ops to hw.module ops.";
   let description = [{
     Converts verif.formal ops into hw.module ops.
--- a/include/circt/Dialect/Verif/Passes.td
+++ b/include/circt/Dialect/Verif/Passes.td
@@ -42,4 +42,16 @@ def VerifyClockedAssertLike : Pass<"verify-clocked-assert-like", "hw::HWModuleOp
   let constructor = "circt::verif::createVerifyClockedAssertLikePass()";
 }
 
+/**
+ * Converts verif.formal ops into hw.module to prepare for export verilog.
+ */
+def PrepareVerifForExportVerilog : 
+    Pass<"prepare-verif-for-export-verilog", "mlir::ModuleOp"> {
+  let summary = "Prepare verif for verilog export by lowering to hw dialect.";
+  let description = [{
+    Converts verif.formal ops into hw.module to reuse the export verilog module
+    logic.
+  }];
+}
+
 #endif // CIRCT_DIALECT_VERIF_PASSES_TD
--- a/lib/Dialect/Verif/Transforms/LowerFormalToHW.cpp
+++ b/lib/Dialect/Verif/Transforms/LowerFormalToHW.cpp
@@ -1,4 +1,4 @@
-//===- PrepareVerifForExportVerilog.cpp - Formal Preparations --*- C++ -*--===//
+//===- LowerFormalToHW.cpp - Formal Preparations --*- C++ -*---------------===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -17,7 +17,7 @@ using namespace circt;
 
 namespace circt {
 namespace verif {
-#define GEN_PASS_DEF_PREPAREVERIFFOREXPORTVERILOG
+#define GEN_PASS_DEF_LOWERFORMALTOHW
 #include "circt/Dialect/Verif/Passes.h.inc"
 } // namespace verif
 } // namespace circt
@@ -26,9 +26,8 @@ using namespace mlir;
 using namespace verif;
 
 namespace {
-struct PrepareVerifForExportVerilog
-    : circt::verif::impl::PrepareVerifForExportVerilogBase<
-          PrepareVerifForExportVerilog> {
+struct LowerFormalToHW
+    : circt::verif::impl::LowerFormalToHWBase<LowerFormalToHW> {
   void runOnOperation() override;
 };
 
@@ -68,7 +67,7 @@ struct FormalOpConversionPattern : public OpConversionPattern<verif::FormalOp> {
 };
 } // namespace
 
-void PrepareVerifForExportVerilog::runOnOperation() {
+void LowerFormalToHW::runOnOperation() {
   auto &context = getContext();
   mlir::ConversionTarget target(context);
   target.addLegalDialect<hw::HWDialect, verif::VerifDialect>();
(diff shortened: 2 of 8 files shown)


lib/Dialect/Verif/Transforms/PrepareVerifForExportVerilog.cpp
```diff
@@ -0,0 +1,91 @@
+//===- PrepareVerifForExportVerilog.cpp - Formal Preparations --------------*- C++ -*--===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//===----------------------------------------------------------------------===//
+//
+// Prepare verif dialect for the verillog export.
```
dtzSiFive, line 8: Nut:
```suggestion
// Prepare verif dialect for the verilog export.
```

lib/Dialect/Verif/Transforms/PrepareVerifForExportVerilog.cpp
```diff
@@ -0,0 +1,91 @@
...
+    // arguments
+    size_t i = 0;
+    for (auto& innerOp : op.getBody().front().getOperations()) {
+      if (dyn_cast<verif::SymbolicValueOp>(innerOp)) {
+        rewriter.replaceAllUsesWith(innerOp.getResult(0), moduleOp.getArgumentForInput(i));
+        i++;
+      } else {
+        rewriter.clone(innerOp);
+      }
+    }
```
fabianschuiki, line 68: I wonder if we could avoid cloning the operations here, since we are going to entirely replace the `verif.formal` op with an `hw.module`. For example, you could try to use the `splice()` function on `Block` to take the operations in the `FormalOp` and insert them into the newly-created module. Something like the following might work:
```c++
Block &moduleBlock = moduleOp.getBody().front();
Block &formalBlock = op.getBody().front();
moduleBlock.getOperations().splice(
  moduleBlock.begin(),
  formalBlock.getOperations()
);
```
After that the formal op's body block should be empty. You'd still have to iterate over the body and delete all the symbolic value ops though.

lib/Dialect/Verif/Transforms/PrepareVerifForExportVerilog.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+struct FormalOpConversionPattern : public OpConversionPattern<verif::FormalOp> {
+  using OpConversionPattern<FormalOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(FormalOp op, OpAdaptor operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Create the ports for all the symbolic values
+    SmallVector<hw::PortInfo> ports;
+    for (auto sym : op.getBody().front().getOps<verif::SymbolicValueOp>()) {
```
uenoku, line 45: Is it guaranteed that verif::SymbolicValueOp appears in the top-level?

uenoku (commented):
Generally LGTM! A few comments:

1. These modules generated from FormalOp will be emitted to default directly (design) so it may need to find a way to emit these to other directory in the future. (probably having a output_file attribute or integrate with Emit dialect would be necessary). 
2. If the pass is specific to ExportVerilog, it may be suitable to put the logic under Conversion/ExportVerilog for consistency.

lib/Dialect/Verif/Transforms/PrepareVerifForExportVerilog.cpp
maerhart: I think `splice` should not be used in conversion patterns. These kinds of transformations need to go through the rewriter. I suggest looking at the various inline methods the rewrite provides instead.

include/circt/Dialect/Verif/Passes.td
```diff
@@ -42,4 +42,17 @@ def VerifyClockedAssertLike : Pass<"verify-clocked-assert-like", "hw::HWModuleOp
...
+ * Converts verif.formal ops into hw.module to prepare for export verilog.
+ */
+def PrepareVerifForExportVerilog : 
+    Pass<"prepare-verif-for-export-verilog", "mlir::ModuleOp"> {
+  let summary = "Prepare verif for verilog export by lowering to hw dialect.";
+  let description = [{
+    Converts verif.formal ops into hw.module to reuse the export verilog module
+    logic.
+  }];
+  let constructor = "circt::verif::createPrepareVerifForExportVerilogPass()";
```
maerhart, line 55: Consider dropping this line. TableGen will auto-gen the constructor for you then.

lib/Dialect/Verif/Transforms/PrepareVerifForExportVerilog.cpp
maerhart: It is because it has the `HasParent` trait.

lib/Dialect/Verif/Transforms/PrepareVerifForExportVerilog.cpp
fabianschuiki: That's an excellent point. 👍

include/circt/Dialect/Verif/Passes.td
fabianschuiki: You'll want to change the def name to `PrepareVerifForExportVerilogPass` as well, otherwise the function created will be called `createPrepareVerifForExportVerilog()`. 😁

fabianschuiki:
> @uenoku 
> 2. If the pass is specific to ExportVerilog, it may be suitable to put the logic under Conversion/ExportVerilog for consistency.

I think it's more of a general verif unit tests to `hw.module` conversion. Not ExportVerilog specific yet 🤔.

lib/Dialect/Verif/Transforms/PrepareVerifForExportVerilog.cpp
leonardt: Updated it to use inlineBlockBefore

uenoku:
> I think it's more of a general verif unit tests to hw.module conversion. Not ExportVerilog specific yet 🤔.

Oh I thought this is ExportVerilog specific because of the pass name `PrepareVerifForExportVerilog`. If the goal is to provide a lowering from Verif formal to HW, maybe simply change the name to `LowerFormalToHW` or something?

lib/Dialect/Verif/Transforms/LowerFormalToHW.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto moduleOp =
+        rewriter.create<hw::HWModuleOp>(op.getLoc(), op.getNameAttr(), ports);
+
+    rewriter.inlineBlockBefore(&op.getBody().front(),
+                               &moduleOp.getBodyBlock()->front(),
+                               op.getBody().getArguments());
+
+    // Replace symbolic values with module arguments
+    size_t i = 0;
+    for (auto symOp : moduleOp.getBodyBlock()->getOps<SymbolicValueOp>()) {
```
uenoku, line 58: Probably we need to use make_early_inc_range to avoid UAF for `symOp`.

lib/Dialect/Verif/Transforms/LowerFormalToHW.cpp
```diff
@@ -0,0 +1,82 @@
+//===- LowerFormalToHW.cpp - Formal Preparations --*- C++ -*---------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//===----------------------------------------------------------------------===//
+//
+// Prepare verif dialect for the verilog export.
```
uenoku, line 8: Maybe obsolete comment?

uenoku (approved):
LGTM！

fabianschuiki (approved):
Really cool! 🥳

test/Dialect/Verif/lower-formal-to-hw.mlir
```diff
@@ -0,0 +1,11 @@
...
+
+hw.module @Foo(in %bar : i32, in %baz : i16) {}
+
+// CHECK-LABEL: hw.module @FormalTop(in %symbolic_value_0 : i32, in %symbolic_value_1 : i16)
+verif.formal @FormalTop {
+  %0 = verif.symbolic_value : i32
+  %1 = verif.symbolic_value : i16
+  // CHECK-NEXT: hw.instance "foo" @Foo(bar: %symbolic_value_0: i32, baz: %symbolic_value_1: i16)
+  hw.instance "foo" @Foo(bar: %0: i32, baz: %1: i16) -> ()
+}
```
fabianschuiki, line 11: Can you try adding a `comb` and `seq` op in here somewhere as well? I think the dialect conversion framework might fail in that case because the comb/seq dialects aren't marked as legal.

leonardt: Yea, are there any other dialects besides comb and seq we want to include?

fabianschuiki: Hmmm, I'm wondering if we can get around listing the dialects explicitly. Since you just have a rewrite pattern that you want to apply exhaustively wherever it can, you could use the greedy pattern rewriter instead of the dialect conversion. That doesn't try to be clever about which ops are legal or illegal; it just applies the pattern everywhere.

This should work:
- Make your pattern inherit from `OpRewritePattern`. (You probably have to move the `make_early_inc_range` down to the loop where you actually call `eraseOp`, because this now applies more eagerly than with dialect conversion.)
- Use `applyPatternsAndFoldGreedily(getOperation(), std::move(patterns))` instead of the dialect conversion framework. (This needs `mlir/Transforms/GreedyPatternRewriteDriver.h`.)
- That should allow you to get rid of the conversion target entirely.

The greedy rewriter should then just apply your pattern until no more `verif.formal` ops are left.

dobios:
I'm curious as to what this particular pass is for? Is the goal to emit verilog testbenches? That would be neat

fabianschuiki:
@dobios Yes exactly! The idea was to turn `verif.formal` ops into the kind of HW module that most of the commercial formal tools and SymbiYosys could use as top-level for a formal proof :smiley:

PR #7709, opened 2024-10-15
[FIRRTL] Don't force non-local trackers in Dedup.

mikeurbach:
The previous behavior for local trackers would be to clone the tracker into duplicate copies for each unique instance. This will fail in LowerClasses, because one path op that used to refer to a single entity as a local reference would now refer to every hierarchical instance of that entity. One path can't refer to multiple things, so if the user specified a local reference, allow that to pass through without expanding it into every hierarchical instance.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -1233,7 +1233,8 @@ private:
   void copyAnnotations(FModuleLike toModule, AnnoTarget to,
                        FModuleLike fromModule, AnnotationSet annos,
                        SmallVectorImpl<Annotation> &newAnnotations,
-                       SmallPtrSetImpl<Attribute> &dontTouches) {
+                       SmallPtrSetImpl<Attribute> &dontTouches,
+                       SmallPtrSetImpl<Attribute> &localTrackerIds) {
     for (auto anno : annos) {
       if (anno.isClass(dontTouchAnnoClass)) {
         // Remove the nonlocal field of the annotation if it has one, since this
@@ -1254,7 +1255,10 @@ private:
       // If the annotation is a local tracker, don't make it non-local. This
       // allows trackers that refer to all instances when the user wants that.
       if (anno.isClass("circt.tracker")) {
-        newAnnotations.push_back(anno);
+        auto [it, inserted] =
+            localTrackerIds.insert(anno.getMember<DistinctAttr>("id"));
+        if (inserted)
+          newAnnotations.push_back(anno);
         continue;
       }
 
@@ -1272,16 +1276,18 @@ private:
     // This is a list of all the annotations which will be added to `to`.
     SmallVector<Annotation> newAnnotations;
 
-    // We have special case handling of DontTouch to prevent it from being
-    // turned into a non-local annotation, and to remove duplicates.
+    // We have special case handling of DontTouch and local trackers to prevent
+    // them from being turned into a non-local annotation, and to remove
+    // duplicates.
     llvm::SmallPtrSet<Attribute, 4> dontTouches;
+    llvm::SmallPtrSet<Attribute, 4> localTrackerIds;
 
     // Iterate the annotations, transforming most annotations into non-local
     // ones.
     copyAnnotations(toModule, to, toModule, toAnnos, newAnnotations,
-                    dontTouches);
+                    dontTouches, localTrackerIds);
     copyAnnotations(toModule, to, fromModule, fromAnnos, newAnnotations,
-                    dontTouches);
+                    dontTouches, localTrackerIds);
 
     // Copy over all the new annotations.
     if (!newAnnotations.empty())
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -1251,6 +1251,13 @@ private:
         targetMap[nla.getAttr()].insert(to);
         continue;
       }
+      // If the annotation is a local tracker, don't make it non-local. This
+      // allows trackers that refer to all instances when the user wants that.
+      if (anno.isClass("circt.tracker")) {
+        newAnnotations.push_back(anno);
+        continue;
+      }
+
       // Otherwise make the annotation non-local and add it to the set.
       makeAnnotationNonLocal(toModule.getModuleNameAttr(), to, fromModule, anno,
                              newAnnotations);
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -916,12 +916,13 @@ struct Deduper {
   using RenameMap = DenseMap<StringAttr, StringAttr>;
 
   Deduper(InstanceGraph &instanceGraph, SymbolTable &symbolTable,
-          NLATable *nlaTable, CircuitOp circuit)
+          NLATable *nlaTable, CircuitOp circuit, bool disableLocalAnnotations)
       : context(circuit->getContext()), instanceGraph(instanceGraph),
         symbolTable(symbolTable), nlaTable(nlaTable),
         nlaBlock(circuit.getBodyBlock()),
         nonLocalString(StringAttr::get(context, "circt.nonlocal")),
-        classString(StringAttr::get(context, "class")) {
+        classString(StringAttr::get(context, "class")),
+        disableLocalAnnotations(disableLocalAnnotations) {
     // Populate the NLA cache.
     for (auto nla : circuit.getOps<hw::HierPathOp>())
       nlaCache[nla.getNamepathAttr()] = nla.getSymNameAttr();
@@ -1254,12 +1255,15 @@ private:
       }
       // If the annotation is a local tracker, don't make it non-local. This
       // allows trackers that refer to all instances when the user wants that.
-      if (anno.isClass("circt.tracker")) {
-        auto [it, inserted] =
-            localTrackerIds.insert(anno.getMember<DistinctAttr>("id"));
-        if (inserted)
-          newAnnotations.push_back(anno);
-        continue;
+      // Check the flag to revert to the previous behavior if necessary.
+      if (!disableLocalAnnotations) {
+        if (anno.isClass("circt.tracker")) {
+          auto [it, inserted] =
+              localTrackerIds.insert(anno.getMember<DistinctAttr>("id"));
+          if (inserted)
+            newAnnotations.push_back(anno);
+          continue;
+        }
       }
 
       // Otherwise make the annotation non-local and add it to the set.
@@ -1449,6 +1453,10 @@ private:
 
   /// A module namespace cache.
   DenseMap<Operation *, hw::InnerSymbolNamespace> moduleNamespaces;
+
+  // Disable local annotations and revert to legacy behavior of making
+  // annotations non-local.
+  bool disableLocalAnnotations;
 };
 
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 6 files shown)


prithayan (approved):
Looks good, but do we need to handle ambiguous paths in LowerClass because of this ?

mikeurbach:
Nope, I don't think I don't think a local target is ambiguous. If a user made a path like `Foo>bar:Bar`, then they are referring to the instance bar within Foo. If Foo is instantiated multiple times, that's ok, this path would still refer to the instance bar within any of those instances of Foo, and lower classes is (now) happy to emit a local hw.hierpath in this case (i.e., the hierpath would be `hw.hierpath [@Foo::@bar]` with a single inner name ref).

The difference in this change is if Foo happened to dedupe with something else, this `makeAnnotationsAbsolute` logic would have duplicated the local tracker annotation once for each instance of Foo. By the time this hits lower classes, lower classes has to error out, because there is one path, which originally referred to the instance of bar within the module Foo, and now refers to the instance of bar with each instance of Foo. Effectively, the path was expanded from one reference into many, which we will fail on. So I think this change is fine from that perspective, if a single local path came in, this makes sure a single local path comes out.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -1251,6 +1251,13 @@ struct Deduper {
         targetMap[nla.getAttr()].insert(to);
         continue;
       }
+      // If the annotation is a local tracker, don't make it non-local. This
+      // allows trackers that refer to all instances when the user wants that.
+      if (anno.isClass("circt.tracker")) {
+        newAnnotations.push_back(anno);
+        continue;
+      }
+
```
youngar, line 1260: I am not totally sure how all the path stuff currently works, but the change doesn't seem totally right - `makeAnnotationNonLocal` is maybe not named that great, but it adds more context to a HierPath if it is needed, and it must be called on non-local annotations to make sure their path stays disambiguated.  How can we tell if a path is local or not? Should you make the condition `anno.isClass("circt.tracker") && !anno.hasMember("circt.nonlocal")` kind of thing? 

Does Chisel emit absolute paths to the top of the circuit? If so I could see this code working in practice, but I think someone could write an example where this is changing the behaviour the behavior. Trying to get the idea out there:
```
circuit Foo:
   public module Foo:
     inst bar0 of Bar0
     inst bar1 of Bar1
  module Bar0:
     output path : Path
     wire w
     connect path, path("Bar0>w")
   module Bar1:
     output path : Path
     wire w
     connect path, path("Bar1>w")
```
So my concert is that when Bar0 and Bar1 dedup, the path doesn't get extended with more context, and we would get a complaint about an ambiguous path (which instance of Bar0 is the path referring to?). Is something like that possible?

mikeurbach: > Should you make the condition anno.isClass("circt.tracker") && !anno.hasMember("circt.nonlocal") kind of thing?

Yep, this is effectively the logic here. This is checking for tracker annotations just after we check if the annotation is already non-local: https://github.com/llvm/circt/blob/2085d0d91e3b540ea765539c58bea305dad5a6aa/lib/Dialect/FIRRTL/Transforms/Dedup.cpp#L1247-L1253

> Does Chisel emit absolute paths to the top of the circuit?

Sometimes, but it is up to the user if they make a hierarchical or local target. In the case of hierarchical targets, nothing changes here, since they're already NLAs by dedup, and the logic I linked just above kicks in. So this is really about handling local targets, just like the example you shared. Specifically, Chisel has started emitting paths very similar to this: https://github.com/chipsalliance/chisel/commit/dd2f57ede872afc2e2b9c00dcdbcd397400bcdfd

I cleaned up your example slightly to make it compile, and to have LowerClasses actually do something interesting:

```
FIRRTL version 4.0.0

circuit Foo:
  public module Foo:
    output path0 : Path
    output path1 : Path
    inst bar0 of Bar0
    inst bar1 of Bar1
    propassign path0, bar0.path
    propassign path1, bar1.path

  module Bar0:
    output path : Path
    wire w : UInt<1>
    invalidate w
    propassign path, path("OMReferenceTarget:~Foo|Bar0>w")

  module Bar1:
    output path : Path
    wire w : UInt<1>
    invalidate w
    propassign path, path("OMReferenceTarget:~Foo|Bar1>w")
```

With this change, I think we're getting exactly what we'd want:

```
  om.class @Foo_Class(%basepath: !om.frozenbasepath) {
    %0 = om.frozenbasepath_create %basepath "Foo/bar0"
    %1 = om.object @Bar0_Class(%0) : (!om.frozenbasepath) -> !om.class.type<@Bar0_Class>
    ...
    %3 = om.frozenbasepath_create %basepath "Foo/bar1"
    %4 = om.object @Bar0_Class(%3) : (!om.frozenbasepath) -> !om.class.type<@Bar0_Class>
    ...
  }
  om.class @Bar0_Class(%basepath: !om.frozenbasepath) {
    %0 = om.frozenpath_create reference %basepath "Bar0>w"
    om.class.field @path, %0 : !om.frozenpath
  }

```

The local path is just referring to the deduped `Bar0>w` after Dedup runs, and LowerClasses [recently](https://github.com/llvm/circt/commit/002427065e8b0751a5c0ed8a980addf99f7ad82a) started supporting passing local paths through to the output without trying to disambiguate them or erroring in the face of multiple instantiation. Note that the different instances of `Bar0_Class` do get different base paths, so the final elaborated paths here get built up hierarchically following the instance graph.

This doesn't show up in your exact example, but in similar cases, without the change in this PR, after Bar0 and Bar1 dedupe, we'd clone the tracker into two: one for each path to the `w` wire. This will fail in LowerClasses here: https://github.com/llvm/circt/blob/2085d0d91e3b540ea765539c58bea305dad5a6aa/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp#L715-L725


This is the general thrust of this change and some of the recent changes to paths--as we start seeing more uses of D/I in Chisel, the notion that you can even make a unique, absolute path from the top of the circuit becomes impossible, so instead we're doing more with local paths, and not forcing ourselves to disambiguate in places we did previously. Instead, paths can safely be local, and compose hierarchically using the basepath infrastructure in place in LowerClasses.

Does that help alleviate your concerns @youngar ?

youngar: Yeah! Either I just can't read code or I forgot how this works - the context adding that I was talking about happens elsewhere and has nothing to do with your change. Sorry for the noise!

youngar:
How do we end up with duplicated local trackers, anyways?  Maybe the inliner could do it?

mikeurbach:
> How do we end up with duplicated local trackers, anyways? Maybe the inliner could do it?

Likely inliner, potentially others. I need to chase this down and resolve it. I added dedupe of local trackers here in Dedup similar to how we do it for the local dont touch annotations, but in some testing downstream I'm seeing other sources of duplicated local trackers that I need to chase down.

mikeurbach:
I think the last commit is still needed here. But the problem I was seeing downstream was actually an interaction with PrefixModules, which can also lead to duplicate annotations in cases like the examples here. That is more fundamental; I don't think there is anything we can do there with the current representation of trackers as annotations, so we'll need to hold this until module prefixes are applied by Chisel and not annotations. I do think we'll want this change at some point.

mikeurbach:
And, now that I think about this, I may have been mistaken; the last commit here was probably _not_ needed, and was me confusing the interaction with PrefixModules with the logic in Dedup.

mikeurbach:
As mentioned earlier, this can only roll out once we've deprecated and stopped relying on `PrefixModules`. In order to support the new feature in this patch independently of the reliance on `PrefixModules`, I've gated this behind a flag, which defaults to the new behavior. That way, we could roll this out, and opt-in to the original behavior downstream until we are ready, or flip back at some point if necessary, etc.

seldridge (commented):
Sorry, only seeing this now.

One nit, one broad concern.

Nit: this option `disable-local-annotations` is inappropriately named and described. This _only_ affects OM tracker annotations and nothing else, correct? Can this be made more obvious that this is only affecting the way that tracker annotations dedup and not _all_ annotations as it seems to imply?

I've pushed back very strongly on this kind of thing before, primarily because it is an information-destroying operation. All annotations are supposed to have the behavior of becoming non-local on deduplication (with the one exception of don't touch... which is bad). This is the information preserving thing to do. Somebody annotated that these two things, they are then made the same by a pass, the correct thing is always to preserve that difference via a path. Downstream passes then need to understand what to do with duplicate information and can apply their interpretation of what a duplicate annotation means.

Usually when this comes up it means that an annotation is being inappropriately used and a different representation is needed (attribute, intrinsic, operation, etc.). I'm fine with this as a stop-gap towards something better (attribute, intrinsic, operation, etc.). Has there been thought on what the replacement could be?

mikeurbach:
Regarding the option name--yes I can make that more explicit, since it's really scoped to OMIR tracker annotations.

Regarding the broader concern...

This is intended to handle cases like what I described in the earlier comments, especially the hidden comments from discussion with @youngar. This will come up more as we start to use D/I in Chisel for more things. Lots of our FIRRTL support for Path properties was built on previous assumptions, like the idea that you could always form a unique hierarchical path to the targeted entity. These assumptions won't hold when you start using definitions and instances, and one valid way to relax these assumptions is to allow local paths, rather than trying to enforce the older hierarchical path assumptions.

I know you're concerned about losing information, but I think this is possible to do without loss of information--if an annotation is local, it is tracking a module or something inside a module. If that module deduplicates with another module, why do we need to preserve these two things that are now the same used to be different?

> Has there been thought on what the replacement could be?

I don't know if this is written down somewhere, but what we really want is to stop using annotations for OMIR, and instead pass values through the IR more like we did with probes. This will require, among other things, the ability to "take a reference" to an instance, which may necessitate changing firrtl.instance to have its own type. This is where @rwy7 started over a year ago, but after a decent amount of yak shaving, we decided to abandon that design point and use annotations "
for now". Now that we are trying to do new things we couldn't do before, perhaps it's time to update the implementation.

---

All that said, the specific thing this was trying to deal with is the situation where have some Path using an OMIR tracker that locally refers to a module or something inside a module. The way we've implemented Paths and OMIR trackers, there is a distinct attribute linking the Path op to the local annotation. This code path could blow out that local annotation to multiple non-local annotations. So, we'll have one Path op, but its distinct attribute shows up multiple times in the IR, and we'll hit this: https://github.com/llvm/circt/blob/097483d892b13a4c2020a37f1016b3f02460efed/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp#L719-L726

Maybe the right thing to do here is not to handle this in LowerClasses... we could keep splatting out local annotations into non-local annotations, and the distinct attribute handling code in LowerClasses could check if they all refer to the same thing.

So I think I want to check a couple things:

* Now that we aren't doing PrefixModules anymore, do we even need this change?
* Could we handle this in LowerClasses instead?

It might be the case that we don't need to relax this behavior of Dedup, and can keep the information for longer, so it is up to the passes to deal with duplicated annotations.

seldridge:
Thanks for the detailed comments and analysis, Mike. If it wasn't obvious, I missed the entire discussion on this!

Yes, I agree that the right thing is to change the instance type (or add an additional return that is a reference to the instance, or have a dedicated op that takes an inner symbol and returns a reference...). I'm also going to run into the instance return type problem with one of the solutions considered for replacing Grand Central Views which requires vector instantiation. There was always general consensus that vector instantiation in FIRRTL made sense. However, the early choice to use multiple return types makes this weird.

Maybe we bite the bullet and do the return type change in preparation for this. 😅 

And, I'm not generally opposed if this PR does come back post-revert (nor did I think it necessarily needed to be reverted). Per-annotation dedup behavior is what I'm generally afraid of as a solution.

Also, it was unstated that I don't think don't touch should be an annotation at this point (which would remove the other special annotation handling case).

mikeurbach:
> Maybe we bite the bullet and do the return type change in preparation for this. 😅

Maybe, at least something to consider for sure.

> And, I'm not generally opposed if this PR does come back post-revert (nor did I think it necessarily needed to be reverted). Per-annotation dedup behavior is what I'm generally afraid of as a solution.

Understood.. I eagerly reverted because I'm now questioning if this was even necessary, and I think it could be re-landed easily, either as-is in a pinch, or ideally with a different approach that doesn't require per-annotation dedup behavior. I think we should be able to do what you were getting at earlier: "Downstream passes then need to understand what to do with duplicate information and can apply their interpretation". I believe I had tried that originally, and got stuck because I was also dealing with the interaction with module prefixing as a FIRRTL transform, and not understanding what the root issue was. With that out of the way, either the problem this PR was trying to solve would go away, or it could be solved in the relevant downstream pass.

PR #7712, opened 2024-10-17
[CombFolds] Preserve two-state attribute in `narrowOperationWidth`

fzi-hielscher:
We're already doing a best-effort preservation of discardable attributes through `narrowOperationWidth`. I cannot think of a reason why the inherent two-state attribute should not be carried over.


fabianschuiki (approved):
LGTM!

PR #7717, opened 2024-10-20
RFC: [AIG] Add an AIG dialect and circt-synth

uenoku:
This proposes AIG(And Inverter Graph) dialect which is meant to be a foundational dialect for logic synthesis. 

Please see `docs/Dialects/AIG/RationaleAIG.md` and `include/Diaclects/AIG/AIGOps.td` for design details.

This PR implements E2E flow for FPGA synthesis with a naive implementation for LUT mapping (`test/circt-synth/lut-size.mlir`).


sequencer:
Since the developing of logic synthesis is really a non-trivial work, I wonder rather than using CIRCT to do the synthesis job by itself, how about exporting the [aiger file](https://fmv.jku.at/aiger/index.html) directly, and then we can consume it by [abc](https://github.com/berkeley-abc/abc) or [mockturtle](https://github.com/lsils/mockturtle) for future use.

uenoku:
> Since the developing of logic synthesis is really a non-trivial work, I wonder rather than using CIRCT to do the synthesis job by itself, how about exporting the [aiger file](https://fmv.jku.at/aiger/index.html) directly, and then we can consume it by [abc](https://github.com/berkeley-abc/abc) or [mockturtle](https://github.com/lsils/mockturtle) for future use.

Implementing AIGER file exporter/importer is trivial once we lower the IR to (low) AIG dialect (= `aig.and_inv` is lowered to have a single bit and two operands). It's totally reasonable to have an integration with AIGER file to leverage other tools and I'm happy to work/help on that! 

That said even though I agree it's non-trivial to logic synthesizer but I believe it would be feasible to get OK results once we implemented necessary passes (Lower to FRAIG, Bit-sensitive IMCP/IMDCE, inlining, re-balancing).

TaoBi22:
Super exciting :smile:

lib/Dialect/AIG/AIGOps.cpp
```diff
@@ -0,0 +1,141 @@
...
+  // TODO: Implement this.
+  return {};
+}
+
+LogicalResult AndInverterOp::canonicalize(AndInverterOp op,
+                                          PatternRewriter &rewriter) {
+  // TODO: Extend to more than 2 inputs.
+  if (op.getInputs().size() == 2) {
+    if ((op.getInputs()[0] == op.getInputs()[1]) &&
+        (op.isInverted(0) == op.isInverted(1))) {
```
uenoku, line 34: Flag check is missing.

```suggestion
        (op.isInverted(0) == op.isInverted(1)) && op.isInverted(0)) {
```
Also we need to implement three cases:
```
a & a -> a
not a & a -> 0
not a & not a -> not a
```

uenoku:
>  One question I had after a cursory look: is it beneficial to allow each operand in aig.and_inv to be inverted, or does that make writing passes more complex?

Inverting on operands would be more closer to traditional AIG format including AIGER (thank you for checking @fabianschuiki!) so the current representation seems to be good way to go.

lib/Dialect/AIG/AIGOps.cpp
fabianschuiki: Maybe this could iterate over the operands in a loop and keep a small map on the side with operands previously seen. And then when you see a duplicate you either discard id (if it's the same inversion), or you replace the gate with cosntant zero.

```
SmallDenseMap<Value, bool> seen;
SmallVector<Value> uniqueValues;
SmallVector<bool> uniqueInverts;
for (auto [value, inverted] : llvm::zip(op.getInputs(), op.getInverted())) {
  auto it = seen.find(value);
  if (it == seen.end()) {
    seen.insert(it, {value, inverted});
    uniqueValues.push_back(value);
    uniqueInverts.push_back(inverted);
  } else if (it->second != inverted) {
    // replace with const 0
    return
  }
}
// build new op with reduced input values
```

lib/Dialect/AIG/AIGOps.cpp
```diff
@@ -0,0 +1,141 @@
...
+  SmallVector<bool> inverts;
+  auto loc = parser.getCurrentLocation();
+
+  while (true) {
+    if (succeeded(parser.parseOptionalKeyword("not"))) {
+      inverts.push_back(true);
+    } else {
+      inverts.push_back(false);
+    }
+    operands.push_back(OpAsmParser::UnresolvedOperand());
```
fabianschuiki, line 58: Could this be done with something like the custom bool keyword parser/printer that we have somewhere in `circt/Support`?

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
```diff
@@ -0,0 +1,297 @@
...
+        newCut.getBodyBlock()->begin(), block->getOperations());
+    rewriter.replaceOp(op, newCut);
+    return success();
+  }
+};
+} // namespace
+
+//===----------------------------------------------------------------------===//
+// Convert Comb to AIG pass
+//===----------------------------------------------------------------------===//
```
fabianschuiki, line 268: Stale comment

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
```diff
@@ -0,0 +1,297 @@
...
+    return failure();
+  }
+};
+
+struct SinkConstantPattern : public mlir::RewritePattern {
+  SinkConstantPattern(MLIRContext *context)
+      : mlir::RewritePattern(aig::CutOp::getOperationName(), 1, context) {}
+
+  LogicalResult matchAndRewrite(Operation *opT,
+                                PatternRewriter &rewriter) const override {
```
fabianschuiki, line 226: Would `OpRewritePattern` clean up some of the `getOperationName()` and `opT -> op` casting?

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
```diff
@@ -0,0 +1,297 @@
...
+      // Always enable a bit for 0 (the original cut).
+      auto result = checkSubsetMerge((i << 1) | 1);
+      if (succeeded(result))
+        return result;
+    }
+    return failure();
+  }
+};
+
+struct SinkConstantPattern : public mlir::RewritePattern {
```
uenoku, line 221: This is always beneficial so it must be moved to CutOp canonicalization

lib/Dialect/AIG/Transforms/LowerCutToLUT.cpp
```diff
@@ -0,0 +1,150 @@
...
+  // Check if every type is i1.
+  for (auto operand : cutOp.getOperands()) {
+    if (operand.getType() != rewriter.getI1Type())
+      return cutOp.emitError("expected i1 type");
+  }
+
+  for (auto result : cutOp.getResults()) {
+    if (result.getType() != rewriter.getI1Type())
+      return cutOp.emitError("expected i1 type");
+  }
```
fabianschuiki, line 67: I'm not sure you're supposed to emit errors inside a rewrite pattern :thinking:. The dialect conversion framework might call these multiple times, and if multiple rewrites apply to the same op, only one has to succeed. But in that case you wouldn't want to see errors printed from the other patterns that failed to apply.

I think you could just move the entire code here into a `getOperation().walk([&](CutOp cutOp){ ... });` call inside the pass itself.

lib/Dialect/AIG/Transforms/LowerVariadic.cpp
```diff
@@ -0,0 +1,117 @@
...
+  target.addIllegalOp<aig::AndInverterOp>();
+  target.addDynamicallyLegalOp<aig::AndInverterOp>(
+      [](AndInverterOp op) { return op.getInputs().size() <= 2; });
+
+  RewritePatternSet patterns(&getContext());
+  populateLowerVariadicPatterns(patterns);
+
+  if (failed(mlir::applyPartialConversion(getOperation(), target,
+                                          std::move(patterns))))
+    return signalPassFailure();
```
fabianschuiki, line 112: Could this be a greedy rewriter instead? I don't think you really need the dialect conversion framework. You could check for `op.getInputs().size() <= 2` in the match and return false to indicate that the pattern doesn't apply one an op has been reduced to 2 or fewer operands.

lib/Dialect/AIG/Transforms/LowerWordToBits.cpp
```diff
@@ -0,0 +1,100 @@
...
+  target.addDynamicallyLegalOp<aig::AndInverterOp>([](AndInverterOp op) {
+    return op.getType().getIntOrFloatBitWidth() <= 1;
+  });
+
+  RewritePatternSet patterns(&getContext());
+  populateLowerWordToBitsPatterns(patterns);
+
+  if (failed(mlir::applyPartialConversion(getOperation(), target,
+                                          std::move(patterns))))
+    return signalPassFailure();
```
fabianschuiki, line 95: Maybe this could be a greedy rewriter as well? Your pattern match already nicely reports failure if the op has an `i1` result, so that should just work.

lib/Dialect/AIG/Transforms/LowerCutToLUT.cpp
uenoku: yeah thanks. These are stale verification which I originally implemented with OpConversionPattern. I'll move these to walk.

test/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(CIRCT_TEST_DEPENDS
   circt-dis
   circt-lec
   circt-opt
+  circt-synth
```
fabianschuiki, line 29: This makes me unreasonably happy :partying_face: :smile:

tools/circt-synth/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+set(LLVM_LINK_COMPONENTS
+  Support
+)
```
fabianschuiki, line 3: If I recall correctly, `Support` is now automatically included for any CIRCT/MLIR tools, such that this is no longer needed.

tools/circt-synth/circt-synth.cpp
```diff
@@ -0,0 +1,218 @@
...
+static llvm::cl::opt<Until> runUntilAfter(
+    "until-after", llvm::cl::desc("Stop pipeline after a specified point"),
+    runUntilValues, llvm::cl::init(UntilEnd), llvm::cl::cat(mainCategory));
+
+// LUT-k parameter. This needs to be unifined to a more fine-grained target
+// architecture information.
+static cl::opt<int> lutSize("lut-size",
+                            cl::desc("Size of LUT to use for mapping"),
+                            cl::init(6), cl::cat(mainCategory));
+//===----------------------------------------------------------------------===//
```
fabianschuiki, line 88:
```suggestion
                            cl::init(6), cl::cat(mainCategory));

//===----------------------------------------------------------------------===//
```

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
uenoku: Yeah I missed OpRewritePattern for some reason ...

tools/circt-synth/circt-synth.cpp
fabianschuiki, line 88:
```suggestion

//===----------------------------------------------------------------------===//
```

docs/Dialects/AIG/RationaleAIG.md
```diff
@@ -0,0 +1,24 @@
...
+
+The ability to represent multibit operations during synthesis is crucial for scalable logic optimization. This approach enables a form of vectorization, allowing for batch processing of logic synthesis when multibit operations are constructed in a similar manner. Such vectorization can significantly improve the efficiency and performance of the synthesis process.
+
+## aig.cut
+
+The `aig.cut` operation represents a "cut" in the logic tree. This operation possesses the `IsolatedAbove` trait and contains a single block. Its input operands represent the input edges, while the returned value represents the output edges.
+
+This operation proves particularly useful for progressive LUT mapping. For instance, a k-input cut can be readily mapped to a k-input LUT. Consequently, the subsequent stages of the pipeline can concentrate on replacing combinational logic with k-input Cut operations, simplifying the overall process.
```
fabianschuiki, line 24: No need to do any of this now, but we might be able to combine this rationale file with the auto-generated docs into a single `docs/Dialects/AIG.md` file. I think the LTL dialect might already do that. This would allow you to move the nice operation rationale here into the op's summary/description fields, and still have them show up in the docs as expected. And the rationale above would be the preface of the dialect doc.

lib/Dialect/AIG/Transforms/LowerCutToLUT.cpp
```diff
@@ -0,0 +1,150 @@
+//===- LowerCutToLUT.cpp ---------------------------------------------===//
```
uenoku, line 1:
```suggestion
//===- LowerCutToLUT.cpp ------------------------------------------------===//
```

fabianschuiki (approved):
This is seriously cool stuff! Can't wait to have this landed and be a starting point for optimizations, tech mapping, potential static timing analysis, etc. Really cool!

lib/Dialect/AIG/AIGOps.cpp
uenoku: We need to modify parseKeywordBool quite a bit since it currently takes BoolAttr& and both trueKeyword/falseKeyword so I think it's fine to keep separated implementation.

lib/Dialect/AIG/AIGOps.cpp
uenoku: Great point, thanks!

docs/Dialects/AIG/RationaleAIG.md
uenoku: Moved, thanks!

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
```diff
@@ -0,0 +1,288 @@
...
+    rewriter.create<aig::OutputOp>(loc, ValueRange{valueToNewValue.at(output)});
+  }
+
+  for (auto oldCut : llvm::reverse(cuts)) {
+    auto oldCutOutput =
+        cast<aig::CutOp>(oldCut).getBodyBlock()->getTerminator();
+    oldCutOutput->erase();
+    cutOp.getBodyBlock()->getOperations().splice(
+        cutOp.getBodyBlock()->begin(),
+        cast<aig::CutOp>(oldCut).getBodyBlock()->getOperations());
```
mikeurbach, line 109: To be safe, should this mutation and the above use the rewriter?

lib/Dialect/AIG/Transforms/LowerVariadic.cpp
```diff
@@ -0,0 +1,112 @@
...
+    else
+      return operands[0];
+  case 2:
+    lhs = operands[0];
+    rhs = operands[1];
+    return rewriter.create<AndInverterOp>(op.getLoc(), lhs, rhs, inverts[0],
+                                          inverts[1]);
+  default:
+    auto firstHalf = operands.size() / 2;
+    lhs = lowerFullyAssociativeOp(op, operands.take_front(firstHalf),
```
mikeurbach, line 58: This seems fine for proof of concept, but is there a potential for stack overflow with large inputs if we recurse along use-def edges?

mikeurbach (approved):
A couple minor questions. This is awesome in general. The motivation and dialect boilerplate make sense to me. I skimmed through the new tool and the initial passes as well, and they make sense as a starting point. Looking forward to seeing more advanced integration here!

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
uenoku: Yeah that's great point. I'll modify it through rewriter.

lib/Dialect/AIG/Transforms/LowerVariadic.cpp
uenoku: The recursion depth is O(log N) where N is operand size so it should be fine (actually we perform the same transformation at PrepareForEmission).

lib/Dialect/AIG/AIGOps.cpp
fabianschuiki: That makes total sense!

lib/Dialect/AIG/Transforms/LowerVariadic.cpp
mikeurbach: I see, fair enough.

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
```diff
@@ -0,0 +1,292 @@
...
+    rewriter.setInsertionPointToEnd(cutOp.getBodyBlock());
+    rewriter.create<aig::OutputOp>(loc, ValueRange{valueToNewValue.at(output)});
+  }
+
+  for (auto oldCut : llvm::reverse(cuts)) {
+    auto *oldCutBlock = cast<aig::CutOp>(oldCut).getBodyBlock();
+    auto oldCutOutput = oldCutBlock->getTerminator();
+    oldCutOutput->erase();
+    // Erase arguments before inlining. Arguments are already replaced.
+    oldCutBlock->eraseArguments([](BlockArgument block) { return true; });
```
maerhart, line 108: I think these things should go through the rewriter and not be done directly on the op/block.

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
```diff
@@ -0,0 +1,292 @@
...
+                                      cloned->getResult(0));
+          changed = true;
+        } else {
+          oldInputs.push_back(in);
+          oldArgs.push_back(block->getArgument(i));
+        }
+      }
+      if (!changed)
+        return failure();
+      block->eraseArguments(eraseArgs);
```
maerhart, line 241: Shouldn't this call and the one further down go through the rewriter?

maerhart (approved):
Really cool to see this coming!

test/Dialect/AIG/canonicalizer.mlir
```diff
@@ -0,0 +1,42 @@
...
+  %4 = aig.and_inv %a, not %c15 : i4
+
+  // a & a & b -> a & b
+  %5 = aig.and_inv %a, %b : i4
+
+  // a & ~a & b -> 0
+  %6 = hw.constant 0 : i4
+
+  // 15 & 15 -> 15
+  %7 = hw.constant -1 : i4
```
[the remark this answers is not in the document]
uenoku: Oh, sorry. Seems like I wrongly added post canonicalization IR as an input..

uenoku:
@maerhart Thank you for really detailed review! and really sorry for a bunch of dirty code remaining in my PR. I'll make sure to double check basic stuffs before removing draft..

include/circt/Dialect/AIG/AIGOps.td
```diff
@@ -0,0 +1,120 @@
...
+    ^bb0(%arg0: i1, %arg1: i1, %arg2: i1, %arg3: i1):
+      %0 = aig.and_inv not %arg0, %arg1 : i1
+      %1 = aig.and_inv %arg1, %arg3 : i1
+      aig.output %0, %1 : i1
+    }
+    ```
+
+  }];
+  let arguments = (ins Variadic<AnyType>:$inputs);
+  let results = (outs Variadic<AnyType>:$results);
```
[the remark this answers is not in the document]
uenoku: Yes, I added TODO for that.

lib/Dialect/AIG/AIGOps.cpp
```diff
@@ -0,0 +1,192 @@
...
+    if (parser.parseCustomTypeWithFallback(type))
+      return mlir::failure();
+    resultRawType = type;
+  }
+
+  result.addTypes(resultTypes);
+  result.addAttribute("inverted",
+                      parser.getBuilder().getDenseBoolArrayAttr(inverts));
+  if (parser.resolveOperands(operands, resultTypes[0], loc, result.operands))
+    return mlir::failure();
```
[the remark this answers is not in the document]
uenoku: Fixed, sorry about that.

lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -0,0 +1,117 @@
...
+};
+
+/// Lower a comb::XorOp operation to AIG operations
+struct CombXorOpConversion : OpConversionPattern<XorOp> {
+  using OpConversionPattern<XorOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(XorOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Xor using And with invert flags: a ^ b = (a | b) & (~a | ~b)
```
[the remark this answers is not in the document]
uenoku: That's really a good boint. This must be lowered to binary op first.

lib/Dialect/AIG/Transforms/GreedyCutDecomp.cpp
uenoku: Thank you for pointing out. oldCutOutput must be erased through rewriter but block arguments should be fine to erase directly. There is no API in rewriter to erase block arguments and upstream does the same (https://github.com/llvm/llvm-project/blob/a285ba7529feaa7c49890e314facb5e9f4d8dc11/mlir/lib/Conversion/TosaToSCF/TosaToSCF.cpp#L38-L39)

uenoku:
Thank you really for review @mikeurbach @fabianschuiki @maerhart! I'm going to merge dialect definitions and circt-synth boilerplate and separate complicated passes (GreedyCutDecomp and CutToLUT)  and canonicalizer into a different PR for the ease of review.

lib/Dialect/AIG/Transforms/LowerWordToBits.cpp
```diff
@@ -0,0 +1,91 @@
...
+      }
+      auto andInverter = rewriter.create<AndInverterOp>(op.getLoc(), operands,
+                                                        op.getInvertedAttr());
+      result.push_back(andInverter);
+    }
+
+    auto concat = rewriter.create<comb::ConcatOp>(op.getLoc(), result);
+    rewriter.replaceOp(op, concat);
+    return success();
+  }
```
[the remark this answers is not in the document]
uenoku: That's fair point, I'll address in the separated PR.

Issue #7716, opened 2024-10-18
[HW] Canonicalization creates invalid aggregate constant


uenoku:
```mlir
$ cat bar.mlir
module {
  hw.module @test(out o : !hw.array<1x!seq.clock>) {
    %0 = seq.const_clock  high
    %1 = hw.array_create %0 : !seq.clock
    hw.output %1 : !hw.array<1x!seq.clock>
  }
}

$ circt-opt bar.mlir -canonicalize 
bar.mlir:3:8: error: 'hw.aggregate_constant' op unknown element type'!seq.clock'
  %0 = hw.array_create %high : !seq.clock
       ^
bar.mlir:3:8: note: see current operation: %0 = "hw.aggregate_constant"() {fields = [#seq<clock_constant high>]} : () -> !hw.array<1x!seq.clock>
```

PR #7718, opened 2024-10-21
[HW][Seq] Allow typed attr to be an element of aggregate_constant and make seq.const_clock typed attr

uenoku:
This fixes https://github.com/llvm/circt/issues/7716. Aggregate constant verifier rejects unknown attributes hence hw.aggregate_constant cannot be used for attributes defined by other dialects (in this case seq.const_clock) was rejected even when hw.aggregate_create allows users to create clock type arrays. This PR loosen the restriction by allowing TypedAttribute. 

This PR also adds TypedAttrInterface to seq.const_clock and support its lowering.

### DIFF
--- a/include/circt/Dialect/Seq/SeqAttributes.td
+++ b/include/circt/Dialect/Seq/SeqAttributes.td
@@ -22,7 +22,13 @@ def ClockConstAttrImpl: I32EnumAttr<"ClockConst", "clock constant",
                                 [ClockLow, ClockHigh]> {
   let genSpecializedAttr = 0;
 }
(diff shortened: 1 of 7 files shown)


mikeurbach (commented):
I'm not too familiar with the seq const clock stuff, but this seems reasonable. It doesn't introduce any new circular dependencies or anything like that right?

uenoku:
> It doesn't introduce any new circular dependencies or anything like that right?

Yes, it uses `TypedAttr` instead of directly using seq.clock to avoid circular dependency between HW and Seq.

fabianschuiki (approved):
This looks reasonable to me!

PR #7719, opened 2024-10-21
[Verif] Add ignore attribute to formal

leonardt:
By default, `circt-test` will not include "ignored" verif.formal ops in the output but has an option `--include-ignored` to emit them.


tools/circt-test/circt-test.cpp
```diff
@@ -66,6 +69,12 @@ Options opts;
 // Tool Implementation
 //===----------------------------------------------------------------------===//
 
+bool shouldIgnore(verif::FormalOp formalOp, Options &opts) {
+  return formalOp->hasAttr("ignore") &&
```
uenoku, line 73: nit: You can use hasAttrOfType to safely cast the attribute
```suggestion
  return formalOp->hasAttrOfType<BoolAttr>("ignore") &&
```

test/circt-test/basic.mlir
```diff
@@ -31,13 +32,21 @@ verif.formal @Some.TestB {}
...
-    wow = false
+    wow = false,
+    ignore = false
+} {}
+
+// CHECK-NOT: "name": "Ignore"
+// JSON-NOT: "name": "Ignore"
+verif.formal @Ignore attributes {
+    ignore = true,
+    this = "should not appear"
```
fabianschuiki, line 49: It might be worthwhile folding the `include-ignored.mlir` file into this one. No need to check all permutations with circt-as again. Maybe just another RUN line with `circt-test %s --include-ignored | FileCheck %s --check-prefix=CHECK-WITH-IGNORED` or similar.

fabianschuiki (approved):
Nice and elegant :slightly_smiling_face: LGTM

tools/circt-test/circt-test.cpp
```diff
@@ -66,6 +69,12 @@ Options opts;
 // Tool Implementation
 //===----------------------------------------------------------------------===//
 
+bool shouldIgnore(verif::FormalOp formalOp, Options &opts) {
+  return formalOp->hasAttrOfType<BoolAttr>("ignore") &&
+         cast<BoolAttr>(formalOp->getAttr("ignore")).getValue() &&
```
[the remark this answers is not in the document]
maerhart: Or, to avoid double lookup:
```
if (auto boolAttr = formalOp->getAttrOfType<BoolAttr>("ignore"))
  return boolAttr.getValue() && !opts.includeIgnored;

return false;
```

uenoku (approved):
LGTM! I'm a bit worried about using discardable attribute for this purpose but I think it's fine as an initial iteration.

leonardt:
If anyone wants to give this a final review: I needed to resolve merge conflicts with Fabian's `sby` runner changes and added logic to `execute` to handle ignored tests

fabianschuiki (approved):

PR #7728, opened 2024-10-24
[circt-bmc] Add simple initial value support to ExternalizeRegisters

TaoBi22:
Adds support for initial values (as long as they're defined directly by a `hw.constant` op inside a `seq.initial`) to `ExternalizeRegisters` - I'll add the support for this in circt-bmc as a follow-up to slim down the diff. Initial values are added in an ArrayAttr on the module op - if there's no initial value then it just adds a unit attr

### DIFF
--- a/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
+++ b/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
@@ -98,6 +98,12 @@ void ExternalizeRegistersPass::runOnOperation() {
           if (auto initVal = regOp.getInitialValue()) {
             // Find the seq.initial op where the initial value is defined and
             // fetch the operation inside that defines the value
+            if (isa<BlockArgument>(initVal) ||
+                !isa<seq::InitialOp>(initVal.getDefiningOp())) {
+              regOp.emitError("registers with initial values not directly "
+                              "defined by a seq.initial op not yet supported");
+              return signalPassFailure();
+            }
             auto *initialOp = regOp.getInitialValue().getDefiningOp();
             auto index =
                 llvm::find(initialOp->getResults(), initVal).getIndex();
@@ -113,7 +119,8 @@ void ExternalizeRegistersPass::runOnOperation() {
               initState = constantOp.getValueAttr();
             } else {
               regOp.emitError("registers with initial values not directly "
-                              "defined by a hw.constant op not yet supported");
+                              "defined by a hw.constant op in a seq.initial op "
+                              "not yet supported");
               return signalPassFailure();
             }
           } else {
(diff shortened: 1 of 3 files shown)


lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -93,9 +94,39 @@ void ExternalizeRegistersPass::runOnOperation() {
...
           }
-          if (regOp.getInitialValue()) {
-            regOp.emitError("registers with initial values not yet supported");
-            return signalPassFailure();
+          mlir::Attribute initState;
+          if (auto initVal = regOp.getInitialValue()) {
+            // Find the seq.initial op where the initial value is defined and
+            // fetch the operation inside that defines the value
+            if (isa<BlockArgument>(initVal) ||
+                !isa<seq::InitialOp>(initVal.getDefiningOp())) {
```
maerhart, line 102:
```suggestion
            auto initialOp = initVal.getDefiningOp<seq::InitialOp>();
            if (!initialOp) {
```

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -93,9 +94,39 @@ void ExternalizeRegistersPass::runOnOperation() {
...
+              regOp.emitError("registers with initial values not directly "
+                              "defined by a seq.initial op not yet supported");
+              return signalPassFailure();
+            }
+            auto *initialOp = regOp.getInitialValue().getDefiningOp();
+            auto index =
+                llvm::find(initialOp->getResults(), initVal).getIndex();
```
maerhart, line 109:
```suggestion
            auto index = cast<OpResult>(initVal).getResultNumber();
```

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -93,9 +94,39 @@ void ExternalizeRegistersPass::runOnOperation() {
...
+            auto initValDef =
+                initialOp->getRegion(0).front().getTerminator()->getOperand(
+                    index);
+            // If it's defined by a constant op then just fetch the constant
+            // value - otherwise unsupported
+            if (auto constantOp =
+                    dyn_cast<hw::ConstantOp>(initValDef.getDefiningOp())) {
```
maerhart, line 116:
```suggestion
                    initValDef.getDefiningOp<hw::ConstantOp>()) {
```

maerhart (approved):
Very cool! Thanks for adding that! 🎉

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
TaoBi22: Nice, thanks!

PR #7729, opened 2024-10-24
[circt-bmc] Add initial_values attribute to BMC op

TaoBi22:
Adds an initial_values attribute to the verif.bmc op and carries the attribute through from the externalize registers version introduced in #7728. Eventually it would be nice to let the init region handle this (along with letting it control input values and letting the loop region set them), but I think this does the trick for now

### DIFF
--- a/include/circt/Dialect/Verif/VerifOps.td
+++ b/include/circt/Dialect/Verif/VerifOps.td
@@ -203,7 +203,7 @@ def BoundedModelCheckingOp : VerifOp<"bmc", [
     match the types yielded by the `init` region).
   }];
 
-  let arguments = (ins I32Attr:$bound, I32Attr:$num_regs);
+  let arguments = (ins I32Attr:$bound, I32Attr:$num_regs, ArrayAttr:$initial_values);
   let regions = (region SizedRegion<1>:$init,
                         SizedRegion<1>:$loop,
                         SizedRegion<1>:$circuit);
@@ -211,7 +211,7 @@ def BoundedModelCheckingOp : VerifOp<"bmc", [
   let results = (outs I1:$result);
 
   let assemblyFormat = [{
-    `bound` $bound `num_regs` $num_regs attr-dict-with-keyword `init` $init `loop` $loop `circuit`
+    `bound` $bound `num_regs` $num_regs `initial_values` $initial_values attr-dict-with-keyword `init` $init `loop` $loop `circuit`
     $circuit
   }];
 
--- a/lib/Tools/circt-bmc/LowerToBMC.cpp
+++ b/lib/Tools/circt-bmc/LowerToBMC.cpp
@@ -87,13 +87,16 @@ void LowerToBMCPass::runOnOperation() {
   // Double the bound given to the BMC op, as a clock cycle takes 2 BMC
   // iterations
   verif::BoundedModelCheckingOp bmcOp;
-  if (auto numRegs = hwModule->getAttrOfType<IntegerAttr>("num_regs"))
+  auto numRegs = hwModule->getAttrOfType<IntegerAttr>("num_regs");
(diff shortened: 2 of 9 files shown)


include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -203,15 +203,15 @@ def BoundedModelCheckingOp : VerifOp<"bmc", [
     match the types yielded by the `init` region).
   }];
 
-  let arguments = (ins I32Attr:$bound, I32Attr:$num_regs);
+  let arguments = (ins I32Attr:$bound, I32Attr:$num_regs, ArrayAttr:$initial_values);
```
maerhart, line 206: Maybe add some comment/TODO to indicate that this attribute is only a temporary solution until we support it in init regions?

lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -87,13 +87,16 @@ void LowerToBMCPass::runOnOperation() {
   // Double the bound given to the BMC op, as a clock cycle takes 2 BMC
   // iterations
   verif::BoundedModelCheckingOp bmcOp;
-  if (auto numRegs = hwModule->getAttrOfType<IntegerAttr>("num_regs"))
+  auto numRegs = hwModule->getAttrOfType<IntegerAttr>("num_regs");
+  auto initialValues = hwModule->getAttrOfType<ArrayAttr>("initial_values");
+  if (numRegs && initialValues)
```
maerhart, line 92: Ideally, we'd also check the attribute types in the array.

maerhart (approved):
LGTM! Just wondering why you decided to add an attribute to the bmc op for now rather than using the init region directly. Is there something that makes using the init region considerably more complicated?

TaoBi22:
> LGTM! Just wondering why you decided to add an attribute to the bmc op for now rather than using the init region directly. Is there something that makes using the init region considerably more complicated?

Specifically initial values for the registers isn't super complicated - my rationale was that we probably also want to give the init/loop regions control over inputs too at some point & there's a question too of whether to let the loop region be responsible for resets (which I'm not totally sure on as it would reduce the need for extra information but could also mean that a mistake generating a BMC op could significantly mess with the expected semantics of the circuit region by not propagating register state correctly). My thinking was that it would probably be nice to make these not-fully-decided changes to the op regions all together so there's a little more consistency (but since you need an initial value to check most meaningful properties it seemed worth adding like this for now)

include/circt/Dialect/Verif/VerifOps.td
uenoku, line 206: This makes `num_regs` redundant since it's always `num_regs == initialValues.size()`. Could you remove and add `iint32_t getNumRegs() { return getInitialValues.size(); }` to external declaration of the op?

uenoku:
I agree with @maerhart. I was expecting initi region's output values correspond to initial values of registers so adding `initial_values` as ODS defined attribute seems a bit weird to me. My understanding is that you wanted to store the externalized register's initial values somewhere on the module for the later lowering to init region, is that correct? In that case maybe adding `initial_values` as discardable might make sense. I think you can remove it when you actually lower `initial_values` to init region.

TaoBi22:
> My understanding is that you wanted to store the externalized register's initial values somewhere on the module for the later lowering to init region, is that correct?

The intention here is that the attribute stays on the op - the init region is specifically for initializing clocks, and the initial_values attribute is consumed by VerifToSMT to construct the initialization logic in the emitted bounded model checking program. 

Happy to be overruled and just add them to the init region now though! It would just require an extra PR to add lowerings for the concrete/symbolic input ops in verif and probably grow the diff here so I can change VerifToSMT at the same time so nothing breaks

TaoBi22:
@uenoku would you be happy for this to be merged as-is given the above?

uenoku (approved):
Yes, this looks good to me!

PR #7736, opened 2024-10-26
Use properties for attributes for many dialects

youngar:
This change enables the use of operation properties for more dialects in CIRCT. The option to *not* use properties is going to be removed in a future release of MLIR, but as well we expect this change to bring a performance improvement.

For all these dialects, the only change required was to properly infer return types.  Some operations were hardcoded to search through the attributes, and the attribute was moved to the property storage. By using the generated adaptor classes, we can abstract over where the interesting attribute is stored.

For future work, some code may be refactored for further performance wins now that we are using properties. In particular usages `setAttr` and `getAttr` may be made faster by using a less generic op transformation interface.

### DIFF
--- a/include/circt/Dialect/Arc/ArcDialect.td
+++ b/include/circt/Dialect/Arc/ArcDialect.td
@@ -22,9 +22,6 @@ def ArcDialect : Dialect {
   let hasConstantMaterializer = 1;
   let useDefaultTypePrinterParser = 1;
 
-  // Opt-out of properties for now, must migrate by LLVM 19.  #5273.
-  let usePropertiesForAttributes = 0;
-
   let extraClassDeclaration = [{
     void registerTypes();
   }];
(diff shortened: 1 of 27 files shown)


uenoku (approved):
Looks awesome!

youngar:
Updated to include all dialects but FIRRTL, it's going to take more work than everything else put together 😂 .

I launched a full nightly integration test on this, just to be sure. It is running [here]( https://github.com/llvm/circt/actions/runs/11529228901).

Testing this on a `FIRRTL` flow, seems like it doesn't affect most passes, but these two stand out: (this is `new/old`)
LowerFIRRTLToHW -> 51% 
VerifyObjectFields -> 74%

uenoku:
That's really impressive!

teqdruid:
I could've sworn we had this turned on for more dialects... I think our of the reasons you don't see more speedup is since it was already turned on for the comb and seq dialects. I suspect you'll see much more when you do this for the firrtl dialect.

Thanks!

PR #7739, opened 2024-10-26
[AIG] Add LowerVariadic and LowerWordToBits passes

uenoku:
This PR adds two basic legalization pass that lowers multibit variadic AndInverter to a single bit binary AndInverter. Once after these lowerings were applied it's trivial to interact with traditional AIG representation.  

This PR is extracted from https://github.com/llvm/circt/pull/7717


lib/Dialect/AIG/Transforms/LowerWordToBits.cpp
```diff
@@ -0,0 +1,103 @@
...
+          // `comb.concat` operands are single-bit.
+          if (concat.getNumOperands() == width &&
+              llvm::all_of(concat.getOperandTypes(), [](Type type) {
+                return type.getIntOrFloatBitWidth() == 1;
+              })) {
+            // Be careful with the endianness here.
+            operands.push_back(concat.getOperand(width - i - 1));
+            continue;
+          }
+        }
```
uenoku, line 64: @maerhart  I addressed the comment https://github.com/llvm/circt/pull/7717#discussion_r1815207000 here. Thank you for the suggestion!

PR #7743, opened 2024-10-28
[AIG] Add CutOp

uenoku:
The `aig.cut` operation represents a cut in the And-Inverter-Graph. This operation is variadic and can take multiple inputs and outputs, which corresponds to the input and output edges in AIG conceptually.

This is separated from https://github.com/llvm/circt/pull/7717 and a verifier is added.


uenoku:
I'm just going to merge this as this is reviewed on https://github.com/llvm/circt/pull/7717.

PR #7744, opened 2024-10-28
[circt-lec] Register Verif dialect

uenoku:
verif.assume can be used to annotate pre-condtion directly in the IR. This is quite useful when verifying transformation which involves undefined values. For example when lowering comb.divu to AIG, the conversion pass could choose arbitary values for divided by zero situation. So we cannot verify the equivalence unless we specify the pre-condtion with verif.assume


maerhart (approved):
LGTM! Thanks!

If we want to make assumptions about the inputs of the modules to be checked for equivalence, we should probably encode this in the `verif.lec` operation such that the same assumptions are made for both modules. But this PR is a necessary step towards that.

PR #7746, opened 2024-10-28
[OM] Add ClassOp region verifier

seldridge:
Add a region verifier to OM dialect's Class Op.  This verifies that the terminator returns the right number of fields with the correct types that match the declared type of the Class Op.

Fixes #7736.


mikeurbach (commented):
Thanks!

PR #7747, opened 2024-10-28
[FIRRTL] Remove validation from type inference code

youngar:
In FIRRTL, all expression operations have a `validate` hook which can be used to check operands and attributes before passing them in to `inferResultTypes`.  FIRParser will verify the same invariants right before calling `validate`, making it redundant.  Since `validate` is only called from the parser, we can delete the entire hook.


youngar:
This is where the operand an attr count are verifier in the parser before calling validate:
https://github.com/llvm/circt/blob/f503add0a60e5c9abe8d787860683cbac02a434f/lib/Dialect/FIRRTL/Import/FIRParser.cpp#L2393-L2426

fabianschuiki (approved):
LGTM! I don't remember why we had that separate validator in the first place :thinking:. This feels a lot cleaner.

PR #7748, opened 2024-10-29
 [OM] Add a new API to update fields of a ClassOp 

prithayan:
Add a new API to update the fields of an existing `FieldsOp` in a class.
This also renames the `addFields` to ensure it is used to add a new Fields Op only.


leonardt (approved):
LGTM

PR #7749, opened 2024-10-29
Bump LLVM to 8193832fb988e3df1e8e726634783805dca8d9b6.

mikeurbach:
This is mostly a straightforward integration of LLVM. One minor change was that DialectConversion materializations have a slightly altered signature since
https://github.com/llvm/llvm-project/commit/f18c3e4e7335df282c468b6dff3d29be1822a96d.

The commit message includes this, but now we return `Value()` in the case where we previously returned `std::nullopt`. Otherwise, this should behave the same.

### DIFF
--- a/lib/Conversion/DCToHW/DCToHW.cpp
+++ b/lib/Conversion/DCToHW/DCToHW.cpp
@@ -111,29 +111,27 @@ public:
   ESITypeConverter() {
     addConversion([](Type type) -> Type { return toESIHWType(type); });
     addConversion([](esi::ChannelType t) -> Type { return t; });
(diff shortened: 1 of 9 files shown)


mikeurbach:
The diff is annoying, because the changes caused clang format to re-flow some functions... it's just `std::optional<Value>` -> `Value` and `std::nullopt` -> `Value()`.

uenoku (approved):
LGTM, I always wondered why it was using `std::optional` instead of `Value`. Glad that it got more consistent!

Dinistro (approved):

PR #7750, opened 2024-10-30
[FIRRTL] enable properties for inherent attributes

youngar:
[[FIRRTL] Clean up inferReturnTypes implementation](https://github.com/llvm/circt/pull/7750/commits/739cf51aa467e0fe6976f234902368fe64e8d20c) 
[739cf51](https://github.com/llvm/circt/pull/7750/commits/739cf51aa467e0fe6976f234902368fe64e8d20c)

This is a big change to clean up our implementation of
`inferReturnTypes` in an effort to get ourselves ready to enable
properties in the FIRRTL dialect. We had structured our API so that
`inferReturnType` would call out to a simpler version of itself with
some of the useless arguments removed.  This prevented us from using op
adaptors to abstract over whether or not an inherent attribute was
contained inside the attr-dict or the properties.

This change keeps the old structure of a two level API, but with
different boundaries: The large API takes all arguments, creates an
adapter, pulls out the necessary attributes and then calls in to the
simpler interface. FIRParser uses the simpler API when inferring return
types.  The simpler interface is now specific to each operation and not
common with other operations.

This last change caused a problem for `parsePrimExp`, which relied on a
generic interface to create all expressions. This function is now
templated over exactly how many arguments the specific prim op takes,
parses exactly that many, and splats them out when calling
`inferReturnType`.  As an upside to this, we can also call a more
specific builder for each operation, which should speed up building
operations when we move to properties.

---

[[FIRRTL] enable properties for inherent attributes](https://github.com/llvm/circt/pull/7750/commits/f38c9b291b9efb035522027a983bc1b21d605f12) 
[f38c9b2](https://github.com/llvm/circt/pull/7750/commits/f38c9b291b9efb035522027a983bc1b21d605f12)
This enables properties in the FIRRTL dialect. Some op parsers which
called `inferReturnTypes` had to be updated to store inherent attributes
as properties on the OperationState, as the call to `inferReturnTypes`
expects properties to have already been populated.  This is a change
that we should make for all operation parsers and builders, to make
things faster, but is not strictly necessary at this time.


youngar:
On a `firtool` run, the change in performance is pretty close to a wash.  Most passes got a little slower, some had bigger changes. This is `newtime/oldtime` again:
PrefixModules: %128
DropName: %36
Will do some profiling to see what is so slow.

fabianschuiki (approved):
Really cool to see this migrated over to properties. LGTM!

rwy7 (approved):

youngar:
Thanks for taking a look!

PR #7751, opened 2024-10-30
[FIRRTL] Add a new FIRRTL annotation to specify type lowering behavior of module body

uenoku:
Add a new annotation to control type lowering behavior for internal signals
within a module, separate from the port convention. This allows more fine-grained
control over how aggregate types are handled inside modules.

The new annotation works similarly to ConventionAnnotation but applies to
internal signals rather than module ports. It supports the same conventions
and includes an 'includeHierarchy' option to apply the setting to all
modules in the hierarchy.


lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -2056,12 +2055,26 @@ void ClassOp::insertPorts(ArrayRef<std::pair<unsigned, PortInfo>> ports) {
...
+Convention FModuleOp::getPortConvention() {
+  return getConventionAttr().getPortConvention();
+}
+
+Convention FModuleOp::getBodyConvention() {
+  return getConventionAttr().getBodyConvention();
+}
+
+Convention FExtModuleOp::getPortConvention() {
+  return getConventionAttr().getPortConvention();
```
uenoku, line 2067: These must go to appropriate places.

fabianschuiki (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
```diff
@@ -298,13 +301,41 @@ static LogicalResult applyConventionAnno(const AnnoPathValue &target,
 
   auto convention = *conventionOpt;
 
+  if (convention == Convention::Internal)
+    // Convention is internal by default so there is nothing to change
+    return success();
+
+  auto includeHierarchy = anno.getAs<BoolAttr>("includeHierarchy");
```
seldridge, line 308: `tryGetAs<BoolAttr>` and checking for failure would be safer.

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
```diff
@@ -298,13 +301,41 @@ static LogicalResult applyConventionAnno(const AnnoPathValue &target,
...
+  auto conventionAttr = ConventionAttr::get(op->getContext(), convention);
+  auto setConvention = [&](Operation *moduleOp) {
+    TypeSwitch<Operation *>(moduleOp)
+        .Case<FModuleOp, FExtModuleOp>([&](auto moduleOp) {
+          if (IsConventionAnno)
+            moduleOp.setConventionAttr(conventionAttr);
+          else
+            moduleOp->setDiscardableAttr("body_type_lowering", conventionAttr);
```
seldridge, line 316: Nice. I like using a discardable attribute here. This is _not_ part of any ABI, so we can just throw it away or ignore it if we want. 👍

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
```diff
@@ -298,13 +301,41 @@ static LogicalResult applyConventionAnno(const AnnoPathValue &target,
...
+        })
+        .Default([](auto) {});
+  };
```
seldridge, line 319: The `TypeSwitch` is kind of overkill here. Is there a trait that this can be cast to instead?

Unrelated, the checking earlier in this function does not guarantee that this is only applied to an `FModuleOp` or FExtModuleOp`. It only checks for `OpAnnoTarget` and then incorrectly says that is a module.

This also does bring up a point of if the internal convention should be allowed to be applied to external modules. It seems like _no_. However, that would mean that duplication as opposed to templating may be better.

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
```diff
@@ -298,13 +301,41 @@ static LogicalResult applyConventionAnno(const AnnoPathValue &target,
...
+      // the hierarchy.
+      for (auto *node :
+           llvm::post_order(state.instancePathCache.instanceGraph[moduleOp])) {
+        if (node && isa<FModuleOp, FExtModuleOp>(*node->getModule()))
+          setConvention(node->getModule());
+      }
+    } else {
+      // Update the convention.
+      setConvention(moduleOp);
+    }
```
seldridge, line 333: Fanning out the convention here seems pretty good. One weirdness that this will create is that doesn't this mean that you can get different behaviors for different annotation orders? Maybe that's okay, just perhaps surprising.

test/Dialect/FIRRTL/lower-types.mlir
```diff
@@ -1405,6 +1405,45 @@ firrtl.circuit "UnrealizedConversion" {
...
+  // AGGREGATE-SAME: %input: !firrtl.vector<uint<8>, 1>
+  // AGGREGATE-NEXT: firrtl.reg
+  // AGGREGATE-SAME: !firrtl.uint<8>
+  firrtl.module private @Conventions4(in %input: !firrtl.vector<uint<8>, 1>, in %clk: !firrtl.clock, out %port: !firrtl.vector<uint<8>, 1>) attributes {convention = #firrtl<convention internal>, body_type_lowering = #firrtl<convention scalarized>}{
+    %r = firrtl.reg interesting_name %clk : !firrtl.clock, !firrtl.vector<uint<8>, 1>
+    firrtl.matchingconnect %r, %input : !firrtl.vector<uint<8>, 1>
+    firrtl.matchingconnect %port, %r : !firrtl.vector<uint<8>, 1>
+  }
+}
+
```
seldridge, line 1446: Nice, exhaustive test. 👍

seldridge (commented):
The `includeHierarchy` field seems unnecessary. I also think this is unnecessary for the `Convention` annotation. My reasoning here is that convention is really supposed to "Port Lowering ABI". This is something that is only defined on public modules and everything else is up to the compiler to decide.

Is there a request for `includeHierarchy` and/or is there a way that we could avoid this and handle it entirely from Chisel, e.g., add some builder state which indicates what the convention is within a scope and then apply it to all modules under that scope?

That said, I am fine with delaying this for now as long as this never becomes a part of an implied ABI that a user relies on.

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
uenoku: There is priority between `scalarized` and `internal` and `scalarized` always wins. So the order doesn't affect the final result.

uenoku:
> The includeHierarchy field seems unnecessary. I also think this is unnecessary for the Convention annotation. My reasoning here is that convention is really supposed to "Port Lowering ABI". This is something that is only defined on public modules and everything else is up to the compiler to decide.

Fair point, I'm totally fine to drop includeHierachy for Convention annotation. For body type lowering, it's useful to annotate the annotation for design module and ensure that module bodies in the design is scalarized.

> add some builder state which indicates what the convention is within a scope and then apply it to all modules under that scope?

That would be great. If we can mark every module in design with this annotation in Chisel, we really don't need includeHierachy. 

There will never be no ABI change regarding body lowering. The intention of the annotation is more like per-module compiler option used in LowerTypes.

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
seldridge: Got it. That is at least stable. 👍

uenoku:
I updated to remove `includeHierarchy` from convention attr and clean up the annotation application. Body type lowering annotation still has `includeHierarchy` since it looks not worth modifying chisel internal for this specific annotation.

lib/Dialect/FIRRTL/Transforms/LowerAnnotations.cpp
uenoku: >  Is there a trait that this can be cast to instead?

I thought FModuleLike was meant to be the interface for that, but I'll just keep the current logic since FModuleLike is implemented for Class or MemModule and the behavior is unclear.  

> This also does bring up a point of if the internal convention should be allowed to be applied to external modules. It seems like no. However, that would mean that duplication as opposed to templating may be better.

Thanks, I changed to duplicate the handler.

lib/Dialect/FIRRTL/Transforms/LowerTypes.cpp
seldridge, line 1: What is interesting about these changes is that we could move in a direction of removing the CLI options for aggregate preservation and instead rely on the annotations / attributes. As a middle ground, it may be better to change the CLI (in a follow-on) to apply the annotations or do it as a part of parsing.

This is more of a thought than a definite direction we should go.

seldridge (approved):
LGTM

Issue #7628, opened 2024-09-25
[MooreToCore] Support string constants


fabianschuiki:
Add a conversion for the `moore.string_constant` op to MooreToCore: https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+11.10+string_bit_array

jpinot:
Hey @fabianschuiki, I'd like to give it a shot! Could you assign me?

fabianschuiki:
Sure thing! Thanks a lot for taking a stab at this :smiley:. Let me know if you run into any trouble.

jpinot:
Hi @fabianschuiki! Sorry for the delayed response—it took me a bit to dive into the CIRCT fundamentals to get oriented :smiley_cat:. To tackle this, I referenced the [Moore test results](https://chipsalliance.github.io/sv-tests-results/?v=moore+11.10+string_bit_array) and tried [converting `moore::StringConstantOp` to `hw::ConstantOp` by translating `StringRef` into a bit vector](https://github.com/jpinot/circt/commit/a1747a3d59dba8124b4f820c358c87d3168511d9), also [I tryed using hw::ArrayCreateOp](https://github.com/jpinot/circt/commit/67c07c95a94cab240485d5ec0659aee71d3e79ca). Here’s the resulting LLHD:

```sh
$ ./bin/circt-verilog string_bit_array.sv
module {
  hw.module @top() {
    %0 = llhd.constant_time <0ns, 0d, 1e>
    %c0_i80 = hw.constant 0 : i80
    %c1415934836_i32 = hw.constant 1415934836 : i32
    %c0_i112 = hw.constant 0 : i112
    %a = llhd.sig %c0_i112 : i112
    llhd.process {
      %1 = comb.concat %c0_i80, %c1415934836_i32 : i80, i32
      llhd.drv %a, %1 after %0 : !hw.inout<i112>
      llhd.halt
    }
    hw.output
  }
}
```

It seems that this module should work as expected since it represents a constant string, which shouldn’t be modified. Still, I’d really appreciate your thoughts on (a) what the expected correct output should look like to confirm I’m on track, and (b) whether there might be a cleaner way to handle the `StringRef` to bit vector conversion. So far, I’ve only managed to iterate through the string and concatenate the bytes, but I’d love to know if there’s a more elegant approach.

Thanks so much for your help guidance!

fabianschuiki:
[...]

PR #7752, opened 2024-10-30
[MooreToCore] Add support for moore::string_constant #7628

jpinot:
Lower `StringConstantOp` to `hw::ConstantOp`. Encodes strings as `APInt` with byte-width based on string length, enabling representation of Moore string constant as bit vectors.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -555,6 +555,34 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
   }
 };
 
+struct StringConstantOpConv : public OpConversionPattern<StringConstantOp> {
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(moore::StringConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    const auto str = op.getValue();
+    const unsigned byteWidth = str.size() * 8;
+    const auto resultType =
+        typeConverter->convertType(op.getResult().getType());
+    if (const auto intType = mlir::dyn_cast<IntegerType>(resultType)) {
+      if (intType.getWidth() < byteWidth) {
+        return rewriter.notifyMatchFailure(op,
+                                           "invalid string constant type size");
+      }
+    } else {
+      return rewriter.notifyMatchFailure(op, "invalid string constant type");
+    }
+    APInt value(byteWidth, 0);
+    for (size_t i = 0; i < str.size(); ++i) {
+      const auto asciiChar = static_cast<uint8_t>(str[i]);
+      value |= APInt(byteWidth, asciiChar) << (8 * (str.size() - 1 - i));
+    }
+    rewriter.replaceOpWithNewOp<hw::ConstantOp>(
+        op, resultType, rewriter.getIntegerAttr(resultType, value));
+    return success();
+  }
+};
+
 struct ConcatOpConversion : public OpConversionPattern<ConcatOp> {
   using OpConversionPattern::OpConversionPattern;
   LogicalResult
@@ -1477,7 +1505,7 @@ static void populateOpConversion(RewritePatternSet &patterns,
     ConversionOpConversion, ReadOpConversion,
     StructExtractOpConversion, StructExtractRefOpConversion,
     ExtractRefOpConversion, StructCreateOpConversion, ConditionalOpConversion,
-    YieldOpConversion, OutputOpConversion,
+    YieldOpConversion, OutputOpConversion, StringConstantOpConv,
 
     // Patterns of unary operations.
     ReduceAndOpConversion, ReduceOrOpConversion, ReduceXorOpConversion,
--- a/test/Conversion/MooreToCore/basic.mlir
+++ b/test/Conversion/MooreToCore/basic.mlir
@@ -868,3 +868,14 @@ moore.module @Assert(in %cond : !moore.l1)  {
   moore.return
   }
 }
+
+// CHECK-LABEL: hw.module @StringConstant
+moore.module @StringConstant() {
+  moore.procedure initial {
+    // CHECK: hw.constant 1415934836 : i32
+    %str = moore.string_constant "Test" : i32
+    // CHECK: hw.constant 0 : i0
+    %str_empty = moore.string_constant "" : i0
+    moore.return
+  }
+}
--- a/test/Dialect/Moore/basic.mlir
+++ b/test/Dialect/Moore/basic.mlir
@@ -302,6 +302,11 @@ moore.module @Expressions(
   // CHECK: moore.struct_inject [[STRUCT1]], "a", [[B]] : struct<{a: i32, b: i32}>, i32
   moore.struct_inject %struct1, "a", %b : struct<{a: i32, b: i32}>, i32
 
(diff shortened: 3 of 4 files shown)


test/Dialect/Moore/basic.mlir
```diff
@@ -302,6 +302,11 @@ moore.module @Expressions(
   // CHECK: moore.struct_inject [[STRUCT1]], "a", [[B]] : struct<{a: i32, b: i32}>, i32
   moore.struct_inject %struct1, "a", %b : struct<{a: i32, b: i32}>, i32
 
+  // CHECK: moore.string_constant "Test" : i128
+  moore.string_constant "Test" : i128
+  // CHECK: moore.string_constant "" : i128
+  moore.string_constant "" : i128
+
```
fabianschuiki, line 309: Thanks for adding these! Can you also add a similar thing to the MooreToCore test, to check that the lowering works?

jpinot: Thank you for pointing out the correct location of the `MooreToCore` test. I've realized that my `StringConstantOpConv` conversion currently ignores the type of the resulting value, which shouldn’t be the case. For example:

```mlir
%str = moore.string_constant "Test" : i2
```

is currently converted to:

```mlir
%c1415934836_i32 = hw.constant 1415934836 : i32
```

when compiled with `circt-opt` using the `--convert-moore-to-core` flag. I'll update the conversion to ensure the resulting type is set correctly.

PD: In this example, I'm uncertain about the expected outcome: should it convert to `hw.constant 1 : i2`, or should it fail to convert?

fabianschuiki: Good question. Either should work. It might be a bit better to fail the conversion. The string being larger than the type should be caught in the frontend already, so the user should never see this error. But erroring out is good to catch compiler bugs 🙂

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -555,6 +555,27 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    const auto str = op.getValue();
+    const unsigned byteWidth = str.size() * 8;
+    Type resultType = typeConverter->convertType(op.getResult().getType());
+    resultType.dump();
+    APInt value(byteWidth, 0);
+    for (size_t i = 0; i < str.size(); ++i) {
+      const auto asciiChar = static_cast<uint8_t>(str[i]);
+      value |= APInt(byteWidth, asciiChar) << (8 * (str.size() - 1 - i));
+    }
```
uenoku, line 571: Maybe we can use `getAsInteger`? https://llvm.org/doxygen/classllvm_1_1StringRef.html#adc9c54d457a9ebc9d909c4b2234c7657

fabianschuiki: I think `getAsInteger` converts the APInt to a string representation of the number. SystemVerilog wants you to take the actual ASCII byte values of each character and concatenate them into a large integer. So it's basically going from a string to an integer, but not by parsing the string as a number like "0x42abc", but by actually taking the ASCII values of the characters and concatenating them 😁. It's pretty wild.

test/Conversion/MooreToCore/basic.mlir
```diff
@@ -868,3 +868,15 @@ moore.module @Assert(in %cond : !moore.l1)  {
...
   }
 }
+
+// CHECK-LABEL: hw.module @StringConstant
+moore.module @StringConstant() {
+  moore.procedure initial {
+    // CHECK: hw.constant 1415934836
+    %str = moore.string_constant "Test" : i2
+    // CHECK: hw.constant 0
+    %str_empty = moore.string_constant "" : i128
```
fabianschuiki, line 878: Maybe add the `: i2` and `: i128` type suffices to the CHECK lines, to make sure the conversion does the right thing.

lib/Conversion/MooreToCore/MooreToCore.cpp
uenoku: Oh, yeah. Thank you for pointing it out 👍

test/Dialect/Moore/basic.mlir
hailongSun2000: Thank @jpinot for your work on `StringConstantOpConv` :partying_face:! @fabianschuiki is all right! Slang can guarantee the correct type width. And You're able to don't worry about the type conversion. Because @maerhart has already handled the related error on https://github.com/llvm/circt/pull/7631. I think maybe you should tweak your test from 
```
%str = moore.string_constant "Test" : i2
%str_empty = moore.string_constant "" : i128
``` 
to 
```
%str = moore.string_constant "Test" : i32
%str_empty = moore.string_constant "" : i8
```

jpinot: Thank you, @fabianschuiki and @hailongSun2000, for taking the time to review my PR. I've made the following updates based on your feedback:

- Adjusted the test to match type with string size.
- Added checks to validate that the type and type size are smaller or equal than the string size.
- Included a new test case to verify this functionality.

Additionally, I noticed an issue with the run command in the test file. I corrected it in a [separate commit](https://github.com/llvm/circt/pull/7752/commits/8aa4794e254680e081daee47c510a91789cf452a). Please let me know if I misunderstood  or if you prefer this change to be in a separate PR.

test/Conversion/MooreToCore/basic.mlir
jpinot: Done. Corrected types as mentioned [above](https://github.com/llvm/circt/pull/7752#discussion_r1825464079) and added type suffixes to the `CHECK` lines.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -555,6 +555,35 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
+      }
+    } else {
+      return rewriter.notifyMatchFailure(op, "invalid string constant type");
+    }
+    APInt value(byteWidth, 0);
+    for (size_t i = 0; i < str.size(); ++i) {
+      const auto asciiChar = static_cast<uint8_t>(str[i]);
+      value |= APInt(byteWidth, asciiChar) << (8 * (str.size() - 1 - i));
+    }
+    const auto type = IntegerType::get(rewriter.getContext(), byteWidth);
```
fabianschuiki, line 580: I think you'll have to use the `resultType`'s width here to avoid converting the `moore.string_constant` op to an `hw.constant` of a different return type.

fabianschuiki (approved):
LGTM besides a minor nit on the `hw.constant` return type 👍 Thanks a lot for doing this work!

lib/Conversion/MooreToCore/MooreToCore.cpp
jpinot: Done, updated type to `resultType`. Good catch :)

fabianschuiki:
Really nice! @jpinot If you want, feel free to [get commit access to CIRCT](https://llvm.org/docs/DeveloperPolicy.html#id20) to merge the PR yourself, or let me know if I should land it for you.

hailongSun2000 (approved):
LGTM! :100:

test/Dialect/Moore/basic.mlir
hailongSun2000: You're right! Thanks for catching this detail.

PR #7754, opened 2024-10-31
[Verif][NFC] Use auto-generated constructors for all passes

fabianschuiki:
Drop the `let constructor` line from the Verif pass definitions. This causes TableGen to automatically generate the constructors for these passes, and also handle pass options if we ever decide to add them.

This requires adding a `...Pass` to the end of each pass def to produce a constructor name of the form `create...Pass()`.

### DIFF
--- a/include/circt/Dialect/Verif/Passes.td
+++ b/include/circt/Dialect/Verif/Passes.td
@@ -1,4 +1,4 @@
-//===-- Passes.td - Verif pass definition file ----------------*- tablegen -*-===//
+//===-- Passes.td - Verif pass definition file -------------*- tablegen -*-===//
 //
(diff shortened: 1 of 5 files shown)


include/circt/Dialect/Verif/Passes.td
```diff
@@ -1,4 +1,4 @@
-//===-- Passes.td - Verif pass definition file ----------------*- tablegen -*-===//
+//===-- Passes.td - Verif pass definition file -------------*- tablegen -*-===//
```
seldridge, line 1: 💯

seldridge (approved):
LGTM

Thanks for this cleanup. This would be great to roll out everywhere.

PR #7755, opened 2024-10-31
[circt-test] Add test discovery classes and `-l` option

fabianschuiki:
Add the `TestSuite` and `Test` classes to represent the discovered tests. Instead of directly printing ops to the output, first traverse the provided MLIR module and populate a test suite object with the discovered tests. Then print the list of tests in the test suite.

A separate test discovery step will later allow us to discover tests via other means, for example, by scanning an entire Verilog directory for marker comments.

### DIFF
--- a/include/circt/Dialect/Verif/Passes.td
+++ b/include/circt/Dialect/Verif/Passes.td
@@ -1,4 +1,4 @@
-//===-- Passes.td - Verif pass definition file ----------------*- tablegen -*-===//
+//===-- Passes.td - Verif pass definition file -------------*- tablegen -*-===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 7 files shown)


include/circt/Dialect/Verif/Passes.td
```diff
@@ -15,40 +15,30 @@
...
 include "mlir/Pass/PassBase.td"
 
-/**
- * Performs preparation transformations for formal verification.
- * This includes things like flattening wires, removing unnecessary information
- * such as OM Classes, handles formal contracts and transforms module instances.
- */
- def PrepareForFormal : Pass<"prepare-for-formal", "hw::HWModuleOp"> {
-  let summary = "Prepares a given top-level circuit for formal verification.";
+def PrepareForFormalPass : Pass<"prepare-for-formal", "hw::HWModuleOp"> {
```
uenoku, line 18: nit: MLIR/CIRCT don't usually give "Pass" suffix to tablegen definition so it would be more consistent to keep that.

tools/circt-test/circt-test.cpp
```diff
@@ -50,58 +50,122 @@ namespace {
...
+class Test {
+public:
+  /// The name of the test. This is also the name of the top-level module passed
+  /// to the formal or simulation tool to be run.
+  StringAttr name;
+  /// The kind of test, such as "formal" or "simulation".
+  TestKind kind;
+  /// An optional location indicating where this test was discovered. This can
+  /// be the location of an MLIR op, or a line in some other source file.
+  LocationAttr loc;
```
uenoku, line 89: nit: `op.getLoc()` return `Location` which a non-null wrapper around LocationAttr so maybe just use `Location` here.

uenoku (approved):
LGTM

include/circt/Dialect/Verif/Passes.td
fabianschuiki: The problem is that the constructors are then called `createPrepareForFormal`, which can be really strange if you have something like a "partition pass", which would create a constructor `createPartition`.

tools/circt-test/circt-test.cpp
fabianschuiki: Good point. I think there will be cases in the future where we don't have a location. For example, if we discover tests by looking for something like "// RUN: ..." lines in SystemVerilog files, we may not have a reliable location. Or maybe we do?

PR #7093, opened 2024-05-27
Flatten memref GetGlobal and Global operations

jiahanxie353:
This PR supports flattening of `memref::GetGlobalOp` and `memref::GlobalOp`.

Test case not yet included because I wasn't sure which script to use to auto update test mlir files.

PR #7756, opened 2024-10-31
[circt-test] Add simple SymbiYosys test runner

fabianschuiki:
Extend circt-test to export the input MLIR file as Verilog and then run all discovered tests through a simple SymbiYosys runner script. This is currently very rigid: all tests are always run, and they all run through the exact same runner script. In the future, we'll want to add filtering mechanisms to include and exclude tests, and pick among multiple runner scripts for each test.

Also add a basic integration test that builds an adder from discrete logic gates and then verifies it through multiple `verif.formal` ops.

To run this you need to have SymbiYosys installed. In the future we'll want to have `circt-bmc` and other CIRCT-based verification tools available as runners as well.

### DIFF
--- a/integration_test/circt-test/basic.mlir
+++ b/integration_test/circt-test/basic.mlir
@@ -1,0 +1,115 @@
+// RUN: circt-test %s -r circt-test-runner-sby.py | FileCheck %s
+// REQUIRES: sby
+
+// CHECK: all 6 tests passed
+
+hw.module @FullAdder(in %a: i1, in %b: i1, in %ci: i1, out s: i1, out co: i1) {
+  %0 = comb.xor %a, %b : i1
+  %1 = comb.xor %0, %ci : i1
+  %2 = comb.and %a, %b : i1
+  %3 = comb.and %0, %ci : i1
+  %4 = comb.or %2, %3 : i1
+  hw.output %1, %4 : i1, i1
+}
+
+hw.module @CustomAdderWithCarry(in %a: i4, in %b: i4, in %ci: i1, out z: i4, out co: i1) {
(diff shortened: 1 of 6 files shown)


mikeurbach (commented):
This is pretty neat!

tools/circt-test/circt-test.cpp
```diff
@@ -193,6 +213,104 @@ static LogicalResult execute(MLIRContext *context) {
...
+  pm.addPass(createExportVerilogPass(verilogFile->os()));
+  if (failed(pm.run(*module)))
+    return failure();
+  verilogFile->keep();
+
+  // Find the runner binary in the search path. Otherwise assume it is a binary
+  // we can run as is.
+  std::string runner = opts.runner;
+  if (auto result = llvm::sys::findProgramByName(runner))
+    runner = result.get();
```
uenoku, line 249: nit: Return failure otherwise?

uenoku (approved):
Looks great!

integration_test/circt-test/basic.mlir
```diff
@@ -0,0 +1,113 @@
+// RUN: circt-test %s -r circt-test-runner-sby.py
+// REQUIRES: sby
```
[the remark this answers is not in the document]
fabianschuiki: Ah cool idea, definitely!

tools/circt-test/circt-test.cpp
fabianschuiki: Good idea, thanks!

fabianschuiki:
Landing this since the only failure is related to #7093.

PR #7757, opened 2024-10-31
[HW] Add port name accessors to HWInstanceLike

maerhart:
HWInstanceLike is currently 'empty' and thus essentially the same as InstanceGraphInstanceOpInterface. On the other hand, HWModuleLike deals with port names and hw::InstanceOp caches them and verifies they match the referenced module's port names. This PR basically makes this behavior part of what it means to be an HWInstanceLike. If one has an instance operation that does not deal with port names, it's still possible to just make it an InstanceGraphInstanceOpInterface directly


uenoku (approved):
Looks great to me!

fabianschuiki (approved):
LGTM

PR #7759, opened 2024-10-31
[LTL] Canonicalize ltl.and to comb.and for i1 properties

fabianschuiki:
Extend `ltl.and`, `ltl.or`, and `ltl.intersect` to infer their return type as `i1` if all operands are `i1`. Also add a canonicalization method that replaces these ops with `comb.and` and `comb.or` if they operate entirely on `i1`.


dobios (approved):
This is a great idea! I will definitely makes things easier to lower into a potential future form of the ltl dialect! LGTM!

PR #7760, opened 2024-10-31
[FIRRTL] Fix folding of when conditions into LTL properties

fabianschuiki:
Make ExpandWhen's folding of the when condition into LTL properties work for more inputs. This used to be very brittle and the condition would be mapped to a new implication around the entire property:

    when (a):
      assert(@(posedge clock) b |-> c)

used to become

    assert(a |-> @(posedge clock) b |-> c)

This commit improves on this by always pushing implications behind clocks, and by merging implications like `a |-> b |-> c` into a single `(a & b) |-> c`. As a result, ExpandWhens now produces the following output, which is more in line with what users expect:

    assert(@(posedge clock) a & b |-> c)

Also add a FIR-to-Verilog test to nail this behavior down and ensure we don't regress on user expectations.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
@@ -551,24 +551,82 @@ private:
 
   /// Concurrent and of a property with the current condition.  If we are in
   /// the outer scope, i.e. not in a WhenOp region, then there is no condition.
-  Value ltlAndWithCondition(Operation *op, Value value) {
-    // 'ltl.and' the value with the current condition.
-    return OpBuilder(op).createOrFold<LTLAndIntrinsicOp>(
-        condition.getLoc(), condition.getType(), condition, value);
+  Value ltlAndWithCondition(Operation *op, Value property) {
+    // Look through nodes.
+    while (auto nodeOp = property.getDefiningOp<NodeOp>())
+      property = nodeOp.getInput();
+
+    // Look through `ltl.clock` ops.
+    if (auto clockOp = property.getDefiningOp<LTLClockIntrinsicOp>()) {
+      auto input = ltlAndWithCondition(op, clockOp.getInput());
+      auto &newClockOp = createdLTLClockOps[{clockOp, input}];
+      if (!newClockOp) {
+        newClockOp = OpBuilder(op).cloneWithoutRegions(clockOp);
+        newClockOp.getInputMutable().assign(input);
+      }
+      return newClockOp;
+    }
+
+    // Otherwise create a new `ltl.and` with the condition.
+    auto &newOp = createdLTLAndOps[{condition, property}];
+    if (!newOp)
+      newOp = OpBuilder(op).createOrFold<LTLAndIntrinsicOp>(
+          condition.getLoc(), property.getType(), condition, property);
+    return newOp;
   }
 
   /// Overlapping implication with the condition as its antecedent and a given
   /// property as the consequent.  If we are in the outer scope, i.e. not in a
   /// WhenOp region, then there is no condition.
-  Value ltlImplicationWithCondition(Operation *op, Value value) {
-    // 'and' the value with the current condition.
-    return OpBuilder(op).createOrFold<LTLImplicationIntrinsicOp>(
-        condition.getLoc(), condition.getType(), condition, value);
+  Value ltlImplicationWithCondition(Operation *op, Value property) {
+    // Look through nodes.
+    while (auto nodeOp = property.getDefiningOp<NodeOp>())
+      property = nodeOp.getInput();
+
+    // Look through `ltl.clock` ops.
+    if (auto clockOp = property.getDefiningOp<LTLClockIntrinsicOp>()) {
+      auto input = ltlImplicationWithCondition(op, clockOp.getInput());
+      auto &newClockOp = createdLTLClockOps[{clockOp, input}];
+      if (!newClockOp) {
+        newClockOp = OpBuilder(op).cloneWithoutRegions(clockOp);
+        newClockOp.getInputMutable().assign(input);
+      }
+      return newClockOp;
+    }
+
+    // Merge condition into `ltl.implication` left-hand side.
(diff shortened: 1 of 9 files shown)


lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
```diff
@@ -551,24 +551,82 @@ class WhenOpVisitor : public LastConnectResolver<WhenOpVisitor> {
...
-    // 'ltl.and' the value with the current condition.
-    return OpBuilder(op).createOrFold<LTLAndIntrinsicOp>(
-        condition.getLoc(), condition.getType(), condition, value);
+  Value ltlAndWithCondition(Operation *op, Value property) {
+    // Look through nodes.
+    while (auto nodeOp = property.getDefiningOp<NodeOp>())
+      property = nodeOp.getInput();
+
+    // Look through `ltl.clock` ops.
+    if (auto clockOp = property.getDefiningOp<LTLClockIntrinsicOp>()) {
```
dobios, line 560: shouldn't this `if` be a `while`? I forget if we somewhere enforce the rule that you can't have 
```
%0 = ltl.clock ...
%1 = ltl.clock %1
...
```
because in that case this still wouldn't fold correctly no?

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
```diff
@@ -551,24 +551,82 @@ class WhenOpVisitor : public LastConnectResolver<WhenOpVisitor> {
...
-    // 'and' the value with the current condition.
-    return OpBuilder(op).createOrFold<LTLImplicationIntrinsicOp>(
-        condition.getLoc(), condition.getType(), condition, value);
+  Value ltlImplicationWithCondition(Operation *op, Value property) {
+    // Look through nodes.
+    while (auto nodeOp = property.getDefiningOp<NodeOp>())
+      property = nodeOp.getInput();
+
+    // Look through `ltl.clock` ops.
+    if (auto clockOp = property.getDefiningOp<LTLClockIntrinsicOp>()) {
```
dobios, line 587: same comment here

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
```diff
@@ -551,24 +551,82 @@ class WhenOpVisitor : public LastConnectResolver<WhenOpVisitor> {
...
-        condition.getLoc(), condition.getType(), condition, value);
+  Value ltlAndWithCondition(Operation *op, Value property) {
+    // Look through nodes.
+    while (auto nodeOp = property.getDefiningOp<NodeOp>())
+      property = nodeOp.getInput();
+
+    // Look through `ltl.clock` ops.
+    if (auto clockOp = property.getDefiningOp<LTLClockIntrinsicOp>()) {
+      auto input = ltlAndWithCondition(op, clockOp.getInput());
+      auto &newClockOp = createdLTLClockOps[{clockOp, input}];
```
dobios, line 562: why are you keeping track of these?

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
```diff
@@ -551,24 +551,82 @@ class WhenOpVisitor : public LastConnectResolver<WhenOpVisitor> {
...
+
+  /// The `ltl.and` operations that have been created.
+  SmallDenseMap<std::pair<Value, Value>, Value> createdLTLAndOps;
+
+  /// The `ltl.implication` operations that have been created.
+  SmallDenseMap<std::pair<Value, Value>, Value> createdLTLImplicationOps;
+
+  /// The `ltl.clock` operations that have been created.
+  SmallDenseMap<std::pair<LTLClockIntrinsicOp, Value>, LTLClockIntrinsicOp>
+      createdLTLClockOps;
```
dobios, line 629: I'm guessing you mean for these to persist, i.e. not clear them?

dobios (approved):
This sounds great! I'm just a bit confused by a few things you're doing, but otherwise this seems like a really good idea (the previous solution might also have been rejected in some implementations of SVA, since the spec mentions that antecedents of an OI can't be a property ^^" I'm surprised nobody ran into any issues

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
fabianschuiki: This should work also in that case, since there's a recursive call on the line just below. So the condition would be pushed all the way through any clocks and `firrtl.node`s :slightly_smiling_face:

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
fabianschuiki: This is just to reduce duplication. Since the condition is inserted _behind_ the clock, we basically have to copy the clock op since it might have other users that we can't modify. This has a tendency of creating redundant copies of the clock if a property is defined outside the `when` and reused for multiple asserts. The map here ensures that we only ever create a single copy per unique clock/input combination.

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
fabianschuiki: Recursion :superhero: :sweat_smile:

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
fabianschuiki: There's a fresh new `WhenOpVisitor` for each block, so these should start out empty for every when or else, and will then get dropped once we're done with the block.

fabianschuiki:
@dobios Yeah we really need to improve the way we describe sequences and properties :grin: This expression grammar with implicit clocks and what not is a big headache.

lib/Dialect/FIRRTL/Transforms/ExpandWhens.cpp
dobios: right I missread my bad ^^"

PR #7762, opened 2024-11-01
[Calyx] Sequential memory support floating point operations

jiahanxie353:
This patch supports floating point operations with SeqMemory in SCFToCalyx.

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -150,7 +150,7 @@ def MemoryOp : CalyxPrimitive<"memory", []> {
   );
 
   let results = (outs
-    Variadic<AnySignlessInteger>:$results
+    Variadic<AnyTypeOf<[AnySignlessInteger, AnyFloat]>>:$results
   );
 
   let assemblyFormat = [{
@@ -220,7 +220,7 @@ def SeqMemoryOp : CalyxPrimitive<"seq_mem", []> {
   );
 
   let results = (outs
-    Variadic<AnySignlessInteger>:$results
+    Variadic<AnyTypeOf<[AnySignlessInteger, AnyFloat]>>:$results
   );
 
   let assemblyFormat = [{
@@ -233,6 +233,12 @@ def SeqMemoryOp : CalyxPrimitive<"seq_mem", []> {
     OpBuilder<(ins
       "StringRef":$sym_name,
       "int64_t":$width,
+      "ArrayRef<int64_t>":$sizes,
+      "ArrayRef<int64_t>":$addrSizes
+    )>,
+    OpBuilder<(ins
+      "StringRef":$sym_name,
+      "Type":$type,
       "ArrayRef<int64_t>":$sizes,
       "ArrayRef<int64_t>":$addrSizes
     )>
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -232,12 +232,6 @@ def SeqMemoryOp : CalyxPrimitive<"seq_mem", []> {
   let builders = [
     OpBuilder<(ins
       "StringRef":$sym_name,
-      "int64_t":$width,
-      "ArrayRef<int64_t>":$sizes,
-      "ArrayRef<int64_t>":$addrSizes
-    )>,
-    OpBuilder<(ins
-      "StringRef":$sym_name,
       "Type":$type,
       "ArrayRef<int64_t>":$sizes,
       "ArrayRef<int64_t>":$addrSizes
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -570,17 +570,10 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
     // for sequential memories will cause a read to take at least 2 cycles,
     // but it will usually be better because combinational reads on memories
     // can significantly decrease the maximum achievable frequency.
-    calyx::RegisterOp reg;
-    if (isa<IntegerType>(loadOp.getMemRefType()))
-      reg = createRegister(
-          loadOp.getLoc(), rewriter, getComponent(),
-          loadOp.getMemRefType().getElementTypeBitWidth(),
-          getState<ComponentLoweringState>().getUniqueName("load"));
-    else
-      reg = createRegister(
-          loadOp.getLoc(), rewriter, getComponent(),
-          loadOp.getMemRefType().getElementType(),
-          getState<ComponentLoweringState>().getUniqueName("load"));
+    calyx::RegisterOp reg = createRegister(
+        loadOp.getLoc(), rewriter, getComponent(),
+        loadOp.getMemRefType().getElementType(),
+        getState<ComponentLoweringState>().getUniqueName("load"));
     rewriter.setInsertionPointToEnd(group.getBodyBlock());
     rewriter.create<calyx::AssignOp>(loadOp.getLoc(), reg.getIn(),
                                      memoryInterface.readData());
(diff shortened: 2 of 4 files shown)


cgyurgyik:
Main question(s): what does a floating point memory look like in native Calyx? My understanding is memories are just treated as vectors of bits, which are subsequently interpreted by users / operations. Is there a benefit to having typed memory here?

(Other than that, just a few smaller comments)

jiahanxie353:
> Main question(s): what does a floating point memory look like in native Calyx? My understanding is memories are just treated as vectors of bits, which are subsequently interpreted by users / operations.

In Calyx native compiler, memory does not care if it's int or float, just a vector of bits, as you said.

>  Is there a benefit to having typed memory here?

It's beneficial for `circt-opt`  but not `circt-translate`, since the former still keeps type information. For example, if we want to store `f32` value into a `SeqMem`, the verifier needs to make sure the source and destination types are the same, thus requiring us to specify the type of each port of the `SeqMem`. But once it's exported to Calyx, you are right that type information no longer matters.

andrewb1999:
Personally I don't think we should have any `f32` types by the time we get to Calyx. Most Calyx ops allow any type, which we probably should change this one to be consistent, but to me it makes sense to treat i32 as a bit vector.

andrewb1999 (changes requested):
SCFToCalyx should cast the `f32` value to an `i32` rather than using the `f32` type in Calyx.

rachitnigam:
Agreed with @andrewb1999 and @cgyurgyik! Let's remove the type-specific tracking for floating-point values.

jiahanxie353:
You have a point, and please convince me on the following example:
```mlir
calyx.cell {
  %float_cst = calyx.const 4.2 : f32
  %mem_0.addr0, ..., %mem_0.write_data, ... = i32, ..., ? what data type for write_data ?
}
calyx.wire {
  %mem_0.write_data = %float_cst.out : ? what data type should we choose here?
}
```
or the floating point adder we just introduced [here](https://github.com/llvm/circt/pull/7089)
```mlir
calyx.cell {
  %float_adder.left, ... = std_addFN  : f32, ...
  %mem_0.addr0, ..., %mem_0.read_data, ... = i32, ..., ? what data type for read_data ?
}
calyx.wire {
  %float_adder.left = %mem_0.read_data : ? what data type should we choose here?
}
```

-----------
Of course, it would be ideal that if MLIR had bit-number only, type-less information like:
```mlir
calyx.cell {
  %float_cst = calyx.const 4.2 : 32
  %mem_0.addr0, ..., %mem_0.write_data, ... = 32, ..., 32
}
calyx.wire {
  %mem_0.write_data = %float_cst.out : 32
}
```

But it does not, so we have to make a decision between:
my solution, which is the work in this patch:
```mlir
calyx.cell {
  %float_cst = calyx.const 4.2 : f32
  %mem_0.addr0, ..., %mem_0.write_data, ... = i32, ..., f32
}
calyx.wire {
  %mem_0.write_data = %float_cst.out : f32
}
```
and yours:
```mlir
calyx.cell {
  %float_cst = calyx.const 4.2 : i32
  %mem_0.addr0, ..., %mem_0.write_data, ... = i32, ..., i32
}
calyx.wire {
  %mem_0.write_data = %float_cst.out : i32
}
```
which cast _everything_ to `i32`, regardless of the intended semantics. But please correct me if I misunderstood anything.

(There's another one, which keeps floating point constants `f32` but has sequential memory ports `i32`. But it obviously doesn't work:
```mlir
calyx.cell {
  %float_cst = calyx.const 4.2 : f32
  %mem_0.addr0, ..., %mem_0.write_data, ... = i32, ..., i32
}
calyx.wire {
  %mem_0.write_data = %float_cst.out : f32
}
```

since `calyx.assign` should verify that type for the left-hand side and RHS should match, as checked [here](https://github.com/llvm/circt/blob/1dc931d2c0968b5bdfb25e7a667d6b350dd5ad0d/include/circt/Dialect/Calyx/CalyxStructure.td#L549)

-------
From my understanding, your solution has one pro and my solution has two:
1. Your solution's pro: it respects Calyx internal representations for values, that is, bit vector.
2. My solution's pro:
    - it respects MLIR's typing semantic, because obviously if we define a value to be `f32`, then it makes more sense to have `%float_cst = calyx.const 4.2 : f32` rather than `%float_cst = calyx.const 4.2 : i32`;
    - it's clearer for Calyx users to see the type information. Imagine a Calyx external user/non-Calyx user, they define an `arith.constant 4.2 : f32` and when they lower it to `calyx.constant`, value `4.2` does not change, but we force cast everything to `i32`, essentially telling the user that: "hey, now 4.2 is an integer type." This can be confusing.
   
Besides, my proposed solution does not completely disrespect Calyx's rule. Because after using `circt-opt --lower-scf-to-calyx`, we are still in the MLIR world, not Calyx native compiler; and it is the point that we do `circt-translate --export-calyx`, that we completely convert to Calyx, completely conform to Calyx's rules, so we remove type information altogether. So mine is not completely off.

Does it make sense?

andrewb1999:
@jiahanxie353 my opinion here is we should change the `calyx.const` op to instead look like 
```
%0 = calyx.const<4.2 : f32> : i32
``` 
and make it have a verifier to ensure the attribute bitwidth and result type bitwidth are the same regardless of type. Essentially with this version we use the `i32` type to just mean any 32 bit type. We don't actually care what the type is once we get to Calyx. If a user wants to define an integer constant and use it as a float, it might give a weird answer but go right ahead. At this level of the compiler infrastructure it's more work than it's worth to do type checking of the type you described.

jiahanxie353:
> we should change the calyx.const op to instead look like `%0 = calyx.const<4.2 : f32> : i32`

this looks awkward, you are keeping two fields to store two types of information: (1) `i32` to indicate that it's a 32 bit vector; (2) `f32` in the attribute to indicate that users are passing in a floating point value. While `%0 = calyx.const 4.2: f32` is simpler and contains both information.

I don't totally agree that we should completely not care about what type it is once we get to Calyx. If this is true, all Calyx `data` json files should be using `bitnum` as the "numeric_type" solely, can you argue that we should also not be using "fixed_point" there at all?

Preserving `f32` type in MLIR-Calyx has no side-effect at all, it subsumes bit number information in `ElementType` and it doesn't even introduce more code. If you are worrying about having an additional `SeqMemoryOp::build`, we can even adopt my latest (experimental) [commit](https://github.com/llvm/circt/pull/7762/commits/89a64dfdf7eba334ef1da01e3b83f0c75489b8c5).

andrewb1999:
> > we should change the calyx.const op to instead look like `%0 = calyx.const<4.2 : f32> : i32`
> 
> this looks awkward, you are keeping two fields to store two types of information: (1) `i32` to indicate that it's a 32 bit vector; (2) `f32` in the attribute to indicate that users are passing in a floating point value. While `%0 = calyx.const 4.2: f32` is simpler and contains both information.

Your version might look simpler, but it is not simpler at all in practice. It ends up meaning drastic changes to the way we use types in the CIRCT Calyx dialect.

> I don't totally agree that we should completely not care about what type it is once we get to Calyx. If this is true, all Calyx `data` json files should be using `bitnum` as the "numeric_type" solely, can you argue that we should also not be using "fixed_point" there at all?

In essence the calyx `data` files are doing the same thing as what I suggest the `calyx.const` operation should do. They are converting the type provided by the user to a sequence of bits. The calyx component itself has no idea what the type of these inputs are. This is an extremely common compilers choice, which is why LLVM/MLIR use signless integers instead of having both signed and unsigned types. The use of a value defines whether it is used as signed or unsigned, not the type itself.

> Preserving `f32` type in MLIR-Calyx has no side-effect at all, it subsumes bit number information in `ElementType` and it doesn't even introduce more code. If you are worrying about having an additional `SeqMemoryOp::build`, we can even adopt my latest (experimental) [commit](https://github.com/llvm/circt/pull/7762/commits/89a64dfdf7eba334ef1da01e3b83f0c75489b8c5).

It doesn't have a side effect in this commit, but it does in terms of maintenance and adding other features. If we allow this, then from now on all operations that might be used to manipulate `f32` values will need to explicitly support it. There is good reason why the native Calyx compiler has made the same decision as what I am suggesting here.

jiahanxie353:
Thank you for the explanation @andrewb1999 , now I'm more convinced.

Please let me ask two more questions,

1. When you say:
> It ends up meaning drastic changes to the way we use types in the CIRCT Calyx dialect.

How drastic can it be? I can only imagine the difference of switching between `getElementType` and `getElementType.getInterWidth`. Can you give some concrete examples?

2. I'was going to support the lowering of[ `CmpFOp`](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithcmpf-arithcmpfop). What do you think about it? One of the differences between `CmpIOp` and`CmpFOp` is that the latter needs to address unordered comparisons such as `NaN`. 
  - Is `f32` vs. `i32` type information important here?
  - Does Calyx primitive library ops address unordered comparison?
Any insight? @andrewb1999 @rachitnigam @cgyurgyik

rachitnigam:
Hi folks, jumping in since this seems like a lively discussion! @jiahanxie353 I generally agree with @andrewb1999 that we should keep the `u32` representation for data when it comes to data in Calyx. I think the design question we're discussing is "what is the value of tracking floating-pointedness of numbers in the Calyx (we don't have to limit ourselves to just CIRCT)?".

There are a couple of key points here:
1. Unlike most software languages, where floating-point refers to IEEE-754, hardware has many different possible floating-point formats. For example, within MLIR, there are a [bunch more](https://mlir.llvm.org/docs/Dialects/Builtin/#float8e4m3b11fnuztype) ad-hoc floating-point representations. Is there any good reason to give IEEE-754 some special preference?
2. Even _if_ the answer to (1) is yes, what does this give us at the Calyx level? We use black-box primitives to perform "floating-point computations" but Calyx has no way to look into the computations or optimize them. Even if we could, [optimizing floating point computations is really complicated](https://herbie.uwplse.org/pldi15-paper.pdf).
3. If we can't do optimizations, then we only get the ability to verify that the computations are flowing into the right type of circuits. This can be done using attributes instead of explicit type-level tracking.

I think fundamentally, the question is not whether or not you can represent things in a compiler; the question is what trade-off does the specific representation make. If (3) is the _best_ thing we get out of adding explicit type-tracking, it is not worth the change in interface and changing norms w.r.t. to the native compiler.

PR #7763, opened 2024-11-01
[circt-test] Add simple runner interface for choosing modes and depth

leonardt:
* Maintains default behavior of using `cover`, `bmc`, and `induction` tasks, but allows the user to supply a custom comma-separated `mode` string selecting specific tasks.  NOTE: The runner script still expects mode to be some subset of the above three.
* Adds a depth attribute/parameter for controlling bound or induction factor.  For now, this parameter is shared across all tasks.

### DIFF
--- a/integration_test/circt-test/basic.mlir
+++ b/integration_test/circt-test/basic.mlir
@@ -86,7 +86,7 @@ verif.formal @ALUCanAdd {
   verif.assert %eq : i1
 }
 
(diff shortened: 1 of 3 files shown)


leonardt:
Maybe we should rename mode to tasks for the attribute and argument.  Also we could make it an array of strings rather than a comma-separated string.

fabianschuiki (approved):
Very cool! Thanks 🥳

PR #7769, opened 2024-11-03
[Calyx] Binary Floating Point MulF Operator

jiahanxie353:
Add support for floating point multiplication in Calyx

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -417,20 +417,35 @@ public:
     }
   }
 
+  template <typename T, typename = void>
+  struct HasGetFloatingPointStandard : std::false_type {};
+
+  template <typename T>
+  struct HasGetFloatingPointStandard<
+      T, std::void_t<decltype(std::declval<T>().getFloatingPointStandard())>>
+      : std::is_same<decltype(std::declval<T>().getFloatingPointStandard()),
+                     FloatingPointStandard> {};
+
   template <typename TLibraryOp>
   TLibraryOp getNewLibraryOpInstance(OpBuilder &builder, Location loc,
                                      TypeRange resTypes) {
     mlir::IRRewriter::InsertionGuard guard(builder);
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
-    auto name = TLibraryOp::getOperationName().split(".").second;
-    if constexpr (TLibraryOp::template hasTrait<FloatingPoint>()) {
-      auto standard = TLibraryOp::getFloatingPointStandard();
-      if (standard == FloatingPointStandard::IEEE754) {
-        assert(name.consume_front("ieee754.") &&
+    std::string name = TLibraryOp::getOperationName().split(".").second.str();
+    if constexpr (HasGetFloatingPointStandard<TLibraryOp>::value) {
+      switch (TLibraryOp::getFloatingPointStandard()) {
+      case FloatingPointStandard::IEEE754: {
+        std::string prefix = "ieee754.";
+        assert(name.find(prefix) == 0 &&
                "IEEE754 type operation's name must begin with 'ieee754'");
-        std::string modified = llvm::join_items("", "std_", name, "FN");
-        name = modified;
+        name.erase(0, prefix.length());
+        name = llvm::join_items(/*separator=*/"", "std_", name, "FN");
+        break;
+      }
+
+      default:
+        llvm_unreachable("Unhandled floating point standard.");
       }
     }
     return builder.create<TLibraryOp>(loc, getUniqueName(name), resTypes);
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -424,10 +424,15 @@ public:
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
     auto name = TLibraryOp::getOperationName().split(".").second;
-    if (std::is_same<calyx::AddFNOp, TLibraryOp>::value)
-      name = "std_addFN";
-    else if (std::is_same<calyx::MulFNOp, TLibraryOp>::value)
-      name = "std_mulFN";
+    if constexpr (TLibraryOp::template hasTrait<FloatingPoint>()) {
+      auto standard = TLibraryOp::getFloatingPointStandard();
+      if (standard == FloatingPointStandard::IEEE754) {
+        assert(name.consume_front("ieee754.") &&
+               "IEEE754 type operation's name must begin with 'ieee754'");
+        std::string modified = llvm::join_items("", "std_", name, "FN");
+        name = modified;
+      }
+    }
     return builder.create<TLibraryOp>(loc, getUniqueName(name), resTypes);
   }
 
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -424,6 +424,10 @@ public:
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
     auto name = TLibraryOp::getOperationName().split(".").second;
+    if (std::is_same<calyx::AddFNOp, TLibraryOp>::value)
+      name = "std_addFN";
+    else if (std::is_same<calyx::MulFNOp, TLibraryOp>::value)
+      name = "std_mulFN";
     return builder.create<TLibraryOp>(loc, getUniqueName(name), resTypes);
   }
 
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -418,10 +418,10 @@ public:
   }
 
   template <typename T, typename = void>
-  struct HasGetFloatingPointStandard : std::false_type {};
+  struct IsFloatingPoint : std::false_type {};
 
   template <typename T>
-  struct HasGetFloatingPointStandard<
+  struct IsFloatingPoint<
       T, std::void_t<decltype(std::declval<T>().getFloatingPointStandard())>>
       : std::is_same<decltype(std::declval<T>().getFloatingPointStandard()),
                      FloatingPointStandard> {};
@@ -433,13 +433,15 @@ public:
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
     std::string name = TLibraryOp::getOperationName().split(".").second.str();
-    if constexpr (HasGetFloatingPointStandard<TLibraryOp>::value) {
+    if constexpr (IsFloatingPoint<TLibraryOp>::value) {
       switch (TLibraryOp::getFloatingPointStandard()) {
       case FloatingPointStandard::IEEE754: {
-        std::string prefix = "ieee754.";
+        constexpr char prefix[] = "ieee754.";
         assert(name.find(prefix) == 0 &&
-               "IEEE754 type operation's name must begin with 'ieee754'");
-        name.erase(0, prefix.length());
+               ("IEEE754 type operation's name must begin with '" +
+                std::string(prefix) + "'")
+                   .c_str());
+        name.erase(0, sizeof(prefix) - 1);
         name = llvm::join_items(/*separator=*/"", "std_", name, "FN");
         break;
       }
--- a/include/circt/Dialect/Calyx/CalyxOps.h
+++ b/include/circt/Dialect/Calyx/CalyxOps.h
@@ -76,16 +76,7 @@ enum class FloatingPointStandard {
 /// Signals that the following operation operates on floating point values.
 template <typename ConcreteType>
 class FloatingPoint
-    : public mlir::OpTrait::TraitBase<ConcreteType, FloatingPoint> {
-public:
-  static constexpr FloatingPointStandard getFloatingPointStandard() {
-    if constexpr (std::is_same<calyx::AddFNOp, ConcreteType>::value ||
-                  std::is_same<calyx::MulFNOp, ConcreteType>::value) {
-      return FloatingPointStandard::IEEE754;
-    }
-    return FloatingPointStandard::IEEE754;
-  }
-};
+    : public mlir::OpTrait::TraitBase<ConcreteType, FloatingPoint> {};
 
 /// The direction of a Component or Cell port. this is similar to the
 /// implementation found in the FIRRTL dialect.
--- a/include/circt/Dialect/Calyx/CalyxOps.h
+++ b/include/circt/Dialect/Calyx/CalyxOps.h
@@ -26,6 +26,8 @@
 namespace circt {
 namespace calyx {
 
+class AddFNOp;
+class MulFNOp;
 // the goPort, donePort, resetPort and clkPort identify the attributes of the
 // go, done, reset and clk port of the circuit.
 static constexpr std::string_view goPort = "go";
@@ -64,6 +66,24 @@ public:
     assert(staticValue == 0 && "If combinational, it should take 0 cycles.");
 
     return success();
+  }
+};
+
+enum class FloatingPointStandard {
+  IEEE754,
+};
+
+/// Signals that the following operation operates on floating point values.
+template <typename ConcreteType>
+class FloatingPoint
+    : public mlir::OpTrait::TraitBase<ConcreteType, FloatingPoint> {
+public:
+  static constexpr FloatingPointStandard getFloatingPointStandard() {
+    if constexpr (std::is_same<calyx::AddFNOp, ConcreteType>::value ||
+                  std::is_same<calyx::MulFNOp, ConcreteType>::value) {
+      return FloatingPointStandard::IEEE754;
+    }
+    return FloatingPointStandard::IEEE754;
   }
 };
 
--- a/include/circt/Dialect/Calyx/CalyxOps.h
+++ b/include/circt/Dialect/Calyx/CalyxOps.h
@@ -26,8 +26,6 @@
 namespace circt {
 namespace calyx {
 
-class AddFNOp;
-class MulFNOp;
 // the goPort, donePort, resetPort and clkPort identify the attributes of the
 // go, done, reset and clk port of the circuit.
 static constexpr std::string_view goPort = "go";
@@ -72,11 +70,6 @@ public:
 enum class FloatingPointStandard {
   IEEE754,
 };
-
-/// Signals that the following operation operates on floating point values.
-template <typename ConcreteType>
-class FloatingPoint
-    : public mlir::OpTrait::TraitBase<ConcreteType, FloatingPoint> {};
 
 /// The direction of a Component or Cell port. this is similar to the
 /// implementation found in the FIRRTL dialect.
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -338,7 +338,7 @@ class ArithBinaryFloatingPointLibraryOp<string mnemonic> :
     SameTypeConstraint<"left", "out">
   ]> {}
 
-def AddFNOp : ArithBinaryFloatingPointLibraryOp<"ieee754.add"> {
+def AddFOpIEEE754 : ArithBinaryFloatingPointLibraryOp<"ieee754.add"> {
   let results = (outs I1:$clk, I1:$reset, I1:$go, I1:$control, I1:$subOp,
     AnySignlessInteger:$left, AnySignlessInteger:$right, AnySignlessInteger:$roundingMode, AnySignlessInteger:$out,
     AnySignlessInteger:$exceptionalFlags, I1:$done);
@@ -378,7 +378,7 @@ def AddFNOp : ArithBinaryFloatingPointLibraryOp<"ieee754.add"> {
   }];
 }
 
-def MulFNOp : ArithBinaryFloatingPointLibraryOp<"ieee754.mul"> {
+def MulFOpIEEE754 : ArithBinaryFloatingPointLibraryOp<"ieee754.mul"> {
   let results = (outs I1:$clk, I1:$reset, I1:$go, I1:$control,
     AnySignlessInteger:$left, AnySignlessInteger:$right, AnySignlessInteger:$roundingMode, AnySignlessInteger:$out,
     AnySignlessInteger:$exceptionalFlags, I1:$done);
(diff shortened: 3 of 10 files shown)


rachitnigam:
Okay, I think we should move the floating point operators into their own separate definition because they are so much more complicated. Can we get the syntax: `calyx.ieee754.add` and `calyx.ieee754.mul` working instead of adding the floating-point operations directly into the same namespace as everything else?

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -378,6 +378,40 @@ def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {
   }];
 }
 
+def MulFNOp : ArithBinaryFloatingPointLibraryOp<"mulFN"> {
```
jiahanxie353, line 381:
```suggestion
def IEEEMulFNOp : ArithBinaryIEEEFloatingPointLibraryOp<"ieeeMulFN"> {
```

Are you imagining something like this? @rachitnigam

test/Dialect/Calyx/emit.mlir
```diff
@@ -331,3 +331,51 @@ module attributes {calyx.entrypoint = "main"} {
...
+// -----
+
+module attributes {calyx.entrypoint = "main"} {
+  // CHECK: import "primitives/float/mulFN.futil";
+  calyx.component @main(%in0: f32, %clk: i1 {clk}, %reset: i1 {reset}, %go: i1 {go}) -> (%out0: f32, %done: i1 {done}) {
+    // CHECK:           std_mulFN_0 = std_mulFN(8, 24, 32);
+    %cst = calyx.constant {sym_name = "cst_0"} 4.200000e+00 : f32
+    %true = hw.constant true
+    %mulf_0_reg.in, %mulf_0_reg.write_en, %mulf_0_reg.clk, %mulf_0_reg.reset, %mulf_0_reg.out, %mulf_0_reg.done = calyx.register @mulf_0_reg : f32, i1, i1, i1, f32, i1
+    %std_mulFN_0.clk, %std_mulFN_0.reset, %std_mulFN_0.go, %std_mulFN_0.control, %std_mulFN_0.left, %std_mulFN_0.right, %std_mulFN_0.roundingMode, %std_mulFN_0.out, %std_mulFN_0.exceptionalFlags, %std_mulFN_0.done = calyx.std_mulFN @std_mulFN_0 : i1, i1, i1, i1, f32, f32, i3, f32, i5, i1
```
jiahanxie353, line 345:
```suggestion
    %std_mulFN_0.clk, ... = calyx.ieee754.mul @std_mulFN_0 : i1, ...
```

similar here? @rachitnigam

include/circt/Dialect/Calyx/CalyxPrimitives.td
andrewb1999: No I think @rachitnigam is imagining changing the op name to `ieee754.mul`, which is totally allowed by MLIR.

test/Conversion/SCFToCalyx/convert_simple.mlir
```diff
@@ -257,3 +257,28 @@ module {
     return %1 : f32
   }
 }
+
+// -----
+
+// Test floating point mul
+
+// CHECK:    %mulf_0_reg.in, %mulf_0_reg.write_en, %mulf_0_reg.clk, %mulf_0_reg.reset, %mulf_0_reg.out, %mulf_0_reg.done = calyx.register @mulf_0_reg : f32, i1, i1, i1, f32, i1
+// CHECK-DAG:    %std_ieee754.mul_0.clk, %std_ieee754.mul_0.reset, %std_ieee754.mul_0.go, %std_ieee754.mul_0.control, %std_ieee754.mul_0.left, %std_ieee754.mul_0.right, %std_ieee754.mul_0.roundingMode, %std_ieee754.mul_0.out, %std_ieee754.mul_0.exceptionalFlags, %std_ieee754.mul_0.done = calyx.std_ieee754.mul @std_ieee754.mul_0 : i1, i1, i1, i1, f32, f32, i3, f32, i5, i1
```
jiahanxie353, line 266: looking good? is this what you are looking for? @rachitnigam @andrewb1999

test/Dialect/Calyx/emit.mlir
```diff
@@ -331,3 +331,52 @@ module attributes {calyx.entrypoint = "main"} {
...
+    %std_ieee754.mul_0.clk, %std_ieee754.mul_0.reset, %std_ieee754.mul_0.go, %std_ieee754.mul_0.control, %std_ieee754.mul_0.left, %std_ieee754.mul_0.right, %std_ieee754.mul_0.roundingMode, %std_ieee754.mul_0.out, %std_ieee754.mul_0.exceptionalFlags, %std_ieee754.mul_0.done = calyx.std_ieee754.mul @std_ieee754.mul_0 : i1, i1, i1, i1, f32, f32, i3, f32, i5, i1
+    %ret_arg0_reg.in, %ret_arg0_reg.write_en, %ret_arg0_reg.clk, %ret_arg0_reg.reset, %ret_arg0_reg.out, %ret_arg0_reg.done = calyx.register @ret_arg0_reg : f32, i1, i1, i1, f32, i1
+    calyx.wires {
+      calyx.assign %out0 = %ret_arg0_reg.out : f32
+      // CHECK-LABEL:    group bb0_0 {
+      // CHECK-NEXT:      std_ieee754.mul_0.left = in0;
+      // CHECK-NEXT:      std_ieee754.mul_0.right = cst_0.out;
+      // CHECK-NEXT:      mulf_0_reg.in = std_ieee754.mul_0.out;
+      // CHECK-NEXT:      mulf_0_reg.write_en = std_ieee754.mul_0.done;
+      // CHECK-NEXT:      std_ieee754.mul_0.go = !std_ieee754.mul_0.done ? 1'd1;
```
jiahanxie353, line 354: also ieee754 namespace here

test/Dialect/Calyx/emit.mlir
rachitnigam: Indeed!

test/Conversion/SCFToCalyx/convert_simple.mlir
rachitnigam: This looks right! Although, why are the names of the ports of the `std_ieee754`? Can we name the module mulFN so that the name is a little more readable?

test/Dialect/Calyx/emit.mlir
```diff
@@ -331,3 +331,52 @@ module attributes {calyx.entrypoint = "main"} {
...
   } {toplevel}
 }
+
+
+// -----
+
+module attributes {calyx.entrypoint = "main"} {
+  // CHECK: import "primitives/float/mulFN.futil";
+  calyx.component @main(%in0: f32, %clk: i1 {clk}, %reset: i1 {reset}, %go: i1 {go}) -> (%out0: f32, %done: i1 {done}) {
+    // CHECK: std_ieee754.mul_0 = std_ieee754.mul(8, 24, 32);
```
rachitnigam, line 341: This check command looks wrong. In native Calyx, the name cannot contain `.` in it...

test/Conversion/SCFToCalyx/convert_simple.mlir
jiahanxie353: Can I double check if I understand you correctly?
In CIRCT-Calyx, we'd like to have:
```mlir
%mulFN_0.clk, %mulFN_0.reset, ... = calyx.std_ieee754.mul @mulFN_0 : i1, i1, ....
```
not:
```mlir
%std_ieee754.mul_0.clk, %std_ieee754.mul_0.reset, ... = calyx.std_ieee754.mul @std_ieee754.mul_0 : i1, i1, ....
```

And after `--export-calyx`, we'd like to have:
```mlir
import "primitives/float/mulFN.futil";
component() {
  cell {
    mulFN_0 = mulFN(8, 24, 32);
  }
  ...
}
```

rachitnigam: Yup! Mostly this is because `mulFN_0 = std_ieee754.mul_0(...)` is not valid Calyx because of the `.` in the name.

jiahanxie353:
marking ready for review again!

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -335,7 +335,7 @@ def XorLibOp  : CombinationalArithBinaryLibraryOp<"xor"> {}
 class ArithBinaryFloatingPointLibraryOp<string mnemonic> : ArithBinaryLibraryOp<mnemonic, [
   SameTypeConstraint<"left", "out">]> {}
 
-def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {
+def AddFNOp : ArithBinaryFloatingPointLibraryOp<"ieee754.add"> {
```
rachitnigam, line 338: Why does this get printed as `calyx.std_ieee754.add` instead of `calyx.ieee754.add`? Can we make it be the latter please?

jiahanxie353: Sure, no problem
Because `ArithBinaryFloatingPointLibraryOp` inherits `CalyxPrimitive`, which automatically prints op names as `"std_"+`, [here](https://github.com/llvm/circt/blob/8a270d72f2e70c1db80db8184ad2ba3a89403d38/include/circt/Dialect/Calyx/CalyxPrimitives.td#L261)

test/Dialect/Calyx/emit.mlir
```diff
@@ -331,3 +331,52 @@ module attributes {calyx.entrypoint = "main"} {
     }
   } {toplevel}
 }
+
+
+// -----
+
+module attributes {calyx.entrypoint = "main"} {
+  // CHECK: import "primitives/float/mulFN.futil";
+  calyx.component @main(%in0: i32, %clk: i1 {clk}, %reset: i1 {reset}, %go: i1 {go}) -> (%out0: i32, %done: i1 {done}) {
```
jiahanxie353, line 340: > Please add a CHECK-NEXT: for this line.

wait it is added here right? I believe you are looking for `std_mulFN_0 = std_mulFN(8, 24, 32)`?

rachitnigam: Ah, can this not use `CHECK-NEXT` since we know exactly which line produces the primitive?

jiahanxie353: you mean we don't need to check `std_mulFN_0 = std_mulFN(8, 24, 32)` at all?

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -424,6 +424,10 @@ class ComponentLoweringStateInterface {
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
     auto name = TLibraryOp::getOperationName().split(".").second;
+    if (std::is_same<calyx::AddFNOp, TLibraryOp>::value)
+      name = "std_addFN";
```
cgyurgyik, line 428: Nit: isn't the better way to do this is remove the `.ieee754` prefix for any `F` operations?

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -1019,11 +1024,15 @@ void Emitter::emitLibraryFloatingPoint(Operation *op) {
     return;
   }
 
-  StringRef opName = op->getName().getStringRef();
+  std::string opName;
+  if (isa<calyx::AddFNOp>(op))
```
cgyurgyik, line 1028: Same here.

cgyurgyik (commented):
A few smaller comments.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
jiahanxie353: The prefix is `calyx.ieee754.add`/`calyx.ieee754.mul`, removing `ieee754` will simply become `add`/`mul`; but we want `std_addFN`/`std_mulFN` after emitting to Calyx native compiler.

cgyurgyik: More concretely, I think there is a better way to do this than manually write a new `if` statement for each floating point operation. It'd be nice to avoid this every time:

```
if (std::is_same<calyx::<MyNewFloatingPointOp, TLibraryOP>::value)
  name = "std_MyNewFloatingPointOp";
```

Ideally, since we're already distinguishing them without a `_std` prefix, we could infer this is some floating point operation, and then check what standard it conforms too.

```cpp
/// Returns the technical floating point this operation conforms to, e.g., "ieee754"
/// (This should probably be an enum, which is subsequently converted to a string.)
std::string standard() { ... }

if (auto fOp = dyn_cast<FloatingPointOp>(op)):
  name.remove(llvm::join(".", fOp.standard()));
  name.prepend("std_");
```

(I'm not saying you should spend a lot of time doing this, it is a mere suggestion. In general, I do think the conforming technical standard of a floating point operation should be a field of the op so it can referenced during pattern rewrites.)

jiahanxie353: I've thought about this, I don't think it's worth it now.
In the foreseeable future, only IEEE754 floating point is integrate; and moreover, only Add and Mul are present in the Berkeley HardFloat.
My opinion is we can address future needs and refactor the code when there is an actual, concrete need.
But I totally get what you are saying.

test/Dialect/Calyx/emit.mlir
rachitnigam: No, can we ensure that the code is generated precisely as a consequence of that one line?

jiahanxie353: Sounds good, I'm updating to `CHECK-DAG` to ensure your request of checking the code is generated by the correct line sequentially; as well as avoiding checking some boilerplate code that will be otherwise generated by `CHECK-NEXT`. Please let me know if it makes sense.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
jiahanxie353: Sorry, let me take it back.
When doing some research on the floating point comparison issue https://github.com/llvm/circt/issues/7821, I found Berkeley HardFloat has division, square root, and comparisons as well.
So I definitely need to use a better design to incorporate all of these `calyx::FloatingPointOp`s, as you suggested.

jiahanxie353:
I'm using `FloatingPoint` trait to check if an operation deals with floating point values; and use an `enum` to distinguish between different standards.

May not be the best design, happy to hear folks' opinions.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -424,6 +424,15 @@ class ComponentLoweringStateInterface {
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
     auto name = TLibraryOp::getOperationName().split(".").second;
+    if constexpr (TLibraryOp::template hasTrait<FloatingPoint>()) {
+      auto standard = TLibraryOp::getFloatingPointStandard();
+      if (standard == FloatingPointStandard::IEEE754) {
+        assert(name.consume_front("ieee754.") &&
+               "IEEE754 type operation's name must begin with 'ieee754'");
+        std::string modified = llvm::join_items("", "std_", name, "FN");
+        name = modified;
```
cgyurgyik, line 433: To avoid a potential copy, just use the r-value:

```suggestion
        name = llvm::join_items("", "std_", name, "FN");
```

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -424,6 +424,15 @@ class ComponentLoweringStateInterface {
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
     auto name = TLibraryOp::getOperationName().split(".").second;
+    if constexpr (TLibraryOp::template hasTrait<FloatingPoint>()) {
+      auto standard = TLibraryOp::getFloatingPointStandard();
+      if (standard == FloatingPointStandard::IEEE754) {
+        assert(name.consume_front("ieee754.") &&
+               "IEEE754 type operation's name must begin with 'ieee754'");
+        std::string modified = llvm::join_items("", "std_", name, "FN");
```
cgyurgyik, line 432: Do we need the empty string in the front?

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -424,6 +424,15 @@ class ComponentLoweringStateInterface {
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
     auto name = TLibraryOp::getOperationName().split(".").second;
+    if constexpr (TLibraryOp::template hasTrait<FloatingPoint>()) {
+      auto standard = TLibraryOp::getFloatingPointStandard();
+      if (standard == FloatingPointStandard::IEEE754) {
```
cgyurgyik, line 429: I think a switch would be appropriate here, so that that if someone adds another case the compiler will complain if we don't update this.
```suggestion
      switch(TLibraryOp::getFloatingPointStandard()) {
        case FloatingPointStandard::IEEE754:
          ...
      }
```

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
cgyurgyik: Still not a fan of this `if-else`, it implies that there will only ever exist two FN functions, add and mul.

include/circt/Dialect/Calyx/CalyxOps.h
```diff
@@ -67,6 +69,24 @@ class Combinational
   }
 };
 
+enum class FloatingPointStandard {
+  IEEE754,
+};
+
+/// Signals that the following operation operates on floating point values.
```
cgyurgyik, line 76:
```suggestion
/// Signals that the following operation operates on floating point values.
/// Unlike other CIRCT dialects, Calyx treats inputs as signless values and
/// leaves their interpretation to the respective operation.
```

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -332,10 +332,17 @@ def AndLibOp  : CombinationalArithBinaryLibraryOp<"and"> {}
...
+}
+
+class ArithBinaryFloatingPointLibraryOp<string mnemonic> :
+  ArithBinaryLibraryOp<mnemonic, "", [
+    FloatingPoint,
+    SameTypeConstraint<"left", "out">
+  ]> {}
 
-def AddFNOp : ArithBinaryFloatingPointLibraryOp<"addFN"> {
+def AddFNOp : ArithBinaryFloatingPointLibraryOp<"ieee754.add"> {
```
cgyurgyik, line 345: I like this better. IMO, a better name for this might be: `AddFOpIeee754` and similar for below. I imagine future ops would follow similar patterns, e.g., `AddFOpBf16`.

cgyurgyik (commented):
Nice, definitely a big improvement. A few more comments.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
jiahanxie353: I don't think `name = llvm::join_items("", "std_", name, "FN");` work
`llvm::join_items` returns a temporary `std::string` so it cannot directly assign to a `StringRef` (`name` here). So I had to assign it to a variable `modified` before assigning to `name`...

Is there any idiomatic way to deal with this, `name = StringRef(llvm::join_items("", "std_", name, "FN"))`?

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
jiahanxie353: I believe so, otherwise it thinks `std_` is the separator

cgyurgyik: Ah yes sorry.

jiahanxie353: sure, good idea!

jiahanxie353:
I have fixed the `StringRef` vs `std::string` for `name` problem;
and have a re-design for distinguish if it's a floating point operation using `FloatingPointInterface` and SFINAE, so that getting the standard and the Calyx library name is more elegant.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -417,13 +417,37 @@ class ComponentLoweringStateInterface {
...
                                      TypeRange resTypes) {
     mlir::IRRewriter::InsertionGuard guard(builder);
     Block *body = component.getBodyBlock();
     builder.setInsertionPoint(body, body->begin());
-    auto name = TLibraryOp::getOperationName().split(".").second;
+    std::string name = TLibraryOp::getOperationName().split(".").second.str();
+    if constexpr (HasGetFloatingPointStandard<TLibraryOp>::value) {
+      switch (TLibraryOp::getFloatingPointStandard()) {
+      case FloatingPointStandard::IEEE754: {
+        std::string prefix = "ieee754.";
```
cgyurgyik, line 439:
```suggestion
        constexpr char[] prefix = "ieee754.";
```

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -417,13 +417,37 @@ class ComponentLoweringStateInterface {
...
+        assert(name.find(prefix) == 0 &&
+               "IEEE754 type operation's name must begin with 'ieee754'");
```
cgyurgyik, line 441: Code duplication nit: use `prefix` instead of repeating the string `ieee754`.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -417,13 +417,37 @@ class ComponentLoweringStateInterface {
     }
   }
 
+  template <typename T, typename = void>
+  struct HasGetFloatingPointStandard : std::false_type {};
```
cgyurgyik, line 421: Suggestion: this name is a bit confusing, maybe just `IsFloatingPoint`?

jiahanxie353:
thanks for the suggestions @cgyurgyik ! I've also changed `AddFNOp`/`MulFNOp` to better naming standards

cgyurgyik (approved):
This looks good to me. I believe Rachit is on vacation, so I think it would be ok to merge (when CI passes) if this is blocking you, and then address any comments later.

jiahanxie353:
> This looks good to me. I believe Rachit is on vacation, so I think it would be ok to merge (when CI passes) if this is blocking you, and then address any comments later.

Sounds good, thank you!

PR #7772, opened 2024-11-04
[FIRRTL] Add new ModulePrefixAnnotation

youngar:
We are in the process of reimplementing the `NestedPrefixModulesAnnotation` directly in Chisel, however there are still some things that Chisel can't do itself and needs to communicate more information to CIRCT. This adds a new `ModulePrefixAnnotation` annotation which can only target memories, which enables firtool to generate memories with correct names.

### DIFF
--- a/include/circt/Dialect/FIRRTL/CHIRRTLOps.td
+++ b/include/circt/Dialect/FIRRTL/CHIRRTLOps.td
@@ -40,7 +40,8 @@ def CombMemOp : CHIRRTLOp<"combmem", [HasCustomSSAName, DeclareOpInterfaceMethod
   let arguments = (ins StrAttr:$name, NameKindAttr:$nameKind,
                        AnnotationArrayAttr:$annotations,
                        OptionalAttr<InnerSymAttr>:$inner_sym,
-                       OptionalAttr<MemoryInitAttr>:$init);
+                       OptionalAttr<MemoryInitAttr>:$init,
+                       OptionalAttr<StrAttr>:$prefix);
   let results = (outs CMemoryType:$result);
   let assemblyFormat = [{(`sym` $inner_sym^)? `` custom<NameKind>($nameKind)
                          `` custom<CombMemOp>(attr-dict) `:` qualified(type($result))}];
@@ -62,7 +63,8 @@ def SeqMemOp : CHIRRTLOp<"seqmem", [HasCustomSSAName, DeclareOpInterfaceMethods<
   let arguments = (ins RUWAttr:$ruw, StrAttr:$name, NameKindAttr:$nameKind,
                        AnnotationArrayAttr:$annotations,
                        OptionalAttr<InnerSymAttr>:$inner_sym,
-                       OptionalAttr<MemoryInitAttr>:$init);
+                       OptionalAttr<MemoryInitAttr>:$init,
+                       OptionalAttr<StrAttr>:$prefix);
   let results = (outs CMemoryType:$result);
   let assemblyFormat = [{(`sym` $inner_sym^)? `` custom<NameKind>($nameKind) $ruw
                          custom<SeqMemOp>(attr-dict) `:` qualified(type($result))}];
--- a/lib/Dialect/FIRRTL/Transforms/LowerCHIRRTL.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerCHIRRTL.cpp
@@ -358,14 +358,14 @@ void LowerCHIRRTLPass::replaceMem(Operation *cmem, StringRef name,
   // Create the memory.
   ImplicitLocOpBuilder memBuilder(cmem->getLoc(), cmem);
   auto symOp = cast<hw::InnerSymbolOpInterface>(cmem);
-  auto prefixAttr = cmem->getAttrOfType<StringAttr>("prefix");
   auto memory = memBuilder.create<MemOp>(
       resultTypes, readLatency, writeLatency, depth, ruw,
       memBuilder.getArrayAttr(resultNames), name,
       cmem->getAttrOfType<firrtl::NameKindEnumAttr>("nameKind").getValue(),
       annotations, memBuilder.getArrayAttr(portAnnotations),
       symOp.getInnerSymAttr(),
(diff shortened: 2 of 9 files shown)


youngar:
There is one difference between the output verilog for this new `ModulePrefixAnnotation ` vs the old `NestedPrefixModulesAnnotation`. When using `ReplSeqMems=true`, we replace a memory operation with an instance of an FModuleOp, which instantiates an FExtModuleOp.  The difference is in the name chosen for the *instance* of the FExtModuleOp: the old (NPMA) name matches the name of the FExtModuleOp before applying the prefix, and the new (MPA) name matches the final, prefix applied, name of the module.  I don't expect this to have any ill-effects, and it might actually be an improvement for those reading the output verilog.

lib/Dialect/FIRRTL/Transforms/LowerCHIRRTL.cpp
```diff
@@ -358,13 +358,14 @@ void LowerCHIRRTLPass::replaceMem(Operation *cmem, StringRef name,
...
   auto symOp = cast<hw::InnerSymbolOpInterface>(cmem);
+  auto prefixAttr = cmem->getAttrOfType<StringAttr>("prefix");
   auto memory = memBuilder.create<MemOp>(
       resultTypes, readLatency, writeLatency, depth, ruw,
       memBuilder.getArrayAttr(resultNames), name,
       cmem->getAttrOfType<firrtl::NameKindEnumAttr>("nameKind").getValue(),
       annotations, memBuilder.getArrayAttr(portAnnotations),
       symOp.getInnerSymAttr(),
-      cmem->getAttrOfType<firrtl::MemoryInitAttr>("init"), StringAttr());
+      cmem->getAttrOfType<firrtl::MemoryInitAttr>("init"), prefixAttr);
```
seldridge, line 368: Nit: the arguments of this create are a mix of two patterns:

1. Single-use variables to improve readability.
2. Inlined expressions as the temporaries would be single-use.

Maybe, let's split the difference and refactor this to use `/*argument=*/` comments?

(I know that I prefer inline and you like outline. 😉) Alternatively, single-use variables for everything. I mostly just don't like the fact that the argument immediately before this is of the exact same form as the new one, but chooses a different format. This can happen in a cleanup or prerequisite.

include/circt/Dialect/FIRRTL/CHIRRTLOps.td
```diff
@@ -40,7 +40,8 @@ def CombMemOp : CHIRRTLOp<"combmem", [HasCustomSSAName, DeclareOpInterfaceMethod
   let arguments = (ins StrAttr:$name, NameKindAttr:$nameKind,
                        AnnotationArrayAttr:$annotations,
                        OptionalAttr<InnerSymAttr>:$inner_sym,
-                       OptionalAttr<MemoryInitAttr>:$init);
+                       OptionalAttr<MemoryInitAttr>:$init,
+                       OptionalAttr<StrAttr>:$prefix);
```
seldridge, line 44: This can represent `OptionalAttr<StrAttr>:$prefix` as `DefaultValuedAttr<StrAttr, "{}">`. I thought there used to be some tradeoff here. Is there or are these the same thing? (Is there any cost to one or the other?)

seldridge (approved):
Generally LGTM.

Aside: given the extreme narrowness of this (an annotation that must be local and can only target three operations, two of which are undocumented in the FIRRTL spec), I'm almost inclined to just put this in the FIRRTL spec. We can revisit this once the Chisel API has landed. The Chisel API keeps everything about the API private, so this can be seamlessly changed later.

include/circt/Dialect/FIRRTL/CHIRRTLOps.td
youngar: Yeah these are a little different, I quickly did some testing for three different options here:
- `OptionalAttr` allows the attribute to be `null`, builders still require it as an argument, `getPrefix()` returns an `optional<StringRef>`
- `DefaultValuedAttr` does not allow the attr to be null, generates a builder which does not require it as an argument, `getPrefix()` returns a `StringRef`
- `DefaultValuedOptionalAttr` allows the attr to be null, builders still require it as an argument, `getPrefix()` returns `StringRef` by using the default value if the attr is null

I used `OptionalAttr` here because that is what is used for `MemOp` and `FirMemOp`.

I don't really know what is best.  Using an optional attribute is kind of confusing because we end up allowing both empty string prefix ("") and null string prefix, and its not clear what the difference is.  We have, in the past, decided not to use Optional attrs when there is a empty value for the attribute, to avoid this kind of confusion, but I know that this can have real overhead costs (but memories are relatively rare).  I see that we use OptionalAttr for `inner_sym`, despite it having a empty value.

mwachs5:
> he difference is in the name chosen for the instance of the FExtModuleOp: the old (NPMA) name matches the name of the FExtModuleOp before applying the prefix, and the new (MPA) name matches the final, prefix applied, name of the module. I don't expect this to have any ill-effects, and it might actually be an improvement for those reading the output verilog.

I am not sure that is acceptable. These instance paths can be a sensitive API for PD scripts which need to target these memories and might not want to rewrite these scripts. Is there a reason we cannot keep the existing functionality?

youngar:
I talked with @jackkoenig about this offline, after going through some of the issues, he think its best if we change how the instance name is selected:
1. the old instance name matched the globally unique name of the memory module after memory-deduplication, but before prefix-modules split it back out in to its own memory.
2. the new instance name will just match the original name of the memory declaration, which is not required to be globally unique
Making this change can be done first, if we find it acceptable?  It can potentially change many of the instance names depending on what memories were deduped and the module name collisions that occurred.

youngar:
I opened a PR with an initial implementation of the above here: https://github.com/llvm/circt/pull/7776

lib/Dialect/FIRRTL/Transforms/LowerCHIRRTL.cpp
youngar: I think this was more of a mistake, from when I was hacking together the initial implementation. I moved it to an inline expression, I think its pretty clear that its just forwarding along the prefix attribute.

youngar:
Since I moved out the instance name changes to a different PR, this PR does not change any pre-existing behaviour anymore.

PR #7776, opened 2024-11-06
[FIRRTL] ReplSeqMems: use the original mem name for the instance name

youngar:
This PR is currently more of an experiment, not sure if it is the direction we will go.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerMemory.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerMemory.cpp
@@ -252,9 +252,9 @@ void LowerMemoryPass::lowerMemory(MemOp mem, const FirMemory &summary,
   auto memModule = getOrCreateMemModule(mem, summary, ports, shouldDedup);
   b.setInsertionPointToStart(wrapper.getBodyBlock());
 
-  auto memInst =
-      b.create<InstanceOp>(mem->getLoc(), memModule, memModule.getModuleName(),
-                           mem.getNameKind(), mem.getAnnotations().getValue());
+  auto memInst = b.create<InstanceOp>(
+      mem->getLoc(), memModule, (mem.getName() + "_ext").str(),
+      mem.getNameKind(), mem.getAnnotations().getValue());
 
   // Wire all the ports together.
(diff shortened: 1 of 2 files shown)


youngar:
It is looking like we will be moving in this direction. Updated the PR with some test updates to pass CI.

lib/Dialect/FIRRTL/Transforms/LowerMemory.cpp
```diff
@@ -252,9 +252,9 @@ void LowerMemoryPass::lowerMemory(MemOp mem, const FirMemory &summary,
   auto memModule = getOrCreateMemModule(mem, summary, ports, shouldDedup);
   b.setInsertionPointToStart(wrapper.getBodyBlock());
 
-  auto memInst =
-      b.create<InstanceOp>(mem->getLoc(), memModule, memModule.getModuleName(),
-                           mem.getNameKind(), mem.getAnnotations().getValue());
+  auto memInst = b.create<InstanceOp>(
+      mem->getLoc(), memModule, (mem.getName() + "_ext").str(),
```
seldridge, line 256: Aside: It's annoying that this needs a string ref and can't be done with a Twine.

seldridge: Two things:

1. Would it be better to just use the original memory name here, i.e., `mem.getName()`?
2. If generating a new name, this can conflict with names in the module namespace and would need to go through a `newName` API to generate a name. (Will this error if I have a wire called `mem_ext` and a memory called `mem`?)

seldridge: For (2), this actually doesn't matter as the name will get uniqued. This is only a problem for symbols.

seldridge (approved):

lib/Dialect/FIRRTL/Transforms/LowerMemory.cpp
jackkoenig: > Would it be better to just use the original memory name here, i.e., mem.getName()?

I think that would be better, but I advised @youngar to keep it _for now_ because I know users rely on `_ext` in brittle regex flows.

PR #7778, opened 2024-11-06
[FreezePaths] Add support for HWInstanceLike, instead of only HWInstanceOp

prithayan:
`PathVisitor::processPath` errors out on other `HWInstanceLike` operations. But other dialects can add paths on such operations, handle `HWInstanceLike` to support operations in other dialects.

### DIFF
--- a/lib/Dialect/OM/Transforms/FreezePaths.cpp
+++ b/lib/Dialect/OM/Transforms/FreezePaths.cpp
@@ -166,14 +166,14 @@ LogicalResult PathVisitor::processPath(Location loc, hw::HierPathOp hierPathOp,
       // If this is our inner ref pair: [Foo::bar]
       // if "bar" is an instance, modules = [Foo::bar], bottomModule = Bar.
       // if "bar" is a wire, modules = [], bottomModule = Foo, component = bar.
-      if (isa<hw::HWInstanceLike>(op)) {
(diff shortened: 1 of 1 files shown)


lib/Dialect/OM/Transforms/FreezePaths.cpp
```diff
@@ -166,14 +166,14 @@ LogicalResult PathVisitor::processPath(Location loc, hw::HierPathOp hierPathOp,
       // If this is our inner ref pair: [Foo::bar]
       // if "bar" is an instance, modules = [Foo::bar], bottomModule = Bar.
       // if "bar" is a wire, modules = [], bottomModule = Foo, component = bar.
-      if (isa<hw::HWInstanceLike>(op)) {
+      if (auto inst = dyn_cast<hw::HWInstanceLike>(op)) {
         // TODO: add support for instance choices.
-        auto inst = dyn_cast<hw::InstanceOp>(op);
-        if (!inst)
+        auto mods = inst.getReferencedModuleNamesAttr();
+        if (mods.size() > 1)
```
youngar, line 172: is it possible for the size to be 0?

youngar (approved):

lib/Dialect/OM/Transforms/FreezePaths.cpp
youngar: sounds good to me 👍

PR #7779, opened 2024-11-07
[FIRRTL] Return an empty ArrayRef<Attr>, not an array of 1 attr

rwy7:


seldridge (approved):
LGTM

Though, I think this is correct because it is no longer returning a reference to a temporary.

PR #7781, opened 2024-11-07
[FIRRTL] infer more return types in asm format

youngar:
Many expressions in FIRRTL implement the InferTypesOpInterface, but we include an explicit result type in the assembly format.  This change removes the explicit result type in the asm format for many FIRRTL expressions, which means less typing.


seldridge:
I think my largest gripe about our FIRRTL Dialect syntax is that it is inconsistent in where this kind of choice is applied. E.g., some operations are very explicit with `(a, b) -> (c)` and others are more abbreviated like `firrtl.probe` and `firrtl.matchingconnect`. I think consistency here is a good thing. This pulls it more in the direction of abbreviation which is most prevalent in newer operations.

This seems reasonable.

@darthscsi: WDYT?

darthscsi:
Although this is a structuralish change and I have no problem with it going in a sweeping change, it might be more manageable to do it op at a time (just to get it unstuck).

PR #7782, opened 2024-11-07
[HW] Add a verifier for HWInstanceLike

prithayan:
Add a verifier to `HWInstanceLike` to ensure the referenced modules list is non empty.

### DIFF
--- a/include/circt/Dialect/HW/HWOpInterfaces.td
+++ b/include/circt/Dialect/HW/HWOpInterfaces.td
@@ -507,14 +507,6 @@ def HWInstanceLike : OpInterface<"HWInstanceLike", [
       "bool", "getDoNotPrint", (ins)
     >
   ];
-
-  let verify = [{
-    if (cast<circt::hw::HWInstanceLike>(op)
-            .getReferencedModuleNamesAttr()
-            .size())
-      return success();
-    return op->emitOpError("has empty list of referenced modules");
-  }];
 }
 
 def InnerRefNamespace : NativeOpTrait<"InnerRefNamespace">;
--- a/include/circt/Dialect/HW/HWOpInterfaces.td
+++ b/include/circt/Dialect/HW/HWOpInterfaces.td
@@ -507,6 +507,14 @@ def HWInstanceLike : OpInterface<"HWInstanceLike", [
       "bool", "getDoNotPrint", (ins)
     >
   ];
+
+  let verify = [{
+    if (cast<circt::hw::HWInstanceLike>(op)
+            .getReferencedModuleNamesAttr()
+            .size())
+      return success();
+    return op->emitOpError("has empty list of referenced modules");
+  }];
 }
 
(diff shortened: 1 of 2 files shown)


include/circt/Dialect/HW/HWOpInterfaces.td
```diff
@@ -507,6 +507,14 @@ def HWInstanceLike : OpInterface<"HWInstanceLike", [
       "bool", "getDoNotPrint", (ins)
     >
   ];
+
+  let verify = [{
+    if (cast<circt::hw::HWInstanceLike>(op)
+            .getReferencedModuleNamesAttr()
+            .size())
+      return success();
+    return op->emitOpError("has empty list of referenced modules");
```
maerhart, line 516: I think you should use `$_op` instead of `op` directly.
I'd also fully qualify everything (i.e., also the `cast` and `success`)

maerhart (approved):
This makes sense to me.

seldridge (commented):
Two things:

1. This is a method that is defined on `InstanceGraphInstanceOpInterface`. ~Interfaces can have verifiers.~ (Edit: obviously as this commit is a verifier of a different interface.) This would be better placed there if possible.
3. This needs a test. 😉

maerhart:
Does the InstanceGraph infrastructure require instances to have at least one referenced module?

prithayan:
Thanks @seldridge , I missed that, will move it to `InstanceGraphInstanceOpInterface`. 
@maerhart , I think it makes sense to assume that at least one referenced module exists.

include/circt/Support/InstanceGraphInterface.td
```diff
@@ -49,6 +49,14 @@ def InstanceGraphInstanceOpInterface : OpInterface<"InstanceOpInterface"> {
       return ::mlir::ArrayAttr::get(attr.getContext(), {attr});
     }]>,
   ];
+
+  let verify = [{
+    if (mlir::cast<ConcreteOp>($_op)
+            .getReferencedModuleNamesAttr()
+            .size())
+      return success();
+    return $_op->emitOpError("has empty list of referenced modules");
```
seldridge, line 58: Mega-nit: error language that is more affirmatively states what the problem is or what the expectation is would be better than an observation (with the implication that this is the problem or the expectation was the inverse). Consider either:

- "expected a non-empty list of referenced modules"
- "must have a non-empty list of referenced modules"

seldridge (commented):
Needs at least one (negative) test!

prithayan:
> Needs at least one (negative) test!
> 
> (Thinking more...) Is there a way to test this?

Yeah, I couldn't figure out a way to test this error, since we can't write lit test with `hw.instance` with zero referenced modules. 
Any other idea  ?

seldridge (approved):
I can't think of a good way to test this. 🙃 LGTM

PR #7783, opened 2024-11-08
[Moore] Add explicit truncation and zero/sign-extension

fabianschuiki:
Add the `moore.trunc` operation to explicitly truncate the bit width of `IntType` values, and `moore.zext` and `moore.sext` to explicitly extend such a value with zeroes or its sign bit.

This requires tweaking the way how ImportVerilog generates conversion ops. Currently `moore.conversion` is used as a catch-all operation that expresses any kind of type conversion. In the future, we'll want to split this up into multiple dedicated operations. These width adjustment ops are the first step in that direction.

Making sign-extension explicit also fixes a long-standing issue where a `$signed` or `signed'(x)` expression would be erroneously converted into a zero-extension.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -610,20 +610,16 @@ struct RvalueExprVisitor {
 
     // Handle left expression.
     builder.setInsertionPointToStart(&trueBlock);
-    auto trueValue = context.convertRvalueExpression(expr.left());
+    auto trueValue = context.convertRvalueExpression(expr.left(), type);
     if (!trueValue)
       return {};
-    if (trueValue.getType() != type)
-      trueValue = builder.create<moore::ConversionOp>(loc, type, trueValue);
     builder.create<moore::YieldOp>(loc, trueValue);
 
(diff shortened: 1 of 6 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1064,3 +1061,43 @@ Value Context::convertToSimpleBitVector(Value value) {
...
+      *dstPacked.getBitSize() != *srcPacked.getBitSize()) {
+    auto dstWidth = *dstPacked.getBitSize();
+    auto srcWidth = *srcPacked.getBitSize();
+
+    // Convert the value to a simple bit vector which we can extend or truncate.
+    auto srcWidthType = moore::IntType::get(value.getContext(), srcWidth,
+                                            srcPacked.getDomain());
+    if (value.getType() != srcWidthType)
+      value = builder.create<moore::ConversionOp>(value.getLoc(), srcWidthType,
+                                                  value);
```
hailongSun2000, line 1084: So cool :sunglasses:!

hailongSun2000 (approved):
I don't have any questions. Perfect :+1:!

PR #7785, opened 2024-11-08
[FIRRTL] Emitting EICG wrapper warning only once

prithayan:
The EICG wrapper warning is be emitted unless the switchover to the intrinsic is made.
There are some designs where we have hundreds of clock gates, the warning clogs up the stdout for such cases.
This commit, ensures the warning is emitted only once.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerIntmodules.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerIntmodules.cpp
@@ -61,6 +61,7 @@ void LowerIntmodulesPass::runOnOperation() {
   auto &ig = getAnalysis<InstanceGraph>();
 
   bool changed = false;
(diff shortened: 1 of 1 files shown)


lib/Dialect/FIRRTL/Transforms/LowerIntmodules.cpp
```diff
@@ -61,6 +61,7 @@ void LowerIntmodulesPass::runOnOperation() {
   auto &ig = getAnalysis<InstanceGraph>();
 
   bool changed = false;
+  bool warningEmittedOnce = false;
```
seldridge, line 64: Mega-nit: can the variable name be more descriptive, i.e., about what warning this is?

seldridge (approved):
LGTM

PR #7787, opened 2024-11-08
[circt-verilog] Support MLIR and MLIRBC input files

fabianschuiki:
Add a `--format` option to circt-verilog that allows it to read SV and MLIR input files. Being able to read MLIR files is useful for debugging or when developing new transformation passes.

Also makes error message use the `WithColor::error()` prefix for somewhat nicer readability.

### DIFF
--- a/tools/circt-verilog/circt-verilog.cpp
+++ b/tools/circt-verilog/circt-verilog.cpp
@@ -14,11 +14,23 @@
 
 #include "circt/Conversion/ImportVerilog.h"
 #include "circt/Conversion/MooreToCore.h"
+#include "circt/Dialect/Comb/CombDialect.h"
+#include "circt/Dialect/Debug/DebugDialect.h"
+#include "circt/Dialect/HW/HWDialect.h"
+#include "circt/Dialect/LLHD/IR/LLHDDialect.h"
+#include "circt/Dialect/Moore/MooreDialect.h"
 #include "circt/Dialect/Moore/MoorePasses.h"
+#include "circt/Dialect/Seq/SeqDialect.h"
+#include "circt/Dialect/Sim/SimDialect.h"
+#include "circt/Dialect/Verif/VerifDialect.h"
 #include "circt/Support/Passes.h"
 #include "circt/Support/Version.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/AsmState.h"
 #include "mlir/IR/BuiltinOps.h"
+#include "mlir/Parser/Parser.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Support/FileUtilities.h"
 #include "mlir/Transforms/Passes.h"
@@ -26,6 +38,7 @@
 #include "llvm/Support/InitLLVM.h"
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/ToolOutputFile.h"
+#include "llvm/Support/WithColor.h"
 
 using namespace llvm;
 using namespace mlir;
@@ -36,6 +49,11 @@ using namespace circt;
 //===----------------------------------------------------------------------===//
 
 namespace {
+enum class Format {
+  SV,
+  MLIR,
+};
+
 enum class LoweringMode {
   OnlyPreprocess,
   OnlyLint,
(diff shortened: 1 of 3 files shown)


tools/circt-verilog/circt-verilog.cpp
```diff
@@ -367,24 +406,73 @@ static LogicalResult executeWithSources(MLIRContext *context,
...
+    hw::HWDialect,
+    llhd::LLHDDialect,
+    moore::MooreDialect,
+    scf::SCFDialect,
+    seq::SeqDialect,
+    sim::SimDialect,
+    verif::VerifDialect
+  >();
+  // clang-format on
+
```
hailongSun2000, line 475: nit: Should we add `cf::ControFlowDialect` :thinking:?

hailongSun2000 (approved):
One nit. LGTM :tada:!

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -14,18 +14,30 @@
...
+#include "circt/Dialect/LLHD/IR/LLHDDialect.h"
+#include "circt/Dialect/Moore/MooreDialect.h"
 #include "circt/Dialect/Moore/MoorePasses.h"
+#include "circt/Dialect/Seq/SeqDialect.h"
+#include "circt/Dialect/Sim/SimDialect.h"
+#include "circt/Dialect/Verif/VerifDialect.h"
 #include "circt/Support/Passes.h"
 #include "circt/Support/Version.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
```
maerhart, line 29: Do we already have all of those dialects in the CMakeLists?

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -367,24 +406,73 @@ static LogicalResult executeWithSources(MLIRContext *context,
...
   outputFile->keep();
   return success();
 }
 
 static LogicalResult execute(MLIRContext *context) {
+  // Abort if there are no input files to be processed.
+  if (opts.inputFilenames.empty()) {
+    WithColor::error() << "no input files\n";
+    return failure();
+  }
```
maerhart, line 420: If we don't allow it to be empty, we should mark the LLVM option as required.
Also, it'd be nice to still support input from stdin. Does it work if we give it "-" as filename?
Maybe only fail here if `format` is also not given? Otherwise, we can process from stdin by default with the specified format?

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -367,24 +406,73 @@ static LogicalResult executeWithSources(MLIRContext *context,
...
     }
     sourceMgr.AddNewSourceBuffer(std::move(buffer), llvm::SMLoc());
   }
 
+  // Register the dialects.
+  // clang-format off
+  context->loadDialect<
+    comb::CombDialect,
+    debug::DebugDialect,
+    func::FuncDialect,
```
maerhart, line 465: I think the func dialect promises the inliner interface. But I guess it's fine if we don't call `inline`? I'm not sure if we do so.

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -367,24 +406,73 @@ static LogicalResult executeWithSources(MLIRContext *context,
...
+    hw::HWDialect,
+    llhd::LLHDDialect,
+    moore::MooreDialect,
+    scf::SCFDialect,
+    seq::SeqDialect,
+    sim::SimDialect,
+    verif::VerifDialect
+  >();
+  // clang-format on
```
maerhart, line 474: Nit: I'm not really a fan of switching off clang-format. If you want to make it easier to understand/read then maybe group them with line comments by abstraction level or sth?

tools/circt-verilog/circt-verilog.cpp
```diff
@@ -367,24 +406,73 @@ static LogicalResult executeWithSources(MLIRContext *context,
...
-      llvm::errs() << errorMessage << "\n";
+      WithColor::error() << errorMessage << "\n";
       return failure();
     }
     sourceMgr.AddNewSourceBuffer(std::move(buffer), llvm::SMLoc());
   }
 
+  // Register the dialects.
+  // clang-format off
+  context->loadDialect<
```
maerhart, line 462: Doesn't the verilog parser explicitly load the dialects it needs and thus registering here would be enough?

maerhart (approved):
LGTM!

tools/circt-verilog/circt-verilog.cpp
fabianschuiki: Oh yeah, great catch! Thanks :+1:

tools/circt-verilog/circt-verilog.cpp
fabianschuiki: Hmmm, the way clang format bunches these up is kind of annoying, because you can't clearly see if something gets added to the list. It's always just a blob of delta. :thinking:

tools/circt-verilog/circt-verilog.cpp
fabianschuiki: That is a good point. It _should_ do that. Let me double-check.

tools/circt-verilog/circt-verilog.cpp
fabianschuiki: I don't think we do. Yet. Maybe something to add once we do?

tools/circt-verilog/circt-verilog.cpp
fabianschuiki: Yeah "-" works here. We could default to reading from stdin if format is present. :+1:

tools/circt-verilog/circt-verilog.cpp
fabianschuiki: :grimacing: No...

PR #7789, opened 2024-11-09
[FIRRTL] Make memory matadata work with layers

seldridge:
Extend the existing notion of how memory metadata is generated to work for memories that are under layers.  Use an interpretation of the "design" excluding layers.  In effect, memories which are under layers now generate metadata as if that instantiation was in the test harness.

I note that the current behavior here is inconsistent.  The metadata generation behavior is different for a memory which is partially in the design (it has instances which are in the design and not in the design) compared to a memory which is wholly not in the design.  However, this is the exact same behavior as before this change.

The details of what happens here are somewhat complex.  For documentation of the current behavior, see the tests that accompany this commit.

In the future, this behavior should be changed to be dramatically simpler.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
@@ -219,8 +219,17 @@ struct ObjectModelIR {
     auto memPaths = instancePathCache.getAbsolutePaths(mem);
     SmallVector<Value> memoryHierPaths;
     SmallVector<Value> finalInstanceNames;
-    // Memory hierarchy is relevant only for memories under DUT.
-    bool inDut = instanceInfo.anyInstanceUnderEffectiveDut(mem);
+    // Add metadata for memory paths that are in the design.  There are two
+    // slightly different code paths here.  If no instance is in the design,
+    // then just skip the memory.  If some instances are in the designs and some
+    // are not, then add paths for memories in the design.  For memories which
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -219,8 +219,17 @@ struct ObjectModelIR {
...
-    // Memory hierarchy is relevant only for memories under DUT.
-    bool inDut = instanceInfo.anyInstanceUnderEffectiveDut(mem);
+    // Add metadata for memory paths that are in the design.  There are two
+    // slightly different code paths here.  If no instance is in the design,
+    // then just skip the memory.  If some instances are in the designs and some
+    // are not, then add paths for memories in the design.  For memories which
+    // are then _not_ in the design, give them unresolvable distinct attribute
+    // paths.  The LowerClasses pass will later treat these as "optimized away"
+    // and create an empty path.
+    //
```
prithayan, line 229: Didn't realize this behavior existed. As far as I can tell, any downstream tool parsing the OMIR would reject such an optimized away empty path, so maybe its safe to assume this case is not executed in any real world design. Thanks for documenting it, in a followup PR I can try to drop such a memory from the metadata.

prithayan (approved):
This makes sense, I think as a followup we can drop memories with an empty path, from the metadata entries.
Also thanks for your previous cleanup of this pass and the lit tests.

PR #7794, opened 2024-11-11
[firtool] mv LoweLayers after CreateSiFiveMetadata

seldridge:
Move the `LowerLayers` later in the `firtool` pass pipeline.  This is working towards moving this pass to the _end_ of the pipeline.  This movement is a prerequisite to enable two features:

  1. Compilation of inline layers
  2. Forcing out of a layerblock


prithayan (approved):
Seems like this should be good to go, once the PRs that update `CreateSiFiveMetadata` to support layers are merged.

PR #7796, opened 2024-11-11
[FIRRTL] Run canonicalizer again after IMCP

rwy7:
IMCP propagates constants across modules, which should enable more
canonicalization.  However, lower-layers, which runs between IMCP and
canonicalization, takes layerblocks and outlines them into their own module.
This effectively undoes the work of IMCP, and prevents us from canonicalizing
as well as we should under layers. This PR adds an additional run of the canonicalizer
immediately after the second run of IMCP, so that we have a better shot at
canonicalization.

In particular, we want to ensure that we are able to canonicalize away dead
assertions and assumptions under layerblocks.


seldridge (approved):
LGTM

PR #7798, opened 2024-11-12
[FIRRTL] Only emit retime metadata for design

seldridge:
Change the `CreateSiFiveMetadata` pass to only create metadata for modules which are instantiated in the design.


prithayan (approved):
Looks good.

PR #7799, opened 2024-11-12
[FIRRTL] Only add SiTest metadata for design

seldridge:
Update the `CreateSiFiveMetadata` pass to treat blackboxes under layers as contributing to test harness JSON metadata.  This brings this metadata in line with other metadata which treats layers as "not design".

Note: for class-based metadata, this pass does not make any distinction between design and test.  Therefore, this commit does not make any change to class-based metadata.  Any blackboxes under layers will be included in class-based metadata.  (I have no idea if this distinction is correct or intended.  However, keeping this the same is at least self-consistent.)


prithayan (approved):
Looks good. Regarding the `SitestBlackBoxMetadata` OM classes, they are currently being ignored by the python tool, that is `firtool` is still generating the metadata files.
The logic in this commit, will be moved over to the OM fields eventually.

PR #7803, opened 2024-11-12
[ESI] Implement basic hostmem reads lacking a bunch of functionality

teqdruid:
Partially implement PyCDE and Cosim BSP of host memory service reads. Has several limitations to make this very simple: 1) one client only, 2) 64-bit reads only (no gearboxing), 3) runtime doesn't do any checking on the read requests. Provides a very basic implementation on which to grow.

All of the cosim stuff (i.e. most of this PR) is subject to change as I get a better idea of how to better implement this in implementing XRT support.


lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
```diff
@@ -415,9 +415,81 @@ class CosimMMIO : public MMIO {
...
+      : acc(acc), ctxt(ctxt), rpcClient(rpcClient) {}
+
+  void start() override {
+    // We have to locate the channels ourselves since this service might be used
+    // to retrieve the manifest.
+
+    // Setup the read side callback.
+    ChannelDesc readArg, readResp;
+    if (!rpcClient->getChannelDesc("__cosim_hostmem_read.arg", readArg) ||
+        !rpcClient->getChannelDesc("__cosim_hostmem_read.result", readResp))
```
mortbopet, line 444: Not the nicest to tightly couple this to `frontends/PyCDE/src/pycde/bsp/cosim.py` through a string constant. I understand that there isn't a neat way to share this string between C++ and Python code, but is this really the best way of doing it?

mortbopet (approved):

lib/Dialect/ESI/runtime/cpp/lib/backends/Cosim.cpp
teqdruid: I can't think of any given that the runtime and PyCDE are not technically coupled. I could imagine a way to keep those constants in the runtime and suck them into CIRCT and/or PyCDE at compile time... but it wouldn't be pretty.

PR #7805, opened 2024-11-12
Revert "[FIRRTL] Don't force non-local trackers in Dedup."

mikeurbach:
Reverts llvm/circt#7709

There were some post-commit questions about this approach, and I'm also not positive we even need the changes any more. If we do, we can either re-land this in its current form, with its warts, or we can solve the problem with more intelligent logic in LowerClasses.


mikeurbach:
Revert is passing CI, so I'll merge this for now.

PR #7811, opened 2024-11-13
[FIRRTL] Document how to deal with specification releases

fabianschuiki:
Add a few lines to the FIRRTL dialect rationale that outline how we deal with new releases of the FIRRTL specification, and how we develop features between CIRCT and the specification.


seldridge (approved):
This is fantastic. Thank you for taking the time to document this!

fabianschuiki:
Happy to do that :+1:

PR #7812, opened 2024-11-13
[Handshake] Adding func instance op for integration

teqdruid:
Adds the ESIInstanceOp. This op allows a non-handshake design to instantiate `handshake.func`s. Since handshake needs elastic inputs and produces elastic outputs, this uses ESI channels.

### DIFF
--- a/include/circt/Dialect/Handshake/HandshakeOps.h
+++ b/include/circt/Dialect/Handshake/HandshakeOps.h
@@ -15,6 +15,7 @@
 
 #include "circt/Dialect/Handshake/HandshakeDialect.h"
 #include "circt/Dialect/Handshake/HandshakeInterfaces.h"
+#include "circt/Dialect/Seq/SeqTypes.h"
 #include "circt/Support/LLVM.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/Builders.h"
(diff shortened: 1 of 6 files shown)


teqdruid:
@mortbopet This is the op we discussed earlier today. Posting it without the lowering to get your eyes on it before you go on vacation. Will add the lowering before merging.

mortbopet:
Some nit comments, but yep, this LGTM. I'm assuming this will lower to a `hw.instance` + `dc.from_esi/dc.to_esi` in [`HandshakeToDC`](https://github.com/llvm/circt/blob/main/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp).

teqdruid:
> I'm assuming this will lower to a `hw.instance` + `dc.from_esi/dc.to_esi` in [`HandshakeToDC`](https://github.com/llvm/circt/blob/main/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp).

Yep. Thanks!

teqdruid:
> > I'm assuming this will lower to a `hw.instance` + `dc.from_esi/dc.to_esi` in [`HandshakeToDC`](https://github.com/llvm/circt/blob/main/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp).
> 
> Yep. Thanks!

Actually, the DC lowering appears to be broken (#7818). I've added it instead to the HandshakeToHW lowering. Will add it to the DC on once that's fixed.

teqdruid:
Will rely on post-merge feedback since @mortbopet is on vacation.

Dinistro (commented):
Sorry for not checking this earlier. I'm no longer deeply involved in handshake and my current job is quite consuming at the moment.

Non-the-less, this change looks reasonable to me.

PR #7814, opened 2024-11-14
[firtool] Run CSE on classes

youngar:
This runs CSE on all operations underneath the circuit. This also moves it up before the rest of the FModule passes run, as the Any pipeline will not merge with them.

### DIFF
--- a/lib/Firtool/Firtool.cpp
+++ b/lib/Firtool/Firtool.cpp
@@ -59,15 +59,14 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
 
   pm.nest<firrtl::CircuitOp>().addPass(firrtl::createInjectDUTHierarchyPass());
 
+  if (!opt.shouldDisableOptimization())
+    pm.nest<firrtl::CircuitOp>().nestAny().addPass(mlir::createCSEPass());
+
   pm.nest<firrtl::CircuitOp>().nest<firrtl::FModuleOp>().addPass(
       firrtl::createPassiveWiresPass());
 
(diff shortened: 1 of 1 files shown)


lib/Firtool/Firtool.cpp
```diff
@@ -59,16 +59,15 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
...
+
   pm.nest<firrtl::CircuitOp>().nest<firrtl::FModuleOp>().addPass(
       firrtl::createPassiveWiresPass());
 
   pm.nest<firrtl::CircuitOp>().nest<firrtl::FModuleOp>().addPass(
       firrtl::createDropNamesPass(opt.getPreserveMode()));
 
-  if (!opt.shouldDisableOptimization())
-    pm.nest<firrtl::CircuitOp>().nest<firrtl::FModuleOp>().addPass(
-        mlir::createCSEPass());
```
seldridge, line 70: Is there any benefit in running twice or keeping this here to run on `FModuleOp`? Or: is CSE blocked on things with different names / does `DropNames` expose work for CSE?

seldridge (approved):

lib/Firtool/Firtool.cpp
youngar: Should be OK I think.  CSE will only remove the ops marked Pure, and anything FNamable is not Pure.

youngar:
@uenoku I merged this as is, but I agree with what you said offline: it seems like it would be a great idea to push this as far to the front of the pipeline as we can.

PR #7819, opened 2024-11-14
[hlstool] Add option to use the DC lowering flow

teqdruid:
Adds a `--dc` flag to use HandshakeToDC (and the DC lowering flow) instead of HandshakeToHW. Not enabling any tests since (AFAICT) none of them pass with this option.

### DIFF
--- a/tools/hlstool/CMakeLists.txt
+++ b/tools/hlstool/CMakeLists.txt
@@ -9,10 +9,14 @@ llvm_update_compile_flags(hlstool)
 target_link_libraries(hlstool
   PRIVATE
 
+  CIRCTDC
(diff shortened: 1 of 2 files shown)


teqdruid:
@mikeurbach @Dinistro Adding you two since @mortbopet is on vacation. Not sure if I'm adding all the necessary passes. It's entirely likely that I'm missing one to get this working...

Dinistro (commented):
I was only tangentially involved in DC, so I'm not sure if this is supposed to work 😅 
Considering that there is either way an issue that needs to be addressed, this will probably be checked while doing so.

mortbopet (commented):
(post-merge approval).

PR #7820, opened 2024-11-15
[FIRRTL] Dedup: hash modules back->front

rwy7:
Hashing modules back->front lets us be a bit leaner with the state we
have to track, and hopefully will give us a nice speed improvement.

In the old algorithm, as we hashed each operation or block in the
module, we would store its position as an index into a side-table.  When
a value is used, we could record the use by hashing-in the result-no and
the defining operation's index (or argument-no and block index, resp).

This index table is a major performance bottleneck for dedup: in a large
module, this table can be massive. The observation made is that values
tend to only be used near their definition. After we hash the last use
of an operation or block, we should be safe to remove its index from the
index table, and keep the index table as small as possible.

This PR modifies the hasher to walk the module backwards. When a value
is first encountered (while hashing a use/operand), we assign an ID to
the defining operation. We use that ID to hash all uses.

When the defining op is hashed, we hash its ID once more (recording the
fact that the ID is defined by the op), and remove the ID from the
table--a value can only be used after it is defined. This ensures that
we only track the ID of an operation for its live range in the IR.

The IDs are assigned according to their "first occurrence" in the
backwards walk of the IR. Since the assignment of IDs is derived from
the structure of the IR, two equivalent modules should assign the same
IDs to the same ops.

This PR also updates the hashing of inner-symbols to be handled in a
similar way. When an inner symbol is referenced, we assign an ID and
record the reference by hashing the ID. When an inner symbol is defined,
we record the definition by, again, hashing the ID. Unlike values, a
symbol can be referenced before it is defined, so we can never free
inner symbol IDs. This corrects an old logical "bug" in dedup, which
never arises in practice because chisel cannot generate it.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -98,20 +98,6 @@ static bool operator==(const ModuleInfo &lhs, const ModuleInfo &rhs) {
          lhs.referredModuleNames == rhs.referredModuleNames;
 }
 
-/// Unique identifier for a value.  All value sources are numbered by apperance,
-/// and values are identified using this numbering (`index`) and an `offset`.
-/// For BlockArgument's, this is the argument number.
-/// For OpResult's, this is the result number.
-struct ValueId {
-  uint64_t index;
-  uint64_t offset;
-};
-
-struct SymbolTarget {
-  ValueId index;
-  uint64_t fieldID;
-};
-
 /// This struct contains constant string attributes shared across different
 /// threads.
 struct StructuralHasherSharedConstants {
@@ -150,7 +136,7 @@ struct StructuralHasherSharedConstants {
 
 struct StructuralHasher {
   explicit StructuralHasher(const StructuralHasherSharedConstants &constants)
-      : constants(constants){};
+      : constants(constants) {}
 
   ModuleInfo getModuleInfo(FModuleLike module) {
     update(&(*module));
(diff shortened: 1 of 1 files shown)


lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -150,14 +136,56 @@ struct StructuralHasherSharedConstants {
...
+      update(getID(op));
+      update(result.getResultNumber());
+      return;
+    }
+    if (auto argument = dyn_cast<BlockArgument>(value)) {
+      auto *block = argument.getOwner();
+      update(getID(block));
+      update(argument.getArgNumber());
+      return;
+    }
```
youngar, line 186: llvm_unreachable here would optimize away the dyn_cast<BlockArgument>

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -318,55 +303,60 @@ struct StructuralHasher {
...
-    }
-
     // Record any op results (types).
     for (auto result : op->getResults())
       update(result);
+
+    // Incorporate the hash of uses we have already built.
+    update(finalizeID(op));
+    update(position);
+    ++position;
```
youngar, line 350: I think its fine, but it might make more sense to do a DF numbering (increment this before recursing into the regions, include it right after the op name).

youngar: I forgot this is doing a backwards walk now

youngar:
This is not your problem, but I think when hashing the attr-dict for operations, we should hash in the name of the attribute for symbols and types. So move this code right below the check for non-essential attributes:
```
      // Hash the interned pointer.
      update(name.getAsOpaquePointer());
```

youngar (approved):

rwy7:
This PR also adds in a "position counter" which is hashed in with each IR object. I think this is unnecessary, but it gives me a bit of extra confidence that two different IRs will push different data through the hasher. We could look at removing it in a followup PR but I think the overhead is negligible.

PR #7823, opened 2024-11-15
[FIRRTL] Refactor class lowering to avoid unnecessary cloning

mikeurbach:
Simplify the class lowering logic by directly moving or cloning operations
instead of cloning the entire module body. This change:

- Removes the need to clone the entire module region
- Only clones instance operations that need to be preserved
- Directly moves property operations to the OM class
- Removes redundant block cleanup

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -1151,12 +1151,17 @@ void LowerClassesPass::lowerClass(om::ClassOp classOp, FModuleLike moduleLike,
     mapping.map(inputValue, parameterValue);
   }
 
-  // Clone the property ops from the FIRRTL Class or Module to the OM Class.
-  SmallVector<Operation *> opsToErase;
-  OpBuilder builder = OpBuilder::atBlockBegin(classOp.getBodyBlock());
-  llvm::SmallVector<mlir::Location> fieldLocs;
-  llvm::SmallVector<mlir::Value> fieldValues;
-  for (auto &op : moduleLike->getRegion(0).getOps()) {
+  // Clone over the body block from the module like to the OM class. This
+  // actually lands the body into the OM class as a second block, which we then
+  // manually splice into the entry block and remove later.
+  moduleLike->getRegion(0).cloneInto(&classOp->getRegion(0),
+                                     classOp->getRegion(0).end(), mapping);
+  classBody->getOperations().splice(
+      classBody->end(), classOp->getRegion(0).back().getOperations());
+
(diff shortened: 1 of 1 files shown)


lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -1201,14 +1232,21 @@ void LowerClassesPass::lowerClass(om::ClassOp classOp, FModuleLike moduleLike,
...
-    for (auto *op : llvm::reverse(opsToErase))
-      op->erase();
+    for (auto &op :
+         llvm::make_early_inc_range(llvm::reverse(moduleLike.getOperation()
+                                                      ->getRegion(0)
+                                                      .getBlocks()
+                                                      .front()
+                                                      .getOperations())))
+      if (isPropertyOp(op) && !isa<InstanceOp>(op))
+        op.erase();
```
youngar, line 1249: Just making sure, is there anything to erase in the original module, since we're moving all the property operations?

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -1169,28 +1170,58 @@ void LowerClassesPass::lowerClass(om::ClassOp classOp, FModuleLike moduleLike,
...
+      auto operand = newOp->getOperand(i);
+      auto it = valueMapping.find(operand);
+      if (it != valueMapping.end())
+        newOp->setOperand(i, it->second);
+    }
+
+    // Move the op into the OM class body.
+    if (needsMove)
+      newOp->moveBefore(classBody, classBody->end());
+  }
```
rwy7, line 1209: Simplification, maybe?

```suggestion
    // Clone instances and register their results to the value mapping.
    if (isa<InstanceOp>(op)) {
      newOp = op.clone();
      for (auto [oldResult, newResult] :
           llvm::zip(op.getResults(), newOp->getResults()))
        valueMapping[oldResult] = newResult;
      classBody->getOperations().insert(classBody->end(), newOp);
      continue;
    }

    // Remap operands using the value mapping.
    for (size_t i = 0; i < newOp->getNumOperands(); ++i) {
      auto operand = newOp->getOperand(i);
      auto it = valueMapping.find(operand);
      if (it != valueMapping.end())
        newOp->setOperand(i, it->second);
    }

    // Move the op into the OM class body.
    newOp->moveBefore(classBody, classBody->end());
  }
```

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
uenoku: Good point, I think this is dead code now.

PR #7824, opened 2024-11-15
[FIRRTL] LowerClasses: drop the IR mapping

rwy7:


lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
```diff
@@ -1238,16 +1194,6 @@ void LowerClassesPass::lowerClass(om::ClassOp classOp, FModuleLike moduleLike,
   // If the module-like is a Class, it will be completely erased later.
   // Otherwise, erase just the property ports and ops.
   if (!isa<firrtl::ClassLike>(moduleLike.getOperation())) {
-    // Erase ops in use before def order, thanks to FIRRTL's SSA regions.
-    for (auto &op :
```
rwy7, line 1242: I'm not 100% if this is safe to delete, see:
https://github.com/llvm/circt/pull/7823/files#r1844569707

cc @mikeurbach @uenoku

uenoku (approved):
LG, thank you for simplifying this further!

lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
rwy7: Seems like it is safe to remove

PR #7825, opened 2024-11-15
[HandshakeToDC] Add clock and reset ports

### LINKED ISSUE #7818: [HandshakeToDC] No clk or rst inserted during lowering
I can't figure out where clock and reset inputs are added with the necessary dc dialect attributes during this lowering. Is this flow just broken?

teqdruid:
After the lowering, add clock and reset ports (with the required attributes) to all of the created modules.

Resolves #7818.

### DIFF
--- a/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
+++ b/lib/Conversion/HandshakeToDC/HandshakeToDC.cpp
@@ -651,9 +651,9 @@ static hw::ModulePortInfo getModulePortInfoHS(const TypeConverter &tc,
   return hw::ModulePortInfo{inputs, outputs};
 }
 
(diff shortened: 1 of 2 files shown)


teqdruid:
@mortbopet This is a kinda hacky way to do this. Ideally there'd be additional ports created during the FuncOp lowering, but I couldn't figure out the right way to merge the blocks such that existing operations didn't have the wrong operands. I think it's the TypeConverter but I didn't want to spend any more time on this.

mortbopet:
LGTM - only nit would be to move the string literals "dc.clock"/"dc.reset" to some static constant in a header. Or, even better, make them [unit attributes](https://mlir.llvm.org/docs/Dialects/Builtin/#unitattr) in DC.

PR #7827, opened 2024-11-15
[FIRRTL] Update, fix parser versions

seldridge:
FIRRTL v4.0.0 was released [[1]].  Update the FIRRTL parser to correctly accept/reject features from this version now that the feature has solidified.  Fix a number of incorrect version checks.

Note: when going through this, I realized that the 3.3.0 version made some breaking changes, notably with respect to "optional groups".  These were added in 3.2.0 and dropped in 3.3.0 (in favor of layers).  This isn't hugely problematic, however, it did violate the FIRRTL specification versioning policy.  Oops.

The proliferation of version checks on FIRRTL features is really starting to make me think that we want to put this into a tabular structure indicating: (1) when a feature was added and (2) when a feature was removed.  We can then just check all these features automatically as opposed to sprinkling the checks around.

[1]: https://github.com/chipsalliance/firrtl-spec/releases/tag/v4.0.0

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRParser.h
+++ b/include/circt/Dialect/FIRRTL/FIRParser.h
@@ -124,7 +124,7 @@ constexpr FIRVersion minimumFIRVersion(2, 0, 0);
 /// new version of the spec is released, all uses of `nextFIRVersion` in the
 /// parser are replaced with the concrete version `{x, y, z}`, and this
 /// declaration here is bumped to the next probable version number.
-constexpr FIRVersion nextFIRVersion(4, 0, 0);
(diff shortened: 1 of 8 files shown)


fabianschuiki (approved):
Really neat! It might make sense to adjust the `nextFIRVersion` in `FIRParser.h` to be something like `4.0.1` or `4.1.0` or `5.0.0`; anything greater than `4.0.0`. Otherwise we'd be exporting FIRRTL labeled as `4.0.0` containing features beyond `4.0.0`.

seldridge:
I bumped the "next" version to 4.1.0 in the `!fixup` commit. Good idea.

Issue #7431, opened 2024-08-03
[ImportVerilog] Create variables for function arguments


fabianschuiki:
The following SV currently crashes because `x` is lowered to an `i32`, which is unassignable, instead of a `ref<i32>`, which would be assignable:
```
function void foo(int x);
  x = 0;
endfunction
```
To fix this, create local variables that shadow the function arguments.

Max-astro:
Hi fabianschuiki,

I recently encountered the same crash you reported. I’m interested in working on this fix and learning more about the CIRCT project, as I’m still a beginner here.

Where you suggested that "create local variables that shadow the function arguments." might be a fix. I reviewed the crash stack, but I'm not fully clear on which function I need to create local variables for function arguments.

Could you explain a bit more about what "shadowing" function arguments involves in this context, or perhaps share an example of what you had in mind? Any additional insights would be greatly appreciated as I work toward understanding and implementing a solution.

Thank you for your help and guidance, and I look forward to hopefully collaborating on this!

fabianschuiki:
Hi @Max-astro, thanks a lot for taking a look at this!

The crash happens because `x` is converted to a function argument, which is just a simple value. In order to be able to assign to it, `x` would have to be converted to a variable instead. This is all handled in the `Context::convertFunction` function in ImportVerilog, in the `lib/Conversion/ImportVerilog/Structure.cpp` file.

In a nutshell, the current implementation is trying to convert the SV
```systemverilog
function void foo(int x);
  x = 0;
endfunction
```
into the MLIR
```mlir
func.func private @foo(%arg0: !moore.i32) {
  %0 = moore.constant 0 : i32
  moore.assign %arg0, %0 : i32
  return
}
```
But this crashes since `moore.assign` requires the destination of the assignment to be a `!moore.ref<i32>`, so something like a `moore.variable`, instead of a simple `i32` like the function argument.

To fix this, the conversion should produce something like the following instead:
```mlir
func.func private @foo(%arg0: !moore.i32) {
  %x = moore.variable %arg0 : <i32>  // <--- a variable that is initialized to the func arg value
  %0 = moore.constant 0 : i32
  moore.assign %x, %0 : i32  // <--- assignment to the variable instead of the func arg
  return
}
```

[...]

PR #7829, opened 2024-11-16
[ImportVerilog]: Create variables for function arguments

Max-astro:
Adds shadow variables for SV function arguments.
Fixes: https://github.com/llvm/circt/issues/7431

Adjusts two test cases ```test/Conversion/ImportVerilog/basic.sv``` and ```test/Conversion/ImportVerilog/builtins.sv```

Adds two test functions in ```test/Conversion/ImportVerilog/builtins.sv```, which will crash in the former ```circt-verilog``` tool.

### DIFF
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -894,6 +894,7 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
   ValueSymbolScope scope(valueSymbols);
 
   // Create a function body block and populate it with block arguments.
+  SmallVector<moore::VariableOp> argVariables;
   auto &block = lowering->op.getBody().emplaceBlock();
   for (auto [astArg, type] :
        llvm::zip(subroutine.getArguments(),
@@ -908,10 +909,11 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
       OpBuilder::InsertionGuard g(builder);
       builder.setInsertionPointToEnd(&block);
 
-      Value shadowArg = builder.create<moore::VariableOp>(
+      auto shadowArg = builder.create<moore::VariableOp>(
           loc, moore::RefType::get(cast<moore::UnpackedType>(type)),
           StringAttr{}, blockArg);
       valueSymbols.insert(astArg, shadowArg);
+      argVariables.push_back(shadowArg);
     }
   }
 
@@ -946,5 +948,16 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
   }
   if (returnVar && returnVar.use_empty())
     returnVar.getDefiningOp()->erase();
+
+  for (auto var : argVariables) {
+    if (llvm::all_of(var->getUsers(),
+                     [](auto *user) { return isa<moore::ReadOp>(user); })) {
+      for (auto *user : llvm::make_early_inc_range(var->getUsers())) {
+        user->getResult(0).replaceAllUsesWith(var.getInitial());
+        user->erase();
+      }
+      var->erase();
+    }
+  }
   return success();
 }
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -900,7 +900,19 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
                  lowering->op.getFunctionType().getInputs())) {
     auto loc = convertLocation(astArg->location);
     auto blockArg = block.addArgument(type, loc);
-    valueSymbols.insert(astArg, blockArg);
+
+    if (isa<moore::RefType>(type)) {
+      valueSymbols.insert(astArg, blockArg);
+    } else {
+      // Convert the body of the function.
+      OpBuilder::InsertionGuard g(builder);
+      builder.setInsertionPointToEnd(&block);
+
+      Value shadowArg = builder.create<moore::VariableOp>(
+          loc, moore::RefType::get(cast<moore::UnpackedType>(type)),
+          StringAttr{}, blockArg);
+      valueSymbols.insert(astArg, shadowArg);
+    }
   }
 
   // Convert the body of the function.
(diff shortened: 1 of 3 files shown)


test/Conversion/ImportVerilog/basic.sv
```diff
@@ -2055,20 +2126,54 @@ endmodule
...
+  
+  // CHECK: [[C2:%.+]] = moore.constant 2 : i32
+  // CHECK: moore.blocking_assign [[Y]], [[C2]] : i32
+  y = 2;
+
+  // CHECK: [[READ_X:%.+]] = moore.read [[X]] : <i32>
+  // CHECK: [[READ_Y:%.+]] = moore.read [[Y]] : <i32>
+  // CHECK: [[ADD:%.+]] = moore.add [[READ_X]], [[READ_Y]] : i32
+  return x+y;
+endfunction
```
hailongSun2000, line 2179: nit: please add a new empty line.

hailongSun2000 (approved):
LGTM! Thank @Max-astro for your work on this error(https://github.com/llvm/circt/issues/7431).

hailongSun2000:
>  @fabianschuiki: And if the user never assigns to the variable, the canonicalizers will get rid of it automatically again later 😃

If you want IR to stay the same when using `--canonicalize` like
```
func.func private @dummyE(%arg0: !moore.i1) -> !moore.i1 {
    return %arg0 : !moore.i1
}
```
You have to implement the `VariableOp::Canonicalize()` manually. But you can add this in the new PR :smile:.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -900,7 +900,19 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
                  lowering->op.getFunctionType().getInputs())) {
     auto loc = convertLocation(astArg->location);
     auto blockArg = block.addArgument(type, loc);
-    valueSymbols.insert(astArg, blockArg);
+
```
fabianschuiki, line 903: If you'd like to get rid of unneeded variables, you could create a list of shadow variables while you build them up:
```suggestion

SmallVector<moore::VariableOp> argVariables;
```

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -900,7 +900,19 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
...
+    if (isa<moore::RefType>(type)) {
+      valueSymbols.insert(astArg, blockArg);
+    } else {
+      // Convert the body of the function.
+      OpBuilder::InsertionGuard g(builder);
+      builder.setInsertionPointToEnd(&block);
+
+      Value shadowArg = builder.create<moore::VariableOp>(
+          loc, moore::RefType::get(cast<moore::UnpackedType>(type)),
+          StringAttr{}, blockArg);
```
fabianschuiki, line 913:
```suggestion
      auto shadowArg = builder.create<moore::VariableOp>(
          loc, moore::RefType::get(cast<moore::UnpackedType>(type)),
          StringAttr{}, blockArg);
      argVariables.push_back(shadowArg);
```

fabianschuiki: And then further down just before `return success();` you could replace the variables that were not needed:
```c++
for (auto var : argVariables) {
  if (llvm::all_of(var.getUsers(), [](auto *user){ return isa<moore::ReadOp>(user); })) {
    for (auto *user : llvm::make_early_inc_range(var.getUsers())) {
      user->replaceAllUsesWith(var.getInitial());
      user->erase();
    }
    var.erase();
  }
}
```

fabianschuiki (approved):
Awesome, thanks for implementing this!

If you want to get rid of all the changes in the SV tests, you could do what @hailongSun2000 has suggested and try to delete the variables again if they were not needed. That should make most tests see no real change at all, which may be beneficial? Let me know what you think :smiley: I'm also very happy with just landing this!

lib/Conversion/ImportVerilog/Structure.cpp
Max-astro: Thank you for your suggestions. I believe eliminating unnecessary variables immediately is more efficient, as it reduces memory usage, and the output MLIR looks cleaner compared to retaining them until the `Canonicalize()`.  (Maybe my instincts aren't right at all)

Should I commit this enhancement to this PR or a new PR?

hailongSun2000: If you don't want to use `Canonicalize()`, maybe "this PR" is perfect.

Max-astro: > If you don't want to use `Canonicalize()`, maybe "this PR" is perfect.

Thanks for your response, could you explain more about the pros and cons of these two implementations?

hailongSun2000: These two implementations are similar. The only difference is that we have to invoke `--canonicalize` using `circt-opt`  explicitly. So when we just use `circt-translate --import-verilog` to test, we cannot hide the function arguments unused by `moore.blocking_assign`. For example:
```
function int simpleFunc3(int a, b);
  {
    // CHECK: [[ARG0:%.+]] = moore.variable %arg0 : <i32>
    // CHECK: [[ARG1:%.+]] = moore.variable %arg1 : <i32>
    // CHECK: [[READ0:%.+]] = moore.read [[ARG0]] : <i32>
    // CHECK: [[READ1:%.+]] = moore.read [[ARG1]] : <i32>
  }
    
    // CHECK: [[TMP:%.+]] = moore.mul [[READ0]], [[READ1]]
    // CHECK: return [[TMP]]
    return a * b;
  endfunction
```
In a nutshell, these two implementations both generate IRs surrounded in a brace(`{ }`). But non-`Canonicalize` will hide them when we use `circt-translate`.

fabianschuiki (approved):
Very cool, thanks a lot for doing all this work!

Max-astro:
> Very cool, thanks a lot for doing all this work!

I really appreciate the guidance you provided, it means a lot to me that you took the time to help.

hailongSun2000 (approved):
So cool!

hailongSun2000:
Can I merge it? @Max-astro

Max-astro:
> Can I merge it? @Max-astro

Sure, I'm glad to hear that.

fabianschuiki:
Fantastic, thanks for merging @hailongSun2000 and thanks for fixing this bug @Max-astro 🥳!

PR #7830, opened 2024-11-16
[Calyx] Lower SCF parallel op to Calyx

jiahanxie353:
This patch lowers `scf::parallel` to Calyx, assuming that all induction variables can be statically derived and that `scf::reduce` does not have operands.

### DIFF
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -1595,7 +1595,6 @@ private:
           // All combinations have been generated
           done = true;
       }
-
       if (done)
         break;
     }
@@ -1662,8 +1661,11 @@ private:
 
         /// Only schedule the 'after' block. The 'before' block is
         /// implicitly scheduled when evaluating the while condition.
-        LogicalResult res = buildCFGControl(path, rewriter, whileBodyOpBlock,
-                                            block, whileOp.getBodyBlock());
+        if (LogicalResult result =
+                buildCFGControl(path, rewriter, whileBodyOpBlock, block,
+                                whileOp.getBodyBlock());
+            result.failed())
+          return result;
 
         // Insert loop-latch at the end of the while group
         rewriter.setInsertionPointToEnd(whileBodyOpBlock);
@@ -1671,9 +1673,6 @@ private:
             getState<ComponentLoweringState>().getWhileLoopLatchGroup(whileOp);
         rewriter.create<calyx::EnableOp>(whileLatchGroup.getLoc(),
                                          whileLatchGroup.getName());
-
-        if (res.failed())
-          return res;
       } else if (auto *parSchedPtr = std::get_if<ParScheduleable>(&group)) {
         auto parOp = parSchedPtr->parOp;
         auto calyxParOp = rewriter.create<calyx::ParOp>(parOp.getLoc());
@@ -1682,11 +1681,11 @@ private:
           rewriter.setInsertionPointToEnd(calyxParOp.getBodyBlock());
           auto seqOp = rewriter.create<calyx::SeqOp>(parOp.getLoc());
           rewriter.setInsertionPointToEnd(seqOp.getBodyBlock());
-          res = scheduleBasicBlock(rewriter, path, seqOp.getBodyBlock(),
-                                   &innerBlock);
+          if (res = scheduleBasicBlock(rewriter, path, seqOp.getBodyBlock(),
+                                       &innerBlock);
+              res.failed())
+            return res;
         }
-        if (res.failed())
-          return res;
       } else if (auto *forSchedPtr = std::get_if<ForScheduleable>(&group);
                  forSchedPtr) {
         auto forOp = forSchedPtr->forOp;
@@ -1701,8 +1700,10 @@ private:
         auto *forBodyOpBlock = forBodyOp.getBodyBlock();
 
         // Schedule the body of the for loop.
-        LogicalResult res = buildCFGControl(path, rewriter, forBodyOpBlock,
-                                            block, forOp.getBodyBlock());
+        if (LogicalResult res = buildCFGControl(path, rewriter, forBodyOpBlock,
+                                                block, forOp.getBodyBlock());
+            res.failed())
+          return res;
 
         // Insert loop-latch at the end of the while group.
         rewriter.setInsertionPointToEnd(forBodyOpBlock);
@@ -1710,8 +1711,6 @@ private:
             getState<ComponentLoweringState>().getForLoopLatchGroup(forOp);
         rewriter.create<calyx::EnableOp>(forLatchGroup.getLoc(),
                                          forLatchGroup.getName());
-        if (res.failed())
-          return res;
       } else if (auto *ifSchedPtr = std::get_if<IfScheduleable>(&group);
                  ifSchedPtr) {
         auto ifOp = ifSchedPtr->ifOp;
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -119,10 +119,15 @@ struct CallScheduleable {
   func::CallOp callOp;
 };
 
+struct ParScheduleable {
+  /// Parallel operation to schedule.
+  scf::ParallelOp parOp;
+};
+
 /// A variant of types representing scheduleable operations.
 using Scheduleable =
     std::variant<calyx::GroupOp, WhileScheduleable, ForScheduleable,
-                 IfScheduleable, CallScheduleable>;
+                 IfScheduleable, CallScheduleable, ParScheduleable>;
 
 class IfLoweringStateInterface {
 public:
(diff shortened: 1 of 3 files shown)


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1093,6 +1103,19 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
   return success();
 }
 
+LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
+                                     scf::ReduceOp reduceOp) const {
+  if (!reduceOp.getOperands().empty())
```
cgyurgyik, line 1108: It might be nice here to emit some error message, since this might be a surprising pre-condition.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1481,6 +1504,75 @@ class BuildIfGroups : public calyx::FuncOpPartialLoweringPattern {
...
+    rewriter.setInsertionPointAfter(scfParOp);
+    scf::ParallelOp newParOp = scfParOp.cloneWithoutRegions();
+    auto loc = newParOp.getLoc();
+    rewriter.insert(newParOp);
+    OpBuilder insideBuilder(newParOp);
+    Block *currBlock = nullptr;
+    auto &region = newParOp.getRegion();
+    IRMapping operandMap;
+    std::function<void(SmallVector<int64_t, 4> &, unsigned)> genIVCombinations;
+    genIVCombinations = [&](SmallVector<int64_t, 4> &indices, unsigned dim) {
```
cgyurgyik, line 1544: Suggestion: what about making this its own function? Its quite large for a lambda function. I would also generally suggest avoid recursion in compilers, as this can lead to stack overflow.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1547,6 +1640,19 @@ class BuildControl : public calyx::FuncOpPartialLoweringPattern {
         rewriter.create<calyx::EnableOp>(whileLatchGroup.getLoc(),
                                          whileLatchGroup.getName());
 
+        if (res.failed())
```
cgyurgyik, line 1643: Why was this added here, and not above where the result is?

```
if (LogicalResult result = buildCFGControl(path, rewriter, whileBodyOpBlock, block, whileOp.getBodyBlock()); result.failed()) {
  return result;
}
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1547,6 +1640,19 @@ class BuildControl : public calyx::FuncOpPartialLoweringPattern {
...
+          return res;
+      } else if (auto *parSchedPtr = std::get_if<ParScheduleable>(&group)) {
+        auto parOp = parSchedPtr->parOp;
+        auto calyxParOp = rewriter.create<calyx::ParOp>(parOp.getLoc());
+        LogicalResult res = LogicalResult::success();
+        for (auto &innerBlock : parOp.getRegion().getBlocks()) {
+          rewriter.setInsertionPointToEnd(calyxParOp.getBodyBlock());
+          auto seqOp = rewriter.create<calyx::SeqOp>(parOp.getLoc());
+          rewriter.setInsertionPointToEnd(seqOp.getBodyBlock());
+          res = scheduleBasicBlock(rewriter, path, seqOp.getBodyBlock(),
```
cgyurgyik, line 1653: Don't you need to check this upon each iteration instead of overwriting it?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1093,6 +1103,19 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
   return success();
 }
 
+LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
+                                     scf::ReduceOp reduceOp) const {
+  if (!reduceOp.getOperands().empty())
+    return failure();
+  return success();
```
cgyurgyik, line 1110: Suggestion:

```suggestion
    return reduceOp.getOperands().empty() ? success() : failure();
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1481,6 +1504,75 @@ class BuildIfGroups : public calyx::FuncOpPartialLoweringPattern {
...
 };
 
+class BuildParGroups : public calyx::FuncOpPartialLoweringPattern {
+  using FuncOpPartialLoweringPattern::FuncOpPartialLoweringPattern;
+
+  LogicalResult
+  partiallyLowerFuncToComp(FuncOp funcOp,
+                           PatternRewriter &rewriter) const override {
+    LogicalResult res = success();
+    funcOp.walk([&](Operation *op) {
```
cgyurgyik, line 1514: Can't you just walk `scf::ParallelOp`s?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1481,6 +1504,75 @@ class BuildIfGroups : public calyx::FuncOpPartialLoweringPattern {
...
+          return WalkResult::interrupt();
+        }
+      return WalkResult::advance();
+    });
+
+    return res;
+  }
+
+private:
+  LogicalResult partialEval(PatternRewriter &rewriter,
```
cgyurgyik, line 1527: Add documentation please so someone understands what you're trying to achieve.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: you are right, i just copied the code from the previous PR

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: making it to a while loop now

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1093,6 +1103,21 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
   return success();
 }
 
+LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
+                                     scf::ReduceOp reduceOp) const {
+  // we don't handle reduce operation and simply return success for now since
+  // because BuildParGroups would have already emitted an error and exited early
```
cgyurgyik, line 1109:
```suggestion
  // BuildParGroups would have already emitted an error and exited early
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1481,6 +1506,104 @@ class BuildIfGroups : public calyx::FuncOpPartialLoweringPattern {
...
 
+class BuildParGroups : public calyx::FuncOpPartialLoweringPattern {
+  using FuncOpPartialLoweringPattern::FuncOpPartialLoweringPattern;
+
+  LogicalResult
+  partiallyLowerFuncToComp(FuncOp funcOp,
+                           PatternRewriter &rewriter) const override {
+    WalkResult walkResult = funcOp.walk([&](scf::ParallelOp scfParOp) {
+      if (!scfParOp.getResults().empty())
+        return scfParOp.emitError("Currently don't support reduce operation"),
```
cgyurgyik, line 1517: (I'm personally not a fan of the comma operator used in this situation, I don't think we gain much from it other than a lack of curly braces I guess)

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1537,18 +1661,31 @@ class BuildControl : public calyx::FuncOpPartialLoweringPattern {
...
             getState<ComponentLoweringState>().getWhileLoopLatchGroup(whileOp);
         rewriter.create<calyx::EnableOp>(whileLatchGroup.getLoc(),
                                          whileLatchGroup.getName());
-
-        if (res.failed())
-          return res;
+      } else if (auto *parSchedPtr = std::get_if<ParScheduleable>(&group)) {
+        auto parOp = parSchedPtr->parOp;
+        auto calyxParOp = rewriter.create<calyx::ParOp>(parOp.getLoc());
+        LogicalResult res = LogicalResult::success();
```
cgyurgyik, line 1679: Is this initialization needed? Can we just initialize `res` in the if statement below?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
cgyurgyik, line 1517:
```suggestion
        return scfParOp.emitError("Reduce operations in scf.parallel is not supported yet"),
```

cgyurgyik (commented):
Nice! A few smaller comments.

cgyurgyik (approved):
LGTM! Thanks for your patience, feel free to merge when CI passes.

PR #7832, opened 2024-11-18
[Ibis] Rename to 'Kanagawa'

teqdruid:


blakep-msft (approved):
LGTM.  Thanks!

mortbopet (approved):

seldridge:
This PR seems to have inadvertently de-bumped LLVM. I'll see about fixing it.

teqdruid:
Oops. It might have been since I let it sit around for so long.

PR #7833, opened 2024-11-18
[RTG] Add Random Test Generation dialect

maerhart:
Please refer to the `RTG.md` file for the rationale.

While CIRCT has a strong set of dialects to generate hardware from various source languages, produce SV, simulate, formally verify, and even synthesize, it lacks tools for system-level verification. This dialect aims to address this by introducing an infrastructure to represent and compile randomized tests. These tests can be ISA instruction sequences to test CPUs, transaction sequences to test modules with a ready-valid interface, etc. A later PR will add another dialect (a test dialect) to demonstrate and test this infrastructure for ISA instruction sequences.

This dialect is different from all the other dialects in CIRCT in that it isn't part of a compilation pipeline from a hardware description. Instead, it is used to write/generate tests for the hardware produced by the other dialects in CIRCT and is thus more disconnected from them. Therefore, we might want to restructure CIRCT a bit to introduce top-level projects like the LLVM repo. However, let's wait and see what happens with the upstream restructuring efforts.


docs/Dialects/RTG.md
```diff
@@ -0,0 +1,275 @@
...
+dialect that represents the actual test constructs, i.e., the parts that get
+randomized. After all the randomization constructs are fully elaborated, the
+resulting test will only consist of those dialects.
+
+Examples for tests that can be randomized, and thus candidates for companion
+dialects, are instruction sequences of any ISA (Instruction Set Architecture),
+transaction sequences over a latency insensitive (ready-valid) channel, input
+sequences to an FSM (Finite State Machine), transaction sequences for
+potentially any other protocol. Albeit, the initial motivation are ISA tests and
+will thus be the best supported, at least initially.
```
TaoBi22, line 24: super-super-super-nit
```suggestion
will thus these be the best supported, at least initially.
```

darthscsi (approved):
LGTM

docs/Dialects/RTG.md
```diff
@@ -0,0 +1,275 @@
...
+  memories in ISA tests, or channels in hardware transaction tests
+* *Contexts*: the context or environment the instuctions are performed in,
+  e.g., on which CPU and in which mode for ISA test.
+
+To express their mapping, the companion dialects must implement the the
+interfaces as follows.
+
+**Instructions**
+
+They implement `InstructionOpInterface`on all operations that represent an
```
fabianschuiki, line 68:
```suggestion
They implement `InstructionOpInterface` on all operations that represent an
```

fabianschuiki (approved):
Very exciting! LGTM

PR #7836, opened 2024-11-19
[FIRRTL] Lower firrtl.formal to verif.formal

fabianschuiki:
Add a lowering from `firrtl.formal` to `verif.formal` in the FIRRTL-to-HW conversion. The lowering creates a `verif.formal` op that simply instantiates the module pointed to by `firrtl.formal`, with symbolic values providing the inputs to the module. We may want to inline the instance in the future, but that is purely cosmetic.

This now allows FIR files to contain `formal` unti tests that can be lowered to HW through `firtool` and then executed using `circt-test`.

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -578,6 +578,8 @@ private:
                               CircuitLoweringState &loweringState);
   LogicalResult lowerModuleOperations(hw::HWModuleOp module,
                                       CircuitLoweringState &loweringState);
+  LogicalResult lowerFormalBody(verif::FormalOp formalOp,
+                                CircuitLoweringState &loweringState);
 };
 
 } // end anonymous namespace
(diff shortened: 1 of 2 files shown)


lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -670,6 +673,16 @@ void FIRRTLModuleLowering::runOnOperation() {
               state.recordModuleMapping(&op, loweredMod);
               return success();
             })
+            .Case<FormalOp>([&](auto oldFormalOp) {
+              auto builder = OpBuilder::atBlockEnd(topLevelModule);
+              auto newFormalOp = builder.create<verif::FormalOp>(
+                  oldFormalOp.getLoc(), oldFormalOp.getNameAttr());
+              newFormalOp->setDiscardableAttrs(oldFormalOp.getParametersAttr());
```
maerhart, line 680: Maybe we want some ODS verified dictionary attribute for the `verif.formal` op instead of using a discardable attribute?

maerhart (approved):
LGTM!

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
fabianschuiki: We we should definitely do that. I'll add that in a follow-up PR.

PR #5224, opened 2023-05-17
[OM] Create OM Nodes for metadata

prithayan:
Create `OM` nodes to represent the `json` metadata in `CreateSifiveMetadata`.
This adds the required infrastructure for including the `om` dialect, to `FIRRTL`. 
The `emit-metadata` pass now generates the metadata as `om.class` in the IR, in addition to the `verbatim`.
The lit tests also check that the om dialect is generated and preserved in the `mlir` after `ExportVerilog`.

Example IR:
Memory metadata: 
```
    firrtl.memmodule private @head_ext(in W0_addr: !firrtl.uint<5>, in W0_en: !firrtl.uint<1>, in W0_clk: !firrtl.clock, in W0_data: !firrtl.uint<5>) attributes {dataWidth = 5 : ui32, depth = 20 : ui64, extraPorts = [], maskBits = 1 : ui32, numReadPorts = 0 : ui32, numReadWritePorts = 0 : ui32, numWritePorts = 1 : ui32, readLatency = 1 : ui32, writeLatency = 1 : ui32}
    firrtl.memmodule private @head_0_ext(in W0_addr: !firrtl.uint<5>, in W0_en: !firrtl.uint<1>, in W0_clk: !firrtl.clock, in W0_data: !firrtl.uint<5>) attributes {dataWidth = 5 : ui32, depth = 20 : ui64, extraPorts = [], maskBits = 1 : ui32, numReadPorts = 0 : ui32, numReadWritePorts = 0 : ui32, numWritePorts = 1 : ui32, readLatency = 1 : ui32, writeLatency = 1 : ui32}

    om.class @MemoryConf(%name: !om.str, %depth: ui64, %width: ui32, %maskBits: ui32, %readPorts: ui32, %writePorts: ui32, %readwritePorts: ui32, %writeLatency: ui32, %readLatency: ui32) {
      om.class.field @name, %name : !om.str
      om.class.field @depth, %depth : ui64
      om.class.field @width, %width : ui32
      om.class.field @maskBits, %maskBits : ui32
      om.class.field @readPorts, %readPorts : ui32
      om.class.field @writePorts, %writePorts : ui32
      om.class.field @readwritePorts, %readwritePorts : ui32
      om.class.field @writeLatency, %writeLatency : ui32
      om.class.field @readLatency, %readLatency : ui32
    }
    om.class @MemoryMetadata() {
      %0 = om.constant #om.str<"head_ext"> : !om.str
      %1 = om.constant 20 : ui64
[...]

PR #7837, opened 2024-11-19
[firtool] Run LowerFormalToHW pass when emitting SV

fabianschuiki:
Add the `LowerFormalToHW` pass to the HW-to-SV pipeline of firtool. This will make any `verif.formal` ops be emitted as additional top-level modules, which is in line with what circt-test also does.

In the future we may want to have additional controls in firtool to disable emission of these formal tests. (But we would likely do this by removing them early in the pipeline to open up new optimization possibilities rather than not emitting them at the final stage.)

This creates a small amount of churn in the `directories.fir` test which was sensitive to _any_ kind of folder running, e.g. through the dialect conversion or greedy rewriter framework. The latter one is used by the formal op lowering, and it's probably a good idea to not have tests fail when a very common optimization runs.

### DIFF
--- a/test/Dialect/FIRRTL/SFCTests/directories.fir
+++ b/test/Dialect/FIRRTL/SFCTests/directories.fir
@@ -140,56 +140,38 @@ circuit TestHarness:
 ; MLIR_OUT:  }
 
 ; MLIR_OUT:  om.class @SitestBlackBoxMetadata(%basepath: !om.basepath) -> [[V1:.+]]: !om.class.type<@SitestBlackBoxModulesSchema>, [[V2:.+]]: !om.class.type<@SitestBlackBoxModulesSchema>, [[V3:.+]]: !om.class.type<@SitestBlackBoxModulesSchema>
-; MLIR_OUT:    %0 = om.constant "Foo_BlackBox" : !om.string
-; MLIR_OUT:    %1 = om.object @SitestBlackBoxModulesSchema(%basepath, %0) : (!om.basepath, !om.string) -> !om.class.type<@SitestBlackBoxModulesSchema>
-; MLIR_OUT:    %2 = om.constant "Bar_BlackBox" : !om.string
-; MLIR_OUT:    %3 = om.object @SitestBlackBoxModulesSchema(%basepath, %2) : (!om.basepath, !om.string) -> !om.class.type<@SitestBlackBoxModulesSchema>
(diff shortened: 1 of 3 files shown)


test/Dialect/FIRRTL/SFCTests/directories.fir
```diff
@@ -140,56 +140,38 @@ circuit TestHarness:
...
+; MLIR_OUT-DAG: [[OBJ2:%.+]] = om.object @MemorySchema
+; MLIR_OUT-DAG: om.constant "bar_m_ext" : !om.string
+; MLIR_OUT-DAG: om.constant #om.integer<2 : ui64> : !om.integer
+; MLIR_OUT-DAG: om.path_create instance %basepath @memNLA_1
+; MLIR_OUT-DAG: om.path_create instance %basepath @memNLA_2
+; MLIR_OUT-DAG: om.list_create
+; MLIR_OUT-DAG: [[OBJ3:%.+]] = om.object @MemorySchema
+; MLIR_OUT-DAG: om.constant "baz_m_ext" : !om.string
+; MLIR_OUT-DAG: om.constant #om.integer<3 : ui64> : !om.integer
+; MLIR_OUT:     om.class.fields [[OBJ1]], [[OBJ2]], [[OBJ3]] : !om.class.type<@MemorySchema>, !om.class.type<@MemorySchema>, !om.class.type<@MemorySchema>
```
seldridge, line 175: This is fine. However, this test in appropriately combined directory testing with OM testing. (Apparently, my past self thought the same thing: https://github.com/llvm/circt/pull/5224#discussion_r1277128769)

If it's cleaner, in a follow-on, to either (1) factor this out of this test or (2) just delete this part of the test, then that would be good. @prithayan: Do you have any thoughts here?

seldridge (approved):
LGTM

test/Dialect/FIRRTL/SFCTests/directories.fir
fabianschuiki: Yeah that would be great!

prithayan: Sorry for the messy lit tests and missing out your original review @seldridge , I think it makes sense to remove the testing from here. I added the om classes just because the original test was checking for the metadata verbatim ops, which will be replaced by the OM. I will take up the task to cleanup the tests.

PR #2110, opened 2021-11-09
[FIRRTL][LowerToHW] Use always instead of always_ff

uenoku:
This PR replaces `always_ff` with `always` at lower-to-hw to fix the invalid usages of always_ff with initialization blocks.
This PR also removes `alwaysFF` option in firtool😀

Will close #1707

PR #7838, opened 2024-11-19
[SeqToSV] Do not use `always_ff` for compreg with initializer

fzi-hielscher:
By default SeqToSV uses `always_ff` processes to lower `seq.compreg` ops. This prohibits any other assignments to the register/variable in another process. [Apparently, some simulators consider an initializer at the declaration to be a process.](https://stackoverflow.com/questions/74392350/the-multiple-driver-error-when-writing-always-ff) Other simulators don't, and it  is also not how I would interpret the SV spec, but what do I know?

I think dodging a simulator error is more important than having the `always_ff`. So, this PR changes the lowering to use `always` instead of `always_ff` in the presence of an initializer, even if the `lower-to-always-ff` flag is set.


teqdruid (approved):
This seems like a reasonable compromise to me, though it makes me sad. Thanks!

youngar:
We had the same problem with FIRRTL https://github.com/llvm/circt/pull/2110

youngar (approved):

PR #7843, opened 2024-11-19
[ExternalizeRegisters][NFC] Clean up initial value fetching

TaoBi22:
Just a cleanup commit that reuses `unwrapImmutableValue` instead of pretty much reimplementing it as I only just found out it existed - thanks for including this @uenoku!

### DIFF
--- a/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
+++ b/lib/Tools/circt-bmc/ExternalizeRegisters.cpp
@@ -98,8 +98,7 @@ void ExternalizeRegistersPass::runOnOperation() {
           if (auto initVal = regOp.getInitialValue()) {
             // Find the constant op that defines the reset value in an initial
             // block (if it exists)
-            if (isa<BlockArgument>(initVal) ||
-                !isa<seq::InitialOp>(initVal.getDefiningOp())) {
+            if (!initVal.getDefiningOp<seq::InitialOp>()) {
               regOp.emitError("registers with initial values not directly "
                               "defined by a seq.initial op not yet supported");
               return signalPassFailure();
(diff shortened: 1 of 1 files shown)


lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -96,22 +96,16 @@ void ExternalizeRegistersPass::runOnOperation() {
...
           if (auto initVal = regOp.getInitialValue()) {
-            // Find the seq.initial op where the initial value is defined and
-            // fetch the operation inside that defines the value
-            auto initialOp =
-                regOp.getInitialValue().getDefiningOp<seq::InitialOp>();
-            if (!initialOp) {
+            // Find the constant op that defines the reset value in an initial
+            // block (if it exists)
+            if (isa<BlockArgument>(initVal) ||
+                !isa<seq::InitialOp>(initVal.getDefiningOp())) {
```
maerhart, line 102: Bit unfortunate that `unwrapImmutableValue` asserts that the passed value comes from an initial op instead of just returning a null value.

Nit: this seems a bit more readable to me:
```suggestion
            if (!initVal.getDefiningOp<seq::InitialOp>()) {
```

maerhart (approved):
Cool improvement. Thanks!

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
TaoBi22: Agreed, it would be nice to move this inside the function at some point, that feels more in line with what we see elsewhere.

Agreed that that's nicer, thanks!

PR #7844, opened 2024-11-19
[HWToBTOR2] Fix crashes on initial value corner cases

TaoBi22:
Currently HWToBTOR2 crashes on both of the included error cases:

1. `unwrapImmutableValue` throws an assert whenever it's given a value that wasn't defined by an initialOp, so I've added a check for that. While the first error case isn't technically legal anyway as instance ops aren't allowed in HWToBTOR2, the pass currently crashes before it gets to that point.
2. Initial values given as a block argument will throw an error in requireSort as they don't have a queriable width. Pretty unlikely edge case, but can't hurt to have it covered! I've added it to the list of sorts ignored by requireSort, and it'll now get mopped up by the same check as point 1 rather than crashing

### DIFF
--- a/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
+++ b/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
@@ -594,8 +594,8 @@ public:
   void visit(hw::PortInfo &port) {
     // Separate the inputs from outputs and generate the first btor2 lines for
     // input declaration We only consider ports with an explicit bit-width (so
-    // ignore clocks)
-    if (port.isInput() && !isa<seq::ClockType>(port.type)) {
+    // ignore clocks and immutables)
+    if (port.isInput() && !isa<seq::ClockType, seq::ImmutableType>(port.type)) {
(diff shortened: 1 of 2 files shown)


lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -880,6 +880,12 @@ struct ConvertHWToBTOR2Pass
     genState(reg, w, regName);
 
     if (init) {
+      if (init && (isa<BlockArgument>(init) ||
+                   !isa<seq::InitialOp>(init.getDefiningOp()))) {
```
dobios, line 884: any reason why you're checking init twice here?

dobios (approved):
LGTM apart for the small nit! Thanks for catching these!

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
TaoBi22: Oops! Just a copy-paste error, thanks for the catch!

PR #7845, opened 2024-11-19
[HWToBTOR2][NFC] update 'PowerOn' in error message and comments

TaoBi22:
Very small NFC just to update the language used in the pass referring to initial values

### DIFF
--- a/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
+++ b/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
@@ -254,7 +254,7 @@ private:
     return constlid;
   }
 
-  // Generates an init statement, which allows for the use of powerOnValue
+  // Generates an init statement, which allows for the use of initial values
   // operands in compreg registers
   void genInit(Operation *reg, Value initVal, int64_t width) {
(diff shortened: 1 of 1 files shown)


lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -884,9 +884,9 @@ struct ConvertHWToBTOR2Pass
       auto initialConstant = circt::seq::unwrapImmutableValue(init)
                                  .getDefiningOp<hw::ConstantOp>();
       if (!initialConstant)
-        reg->emitError("PowerOn Value must be constant!!");
+        reg->emitError("Initial value must be constant!!");
```
fabianschuiki, line 887: Maybe no exclamation marks 😁?
```suggestion
        reg->emitError("initial value must be constant");
```

fabianschuiki (approved):

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
TaoBi22: Sounds wise to me!

PR #7846, opened 2024-11-19
[Handshake][Python] Add Python bindings for Handshake

teqdruid:


teqdruid:
This one was odd since the Handshake tablegen/printer/parser are still in an old style and missing features. I don't want to go down the rat hole of touching the internals, which would be necessary to clean this up.

Dinistro (approved):
I'm not too familiar with the python bindings, but given that the test passes, this is probably fine 😅 .

PR #7848, opened 2024-11-19
[RTG] Add set type and operations

maerhart:


include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -95,3 +95,44 @@ def InvokeSequenceOp : RTGOp<"invoke_sequence", []> {
...
+  let hasVerifier = 1;
+}
+
+def SetSelectRandomOp : RTGOp<"set_select_random", [
+  Pure,
+  TypesMatchWith<"output must be of the element type of input set", 
+                 "set", "output",
+                 "llvm::cast<rtg::SetType>($_self).getElementType()">
+]> {
+  let summary = "selects an element uniformly at random from a set";
```
youngar, line 117: What does this return if the set is empty?

youngar (approved):

include/circt/Dialect/RTG/IR/RTGOps.td
maerhart: This is currently undefined behavior. My plan is to add another operation to query the number of elements in the set. `scf.if` could then be used to skip random selections from empty sets. Alternatively, we could pass a default value to be chosen if the set is empty. Feel free to share your thoughts/opinion on how this should ideally be handled.

PR #7849, opened 2024-11-19
[PyCDE][Handshake] Add bindings for Handshake functions

teqdruid:
Exposes the Handshake dialect's FuncOp in the same style as an HWModuleOp:

```py

class TestFunc(Func):
  a = Input(Bits(8))
  x = Output(Bits(8))

  @generator
  def build(ports):
    ports.x = ports.a & Bits(8)(0xF)
```

And instantiates them via ESI channels:

```py
class Top(Module):
  clk = Clock()
  rst = Input(Bits(1))

  a = Input(Channel(Bits(8)))
  x = Output(Channel(Bits(8)))

  @generator
  def build(ports):
    test = TestFunc(clk=ports.clk, rst=ports.rst, a=ports.a)
    ports.x = test.x

```


mikeurbach:
Very cool!

mortbopet (approved):
LGTM modulo "locking things down" and only allowing `comb` and `handshake` operations inside the `Func` (... for now).

teqdruid:
> LGTM modulo "locking things down" and only allowing `comb` and `handshake` operations inside the `Func` (... for now).

The reason I didn't is that I'm not quite sure the best way to do this. I don't want the `signals` module to know about the `handshake` module... so I'd have to create a new infrastructure for the containing module to hook into any operation creation to validate or override it. I think that's a bit too much for this PR. Unless you have any suggestions for a simpler way to do this...?

PR #7855, opened 2024-11-20
[RTG] Add op and type visitors

maerhart:
Add an RTG visitor ahead of adding a pass using it to keep the pass PR small(er).


youngar (approved):
Code looks good. We have seen that this visitor pattern does not compile very well, especially when there are lots of ops you don't care about, so I am a little worried that this is actually an anti-pattern. But that is a problem for another day :)

PR #7856, opened 2024-11-20
[RTG] Add TestOp, TargetOp, and DictType

maerhart:


include/circt/Dialect/RTG/IR/RTGTypes.td
```diff
@@ -47,4 +47,24 @@ class SetTypeOf<Type elementType> : ContainerType<
...
+    This type is a dictionary with a static set of entries. This datatype does
+    not make any assumptions about how the values are stored (could be a struct,
+    a map, etc.). Furthermore, two values of this type should be considered
+    equivalent if they have the same set of entry names and types and the values
+    match for each entry, independent of the order.
+  }];
+
+  let parameters = (ins
+    ArrayRefParameter<"mlir::StringAttr", "dict entry names">:$entryNames,
+    ArrayRefParameter<"mlir::Type", "dict entry types">:$entryTypes);
```
youngar, line 62: you might consider making this a single array of a pair-like data structure.  You wouldn't need the verifier anymore.

lib/Dialect/RTG/IR/RTGTypes.cpp
```diff
@@ -19,6 +19,77 @@ using namespace rtg;
...
+        auto fieldLoc = p.getCurrentLocation();
+        if (p.parseKeywordOrString(&name) || p.parseColon() ||
+            p.parseType(type))
+          return failure();
+
+        if (!nameSet.insert(name).second) {
+          p.emitError(fieldLoc, "duplicate entry name \'" + name + "\'");
+          // Continue parsing to print all duplicates, but make sure to error
+          // eventually
+          hasDuplicateName = true;
```
youngar, line 65: Can you just return failure() here, and remove the `hasDuplicateName` local?
```
        if (!nameSet.insert(name).second) {
          p.emitError(fieldLoc, "duplicate entry name \'" + name + "\'");
```

lib/Dialect/RTG/IR/RTGTypes.cpp
```diff
@@ -19,6 +19,77 @@ using namespace rtg;
...
+      result = failure();
+      emitError() << "duplicate entry name '" << name.getValue()
+                  << "' in rtg.dict type";
+    }
+  }
+  return result;
+}
+
+Type DictType::parse(AsmParser &p) {
+  llvm::StringSet<> nameSet;
```
youngar, line 46: Since we're interning the field names anyways, I think it would make more sense to use some kind of StringAttrs instead.

lib/Dialect/RTG/IR/RTGTypes.cpp
```diff
@@ -19,6 +19,77 @@ using namespace rtg;
...
+        return success();
+      });
+
+  if (hasDuplicateName)
+    return Type();
+
+  if (failed(parseResult))
+    return Type();
+
+  return get(p.getContext(), names, types);
```
youngar, line 79: Do we need to sort the fields by name, so that we can guarantee that two dictionary types with the same fields intern to the same thing?

youngar (approved):

PR #7857, opened 2024-11-20
[circt-test] Add runner for circt-bmc

fabianschuiki:
Add a runner script for `circt-bmc` to circt-test. This allows CIRCT's own model checker to be used to verify `verif.formal` operations in MLIR inputs. The user currently has to pass the runner to circt-test explicitly using the `-r` option. In the future, we'll want to add a mechanism to circt-test to list available runners, check if the needed software is installed on the system, and configure new local runners through some config file.

### DIFF
--- a/tools/circt-test/CMakeLists.txt
+++ b/tools/circt-test/CMakeLists.txt
@@ -29,5 +29,16 @@ target_link_libraries(circt-test PRIVATE ${libs})
 llvm_update_compile_flags(circt-test)
 mlir_check_all_link_libraries(circt-test)
 
-configure_file(circt-test-runner-sby.py ${CIRCT_TOOLS_DIR}/circt-test-runner-sby.py)
-add_custom_target(circt-test-runner-sby SOURCES ${CIRCT_TOOLS_DIR}/circt-test-runner-sby.py)
+set(all_runner_paths "")
+foreach(runner sby circt-bmc)
+  set(src_path ${CMAKE_CURRENT_SOURCE_DIR}/circt-test-runner-${runner}.py)
+  set(dst_path ${CIRCT_TOOLS_DIR}/circt-test-runner-${runner}.py)
+  add_custom_command(
+    OUTPUT ${dst_path}
+    DEPENDS ${src_path}
+    COMMAND ${CMAKE_COMMAND} -E copy ${src_path} ${dst_path}
+  )
+  list(APPEND all_runner_paths ${dst_path})
+endforeach()
+add_custom_target(circt-test-runners DEPENDS ${all_runner_paths})
+add_dependencies(circt-test circt-test-runners)
--- a/tools/circt-test/circt-test-runner-circt-bmc.py
+++ b/tools/circt-test/circt-test-runner-circt-bmc.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+from pathlib import Path
+import argparse
+import os
+import shlex
+import subprocess
+import sys
+
+parser = argparse.ArgumentParser()
+parser.add_argument("mlir")
+parser.add_argument("-t", "--test", required=True)
+parser.add_argument("-d", "--directory", required=True)
+parser.add_argument("-m", "--mode")
+parser.add_argument("-k", "--depth", type=int)
+args = parser.parse_args()
+
+# Use circt-opt to lower any `verif.formal` ops to `hw.module`s. Once circt-bmc
+# natively supports `verif.formal`, we can directly run it on the input MLIR.
+lowered_mlir = Path(args.directory) / "lowered.mlir"
+cmd = ["circt-opt", args.mlir, "-o", lowered_mlir, "--lower-formal-to-hw"]
+sys.stderr.write("# " + shlex.join(str(c) for c in cmd) + "\n")
+sys.stderr.flush()
+result = subprocess.call(cmd)
+if result != 0:
+  sys.exit(result)
+
+# Run circt-bmc. We currently have to capture the output and look for a specific
+# string to know if the verification passed or failed. Once the tool properly
+# exits on test failure, we can skip this.
+cmd = ["circt-bmc", lowered_mlir]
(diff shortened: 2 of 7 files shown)


tools/circt-test/CMakeLists.txt
```diff
@@ -29,5 +29,13 @@ target_link_libraries(circt-test PRIVATE ${libs})
 llvm_update_compile_flags(circt-test)
 mlir_check_all_link_libraries(circt-test)
 
-configure_file(circt-test-runner-sby.py ${CIRCT_TOOLS_DIR}/circt-test-runner-sby.py)
-add_custom_target(circt-test-runner-sby SOURCES ${CIRCT_TOOLS_DIR}/circt-test-runner-sby.py)
+foreach(runner sby circt-bmc)
+  configure_file(
+    circt-test-runner-${runner}.py
+    ${CIRCT_TOOLS_DIR}/circt-test-runner-${runner}.py
+  )
```
maerhart, line 36: Nit: Using `configure_file` adds a dependency between this source file and the configuration step, i.e., whenever the python file is modified, this cmake step is run again instead of just copying the file. I don't think this is necessary.
I *think* the way to avoid this is to use `add_custom_command`. You can take a look at how this is done for the python bindings:
https://github.com/llvm/llvm-project/blob/83c7784c35918ce037823f29d29918c5542cdf9c/mlir/cmake/modules/AddMLIRPython.cmake#L598-L605

tools/circt-test/circt-test-runner-circt-bmc.py
```diff
@@ -0,0 +1,47 @@
...
+parser.add_argument("-d", "--directory", required=True)
+parser.add_argument("-m", "--mode", type=int)
+parser.add_argument("-k", "--depth")
+args = parser.parse_args()
+
+if args.mode is not None and args.mode != "bmc":
+  sys.stderr.write("error: circt-bmc only supports `bmc` mode")
+
+# Use circt-opt to lower any `verif.formal` ops to `hw.module`s. Once circt-bmc
+# natively supports `verif.formal`, we can directly run it on the input MLIR.
```
maerhart, line 21: I think once we have this connection, we'd ideally just instantiate the circt-bmc pipeline and jit exec call in circt-test directly instead of going through a python script.

tools/circt-test/circt-test-runner-circt-bmc.py
```diff
@@ -0,0 +1,47 @@
...
+if result != 0:
+  sys.exit(result)
+
+# Run circt-bmc. We currently have to capture the output and look for a specific
+# string to know if the verification passed or failed. Once the tool properly
+# exits on test failure, we can skip this.
+cmd = ["circt-bmc", lowered_mlir]
+cmd += ["-b", str(args.depth or 20)]
+cmd += ["--module", args.test]
+cmd += ["--shared-libs", "/usr/lib/libz3.so"]
```
maerhart, line 36: Maybe get this path as an argument instead of hardcoding it?

tools/circt-test/circt-test.cpp
```diff
@@ -86,6 +86,11 @@ struct Options {
   cl::opt<std::string> runner{
       "r", cl::desc("Program to run individual tests"), cl::value_desc("bin"),
       cl::init("circt-test-runner-sby.py"), cl::cat(cat)};
+
+  cl::opt<bool> runnerReadsMLIR{
+      "mlir-runner",
+      cl::desc("Pass the MLIR file to the runner instead of Verilog"),
+      cl::init(false), cl::cat(cat)};
```
maerhart, line 93: Maybe make this an enum "runnerInputFormat" because BTOR2 could also be an option?

maerhart (approved):
This is very cool! It will be even better once we have the more direct lowering pipeline in place and don't need to rely on a python script in between.

maerhart:
Also, could you maybe add a small integration test for that?

tools/circt-test/circt-test.cpp
fabianschuiki: Good point. This option is currently just a hack to manually get the runners going. My idea was to have the runners themselves report what kind of input format they need, if the corresponding tool is installed on the system and available, and for which kinds of tests (formal, simulation, etc.) the runner is used for. This and the `-r` option will go away at that point, and will be replaced with a way for the user to pick a tool by name, or provide a list of tool preferences, e.g. `--tool sby,circt-bmc,vcf,jg`.

tools/circt-test/circt-test-runner-circt-bmc.py
fabianschuiki: That would be nice. I'm not sure how to do that exactly, though. The idea was to prevent the user from having to configure the runners. It must be possible to type `circt-test input.mlir` and just have all the tests in there run. So maybe this should be an environment variable, or part of a local configuration file somehow? Or maybe the runner script should just try a few common locations for Z3 and otherwise report that circt-bmc is not available. But this should definitely be configurable somehow, but maybe just not as a command line argument on `circt-test`.

tools/circt-test/circt-test-runner-circt-bmc.py
fabianschuiki: That's definitely an option. Although I'm also a bit hesitant to conflate test discovery and scheduling with the actual implementation of a formal verification tool. Maybe something like `circt-formal` or `circt-verify` could implement the collection of LEC, BMC, induction, and other proofs, and `circt-test` would just call out to that tool? That way we could get rid of the runner in favor of just calling the tools directly. But a separation may be valuable regardless.

tools/circt-test/circt-test-runner-circt-bmc.py
maerhart: Environment variable sounds even better if that's possible.

fabianschuiki: Done! Turned this into an environment variable. Funny how integration tests flush out missing features and find issues 😬

PR #7858, opened 2024-11-20
[HandshakeToDC] Getting some working tests

teqdruid:
- Adding port name propagation to HandshakeToDC
- Configurable port names for clock and reset
- Add ESIPrimitives.sv to the list of SV files
- Adding test runs for a few integration tests which look like they should work, but then disabling running if they don't.


teqdruid:
Many of the failing tests are due to a compile error: 

```
/home/jodemme/circt/integration_test/Dialect/Handshake/multiple_loops/multiple_loops.mlir:16:1: error: 'dc.fork' op DCToHW: value %3:3 = "dc.fork"(%arg1) : (!dc.token) -> (!dc.token, !dc.token, !dc.token) is unused.
func.func @top(%N: i64) -> (i64, i64) {
^
/home/jodemme/circt/integration_test/Dialect/Handshake/multiple_loops/multiple_loops.mlir:16:1: note: see current operation: %3:3 = "dc.fork"(%arg1) : (!dc.token) -> (!dc.token, !dc.token, !dc.token)
/home/jodemme/circt/integration_test/Dialect/Handshake/multiple_loops/multiple_loops.mlir:0:0: error: 'builtin.module' op DCToHW: failed to verify that all values are used exactly once. Remember to run the fork/sink materialization pass before HW lowering.
```

I am running that pass and I think it's nested correctly... am I not running it correctly?

mortbopet:
> I am running that pass and I think it's nested correctly... am I not running it correctly?

Looks correct to me. How does the IR look prior to `DCToHW`?

teqdruid:
@mortbopet Let's merge this with the TODOs in the tests to fix the fork issue. Dunno when I'll get around to fixing that bug, but I want to merge all my outstanding work before Thanksgiving weekend and I forget everything.

mortbopet (approved):

Issue #7821, opened 2024-11-15
[Calyx] Floating Point Comparisons


jiahanxie353:
I'm going to support the lowering of [`arith.maximumf`](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithcmpf-arithcmpfop) to Calyx dialect, for the sake of ReLU operations in neural networks. It's going to be a bit tricky since it involves ordered and unordered comparisons, where there might be NaNs in the latter case. 

What do you think about it? If we have the bit vector representation, it's easy to determine NaN by looking at the relevant bits; but how should Calyx determine if it's NaN given that we are passing regular decimal numbers as inputs (like 4.20)? Let's assume it's IEEE754 standard for now.

And I took some inspirations from the existing pass, [`-arith-expand`](https://mlir.llvm.org/docs/Passes/#-arith-expand) path, which legalizes operations such as `maximumf` to the operations that LLVM can support,  from:
```mlir
%2 = arith.maximumf %0, %1 : f32
```
we get:
```mlir
%2 = arith.cmpf ugt, %0, %1 : f32
%3 = arith.select %2, %0, %1 : f32
%4 = arith.cmpf uno, %1, %1 : f32
%5 = arith.select %4, %1, %3 : f32
```

If we have to use similar comparison operators, how does Calyx library ops support unordered comparisons?

Thoughts? @rachitnigam @andrewb1999 @cgyurgyik

jiahanxie353:
Can I tag @ekiwi as well given your expertise? Thanks!

cgyurgyik:
Either have a dedicated unordered/ordered comparison (based on some standard, eg ieee754) operations if the hardware supports it, or decompose it into a series of smaller operations

jiahanxie353:
I found this in the Berkeley HardFloat:
```verilog
module
    compareRecFN#(parameter expWidth, parameter sigWidth) (
        input [(expWidth + sigWidth):0] a,
        input [(expWidth + sigWidth):0] b,
        input signaling,
        output lt,
        output eq,
        output gt,
        output unordered,
        output [4:0] exceptionFlags
```

How does the following plan sound. 
1. Based on `compareRecFN`, we create three wrappers in `calyx/primitives/float` for IEEE754 standards:
    - `std_ltFN` by extracting `output lt` and `output unordered`;
    - `std_eqFN` by extracting `output eq` and `output unordered`;
    - `std_gtFN` by extracting `output gt` and `output unordered`.
[...]

PR #7860, opened 2024-11-21
[Calyx] Lower Arith CmpFOp to Calyx

jiahanxie353:
This patch tries to lowers `arith::cmpf` to Calyx. Relevant issue: https://github.com/llvm/circt/issues/7821

One thing I'm still trying to figure out is that, taking [`ult`](https://github.com/jiahanxie353/circt/blob/1ceae59a2921c63260b218bb20540b68b33053a0/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp#L820) as an example, I tried to create a `CombGroup` to evaluate `std_compareFN.lt || std_compareFN.unordered`.

My test case is:
```mlir
module {
  func.func @main(%arg0 : f32) -> i1 {
    %0 = arith.constant 4.2 : f32
    %1 = arith.cmpf ult, %arg0, %0 : f32

    return %1 : i1
  }
}
```

and I `dump` intermediate result after `BuildOpGroups`, it shows:
```mlir
"calyx.component"() <{function_type = (i32, i1, i1, i1, i1, i1) -> (), portAttributes = [{}, {clk}, {reset}, {go}, {}, {done}], portDirections = -16 : i6, portNames = ["in0", "clk", "reset", "go", "out0", "done"]}> ({
^bb0(%arg0: i32, %arg1: i1, %arg2: i1, %arg3: i1, %arg4: i1, %arg5: i1):
  %0 = "hw.constant"() <{value = true}> : () -> i1
  %1 = "hw.constant"() <{value = true}> : () -> i1
  %2:6 = "calyx.register"() <{sym_name = "cmpf_0_reg"}> : () -> (i1, i1, i1, i1, i1, i1)
  %3:3 = "calyx.std_or"() <{sym_name = "std_or_0"}> : () -> (i1, i1, i1)
  %4 = "hw.constant"() <{value = true}> : () -> i1
  %5:12 = "calyx.ieee754.compare"() <{sym_name = "std_compareFN_0"}> : () -> (i1, i1, i1, i32, i32, i1, i1, i1, i1, i1, i5, i1)
  %6:6 = "calyx.register"() <{sym_name = "ret_arg0_reg"}> : () -> (i1, i1, i1, i1, i1, i1)
  %7 = "calyx.constant"() <{sym_name = "cst_0", value = 4.200000e+00 : f32}> : () -> i32
  "calyx.wires"() ({
    "calyx.assign"(%arg4, %6#4) : (i1, i1) -> ()
    "calyx.comb_group"() <{sym_name = "compute_compare_0_out"}> ({
      "calyx.assign"(%3#0, %5#9) : (i1, i1) -> ()
      "calyx.assign"(%3#1, %5#6) : (i1, i1) -> ()
    }) : () -> ()
    "calyx.group"() <{sym_name = "bb0_0"}> ({
      "calyx.assign"(%5#3, %arg0) : (i32, i32) -> ()
      "calyx.assign"(%5#4, %7) : (i32, i32) -> ()
      "calyx.assign"(%2#0, %3#2) : (i1, i1) -> ()
      "calyx.assign"(%2#1, %5#11) : (i1, i1) -> ()
      %8 = "hw.constant"() <{value = true}> : () -> i1
      %9 = "comb.xor"(%5#11, %8) : (i1, i1) -> i1
      "calyx.assign"(%5#2, %1, %9) : (i1, i1, i1) -> ()
      "calyx.group_done"(%2#5) : (i1) -> ()
    }) : () -> ()
    "calyx.group"() <{sym_name = "ret_assign_0"}> ({
      "calyx.assign"(%6#0, %3#2) : (i1, i1) -> ()
      "calyx.assign"(%6#1, %0) : (i1, i1) -> ()
      "calyx.group_done"(%6#5) : (i1) -> ()
    }) : () -> ()
  }) : () -> ()
  "calyx.control"() ({
  ^bb0:
  }) : () -> ()
}) {sym_name = "main", toplevel} : () -> ()
```
so clearly we can see:
```mlir
    "calyx.comb_group"() <{sym_name = "compute_compare_0_out"}> ({
      "calyx.assign"(%3#0, %5#9) : (i1, i1) -> ()
      "calyx.assign"(%3#1, %5#6) : (i1, i1) -> ()
    }) : () -> ()
```

But after the entire execution, it finishes with:
```mlir
module attributes {calyx.entrypoint = "main"} {
  calyx.component @main(%in0: i32, %clk: i1 {clk}, %reset: i1 {reset}, %go: i1 {go}) -> (%out0: i1, %done: i1 {done}) {
    %cst = calyx.constant @cst_0 <4.200000e+00 : f32> : i32
    %true = hw.constant true
    %cmpf_0_reg.in, %cmpf_0_reg.write_en, %cmpf_0_reg.clk, %cmpf_0_reg.reset, %cmpf_0_reg.out, %cmpf_0_reg.done = calyx.register @cmpf_0_reg : i1, i1, i1, i1, i1, i1
    %std_or_0.left, %std_or_0.right, %std_or_0.out = calyx.std_or @std_or_0 : i1, i1, i1
    %std_compareFN_0.clk, %std_compareFN_0.reset, %std_compareFN_0.go, %std_compareFN_0.left, %std_compareFN_0.right, %std_compareFN_0.signaling, %std_compareFN_0.lt, %std_compareFN_0.eq, %std_compareFN_0.gt, %std_compareFN_0.unordered, %std_compareFN_0.exceptionalFlags, %std_compareFN_0.done = calyx.ieee754.compare @std_compareFN_0 : i1, i1, i1, i32, i32, i1, i1, i1, i1, i1, i5, i1
    %ret_arg0_reg.in, %ret_arg0_reg.write_en, %ret_arg0_reg.clk, %ret_arg0_reg.reset, %ret_arg0_reg.out, %ret_arg0_reg.done = calyx.register @ret_arg0_reg : i1, i1, i1, i1, i1, i1
    calyx.wires {
      calyx.assign %out0 = %ret_arg0_reg.out : i1
      calyx.group @bb0_0 {
        calyx.assign %std_compareFN_0.left = %in0 : i32
        calyx.assign %std_compareFN_0.right = %cst : i32
        calyx.assign %cmpf_0_reg.in = %std_or_0.out : i1
        calyx.assign %cmpf_0_reg.write_en = %std_compareFN_0.done : i1
        %0 = comb.xor %std_compareFN_0.done, %true : i1
        calyx.assign %std_compareFN_0.go = %0 ? %true : i1
        calyx.group_done %cmpf_0_reg.done : i1
      }
      calyx.group @ret_assign_0 {
        calyx.assign %ret_arg0_reg.in = %std_or_0.out : i1
        calyx.assign %ret_arg0_reg.write_en = %true : i1
        calyx.group_done %ret_arg0_reg.done : i1
      }
    }
    calyx.control {
      calyx.seq {
          calyx.enable @bb0_0
          calyx.enable @ret_assign_0
      }
    }
  } {toplevel}
}
```

There's no appearance of that combinational group. Any insight why could this happen? @cgyurgyik @andrewb1999 @rachitnigam


jiahanxie353:
> My first guess is this is being inlined because of [InlineCombGroups](https://github.com/llvm/circt/blob/bdf1cc94dbf81f116784c41a13a1ee75bbd528d5/include/circt/Dialect/Calyx/CalyxLoweringUtils.h#L715).

Makes sense, and I think I shouldn’t put them into comb groups in the first place since `std_compareFN.lt/unordered` are sequential. And does that mean I should make registers to hold `std_compareFN.lt/unordered`  values?

cgyurgyik:
> Makes sense, and I think I shouldn’t put them into comb groups in the first place since `std_compareFN.lt/unordered` are sequential. And does that mean I should make registers to hold `std_compareFN.lt/unordered` values?

Yes, if you want these to be read in sequential order then they need to be placed in registers.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -729,6 +731,164 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+
+    if (cmpf.getPredicate() == CmpFPredicate::AlwaysFalse) {
+      Value constantZero = createConstant(loc, rewriter, getComponent(), 1, 0);
+      cmpf.getResult().replaceAllUsesWith(constantZero);
+      return success();
+    }
+  }
+
+  // General case
+  StringRef opName = cmpf.getOperationName().split(".").second;
```
jiahanxie353, line 782: Maybe I should refactor with `buildLibraryBinaryPipeOp`, it's tricky though

cgyurgyik (commented):
A few smaller comments.

jiahanxie353:
thanks @cgyurgyik ! I've made changes based on your review

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -779,6 +779,26 @@ class BuildCallInstance : public calyx::FuncOpPartialLoweringPattern {
...
+    enum Port { EQ, GT, LT, UNORDERED } port;
+    // Whether we should invert the port before passing as inputs to the `op`
+    // field. For example, we should invert both the `lt` and the `unordered`
+    // port just extracted for predicate `oge`.
+    bool invert;
+  };
+
+  // The combinational logic to apply to the input ports. For example, we should
+  // apply `AND` logic to the two input ports for predicate `oge`.
+  enum CombLogic { AND, OR, SPECIAL } logic;
```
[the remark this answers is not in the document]
jiahanxie353: changed it to `noComb` (hopefully the intention that it stands for "no combinational logic" is more obvious)

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -779,6 +779,28 @@ class BuildCallInstance : public calyx::FuncOpPartialLoweringPattern {
...
+    Port port;
+    // Whether we should invert the port before passing as inputs to the `op`
+    // field. For example, we should invert both the `lt` and the `unordered`
+    // port just extracted for predicate `oge`.
+    bool invert;
+  };
+
+  // The combinational logic to apply to the input ports. For example, we should
+  // apply `logicAnd` to the two input ports for predicate `oge`.
+  enum class CombLogic { logicAnd, logicOr, noComb };
```
cgyurgyik, line 797: Suggestion: It might even be simpler to say None, i.e., `CombLogic::None` seems like a less verbose version of `CombLogic::noComb`. Usually the NONE case is placed first as well, since it corresponds to the integer 0.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -729,6 +730,172 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+  case CombLogic::logicOr: {
+    auto outputLibOp = getState<ComponentLoweringState>()
+                           .getNewLibraryOpInstance<calyx::OrLibOp>(
+                               rewriter, loc, {one, one, one});
+    rewriter.create<calyx::AssignOp>(loc, outputLibOp.getLeft(),
+                                     inputRegs[0].getOut());
+    rewriter.create<calyx::AssignOp>(loc, outputLibOp.getRight(),
+                                     inputRegs[1].getOut());
+
+    outputValue = outputLibOp.getOut();
```
cgyurgyik, line 860: should there be a `break` here in case a default / another case is added?

test/Conversion/SCFToCalyx/convert_simple.mlir
```diff
@@ -438,3 +438,705 @@ module {
   }
 }
 
+// -----
+
+// Test floating point OEQ
```
cgyurgyik, line 443: Let's create a new file `float_compare.mlir` or similar containing these tests.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -779,6 +779,28 @@ class BuildCallInstance : public calyx::FuncOpPartialLoweringPattern {
   ComponentOp getCallComponent(mlir::func::CallOp callOp) const;
```
cgyurgyik, line 779: Nice, looking much better; I've added a few smaller comments. I am curious if you've verified at all the output of this code besides manually checking it, e.g., have you run the emitted code through the native Calyx compiler and checked the outputs via some backend like Icarus or SysVerilog?

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
jiahanxie353: ah indeed, i was concerned `none` was a keyword or something

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: good point

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
jiahanxie353: I went to test in Calyx and still trying to debug. I have a `oge` test case and the lowered result is:
```mlir
    group bb0_0 {
      std_compareFN_0.left = 32'd1;
      std_compareFN_0.right = 32'd9;
      std_compareFN_0.signaling = 1'b1;
      compare_port_0_reg.write_en = std_compareFN_0.done;
      compare_port_0_reg.in = (std_compareFN_0.lt == 1'd0) ? 1'd1;
      unordered_port_0_reg.write_en = std_compareFN_0.done;
      unordered_port_0_reg.in = (std_compareFN_0.unordered == 1'd0) ? 1'd1;
      std_and_0.left = compare_port_0_reg.out;
      std_and_0.right = unordered_port_0_reg.out;
      std_and_1.left = compare_port_0_reg.done;
      std_and_1.right = unordered_port_0_reg.done;
      cmpf_0_reg.in = std_and_0.out;
      cmpf_0_reg.write_en = std_and_1.out;
      std_compareFN_0.go = !std_compareFN_0.done ? 1'd1;
      bb0_0[done] = cmpf_0_reg.done;
    }
```
but somehow `compare_port_0_reg.in = (std_compareFN_0.lt == 1'd0) ? 1'd1` always evaluates to `1` for some reason, even though `std_compareFN_0.left = 32'd1 < std_compareFN_0.right = 32'd9`(actually it's always `1` regardless of input...)

jiahanxie353: okay, `std_reg.in` is marked with `@data`, so guarded assignments like `compare_port_0_reg.in = (std_compareFN_0.lt == 1'd0) ? 1'd1;` will all be rewritten to `compare_port_0_reg.in =  1'd1;`, according to https://docs.calyxir.org/lang/attributes.html?highlight=data#data

jiahanxie353:
another round of revision based on the feedback, thanks so much @cgyurgyik for your timely review each time!

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -729,6 +730,201 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+  case CmpFPredicate::UGE:
+  case CmpFPredicate::ULT:
+  case CmpFPredicate::ULE:
+  case CmpFPredicate::OGT:
+  case CmpFPredicate::OGE:
+  case CmpFPredicate::OLT:
+  case CmpFPredicate::OLE:
+    signalingFlag = true;
+    break;
+  default:
```
cgyurgyik, line 815: Suggestion: Just add the other cases here with `signalingFlag = false;`, to future-proof for any changes to CmpFPredicate.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -729,6 +730,201 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+    break;
+  }
+
+  // The IEEE Standard mandates that equality comparisons ordinarily are quiet,
+  // while inequality comparisons ordinarily are signaling.
+  if (signalingFlag)
+    rewriter.create<calyx::AssignOp>(loc, calyxCmpFOp.getSignaling(), c1);
+  else
+    rewriter.create<calyx::AssignOp>(loc, calyxCmpFOp.getSignaling(), c0);
```
cgyurgyik, line 824:
```suggestion
  rewriter.create<calyx::AssignOp>(loc, calyxCmpFOp.getSignaling(), signalingFlag  ? c1 : c0);
```

lib/Dialect/Calyx/Transforms/CalyxLoweringUtils.cpp
```diff
@@ -843,5 +843,102 @@ BuildCallInstance::getCallComponent(mlir::func::CallOp callOp) const {
...
+  using CombLogic = PredicateInfo::CombLogic;
+  using Port = PredicateInfo::InputPorts::Port;
+  PredicateInfo info;
+  switch (pred) {
+  case CmpFPredicate::OEQ: {
+    return PredicateInfo{.logic = CombLogic::And,
+                         .inputPorts = {{Port::Eq, /*inverted=*/false},
+                                        {Port::Unordered, /*inverted=*/true}}};
+  }
+
```
cgyurgyik, line 856: Nit: remove these newlines between `case`s.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
cgyurgyik: Suggestion: minimize the predication to the field it affects.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -779,6 +779,28 @@ class BuildCallInstance : public calyx::FuncOpPartialLoweringPattern {
...
 };
 
+/// Predicate information for the floating point comparisons
```
cgyurgyik, line 782: Please update the description in the initial comment to be more declarative rather than "This patch tries to...", e.g.,

`Adds floating points comparison op with the Berkeley HardFloat interface.`

Also, if the issue below it is no longer relevant, I'd remove that too.

cgyurgyik (approved):
Nice job! LGTM after addressing a few small suggestions.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -729,6 +730,173 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+                  cmpf.getLhs().getType().getIntOrFloatBitWidth());
+  auto calyxCmpFOp = getState<ComponentLoweringState>()
+                         .getNewLibraryOpInstance<calyx::CompareFOpIEEE754>(
+                             rewriter, loc,
+                             {one, one, one, width, width, one, one, one, one,
+                              one, five, one});
+  hw::ConstantOp c1 = createConstant(loc, rewriter, getComponent(), 1, 1);
+  rewriter.setInsertionPointToStart(getComponent().getBodyBlock());
+
+  auto createSignalRegister =
```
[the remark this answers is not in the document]
cgyurgyik: I'm not saying you need to have a function in CalyxLoweringUtil. I'm suggesting it should be a class method, for a few reasons. 

1. You have 9 arguments to this function. That's a lot, regardless if its a lambda call. Can't this be reduced by making it a method, e.g., just call `getComponent()` instead of have a `component` argument?
2. Modular code is good; that's why functions exist. We could obviously inline everything into a `main` function, but then it can be quite difficult to grasp the high level understanding of the program. 
3. I really believe lambdas should be short and sweet. If they're large, they should be contained in their own function. This is personal opinion though, hence why it is a nit.



(You can also put this locally as a static function. See: https://llvm.org/docs/CodingStandards.html#anonymous-namespaces)

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
cgyurgyik: Here I'm saying that the following is more concise:

```cpp
foo(A, cond ? B : C);
```
than 

```cpp
if (cond)
  foo(A, B);
else
  foo(A, C);
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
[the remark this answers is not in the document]
jiahanxie353: makes senses, just changed that!

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
jiahanxie353: sure, I'll change the final squash commit message to be more informative as you suggested

lib/Dialect/Calyx/Transforms/CalyxLoweringUtils.cpp
```diff
@@ -843,5 +843,88 @@ BuildCallInstance::getCallComponent(mlir::func::CallOp callOp) const {
...
 
+PredicateInfo getPredicateInfo(CmpFPredicate pred) {
+  using CombLogic = PredicateInfo::CombLogic;
+  using Port = PredicateInfo::InputPorts::Port;
+  PredicateInfo info;
+  switch (pred) {
+  case CmpFPredicate::OEQ: {
+    return PredicateInfo{.logic = CombLogic::And,
+                         .inputPorts = {{Port::Eq, /*inverted=*/false},
+                                        {Port::Unordered, /*inverted=*/true}}};
```
uenoku, line 854: This designated initializer is c++ 20 feature so can you replace it with plain one?

Issue #7861, opened 2024-11-21
[SCFToCalyx] `buildLibraryBinaryPipeOp` should replace `TSrcOp`'s result with output with the register output


jiahanxie353:
When lowering the example:
```mlir
module {
  func.func @main(%arg0 : i32) -> i32 {
    %0 = arith.constant 4 : i32
    %1 = arith.muli %arg0, %0 : i32

    return %1 : i32
  }
}
```
current `--lower-scf-to-calyx` gives:
```mlir
module attributes {calyx.entrypoint = "main"} {
  calyx.component @main(%in0: i32, %clk: i1 {clk}, %reset: i1 {reset}, %go: i1 {go}) -> (%out0: i32, %done: i1 {done}) {
    %c4_i32 = hw.constant 4 : i32
    %true = hw.constant true
    %muli_0_reg.in, %muli_0_reg.write_en, %muli_0_reg.clk, %muli_0_reg.reset, %muli_0_reg.out, %muli_0_reg.done = calyx.register @muli_0_reg : i32, i1, i1, i1, i32, i1
    %std_mult_pipe_0.clk, %std_mult_pipe_0.reset, %std_mult_pipe_0.go, %std_mult_pipe_0.left, %std_mult_pipe_0.right, %std_mult_pipe_0.out, %std_mult_pipe_0.done = calyx.std_mult_pipe @std_mult_pipe_0 : i1, i1, i1, i32, i32, i32, i1
    %ret_arg0_reg.in, %ret_arg0_reg.write_en, %ret_arg0_reg.clk, %ret_arg0_reg.reset, %ret_arg0_reg.out, %ret_arg0_reg.done = calyx.register @ret_arg0_reg : i32, i1, i1, i1, i32, i1
    calyx.wires {
      calyx.assign %out0 = %ret_arg0_reg.out : i32
      calyx.group @bb0_0 {
        calyx.assign %std_mult_pipe_0.left = %in0 : i32
        calyx.assign %std_mult_pipe_0.right = %c4_i32 : i32
        calyx.assign %muli_0_reg.in = %std_mult_pipe_0.out : i32
        calyx.assign %muli_0_reg.write_en = %std_mult_pipe_0.done : i1
        %0 = comb.xor %std_mult_pipe_0.done, %true : i1
        calyx.assign %std_mult_pipe_0.go = %0 ? %true : i1
        calyx.group_done %muli_0_reg.done : i1
      }
      calyx.group @ret_assign_0 {
[...]

PR #7862, opened 2024-11-21
[SCFToCalyx] `buildLibraryBinaryPipeOp` source operation result replacement

jiahanxie353:
This patch fixes https://github.com/llvm/circt/issues/7861.


cgyurgyik (approved):
Nice find and fix!

Issue #7863, opened 2024-11-21
[HWToBTOR2] Builds are Flaky


seldridge:
All builds seem to have destabilized after one of the recent changes to HWToBTOR2. The CI windows builds seem to hit this most regularly, though I've experienced it on Linux and macOS.

An example of the failure is in this build log: https://github.com/llvm/circt/actions/runs/11945441683/job/33298106177

Here is a macOS failure:

```
******************** TEST 'CIRCT :: Conversion/HWToBTOR2/errors.mlir' FAILED ********************
Exit Code: 134

Command Output (stdout):
--
1 sort bitvec 32
2 constd 1 
--
Command Output (stderr):
--
RUN: at line 1: /Users/schuylereldridge/repos/github.com/llvm/circt/build/bin/circt-opt /Users/schuylereldridge/repos/github.com/llvm/circt/test/Conversion/HWToBTOR2/errors.mlir --convert-hw-to-btor2 --verify-diagnostics --split-input-file -o /dev/null
+ /Users/schuylereldridge/repos/github.com/llvm/circt/build/bin/circt-opt /Users/schuylereldridge/repos/github.com/llvm/circt/test/Conversion/HWToBTOR2/errors.mlir --convert-hw-to-btor2 --verify-diagnostics --split-input-file -o /dev/null
Assertion failed: (GetNumBytesInBuffer() == 0 && "Current buffer is non-empty!"), function SetBufferAndMode, file raw_ostream.cpp, line 115.
PLEASE submit a bug report to https://github.com/llvm/circt and include the crash backtrace.
 #0 0x0000000104cce984 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/Users/schuylereldridge/repos/github.com/llvm/circt/build/bin/circt-opt+0x1000da984)
 #1 0x0000000104cccea0 llvm::sys::RunSignalHandlers() (/Users/schuylereldridge/repos/github.com/llvm/circt/build/bin/circt-opt+0x1000d8ea0)
 #2 0x0000000104ccefd8 SignalHandler(int) (/Users/schuylereldridge/repos/github.com/llvm/circt/build/bin/circt-opt+0x1000dafd8)
 #3 0x0000000199356584 (/usr/lib/system/libsystem_platform.dylib+0x18047a584)
 #4 0x0000000199325c20 (/usr/lib/system/libsystem_pthread.dylib+0x180449c20)
[...]

PR #7864, opened 2024-11-21
[NFCI][HWToBTOR2] Avoid running HWToBTOR2 pass in parallel

fzi-hielscher:
Change the base operation for the HWToBTOR2 pass from `hw::HWModuleOp` to `mlir::ModuleOp`.

Running this pass in parallel on HWModules will just produce garbled output on the provided output stream. It is also likely the cause for the current CI failures (#7852, #7859, #7863). Multiple pass instances get instantiated with the same `raw_ostream` singleton returned by `llvm::outs()`. I do not think it is thread safe (why would it be?) and the CI stack traces point  quite clearly to a race condition in `raw_ostream::write`.

### DIFF
--- a/include/circt/Conversion/Passes.td
+++ b/include/circt/Conversion/Passes.td
@@ -582,7 +582,7 @@ def LowerHWToSV : Pass<"lower-hw-to-sv", "hw::HWModuleOp"> {
 // HWToBTOR2
 //===----------------------------------------------------------------------===//
 
-def ConvertHWToBTOR2 : Pass<"convert-hw-to-btor2", "hw::HWModuleOp"> {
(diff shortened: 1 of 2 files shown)


maerhart (approved):
LGTM! Interesting. that this didn't cause problems earlier.

The pass also seems to just assume there is only one `hw.module` in the IR and the pass class contains fields that are not reset which can cause issues if the class instance is reused. Seems like this pass needs some bigger cleanup at some point. But this is a good start 👍

dobios:
this is great, I am trying to get this to run on `verif.formal` in the future and hopefully produce a different output per `verif.formal` (ideally a tool would make this into multiple files per `verif.formal`) this is probably necessary for contracts to be useful. But this should work for now!

PR #7865, opened 2024-11-21
[SMT] Add push/pop operations

TaoBi22:
Adds push and pop ops to add/remove assertions from the assertion stack as per SMTLIB.

Unfortunately SMTLIB specifies that each of them takes a number of levels to push/pop but the Z3 API only provides such an argument for pop, so we have to make multiple calls for a big push (but I can't imagine that this will ever really be used in practice).


maerhart (approved):
A short integration test exercising the Z3 calls might be nice. But otherwise this looks fantastic! 🎉

TaoBi22:
> A short integration test exercising the Z3 calls might be nice.

Good point! Added now, thanks!

PR #7866, opened 2024-11-21
[HWToBTOR2] Error on variadic ops (for now)

TaoBi22:
Currently HWToBTOR2 (mostly) silently produces incorrect output whenever a variadic op is encountered as it assumes that there are always two operands. For the most part they should be pretty simple to support (with the exception of concat since you have to introduce some additional sorts) and I'll work on getting a PR open for that soon, but in the meantime we probably don't want this incorrect output silently produced.


dobios (approved):
LGTM!

PR #7867, opened 2024-11-21
[Verif] Make verif.formal parameters an ODS property

fabianschuiki:
Instead of using the discardable attributes dictionary of `verif.formal` as vessel for user-defined test parameters, add a dedicated `parameters` field to the op itself. This makes the parameters non-discardable and a proper part of the test.

### DIFF
--- a/include/circt/Dialect/Verif/VerifOps.td
+++ b/include/circt/Dialect/Verif/VerifOps.td
@@ -295,7 +295,7 @@ def FormalOp : VerifOp<"formal", [
 
     ### Example
     ```
-    verif.formal @AdderTest {
+    verif.formal @AdderTest {myParam = 42, myTag = "hello"} {
       %0 = verif.symbolic_value : i42
       %1 = verif.symbolic_value : i42
       %2 = hw.instance "dut" @Adder(a: %0: i42, b: %1: i42) -> (c: i42)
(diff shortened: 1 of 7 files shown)


include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -295,7 +295,7 @@ def FormalOp : VerifOp<"formal", [
 
     ### Example
```
maerhart, line 296: A short paragraph in the documentation describing what parameters mean for frontends and for backends would be great. E.g., what kind of parameters can a frontend add and what can it expect/what guarantees does it have? What do backends need to do with the parameters? Can they just ignore them at will?

maerhart (approved):
LGTM! Slightly more documentation would be nice 😁

include/circt/Dialect/Verif/VerifOps.td
fabianschuiki: 😁 Yeah we'll definitely need more details in the documentation. The problem right now is that we don't know yet which parameters are relevant or used. I expect us to figure this out as we build out the runners and the circt-test tool. For example, it looks like a "mode" and "bound" parameter makes sense for formal tests. But maybe also not really. My reflex here was to leave this out until we actually know which parameters are guaranteed to be understood by the runners, instead of promising something up front that then doesn't happen. 😬

PR #7868, opened 2024-11-21
[FIRRTL] FoldRegMems: insert new ops into same block as memory

rwy7:
Before this PR, FoldRegMems would construct new ops in the "body of the parent FModuleOp". We need to place these ops in the same block as the memory.

This PR fixes a bug where, when a memory under a layerblock was canonicalized to a register, the register would be placed at the original location of the memory (under the layerblock), but its readers would be placed outside the layerblock, resulting in a dominance checking error.

### DIFF
--- a/test/Dialect/FIRRTL/simplify-mems.mlir
+++ b/test/Dialect/FIRRTL/simplify-mems.mlir
@@ -408,10 +408,21 @@ firrtl.circuit "OneAddressNoMask" {
       in %in_rwen: !firrtl.uint<1>,
       out %result_read: !firrtl.uint<32>,
       out %result_rw: !firrtl.uint<32>) {
+    %c1_ui1 = firrtl.constant 1 : !firrtl.uint<1>
+    %Memory_read, %Memory_rw, %Memory_write = firrtl.mem Undefined
+      {
+        depth = 1 : i64,
+        name = "Memory",
+        portNames = ["read", "rw", "write"],
+        readLatency = 2 : i32,
+        writeLatency = 4 : i32
+      } :
+        !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data flip: uint<32>>,
+        !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, rdata flip: uint<32>, wmode: uint<1>, wdata: uint<32>, wmask: uint<1>>,
+        !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data: uint<32>, mask: uint<1>>
 
-    // Pipeline the inputs.
-    // TODO: It would be good to de-duplicate these either in the pass or in a canonicalizer.
-
+    // CHECK: %Memory = firrtl.reg %clock : !firrtl.clock, !firrtl.uint<32>
+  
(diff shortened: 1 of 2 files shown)


test/Dialect/FIRRTL/simplify-mems.mlir
```diff
@@ -497,3 +492,105 @@ firrtl.circuit "OneAddressNoMask" {
...
+      %write_addr = firrtl.subfield %Memory_write[addr] : !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data: uint<32>, mask: uint<1>>
+      firrtl.connect %write_addr, %addr : !firrtl.uint<1>, !firrtl.uint<1>
+      %write_en = firrtl.subfield %Memory_write[en] : !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data: uint<32>, mask: uint<1>>
+      firrtl.connect %write_en, %in_wen : !firrtl.uint<1>, !firrtl.uint<1>
+      %write_clk = firrtl.subfield %Memory_write[clk] : !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data: uint<32>, mask: uint<1>>
+      firrtl.connect %write_clk, %clock : !firrtl.clock, !firrtl.clock
+      %write_data = firrtl.subfield %Memory_write[data] : !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data: uint<32>, mask: uint<1>>
+      firrtl.connect %write_data, %in_data : !firrtl.uint<32>, !firrtl.uint<32>
+      %write_mask = firrtl.subfield %Memory_write[mask] : !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data: uint<32>, mask: uint<1>>
+      firrtl.connect %write_mask, %c1_ui1 : !firrtl.uint<1>, !firrtl.uint<1>
```
seldridge, line 593: Can any of these connections be dropped without compromising the integrity of the test?

test/Dialect/FIRRTL/simplify-mems.mlir
```diff
@@ -497,3 +492,105 @@ firrtl.circuit "OneAddressNoMask" {
...
+        {
+          depth = 1 : i64,
+          name = "Memory",
+          portNames = ["read", "rw", "write"],
+          readLatency = 2 : i32,
+          writeLatency = 2 : i32
+        } :
+          !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data flip: uint<32>>,
+          !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, rdata flip: uint<32>, wmode: uint<1>, wdata: uint<32>, wmask: uint<1>>,
+          !firrtl.bundle<addr: uint<1>, en: uint<1>, clk: clock, data: uint<32>, mask: uint<1>>
```
seldridge, line 530: The test may be able to be golfed down by only using a single port.

seldridge (approved):
Given that this canonicalizer already exists and is running, it makes sense to fix this up so that it works with layers.

That said, I agree that this is a pretty weird thing to be doing especially given the unknown expectations around memories and metadata.

seldridge:
I'm going to go ahead and merge this so that I can get the release going. Please loop back and take care of the review comments in post-merge.

PR #7870, opened 2024-11-21
[Verif] LowerContractsPass

leonardt:
* Adds pass to convert hw modules containing contracts into verif.formal tests with contracts inlined
  * ensure -> assert, require -> assume
* Adds tests for multiplier, carry save compressor, and shift left


lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+struct HWOpRewritePattern : public OpRewritePattern<HWModuleOp> {
+  using OpRewritePattern<HWModuleOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(HWModuleOp op,
+                                PatternRewriter &rewriter) const override {
+    auto formalOp = rewriter.create<verif::FormalOp>(
+        op.getLoc(), op.getNameAttr(), rewriter.getDictionaryAttr({}));
```
maerhart, line 54: What if there's no `verif.contract` op in the module? In that case we want to just leave the module as-is, right? Because the top-level of the test is already wrapped in a `verif.formal` and just arbitrarily splitting out this module and proving it separately won't work.

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    // Clone module body into fomal op body
```
maerhart, line 56:
```suggestion
    // Clone module body into formal op body
```

include/circt/Dialect/Verif/Passes.td
```diff
@@ -42,4 +42,12 @@ def LowerFormalToHWPass : Pass<"lower-formal-to-hw", "mlir::ModuleOp"> {
   }];
 }
 
+def LowerContractsPass : Pass<"lower-contracts", "mlir::ModuleOp"> {
+  let summary = "Lower contracts into formal tests";
+  let description = [{
+    Converts `hw.module` ops containing a `verif.contract` into a
+    `verif.formal` op.
```
maerhart, line 49: Might be nice to have a slightly more detailed explanation (and maybe an example) here (it's essentially already in the rationale, but it might be nice to repeat the most important points here.

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
+//===- LowerContracts.cpp - Formal Preparations --*- C++ -*----------------===//
```
maerhart, line 1:
```suggestion
//===- LowerContracts.cpp - Formal Preparations -----------------*- C++ -*-===//
```

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+};
+
+template <typename FROM, typename TO>
+void replaceContractOp(PatternRewriter &rewriter, Block *body) {
+  for (auto op : llvm::make_early_inc_range(body->getOps<FROM>())) {
+    auto label = op.getLabel();
+    StringAttr labelAttr;
+    if (label) {
+      labelAttr = rewriter.getStringAttr(label.value());
+    }
```
maerhart, line 42:
```suggestion
    StringAttr labelAttr;
    if (auto label = op.getLabel())
      labelAttr = rewriter.getStringAttr(label.value());

```

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace {
+struct LowerContractsPass
+    : verif::impl::LowerContractsPassBase<LowerContractsPass> {
+  void runOnOperation() override;
+};
+
+template <typename FROM, typename TO>
+void replaceContractOp(PatternRewriter &rewriter, Block *body) {
+  for (auto op : llvm::make_early_inc_range(body->getOps<FROM>())) {
```
maerhart, line 37: Might want a walk here or error out if there's an op with a region that could contain an op to convert.

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    // Inline contract ops
+    for (auto contractOp :
+         llvm::make_early_inc_range(bodyBlock->getOps<verif::ContractOp>())) {
+
+      // Convert ensure to assert, require to assume
+      rewriter.setInsertionPointToEnd(&contractOp.getBody().front());
+      Block *contractBlock = &contractOp.getBody().front();
+      replaceContractOp<EnsureOp, AssertOp>(rewriter, contractBlock);
+      replaceContractOp<RequireOp, AssumeOp>(rewriter, contractBlock);
```
maerhart, line 82: You're walking the block twice here, instead consider walking it once and checking for both conversions at the same time.

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+    bodyBlock->eraseArguments(0, bodyBlock->getNumArguments());
+
+    // Inline contract ops
+    for (auto contractOp :
+         llvm::make_early_inc_range(bodyBlock->getOps<verif::ContractOp>())) {
+
+      // Convert ensure to assert, require to assume
+      rewriter.setInsertionPointToEnd(&contractOp.getBody().front());
+      Block *contractBlock = &contractOp.getBody().front();
```
maerhart, line 80:
```suggestion
      Block *contractBlock = &contractOp.getBody().front();
      rewriter.setInsertionPointToEnd(contractBlock);
```

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+      replaceContractOp<EnsureOp, AssertOp>(rewriter, contractBlock);
+      replaceContractOp<RequireOp, AssumeOp>(rewriter, contractBlock);
+
+      // Inline body
+      rewriter.inlineBlockBefore(&contractOp.getBody().front(),
```
maerhart, line 85:
```suggestion
      rewriter.inlineBlockBefore(contractBlock,
```

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+                                 &formalOp.getBody().front(),
+                                 formalOp.getBody().front().end());
```
maerhart, line 87:
```suggestion
                                 bodyBlock,
                                 bodyBlock->end());
```

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+      // Replace results with inputs and erase
+      for (auto [input, result] :
+           llvm::zip(contractOp.getResults(), contractOp.getInputs())) {
+        rewriter.replaceAllUsesWith(input, result);
+      }
+      rewriter.eraseOp(contractOp);
+    }
+
+    rewriter.eraseOp(op);
```
maerhart, line 97: You cannot just delete the module here, instead you want to apply the contracts (i.e., assume they hold to simplify the body of the module).

lib/Dialect/Verif/Transforms/LowerContracts.cpp
```diff
@@ -0,0 +1,109 @@
...
+  }
+};
+} // namespace
+
+void LowerContractsPass::runOnOperation() {
+  RewritePatternSet patterns(&getContext());
+  patterns.add<HWOpRewritePattern>(patterns.getContext());
+
+  if (failed(applyPatternsAndFoldGreedily(getOperation(), std::move(patterns))))
+    signalPassFailure();
```
maerhart, line 108: Since you are only matching on modules, you can also just have a `for` loop iterating over the modules, which would be a bit more efficient.

test/Dialect/Verif/lower-contracts.mlir
maerhart, line 1: This file contains tests that are great as integration tests, but for regression tests we want to check the corner cases and keep the tests small (they don't have to do something that makes sense from a design implementation/logics perspective). For example, there's no test of a module with more than one contract inside, with zero contracts inside. Deleting the module also only works because you have no other module instantiating it.

fabianschuiki:
Really cool! :partying_face:

Instead of converting modules to `verif.formal` ops directly, it would be great to create a separate `verif.formal` op for each `verif.contract` op in the design. You could for example do a `mlirModule.walk([&](ContractOp op){ ... })` to find all contract ops in the design. Each could then go into its own `verif.formal` op. You could do a depth-first traversal starting at the contract and traversing its operands until you hit primary module inputs, and clone all ops that you encounter into the `verif.formal` op. There's a handy `IRMapper` that you can pass to `Operation::clone` that will automatically update operands for you :slightly_smiling_face:.

If a contract has another contract somewhere in its fan-in cone of operands, we should be able to just assume that that other contract holds by directly inlining it. Maybe there's an interleaving of steps that would do all of this, something like:

1. Collect all `verif.contract` ops into a worklist.
2. Replace uses of contract results with the contract body.
3. Outline all of the contracts into dedicated `verif.formal` ops, cloning all ops in their fan-in cone into the formal op.

I think that would get you the inlining of contracts everywhere, except for the one formal test that checks a contract, where we would check the actual design in the module against the contract body.

leonardt:
This is ready for another round of reviews.  It's been updated to handle more than one contract using a more general algorithm and also adds the logic to rewrite the hw module to assume the contracts hold.

include/circt/Dialect/Verif/Passes.td
```diff
@@ -42,4 +42,33 @@ def LowerFormalToHWPass : Pass<"lower-formal-to-hw", "mlir::ModuleOp"> {
...
+    Contracts are checked by turning `require`s into `assume`s and `ensure`s
+    into `asserts`s.  Inputs to the `hw.module` op are replaced with
+    `verif.symbolic_value` ops and other contracts are assumed to hold by
+    inlining the contract body.
+
+    Contracts are applied by turning `require`s into `assert`s and `ensure`s
+    into `assume`s in the original `hw.module` op body.  Results of the
+    contract are replaced with `verif.symbolic_value` ops.  In some cases,
+    using the `SimplifyAssumeEqPass` after lowering contracts and before
+    canonicalization and cse can result in a greatly simplify `hw.module` op
```
fabianschuiki, line 61:
```suggestion
    canonicalization and cse can result in a greatly simplified `hw.module` op
```

lib/Dialect/Verif/Transforms/SimplifyAssumeEq.cpp
```diff
@@ -0,0 +1,88 @@
...
+  LogicalResult matchAndRewrite(AssumeOp op,
+                                PatternRewriter &rewriter) const override {
+    auto prop = op.getProperty();
+    auto *definingOp = prop.getDefiningOp();
+    if (!definingOp)
+      return llvm::failure();
+
+    auto icmp = dyn_cast<comb::ICmpOp>(definingOp);
+    if (!icmp)
+      return llvm::failure();
```
fabianschuiki, line 64: Nit: getDefiningOp has a templated variant that does the dyn cast automatically if the op is non-null:
```suggestion
    auto icmp = prop.getDefiningOp<comb::ICmpOp>();
    if (!icmp)
      return llvm::failure();
```

lib/Dialect/Verif/VerifOpInterfaces.cpp
```diff
@@ -0,0 +1,15 @@
+//===- VerifOpInterfacfes.cpp - TODO -===//
```
fabianschuiki, line 1:
```suggestion
//===- VerifOpInterfacfes.cpp ---------------------------------------------===//
```

lib/Dialect/Verif/VerifOpInterfaces.cpp
```diff
@@ -0,0 +1,15 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file contains the Object Model operation interface definitions.
+//
+//===----------------------------------------------------------------------===//
```
fabianschuiki, line 11:
```suggestion
```

fabianschuiki (approved):
This looks great! 🥳 Thanks for fleshing this out in detail and adding all the tests 💯

lib/Dialect/Verif/Transforms/SimplifyAssumeEq.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+struct AssumeEqRewritePattern : public OpRewritePattern<verif::AssumeOp> {
+  using OpRewritePattern<AssumeOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AssumeOp op,
+                                PatternRewriter &rewriter) const override {
+    auto prop = op.getProperty();
+    auto icmp = prop.getDefiningOp<comb::ICmpOp>();
+    if (!icmp)
+      return llvm::failure();
```
uenoku, line 60: Do we need to check enable is always true?

fabianschuiki: Oh yeah great point, completely forgot about the enable. We could check if it isn't there at all. The canonicalizer should fold away constant-1 enables.

dobios:
Thank you so much for doing this! This is really great!! You can probably merge this with the PrepareForFormal pass now tbh

lib/Dialect/Verif/Transforms/SimplifyAssumeEq.cpp
leonardt: Hmm, in the current test I run canonicalization after the assume eq, would it make sense to run canonicalize before? Or before/after?  If we run it before, then we could just check if enable exists and return if so (so only simplify if enable doesn't exist)

PR #7871, opened 2024-11-21
[HandshakeToDC] Implement ESIInstanceOp lowering

teqdruid:
Lower the ESIInstanceOp to `hw.instance` with `dc.from_esi` and `dc.to_esi` to convert the args/results.


teqdruid:
Merging without review so that I can progress. This seems simple enough and will rely on post-vacation review from @mortbopet.

mortbopet (commented):
LGTM.

PR #7872, opened 2024-11-21
 [ExtractInstances] Append original instance name to path in metadata 

prithayan:
Update the metadata generated by `ExtractInstances`. 
Currently the path in the metadata, doesn't contain the original instance name, which causes difficulty in distinguishing extracted instances within the same module.
This PR updates the metadata to append the original instance name to the path.
Note that this instance name is of the instance, that has been extracted, hence it cannot be represented by the Hierarchical path.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
@@ -165,6 +165,7 @@ struct ExtractInstancesPass
   /// Cache of the inner ref to the new instances created. Will be used to
   /// create a path to the instance
   DenseMap<InnerRefAttr, InstanceOp> innerRefToInstances;
+  Type stringType, pathType;
 };
 } // end anonymous namespace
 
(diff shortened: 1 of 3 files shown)


mikeurbach (approved):
Seems straightforward to me.

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -1175,9 +1186,10 @@ void ExtractInstancesPass::createSchema() {
   mlir::Type portsType[] = {
       StringType::get(context), // name
       PathType::get(context),   // extracted instance path
-      StringType::get(context)  // filename
+      StringType::get(context), // filename
```
darthscsi, line 1189: Please call get once and reuse the value.  Type uniquing has costs we don't need to pay here.

prithayan: Thanks for spotting this, fixed it.

prithayan:
Thanks for the review, and Sorry for the delay in resolving this PR. I was waiting on some other changes to the client consuming the metadata files. Going to merge this after the CI passes.

PR #7873, opened 2024-11-22
[SMT] Add SMTLIB lowering for push & pop operations

TaoBi22:
Just adds an SMTLIB export for push/pop ops added in #7865


maerhart (approved):
Very nice! Thanks for the thorough tests! 🎉

PR #7876, opened 2024-11-22
[RTG] Add ElaborationPass

maerhart:
Add a pass that lowers the random parts of RTG IR.

### DIFF
--- a/include/circt/Dialect/RTG/Transforms/RTGPasses.h
+++ b/include/circt/Dialect/RTG/Transforms/RTGPasses.h
@@ -1,0 +1,31 @@
+//===- RTGPasses.h - RTG pass entry points ----------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This header file defines prototypes that expose pass constructors.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_DIALECT_RTG_TRANSFORMS_RTGPASSES_H
+#define CIRCT_DIALECT_RTG_TRANSFORMS_RTGPASSES_H
+
+#include "mlir/Pass/Pass.h"
+#include <memory>
+
+namespace circt {
+namespace rtg {
+
+/// Generate the code for registering passes.
+#define GEN_PASS_REGISTRATION
+#define GEN_PASS_DECL
+#include "circt/Dialect/RTG/Transforms/RTGPasses.h.inc"
+#undef GEN_PASS_REGISTRATION
+
+} // namespace rtg
+} // namespace circt
+
+#endif // CIRCT_DIALECT_RTG_TRANSFORMS_RTGPASSES_H
--- a/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
+++ b/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
@@ -1,0 +1,624 @@
+//===- ElaborationPass.cpp - RTG ElaborationPass implementation -----------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This pass elaborates the random parts of the RTG dialect.
+// It performs randomization top-down, i.e., random constructs in a sequence
+// that is invoked multiple times can yield different randomization results
+// for each invokation.
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/RTG/IR/RTGOps.h"
+#include "circt/Dialect/RTG/IR/RTGVisitors.h"
+#include "circt/Dialect/RTG/Transforms/RTGPasses.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/IR/IRMapping.h"
+#include "mlir/IR/PatternMatch.h"
+#include "llvm/Support/Debug.h"
+#include <deque>
+#include <random>
+
+namespace circt {
+namespace rtg {
+#define GEN_PASS_DEF_ELABORATIONPASS
+#include "circt/Dialect/RTG/Transforms/RTGPasses.h.inc"
+} // namespace rtg
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+using namespace circt::rtg;
+
+#define DEBUG_TYPE "rtg-elaboration"
+
+//===----------------------------------------------------------------------===//
+// Uniform Distribution Helper
+//
+// Simplified version of
+// https://github.com/llvm/llvm-project/blob/main/libcxx/include/__random/uniform_int_distribution.h
+// We use our custom version here to get the same results when compiled with
+// different compiler versions and standard libraries.
+//===----------------------------------------------------------------------===//
+
+static uint32_t computeMask(size_t w) {
+  size_t n = w / 32 + (w % 32 != 0);
+  size_t w0 = w / n;
+  return w0 > 0 ? uint32_t(~0) >> (32 - w0) : 0;
+}
+
+/// Get a number uniformly at random in the in specified range.
+static uint32_t getUniformlyInRange(std::mt19937 &rng, uint32_t a, uint32_t b) {
+  const uint32_t diff = b - a + 1;
+  if (diff == 1)
+    return a;
+
+  const uint32_t digits = std::numeric_limits<uint32_t>::digits;
+  if (diff == 0)
+    return rng();
+
+  uint32_t width = digits - llvm::countl_zero(diff) - 1;
+  if ((diff & (std::numeric_limits<uint32_t>::max() >> (digits - width))) != 0)
+    ++width;
+
+  uint32_t mask = computeMask(diff);
+  uint32_t u;
+  do {
+    u = rng() & mask;
+  } while (u >= diff);
+
+  return u + a;
+}
+
+//===----------------------------------------------------------------------===//
+// Elaborator Values
+//===----------------------------------------------------------------------===//
+
+namespace {
+
+/// The abstract base class for elaborated values.
+struct ElaboratorValue {
+public:
+  enum class ValueKind { Attribute, Set };
+
+  ElaboratorValue(ValueKind kind) : kind(kind) {}
+  virtual ~ElaboratorValue() {}
+
+  virtual llvm::hash_code getHashValue() const = 0;
+  virtual bool isEqual(const ElaboratorValue &other) const = 0;
+
+#ifndef NDEBUG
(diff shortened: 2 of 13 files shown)


lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,549 @@
...
+      return signalPassFailure();
+}
+
+void ElaborationPass::cloneTargetsIntoTests() {
+  auto moduleOp = getOperation();
+  for (auto target : llvm::make_early_inc_range(moduleOp.getOps<TargetOp>())) {
+    for (auto test : moduleOp.getOps<TestOp>()) {
+      // If the test requries nothing from a target, we can always run it.
+      if (test.getTarget().getEntryTypes().empty())
+        continue;
```
youngar, line 505: So if a test does not have any requirements we won't run it on each target, just once in general?

test/Dialect/RTG/Transform/elaboration.mlir
```diff
@@ -0,0 +1,123 @@
...
+^bb0(%arg0: !rtg.sequence):
+  %0 = rtg.sequence_closure @seq0
+  %set = rtg.set_create %arg0, %0 : !rtg.sequence
+  // expected-warning @below {{set contained 1 duplicate value(s), the value at index 0 might not be the intended one}}
+  %seq = rtg.set_select_random %set : !rtg.set<!rtg.sequence> {rtg.elaboration = 0}
+  rtg.invoke_sequence %seq
+  rtg.invoke_sequence %seq
+}
+
+// Test the set operations and passing a sequence ot another one via argument
```
youngar, line 19: "sequence ot"

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,549 @@
...
+  void runOnOperation() override;
+  void cloneTargetsIntoTests();
+
+private:
+  ElaborationOptions options;
+  Pass::Option<unsigned> seed{*this, "seed",
+                              llvm::cl::desc("Seed for the RNG.")};
+  Pass::Option<bool> debugMode{*this, "debug",
+                               llvm::cl::desc("Enable debug mode."),
+                               llvm::cl::init(false)};
```
youngar, line 479: Not sure what the best thing is, but I don't think it makes sense to store this duplicate information. Maybe just copy it out of the ElaborationOptions and in to the Pass::Options, which seems to be what happens for ODS declared options.

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,549 @@
...
+
+    if (debug)
+      for (auto [i, val] : llvm::enumerate(set))
+        debugMap[i] = val;
+
+    // Make sure the vector is sorted and has no duplicates.
+    llvm::sort(this->set);
+    this->set.erase(std::unique(this->set.begin(), this->set.end()),
+                    this->set.end());
+  }
```
youngar, line 102: SmallVector has a single-iterator version of erase: `this->set.erase(llvm::unique(this->set))`

youngar (commented):
Will keep looking tomorrow, its looking great! 😁

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
youngar: I don't really know whats best, what you had originally is probably good!

include/circt/Dialect/RTG/Transforms/RTGPasses.h
```diff
@@ -0,0 +1,49 @@
...
+namespace circt {
+namespace rtg {
+
+/// Options for the ElaborationPass
+struct ElaborationOptions {
+  /// The seed for any RNG constructs used in the pass. If `std::nullopt` is
+  /// passed, no seed is used and thus the IR after each invocation of this pass
+  /// will be different non-deterministically. However, if a nubmer is provided,
+  /// the pass is guaranteed to produce the same IR every time.
+  std::optional<unsigned> seed;
```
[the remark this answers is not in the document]
maerhart: Totally agree with this. Thanks for pointing it out!

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,520 @@
...
+inline llvm::hash_code hash_value(const ElaboratorValue &val) {
+  return val.getHashValue();
+}
+
+struct InternMapInfo : public DenseMapInfo<ElaboratorValue *> {
+  static unsigned getHashValue(const ElaboratorValue *value) {
+    auto *tk = getTombstoneKey();
+    auto *ek = getEmptyKey();
+    if (value == tk || value == ek)
+      return DenseMapInfo<ElaboratorValue *>::getHashValue(value);
```
youngar, line 202: should this just assert that we are not inserting the TK or EK?

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,568 @@
...
+                                op.getSet().getType());
+    return DeletionKind::Delete;
+  }
+
+  FailureOr<DeletionKind>
+  visitOp(SetSelectRandomOp op, function_ref<void(Operation *)> addToWorklist) {
+    auto *set = cast<SetValue>(state.at(op.getSet()));
+
+    size_t selected;
+    std::uniform_int_distribution<size_t> dist(0, set->getSet().size() - 1);
```
maerhart, line 348: This distribution is apparently not very well specified and may behave differently for different compilers and libraries. So it seems like I need to run my own distribution...

maerhart: Seems like it was a great idea to change to `rtg.elaboration_custom_seed` otherwise I probably wouldn't have noticed that for a long time.

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,568 @@
...
+
+    while (!worklist.empty()) {
+      auto *curr = worklist.back();
+      if (visited.contains(curr)) {
+        worklist.pop_back();
+        continue;
+      }
+
+      if (curr->getNumRegions() != 0)
+        return curr->emitError("nested regions not supported");
```
maerhart, line 414: Needs a test.

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,568 @@
...
+    if (op->hasTrait<OpTrait::ConstantLike>()) {
+      SmallVector<OpFoldResult, 1> result;
+      auto foldResult = op->fold(result);
+      (void)foldResult; // Make sure there is a user when assertions are off.
+      assert(succeeded(foldResult) &&
+             "constant folder of a constant-like must always succeed");
+      auto attr = dyn_cast<TypedAttr>(result[0].dyn_cast<Attribute>());
+      if (!attr)
+        return op->emitError(
+            "only typed attributes supported for constant-like operations");
```
maerhart, line 386: Should probably add a constant op with a non-typed attribute to `rtgtest` to test this.

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,568 @@
...
+                  function_ref<void(Operation *)> addToWorklist) {
+    // TODO: we only have this to be able to write tests for this pass without
+    // having to add support for more operations for now, so it should be
+    // removed once it is not necessary anymore for writing tests
+    if (op->use_empty()) {
+      for (auto &operand : op->getOpOperands()) {
+        Value val = materializer.materialize(state.at(operand.get()),
+                                             op->getBlock(), op->getLoc());
+        if (!val)
+          return op->emitError("failed to materialize value");
```
maerhart, line 323: I'd probably be a good idea to pass some diagnostic to the materializer and print more specific errors.

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,598 @@
...
+    auto *e = ptr.get();
+    auto [iter, _] = interned.insert({e, std::move(ptr)});
+    state[val] = iter->second.get();
+  }
+
+  /// Print a nice error message for operations we don't support yet.
+  FailureOr<DeletionKind>
+  visitUnhandledOp(Operation *op,
+                   function_ref<void(Operation *)> addToWorklist) {
+    return op->emitError("elaboration not supported");
```
fabianschuiki, line 340: Maybe `emitOpError`?

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,598 @@
...
+  for (auto testOp : moduleOp.getOps<TestOp>())
+    if (failed(elaborator.elaborate(testOp)))
+      return signalPassFailure();
+}
+
+void ElaborationPass::cloneTargetsIntoTests(SymbolTable &table) {
+  auto moduleOp = getOperation();
+  for (auto target : llvm::make_early_inc_range(moduleOp.getOps<TargetOp>())) {
+    for (auto test : moduleOp.getOps<TestOp>()) {
+      // If the test requries nothing from a target, we can always run it.
```
fabianschuiki, line 559:
```suggestion
      // If the test requires nothing from a target, we can always run it.
```

test/Dialect/RTG/Transform/elaboration.mlir
```diff
@@ -0,0 +1,54 @@
...
+  // CHECK-NEXT: }
+  %0 = arith.constant 2 : i32
+  %1 = arith.constant 3 : i32
+  %2 = arith.constant 4 : i32
+  %3 = arith.constant 5 : i32
+  %set = rtg.set_create %0, %1, %2, %0 : i32
+  %4 = rtg.set_select_random %set : !rtg.set<i32> {rtg.elaboration_custom_seed = 0}
+  %new_set = rtg.set_create %3, %4 : i32
+  %diff = rtg.set_difference %set, %new_set : !rtg.set<i32>
+  %5 = rtg.set_select_random %diff : !rtg.set<i32> {rtg.elaboration_custom_seed = 0}
```
fabianschuiki, line 24: Would it make sense to have custom seed values that are distinct from the `seed=0` passed to the `--rtg-elaborate` pass? To make sure that the correct seed value is used?

fabianschuiki (approved):
I'm not 100% familiar with the details of the RTG dialect, but the implementation makes sense to me :+1:

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -0,0 +1,614 @@
...
+      return iter->second;
+
+    auto [builderIter, _] =
+        builderPerBlock.insert({block, OpBuilder::atBlockBegin(block)});
+    OpBuilder builder = builderIter->second;
+
+    return TypeSwitch<ElaboratorValue *, Value>(val)
+        .Case<AttributeValue, SetValue>(
+            [&](auto val) { return visit(val, builder, loc, emitError); })
+        .Default([](auto val) { return Value(); });
```
maerhart, line 253: There should be an error message here.

PR #7877, opened 2024-11-22
[FIRRTL] Dedup: dedup private modules into public ones

### LINKED ISSUE #7875: [FIRRTL] Allow non-public modules to dedup with public modules
The FIRRTL spec allows for an amount of wiggle room on what can be done with deduplication and public modules. Currently, we don't do something as aggressive as we could. Consider the following:

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  module A:
  module B:
  module C:
  public module D:
  public module Foo:
    inst a of A
    inst b of B
    inst c of C
    inst d of D
```

Currently, only modules `A`, `B`, and `C` dedup. However, it is legal to have these dedup with `D`.

Modify the dedup pass to dedup `A`, `B`, and `C` into `D`. In the final Verilog, there should only be two modules.

You can make a call and do anything legal if there are multiple possible dedup targets. E.g., if there is `public module D` and `public module E`. This should incorporate the must dedup annotations to do what they say.

youngar:
This change allows private modules to deduplicate in to public modules.  It does this by swapping a private canonical module with a public module when found.  Future public modules will not deduplicate into this group, as public modules can not dedup with other public modules in general.  This also begins ignoring the convention of private modules, as it doesn't really carry any meaning.  In the future we might want to disallow conventions to be specified on private modules.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -51,14 +51,23 @@ using hw::InnerRefAttr;
 // Utility function for classifying a Symbol's dedup-ability.
 //===----------------------------------------------------------------------===//
 
+/// Returns true if the module can be removed.
 static bool checkVisibility(mlir::SymbolOpInterface symbol) {
-  // If module has symbol (name) that must be preserved even if unused,
-  // skip it. All symbol uses must be supported, which is not true if
-  // non-private.
-  // Special handling for class-like's and if symbol reports cannot be
-  // discarded.
-  return symbol.isPrivate() &&
-         (symbol.canDiscardOnUseEmpty() || isa<ClassLike>(*symbol));
+  // If the symbol is not private, it cannot be removed.
+  if (!symbol.isPrivate())
+    return false;
+  // Classes may be referenced in object types, so can not normally be removed
+  // if we can't find any symbol uses. Since we know that dedup will update the
+  // types of instances appropriately, we can ignore that and return true here.
+  if (isa<ClassLike>(*symbol))
+    return true;
+  // If module can not be removed even if no uses can be found, we can not
+  // delete it. The implication is that there are hidden symbol uses that dedup
+  // will not properly update.
+  if (!symbol.canDiscardOnUseEmpty())
+    return false;
+  // The module can be deleted.
+  return true;
 }
 
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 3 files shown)


rwy7 (approved):

test/Dialect/FIRRTL/dedup.mlir
```diff
@@ -37,6 +37,36 @@ firrtl.circuit "Simple" {
...
+  firrtl.module @PublicAndPrivateModulesDedup() {
+    // CHECK: firrtl.instance a @B()
+    // CHECK: firrtl.instance b @B()
+    // CHECK: firrtl.instance c @B()
+    firrtl.instance a @A()
+    firrtl.instance b @B()
+    firrtl.instance c @C()
+  }
+}
+
```
seldridge, line 69: Can the PR include tests of these three annotation-based dedup features and how this works with public/private:

1. `firrtl.transforms.MustDedupAnnotation`
2. `firrtl.transforms.NoDedupAnnotation`
3. `firrtl.transforms.DedupGroupAnnotation`

It's possible that the tests could be elided if there is sufficient coverage or the code paths don't warrant it.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -51,14 +51,23 @@ using hw::InnerRefAttr;
 // Utility function for classifying a Symbol's dedup-ability.
 //===----------------------------------------------------------------------===//
 
+/// Returns true if the module can be removed.
 static bool checkVisibility(mlir::SymbolOpInterface symbol) {
```
seldridge, line 55: Nit: renaming the function to better match what it does now would be helpful.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -1705,9 +1716,6 @@ class DedupPass : public circt::firrtl::impl::DedupBase<DedupPass> {
               ext && !ext.getDefname().has_value())
             return success();
 
-          if (!checkVisibility(module))
-            return success();
```
seldridge, line 1709: Something to keep an eye on: now that there is no early exit here, all modules will be hashed. This may affect performance, specifically for private classes.

test/Dialect/FIRRTL/dedup.mlir
youngar: I don't think the code paths really warrant more tests, the way they operate stays the same with this new feature. MustDedup is a post-transform check that the given list of modules deduped, it doesn't in any way force modules to dedup.  NoDedup modules are still never hashed.  The DedupGroup attribute is hashed like the rest of the attributes on a module, e.g. portTypes.

seldridge: Fantastic. 💯  Thanks for thinking through this.

PR #7878, opened 2024-11-22
[VerifToSMT] Only update registers on clock posedge

TaoBi22:
Fixes a bug in VerifToSMT where register states are updated on each timestep, rather than only on clock posedges. This requires locking the pass down to single-clock BMC ops for now, until we're able to associate register inputs/outputs with specific clocks.

### DIFF
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -274,6 +274,16 @@ struct VerifBoundedModelCheckingOpConversion
       }
     }
 
+    // TODO: this can be removed once we have a way to associate reg ins/outs
+    // with clocks
+    if (clockIndexes.size() > 1)
+      return op->emitError(
+          "only modules with one clock are currently supported");
+
+    // If we have no clocks then num_regs has to be 0
+    if (clockIndexes.size() == 0 && numRegs != 0)
+      return op->emitError("num_regs is non-zero but found no clock");
+
     auto numStateArgs = initVals.size() - initIndex;
     // Add the rest of the init vals (state args)
     for (; initIndex < initVals.size(); ++initIndex)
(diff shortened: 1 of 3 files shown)


lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -274,6 +274,16 @@ struct VerifBoundedModelCheckingOpConversion
...
 
+    // TODO: this can be removed once we have a way to associate reg ins/outs
+    // with clocks
+    if (clockIndexes.size() > 1)
+      return op->emitError(
+          "only modules with one clock are currently supported");
+
+    // If we have no clocks then num_regs has to be 0
+    if (clockIndexes.size() == 0 && numRegs != 0)
+      return op->emitError("num_regs is non-zero but found no clock");
```
maerhart, line 285: We shouldn't use `emitError` in conversion patterns. In theory, only if this pattern application fails doesn't mean that the whole pass fails. If. you. want error messages you need to do the checks in a walk before the conversion patterns are applied.

maerhart, line 285: That looks like something that should be verified by the BMC op, right?

test/Conversion/VerifToSMT/verif-to-smt.mlir
```diff
@@ -109,7 +109,12 @@ func.func @test_lec(%arg0: !smt.bv<1>) -> (i1, i1, i1) {
 // CHECK:        [[ORI:%.+]] = arith.ori [[SMTCHECK]], [[ARG5]]
 // CHECK:        [[LOOP:%.+]]:2 = func.call @bmc_loop([[ARG1]], [[ARG4]])
 // CHECK:        [[F2:%.+]] = smt.declare_fun : !smt.bv<32>
-// CHECK:        scf.yield [[LOOP]]#0, [[F2]], [[CIRCUIT]]#1, [[LOOP]]#1, [[ORI]]
+// CHECK:        [[OLDCLOCKLOW:%.+]] = smt.bv.not %arg1
```
maerhart, line 112: Super-nit: Inconsistent use of `%arg1` vs `[[ARG1]]`

maerhart (approved):
Thanks for fixing this!

lib/Conversion/VerifToSMT/VerifToSMT.cpp
TaoBi22: Ah cool, makes sense! I'll swap to it to a pre-application check.

TaoBi22: Yeah good point, I'll drop it here and add it to the verifier in another PR

test/Conversion/VerifToSMT/verif-to-smt.mlir
TaoBi22: Oops, thanks for the catch!

PR #7879, opened 2024-11-22
[FIRRTL] Rewrite, extend CheckLayers for GC Views

seldridge:
Change the InstanceInfo analysis to treat GrandCentral views as being
"under a layer".  This is the consistent, inclusive interpretation of
"under" with respect to the design.  I.e., since the
design-under-test (DUT) is under the design, a GrandCentral view should be
under a layer.

Rewrite the `CheckLayers` pass to more directly rely on the `InstanceInfo`
analysis.  Instead of the top-down walk (which was there from the original
architecture which did not have access to `InstanceInfo` and was
originally pushing layer information down), use only a bottom-up walk and
rely on the `InstanceInfo` analysis to determine when modules could be
problematic.  This preserves a similar style of the original verbose
errors/notes which reports transitive errors (modules which do not contain
layers, but instantiate modules that do), but switches to something that
is easier to work with.

Extend the `CheckLayers` pass to now properly handle Grand Central
companion modules.  These are conceptually the same as layers and are
now lumped under the "under layer" bucket by `InstanceInfo`.

### DIFF
--- a/lib/Analysis/FIRRTLInstanceInfo.cpp
+++ b/lib/Analysis/FIRRTLInstanceInfo.cpp
@@ -127,6 +127,7 @@ InstanceInfo::InstanceInfo(Operation *op, mlir::AnalysisManager &am) {
       if (isGCCompanion) {
         attributes.inDesign.mergeIn(false);
         attributes.inEffectiveDesign.mergeIn(false);
+        attributes.underLayer.mergeIn(true);
       }
 
       // Merge in values based on the instantiations of this module.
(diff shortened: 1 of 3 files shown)


seldridge:
Please give this a post-merge look. I'm getting ready to test a new release with this in it and I don't want to have https://github.com/llvm/circt/commit/33fae2fe39b72fe1822d313954862bf8c15bded8 without this patch. (That earlier commit removed some of the safety in Grand Central and this adds it back in.)

PR #7881, opened 2024-11-23
[FIRRTL] Make IMDCE work for ops w/ regions/blocks

seldridge:
Fix a bug in FIRRTL's IMDCE Pass where it would not visit the blocks of operations.  This can result in crashes for blocks which contain users of FIRRTL modules, e.g., instances inside layerblocks of sv.ifdef.

This conceptually is two changes: (1) when marking a block live, the block needs to be recursively walked and (2) when erasing ops, a recursive walk is needed.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
@@ -14,6 +14,7 @@
 #include "circt/Dialect/HW/InnerSymbolTable.h"
 #include "circt/Support/Debug.h"
 #include "mlir/IR/ImplicitLocOpBuilder.h"
+#include "mlir/IR/Iterators.h"
 #include "mlir/IR/Threading.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Pass/Pass.h"
@@ -240,15 +241,16 @@ void IMDeadCodeElimPass::markBlockExecutable(Block *block) {
   if (!executableBlocks.insert(block).second)
     return; // Already executable.
 
-  auto fmodule = cast<FModuleOp>(block->getParentOp());
-  if (fmodule.isPublic())
+  auto fmodule = dyn_cast<FModuleOp>(block->getParentOp());
+  if (fmodule && fmodule.isPublic())
     markAlive(fmodule);
 
   // Mark ports with don't touch as alive.
   for (auto blockArg : block->getArguments())
     if (hasDontTouch(blockArg)) {
       markAlive(blockArg);
-      markAlive(fmodule);
+      if (fmodule)
+        markAlive(fmodule);
     }
 
   for (auto &op : *block) {
@@ -261,8 +263,14 @@ void IMDeadCodeElimPass::markBlockExecutable(Block *block) {
     else if (isa<FConnectLike>(op))
       // Skip connect op.
       continue;
-    else if (hasUnknownSideEffect(&op))
+    else if (hasUnknownSideEffect(&op)) {
       markUnknownSideEffectOp(&op);
+      // Recursively mark any blocks contained within these operations as
+      // executable.
+      for (auto &region : op.getRegions())
+        for (auto &block : region.getBlocks())
+          markBlockExecutable(&block);
+    }
 
     // TODO: Handle attach etc.
   }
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
```diff
@@ -561,33 +569,35 @@ void IMDeadCodeElimPass::rewriteModuleBody(FModuleOp module) {
...
-      continue;
-    }
-
-    // Remove non-sideeffect op using `isOpTriviallyDead`.
-    if (mlir::isOpTriviallyDead(&op)) {
-      op.erase();
-      ++numErasedOps;
-    }
-  }
+  module.walk<mlir::WalkOrder::PostOrder, mlir::ReverseIterator>(
```
fabianschuiki, line 572: Nice use of iterators :sunglasses:

lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
```diff
@@ -264,6 +266,12 @@ void IMDeadCodeElimPass::markBlockExecutable(Block *block) {
     else if (hasUnknownSideEffect(&op))
       markUnknownSideEffectOp(&op);
 
+    // Recursively mark any blocks contained within these operations as
+    // executable.
+    for (auto &region : op.getRegions())
+      for (auto &block : region.getBlocks())
+        markBlockExecutable(&block);
```
fabianschuiki, line 273: This recursion can blow up the stack for extremely deeply-nested blocks. I don't think that's an actual issue in practice, especially since we probably parse these blocks with recursive function calls in the parser, so this doesn't add new scaling issues. WDYT?

fabianschuiki (approved):
LGTM! Nice mechanical change.

lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
seldridge: I was worried about this when reading it... It is narrowly, recursion over regions (which is allowed by the MLIR developer policy). However, I agree that it is risky to do it this way. I can see about rewriting this to be iterative (here or in a follow-on).

youngar, line 273: I think we should move this under `hasUnknownSideEffect` op.  I consider this the conservative codepath for when we don't know what an operation is. Layerblocks fall under this umbrella.

If we ever had an operation like this totally made up example:
```
sv.ifdef 0 {
  sv.assert(clock, 0)
}
```
We would not want to unconditionally visit its regions.

youngar (approved):
Feel free to ignore the comments, I'm not sure they are valid.

lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
fabianschuiki: :+1: Using recursion on blocks feels fine :+1:

lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
seldridge: I really like not having to handle this manually and also not having to worry about futre-proofing it.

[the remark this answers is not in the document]
seldridge: It could be, however, this would require us to better define whether or not all the ops in the FIRRTL dialect have side effects. Consider the following example:

``` mlir
firrtl.circuit "Foo" {
  firrtl.layer @A bind {
  }
  firrtl.module private @Bar(in %clock: !firrtl.clock, in %a: !firrtl.uint<1>, out %b: !firrtl.probe<uint<1>, @A>, out %c: !firrtl.uint<1>) {
    %0 = firrtl.not %a : (!firrtl.uint<1>) -> !firrtl.uint<1>
    firrtl.matchingconnect %c, %0 : !firrtl.uint<1>
    firrtl.layerblock @A {
      %1 = firrtl.not %a : (!firrtl.uint<1>) -> !firrtl.uint<1>
      %a_not = firrtl.node %1 : !firrtl.uint<1>
      %2 = firrtl.ref.send %a_not : !firrtl.uint<1>
      %3 = firrtl.ref.cast %2 : (!firrtl.probe<uint<1>>) -> !firrtl.probe<uint<1>, @A>
      firrtl.ref.define %b, %3 : !firrtl.probe<uint<1>, @A>
    }
  }
  firrtl.module @Foo(in %clock: !firrtl.clock, in %a: !firrtl.uint<1>, out %c: !firrtl.uint<1>) attributes {convention = #firrtl<convention scalarized>} {
    %bar_clock, %bar_a, %bar_b, %bar_c = firrtl.instance bar @Bar(in clock: !firrtl.clock, in a: !firrtl.uint<1>, out b: !firrtl.probe<uint<1>, @A>, out c: !firrtl.uint<1>)
    firrtl.matchingconnect %bar_clock, %clock : !firrtl.clock
    firrtl.matchingconnect %bar_a, %a : !firrtl.uint<1>
    firrtl.matchingconnect %c, %bar_c : !firrtl.uint<1>
  }
}
```

This has a path through a `layerblock` via a layer-colored probe. I have an out-of-tree change that I will push that makes `layerblock` ops `RecursiveMemoryEffects` and `RecursivelySpeculatable` (which they arguably should be). With that change, if I do a post-order walk, then I will delete the `layerblock` op since it is trivially dead after all the operations are removed (which are not side effect free as defined in ODS). If I do a pre-order walk, then I will not delete the `layerblock` op.

Given the way that we've defined FIRRTL operations, the post-order walk seems to be the only way to do this if we want to delete the `layerblock` as part of this pass as opposed to relying on something else to delete it, e.g., a canonicalizer.

h/t @rwy7 for working through this example.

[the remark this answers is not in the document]
seldridge: Change to add the missing traits to `layerblock`s here: https://github.com/llvm/circt/pull/7881/commits/8f8acc233e87fc745500780de0aa0b02960abf7b

lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
```diff
@@ -240,15 +241,16 @@ void IMDeadCodeElimPass::markBlockExecutable(Block *block) {
...
+  if (fmodule && fmodule.isPublic())
     markAlive(fmodule);
 
   // Mark ports with don't touch as alive.
   for (auto blockArg : block->getArguments())
     if (hasDontTouch(blockArg)) {
       markAlive(blockArg);
-      markAlive(fmodule);
+      if (fmodule)
+        markAlive(fmodule);
```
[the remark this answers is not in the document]
seldridge: What would you say are the defined semantics of `DontTouchAnnotation`? 😉 

From an annotation perspective, this can only be applied to certain named things (modules, wires, registers, memories, and ports). This annotation has the semantics of, when applied to a module, is equivalent to don't touching all the ports. Given that interpretation, then this is doing the right thing. Or: it's reasonable to expect that this is a module port and not a general block arg as this was never intended to be defined for block args, just for ports.

That said, this annotation has been _begging_ for a better in-IR representation and we've been skirting around this for a while. The best thing to do is to fit this into the FIRRTL spec and then figure out how to represent it here as opposed to trying to generalize don't touch on ports to don't touch on block args (as nobody is asking for that).

lib/Dialect/FIRRTL/Transforms/IMDeadCodeElim.cpp
seldridge: I think this seems right. It seems like the better thing would be to make this the default for any unknown operation and not just `hasUnknownSideEffect`. That is different from how the PR is written, though, and means that there should be early exits from all the prior `if`/`else if` clauses.

I.e., the original problem with this pass was that unless this is the fallback _for all unknown operations_ then the pass will incorrectly delete declarations without deleting their uses.

Switched to your suggestion in: https://github.com/llvm/circt/pull/7881/commits/c4d490a51d536bfdd11cf11ab275522de081f32e

PR #7882, opened 2024-11-23
[RTG][RTGTest] Add CAPI and a basic lowering pipeline

maerhart:
We need a CAPI for the Python entry


fabianschuiki (approved):
This looks like a nice continuation of the `circt-capi-*-test` pattern :partying_face: Awesome that you're testing the CAPI thoroughly :100:. API LGTM.

PR #7884, opened 2024-11-24
[circt-test] fix lit config for circt-bmc

unlsycn:
#7857 introduces a Python script to test `circt-bmc`, but there are two problems:
1. We need to add the script to the substitution list in `lit.cfg.py` so that lit can correctly substitute the path in the test.
2. The script invokes `circt-opt` but the executable is not in the PATH variable causing python cannot find it. This PR simply adds `circt_tool_dir` to PATH to resolve it, but I think a better way is to get rid of the python script and allow lit to substitute the path.

### DIFF
--- a/integration_test/lit.cfg.py
+++ b/integration_test/lit.cfg.py
@@ -79,8 +79,9 @@ tool_dirs = [
 ]
 tools = [
     'arcilator', 'circt-opt', 'circt-translate', 'firtool', 'circt-rtl-sim.py',
-    'equiv-rtl.sh', 'handshake-runner', 'hlstool', 'kanagawatool', 'circt-lec',
-    'circt-bmc', 'circt-test', 'circt-test-runner-sby.py'
+    'equiv-rtl.sh', 'handshake-runner', 'hlstool', 'ibistool', 'circt-lec',
+    'circt-bmc', 'circt-test', 'circt-test-runner-sby.py',
+    'circt-test-runner-circt-bmc.py'
 ]
 
 # Enable python if its path was configured
@@ -230,6 +231,7 @@ if config.arcilator_jit_enabled:
   config.available_features.add('arcilator-jit')
 
 config.substitutions.append(('%driver', f'{config.driver}'))
+config.substitutions.append(('%circt-tools-dir', f'{config.circt_tools_dir}'))
 llvm_config.add_tool_substitutions(tools, tool_dirs)
 
 # cocotb availability


unlsycn:
cc @fabianschuiki

unlsycn:
One confusing point is that the Github CI can pass this test without paht substitution and the additional PATH variable, which seems to be an existing CIRCT binary path in the PATH.
https://github.com/llvm/circt/actions/runs/11998261862/job/33447711728

fabianschuiki:
Hmmm, this is interesting. The `circt-test-runner-*.py` scripts are installed alongside the `circt-test` executable. But you're right, if `circt-test` is expanded by lit instead of found through the PATH, then the scripts, which are searched for in the PATH, won't be found either. Can we tweak lit to add the binary build directory to the PATH, such that we don't have to muck with the PATH variable in RUN lines?

unlsycn:
> Ah sorry, I just now realized that this is for `circt-opt`, not the benchmark script. In the lit config, it should be possible to add the CIRCT tool build path to the `config.path` list, which will be used to populate the PATH variable. That might fix things.

Yeah, it's a possible fix that I wasn't sure made sense before. Maybe we should just consider it a workaround to work with the Python script rather than using it to replace the substitution flow.

integration_test/lit.cfg.py
```diff
@@ -65,6 +65,7 @@
 
 # Tweak the PATH to include the tools dir.
 llvm_config.with_environment('PATH', config.llvm_tools_dir, append_path=True)
+config.environment['PATH'] = os.pathsep.join([config.circt_tools_dir, config.environment['PATH']])
```
fabianschuiki, line 68: Does the following work as well?
```suggestion
config.path.append(config.circt_tools_dir)
```

unlsycn: TestingConfig should not have this memver?

fabianschuiki: Never mind, I think I was looking at a very old version of lit. Just found the following in LLVM though:
https://github.com/llvm/llvm-project/blob/main/llvm/test/lit.cfg.py#L38-L39
```suggestion
# Tweak the PATH to include the tools dir.
llvm_config.with_environment("PATH", config.llvm_tools_dir, append_path=True)
```
Seems like we're in good company.

unlsycn: My fault, I just misunderstood the meaning of `llvm_config.with_environmet'.

fabianschuiki (approved):
Really nice, thanks!

fabianschuiki:
Would you like to merge this yourself, or do you want me to do it?

unlsycn:
> Would you like to merge this yourself, or do you want me to do it?
I'd be really grateful if you could help me merge them for now. Thanks so much!

unlsycn:
Oh..sorry, my wrong rebase screwed everything up!

PR #7887, opened 2024-11-24
[RTG] Add BagType and operations

maerhart:


lib/Dialect/RTG/IR/RTGOps.cpp
```diff
@@ -78,6 +78,78 @@ LogicalResult SetCreateOp::verify() {
...
+    p << "(";
+  llvm::interleaveComma(llvm::zip(getElements(), getMultiples()), p,
+                        [&](auto elAndMultiple) {
+                          auto [el, multiple] = elAndMultiple;
+                          p << multiple << " x " << el;
+                        });
+  if (!getElements().empty())
+    p << ")";
+
+  p.printOptionalAttrDict((*this)->getAttrs());
```
uenoku, line 138: nit: Is it intentional that position of optional attribute is different from other operations (bag_selecet and bag_random)?

uenoku (approved):
LGTM

PR #7889, opened 2024-11-25
[RTG] Add some additional debug printing

maerhart:
Also, preserve the set order when computing set differences.


lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -73,6 +74,15 @@ class ElaboratorValue {
...
   }
 
+  virtual std::string toString() const {
+    std::string out;
+    llvm::raw_string_ostream stream(out);
+    stream << "<opaque ";
+    value.print(stream);
+    stream << " at " << this << ">";
+    return out;
+  }
```
seldridge, line 84: Can this be done by defining the `<<` operator. This can also, if only used in debug, can be guarded under `NDEBUG`. E.g., something like https://github.com/llvm/circt/blob/main/include/circt/Analysis/FIRRTLInstanceInfo.h#L198

maerhart: That's a great point! Thanks for reviewing!

maerhart:
Closing this because everything here was integrated into https://github.com/llvm/circt/pull/7876

PR #7890, opened 2024-11-25
[RTG][Elaboration] Add support for arith constants

maerhart:
This prepares for `!rtg.bag` support in the elaboration pass which requires us to interpret `index` values.

I've added a visitor for Arith ops to the RTG dialect since this visitor pattern is not used upstream. It probably doesn't make sense to upstream this due to what @youngar mentioned in https://github.com/llvm/circt/pull/7855#pullrequestreview-2449567773

### DIFF
--- a/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
+++ b/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
@@ -13,6 +13,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "circt/Dialect/RTG/IR/ArithVisitors.h"
 #include "circt/Dialect/RTG/IR/RTGOps.h"
 #include "circt/Dialect/RTG/IR/RTGVisitors.h"
 #include "circt/Dialect/RTG/Transforms/RTGPasses.h"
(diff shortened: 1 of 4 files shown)


lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -431,6 +501,15 @@ class Elaborator : public RTGOpVisitor<Elaborator, FailureOr<DeletionKind>,
     return DeletionKind::Delete;
   }
 
+  FailureOr<DeletionKind>
+  dispatchOpVisitor(Operation *op,
+                    function_ref<void(Operation *)> addToWorklist) {
+    if (op->getDialect() == op->getContext()->getLoadedDialect<RTGDialect>())
```
darthscsi, line 507: Sigh.  I don't suppose we can get around this.

darthscsi (approved):
Please upstream the  ArithOpVisitor.  Otherwise LGTM.

youngar (approved):

maerhart:
This was superseded by index dialect support and generalized constant-like support.

PR #7892, opened 2024-11-25
[RTG][Elaboration] Elaboration support for Bags

maerhart:


fabianschuiki (approved):
I'm not too familiar with the RTG details, but the code and tests look good to me! :+1:

PR #7893, opened 2024-11-25
[MooreToCore] Lower the unpacked array type to `hw.array`

slowlime:
This add a type conversion for unpacked arrays, mapping them to `hw.array`. The approach is analogous to how unpacked structs are currently handled.

As far as I understand, `hw.uarray` would be more appropriate in the future, especially once it's moved to SV. However, no operation accepts this type yet. I'm not sure what the consensus is as to whether the operation set should be extended to accommodate unpacked arrays (by adding new operations or "overloading" existing ones that deal with `hw.array`s). Using packed arrays for the time being sidesteps this question.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -1447,6 +1447,16 @@ static void populateTypeConversion(TypeConverter &typeConverter) {
     return {};
   });
 
+  // FIXME: Unpacked arrays support more element types than their packed
+  // variants, and as such, mapping them to hw::Array is somewhat naive. See
+  // also the analogous note below concerning unpacked struct type conversion.
+  typeConverter.addConversion(
+      [&](UnpackedArrayType type) -> std::optional<Type> {
+        if (auto elementType = typeConverter.convertType(type.getElementType()))
+          return hw::ArrayType::get(elementType, type.getSize());
+        return {};
+      });
+
   typeConverter.addConversion([&](StructType type) -> std::optional<Type> {
     SmallVector<hw::StructType::FieldInfo> fields;
     for (auto field : type.getMembers()) {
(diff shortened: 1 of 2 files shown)


fabianschuiki (approved):
LGTM! Thanks for adding this :+1:. I like the idea of just treating unpacked arrays as packed for now, and then build out the subtle differences later on.

fabianschuiki:
Feel free to get [LLVM commit access](https://llvm.org/docs/DeveloperPolicy.html#obtaining-commit-access) such that the CI builds for your PRs run automatically and you can merge your own PRs one you've gotten reviews and are happy with everything :slightly_smiling_face:.

hailongSun2000:
Hey, @slowlime. I ever attempted to convert `moore.unpacked_array` into `hw.array`. But I failed if I remember correctly. Maybe you can test `logic [3:0][7:0] uarr [0:3][0:7];`.

slowlime:
> I ever attempted to convert `moore.unpacked_array` into `hw.array`. But I failed if I remember correctly. Maybe you can test `logic [3:0][7:0] uarr [0:3][0:7];`.

@hailongSun2000 Hm, that's interesting because the type conversion itself succeeds even given nested arrays — I've just pushed a commit to test this too. (FWIW, I've also tried running `circt-verilog` on a SV module containing the code you've suggested, along with stuff to prevent it from getting DCE'd, and found it works fine.)

On the other hand, multidimensional arrays don't seem usable currently because `moore.extract` is translated to `hw.array_slice` instead of `hw.array_get` (#7918).

hailongSun2000:
Yeah, great catch :smiley:. The `moore. unpacked_array` is unavailable now in `moore.extract(_ref)`.

slowlime:
Hey, I was wondering if you could merge this, please? I'm not sure if I want commit access for myself yet.

PR #7898, opened 2024-11-26
[circt-bmc] Drop outdated integration test

TaoBi22:
This test was made for an outdated version of circt-bmc and only passes now because of a bug (fix incoming!)


maerhart (approved):
LGTM!

PR #7899, opened 2024-11-26
[MooreToCore] Support pows and powu op

### LINKED ISSUE #7626: [MooreToCore] Support pow op
Add support for the `moore.pow*` ops in MooreToCore: https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+11.4.3+binary_op_pow

Max-astro:
Fixes #7626

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -1174,6 +1174,84 @@ struct ShrOpConversion : public OpConversionPattern<ShrOp> {
   }
 };
 
+struct PowUOpConversion : public OpConversionPattern<PowUOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(PowUOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Type resultType = typeConverter->convertType(op.getResult().getType());
+
+    Location loc = op.getLoc();
+    auto intType = cast<IntType>(op.getRhs().getType());
+
+    // transform a ** b into scf.for 0 to b step 1 { init *= a }, init = 1
+    Type integerType = rewriter.getIntegerType(intType.getWidth());
+    Value lowerBound = rewriter.create<hw::ConstantOp>(loc, integerType, 0);
+    Value upperBound =
+        rewriter.create<ConversionOp>(loc, integerType, op.getRhs());
+    Value step = rewriter.create<hw::ConstantOp>(loc, integerType, 1);
+
+    Value initVal = rewriter.create<hw::ConstantOp>(loc, resultType, 1);
+    Value lhsVal = rewriter.create<ConversionOp>(loc, resultType, op.getLhs());
+
+    auto forOp = rewriter.create<scf::ForOp>(
+        loc, lowerBound, upperBound, step, ValueRange(initVal),
+        [&](OpBuilder &builder, Location loc, Value i, ValueRange iterArgs) {
+          Value loopVar = iterArgs.front();
+          Value mul = rewriter.create<comb::MulOp>(loc, lhsVal, loopVar);
+          rewriter.create<scf::YieldOp>(loc, ValueRange(mul));
+        });
+
+    rewriter.replaceOp(op, forOp.getResult(0));
+
+    return success();
+  }
+};
+
+struct PowSOpConversion : public OpConversionPattern<PowSOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(PowSOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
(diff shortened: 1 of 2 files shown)


Max-astro:
Hi @fabianschuiki , I have incorporated the advice from you and @maerhart . Please let me know if there are any issues with my code or the test cases.

Thank you so much for taking the time to review this. I really appreciate your feedback! :)

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1174,6 +1174,111 @@ struct ShrOpConversion : public OpConversionPattern<ShrOp> {
...
+
+    Location loc = op.getLoc();
+    auto intType = cast<IntType>(op.getRhs().getType());
+    if (auto baseOp = op.getLhs().getDefiningOp<ConstantOp>()) {
+      if (baseOp.getValue() == 2) {
+        Value constOne = rewriter.create<ConstantOp>(loc, intType, 1);
+        rewriter.replaceOpWithNewOp<ShlOp>(op, constOne, op.getRhs());
+        return success();
+      }
+    }
```
fabianschuiki, line 1193: This would be fantastic to have as a canonicalization in the Moore dialect! Could you move this into the canonicalizer of `moore::PowUOp`? (You might have to add a `let hasCanonicalizeMethod = 1` to the `PowUOp` definition in the `MooreOps.td` file.)

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1174,6 +1174,111 @@ struct ShrOpConversion : public OpConversionPattern<ShrOp> {
...
+            op, op.getLhs().getType(), isNegative);
+        Block *thenBlock = rewriter.createBlock(&condOp.getTrueRegion());
+        rewriter.setInsertionPointToStart(thenBlock);
+        rewriter.create<YieldOp>(loc, constZero);
+        Block *elseBlock = rewriter.createBlock(&condOp.getFalseRegion());
+        rewriter.setInsertionPointToStart(elseBlock);
+        rewriter.create<YieldOp>(loc, shift);
+        return success();
+      }
+    }
```
fabianschuiki, line 1245: This would be fantastic to have as a canonicalization in the Moore dialect! Could you move this into the canonicalizer of `moore::PowSOp`? (You might have to add a `let hasCanonicalizeMethod = 1` to the `PowSOp` definition in the `MooreOps.td` file.)

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1174,6 +1174,111 @@ struct ShrOpConversion : public OpConversionPattern<ShrOp> {
...
+                                                     rhsVal, constZero);
+
+    // if the exponent is negative, lhsVal = 1 / lhs, upperBound = -rhs
+    Value invLhs = rewriter.create<comb::DivUOp>(
+        loc, rewriter.create<hw::ConstantOp>(loc, resultType, 1), lhsVal);
+    Value negRhs = rewriter.create<comb::SubOp>(loc, constZero, rhsVal);
+
+    lhsVal = rewriter.create<comb::MuxOp>(loc, isNegative, invLhs, lhsVal);
+    Value upperBound =
+        rewriter.create<comb::MuxOp>(loc, isNegative, negRhs, rhsVal);
```
fabianschuiki, line 1262: If the exponent is negative, can this ever produce anything besides 0? Since these are integers, I'd assume `1 / lhs` to always return 0, since there are no fractional numbers. We'll have to figure out a way to represent real numbers, but you could check if the `PowSOp` type is a `moore::IntType`, and only implement integers for now. (We don't really have a lowering for SystemVerilog's `real` type yet. That would be cool to add at some point -- but we'd probably have to map that to the `arith` MLIR dialect and MLIR's standard floating-point type.)

lib/Conversion/MooreToCore/MooreToCore.cpp
Max-astro: Thanks again for your review. 

I just wanted to let you know that I followed this suggestion. 
> @maerhart : Unfortunately, canonicalization patterns and conversion patterns are not compatible, so you also need to replicate the existing canonicalization patterns functionality inside the conversion pattern.

The functions `PowSOp::canonicalize` and `PowUOp::canonicalize` already exist, and the two sections of code mentioned above were duplicated from them.

lib/Conversion/MooreToCore/MooreToCore.cpp
Max-astro: This is a really good point, and I hadn't considered this question before. Is there a better way to handle the negative exponent situation? Should I remove the `comb::DivUOp` and skip the `scf.for` loop, then return const zero like the `PowSOp::canonicalize` function did?

Max-astro: Thank you so much for your guidance and patience!  Based on your suggestions, I've made some updates to the code and tests. 😃

fabianschuiki (approved):
Really cool! Thanks for implementing this :smiley:

PR #7900, opened 2024-11-26
[circt-bmc][VerifToSMT] Pop existing assertions on each cycle

TaoBi22:
This fixes a bug with circt-bmc where assertions on each cycle are never popped from the stack - as a result, if a property holds on the first cycle, it will always hold (as the unsatisfiable assertion never goes away). I haven't included an integration test in this PR because a meaningful integration test needs initial value support in circt-bmc - I have a branch to add this that I'll PR once this is merged.


maerhart (approved):
LGTM! Thanks for fixing this bug 💯

PR #7903, opened 2024-11-26
[circt-bmc][VerifToSMT] Add initial value support

TaoBi22:
Changes circt-bmc to use the initial values added to the BMC op in #7729 to initialize register states - this should mean circt-bmc can now check meaningful properties! :tada: 

The added integration test checks whether the counter value always stays below a certain value with different time bounds, which also functions as a check that assertions are getting popped as mentioned in #7900

### DIFF
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -9,6 +9,7 @@
 #include "circt/Conversion/VerifToSMT.h"
 #include "circt/Conversion/HWToSMT.h"
 #include "circt/Dialect/SMT/SMTOps.h"
+#include "circt/Dialect/SMT/SMTTypes.h"
 #include "circt/Dialect/Seq/SeqTypes.h"
 #include "circt/Dialect/Verif/VerifOps.h"
 #include "circt/Support/Namespace.h"
@@ -208,6 +209,7 @@ struct VerifBoundedModelCheckingOpConversion
 
     unsigned numRegs =
         cast<IntegerAttr>(op->getAttr("num_regs")).getValue().getZExtValue();
+    auto initialValues = cast<ArrayAttr>(op->getAttr("initial_values"));
 
     auto initFuncTy = rewriter.getFunctionType({}, initOutputTy);
     // Loop and init output types are necessarily the same, so just use init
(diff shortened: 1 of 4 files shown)


lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -208,6 +209,7 @@ struct VerifBoundedModelCheckingOpConversion
 
     unsigned numRegs =
         cast<IntegerAttr>(op->getAttr("num_regs")).getValue().getZExtValue();
+    auto initialValues = cast<ArrayAttr>(op->getAttr("initial_values"));
```
maerhart, line 212: There should be an ODS generated getter for the attribute. It's not great to hardcode the attribute name here.

lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -272,9 +274,19 @@ struct VerifBoundedModelCheckingOpConversion
...
-        inputDecls.push_back(rewriter.create<smt::DeclareFunOp>(loc, newTy));
+        continue;
       }
+      if (curIndex >= oldCircuitInputTy.size() - numRegs) {
+        auto initVal =
+            initialValues[curIndex - oldCircuitInputTy.size() + numRegs];
+        if (auto initIntAttr = dyn_cast<IntegerAttr>(initVal)) {
+          inputDecls.push_back(rewriter.create<smt::BVConstantOp>(
+              loc, initIntAttr.getValue().getSExtValue(),
+              dyn_cast<smt::BitVectorType>(newTy).getWidth()));
```
maerhart, line 285: If we are sure it can only be a BV we can to a `cast`, but don't we have to check if the type is a BV and error out otherwise? Are we doing this already somewhere?
It might also be possible to query the dialect's constant materializer.
```suggestion
              cast<smt::BitVectorType>(newTy).getWidth()));
```

maerhart (approved):
LGTM! Very nice that these examples work properly now. 🎉

I think we should change the assembly format of the bmc op a bit in the future. It's a bit confusing to see which of the block arguments and terminator operands are regular inputs/outputs, registers, state values, etc.

lib/Conversion/VerifToSMT/VerifToSMT.cpp
TaoBi22: Good point, I'll swap that and the num_regs one over!

lib/Conversion/VerifToSMT/VerifToSMT.cpp
TaoBi22: Good point! I'll try and swap to the constant materializer, that should be a lot nicer

PR #7670, opened 2024-10-06
SCF IndexSwitch to nested If-Else

jiahanxie353:
This transform is meant for dialects that inherently does not have a `switch` control statement (like Calyx)

PR #7905, opened 2024-11-26
SCF IndexSwitch to nested If-Else

jiahanxie353:
This transform is meant for dialects that inherently does not have a switch control statement (like Calyx).
And fixes valgrind issue, which was a problem here: https://github.com/llvm/circt/pull/7670


jiahanxie353:
I fixed the valgrind locally without major changes; so I'll just merge it after CI passes and hopefully it should pass the nightly integration test
cc @seldridge @cgyurgyik @rachitnigam

cgyurgyik:
Curiously, what was the issue?

jiahanxie353:
> Curiously, what was the issue?

I directly erased the block still owned by operations, causing invalid memory access. So I used `inlineRegion` to handle the ownership and block lists issue

PR #7907, opened 2024-11-26
[FIRRTL] Remove all traces of OMIR JSON support.

mikeurbach:
This legacy system for representing an object model has been replaced by FIRRTL classes, objects, and properties. This removes the old system completely.

The main change is removing the EmitOMIR pass completely, but there are several other things related to OMIR JSON to clean up. This includes command line options and annotations for specifying input and output OMIR JSON, as well as support for parsing the JSON and scattering OMIR tracker annotations. Any tests related to the above were also removed.


include/circt/Dialect/FIRRTL/AnnotationDetails.h
```diff
@@ -48,10 +48,6 @@ constexpr const char *traceNameAnnoClass =
 constexpr const char *traceAnnoClass =
     "chisel3.experimental.Trace$TraceAnnotation";
 
-constexpr const char *omirAnnoClass =
-    "freechips.rocketchip.objectmodel.OMIRAnnotation";
-constexpr const char *omirFileAnnoClass =
-    "freechips.rocketchip.objectmodel.OMIRFileAnnotation";
 constexpr const char *omirTrackerAnnoClass =
     "freechips.rocketchip.objectmodel.OMIRTracker";
```
seldridge, line 52: This can also go. If you don't handle it in this PR, I can deal with with it in a follow-on. 

This does mean that the `shouldBeDeleted` (or can be) function needs to be updated.

seldridge (approved):
Awesome. If there's any residue after this I am more than happy to deal with it.

include/circt/Dialect/FIRRTL/AnnotationDetails.h
mikeurbach: Yep, I'll pull this thread

mikeurbach: This is now going deeper than I wanted... I'm starting to look at the changes to IMDCE and I'm nervous I've missed something... So, I broke it out and I've sent a draft PR here, based on this branch for now: https://github.com/llvm/circt/pull/7908

PR #7908, opened 2024-11-26
[FIRRTL] Rip out OMIRTracker and logic that uses it.

mikeurbach:
This was used in the old legacy OMIR JSON system. Now that is gone, we don't need to support this annotation. Furthermore, the canBeDeleted logic on annotations will always return false, since we don't actually see this annotation any more. That means we can just get rid of that method, and fix up all the call sites to reflect the fact that it would have always returned false.


mikeurbach:
Thanks for pushing this through!

PR #7909, opened 2024-11-27
[FIRRTL] FoldRegMems: insert new ops into same block as memory

rwy7:
Before this PR, FoldRegMems would construct new ops in the "body of the parent FModuleOp". We need to place these ops in the same block as the memory.

This PR fixes a bug where, when a memory under a layerblock was canonicalized to a register, the register would be placed at the original location of the memory (under the layerblock), but its readers would be placed outside the layerblock, resulting in a dominance checking error.


seldridge (approved):
I see utility and simplicity in this approach of relocating the register to the end. This will likely produce worse Verilog in the output unless a later pass comes and reorders things to remove the temporary wire. That said, this optimization is so rare anyway and this patch fixes a known bug re: layers.

LGTM

PR #4652, opened 2023-02-13
[FIRRTL][MemOp] Removed unused bits from memories

nandor:
After memories are flattened, subsequent passes can render certain bits of their outputs unused.

This PR introduces a canonicalizer which identifies the bits of a memory output that are used (by looking through `BitsPrimOp` oeprations). Afterwards, the memory is narrowed and write inputs are built by concatenating the chunks of the selected inputs.

PR #7911, opened 2024-11-27
[FIRRTL] fix missing type conversion in FoldRegMems

unlsycn:
Fix https://github.com/chipsalliance/chisel/issues/4536

### DIFF
--- a/test/Dialect/FIRRTL/simplify-mems.mlir
+++ b/test/Dialect/FIRRTL/simplify-mems.mlir
@@ -497,3 +497,47 @@ firrtl.circuit "OneAddressNoMask" {
     firrtl.connect %write_mask, %c1_ui1 : !firrtl.uint<1>, !firrtl.uint<1>
   }
 }
+
+// -----
+
+firrtl.circuit "SIntOneAddress" {
+  firrtl.module @SIntOneAddress(
+      in %clock: !firrtl.clock,
+      in %reset: !firrtl.uint<1>,
+      in %io_dataIn: !firrtl.sint<32>,
+      out %io_dataOut: !firrtl.sint<32>) {
+
+    // CHECK: %mem = firrtl.reg %clock : !firrtl.clock, !firrtl.sint<32>
+    // CHECK: firrtl.matchingconnect %mem, {{%.+}} : !firrtl.sint<32>
+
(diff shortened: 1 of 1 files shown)


rwy7:
Hi, the underlying issue seems to be in FoldUnusedBits (introduced in https://github.com/llvm/circt/pull/4652), a different memory canonicalizer, which wasn't correctly handling signed data in a memory. The bitcast op added in this PR hides the problem. We found a different way to produce the error which does not involve the 1-element-memory -> register fold. I have a different fix over here https://github.com/llvm/circt/pull/7913, please take a look!

Thanks for all your help!

unlsycn:
> Hi, the underlying issue seems to be in FoldUnusedBits (introduced in #4652), a different memory canonicalizer, which wasn't correctly handling signed data in a memory. The bitcast op added in this PR hides the problem. We found a different way to produce the error which does not involve the 1-element-memory -> register fold. I have a different fix over here #7913, please take a look!
> 
> Thanks for all your help!

Hi, thanks for your explanation! #7913 won't fix https://github.com/chipsalliance/chisel/issues/4536 since the actual lack of type cast in the original issue is that `dataPart` and `nextPart` produced by `firrtl.bits` are unsigned but connect to the signed cast.  While #7913 is obviously a nice patch, it doesn't seems to be related to https://github.com/chipsalliance/chisel/issues/4536.

rwy7:
Ugh, thanks, looks like I managed to golf the example to a different issue, sorry. I think you're right. Do you think it makes more sense to cast when applying the mask, right here: https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/FIRRTLFolds.cpp#L3036

Either way, LGTM.

rwy7 (approved):

seldridge:
I rebased this and merged it here: https://github.com/llvm/circt/commit/996b1283d861d4f373d8b854c7f84ede8e227b1a Thanks for the PR.

unlsycn:
> I rebased this and merged it here: [996b128](https://github.com/llvm/circt/commit/996b1283d861d4f373d8b854c7f84ede8e227b1a) Thanks for the PR.

Thank you so much, I am sorry that I forgot about this PR.

PR #7912, opened 2024-11-27
[FIRRTL] Support layers in MergeConnections

seldridge:
Modify the `MergeConnections` pass to walk into regions and blocks. Because the algorithm used in this pass wants to insert operations _after_ the operation that is currently being processed _and then_ visit them, this cannot use a normal `Operation::walk` (since this will pre-increment and not visit these operations).  Keep this existing algorithm, but use an iterator worklist to make this behave like a walk.

This is done as part of moving `LowerLayers` to the end of the FIRRTL pipeline so that inline layer support (which requires producing `sv.ifdef` operations which we would like to hide from the rest of the FIRRTL pipeline) can be turned on.


uenoku (approved):
LGTM

PR #7913, opened 2024-11-27
[FIRRTL] FoldUnusedBits: Cast compressed data back to signed integer

rwy7:
The FoldUnusedBits folder will, when it sees a memory with obviously unused bits (all readers are bitselect ops), will compress that memory to a smaller integer. The data written to the memory has to be compressed to just the relevant bits, which is done using a sequence of bit-select and bit-concat ops on the data. The result of these ops is unsigned. If the memory itself stores a signed integer, then we need to cast this data back to a signed value.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -2849,6 +2849,15 @@ struct FoldUnusedBits : public mlir::RewritePattern {
           catOfSlices = slice;
         }
       }
+
+      // If the original memory held a signed integer, then the compressed
+      // memory will be signed too. Since the catOfSlices is always unsigned,
+      // cast the data to a signed integer if needed before connecting back to
+      // the memory.
+      if (type.isSigned())
+        catOfSlices =
+            rewriter.createOrFold<AsSIntPrimOp>(writeOp.getLoc(), catOfSlices);
+
       rewriter.replaceOpWithNewOp<MatchingConnectOp>(writeOp, writeOp.getDest(),
                                                      catOfSlices);
     }
--- a/test/Dialect/FIRRTL/simplify-mems.mlir
+++ b/test/Dialect/FIRRTL/simplify-mems.mlir
@@ -336,6 +336,36 @@ firrtl.circuit "UnusedBitsAtEnd" {
 
 // -----
 
+firrtl.circuit "UnusedBitsOfSignedInteger" {
+  firrtl.module public @UnusedBitsOfSignedInteger(
+      in %in_data: !firrtl.sint<42>,
+      out %result_read: !firrtl.sint<5>) {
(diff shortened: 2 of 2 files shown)


youngar (approved):

test/Dialect/FIRRTL/simplify-mems.mlir
```diff
@@ -336,6 +336,54 @@ firrtl.circuit "UnusedBitsAtEnd" {
...
+    firrtl.connect %write_addr, %addr : !firrtl.uint<4>, !firrtl.uint<4>
+    %write_en = firrtl.subfield %Memory_write[en] : !firrtl.bundle<addr: uint<4>, en: uint<1>, clk: clock, data: sint<42>, mask: uint<1>>
+    firrtl.connect %write_en, %c1_ui1 : !firrtl.uint<1>, !firrtl.uint<1>
+    %write_clk = firrtl.subfield %Memory_write[clk] : !firrtl.bundle<addr: uint<4>, en: uint<1>, clk: clock, data: sint<42>, mask: uint<1>>
+    firrtl.connect %write_clk, %clock : !firrtl.clock, !firrtl.clock
+    %write_data = firrtl.subfield %Memory_write[data] : !firrtl.bundle<addr: uint<4>, en: uint<1>, clk: clock, data: sint<42>, mask: uint<1>>
+    firrtl.connect %write_data, %in_data : !firrtl.sint<42>, !firrtl.sint<42>
+    %write_mask = firrtl.subfield %Memory_write[mask] : !firrtl.bundle<addr: uint<4>, en: uint<1>, clk: clock, data: sint<42>, mask: uint<1>>
+    firrtl.connect %write_mask, %c1_ui1 : !firrtl.uint<1>, !firrtl.uint<1>
+  }
```
seldridge, line 382: Usual nit about playing test golf to get this as minimal (maintainable) as possible.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -2849,6 +2849,15 @@ struct FoldUnusedBits : public mlir::RewritePattern {
...
         }
       }
+
+      // If the original memory held a signed integer, then the compressed
+      // memory will be signed too. Since the catOfSlices is always unsigned,
+      // cast the data to a signed integer if needed before connecting back to
+      // the memory.
+      if (type.isSigned())
+        catOfSlices =
+            rewriter.createOrFold<AsSIntPrimOp>(writeOp.getLoc(), catOfSlices);
```
seldridge, line 2859: Nice clean fix. 👍

seldridge (approved):

unlsycn:
This patch doesn't seem to fix https://github.com/chipsalliance/chisel/issues/4536, see [this comment](https://github.com/llvm/circt/pull/7911#issuecomment-2505236388)

PR #7915, opened 2024-11-28
[HW] ExportHier: do not include bound in modules

youngar:


seldridge (approved):
SGTM

This is a file format we define and if we want to change it in this way, that's fine.

Does the test fail without the change? Some of those `CHECK-SAME` checks may need one or more `CHECK-NOT`s interleaved.

youngar:
Woops yeah, I think I was reading those as check-next and not check-same.

PR #7916, opened 2024-11-28
[RTG] Add set union operation

maerhart:


include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -141,6 +142,23 @@ def SetDifferenceOp : RTGOp<"set_difference", [
   }];
 }
 
+def SetUnionOp : RTGOp<"set_union", [
+  Pure, SameOperandsAndResultType, Commutative
+]> {
+  let summary = "computes the union of sets";
+  let description = [{
+    Computes the union of the given sets. If the list of sets must contain at
+    least one element.
```
fabianschuiki, line 151: Does the second sentence want to say "The list of sets must contain at least one element."? If yes, it might be worth checking for that in the verifier, or if empty sets are valid, maybe just return an empty set in that case?

fabianschuiki (approved):
LGTM

include/circt/Dialect/RTG/IR/RTGOps.td
maerhart: Yes that's what it should say. It originally said that if there are zero operands, it returns the empty set but then changed it because the `SameOperandsAndResultType` trait verifies that there is at least one operand (there is also a test for that already added). We could go back to allowing zero operands, but I'm not sure it it's beneficial (it would just fold to a zero-operand `set_create`).

fabianschuiki: Oh nice! I was wondering where the error message came from :grinning: No need to change anything, disallowing empty operand lists feels very reasonable. Especially since you'd have two different ways of creating a  constant empty set afterwards, which would be pretty weird.

PR #7917, opened 2024-11-28
[RTG] Add bag union operation

maerhart:


include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -224,6 +224,23 @@ def BagDifferenceOp : RTGOp<"bag_difference", [
   }];
 }
 
+def BagUnionOp : RTGOp<"bag_union", [
+  Pure, SameOperandsAndResultType, Commutative
+]> {
+  let summary = "computes the union of bags";
+  let description = [{
+    Computes the union of the given bags. If the list of sets must contain at
+    least one element.
```
fabianschuiki, line 233:
```suggestion
    Computes the union of the given bags. The list of sets must contain at
    least one element.
```

fabianschuiki (approved):
LGTM! I think that spurious CI failure in HWToBTOR has been fixed on main. Might just work if you kick the run off again.

PR #7920, opened 2024-11-29
[RTG] Add set_size op for sets and bag_unique_size op for bags

maerhart:


include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -241,6 +252,21 @@ def BagUnionOp : RTGOp<"bag_union", [
   }];
 }
 
+def BagGetSizeOp : RTGOp<"bag_get_size", [Pure]> {
+  let summary = "returns the number of unique elements in the bag";
+  let description = [{
+    This operation returns the number of unique elements in the bag, i.e., for
+    the bag `{a, a, b, c, c}` it returns 3.
+  }];
```
fabianschuiki, line 260: Random thought: Would there ever be a need for an operation that returns 5 for `{a, a, b, c, c}`? Just wondering if there's a need to distinguish "size" and maybe something like "count unique" at some point.

fabianschuiki (approved):
LGTM!

include/circt/Dialect/RTG/IR/RTGOps.td
maerhart: I think it's more likely that we want something that can give us the number of duplicates of a given object in a bag. But I can rename this op to make it more obvious what it does and keep the option open to have a "size" op that does what you describe.

fabianschuiki: No worries, no need to change any op names 😃. There's a summary/description that says exactly what the op does, so that's perfect!

PR #7923, opened 2024-11-29
[VerifToSMT] Exit early after too many clocks error

TaoBi22:
Tiny change, just makes it so VerifToSMT exits early after throwing the too many clocks error (rather than still trying to apply the pass).

### DIFF
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -445,6 +445,7 @@ void ConvertVerifToSMTPass::runOnOperation() {
             op->emitError(
                 "only modules with one or zero clocks are currently supported");
             signalPassFailure();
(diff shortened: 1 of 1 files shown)


lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -445,6 +445,7 @@ void ConvertVerifToSMTPass::runOnOperation() {
             op->emitError(
                 "only modules with one or zero clocks are currently supported");
             signalPassFailure();
```
maerhart, line 447:
```suggestion
```

maerhart (approved):
Nice!

lib/Conversion/VerifToSMT/VerifToSMT.cpp
TaoBi22: Good point, thanks!

PR #7924, opened 2024-11-29
[RTGTest] Add some registers

maerhart:
This adds some registers to later test passes such as register alloction, assembly emission, correct behavior of the elaboration pass in presence of registers, etc.


include/circt/Dialect/RTGTest/IR/RTGTestAttributes.td
```diff
@@ -0,0 +1,60 @@
...
+    I32EnumAttrCase<"s6",  22>,
+    I32EnumAttrCase<"s7",  23>,
+    I32EnumAttrCase<"s8",  24>,
+    I32EnumAttrCase<"s9",  25>,
+    I32EnumAttrCase<"s10",  26>,
+    I32EnumAttrCase<"s11",  27>,
+    I32EnumAttrCase<"t3",  28>,
+    I32EnumAttrCase<"t4",  29>,
+    I32EnumAttrCase<"t5",  30>,
+    I32EnumAttrCase<"t6",  31>,
```
fabianschuiki, line 54: This looks suspiciously like RISC-V :grinning:. Is the plan for the RTG dialect to at some point also "standardize" some of these things in CIRCT? Or are users of the RTG dialect expected to always provide their own ISA-describing ops?

include/circt/Dialect/RTGTest/IR/RTGTestOps.td
```diff
@@ -33,3 +35,21 @@ def CPUDeclOp : RTGTestOp<"cpu_decl", [
...
+def RegisterOp : RTGTestOp<"reg", [
+  DeclareOpInterfaceMethods<RegisterOpInterface>,
+  DeclareOpInterfaceMethods<InferTypeOpInterface>,
+]> {
+  let summary = "returns a value representing a register";
+  let description = [{
+    This operation creates a value representing the register given as the 'reg'
+    attribute. A register can be a concrete register or a virtual register.
+    Virtual registers will be assigned a concrete register when running register
+    allocation.
```
fabianschuiki, line 48: This probably doesn't make sense for a test dialect, but in practice, would you have virtual registers as a separate op that cannot be CSEd, and actual ISA register as something "pure" or maybe even "constant-like"?

fabianschuiki (approved):
LGTM!

include/circt/Dialect/RTGTest/IR/RTGTestAttributes.td
maerhart: Well observed. The idea is that we add a description of RV32I or a subset of it to test the RTG infra. We chose an actual ISA because it allows us to have E2E integration tests upstream so that breaking changes to core RTG stuff are detected as early as possible. It *could* also serve as a small example of how to use the RTG infra, but it is not meant to "standardize" anything. Users are expected to come up with their own dialect describing what they want to generate random tests for.

include/circt/Dialect/RTGTest/IR/RTGTestOps.td
maerhart: Good point! Having concrete registers CSE would be very nice. Having separate ops for concrete and virtual registers complicates register allocation a bit but makes elaboration easier. Let me think about that a bit more.

include/circt/Dialect/RTGTest/IR/RTGTestAttributes.td
```diff
@@ -0,0 +1,60 @@
...
+
+#ifndef CIRCT_DIALECT_RTGTEST_IR_RTGTESTATTRIBUTES_TD
+#define CIRCT_DIALECT_RTGTEST_IR_RTGTESTATTRIBUTES_TD
+
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/EnumAttr.td"
+
+// Flat allocation of unique IDs to all registers. The actual ID value does not
+// matter.
+def RegisterAttr : I32EnumAttr<
```
maerhart, line 21: I think ideally this attribute should be of IntegerRegisterType, not I32. It's not a big deal for now though.

include/circt/Dialect/RTGTest/IR/RTGTestAttributes.td
fabianschuiki: That makes sense!

maerhart:
I'll merge this as-is for now and will follow up with another PR to fix the above two points.

PR #7926, opened 2024-11-29
[Calyx] Avoid using designated initializers

TaoBi22:
#7860 broke Windows CI as it uses designated initializers which are c++ 20 - this just reformats the bits where they were used.


cgyurgyik (approved):
That was my fault; I suggested it. Thanks for fixing it!

jiahanxie353:
thanks for fixing it!

uenoku:
Awesome, thanks!

PR #7927, opened 2024-11-29
[SMT] Add set_logic operation

TaoBi22:
Adds an operation to set the logic to be used by the SMT solver. I'll put the lowering to Z3 in a follow-up PR after this one since it requires some extra checks due to limitations on what Z3 allows so it has a larger diff.


maerhart (approved):
LGTM!

PR #7928, opened 2024-12-01
Add case inside

ankolesn:
Implementation of functionality for checking whether a value belongs to a set of expressions (like inside in SystemVerilog)

### DIFF
--- a/lib/Conversion/ImportVerilog/Statements.cpp
+++ b/lib/Conversion/ImportVerilog/Statements.cpp
@@ -208,8 +208,32 @@ struct StmtVisitor {
           cond = builder.create<moore::CaseZEqOp>(itemLoc, caseExpr, value);
           break;
         case CaseStatementCondition::Inside:
-          mlir::emitError(loc, "unsupported set membership case statement");
-          return failure();
+          std::vector<Value> values;
+          values.reserve(item.expressions.size());
+          for (const auto *expr : item.expressions) {
+              auto value = context.convertRvalueExpression(*expr);
+              if (!value)
+                  return failure();
+              values.push_back(value);
+          }
+
+          if (values.empty()) {
+              mlir::emitError(loc, "empty set in inside case statement");
+              return failure();
+          }
+
+          if (values.size() == 1) {
+              cond = builder.create<moore::CaseEqOp>(
+                  loc, caseExpr, values.front());
+          } else {
+              cond = builder.create<moore::CaseEqOp>(loc, caseExpr, values[0]);
+              for (size_t i = 1; i < values.size(); ++i) {
+                  auto nextCond = builder.create<moore::CaseEqOp>(
+                      loc, caseExpr, values[i]);
+                  cond = builder.create<mlir::arith::OrIOp>(loc, builder.getI1Type(), cond,
+                                                    nextCond);
+              }
+          }
         }
         cond = builder.create<moore::ConversionOp>(itemLoc, builder.getI1Type(),
                                                    cond);
--- a/test/Conversion/ImportVerilog/basic.sv
+++ b/test/Conversion/ImportVerilog/basic.sv
@@ -2232,3 +2232,21 @@ function int AssignFuncArgs2(int x, int y);
   // CHECK: [[ADD:%.+]] = moore.add [[READ_X]], [[READ_Y]] : i32
   return x+y;
 endfunction
+
+// CHECK-LABEL: moore.module @CaseStatementTest()
+module CaseStatementTest;
+  bit [31:0] caseExpr, result;
+  // CHECK: [[CASE_EXPR:%.+]] = moore.read %caseExpr : <i32>
+  // CHECK: [[COND1:%.+]] = moore.caseeq [[CASE_EXPR]], {{.*}} : !moore.i32
+  // CHECK: [[COND2:%.+]] = moore.caseeq [[CASE_EXPR]], {{.*}} : !moore.i32
+  // CHECK: [[COND3:%.+]] = moore.caseeq [[CASE_EXPR]], {{.*}} : !moore.i32
+  // CHECK: [[OR1:%.+]] = arith.or [[COND1]], [[COND2]] : i1
+  // CHECK: [[FINAL_COND:%.+]] = arith.or [[OR1]], [[COND3]] : i1
+  // CHECK: moore.assign %result, [[FINAL_COND]] : i1
+  always_comb begin
+    case (caseExpr)
+      32'h1, 32'h2, 32'h3: result = 1;
+      default: result = 0;
+    endcase
+  end
+endmodule


lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -208,8 +208,32 @@ struct StmtVisitor {
...
+          }
+
+          if (values.size() == 1) {
+              cond = builder.create<moore::CaseEqOp>(
+                  loc, caseExpr, values.front());
+          } else {
+              cond = builder.create<moore::CaseEqOp>(loc, caseExpr, values[0]);
+              for (size_t i = 1; i < values.size(); ++i) {
+                  auto nextCond = builder.create<moore::CaseEqOp>(
+                      loc, caseExpr, values[i]);
```
hailongSun2000, line 232: Based on the SystemVerilog IEEE Std 1800-2017 § 12.5.4, we should use wildcard equality ops(`moore.wildcard_eq`/`moore.wildcard_ne`), rather than `moore.case_eq`.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -2232,3 +2232,21 @@ function int AssignFuncArgs2(int x, int y);
...
+  // CHECK: [[OR1:%.+]] = arith.or [[COND1]], [[COND2]] : i1
+  // CHECK: [[FINAL_COND:%.+]] = arith.or [[OR1]], [[COND3]] : i1
+  // CHECK: moore.assign %result, [[FINAL_COND]] : i1
+  always_comb begin
+    case (caseExpr)
+      32'h1, 32'h2, 32'h3: result = 1;
+      default: result = 0;
+    endcase
+  end
+endmodule
```
hailongSun2000, line 2252: nit: Need a new empty line here.

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -208,8 +208,32 @@ struct StmtVisitor {
...
+                  cond = builder.create<mlir::arith::OrIOp>(loc, builder.getI1Type(), cond,
+                                                    nextCond);
```
hailongSun2000, line 234: Moore dialect has its own OrOp(`moore.or`). We should as far as avoid introducing other dialects.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -2232,3 +2232,21 @@ function int AssignFuncArgs2(int x, int y);
...
+  // CHECK: [[COND3:%.+]] = moore.caseeq [[CASE_EXPR]], {{.*}} : !moore.i32
+  // CHECK: [[OR1:%.+]] = arith.or [[COND1]], [[COND2]] : i1
+  // CHECK: [[FINAL_COND:%.+]] = arith.or [[OR1]], [[COND3]] : i1
+  // CHECK: moore.assign %result, [[FINAL_COND]] : i1
+  always_comb begin
+    case (caseExpr)
+      32'h1, 32'h2, 32'h3: result = 1;
+      default: result = 0;
+    endcase
+  end
```
hailongSun2000, line 2251: Your example is not a **case inside statement.** Maybe the example in the SystemVerilog IEEE Std § 12.5.4 is better. What do @ankolesn think :smiley:?
```
logic [2:0] status;
always @(posedge clock)
priority case (status) inside
1, 3 : task1; // matches 'b001 and 'b011
3'b0?0, [4:7]: task2;  // matches 'b000 'b010 'b0x0 'b0z0 'b100 'b101 'b110 'b111
endcase  // priority case fails all other values including 'b00x 'b01x 'bxxx
```

ankolesn: Thanks for your comments! I tried to fix it, you can check it :smiley:

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -2232,3 +2232,72 @@ function int AssignFuncArgs2(int x, int y);
...
+        // CHECK: Entering priority case
+        priority case (status)
+            // CHECK: Matches 3'b001 or 3'b011, executes task1
+            3'b001, 3'b011: task1;
+            // CHECK: Matches 3'b0?0 or [3'b100:3'b111], executes task2
+            3'b0?0, [3'b100:3'b111]: task2;
+            // CHECK: No match, default case
+            default: $display("No match for status = %b at time %0t", status, $time);
+        endcase
+    end
```
fabianschuiki, line 2302: Hey @ankolesn, thanks a lot for working on this! One comment regarding your unit test: we don't simulate the design as part of the unit test, so things like `$display(...)` will not show up as prints in the output. Instead, the unit tests check that the expected IR operations get generated. For your PR, I'd add different examples that exercise interesting corner cases of your implementation. If you look through `basic.sv`, you should find existing tests for `case` statements that can give you some inspiration on how to do that. Then you can check if your code generates the correct IR ops for `inside` case statements.

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -207,9 +207,32 @@ struct StmtVisitor {
         case CaseStatementCondition::WildcardJustZ:
           cond = builder.create<moore::CaseZEqOp>(itemLoc, caseExpr, value);
           break;
-        case CaseStatementCondition::Inside:
-          mlir::emitError(loc, "unsupported set membership case statement");
-          return failure();
+          std::vector<Value> values;
```
darthscsi, line 210: SmallVector per MLIR coding guidelines.

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -207,9 +207,32 @@ struct StmtVisitor {
...
+          }
+
+          if (values.empty()) {
+              mlir::emitError(loc, "empty set in inside case statement");
+              return failure();
+          }
+
+          if (values.size() == 1) {
+            cond = builder.create<moore::WildcardEqOp>(loc, caseExpr,
+                                                       values.front());
```
darthscsi, line 226: nit: why use front here and [0] two lines later?

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -207,9 +207,32 @@ struct StmtVisitor {
...
+                  return failure();
+              values.push_back(value);
+          }
+
+          if (values.empty()) {
+              mlir::emitError(loc, "empty set in inside case statement");
+              return failure();
+          }
+
+          if (values.size() == 1) {
```
darthscsi, line 224: This conditional could be completely eliminated.

lib/Conversion/ImportVerilog/Statements.cpp
ankolesn: I wanted to show that values.front() - means exactly the last element 😃 
I fixed it.

test/Conversion/ImportVerilog/basic.sv
ankolesn: This is the first time I've encountered this type of test, so I'm sorry that I have to upload new versions many times. 
I tried to rewrite it. 😄

PR #7931, opened 2024-12-02
[firtool] Add an option to disable CSE in classes

prithayan:
This PR adds the option to run CSE only on `firrtl.module`, and ignore everything else.
There are some tools, that are not able to handle CSE of `firrtl.object` in `firrtl.class`. 
Specifically, the reuse of `firrtl.object` is not modeled properly.
This is a temporary workaround to disable the CSE of operations inside `firrtl.class`, which can be removed once the actual issue in the tool is fixed.


seldridge (approved):
SGTM

If this helps temporarily make it easier for downstream to get onboard, that is fine.

seldridge (commented):
One comment: can this also keep `include/circt-c/Firtoo/Firtool.h` up-to-date?

PR #7933, opened 2024-12-02
[InstanceChoice] Add the instance choice specialization pass to firtool

prithayan:
This PR adds the instance choice specialization pass to the `firtool` pipeline.
Add optional arguments to `firtool` to select the specialization option for the instance choice.
Then update the pass to use the default module, when no options are specified.
Update lit tests to check for valid instance choice lowering.


lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
```diff
@@ -68,33 +64,46 @@ struct SpecializeOptionPass
...
 
+    bool analysisPreserved = numInstances == 0;
+    circuit->walk([&](OptionOp optionOp) {
+      optionOp->erase();
+      analysisPreserved = false;
+    });
+    if (analysisPreserved)
+      markAllAnalysesPreserved();
+
     if (failed)
```
uenoku, line 97: failed is dead now?

include/circt/Firtool/Firtool.h
seldridge, line 1: Please have the PR reflect any changes to this file in the corresponding area of the C-API.

include/circt/Dialect/FIRRTL/Passes.h
```diff
@@ -202,7 +202,8 @@ std::unique_ptr<mlir::Pass> createProbesToSignalsPass();
 
 std::unique_ptr<mlir::Pass> createSpecializeLayersPass();
 
-std::unique_ptr<mlir::Pass> createSpecializeOptionPass();
+std::unique_ptr<mlir::Pass> createSpecializeOptionPass(
```
uenoku, line 205: I think we can remove `std::optional`. You can use empty array as `nullopt`.

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
```diff
@@ -68,33 +64,46 @@ struct SpecializeOptionPass
         &getContext(), circuit.getOps<FModuleOp>(), [&](auto module) {
           module.walk([&](firrtl::InstanceChoiceOp inst) {
             auto it = selected.find(inst.getOptionNameAttr());
+            FlatSymbolRefAttr target;
             if (it == selected.end()) {
-              inst.emitError("missing specialization for option ")
+              target = inst.getDefaultTargetAttr();
+              inst->emitWarning("missing specialization for option ")
                   << inst.getOptionNameAttr();
```
uenoku, line 71: Do we want to keep this warning if we decided to use default lowering?

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
```diff
@@ -68,33 +64,46 @@ struct SpecializeOptionPass
...
-              failed = true;
-              return;
-            }
```
seldridge, line 76: I realize that this all-or-nothing behavior is originally how this worked.

However, it seems odd to force a user to specialize everything or nothing as opposed to just specializing what the user requested.

uenoku (approved):
LGTM, please update capi as well.

test/firtool/instchoice.fir
```diff
@@ -26,6 +28,8 @@ circuit Foo:
     ; CHECK-SAME:   @FPGA -> @FPGATarget,
     ; CHECK-SAME:   @ASIC -> @ASICTarget
     ; CHECK-SAME: } (in clock: !firrtl.clock)
+    ; ASIC: hw.instance "inst" @ASICTarget(clock: %clock: !seq.clock) -> ()
+    ; DEFAULT: hw.instance "inst" @DefaultTarget(clock: %clock: !seq.clock) -> ()
```
seldridge, line 32: 🤔 Making the default behavior be "replace instance choices with the default" also seems slightly odd. It seems like the default should be "do nothing". I realize that instance choice is unusable right now unless specialization of every option is run, though.

seldridge (commented):
The roughly sister pass of this is `SpecializeLayers` and that went a slightly different direction than adding an option. Instead that puts the information directly onto the `CircuitOp` via a mandatory attribute.

While I can't recall the exact reasons for this design point, consistency between the two approaches may be good.

@youngar or @rwy7: Do you have any thoughts about the above?

test/firtool/instchoice.fir
prithayan: Maybe it depends on how we define it, the default target, for an instance choice can be defined as the target when no option is specified. Also, if nothing is done here, when should the instance choice be handled in the pipeline ?

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
prithayan: In the current implementation, the user need not specialize everything, it would be lowered to the default target, if nothing is specified for an instance choice. The user can specify only the choices that are relevant, and the rest can be lowered to the default target.

test/firtool/instchoice.fir
seldridge: If no default is specified, then the instance choice gets lowered to how the FIRRTL ABI defines this lowering. (This is not specified right now. However, I have one such ABI PR'd here: https://github.com/chipsalliance/firrtl-spec/pull/155.)

The underlying problem is that instance choice was only handled for this narrow, internal use case of CIRCT choosing a value for all instance choices as opposed to the more general user request for Verilog elaboration-time selection. 🥲 

The motivation here is more about aligning this with how layers are enabled and disabled, e.g., `--enable-layer=Verification` specializes away the `Verification` layer, but it leaves any other layers in place. It would seem better to have instance choice specialization work the same way, e.g., `--specialize-instance-choice=Performance=ASIC` would not affect any other instance choice in the circuit.

We do want to move in this direction. However, we don't have to move there right now with this PR.

WDYT?

davidbiancolin: I think this makes more sense, but i'm assuming you meant to write: 

`--specialize-instance-choice=Platform=ASIC`

To specify which target is being specialized? 

Also, not to derail this (and i see there was some discussion of this in the spec PR) but it is useful to be able to specialize different instances of the same `instance_choice` to write miter-style circuits for testing the choices themselves (esp. when they are used in part of a larger circuit under test). I guess here it might just be fine to do the specialization at VCS-compile time though. For this reason i also like that default be explicitly named (as also discussed in the spec PR).

davidbiancolin: So in the short run, we'd need a flag to achieve the effect of specializing all modules to their default and just provide that when we're passing in circuits that contain instance choices for now?

seldridge: > I think this makes more sense, but i'm assuming you meant to write: 
> 
> `--specialize-instance-choice=Platform=ASIC`
> 
> To specify which target is being specialized? 

Yes!

> it is useful to be able to specialize different instances of the same instance_choice to write miter-style circuits for testing the choices themselves (esp. when they are used in part of a larger circuit under test)

Per-instance specialization currently works. It does create a blow-up in the number of options. This is an intentional choice that complicates the design. E.g., consider instance choice used in a multicore design to replace a core with a model. The designer needs to decide what swapping flexibility they want to have. If they want to have complete flexibility to swap every core arbitrarily in an n-core design, that would have to be n-different options. They likely don't want that and will choose to have one instance choice which swaps all of the cores except one.

I'd prefer to not support arbitrary per-instance after-the-fact specialization as that amounts to mandated module duplication which then has interactions with must-dedup and designer expectations around deduplication. I'm more comfortable with this if CIRCT-mandated deduplication (must-dedup) is removed.

> So in the short run, we'd need a flag to achieve the effect of specializing all modules to their default and just provide that when we're passing in circuits that contain instance choices for now?

I'm good with either:

1. All instance choice specializations are specified up front. 
2. There is a catch-all `-specialize-to-default-if-no-override` which will, after applying selected choices, will then convert everything to the default choice.

davidbiancolin: After more thought, i think strictness of (1) is probably desirable to avoid accidental footguns? We can always add the flag later. I'll defer to you and @prithayan on this one.

prithayan: I've decided to go with the option 1, unless all the instance choice specializations are specified, the pass will error out. 
Once the lowering of instance choice through the pipeline to verilog is supported, we can get rid of the error.

include/circt/Firtool/Firtool.h
prithayan: I'm trying to figure out the corresponding type for `ArrayRef` in C-API, any idea ?

seldridge: I don't see any uses of `ArrayRef` for the C-API. This may need to get funneled through an `ArrayAttr`.

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
```diff
@@ -89,12 +85,24 @@ struct SpecializeOptionPass
           });
         });
 
+    bool analysisPreserved = numInstances == 0;
+    circuit->walk([&](OptionOp optionOp) {
+      optionOp->erase();
```
prithayan, line 90: @seldridge , @uenoku , just wanted to make sure, this makes sense right ?
 The `OptionOp`  wouldn't be useful for any other passes downstream, even if we ignore an `InstanceChoiceOp`, when it is not specialized, and left in IR after this pass.

include/circt/Dialect/FIRRTL/Passes.h
prithayan: Fixed.

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
seldridge: This would only be safe for options that were specialized. If any `InstanceChoiceOp` is left in, e.g., if the circuit is partially instance-choice specialized, then this would create an invalid symbol reference.

prithayan: Yes, that makes sense. But since, there are no other passes that can handle the `OptionOp` maybe its okay to delete them. And this pass also errors out on partially specified instance choice. WDYT ?

prithayan:
PR is ready for another review, @uenoku , @seldridge please take a look if you get a chance.

- I updated the select option to follow the specialize layers style.
- Removed the lowering to default and instead error out on partially specified choice.

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
seldridge: All the other passes are _supposed_ to handle these things, just it's not implemented, yet. 🫠 The concern is more about if this pass is guaranteed to specialize away everything (which we eventually don't want), then it's fine. However, if it is going to leave anything in, then it shouldn't delete everything.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5825,10 +5825,24 @@ ParseResult FIRCircuitParser::parseCircuit(
...
 
+  auto arrayToAttr = [&](const ArrayRef<std::string> strArray) {
+    SmallVector<StringRef> attr;
+    for (const auto &str : strArray)
+      attr.push_back(str);
+    OpBuilder builder(getContext());
+    if (attr.empty())
+      return ArrayAttr();
+    return builder.getStrArrayAttr(attr);
+  };
```
seldridge, line 5836: This is super close to `parseLayers`. Can these be combined?

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
```diff
@@ -27,40 +27,38 @@ namespace {
...
-    for (const auto &optionAndCase : select) {
-      size_t eq = optionAndCase.find("=");
-      if (eq == std::string::npos) {
-        mlir::emitError(circuit.getLoc(),
-                        "invalid option format: \"" + optionAndCase + '"');
-        return signalPassFailure();
-      }
+    if (auto choiceAttr = circuit.getSelectInstChoiceAttr()) {
+      for (auto attr : choiceAttr.getAsRange<StringAttr>()) {
+        const auto &optionAndCase = attr.getValue().str();
```
seldridge, line 37: Is a `const std::string &` safe to take for a `.str()` that isn't given a temporary?

lib/Firtool/Firtool.cpp
```diff
@@ -55,6 +55,7 @@ LogicalResult firtool::populatePreprocessTransforms(mlir::PassManager &pm,
 LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
                                                   const FirtoolOptions &opt,
                                                   StringRef inputFilename) {
+
```
seldridge, line 58: Mega-nit: unrelated whitespace change.

lib/Firtool/Firtool.cpp
```diff
@@ -133,6 +134,7 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
 
   // Must run this pass after all diagnostic passes have run, otherwise it can
   // hide errors.
```
seldridge, line 136: Mega-nit: can you update the comment to indicate that this is plural?

test/Dialect/FIRRTL/specialize-option-errors.mlir
seldridge, line 1: Can this PR include a test of the error if the specialization is not fully specified?

seldridge (approved):
Nice. Some comments/nits throughout. 

Can the PR add a test of the error behavior if the specialization is not fully specified?

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
prithayan: Right, for now, this pass currently doesn't leave any `InstanceChoiceOp`, it should error out if a choice is not selected. So, it should be fine to delete it.
But the first pass that needs to handle `OptionOp` is `LowerToHW`, there is no `HW` counterpart for it.

test/firtool/instchoice.fir
davidbiancolin: So if you went with (1) how do you specialize to `Default`?

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
seldridge: Yep... just showing something end-to-end would be cool. FWIW, I am totally fine with `LowerToHW` erroring with "unsupported operation". That's all part of getting a feature turned on incrementally. 😉

test/Dialect/FIRRTL/specialize-option-errors.mlir
prithayan: Added the test.

lib/Firtool/Firtool.cpp
prithayan: Updated the comment.

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
prithayan: Thanks for spotting this. Fixed.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
prithayan: Merged them.

prithayan:
Thanks for the review and discussion everyone. I'm planning to split this into a separate PR for the parser changes and will merge them as two commits.

seldridge:
> Thanks for the review and discussion everyone. I'm planning to split this into a separate PR for the parser changes and will merge them as two commits.

👍 Always feel empowered to slice things up into smaller commit chunks. 💯

PR #7936, opened 2024-12-02
[DC] Add unused fork result elimination canonicalizer

teqdruid:
Adding a canonicalizer to remove unused results from forks. Required to get valid post-canonicalization IR. Also, narrow the scope of "XFAIL"s in canonicalizer test to add new one.

### DIFF
--- a/lib/Dialect/DC/DCOps.cpp
+++ b/lib/Dialect/DC/DCOps.cpp
@@ -138,7 +138,7 @@ public:
 
         // We have a fork feeding into another fork. Replace the output fork by
         // adding more outputs to the current fork.
-        size_t totalForks = fork.getNumResults() + userFork.getNumResults() - 1;
+        size_t totalForks = fork.getNumResults() + userFork.getNumResults();
 
         auto newFork = rewriter.create<dc::ForkOp>(fork.getLoc(),
                                                    fork.getToken(), totalForks);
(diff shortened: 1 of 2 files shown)


lib/Dialect/DC/DCOps.cpp
```diff
@@ -180,10 +180,38 @@ class EliminateForkOfSourcePattern : public OpRewritePattern<ForkOp> {
...
 
+struct EliminateUnusedForkResultsPattern : mlir::OpRewritePattern<ForkOp> {
+  using mlir::OpRewritePattern<ForkOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(ForkOp op,
+                                PatternRewriter &rewriter) const override {
+    std::set<unsigned> unusedIndexes;
+
+    for (auto res : llvm::enumerate(op.getResults()))
+      if (res.value().getUses().empty())
```
mortbopet, line 191: `res.value().use_empty()`

mortbopet (approved):
Pattern LGTM. Strange about the now failing tests 🤔 but, sounds like you'll investigate, so I'll approve :).

teqdruid:
> Strange about the now failing tests 🤔

It's probably been failing for some time. The test for it was just disabled.

I'm gonna merge this without the fix and investigate shortly.

Issue #7932, opened 2024-12-02
[RTG] broken documentation


youngar:
The Operations and Types section of the documentation are not rendering correctly: https://circt.llvm.org/docs/Dialects/RTG/#operations.

PR #7940, opened 2024-12-03
[RTG] Generate separate doc files for ops and types

maerhart:
This makes it properly includable in the rationale file. Overlooked this in the fix for https://github.com/llvm/circt/issues/7932. Sorry.


fabianschuiki (approved):
LGTM

PR #7944, opened 2024-12-04
[FIRRTL] Remove NestedPrefixModulesAnnotation

seldridge:
Now that Chisel supports native prefixing [[1]], support for the SiFive-internal `NestedPrefixModulesAnnotation` and it's accompanying pass, `PrefixModules`, can be dropped.  This removes the annotation, the pass, and updates or deletes tests now that this support is no longer there.

[1]: https://github.com/chipsalliance/chisel/commit/0e191719107b2d9ee9a98170a6c3419784152307


seldridge:
CC: @jackkoenig

PR #7945, opened 2024-12-04
[DC] Initial values were being ignored

teqdruid:
Pass them through to DC, then error out in the DCToHW lowering since
nothing simple supports loading content on reset.

### DIFF
--- a/integration_test/Dialect/Handshake/buffer_arg_init/buffer_arg_init.mlir
+++ b/integration_test/Dialect/Handshake/buffer_arg_init/buffer_arg_init.mlir
@@ -1,0 +1,28 @@
+// REQUIRES: iverilog,cocotb
+
+// Test the original HandshakeToHW flow.
+
+// RUN: hlstool %s --dynamic-hw --buffering-strategy=cycles --verilog --lowering-options=disallowLocalVariables > %t.sv && \
+// RUN: circt-cocotb-driver.py --objdir=%T --topLevel=top --pythonModule=buffer_arg_init --pythonFolder="%S,%S/.." %t.sv 2>&1 | FileCheck %s
+
+// RUN: hlstool %s --dynamic-hw --buffering-strategy=all --verilog --lowering-options=disallowLocalVariables > %t.sv && \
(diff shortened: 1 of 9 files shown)


teqdruid:
@mortbopet There is some question in my mind as to whether or not `dc.buffer` should support initial values at all. It's only used by HLS, correct? If so, we should probably remove it from DC and do something about it in the HandshakeToDC conversion. Just to keep things abstract, we should maybe introduce some sort of `dc.init_values` which just spits out a series of tokens.

mortbopet:
@teqdruid as of right now, yes, it is only used for HLS... but that's because HLS is the only thing that uses **it**! Based on our offline discussions about buffers, yes, i could see how there may be an argument against buffers. But, DC is intended to be able to represent arbitrary control-flow graphs - HLS is just one user of that. My biggest worry about removing `dc.buffer` is that it's (again, from DHLS' point of view) required to be able to break control-flow loops. But, whether that is constraint that enters at the Handshake level (a generic dataflow constraint), or at the (Handshake | DC) -> HW level is a good question.

I'd be interest in hearing @luisacicolini's thoughts on this. AFAIK, you made an attempt at modelling DC semantics, correct? did you have any interesting conclusions wrt. `dc.buffer`? (did you find it a requirement for the dialect to be able to express the things you needed to express? was it "awkward" to fit it into the abstraction level you were modelling? ...).

mortbopet (approved):

luisacicolini:
Hi @mortbopet @teqdruid :) 
So in our semantics we are currently ignoring `dc.buffer`, under the assumption that the behavior of the overall circuit is in fact _independent_ of the buffers and their placement - also looking at some literature in the area. However, this is still very much WIP, and reasoning about the role of buffers is definitely one of the things that we are stuggling with, too. CC: @ymherklotz

teqdruid:
I'm not arguing against `dc.buffer`. I'm arguing to remove the `initialValues` option from it. Then if HandshakeToDC needs to add some initial values, it is responsible for adding some widget to populate the initial values at every reset.

PR #7947, opened 2024-12-04
[ExportVerilog] Fix ifdef of macro w/ Verilog name

seldridge:
Add a utility function to `sv.macro.decl` that can be used to get its
Verilog textual macro identifier name.  This is added for convenience to
avoid having to repeat the patter of getting its optional Verilog name or
symbol name.

This is done to fix an issue related to lowering of FIRRTL inline layers
where the `LowerLayers` pass generates new `sv.macro.decl` symbols that
will replace existing `firrtl.layer` symbols.  However, it relies on a
namespace to generate these and will result in suffixed symbol names.  (It
could obviously do more work to conserve symbols or generate in-pass
invalid IR with duplicate symbols.)  Nonetheless, the real bug here is
that the `ExportVerilog` conversion is not doing the right thing with the
Verilog name attribute.


uenoku (approved):

uenoku:
LGTM!

darthscsi (approved):

PR #7950, opened 2024-12-05
[SMT] Require SolverOp parent for SetLogicOp

TaoBi22:
Just adds a HasParent<SolverOp> requirement to SetLogicOp


maerhart (approved):
Makes sense to me to restrict this for now.

PR #7952, opened 2024-12-05
[DC] Add + re-enable canonicalization patterns

mortbopet:
Various canonicalization patterns were implemented as fold methods. However, these had been turned off a while ago, due to an MLIR bug.

This PR moves the fold methods to actual canonicalization patterns, adds a new one (join on branch), and re-enables previously disabled tests.


mortbopet:
CC @luisacicolini

teqdruid (approved):

PR #7955, opened 2024-12-05
[InstanceChoice] Add a default override for unspecified options

prithayan:
Add an override that, instead of emitting an error, selects the default target to specialize an instance choice when the option is unspecified.
By default a partially specified instance choice is not allowed. But we can override the error and select the default target, by using `--specialize-to-default-if-no-override` with `firtool`.
This enables a user to specialize the instance choice for a particular build flow, and fall-back to the default for all others.
Relevant discussion: https://github.com/llvm/circt/pull/7933

### DIFF
--- a/include/circt/Dialect/FIRRTL/Passes.td
+++ b/include/circt/Dialect/FIRRTL/Passes.td
@@ -876,6 +876,10 @@ def SpecializeOption : Pass<"firrtl-specialize-option", "firrtl::CircuitOp"> {
   }];
   let constructor = "circt::firrtl::createSpecializeOptionPass()";
 
+  let options = [
+    Option<"selectDefault", "specialize-to-default-if-no-override", "bool", "false",
+      "Specialize instance choice to default, if no option selected.">,
+  ];
   let statistics = [
     Statistic<"numInstances", "num-instances",
       "Number of instances specialized">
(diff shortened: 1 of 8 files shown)


include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -876,6 +876,10 @@ def SpecializeOption : Pass<"firrtl-specialize-option", "firrtl::CircuitOp"> {
   }];
   let constructor = "circt::firrtl::createSpecializeOptionPass()";
 
+  let options = [
+    Option<"selectDefault", "specialize-to-default-if-no-override", "bool", "false",
```
seldridge, line 880: Can the option name (both CLI and variable) be more descriptive than just "specialize"? This is a term that is used also for layers and could mean (roughly) and removal of a parameter. Or: include instance choice in the name.

seldridge (approved):
LGTM

include/circt/Dialect/FIRRTL/Passes.td
prithayan: Renamed both to be a bit more descriptive. Hopefully its not too long,
`select-default-for-unspecified-instance-choice`

seldridge: SGTM. Overly descriptive and obvious is fine-by-me. 💯

PR #7956, opened 2024-12-05
[SCFToCalyx] replace shell command with %T for Windows in the test file

jiahanxie353:
trying to run the workflow


uenoku:
Thanks a lot!

PR #7959, opened 2024-12-06
Fix temporal code motion

AndreyVV-100:
This commit fixes the pass in some cases where llhd.drv value was created
in the same block. The case can be the result of simple verilog code.

For example:

```verilog
module Mod(input a, input clk, output logic b);
always @(posedge clk) begin
    b <= ~a;
end
endmodule
```

Bug can be reproduced by command:

```bash
circt-verilog example.sv | circt-opt --llhd-temporal-code-motion
```

### DIFF
--- a/lib/Dialect/LLHD/Transforms/TemporalCodeMotionPass.cpp
+++ b/lib/Dialect/LLHD/Transforms/TemporalCodeMotionPass.cpp
@@ -88,6 +88,40 @@ getBranchDecisionsFromDominatorToTarget(OpBuilder &builder, Block *driveBlock,
   return mem[driveBlock];
 }
 
+/// Try to move all operations related to $value. This function makes lookup
+/// only in drvOp block. I hope that other dependencies are moveBefore
+/// dominators. It's almost true due to hw.constant and llhd.sig ops are
+/// placed before llhd.process.
+static void moveValueDependencies(llhd::DrvOp drvOp, Operation *moveBefore) {
+  auto *drvBlock = drvOp->getBlock();
+  auto value = drvOp.getValue();
+  if (value.getParentBlock() != drvBlock || !value.getDefiningOp())
+    return;
+  llvm::SmallVector<Operation *> moveSet;
+  moveSet.push_back(value.getDefiningOp());
+
+  // Algorithm is finite due to recursive dependencies can't exist
+  // in one linear block
+  for (unsigned currOp = 0; currOp != moveSet.size(); currOp++) {
+    for (auto arg : moveSet[currOp]->getOperands()) {
+      if (arg.getParentBlock() == drvBlock) {
+        auto *op = arg.getDefiningOp();
+        assert(op && "Using block argument as llhd.drv value is unexpected");
+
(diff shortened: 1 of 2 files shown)


lib/Dialect/LLHD/Transforms/TemporalCodeMotionPass.cpp
```diff
@@ -88,6 +88,45 @@ getBranchDecisionsFromDominatorToTarget(OpBuilder &builder, Block *driveBlock,
...
+  llvm::SmallVector<Operation *> moveSet;
+  moveSet.push_back(value.getDefiningOp());
+
+  // Algorithm is finite due to recursive dependencies can't exist
+  // in single linear block
+  for (unsigned currOp = 0; currOp != moveSet.size(); currOp++) {
+    for (auto arg : moveSet[currOp]->getOperands()) {
+      if (arg.getParentBlock() == drvBlock) {
+        auto *op = arg.getDefiningOp();
+        assert(op && "Using block argument as llhd.drv value is unexpected");
```
maerhart, line 110: This assumes that passes like `block-argument-to-mux` are called beforehand. But the pass shouldn't crash if that's not the case, i.e., this should fail with a proper error message or be supported.

maerhart (commented):
Thanks a lot for finding and fixing this issue!

If you run the `llhd-early-code-motion` pass before temporal code motion, this will not fail. I was thinking about merging the early code motion and temporal code motion passes into an `llhd-simplify-processes` pass. Maybe we should do that instead of repeating some of its logic here. It should already properly handle block arguments and CFG loops. WDYT?

AndreyVV-100:
> If you run the `llhd-early-code-motion` pass before temporal code motion, this will not fail.

That looks it strange but two days ago I tried to use the pass before `llhd-temporal-code-motion` and it didn't help me (and my colleague too). I can't reproduce problem now. I think that's my inattention, sorry for useless PR.

> I was thinking about merging the early code motion and temporal code motion passes into an `llhd-simplify-processes` pass

I agree with you. It's unclear that `llhd-temporal-code-motion` must be executed after `llhd-early-code-motion`.

AndreyVV-100:
> > If you run the `llhd-early-code-motion` pass before temporal code motion, this will not fail.
> 
> That looks it strange but two days ago I tried to use the pass before `llhd-temporal-code-motion` and it didn't help me (and my colleague too). I can't reproduce problem now. I think that's my inattention, sorry for useless PR.
> 
> > I was thinking about merging the early code motion and temporal code motion passes into an `llhd-simplify-processes` pass
> 
> I agree with you. It's unclear that `llhd-temporal-code-motion` must be executed after `llhd-early-code-motion`.

I think that I was confused due to #7963 issue. The PR may be closed.

lib/Dialect/LLHD/Transforms/TemporalCodeMotionPass.cpp
AndreyVV-100: > I was thinking about merging
Can I try to implement this? Should old passes be deprecated or removed? Should corresponding files be merged or can we just call passes successively in third file?

PR #7960, opened 2024-12-07
[FIRRTL][CAPI] Add more functions for discriminating and querying type

SpriteOvO:
This PR adds functions:
- for const `firrtlTypeIsConst`, `firrtlTypeGetConstType`
- for bit width `firrtlTypeGetBitWidth`
- for discriminating `firrtlTypeIsA{UInt,SInt,Clock,Reset,...}`
- for vector `firrtlTypeGetVectorElement`, `firrtlTypeGetVectorNumElements`
- for bundle `firrtlTypeGetBundleNumFields`, `firrtlTypeGetBundleFieldByIndex`

CC @sequencer, could you check if there are any missing?

### DIFF
--- a/include/circt-c/Dialect/FIRRTL.h
+++ b/include/circt-c/Dialect/FIRRTL.h
@@ -105,27 +105,73 @@ MLIR_DECLARE_CAPI_DIALECT_REGISTRATION(FIRRTL, firrtl);
 // Type API.
 //===----------------------------------------------------------------------===//
 
+/// Returns `true` if this is a const type whose value is guaranteed to be
+/// unchanging at circuit execution time.
+MLIR_CAPI_EXPORTED bool firrtlTypeIsConst(MlirType type);
+
+/// Returns a const or non-const version of this type.
+MLIR_CAPI_EXPORTED MlirType firrtlTypeGetConstType(MlirType type, bool isConst);
+
+/// Gets the bit width for this type, returns -1 if unknown.
+///
+/// It recursively computes the bit width of aggregate types. For bundle and
+/// vectors, recursively get the width of each field element and return the
+/// total bit width of the aggregate type. This returns -1, if any of the bundle
+/// fields is a flip type, or ground type with unknown bit width.
+MLIR_CAPI_EXPORTED int64_t firrtlTypeGetBitWidth(MlirType type,
+                                                 bool ignoreFlip);
+
+/// Checks if this type is a unsigned integer type.
+MLIR_CAPI_EXPORTED bool firrtlTypeIsAUInt(MlirType type);
+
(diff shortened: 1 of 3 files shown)


include/circt-c/Dialect/FIRRTL.h
```diff
@@ -105,28 +105,73 @@ MLIR_DECLARE_CAPI_DIALECT_REGISTRATION(FIRRTL, firrtl);
...
+MLIR_CAPI_EXPORTED bool firrtlTypeIsConst(MlirType type);
+
+/// Returns a const or non-const version of this type.
+MLIR_CAPI_EXPORTED MlirType firrtlTypeGetConstType(MlirType type, bool isConst);
+
+/// Gets the bit width for this type, returns -1 if unknown.
+///
+/// It recursively compute the bit width of aggregate types. For bundle and
+/// vectors, recursively get the width of each field element and return the
+/// total bit width of the aggregate type. This returns -1, if any of the bundle
```
sequencer, line 119: > This returns -1, if any of the bundle fields is a flip type

why

sequencer (commented):
The capi definition basically looks good to me.

include/circt-c/Dialect/FIRRTL.h
SpriteOvO: This is the design of https://github.com/llvm/circt/blob/cd02a733b22cd8a8f45b860c6caebd5621347b6e/include/circt/Dialect/FIRRTL/FIRRTLTypes.h#L376-L383

But I forgot to present the `ignoreFlip` parameter in C-API, I should add it as well.

SpriteOvO: Updated~

test/CAPI/firrtl.c
```diff
@@ -186,6 +186,131 @@ void testAttrGetIntegerFromString(MlirContext ctx) {
                       mlirStringRefCreateFromCString("114514"), 10));
 }
 
+void testTypeDiscriminantsAndQuarries(MlirContext ctx) {
```
mikeurbach, line 189: nit: I think you meant `...AndQueries`

mikeurbach (approved):
One minor nitpick on the name of the test function, but this looks good to me. Thanks for these new C APIs... it's annoying but this is still the best way to implement them. And thanks for including unit tests!

test/CAPI/firrtl.c
SpriteOvO: Yeah, right. Fixed

SpriteOvO:
@mikeurbach Thank you for reviewing this!

seldridge (commented):
Windows builds are failing with this for a reason that wasn't immediately obvious to me.

PR #7965, opened 2024-12-10
[OM] Generalize handling for list creation ops in FreezePaths.

mikeurbach:
We previously handled the ListCreateOp specifically, because lists of paths need to become lists of frozen paths. But there are other list creation ops that need to be considered and handled similarly, like the recently added ListConcatOp.

To handle this, the process method for ListCreateOp was updated to work on any generic Operation * that returns a ListType. The typeswitch that dispatches to the process methods was updated to use this generic processor for both ListCreateOp and ListConcatOp. I thought about writing a generic check instead of listing out the supported Operation classes, but that seems like a fragile tradeoff that might not be worth the cost relative to keeping this list up to date.


uenoku (approved):
LGTM

PR #7967, opened 2024-12-11
[CombToAIG] Add a pattern for lowering varidaic operations

uenoku:
Some variadic operations (xor, add, mul etc) cannot directly lowered into varidic AIG operations. So this adds a pattern to pre-lower variadic operations.


fabianschuiki (approved):
Nice, LGTM!

PR #7971, opened 2024-12-11
[ImportVerilog] Support for sized unpacked arrays in 'inside' expressions

ladisgin:
Support for sized unpacked arrays in 'inside' expressions. Implementation recursively traverses the array to find the singular value and compares all found ones with the lhs, adding the condition results to the original conditions vector.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -554,19 +554,35 @@ struct RvalueExprVisitor {
         // Handle expressions.
         if (!listExpr->type->isSimpleBitVector()) {
           if (listExpr->type->isUnpackedArray()) {
+            if (listExpr->type->isFixedSize()) {
+              const auto &uaType =
+                  listExpr->type->as<slang::ast::FixedSizeUnpackedArrayType>();
+              auto value = context.convertRvalueExpression(*listExpr);
+              if (!value)
+                return {};
+              context.collectConditionsForUnpackedArray(uaType, value,
+                                                        conditions, lhs, loc);
+              cond = conditions.back();
+              conditions
+                  .pop_back(); // avoiding repetition of cond in the vector
+            } else {
+              mlir::emitError(loc, "unsized unpacked arrays in 'inside' "
+                                   "expressions not supported");
+              return {};
+            }
+          } else {
             mlir::emitError(
-                loc, "unpacked arrays in 'inside' expressions not supported");
+                loc,
+                "only simple bit vectors supported in 'inside' expressions");
             return {};
           }
-          mlir::emitError(
-              loc, "only simple bit vectors supported in 'inside' expressions");
-          return {};
+        } else {
+          auto value = context.convertToSimpleBitVector(
+              context.convertRvalueExpression(*listExpr));
+          if (!value)
+            return {};
+          cond = builder.create<moore::WildcardEqOp>(loc, lhs, value);
         }
-        auto value = context.convertToSimpleBitVector(
-            context.convertRvalueExpression(*listExpr));
-        if (!value)
-          return {};
-        cond = builder.create<moore::WildcardEqOp>(loc, lhs, value);
       }
       conditions.push_back(cond);
     }
@@ -1100,4 +1116,40 @@ Value Context::materializeConversion(Type type, Value value, bool isSigned,
   if (value.getType() != type)
     value = builder.create<moore::ConversionOp>(loc, type, value);
   return value;
+}
+
+void Context::collectConditionsForUnpackedArray(
+    const slang::ast::FixedSizeUnpackedArrayType &slangType,
+    Value upackedArrayValue, SmallVector<Value> &conditions, Value lhs,
+    Location loc) {
+  Value cond;
+  auto type = convertType(slangType);
+  if (!type) {
+    mlir::emitError(loc, "can't convert slang::ast::FixedSizeUnpackedArrayType "
+                         "to moore::UnpackedArrayType");
+  }
+  auto mooreType = dyn_cast<moore::UnpackedArrayType>(type);
+  const auto &elementType = slangType.elementType;
+  for (slang::int32_t i = slangType.getFixedRange().lower();
+       i <= slangType.getFixedRange().upper(); i++) {
+    auto elemValue = builder.create<moore::ExtractOp>(
+        loc, mooreType.getElementType(), upackedArrayValue, i);
+    if (elementType.isUnpackedArray()) {
+      collectConditionsForUnpackedArray(
+          elementType.as<slang::ast::FixedSizeUnpackedArrayType>(), elemValue,
+          conditions, lhs, loc);
+    } else if (elementType.isSingular()) {
+      if (elementType.isIntegral()) {
+        cond = builder.create<moore::WildcardEqOp>(loc, lhs, elemValue);
+      } else {
+        cond = builder.create<moore::EqOp>(loc, lhs, elemValue);
+      }
+      conditions.push_back(cond);
+    } else {
+      mlir::emitError(loc,
+                      "only singular values and fixed-size unpacked arrays "
+                      "allowed as elements of unpacked arrays in 'inside' "
+                      "expressions");
+    }
+  }
 }
--- a/test/Conversion/ImportVerilog/errors.sv
+++ b/test/Conversion/ImportVerilog/errors.sv
@@ -68,8 +68,8 @@ endmodule
 
 // -----
 module Foo;
-  int a, b[3];
-  // expected-error @below {{unpacked arrays in 'inside' expressions not supported}}
+  int a, b[];
+  // expected-error @below {{unsized unpacked arrays in 'inside' expressions not supported}}
   int c = a inside { b };
 endmodule
 
--- a/lib/Conversion/ImportVerilog/ImportVerilogInternals.h
+++ b/lib/Conversion/ImportVerilog/ImportVerilogInternals.h
@@ -135,6 +135,13 @@ struct Context {
   Value materializeConstant(const slang::ConstantValue &constant,
                             const slang::ast::Type &type, Location loc);
 
+  /// Helper function to construct conditions for a fixed-size unpacked array in
+  /// an `inside` expression.
+  void collectConditionsForUnpackedArray(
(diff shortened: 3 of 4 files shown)


hailongSun2000 (approved):
LGTM :100: Thanks @ladisgin a lot for this work :smile:!
@fabianschuiki WDYT :thinking:

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -554,19 +554,35 @@ struct RvalueExprVisitor {
...
+              const auto &uaType =
+                  listExpr->type->as<slang::ast::FixedSizeUnpackedArrayType>();
+              auto value = context.convertRvalueExpression(*listExpr);
+              if (!value)
+                return {};
+              context.collectConditionsForUnpackedArray(uaType, value,
+                                                        conditions, lhs, loc);
+              cond = conditions.back();
+              conditions
+                  .pop_back(); // avoiding repetition of cond in the vector
```
darthscsi, line 567: super nit:  comment before the line to keep the code on one line.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -554,19 +554,35 @@ struct RvalueExprVisitor {
...
+            } else {
+              mlir::emitError(loc, "unsized unpacked arrays in 'inside' "
```
darthscsi, line 569: When you do the LogicalResult conversion, you can just `return mlir::emitError`.

It might be useful to emit the error on the op?  I'm not sure, I haven't thought about it.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1101,3 +1117,39 @@ Value Context::materializeConversion(Type type, Value value, bool isSigned,
...
+void Context::collectConditionsForUnpackedArray(
+    const slang::ast::FixedSizeUnpackedArrayType &slangType,
+    Value upackedArrayValue, SmallVector<Value> &conditions, Value lhs,
+    Location loc) {
+  Value cond;
+  auto type = convertType(slangType);
+  if (!type) {
+    mlir::emitError(loc, "can't convert slang::ast::FixedSizeUnpackedArrayType "
+                         "to moore::UnpackedArrayType");
+  }
```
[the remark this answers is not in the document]
darthscsi: And to add to Fabian's comment, make sure your tests hit all these error message cases.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1101,3 +1117,39 @@ Value Context::materializeConversion(Type type, Value value, bool isSigned,
     value = builder.create<moore::ConversionOp>(loc, type, value);
   return value;
 }
+
+void Context::collectConditionsForUnpackedArray(
+    const slang::ast::FixedSizeUnpackedArrayType &slangType,
+    Value upackedArrayValue, SmallVector<Value> &conditions, Value lhs,
+    Location loc) {
+  Value cond;
```
darthscsi, line 1125: This looks like it can be sunk into the condition branch in the loop.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1101,3 +1117,39 @@ Value Context::materializeConversion(Type type, Value value, bool isSigned,
...
+      collectConditionsForUnpackedArray(
+          elementType.as<slang::ast::FixedSizeUnpackedArrayType>(), elemValue,
+          conditions, lhs, loc);
+    } else if (elementType.isSingular()) {
+      if (elementType.isIntegral()) {
+        cond = builder.create<moore::WildcardEqOp>(loc, lhs, elemValue);
+      } else {
+        cond = builder.create<moore::EqOp>(loc, lhs, elemValue);
+      }
+      conditions.push_back(cond);
```
darthscsi, line 1147: From the comment above, the cond can sink towards here.  Or be eliminated by switching to a trinary instead of a if.  Or elminated by duplicating the push_back.

test/Conversion/ImportVerilog/errors.sv
```diff
@@ -68,8 +68,8 @@ endmodule
 
 // -----
 module Foo;
-  int a, b[3];
-  // expected-error @below {{unpacked arrays in 'inside' expressions not supported}}
+  int a, b[];
```
darthscsi, line 71: you added a lot more errors.  Please add checks for them.

hailongSun2000 (commented):
Roll back a approve

hailongSun2000 (changes requested):
Roll back an approval.

ladisgin:
Hi, sorry for the long reply. I added the return value `LogicalResult`. But I do not know what to do with the tests for my errors. It looks like these type inconsistencies are checked in slang. Should I change them to assert?

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1101,3 +1115,40 @@ Value Context::materializeConversion(Type type, Value value, bool isSigned,
...
+        conditions.push_back(
+            builder.create<moore::WildcardEqOp>(loc, lhs, elemValue));
+      } else {
+        conditions.push_back(builder.create<moore::EqOp>(loc, lhs, elemValue));
+      }
+    } else {
+      return mlir::emitError(
+          loc, "only singular values and fixed-size unpacked arrays "
+               "allowed as elements of unpacked arrays in 'inside' "
+               "expressions");
```
hailongSun2000, line 1150: I guess you need to add a case to check/trigger it. Every `mlir::emitError` manually added needs to be tested.

ladisgin: I can't trigger it, slang already has same check for `inside` in `bindMembershipExpressions` https://github.com/MikePopoloski/slang/blob/7efcca2e348ecf1e01d1859f7ea85d6202fc9f91/source/ast/expressions/OperatorExpressions.cpp#L216

hailongSun2000: Namely, slang will throw an error if the `elementType` is not either fixed-size or singular. We needn't check repeatedly. So you can remove this `mlir::emitError` :smile: .

ladisgin:
@hailongSun2000
Can you take a look?

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1410,3 +1424,39 @@ Context::convertSystemCallArity1(const slang::ast::SystemSubroutine &subroutine,
...
+
+LogicalResult Context::collectConditionsForUnpackedArray(
+    const slang::ast::FixedSizeUnpackedArrayType &slangType,
+    Value upackedArrayValue, SmallVector<Value> &conditions, Value lhs,
+    Location loc) {
+  auto type = convertType(slangType);
+  if (!type) {
+    return failure();
+  }
+  auto mooreType = dyn_cast<moore::UnpackedArrayType>(type);
```
sunhailong2001, line 1436: nit: There will be a potential error. We must check the result of `dyn_cast`.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1410,3 +1424,39 @@ Context::convertSystemCallArity1(const slang::ast::SystemSubroutine &subroutine,
           .Default([&]() -> Value { return {}; });
   return systemCallRes();
 }
+
+LogicalResult Context::collectConditionsForUnpackedArray(
+    const slang::ast::FixedSizeUnpackedArrayType &slangType,
+    Value upackedArrayValue, SmallVector<Value> &conditions, Value lhs,
```
sunhailong2001, line 1430: nit: `upackedArrayValue` ---> `unpackedArrayValue`

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1410,3 +1424,39 @@ Context::convertSystemCallArity1(const slang::ast::SystemSubroutine &subroutine,
...
+  auto mooreType = dyn_cast<moore::UnpackedArrayType>(type);
+  const auto &elementType = slangType.elementType;
+  for (slang::int32_t i = slangType.getFixedRange().lower();
+       i <= slangType.getFixedRange().upper(); i++) {
+    auto elemValue = builder.create<moore::ExtractOp>(
+        loc, mooreType.getElementType(), upackedArrayValue, i);
+    if (elementType.isUnpackedArray()) {
+      if (failed(collectConditionsForUnpackedArray(
+              elementType.as<slang::ast::FixedSizeUnpackedArrayType>(),
+              elemValue, conditions, lhs, loc))) {
```
sunhailong2001, line 1445: I think `elementType.as<>` may cause some troubles. Because `.as<>` is similar to `cast<>`. Could you test `int(or other types) [3:0][1:0] ua2d[2][2];`?

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -606,26 +605,41 @@ struct RvalueExprVisitor {
...
-          mlir::emitError(
-              loc, "only simple bit vectors supported in 'inside' expressions");
-          return {};
+        } else {
+          auto value = context.convertToSimpleBitVector(
+              context.convertRvalueExpression(*listExpr));
+          if (!value)
+            return {};
+          conditions.push_back(
+              builder.create<moore::WildcardEqOp>(loc, lhs, value));
```
sunhailong2001, line 640: nit: Could you attempt whether `listExpr->type>as_if<slang::ast::FixedSizeUnpackedArrayType>()` is valid?

My thought is that maybe we can tweak this code snippet into 
```
else {
  auto value = context.convertRvalueExpression(*listExpr));
  if (!value)
    return {};
  if(listExpr->type->isSimpleBitVector()) {
    value = context.convertToSimpleBitVector()
    if(!value)
      return {};
      ...
  }
  else if (listExpr->type->as_if<slang::ast::FixedSizeUnpackedArrayType>()) {
    ...
  }
  else {
    // my thought: Merge all into one (Is it better to reduce if-else layers judgments 🤔 ?)
    mlir::emitError(loc, 'inside' expressions not support)<< listExpr->type;
  }
}
```
What do @fabianschuiki think about reducing if-else layers 😃 ?

sunhailong2001 (commented):
I'm so sorry for the late reply! Due to job changes, I just have a little spare time right now.

Issue #7711, opened 2024-10-17
[circt-lec] Report failing points


uenoku:
```mlir
$ cat bar.mlir
hw.module @A(out out_1: i12, out out_2: i12) {
  %0 = hw.constant 2: i12
  hw.output %0, %0 : i12, i12
}

hw.module @B(out out_1 : i12, out out_2: i12) {
  %0 = hw.constant 2: i12
  %1 = hw.constant 3: i12
  hw.output %0, %1 : i12, i12
}
$ circt-lec bar.mlir --c1 A --c2 B --shared-libs=z3/build/libz3.so
c1 != c2
```

It would be nice to report failing points (in this case `out_2`).

hovind:
> ```mlir
> $ cat bar.mlir
> hw.module @A(out out_1: i12, out out_2: i12) {
>   %0 = hw.constant 2: i12
>   hw.output %0, %0 : i12, i12
> }
> 
> hw.module @B(out out_1 : i12, out out_2: i12) {
>   %0 = hw.constant 2: i12
>   %1 = hw.constant 3: i12
>   hw.output %0, %1 : i12, i12
> }
> $ circt-lec bar.mlir --c1 A --c2 B --shared-libs=z3/build/libz3.so
> c1 != c2
> ```
> 
> Right now it only tells us yes or no, so it would be nice to report failing points ideally with (executable) counter examples

A very crude way I've done this locally is
```diff
diff --git a/include/circt/Dialect/Verif/VerifOps.td b/include/circt/Dialect/Verif/VerifOps.td
index 7b83a6dcc..27e3cfee9 100644
--- a/include/circt/Dialect/Verif/VerifOps.td
+++ b/include/circt/Dialect/Verif/VerifOps.td
@@ -237,6 +237,8 @@ def LogicEquivalenceCheckingOp : VerifOp<"lec", [
   }];
 
   let results = (outs I1:$areEquivalent);
+  let arguments = (ins StrArrayAttr:$inputNames);
+
   let regions = (region SizedRegion<1>:$firstCircuit,
                         SizedRegion<1>:$secondCircuit);
 
diff --git a/lib/Conversion/VerifToSMT/VerifToSMT.cpp b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
index 57ef512d1..ee5b86666 100644
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
[...]

PR #7974, opened 2024-12-12
[circt-bmc] Add option to print solver output & assertions

TaoBi22:
Have found myself enabling debug on LowerSMTToZ3LLVM quite a lot for debugging, figured it's probably useful enough to have as a flag in circt-bmc (e.g. to figure out why a design's assertions are failing, though obviously as in #7711
ideally there'll be a nicer way to do this eventually).


maerhart (approved):
LGTM!

PR #7977, opened 2024-12-12
[SCFToCalyx] buildLibraryOp cast floating point to integer types

jiahanxie353:
In `buildLibraryOp`, we cast floating point types to integer types with the same width, as discussed here: https://github.com/llvm/circt/pull/7762

The testing case is just the result of running the `-arith-expand` pass on the following input:
```mlir
module {
  func.func @main(%b: f64, %c: f64) -> f64 {
    %a = arith.maximumf %b, %c : f64
    return %a : f64
  }
}
```

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -62,9 +62,10 @@ bool noStoresToMemory(Value memoryReference);
 // Get the index'th output port of compOp.
 Value getComponentOutput(calyx::ComponentOp compOp, unsigned outPortIdx);
 
-// If the provided type is an index type, converts it to i32, else, returns the
-// unmodified type.
-Type convIndexType(OpBuilder &builder, Type type);
+// If the provided type is an index type, converts it to i32; else if the
+// provided is an integer of floating point, bitcasts to a signless integer
+// type; otherwise, returns the unmodified type.
+Type normalizeType(OpBuilder &builder, Type type);
 
 // Creates a new calyx::CombGroupOp or calyx::GroupOp group within compOp.
 template <typename TGroup>
@@ -837,6 +838,20 @@ struct PredicateInfo {
 };
 
 PredicateInfo getPredicateInfo(mlir::arith::CmpFPredicate pred);
+
+/// Performs a bit cast from a non-signless integer type value, such as a
+/// floating point value, to a signless integer type. Calyx treats everything as
+/// bit vectors, and leaves their interpretation to the respective operation
+/// using it. In CIRCT Calyx, we use signless `IntegerType` to represent a bit
+/// vector.
+template <typename T>
+Type toBitVector(T type) {
+  if (!type.isSignlessInteger()) {
+    unsigned bitWidth = cast<T>(type).getIntOrFloatBitWidth();
+    return IntegerType::get(type.getContext(), bitWidth);
+  }
+  return type;
+};
 
 } // namespace calyx
 } // namespace circt
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -63,7 +63,7 @@ bool noStoresToMemory(Value memoryReference);
 Value getComponentOutput(calyx::ComponentOp compOp, unsigned outPortIdx);
 
 // If the provided type is an index type, converts it to i32; else if the
-// provided is an integer of floating point, bitcasts to a signless integer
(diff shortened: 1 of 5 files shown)


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -396,9 +396,19 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
   template <typename TGroupOp, typename TCalyxLibOp, typename TSrcOp>
   LogicalResult buildLibraryOp(PatternRewriter &rewriter, TSrcOp op,
                                TypeRange srcTypes, TypeRange dstTypes) const {
+    auto castToInteger = [](Type type) -> Type {
```
cgyurgyik, line 399:
```suggestion
    /// Performs a bit cast from the floating point value to a signless integer type. Calyx treats everything as bit vectors, and leaves their interpretation to the respective operation using it. In CIRCT Calyx, we use `IntegerType` to represent a bit vector.
    auto castToInteger = [](Type type) -> Type {
```

cgyurgyik: I would add this comment to make it clear. I think this function would be useful in the `calyx_utils` file too, since other folks will likely use it.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -396,9 +396,19 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
...
+      if (isa<FloatType>(type)) {
```
cgyurgyik, line 400: Maybe make this a generic function as well, i.e.,

```
template <typename Type>
...
if (!isa<IntegerType>(Type)) {
  ...
}
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
cgyurgyik, line 399: Nit: maybe `toBitVector` or `bitCast` or something similar to make it clear we're performing a bit cast and not changing the value.

test/Conversion/SCFToCalyx/convert_simple.mlir
```diff
@@ -438,3 +438,72 @@ module {
...
+// CHECK-DAG:        calyx.assign %std_mux_0.cond = %cmpf_0_reg.out : i1
+// CHECK-DAG:        calyx.assign %std_mux_0.tru = %in0 : i64
+// CHECK-DAG:        calyx.assign %std_mux_0.fal = %in1 : i64
+// CHECK-DAG:        calyx.group_done %ret_arg0_reg.done : i1
+// CHECK-DAG:      }
+// CHECK-DAG:    }
+
+module {
+  func.func @main(%arg0: f64, %arg1: f64) -> f64 {
+    %0 = arith.cmpf ugt, %arg0, %arg1 : f64
```
cgyurgyik, line 502: Can you add a test for signed integers as well please?

cgyurgyik (commented):
Some small nits :)

test/Conversion/SCFToCalyx/convert_simple.mlir
jiahanxie353: I just tried, and strange enough, every operation in the arith dialect either takes floating point or signless integer type. There is even no way to do: `divsi` on two `si64` types..

jiahanxie353: nvm `select` takes `anyType`s

jiahanxie353:
Thanks @cgyurgyik ! 

Made some changes based on your feedback; also changed `convIndex` to a more generic function by utilizing the `toBitVector` function.

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -62,9 +62,10 @@ bool noStoresToMemory(Value memoryReference);
 // Get the index'th output port of compOp.
 Value getComponentOutput(calyx::ComponentOp compOp, unsigned outPortIdx);
 
-// If the provided type is an index type, converts it to i32, else, returns the
-// unmodified type.
-Type convIndexType(OpBuilder &builder, Type type);
+// If the provided type is an index type, converts it to i32; else if the
+// provided is an integer of floating point, bitcasts to a signless integer
```
cgyurgyik, line 66:
```suggestion
// provided is an integer or floating point, bitcasts it to a signless integer
```

cgyurgyik (approved):
Awesome, thanks!

PR #7978, opened 2024-12-12
[RTG][Elaboration] Add support for 'index.add' and 'index.cmp'

maerhart:
* Use the index dialect instead of arith because it is a smaller dependency and simpler, but still has everything we need.
* Add support for `index.add` and `index.cmp`
* Add specialized elaborator values for indices and booleans because creating attributes for intermediates is quite bad performance-wise (ideally we'd have some evaluation interface at some point that we can also use for folding, or better folders that we can call directly)

### DIFF
--- a/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
+++ b/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
@@ -17,7 +17,8 @@
 #include "circt/Dialect/RTG/IR/RTGVisitors.h"
 #include "circt/Dialect/RTG/Transforms/RTGPasses.h"
 #include "circt/Support/Namespace.h"
-#include "mlir/Dialect/Arith/IR/Arith.h"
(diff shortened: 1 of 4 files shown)


lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -85,7 +86,7 @@ namespace {
 /// The abstract base class for elaborated values.
 struct ElaboratorValue {
```
darthscsi, line 87: This seems like just a std::variant.

darthscsi (approved):
LGTM.  I think the elaboration is reinventing std::variant, but that is separate from this PR.

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
maerhart: I'll address this in https://github.com/llvm/circt/pull/7997

PR #7979, opened 2024-12-12
[RTG] Add InstructionOpInterface

maerhart:


include/circt/Dialect/RTG/IR/RTGISAAssemblyInterfaces.td
```diff
@@ -49,9 +49,42 @@ def RegisterOpInterface : OpInterface<"RegisterOpInterface"> {
...
+      "::llvm::APInt", "getInstructionBinary",
+      (ins "llvm::ArrayRef<llvm::APInt>":$operands)>,
+    InterfaceMethod<[{
+        Emits the ISA assembly representation of the instruction to the provided
+        stream. The format should be the one understood by common assembler
+        tools.
+      }],
+      "void", "printInstructionAssembly",
+      (ins "llvm::raw_ostream &":$os,
+           "llvm::ArrayRef<llvm::function_ref<void(llvm::raw_ostream &)>>":
```
darthscsi, line 79: taking arrays of lambdas is general, but low performance.  Emiting shouldn't be the bottleneck, but worth keeping in mind.  

It seems that since the dialect defines the type of each op's operands, the operand would know how to print it's operands without indirection.

I'd also avoid having a default implementation as there is no sensible default.

youngar (approved):

PR #7981, opened 2024-12-13
[firtool] Add option to disable layer sink

youngar:
The layer sink pass is showing some performance problems, and we want the ability to turn it off to work around the issue, until we can properly fix it.

### DIFF
--- a/include/circt/Firtool/Firtool.h
+++ b/include/circt/Firtool/Firtool.h
@@ -96,6 +96,7 @@ public:
   }
   bool shouldConvertProbesToSignals() const { return probesToSignals; }
   bool shouldReplaceSequentialMemories() const { return replSeqMem; }
+  bool shouldDisableLayerSink() const { return disableLayerSink; }
   bool shouldDisableOptimization() const { return disableOptimization; }
   bool shouldAdvancedLayerSink() const { return advancedLayerSink; }
   bool shouldLowerMemories() const { return lowerMemories; }
@@ -186,6 +187,11 @@ public:
 
   FirtoolOptions &setBuildMode(BuildMode value) {
     buildMode = value;
+    return *this;
+  }
+
+  FirtoolOptions &setDisableLayerSink(bool value) {
+    disableLayerSink = value;
     return *this;
   }
 
@@ -381,6 +387,7 @@ private:
   firrtl::PreserveValues::PreserveMode preserveMode;
   bool enableDebugInfo;
   BuildMode buildMode;
+  bool disableLayerSink;
   bool disableOptimization;
   bool exportChiselInterface;
   std::string chiselInterfaceOutDirectory;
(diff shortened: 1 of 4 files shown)


include/circt/Firtool/Firtool.h
seldridge, line 1: Even though this is short lived, should this include updates to the C-API?

lib/Firtool/Firtool.cpp
```diff
@@ -228,11 +228,14 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
...
-    pm.nest<firrtl::CircuitOp>().nest<firrtl::FModuleOp>().addPass(
-        firrtl::createLayerSinkPass());
+  if (opt.shouldDisableLayerSink()) {
+    if (opt.shouldAdvancedLayerSink())
+      pm.nest<firrtl::CircuitOp>().addPass(
+          firrtl::createAdvancedLayerSinkPass());
+    else
+      pm.nest<firrtl::CircuitOp>().nest<firrtl::FModuleOp>().addPass(
+          firrtl::createLayerSinkPass());
+  }
```
seldridge, line 238: Given that we now have two options, it doesn't seem great that ALS is gated by another option. I.e., `-advanced-layer-sink` has no effect if `-disable-layer-sink` is set.

Would it be better to instead just make two simple options:

1. `-advanced-layer-sink` : schedules advanced layer sink
2. `-layer-sink` : schedules layer sink

Alternatively an enum that has the three actually interesting states: `-layer-sink=none,simple,advanced`.

seldridge (approved):
Some comments.

This is expected to be short-lived while ALS gets turned on, so the design isn't hugely important.

lib/Firtool/Firtool.cpp
youngar: I figured since we're going to fully replace LS with ALS and removing `--advanced-layer-sink`, that this would have the least churn. Then, we could keep the `disable-layer-sink` flag as long as we find it useful.  Moving to an enum would have to be reversed when we remove either flag.  I'm not sure about the first idea, I guess we would have to disable layersink by default to make it work?

include/circt/Firtool/Firtool.h
youngar: I don't think the CAPI has been kept up to date properly, but I can try to figure this out.

lib/Firtool/Firtool.cpp
seldridge: Yeah, it would have to be default of LS on and ALS off if there are two options.

rwy7 (approved):

include/circt/Firtool/Firtool.h
youngar: updated

dtzSiFive:
LGTM! Just a thought, maybe disable layer sink entirely if it's not used presently and problematic (until ALS is enabled, which already has an option)?

PR #7983, opened 2024-12-13
[FIRRTL] Use PRINTF_FD macro instead of 0x80000002 as printf fd

Clo91eaf:
updates the FIRRTLToHW conversion to use the `PRINTF_FD` macro for specifying the file descriptor, allowing the fd to be obtained externally. This change enhances flexibility by enabling the file descriptor to be defined outside the conversion logic.

now we can define a log package like this:
```
package log_pkg;
  bit log_cond;
  int log_fd;

  function automatic void log_open(string log_path);
    log_fd = $fopen(log_path, "w");
    if (log_fd == 0) $fatal(1, "failed to open file for write");
    log_cond = 1'b1;
  endfunction

  function automatic void log_close();
    if (log_cond) begin
      $fclose(log_fd);
      log_cond = 1'b0;
      log_fd = 0;
    end
  endfunction
endpackage

initial begin
  log_pkg::log_open("xxx");
end

final begin
  log_pkg::log_close();
end
```

and add compile option to import log_fd
```
+define+PRINTF_FD=log_pkg::log_fd
+define+PRINTF_COND=log_pkg::log_cond
```

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -209,7 +209,7 @@ struct FIRRTLModuleLowering;
 /// This is state shared across the parallel module lowering logic.
 struct CircuitLoweringState {
   // Flags indicating whether the circuit uses certain header fragments.
-  std::atomic<bool> usedPrintfCond{false};
+  std::atomic<bool> usedPrintf{false};
   std::atomic<bool> usedAssertVerboseCond{false};
   std::atomic<bool> usedStopCond{false};
 
@@ -809,7 +809,19 @@ void FIRRTLModuleLowering::lowerFileHeader(CircuitOp op,
         guard, []() {}, body);
   };
 
-  if (state.usedPrintfCond) {
+  if (state.usedPrintf) {
+    b.create<sv::MacroDeclOp>("PRINTF_FD");
+    b.create<sv::MacroDeclOp>("PRINTF_FD_");
+    b.create<emit::FragmentOp>("PRINTF_FD_FRAGMENT", [&] {
+      b.create<sv::VerbatimOp>(
+          "\n// Users can define 'PRINTF_FD' to add a specified fd to "
+          "prints.");
+      emitGuard("PRINTF_FD_", [&]() {
+        emitGuardedDefine("PRINTF_FD", "PRINTF_FD_", "(`PRINTF_FD)",
+                          "32'h80000002");
+      });
+    });
+
     b.create<sv::MacroDeclOp>("PRINTF_COND");
     b.create<sv::MacroDeclOp>("PRINTF_COND_");
     b.create<emit::FragmentOp>("PRINTF_COND_FRAGMENT", [&] {
@@ -4416,7 +4428,8 @@ LogicalResult FIRRTLLowering::visitStmt(PrintFOp op) {
   circuitState.addMacroDecl(builder.getStringAttr("SYNTHESIS"));
   addToIfDefBlock("SYNTHESIS", std::function<void()>(), [&]() {
     addToAlwaysBlock(clock, [&]() {
-      circuitState.usedPrintfCond = true;
+      circuitState.usedPrintf = true;
+      circuitState.addFragment(theModule, "PRINTF_FD_FRAGMENT");
       circuitState.addFragment(theModule, "PRINTF_COND_FRAGMENT");
 
       // Emit an "sv.if '`PRINTF_COND_ & cond' into the #ifndef.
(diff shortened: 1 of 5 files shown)


seldridge:
@darthscsi: What do you think about this?

We can just make this part of the FIRRTL ABI (which `PRINTF_COND` already should be).

This would be less configurable than other proposals (e.g., https://github.com/chipsalliance/firrtl-spec/pull/213). However, it avoids some of the problems of things like introducing file descriptors into FIRRTL.

The downside of this is that it would take some clever SystemVerilog to be configurable. E.g., I expect that we could include an evolving SystemVerilog file which uses the `PRINTF_COND` and `PRINTF_FD` to control what prints are enabled and to which files certain prints go to. This would require the SystsemVerilog to be more complicated. However, that is likely much better than complicating the FIRRTL unnecessarily.

WDYT?

seldridge:
CC: @girishpai: As we circle back to ideas you have had about a "logging library" for Chisel, I am wondering if only `PRINTF_COND` and `PRINTF_FD` with sufficient clever SystemVerilog is enough. SystemVerilog is such a large language that I expect we could do something here. Anyone, I'm curious if you have thoughts about this, too.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -805,11 +805,22 @@ void FIRRTLModuleLowering::lowerFileHeader(CircuitOp op,
...
+    b.create<sv::IfDefOp>(guard, []() {}, body);
   };
 
-  if (state.usedPrintfCond) {
+  if (state.usedPrintf) {
+    b.create<sv::MacroDeclOp>("PRINTF_FD");
+    b.create<sv::MacroDeclOp>("PRINTF_FD_");
+    b.create<emit::FragmentOp>("PRINTF_FD_FRAGMENT", [&] {
+      b.create<sv::VerbatimOp>(
+          "\n// Users can define 'PRINTF_FD' to add a specified fd to "
```
darthscsi, line 816: Do we add comments for other chisel macros?  I didn't think we do.  There should be a user friendly form of the chisel abi which outlines the macros created to control these things.

darthscsi:
I don't have an opposition to a minimal change like this.  My concern is it might not be general enough to really accomplish what everyone wants.  But it is probably better to use the macro than a constant.  As long as everyone involved is happy to replace it in the nearish future with something a bit better, I'm fine.  It doesn't need to touch the firrtl abi as it is just providing an implementation dependent hook (yes, yes) for dynamically changing the abi behavior.

fzi-hielscher:
I'm also fine with this as a short term solution. It could also easily be adopted in #7973 .

My half-baked plan for the future is to add an optional symbol reference operand to the `sim.proc.print` operation, representing an abstract output stream. The only hard requirement would be, that the backend has to provide _some_ way of discriminating the streams. Whether the simulation will then actually use (and manage) different IO resources, or just dump everything to `stdout` with a unique prefix is up to the lowering passes and the simulator environment. We could then add (discardable) attributes to the operation declaring the symbol, which provide a hint on the desired specifics of that stream. E.g., "this is an error log", "this should go to a file called xyz.log", etc.

sequencer:
I think we could add a new attribute to `printf` and related IR, which can be assigned and frontend. This attribute should have a default implementation like what we are currently assigning with stderr. User can provide enum, and these enums can be used in the lowering path, e.g. synthesis to different macro or goes to different XDMA channel in FPGA printf synthesis.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
sequencer: > Do we add comments for other chisel macros?

These macro are not exposed in Chisel, I think Chisel doesn't have a clear ABI for now.

But instead we should add these to firrtl spec, and chisel users can be point to there.

seldridge (commented):
This needs tests. Also, I agree that there probably isn't a good reason to add comments about what this is as it just bloats the IR.

Otherwise, this is good to proceed with.

Clo91eaf:
@seldridge: Sorry, I only had time today to fix the issue in the test. Could you please take a look for me?

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
seldridge: This is pretty consistent with how we currently handle other un-documented things. E.g., `PRINTF_COND` has the exact same language.

seldridge (approved):
LGTM

I ran this locally and inspected the output. This looks good. It's also consistent with the existing way that other things like this work. E.g., consider the following:

``` firrtl
FIRRTL version 4.1.0
circuit Foo:
  public module Foo:
    input clock: Clock
    input a: SInt<2>
    input b: SInt<2>
    output c: SInt<3>
    input cond: UInt<1>

    connect c, add(a, b)

    when cond:
      printf(clock, UInt<1>(1), "%d", a)
```

With this patch, the above compiles to:

``` verilog
// Generated by CIRCT firtool-1.99.2-8-g7e4889f56

// Users can define 'PRINTF_FD' to add a specified fd to prints.
`ifndef PRINTF_FD_
  `ifdef PRINTF_FD
    `define PRINTF_FD_ (`PRINTF_FD)
  `else  // PRINTF_FD
    `define PRINTF_FD_ 32'h80000002
  `endif // PRINTF_FD
`endif // not def PRINTF_FD_

`ifndef PRINTF_COND_
  `ifdef PRINTF_COND
  `else  // PRINTF_COND
    `define PRINTF_COND_ 1
  `endif // PRINTF_COND
module Foo(
  input        clock,
  input  [1:0] a,
               b,
  output [2:0] c,
  input        cond
);

  `ifndef SYNTHESIS
    always @(posedge clock) begin
      if ((`PRINTF_COND_) & cond)
        $fwrite(`PRINTF_FD_, "%d", $signed(a));
    end // always @(posedge)
  `endif // not def SYNTHESIS
  assign c = {a[1], a} + {b[1], b};
endmodule
```

PR #7997, opened 2024-12-16
[RTG][Elaboration] Do not internalize primitive values

maerhart:
* Refactor the pass to not internalize `ElaboratorValue` directly, but let it have a pointer to a storage which is internalized. This storage pointer can be used to directly store primitive values, thus avoiding internalization:
    *  `index` type values
    * `i1` type values
    *  MLIR Attributes by storing their opaque pointer (because they are already internalized by MLIR)
* Instead of the unique pointer trick, use a bump allocator in the internalization logic
* I tested 3 possible implementation for the sets that store the internalized values:
    * `DenseSet<StorageTy *>` with a custom MapInfo that does actual value comparison instead of comparing pointers: performs very well when all values internalized are unique (i.e., no equivalent object is already in the set, ever), but not very well when the object is equivalent to an already internalized one most of the time
    * `FoldingSet<StorageTy>`: performs worse than the above in the first case, but considerably better when the object is already in the set, because it avoids unnecessary malloc+free
    * The approach in this PR: combines the advantages of the above two, i.e., performs best in all cases, and is also the approach taken in MLIRs `StorageUniquer.cpp`
* Improves performance of the example in https://github.com/llvm/circt/pull/7986 from ~5.4 sec to ~0.78 sec (~6.9x speedup)


maerhart:
Temporarily converting to draft to evaluate the comments re `std::variant` in earlier PRs

darthscsi (approved):
LGTM

PR #7998, opened 2024-12-16
[RTG][Elaboration] Move ConstantLike check after TypeSwitch for better performance

maerhart:
The `hasTrait` check is quite expensive. Most operations in the IR are likely not constants, thus doing the TypeSwitch and Visitor dispatch first, should lead to better performance in most cases.

Reduces runtime of example in https://github.com/llvm/circt/pull/7986 but with 10x more iterations from ~6.8 sec to ~3.7 sec (1.84x speedup).
Increases runtime of an example containing a single `rtg.test` containing 10 mio `index.constant` (and nothing else) from 1.67 sec to 1.85 sec (10% slowdown).

### DIFF
--- a/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
+++ b/lib/Dialect/RTG/Transforms/ElaborationPass.cpp
@@ -696,6 +696,28 @@ public:
   }
 
   FailureOr<DeletionKind> visitExternalOp(Operation *op) {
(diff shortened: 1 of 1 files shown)


fabianschuiki (approved):
1.84x best case versus 0.9x worst case feels like a nice improvement! :+1:

youngar (approved):

PR #8004, opened 2024-12-17
[ESI] FIFO with ESI channels

teqdruid:
Adds an op which provides a FIFO with ESI channels (FIFO signaling semantics only for now) on both ends.


lib/Dialect/ESI/ESIOps.cpp
```diff
@@ -72,6 +72,22 @@ LogicalResult ChannelBufferOp::verify() {
...
+LogicalResult FIFOOp::verify() {
+  ChannelType inputType = getInput().getType();
+  if (inputType.getSignaling() != ChannelSignaling::FIFO)
+    return emitOpError("only supports FIFO signaling on input");
+  ChannelType outputType = getOutput().getType();
+  if (outputType.getSignaling() != ChannelSignaling::FIFO)
+    return emitOpError("only supports FIFO signaling on output");
+  if (outputType.getInner() != inputType.getInner())
+    return emitOpError("input and output types must match");
+  return success();
```
mortbopet, line 88: Missing verifiers for these

mortbopet (approved):
LGTM % comment.

lib/Dialect/ESI/ESIOps.cpp
teqdruid: You mean verifier tests?

mortbopet: Yep!

teqdruid: I'm only doing the inner type check test since the first two restrictions are being removed shortly.

PR #8005, opened 2024-12-18
[CD] Install nanobind when building wheels.

mikeurbach:
We recently started supporting nanobind as a dependency in upstream MLIR. This is installed in our CI images, but not the Python wheel CI environment. This adds that dependency to the package used for the Python wheel build.

### DIFF
--- a/lib/Bindings/Python/pyproject.toml
+++ b/lib/Bindings/Python/pyproject.toml
@@ -7,6 +7,7 @@ requires = [
     # MLIR build depends.
     "numpy",
     "pybind11>=2.11,<=2.12",
+    "nanobind==2.4.0",
     "PyYAML",
 ]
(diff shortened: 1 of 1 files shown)


teqdruid (approved):

mikeurbach:
Oops forgot to post it, but there was a successful run here: https://github.com/llvm/circt/actions/runs/12387120557

Thanks for the review @teqdruid.

teqdruid:
I'm probably gonna have to make the same changes to PyCDE when I do the next release, so I'm paying attention.

mikeurbach:
Ok sounds good. Sorry that I haven't also been updating PyCDE as I go, but there were other issues in my environment. I think at this point you can just grep for nanobind in this repo and circt/images to see where it's needed:

* CI environments used by CIRCT
* Wheel building environments

PR #8006, opened 2024-12-18
[VerifToSMT] Fix incorrect loop region result indexing

TaoBi22:
Fixes a bug introduced in #7878 where the wrong index is used to access the results of the loop region, often leading to a crash when the clock isn't the first input to the original module. Also includes a breaking example as a test - I've added it in a new file to avoid the different BMC circuit/loop/init functions from this and the other BMC test interleaving, can merge it into the existing test file if preferred and just interleave the Filecheck commands.

### DIFF
--- a/test/Conversion/VerifToSMT/bmc-clock-not-first.mlir
+++ b/test/Conversion/VerifToSMT/bmc-clock-not-first.mlir
@@ -1,0 +1,88 @@
+// RUN: circt-opt %s --convert-verif-to-smt --reconcile-unrealized-casts -allow-unregistered-dialect | FileCheck %s
+
+// CHECK-LABEL:  func.func @test_bmc_clock_not_first() -> i1 {
+// CHECK:    [[BMC:%.+]] = smt.solver
+// CHECK:      [[INIT:%.+]] = func.call @bmc_init()
+// CHECK:      smt.push 1
+// CHECK:      [[F0:%.+]] = smt.declare_fun : !smt.bv<32>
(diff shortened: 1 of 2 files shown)


test/Conversion/VerifToSMT/bmc-clock-not-first.mlir
```diff
@@ -0,0 +1,88 @@
...
+  circuit {
+  ^bb0(%arg0: i32, %clk: !seq.clock, %state0: i32):
+    %c-1_i32 = hw.constant -1 : i32
+    %0 = comb.add %arg0, %state0 : i32
+    // %state0 is the result of a seq.compreg taking %0 as input
+    %2 = comb.xor %state0, %c-1_i32 : i32
+    verif.yield %2, %0 : i32, i32
+  }
+  func.return %bmc : i1
+}
```
maerhart, line 88: Would it be possible to trim down this test a bit and only check the small part that is actually affected, i.e., just the following part, right?
```
// CHECK:        [[LOOP:%.+]] = func.call @bmc_loop([[ARG2]])
// CHECK:        [[BVPOSEDGE:%.+]] = smt.bv.and {{%.*}}, [[LOOP]]
```

maerhart (approved):
LGTM! Thanks for catching this!

test/Conversion/VerifToSMT/bmc-clock-not-first.mlir
TaoBi22: Yeah good point, I think it'll just affect that plus the yielding. Thanks, I'll trim it down!

PR #8007, opened 2024-12-18
[LowerToBMC] Topologically sort module body before inlining to BMC op

TaoBi22:
Allows circt-bmc to support modules that contain non-dominance but can be topologically sorted. Doesn't cover every edge case (i.e. when modules get involved and there may be cycles), but I've run into needing this a few times so it's probably useful to have upstream.

### DIFF
--- a/lib/Tools/circt-bmc/LowerToBMC.cpp
+++ b/lib/Tools/circt-bmc/LowerToBMC.cpp
@@ -61,7 +61,10 @@ void LowerToBMCPass::runOnOperation() {
     return signalPassFailure();
   }
 
-  sortTopologically(&hwModule.getBodyRegion().front());
+  if (!sortTopologically(&hwModule.getBodyRegion().front())) {
+    hwModule->emitError("could not resolve cycles in module");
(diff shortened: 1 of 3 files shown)


lib/Tools/circt-bmc/LowerToBMC.cpp
```diff
@@ -60,6 +61,8 @@ void LowerToBMCPass::runOnOperation() {
     return signalPassFailure();
   }
 
+  sortTopologically(&hwModule.getBodyRegion().front());
```
uenoku, line 64: nit: Can you check if they are topologically sorted? 
```suggestion
  if(!sortTopologically(&hwModule.getBodyRegion().front())) {
    moduleOp.emitError("...");
    return signalPassFailure();
   }
```

uenoku (approved):
LGTM

lib/Tools/circt-bmc/LowerToBMC.cpp
TaoBi22: good point, thanks!

PR #8008, opened 2024-12-18
[PyCDE] Update build flow

teqdruid:
Fix the PyCDE build flow. Added nanobind and fixed dll install on Windows.


teqdruid:
For some reason, the old dll install didn't work anymore. It didn't appear to be related to the nanobind change. I haven't built PyCDE in about two months, so the pipeline probably just bit rotted.

teqdruid:
https://github.com/llvm/circt/actions/runs/12402656601

mikeurbach (commented):
Looks good to me.

Issue #1624, opened 2021-08-22
[Comb] `or` of exclusive concats can be merged


lattner:
I'm seeing a bunch of variants of this pattern:

// x is 6 bits, y is 2 bits.
`wire [9:0] _tmp_122 = {3'h0, x} | {y, 7'h0};`

which could be simplified to:

`wire [9:0] _tmp_122 = {y, 1'h0, x};`

A similar one is:

`{ 5'0, bit}|{4'0, cond?2:0}` ==> `{4'h0, cond ? 1 : 0, bit}` ==> `{4'h0, cond, bit}`

lattner:
Random thought - the second pattern can be handled by canonicalizing `cond?2:0` -> `concat(cond, 0)` which would then cause it to compose into the existing transformation.

youngar:
https://github.com/llvm/circt/pull/1701 is causing performance issues and I am going to revert it until the problems can be addressed.

mikeurbach:
Funny enough we once hit a performance cliff in verilator for a similar situation https://github.com/verilator/verilator/issues/3405. I think the fix there ended up being some kind of "optimization fuel" like approach where it just stopped if the expression was large... That could maybe work, but we should probably come up with something that's both general and performant for ourselves.

PR #8014, opened 2024-12-23
[Comb] delete slow canonicalizer

youngar:
This canonicalizer does not have a functional issue, but it has problematic performance.  This change removes it until it can be fixed properly.


youngar:
I'll reopen the original ticket: https://github.com/llvm/circt/issues/1624

PR #8016, opened 2024-12-24
[CombToAIG] Lower comb.icmp

uenoku:
This commit adds a pattern for icmp. Lowering is straight-forward but fairly complicated. LEC is verified. Test is modified to run CSE for simplifying lit testing.


fabianschuiki (approved):
Very neat :sunglasses:

PR #8017, opened 2024-12-26
[ImportVerilog] add forloop statement support

chenbo-again:
Add forloop statement support, due to lack of dyn array support, currently static-length array is supported.

### DIFF
--- a/lib/Conversion/ImportVerilog/Statements.cpp
+++ b/lib/Conversion/ImportVerilog/Statements.cpp
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "ImportVerilogInternals.h"
+#include "mlir/IR/Diagnostics.h"
 #include "slang/ast/SystemSubroutine.h"
 #include "llvm/ADT/ScopeExit.h"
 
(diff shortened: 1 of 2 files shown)


hailongSun2000 (commented):
Maybe you can refer to the implementation of `Handle repeat loops` by @fabianschuiki. We can pass the cond to the BB1 when we enter.

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "ImportVerilogInternals.h"
+#include "mlir/IR/Diagnostics.h"
```
hailongSun2000, line 10: nit: Don't include extra unnecessary headers.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -494,6 +494,66 @@ function void ForeverLoopStatements(bit x, bit y);
...
+// CHECK: moore.blocking_assign %[[J]], %[[ADD1]] : i32
+// CHECK: cf.br ^[[BB3]]
+// CHECK: ^[[BB8]]:
+// CHECK: cf.br ^[[BB9:.*]]
+// CHECK: ^[[BB9]]:
+// CHECK: %[[I_VAL2:.*]] = moore.read %[[I]] : <i32>
+// CHECK: %[[C1_2:.*]] = moore.constant 1 : i32
+// CHECK: %[[ADD2:.*]] = moore.add %[[I_VAL2]], %[[C1_2]] : i32
+// CHECK: moore.blocking_assign %[[I]], %[[ADD2]] : i32
+// CHECK: cf.br ^[[BB1]]
```
hailongSun2000, line 550: Referring to `repeat loops`, we can avoid creating `blocking_assign`.

hailongSun2000 (commented):
Please add a few annotations. It's important for others. And thanks for your work on this :smile:.

test/Conversion/ImportVerilog/basic.sv
chenbo-again: The repeat loop loops for a certain number of times, while foreach is more like a for loop. It binds the number of times of the loop to a variable. Perhaps we have to perform block_assignment repeatedly.

```systemverilog
foreach (array[i])
 	$display ("array[%0d] = %0d", i, array[i]);
```

I can't find a way to achieve that if no assignment performed.

hailongSun2000: Oh~ I learned it. We cannot avoid using the iterator `i`.

hailongSun2000 (approved):
LGTM! Please wait for @fabianschuiki to review it again.

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -309,6 +396,14 @@ struct StmtVisitor {
     return success();
   }
 
+  LogicalResult visit(const slang::ast::ForeachLoopStatement &stmt) {
+    for (uint32_t level = 0; level < stmt.loopDims.size(); level++) {
+      if (stmt.loopDims[level].loopVar)
+        return recursiveForeach(stmt, level);
+    }
+    return failure();
```
fabianschuiki, line 404: It might make sense to return `success()` here since Slang probably already checks that there's at least one `loopVar`, or if it doesn't it might still be correct to simply not do anything for a loop where no dimensions are actually looped over. If you want to return `failure()`, it would be good to emit an error indicating to the user what happened. Otherwise the compiler will exit with a non-zero exit code, but potentially without explanation to the user, which would be bad.

fabianschuiki (approved):
LGTM, with a small comment on error reporting. Thanks a lot for working on this! :partying_face:

fabianschuiki:
Looks great, thanks for the fix! Feel free to land this whenever you're happy with it 😃. Or let me know if you want me to do it.

PR #8018, opened 2024-12-27
Add the ICE40 dialect.

darthscsi:
The ICE40 dialect is a synthesis target representing the Lattice ICE40 family of fpgas.


teqdruid:
Hobby or SiFive work?

darthscsi:
> Hobby or SiFive work?

Hobby.  Want to push things a bit and build on @uenoku 's synthesis work (also hobby, but hopefully I can fund it).

lib/Dialect/ICE40/ICE40Ops.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "mlir/IR/ImplicitLocOpBuilder.h"
+
+using namespace circt;
+using namespace ice40;
+
+//===----------------------------------------------------------------------===//
+// SBLut4Op
+//===----------------------------------------------------------------------===//
+
+LogicalResult SBLut4Op::verify() { return success(); }
```
TaoBi22, line 24: This is really exciting! One nit - is it worth verifying that `lut_init` has 16 elements? Or is this just a placeholder verifier

KelvinChung2000:
Would it be better to do it as a generic FPGA dialect?

teqdruid:
> Would it be better to do it as a generic FPGA dialect?

There's no way to capture all the details of every FPGA in a set of common ops. There is something to be said for one dialect containing ops for each FPGA family, but it's easier to start separately then merge dialects if need be.

fabianschuiki (approved):
This is really cool :sunglasses: Blinkenlights wen?

include/circt/Dialect/ICE40/ICE40Ops.td
```diff
@@ -0,0 +1,74 @@
...
+  }];
+}
+
+def SBLut4CarryOp : ICE40Op<"sb_lut4_carry", []> {
+  let summary = "LUT which also utilizes the carry logic";
+  let description = [{
+      carry_out(i0, i1, ci) = 1 iff popcount(concat(i0,i1,ci)) >= 2.
+      result(i0, i1, i2, i3) = lut_init[concat(i3,i2,i1,i0)].
+    There are constraints on the use, as i0 and i1 have to be the i0 and i1 input of a single lut.
+    This op doesn't check this.  There is a combined op that handles that.
```
teqdruid, line 61: Isn't this the combined op?

lib/Dialect/ICE40/ICE40Ops.cpp
teqdruid: Agreed it should check.

lib/Dialect/ICE40/ICE40Ops.cpp
```diff
@@ -0,0 +1,44 @@
...
+// SBCarryOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult SBCarryOp::verify() { return success(); }
+
+//===----------------------------------------------------------------------===//
+// SBLut4CarryOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult SBLut4CarryOp::verify() { return success(); }
```
teqdruid, line 36: Same comment re checking here.

test/Dialect/ICE40/basic.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: circt-opt %s | FileCheck %s
```
teqdruid, line 1: When you add the verifier, create an error test.

teqdruid (approved):
Exciting!

PR #8023, opened 2025-01-01
[Comb] Don't try to canonicalize muxes indefinitely

maerhart:
Adds a few checks to avoid infinite canonicalizer runtime in some corner cases.


uenoku (approved):
LGTM!

PR #8026, opened 2025-01-03
[FIRRTL] Add View Intrinsic

seldridge:
Add a preliminary ViewIntrinsic that is an intended replacement for the Grand Central View AugmentedBundleType Annotation.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
@@ -239,8 +239,8 @@ def ViewIntrinsicOp : FIRRTLOp<"view", []> {
     debugging in a waveform.  This is _not_ a true SystemVerilog Interface, it
     is only lowered to one.
   }];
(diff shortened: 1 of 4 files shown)


dtzSiFive:
I tweaked the operation a little, LGTM re:op addition.

Needs entry in FIRRTLVisitor (Statement?).

Verification might be reasonable to add to the operation, depending on how big these get (mostly sanity checking the attribute and that operand count matches, I think)?

While these have (and introduce a name, as a sort of declaration) they aren't expected to really act that way, I think is where things landed.
(can't target with annotations, so on)

seldridge (commented):
LGTM

A verifier (in a future PR) that checks the shape of the augmented bundle type and the number of variadic arguments would be nice.

dtzSiFive:
> A verifier (in a future PR) that checks the shape of the augmented bundle type and the number of variadic arguments would be nice.

Yes!  And maybe tweak the attributes to more directly contain what's needed (vs DictionaryAttr), reducing need to inspect their contents as part of the operation's verification other than minor details (and asking/counting how many leaves)...?

I'll tweak the operation to check all operands are ground types before merging 👍 .

PR #8028, opened 2025-01-03
[HWToBTOR2] Fix incorrect le/ge predicate name emission

TaoBi22:
Found a couple of missing special cases where the CIRCT and BTOR2 names for predicates differ (we abbreviate "than or equal" to just "e", they do it to "te") so invalid BTOR2 is produced - breaking examples in the added test checks.


dobios:
Great catch as usual! It's really great that you're stress testing this as much as you are!!

dobios (approved):

PR #8033, opened 2025-01-06
[MemoryBanking] Support multi-dimension memory banking

jiahanxie353:
This patch supports multi-dimensional memory banking in the memory banking pass.

### DIFF
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -376,9 +376,10 @@ void MemoryBankingPass::runOnOperation() {
     DenseSet<Value> memrefsInPar = collectMemRefs(parOp);
 
     for (auto memrefVal : memrefsInPar) {
-      if (!memoryToBanks.contains(memrefVal))
-        memoryToBanks[memrefVal] =
-            createBanks(memrefVal, bankingFactor, bankingDimension);
+      auto [it, inserted] =
+          memoryToBanks.insert(std::make_pair(memrefVal, SmallVector<Value>{}));
+      if (inserted)
+        it->second = createBanks(memrefVal, bankingFactor, bankingDimension);
     }
   });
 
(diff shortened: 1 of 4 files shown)


lib/Transforms/MemoryBanking.cpp
```diff
@@ -358,17 +377,21 @@ void MemoryBankingPass::runOnOperation() {
   getOperation().walk([&](mlir::affine::AffineParallelOp parOp) {
     DenseSet<Value> memrefsInPar = collectMemRefs(parOp);
 
-    for (auto memrefVal : memrefsInPar)
-      memoryToBanks[memrefVal] = createBanks(memrefVal, bankingFactor);
+    for (auto memrefVal : memrefsInPar) {
+      if (!memoryToBanks.contains(memrefVal))
+        memoryToBanks[memrefVal] =
```
cgyurgyik, line 382: I think 'try_emplace' is a better fit here, if this container supports it. This avoids the double lookup.

cgyurgyik (approved):
LGTM with one small comment

lib/Transforms/MemoryBanking.cpp
jiahanxie353: Agree, I tried it out but the problem with it is that arguments will be evaluated so do something like `try_emplace(memrefVal, createBanks(...))` will invoke `createBanks` function, which changes the block arguments in-place: https://github.com/llvm/circt/blob/8fa4aad9a3e67158c683bdeb5dc090a25ab2a2ed/lib/Transforms/MemoryBanking.cpp#L100

cgyurgyik: Yeah I had more faith in `try_emplace` than I should have. Resolving.

jiahanxie353: seems like `DenseMap`s don't have `emplace`, but it has `insert`: https://llvm.org/doxygen/classllvm_1_1DenseMapBase.html#adb33f3ede2f5bfc9b3ee658aaf648492
so we can do something like:
```c++
auto [it, inserted] = memoryToBanks.insert(std::make_pair(memrefVal, SmallVector<Value>{}));
if (inserted) {
    it->second = createBanks(memrefVal, bankingFactor, bankingDimension);
}
```
it requires making a `std::pair` though

jiahanxie353:
Merging once passed the CI

PR #8034, opened 2025-01-06
[RTG] Add context resource attribute interface

maerhart:
Provides a way to specify context resources with an attribute that works well in combination with the ContextResourceTypeInterface. Currently, the `cpu_decl` operation has an index attr but returns a ContextResourceTypeInterface type, which makes folding difficult (and thus declaring it ConstantLike).
Further down the line, we could consider removing the `ContextResourceOpInterface` in favour of a single op in the RTG dialect.

### DIFF
--- a/include/circt-c/Dialect/RTGTest.h
+++ b/include/circt-c/Dialect/RTGTest.h
@@ -31,6 +31,20 @@ MLIR_CAPI_EXPORTED bool rtgtestTypeIsACPU(MlirType type);
 /// Creates an RTGTest CPU type in the context.
 MLIR_CAPI_EXPORTED MlirType rtgtestCPUTypeGet(MlirContext ctxt);
 
+//===----------------------------------------------------------------------===//
+// Attribute API.
+//===----------------------------------------------------------------------===//
+
+/// If the type is an RTGTest CPUAttr.
+MLIR_CAPI_EXPORTED bool rtgtestAttrIsACPU(MlirAttribute attr);
+
+/// Creates an RTGTest CPU attribute in the context.
+MLIR_CAPI_EXPORTED MlirAttribute rtgtestCPUAttrGet(MlirContext ctxt,
+                                                   unsigned id);
+
+/// Returns the core ID represented by the CPU attribute.
+MLIR_CAPI_EXPORTED unsigned rtgtestCPUAttrGetId(MlirAttribute attr);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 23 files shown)


lib/Bindings/Python/RTGTestModule.cpp
```diff
@@ -31,4 +31,14 @@ void circt::python::populateDialectRTGTestSubmodule(py::module &m) {
...
+
+  mlir_attribute_subclass(m, "CPUAttr", rtgtestAttrIsACPU)
+      .def_classmethod(
+          "get",
+          [](py::object cls, unsigned id, MlirContext ctxt) {
+            return cls(rtgtestCPUAttrGet(ctxt, id));
+          },
+          py::arg("self"), py::arg("id"), py::arg("ctxt") = nullptr)
+      .def_property_readonly(
+          "id", [](MlirAttribute self) { return rtgtestCPUAttrGetId(self); });
```
maerhart, line 43: Needs a test

include/circt-c/Dialect/RTGTest.h
```diff
@@ -31,6 +31,20 @@ MLIR_CAPI_EXPORTED bool rtgtestTypeIsACPU(MlirType type);
...
+
+/// If the type is an RTGTest CPUAttr.
+MLIR_CAPI_EXPORTED bool rtgtestAttrIsACPU(MlirAttribute attr);
+
+/// Creates an RTGTest CPU attribute in the context.
+MLIR_CAPI_EXPORTED MlirAttribute rtgtestCPUAttrGet(MlirContext ctxt,
+                                                   unsigned id);
+
+/// Returns the core ID represented by the CPU attribute.
+MLIR_CAPI_EXPORTED unsigned rtgtestCPUAttrGetId(MlirAttribute attr);
```
maerhart, line 46: Needs tests

lib/Dialect/RTG/IR/CMakeLists.txt
```diff
@@ -15,6 +16,7 @@ add_circt_dialect_library(CIRCTRTGDialect
   CIRCTRTGOpInterfacesIncGen
   CIRCTRTGISAAssemblyOpInterfacesIncGen
   CIRCTRTGTypeInterfacesIncGen
+  CIRCTRTGAttrInterfacesIncGen
```
uenoku, line 19: Super nit: alpha-sort

uenoku (approved):
LGTM!

PR #8037, opened 2025-01-07
Use MLIR python detection environment

cbalint13:
This PR make CIRCT to relay on the very MLIR cmake python detection macros.

It now [properly picks](https://github.com/llvm/llvm-project/blob/ea14bdb0356cdda727ac032470f6a0a2102d1281/mlir/cmake/modules/MLIRDetectPythonEnv.cmake#L4-L48) up ```python``` components including ```pybind11``` and the recently introduced ```nanobind```.

Here is how this [is used upstream](https://github.com/llvm/llvm-project/blob/3b19e787fc5da27dfcc9ac6552b06a763f12ea03/mlir/examples/standalone/CMakeLists.txt#L35) .

---

Here is the cmake process output:

```
{...}
-- CIRCT Python bindings are enabled.
-- Found Python3: /usr/bin/python3.13 (found version "3.13.1") found components: Interpreter Development.Module
-- Found Python: /usr/bin/python3.13 (found version "3.13.1") found components: Interpreter Development.Module
-- Found python include dirs: /usr/include/python3.13
-- Found python libraries: 
-- Found numpy v: 
-- Checking for pybind11 in python path...
-- found (/usr/share/cmake/pybind11)
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- Found pybind11: /usr/include (found version "2.13.6")
-- Found pybind11 v2.13.6: /usr/include
-- Python prefix = '', suffix = '', extension = '.cpython-313-x86_64-linux-gnu.so
-- Checking for nanobind in python path...
-- found (/usr/lib/python3.13/site-packages/nanobind/cmake)
-- Found nanobind v2.4.0: 
-- Python prefix = '', suffix = '', extension = '.cpython-313-x86_64-linux-gnu.so
{...}
```

Thank you !

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -514,11 +514,8 @@ option(CIRCT_BINDINGS_PYTHON_ENABLED "Enables CIRCT Python bindings." OFF)
 if(CIRCT_BINDINGS_PYTHON_ENABLED)
   message(STATUS "CIRCT Python bindings are enabled.")
   set(MLIR_DISABLE_CONFIGURE_PYTHON_DEV_PACKAGES 0)
-  mlir_detect_pybind11_install()
-  # Prime the search like mlir_configure_python_dev_modules
-  find_package(Python3 3.8 COMPONENTS Interpreter Development)
-  find_package(Python3 3.8 COMPONENTS Interpreter Development.Module)
-  find_package(pybind11 2.10 CONFIG REQUIRED)
+  include(MLIRDetectPythonEnv)
+  mlir_configure_python_dev_packages()
 else()
   message(STATUS "CIRCT Python bindings are disabled.")
   # Lookup python either way as some integration tests use python without the


mikeurbach:
I think this may have broken the ability to build wheels on Linux. I actually updated this to _not_ use the upstream helpers based on the discussion here: https://discourse.llvm.org/t/psa-python-binding-dependencies-changing/83376.

That did break the ability to build wheels on MacOS, and I was trying to test some fixes for that, when I noticed the Linux wheel build is now broken: https://github.com/llvm/circt/actions/runs/12659976325/job/35280207153#step:6:7841

It's interesting that the failure matches the failures we've been seeing on MacOS... it seems like we are getting closer to understanding what's going wrong.

I'm not positive this change was the culprit, so I'll open a revert PR and kick off a wheel build on that PR to see if it succeeds.

cbalint13:
> I think this may have broken the ability to build wheels on Linux. I actually updated this to _not_ use the upstream helpers based on the discussion here: https://discourse.llvm.org/t/psa-python-binding-dependencies-changing/83376.
> 

Original issue (on my side, git-tag 1.99.2) was:
```
-- CIRCT Python bindings are enabled.
CMake Error at CMakeLists.txt:517 (mlir_detect_pybind11_install):
  Unknown CMake command "mlir_detect_pybind11_install".
```

* Somehow the ```mlir_detect_pybind11_install()``` (available within ```include(MLIRDetectPythonEnv)```) was not picked up.
* Also the new ```nanobind``` dependency was not picked up at all.

I noticed the proper setup of python environment [here](https://github.com/llvm/llvm-project/blob/3b19e787fc5da27dfcc9ac6552b06a763f12ea03/mlir/examples/standalone/CMakeLists.txt#L35) for booth external (non mono-repo) and internal projects.

> That did break the ability to build wheels on MacOS, and I was trying to test some fixes for that, when I noticed the Linux wheel build is now broken: https://github.com/llvm/circt/actions/runs/12659976325/job/35280207153#step:6:7841


* I did't saw the failure on MacOS, CI was in green.

> It's interesting that the failure matches the failures we've been seeing on MacOS... it seems like we are getting closer to understanding what's going wrong.
> 
> I'm not positive this change was the culprit, so I'll open a revert PR and kick off a wheel build on that PR to see if it succeeds.

mikeurbach:
> Somehow the mlir_detect_pybind11_install() (available within include(MLIRDetectPythonEnv)) was not picked up.
> Also the new nanobind dependency was not picked up at all.

Hmm that is interesting, this is covered in the pre-merge CI. I think we'd need to know more about your environment to understand why it didn't work. I very much want to use the same setup example as the standalone project... if we can make it work for building wheels.

> I did't saw the failure on MacOS, CI was in green.

Yeah, unfortunately the issue only popped up in the Upload Wheels job, which is not tested before merge. It runs weekly, or on release. This was failing on MacOS after my original changes in CMakeLists.txt, and started failing on Linux as well after this PR.

PR #8038, opened 2025-01-07
Revert "Use MLIR python detection environment (#8037)"

mikeurbach:
This reverts commit ab5dc6c797583e2d292aaa8366531177de8d1bd4.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -514,8 +514,11 @@ option(CIRCT_BINDINGS_PYTHON_ENABLED "Enables CIRCT Python bindings." OFF)
 if(CIRCT_BINDINGS_PYTHON_ENABLED)
   message(STATUS "CIRCT Python bindings are enabled.")
   set(MLIR_DISABLE_CONFIGURE_PYTHON_DEV_PACKAGES 0)
-  include(MLIRDetectPythonEnv)
-  mlir_configure_python_dev_packages()
+  mlir_detect_pybind11_install()
+  # Prime the search like mlir_configure_python_dev_modules
+  find_package(Python3 3.8 COMPONENTS Interpreter Development)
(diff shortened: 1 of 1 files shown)


mikeurbach:
Wheel builds are running here: https://github.com/llvm/circt/actions/runs/12661341210

Note that only Linux are expected to actually pass.

mikeurbach:
I'm also looking into actually solving the root problem here, which is now affecting Linux and MacOS on main. I'd much rather keep the upstream helpers if we can.

mikeurbach:
@cbalint13 @teqdruid do you guys mind if I merge this for now, since the wheel building succeeds. Was there a problem in the configuration that required https://github.com/llvm/circt/pull/8037? It works for me locally and in the cibuildwheel environments.

teqdruid:
I do not mind. Sorry to break stuff.

cbalint13:
> @cbalint13 @teqdruid do you guys mind if I merge this for now, since the wheel building succeeds. Was there a problem in the configuration that required #8037? It works for me locally and in the cibuildwheel environments.

It is fine, sorry to break stuff there, but I expected to see that failure on CI at very PR time.
I try to look at the new issue too, address a fix altogether, if there will be a PR on this will request review @mikeurbach .

mikeurbach:
Alright, thanks folks, I will merge this for now.

@cbalint13 if you want to try re-landing this, you can manually kick off this job on your PR to test the fix still works with the wheel building: https://github.com/llvm/circt/actions/workflows/uploadWheels.yml

It should at least pass on Linux, since that should be the case after this revert PR merges. Since the failures I saw on Linux had the same signature as the already failing MacOS builds, it would be great if we could fix both at the same time. I'm still not sure why that is happening.

PR #8039, opened 2025-01-08
[MemoryBanking] Add a new field to keep track values to be erased

jiahanxie353:
This patch adds a new `oldMemRefVals` field to keep track of the original `memref` values that should be erased, as opposed to the old approach that derives `oldMemRefVals` simply from `memoryToBanks` map: https://github.com/llvm/circt/blob/acc507542bdcd91ca45a8b6fa8bd65e6629f513d/lib/Transforms/MemoryBanking.cpp#L382

The issue with the original approach is that some keys in the `memoryToBanks` map might become dangling references because they are replaced: https://github.com/llvm/circt/blob/acc507542bdcd91ca45a8b6fa8bd65e6629f513d/lib/Transforms/MemoryBanking.cpp#L182

We add `loadOp.memref()` to `oldMemRefVals` only when the value is a `BlockArgument` because only the `BlockArgument` will remain in the `cleanUpOldMemRefs` function; and the values that are not `BlockArgument` will all gets replaced and naturally no longer exists, thus not requiring any additional clean-ups.

### DIFF
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -50,8 +50,8 @@ private:
   // map from original memory definition to newly allocated banks
   DenseMap<Value, SmallVector<Value>> memoryToBanks;
   DenseSet<Operation *> opsToErase;
-  // old/original memref values that need to be cleaned up after all rewrite
-  // patterns
+  // Track memory references that need to be cleaned up after memory banking is
+  // complete.
   DenseSet<Value> oldMemRefVals;
 };
 } // namespace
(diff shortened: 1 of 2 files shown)


lib/Transforms/MemoryBanking.cpp
```diff
@@ -179,6 +182,8 @@ struct BankAffineLoadPattern
     auto defaultValue = rewriter.create<arith::ConstantOp>(loc, zeroAttr);
     rewriter.create<scf::YieldOp>(loc, defaultValue.getResult());
 
+    if (isa<BlockArgument>(loadOp.getMemref()))
+      oldMemRefVals.insert(loadOp.getMemref());
```
cgyurgyik, line 186:
```suggestion
    if (Value memRef = loadOp.getMemref(); isa<BlockArgument>(memRef))
      oldMemRefVals.insert(memRef);
```

lib/Transforms/MemoryBanking.cpp
```diff
@@ -179,6 +182,8 @@ struct BankAffineLoadPattern
     auto defaultValue = rewriter.create<arith::ConstantOp>(loc, zeroAttr);
     rewriter.create<scf::YieldOp>(loc, defaultValue.getResult());
 
+    if (isa<BlockArgument>(loadOp.getMemref()))
```
cgyurgyik, line 185: > We add loadOp.memref() to oldMemRefVals only when the value is a BlockArgument because only the BlockArgument will remain in the cleanUpOldMemRefs function; and the values that are not BlockArgument will all gets replaced and naturally no longer exists, thus not requiring any additional clean-ups.

It probably would be good to add this as a (bit more concise) comment here. It isn't clear otherwise to someone reading the code.

lib/Transforms/MemoryBanking.cpp
jiahanxie353: sounds good, resolving

jiahanxie353:
> I've added some suggested changes, do with them as you please. LGTM

committing them, thanks for the review!

Scutua (approved):

PR #8041, opened 2025-01-08
[SMT] Add int2bv operation

TaoBi22:
Adds the `smt.int2bv` operation which converts an `smt.int` to an `smt.bv`. This adds a lowering to Z3 LLVM, but no lowering to SMTLIB as it's not part of the SMTLIB standard (int/bitvector conversions in SMTLIB are not super standardised[^1] in SMT solvers, so @maerhart and I decided this was probably the best way to provide support for now).

[^1]: See [this Google Group discussion](https://groups.google.com/g/smt-lib/c/-GJG1Pq61hI) for more info


include/circt/Dialect/SMT/SMTIntOps.td
```diff
@@ -119,4 +119,13 @@ def IntCmpOp : SMTIntOp<"cmp", [Pure]> {
   let assemblyFormat = "$pred $lhs `,` $rhs attr-dict";
 }
 
+def Int2BVOp : SMTOp<"int2bv", [Pure]> {
+  let summary = "Convert an integer to an n-wide bitvector";
+  let arguments = (ins IntType:$input, ConfinedAttr<I32Attr, [IntNonNegative]>:$width);
```
maerhart, line 124: Why do we need a `width` attribute? The BV type already contains the width.

include/circt/Dialect/SMT/SMTIntOps.td
```diff
@@ -119,4 +119,13 @@ def IntCmpOp : SMTIntOp<"cmp", [Pure]> {
   let assemblyFormat = "$pred $lhs `,` $rhs attr-dict";
 }
 
+def Int2BVOp : SMTOp<"int2bv", [Pure]> {
+  let summary = "Convert an integer to an n-wide bitvector";
```
maerhart, line 123: Can you add a small description describing what happens if the integer does not fix into the BV of the given width?
Are negative integers converted to two-complement?
Since all these conversions are not very standardized, it'd be nice to describe the corner cases in a bit more detail. Since you lower it 1-1 to Z3, I guess that'd just be a copy-paste of Z3's spec?

integration_test/Dialect/SMT/basic.mlir
```diff
@@ -194,6 +194,17 @@ func.func @entry() {
...
+  // CHECK: Res: -1
+  smt.solver () : () -> () {
+    %c4 = smt.int.constant 4
+    %c4_bv16 = smt.bv.constant #smt.bv<4> : !smt.bv<16>
+    %int2bv = smt.int2bv %c4 width 16 : !smt.bv<16>
+    %eq = smt.distinct %c4_bv16, %int2bv : !smt.bv<16>
+    func.call @check(%eq) : (!smt.bool) -> ()
+    smt.yield
+  }
+
```
maerhart, line 207: Nice test!

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -635,6 +635,10 @@ static LogicalResult emit(SolverOp solver, const SMTEmissionOptions &options,
       return op->emitError()
              << "solver must not contain any non-SMT operations";
 
+    if (isa<smt::Int2BVOp>(op))
+      return op->emitError(
+          "int2bv operations are not supported for SMTLIB emission");
```
maerhart, line 640: Maybe move this error message to `visitUnhandledSMTOp` in ExpressionVisitor (I don't quite remember why we just return success there, returning failure would make more sense (and explicitly handle ops for which we just want to return success, if there are any, might also just be a copy-paste mistake from StatementVisitor).

include/circt/Dialect/SMT/SMTIntOps.td
TaoBi22: Good point :woman_facepalming: thanks!

include/circt/Dialect/SMT/SMTIntOps.td
TaoBi22: Will do!

PR #8042, opened 2025-01-08
[SMT] Parse SMT bitvector width as signed

TaoBi22:
Fixes an issue where SMT bitvector widths being parsed as unsigned meant that e.g. !smt.bv<-1> would wrap around and be interpreted as a 18446744073709551615-wide bitvector. The downside of this fix is that the error messages for bitvectors that are too wide become a little less clear (see the updated test) - does this still look like an ok fix to you @maerhart? Let me know if I should also drop the code that's emitting that error in the RepeatOp verifier, don't think it can be triggered after this but I may have missed a case.

### DIFF
--- a/test/Dialect/SMT/bitvector-errors.mlir
+++ b/test/Dialect/SMT/bitvector-errors.mlir
@@ -7,6 +7,13 @@ func.func @at_least_size_one(%arg0: !smt.bv<0>) {
 
 // -----
 
+// expected-error @below {{bit-vector must have at least a width of one}}
+func.func @at_least_size_one(%arg0: !smt.bv<-1>) {
+  return
+}
+
+// -----
+
 func.func @attr_type_and_return_type_match() {
   // expected-error @below {{inferred type(s) '!smt.bv<1>' are incompatible with return type(s) of operation '!smt.bv<32>'}}
   // expected-error @below {{failed to infer returned types}}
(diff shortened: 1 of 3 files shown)


test/Dialect/SMT/bitvector-errors.mlir
```diff
@@ -89,8 +96,8 @@ func.func @repeat_count_too_large(%arg0: !smt.bv<32>) {
 
 // -----
 
-func.func @repeat_result_type_bitwidth_too_large(%arg0: !smt.bv<18446744073709551612>) {
-  // expected-error @below {{result bit-width (provided integer times bit-width of the input type) must fit into 64 bits}}
-  smt.bv.repeat 2 times %arg0 : !smt.bv<18446744073709551612>
```
maerhart, line 94: In the repeat parser, 64 is hardcoded as the width of the unsigned integer used. We'd need to adjust this and keep this error test but with a smaller number.

maerhart (approved):
It might be cleaner to write a custom parser that avoids the implicit conversion (basically parsing an APInt and doing the necessary checks before converting it to uint64_t).
When creating this type in C++ you probably give it an unsigned int, so the implicit conversion to a signed int could be a bit annoying there (but also only in corner cases).
But this approach is also fine for me.

I assume something like https://github.com/llvm/circt/issues/8062 can still happen after this change (just the bitwidth to trigger it would be smaller)?

test/Dialect/SMT/bitvector-errors.mlir
TaoBi22: Cool, I've updated that now.

TaoBi22:
> It might be cleaner to write a custom parser that avoids the implicit conversion (basically parsing an APInt and doing the necessary checks before converting it to uint64_t).

Agreed that would be nicer - I'll merge this for now (once CI has passed) so we have a fix in but I'll try and come back and fix it up properly at a later date.

> I assume something like #8062 can still happen after this change (just the bitwidth to trigger it would be smaller)?

Yes, after this fix circt-opt spits out:

```
hw.module @ImplicitOverflow() {
  %0 = smt.bv.constant #smt.bv<1> : !smt.bv<9223372036854775803>
}
```

as:

```
module {
  hw.module @ImplicitOverflow() {
    %c1_bv4294967291 = smt.bv.constant #smt.bv<1> : !smt.bv<4294967291>
    hw.output
  }
}
```

PR #8047, opened 2025-01-09
[MemoryBanking] Support memory banking for`GetGlobalOp`

jiahanxie353:
This patch supports memory banking for multi-dimensional `GetGlobalOp`s and restructure the data in their corresponding `GlobalOp`s by the banking scheme (such as the banking factor and the dimension).

### DIFF
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -25,6 +25,7 @@
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/FormatVariadic.h"
+#include <numeric>
 
 namespace circt {
 #define GEN_PASS_DEF_MEMORYBANKING
@@ -113,14 +114,8 @@ SmallVector<SmallVector<Attribute>> sliceSubBlock(ArrayRef<Attribute> allAttrs,
                                                   ArrayRef<int64_t> memShape,
                                                   unsigned bankingDimension,
                                                   unsigned bankingFactor) {
-  auto totalSize = [](auto &&shape) {
-    size_t prod = 1;
-    for (auto s : shape)
-      prod *= s;
-    return prod;
-  };
-
-  size_t numElements = totalSize(memShape);
+  size_t numElements = std::reduce(memShape.begin(), memShape.end(), 1,
+                                   std::multiplies<size_t>());
   // `bankingFactor` number of flattened attributes that store the information
   // in the original globalOp
   SmallVector<SmallVector<Attribute>> subBlocks;
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -89,18 +89,20 @@ MemRefType computeBankedMemRefType(MemRefType originalType,
   return newMemRefType;
 }
 
-// Updates n-dimensional index array `ndIndex` in-place by decoding the flat
-// index `linIndex` given the shape of the array. Assume row-major order.
-void decodeIndex(int64_t linIndex, ArrayRef<int64_t> shape,
-                 SmallVector<int64_t> &ndIndex) {
+// Decodes the flat index `linIndex` into an n-dimensional index based on the
+// given `shape` of the array in row-major order. Returns an array to represent
+// the n-dimensional indices.
+SmallVector<int64_t> decodeIndex(int64_t linIndex, ArrayRef<int64_t> shape) {
   const unsigned rank = shape.size();
-  ndIndex.resize(rank);
+  SmallVector<int64_t> ndIndex(rank, 0);
 
-  // compute from last dimension to first because we assume row-major.
-  for (int d = rank - 1; d >= 0; --d) {
+  // Compute from last dimension to first because we assume row-major.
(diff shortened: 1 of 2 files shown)


test/Transforms/memory_banking_multi_dim.mlir
```diff
@@ -1,4 +1,5 @@
 // RUN: circt-opt %s -split-input-file -memory-banking="banking-factor=2 dimension=1" | FileCheck %s --check-prefix RANK2-BANKDIM1
+// RUN: circt-opt %s -split-input-file -memory-banking="banking-factor=2" | FileCheck %s --check-prefix GETGLOBAL
```
cgyurgyik, line 2: You may have handled this in an earlier PR, but what about corner cases, e.g., `banking-factor=0`, `dim_size % banking-factor != 0`, etc.?

lib/Transforms/MemoryBanking.cpp
```diff
@@ -125,6 +171,76 @@ SmallVector<Value, 4> createBanks(Value originalMem, uint64_t bankingFactor,
             banks.push_back(bankAllocaOp);
           }
         })
+        .Case<memref::GetGlobalOp>([&](memref::GetGlobalOp getGlobalOp) {
```
cgyurgyik, line 174: How difficult would it be to place all this code in a separate function and just call it here? It is quite large, but unsure if that would require too many arguments.

cgyurgyik (commented):
Nice job! Mostly style nits and a few smaller comments.

test/Transforms/memory_banking_multi_dim.mlir
jiahanxie353: Yes, in these cases, it will early signal failure: https://github.com/llvm/circt/blob/4a062a2a1cda5052f61c700aeb204ac4ba7d2f55/lib/Transforms/MemoryBanking.cpp#L381
or raise an assertion error: https://github.com/llvm/circt/blob/4a062a2a1cda5052f61c700aeb204ac4ba7d2f55/lib/Transforms/MemoryBanking.cpp#L80

lib/Transforms/MemoryBanking.cpp
```diff
@@ -88,6 +89,51 @@ MemRefType computeBankedMemRefType(MemRefType originalType,
   return newMemRefType;
 }
 
+// Updates n-dimensional index array `ndIndex` in-place by decoding the flat
+// index `linIndex` given the shape of the array. Assume row-major order.
+void decodeIndex(int64_t linIndex, ArrayRef<int64_t> shape,
```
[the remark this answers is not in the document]
jiahanxie353: sounds good!

lib/Transforms/MemoryBanking.cpp
```diff
@@ -88,6 +89,51 @@ MemRefType computeBankedMemRefType(MemRefType originalType,
...
+
+// Performs multi-dimensional slicing on `allAttrs` by extracting all elements
+// whose coordinates range from `bankCnt`*`bankingDimension` to
+// (`bankCnt`+1)*`bankingDimension` from `bankingDimension`'s dimension, leaving
+// other dimensions alone.
+SmallVector<SmallVector<Attribute>> sliceSubBlock(ArrayRef<Attribute> allAttrs,
+                                                  ArrayRef<int64_t> memShape,
+                                                  unsigned bankingDimension,
+                                                  unsigned bankingFactor) {
+  auto totalSize = [](auto &&shape) {
```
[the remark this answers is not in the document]
jiahanxie353: indeed!

jiahanxie353:
Thanks @cgyurgyik for the review and suggestions! I've finished modifying the code based on the request. And it's ready for another round of review!

lib/Transforms/MemoryBanking.cpp
```diff
@@ -88,6 +90,121 @@ MemRefType computeBankedMemRefType(MemRefType originalType,
...
+  builder.setInsertionPointAfter(globalOp);
+  OpBuilder::InsertPoint globalOpsInsertPt = builder.saveInsertionPoint();
+  builder.setInsertionPointAfter(getGlobalOp);
+  OpBuilder::InsertPoint getGlobalOpsInsertPt = builder.saveInsertionPoint();
+
+  for (size_t bankCnt = 0; bankCnt < bankingFactor; ++bankCnt) {
+    // Prepare relevant information to create a new GlobalOp
+    auto newMemRefTy = MemRefType::get(newShape, globalOpTy.getElementType());
+    auto newTypeAttr = TypeAttr::get(newMemRefTy);
+    std::string newNameStr =
```
cgyurgyik, line 175: Nit: It is clear from the type already (std::string), IMO we don't need the `Str` suffix here.
```suggestion
    std::string newName =
```

test/Transforms/memory_banking_multi_dim.mlir
cgyurgyik: Awesome!

cgyurgyik (approved):
LGTM, thanks!

lib/Transforms/MemoryBanking.cpp
jiahanxie353: sounds reasonable!

Plan to merge the patch once the CI has passed. Thanks again for the review :-)

PR #8048, opened 2025-01-09
[LoopScheduleToCalyx] Lower pipeline register to register writes.

cgyurgyik:
Lowers reads from a register to a pipeline register in the next stage. See the [Calyx Zulip discussion](https://calyx.zulipchat.com/#narrow/channel/423600-development/topic/hlstool/near/489718133).


mikeurbach (approved):
It's been a while since I looked at this, but it makes sense to me.

PR #8049, opened 2025-01-09
[SMT] Add bv2int op

TaoBi22:
Adds an operation to convert between SMT bitvectors and SMT integers. As discussed further in #8041, support for these operations varies between solvers - technically this one is outlined in the SMTLIB spec but AFAIU only to describe the semantics of other operations, rather than as something solvers are expected to implement, so I've made ExportSMTLIB error out on this one too since it's not _really_ part of the spec for the language itself.

### DIFF
--- a/include/circt/Dialect/SMT/SMTBitVectorOps.td
+++ b/include/circt/Dialect/SMT/SMTBitVectorOps.td
@@ -238,4 +238,20 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
   }];
 }
 
+def BV2IntOp : SMTOp<"bv2int", [Pure]> {
+  let summary = "Convert an SMT bitvector to an SMT integer.";
+  let description = [{
+    Designed to lower directly to an operation of the same name in Z3. The Z3
+    C API describes the semantics as follows:
+    Create an integer from the bit-vector argument t1.
+    If is_signed is false, then the bit-vector t1 is treated as unsigned.
+    So the result is non-negative and in the range [0..2^N-1], where N are the
+    number of bits in t1. If is_signed is true, t1 is treated as a signed
+    bit-vector.
+  }];
+  let arguments = (ins BitVectorType:$input, BoolAttr:$is_signed);
+  let results = (outs IntType:$result);
+  let assemblyFormat = [{$input custom<KeywordBool>($is_signed, "\"signed\"", "\"unsigned\"") attr-dict `:` qualified(type($input))}];
+}
+
 #endif // CIRCT_DIALECT_SMT_SMTBITVECTOROPS_TD
--- a/include/circt/Dialect/SMT/SMTBitVectorOps.td
+++ b/include/circt/Dialect/SMT/SMTBitVectorOps.td
@@ -248,10 +248,9 @@ def BV2IntOp : SMTOp<"bv2int", [Pure]> {
     N are the number of bits in `input`. If `is_signed` is true, `input` is
     treated as a signed bit-vector.
   }];
-  let arguments = (ins BitVectorType:$input, BoolAttr:$is_signed);
+  let arguments = (ins BitVectorType:$input, UnitAttr:$is_signed);
   let results = (outs IntType:$result);
-  let assemblyFormat = [{$input custom<KeywordBool>($is_signed, "\"signed\"",
-    "\"unsigned\"") attr-dict `:` qualified(type($input))}];
+  let assemblyFormat = [{$input (`signed` $is_signed^)? attr-dict `:` qualified(type($input))}];
 }
 
 #endif // CIRCT_DIALECT_SMT_SMTBITVECTOROPS_TD
--- a/include/circt/Dialect/SMT/SMTBitVectorOps.td
+++ b/include/circt/Dialect/SMT/SMTBitVectorOps.td
@@ -239,19 +239,19 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
 }
 
 def BV2IntOp : SMTOp<"bv2int", [Pure]> {
-  let summary = "Convert an SMT bitvector to an SMT integer.";
+  let summary = "Convert an SMT bit-vector to an SMT integer.";
   let description = [{
-    Designed to lower directly to an operation of the same name in Z3. The Z3
-    C API describes the semantics as follows:
-    Create an integer from the bit-vector argument t1.
-    If is_signed is false, then the bit-vector t1 is treated as unsigned.
-    So the result is non-negative and in the range [0..2^N-1], where N are the
-    number of bits in t1. If is_signed is true, t1 is treated as a signed
-    bit-vector.
+    Adapted from the Z3 documentation:
+    Create an integer from the bit-vector argument `input`.
+    If `is_signed` is false, then the bit-vector `input` is treated as
+    unsigned. So the result is non-negative and in the range [0..2^N-1], where
+    N are the number of bits in `input`. If `is_signed` is true, `input` is
+    treated as a signed bit-vector.
   }];
   let arguments = (ins BitVectorType:$input, BoolAttr:$is_signed);
   let results = (outs IntType:$result);
-  let assemblyFormat = [{$input custom<KeywordBool>($is_signed, "\"signed\"", "\"unsigned\"") attr-dict `:` qualified(type($input))}];
+  let assemblyFormat = [{$input custom<KeywordBool>($is_signed, "\"signed\"",
+    "\"unsigned\"") attr-dict `:` qualified(type($input))}];
 }
 
 #endif // CIRCT_DIALECT_SMT_SMTBITVECTOROPS_TD
--- a/integration_test/Dialect/SMT/basic.mlir
+++ b/integration_test/Dialect/SMT/basic.mlir
@@ -210,7 +210,7 @@ func.func @entry() {
   smt.solver () : () -> () {
     %c4 = smt.int.constant 4
     %c4_bv16 = smt.bv.constant #smt.bv<4> : !smt.bv<16>
-    %bv2int = smt.bv2int %c4_bv16 unsigned : !smt.bv<16>
+    %bv2int = smt.bv2int %c4_bv16 : !smt.bv<16>
     %eq = smt.distinct %c4, %bv2int : !smt.int
     func.call @check(%eq) : (!smt.bool) -> ()
     smt.yield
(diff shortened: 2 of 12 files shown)


lib/Conversion/SMTToZ3LLVM/LowerSMTToZ3LLVM.cpp
```diff
@@ -1211,6 +1211,25 @@ struct Int2BVOpLowering : public SMTLoweringPattern<Int2BVOp> {
...
+/// Z3_ast Z3_API Z3_mk_bv2int(Z3_context c, Z3_ast t1, bool is_signed)
+/// ```
+struct BV2IntOpLowering : public SMTLoweringPattern<BV2IntOp> {
+  using SMTLoweringPattern::SMTLoweringPattern;
+
+  LogicalResult
+  matchAndRewrite(BV2IntOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    Value isSignedConst = rewriter.create<LLVM::ConstantOp>(
+        op->getLoc(), rewriter.getI1Type(), op.getIsSigned());
```
maerhart, line 1225: I'm not sure if we can safely use the `i1` type here. I've seen `bool` being compiled to `i8` in LLVM, and I'm not sure if `i32` could also occur.
We should eventually insert a runtime library between this lowering and the Z3 API to have more control over that.
I'd at least insert some TODO or FIXME here.

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -602,9 +602,9 @@ struct StatementVisitor
                                     mlir::raw_indented_ostream &stream,
                                     ValueMap &valueMap) {
     // Ignore operations which are handled in the Expression Visitor.
-    if (isa<smt::Int2BVOp>(op))
+    if (isa<smt::Int2BVOp, BV2IntOp>(op))
       return op->emitError(
-          "int2bv operations are not supported for SMTLIB emission");
+          "int2bv and bv2int operations are not supported for SMTLIB emission");
 
     return success();
```
maerhart, line 609: Would this work?
```
return op->emitOpError("operation not supported for SMTLIB emission");
```

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -238,4 +238,20 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
   }];
 }
 
+def BV2IntOp : SMTOp<"bv2int", [Pure]> {
+  let summary = "Convert an SMT bitvector to an SMT integer.";
+  let description = [{
+    Designed to lower directly to an operation of the same name in Z3. The Z3
+    C API describes the semantics as follows:
+    Create an integer from the bit-vector argument t1.
```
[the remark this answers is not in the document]
TaoBi22: Oh yeah, good point!

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -238,4 +238,20 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
...
+    If is_signed is false, then the bit-vector t1 is treated as unsigned.
+    So the result is non-negative and in the range [0..2^N-1], where N are the
+    number of bits in t1. If is_signed is true, t1 is treated as a signed
+    bit-vector.
+  }];
+  let arguments = (ins BitVectorType:$input, BoolAttr:$is_signed);
+  let results = (outs IntType:$result);
+  let assemblyFormat = [{$input custom<KeywordBool>($is_signed, "\"signed\"", "\"unsigned\"") attr-dict `:` qualified(type($input))}];
```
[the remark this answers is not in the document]
TaoBi22: Oops, thanks!

integration_test/Dialect/SMT/basic.mlir
```diff
@@ -205,6 +205,40 @@ func.func @entry() {
...
+  // CHECK: unsat
+  // CHECK: Res: -1
+  smt.solver () : () -> () {
+    %c-4 = smt.int.constant -4
+    %c-4_bv16 = smt.bv.constant #smt.bv<-4> : !smt.bv<16>
+    %bv2int = smt.bv2int %c-4_bv16 unsigned : !smt.bv<16>
+    %eq = smt.eq %c-4, %bv2int : !smt.int
+    func.call @check(%eq) : (!smt.bool) -> ()
+    smt.yield
+  }
```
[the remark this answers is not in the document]
TaoBi22: I'd personally prefer to keep both - given this test only checks it doesn't do _one_ thing we don't expect it to, rather than checking it does what we expect it to, I'd be hesitant to make it the only test

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -238,4 +238,20 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
   }];
 }
 
+def BV2IntOp : SMTOp<"bv2int", [Pure]> {
+  let summary = "Convert an SMT bitvector to an SMT integer.";
+  let description = [{
+    Designed to lower directly to an operation of the same name in Z3. The Z3
+    C API describes the semantics as follows:
```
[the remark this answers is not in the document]
TaoBi22: Makes sense - I'll change how I've phrased it, but I'll leave in a mention of Z3 since I don't want to lift from their documentation and not say that's where it's from (and the SMTLIB spec only concerns the signed variant).

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -238,4 +238,20 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
...
+    Create an integer from the bit-vector argument t1.
+    If is_signed is false, then the bit-vector t1 is treated as unsigned.
+    So the result is non-negative and in the range [0..2^N-1], where N are the
+    number of bits in t1. If is_signed is true, t1 is treated as a signed
+    bit-vector.
+  }];
+  let arguments = (ins BitVectorType:$input, BoolAttr:$is_signed);
```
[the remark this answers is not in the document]
TaoBi22: Is it possible to have it be a UnitAttr and still have it require a keyword either way? I chose this format because it means the op always has to specify, which I prefer as I'm not sure either signed or unsigned seems like an obvious default behaviour - happy to be overridden though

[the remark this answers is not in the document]
maerhart: Not sure if there's a builtin way to do this. But definitely possible with a custom directive.
I was thinking about choosing `unsigned` as the default and adding `signed` if it should be interpreted with a sign bit to keep it simple (it's also what `smt.fmt.dec` chose, for example).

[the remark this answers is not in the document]
TaoBi22: Ah cool, if that's what's done elsewhere in the dialect I'll keep it the same here for consistency

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
```diff
@@ -602,9 +602,8 @@ struct StatementVisitor
                                     mlir::raw_indented_ostream &stream,
                                     ValueMap &valueMap) {
     // Ignore operations which are handled in the Expression Visitor.
-    if (isa<smt::Int2BVOp>(op))
-      return op->emitError(
-          "int2bv operations are not supported for SMTLIB emission");
+    if (isa<smt::Int2BVOp, BV2IntOp>(op))
+      return op->emitError("operation not supported for SMTLIB emission");
 
     return success();
```
maerhart, line 608: I think just
```
  return op->emitError("operation not supported for SMTLIB emission");
```
should work instead of
```
    if (isa<smt::Int2BVOp, BV2IntOp>(op))
      return op->emitError("operation not supported for SMTLIB emission");

    return success();
```
to generically emit this error if we add new operations that we don't explicitly add support for in ExportSMTLIB (i.e., should be more robust)

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -238,4 +238,19 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
...
+    Adapted from the Z3 documentation:
+    Create an integer from the bit-vector argument `input`.
+    If `is_signed` is false, then the bit-vector `input` is treated as
+    unsigned. So the result is non-negative and in the range [0..2^N-1], where
+    N are the number of bits in `input`. If `is_signed` is true, `input` is
+    treated as a signed bit-vector.
+  }];
+  let arguments = (ins BitVectorType:$input, UnitAttr:$is_signed);
+  let results = (outs IntType:$result);
+  let assemblyFormat = [{$input (`signed` $is_signed^)? attr-dict `:` qualified(type($input))}];
```
maerhart, line 253: 80 cols

include/circt/Dialect/SMT/SMTBitVectorOps.td
```diff
@@ -238,4 +238,19 @@ def RepeatOp : SMTBVOp<"repeat", [Pure]> {
...
+def BV2IntOp : SMTOp<"bv2int", [Pure]> {
+  let summary = "Convert an SMT bit-vector to an SMT integer.";
+  let description = [{
+    Adapted from the Z3 documentation:
+    Create an integer from the bit-vector argument `input`.
+    If `is_signed` is false, then the bit-vector `input` is treated as
+    unsigned. So the result is non-negative and in the range [0..2^N-1], where
+    N are the number of bits in `input`. If `is_signed` is true, `input` is
+    treated as a signed bit-vector.
+  }];
```
maerhart, line 250: I think this would be a bit more clear given the change to UnitAttr:
```
  let description = [{
    Create an integer from the bit-vector argument `input`. If `is_signed` is
    present, the bit-vector is treated as two's complement signed.  Otherwise,
    it is treated as an unsigned integer in the range [0..2^N-1], where N is the
    number of bits in `input`.
  }];
```

TaoBi22: Ah yeah meant to change that, thanks!

TaoBi22:
I'll push fixes to main, thanks!

lib/Target/ExportSMTLIB/ExportSMTLIB.cpp
TaoBi22: Ah I see, sorry for the misunderstanding before!

TaoBi22: Just checked and this actually breaks all the ExportSMTLIB tests - I'm not super clear on the Visitor infrastructure so I'm not sure why yet, but from a brief scan it looks like getting rid of the condition causes the error to throw on the first non-ConstantLike op the export pass encounters.

Scutua (approved):

PR #8051, opened 2025-01-09
[Python] Update wheels we build in CI/CD.

mikeurbach:
This removes MacOS, which has been failing, and no one seems to care about anyway. This also adds a newer Python 3.13 wheel for a more modern Python version. I have not yet removed the Python 3.8 wheel, but that is EOL, and we can consider removing that in a follow up.

### DIFF
--- a/.github/workflows/uploadWheels.yml
+++ b/.github/workflows/uploadWheels.yml
@@ -23,13 +23,9 @@ jobs:
       matrix:
         config:
           - os: ubuntu-20.04
-            cibw_build: cp38-manylinux_x86_64
-          - os: ubuntu-20.04
             cibw_build: cp310-manylinux_x86_64
-          - os: macos-13
-            cibw_build: cp38-macosx_x86_64
-          - os: macos-13
-            cibw_build: cp310-macosx_x86_64
+          - os: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


mikeurbach:
@teqdruid do you care about Python 3.8 wheels for the core bindings package? I might remove those as well.

mikeurbach:
Kicked off a sanity test wheel build here: https://github.com/llvm/circt/actions/runs/12697878099.

teqdruid:
> @teqdruid do you care about Python 3.8 wheels for the core bindings package? I might remove those as well.

I don't use the core bindings package, so no.

I do, however, care about being able to build the Python 3.8 bindings as PyCDE continues to support 3.8.

teqdruid (approved):

mikeurbach:
Thanks, makes sense. I don't think anyone is doing anything that won't be compatible with Python 3.8. We can keep using the lowest common denominator Python features (e.g., avoid the fancier new generics syntax in Python 3.13). If no one cares about this version of the core bindings wheels, I'll remove that in this PR.

mikeurbach:
Wheel build is passing for Python 3.13 with the cibuildwheel update. I'll go ahead and merge this now.

PR #8055, opened 2025-01-09
[LoopScheduleToCalyx] deduplicate groups within a ParOp.

### LINKED ISSUE #8050: [LoopScheduleToCalyx] Fail due to op still in use
When calling `circt-opt --lower-loopschedule-to-calyx ./test.mlir`,  `circt-opt` fail due to 
```
circt-opt: /home/kelvin/circt/llvm/mlir/lib/IR/PatternMatch.cpp:182: auto mlir::RewriterBase::eraseOp(mlir::Operation *)::(anonymous class)::operator()(mlir::Operation *) const: Assertion `mayBeGraphRegion(*op->getParentRegion()) && "expected that op has no uses"' failed.
```

test.mlir:
```mlir
module {
  func.func @foo(%arg0: i32) attributes {} {
    %const = arith.constant 1 : index
    loopschedule.pipeline II =  2 trip_count =  20 iter_args(%counter = %const) : (index) -> () {
      %latch = arith.cmpi ult, %counter, %const : index
      loopschedule.register %latch : i1
    } do {
      %stage0:2 = loopschedule.pipeline.stage start = 0 {
        %op = arith.addi %counter, %const : index
        %count = arith.addi %counter, %const : index
        loopschedule.register %op, %count : index, index
      } : index, index
      %stage1 = loopschedule.pipeline.stage start = 1 {
        loopschedule.register %stage0#0: index
      } : index
      %stage3 = loopschedule.pipeline.stage start = 7 {
        %moreOp = arith.addi %stage1, %const : index
        loopschedule.register %moreOp : index
      } : index
      loopschedule.terminator iter_args(%stage0#1), results() : (index) -> ()
    }
    return
  }
}
```

I have spent some time looking at what led to this, and it seems like the `%stage1` value is leading to the problem.

cgyurgyik:
Deduplicates multiple occurrences of the same Enable in a `ParOp`. This can be introduced automatically by a compiler lowering phase, so we add this as a canonicalization pattern. Additionally, ensures that the pipeline register (and not the pipeline stage) is used in Calyx computation assignments.

Fixes #8050.


mikeurbach (commented):
I think this makes sense as a canonicalization, but is the error avoided with the change in LoopScheduleToCalyx? In general, I think we strive for a world where the canonicalizations are not required for success / correctness.

cgyurgyik:
> I think this makes sense as a canonicalization, but is the error avoided with the change in LoopScheduleToCalyx? In general, I think we strive for a world where the canonicalizations are not required for success / correctness.

It is not, that change avoids a use after free issue. I have not totally fixed the error yet though so going back to draft mode

cgyurgyik:
@mikeurbach I've updated this so that canonicalization isn't used as an "operation fixer-upper".

mikeurbach (approved):
awesome, thanks for that. looks good to me.

cgyurgyik:
Thanks! This is branched off #8048 so if you have a chance to review that, it'd be great :-)

PR #8056, opened 2025-01-09
[RTGTest] Add a few instructions

maerhart:
Tests for the `printInstructionBinary` and `printInstructionAssembly` implementations will follow in the next PR.


uenoku (approved):
LGTM.  Nice use of a FoldAdaptor! One nit in my mind was these instructions look very RV32I specific even though rtg/rtgtest sound like generic random test dialects. Not sure worth modifying but I was thinking perhaps it's more clear if these instructions were named like `rtgtest.rv32i.*`. It looks great as is so I think it's up to you.

darthscsi (approved):

PR #8066, opened 2025-01-10
[FIRRTL] Reuse XMRRefOps in LowerXMR

seldridge:
Change the `LowerXMR` pass to reuse previously created `XMRRefOp`s within a module.  This is done for later, easier processing of these operations in `LowerToHW`.  While these operations could be made to be CSE'd (by giving them the `Pure` trait), it is advantageous to instead not create duplicate operations as this avoids needing to run schedule a mostly useless CSE.

This is primarily done to allow for Verilator-compatible lowering of `force_initial` and `release_initial` in `LowerToHW` where it would be useful to know when a force and release are referring to the same reference.  (Yes, full alias analysis is necessary to do this correctly and relying on the CSE of `LowerXMR` is going to be buggy.)

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerXMR.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerXMR.cpp
@@ -76,6 +76,50 @@ struct XMRNode {
   os << ", next=" << node.next << ")";
   return os;
 }
+
+/// Track information about operations being created in a module.  This is used
+/// to generate more compact code and reuse operations where possible.
+class ModuleState {
+
+public:
+  ModuleState(FModuleOp &moduleOp) : body(moduleOp.getBodyBlock()) {}
+
+  /// Return the existing XMRRefOp for this type, symbol, and suffix for this
+  /// module.  Otherwise, create a new one.  The first XMRRefOp will be created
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LowerXMR.cpp
```diff
@@ -488,7 +534,10 @@ class LowerXMRPass : public circt::firrtl::impl::LowerXMRBase<LowerXMRPass> {
               if (failed(resolveReference(op.getDest(), builder, ref, str)))
                 return failure();
 
-              Value xmr = builder.create<XMRRefOp>(destType, ref, str);
+              Value xmr =
+                  moduleStates.find(op->template getParentOfType<FModuleOp>())
```
seldridge, line 538: This `template` can cause problems with MSVCC. I'm checking if it works here: https://github.com/llvm/circt/actions/runs/12718725232

seldridge: This looks okay!

prithayan (approved):
Looks good.

Scutua (approved):

PR #8068, opened 2025-01-12
[HW][circt-synth] Implement AggregateToComb pass and add to circt-synth pipeline

uenoku:
This PR adds a new pass `HWAggregateToComb` that lowers array operations to comb operations. The pass converts:

- `hw.array_get` to a mux tree using comb operations
- `hw.array_create` and `hw.array_concat` to `comb.concat`
- `hw.aggregate_constant` to `comb.concat` of individual constants

The PR also moves two utility functions from CombToAIG to the Comb dialect:
- `extractBits`: Extracts individual bits from a value
- `constructMuxTree`: Builds a multiplexer tree from selectors and leaf nodes

The pass is required to run before CombToAIG to ensure array operations are
properly lowered to AIG. Strictly speaking other than array_get operations can be preserved
but array values prevent optimizations at AIG level.  So for the simplicity, I think it's reasonable
to lower array operations within circt-synth pipeline. 

I currently added the pass under `HW/Transforms` considering `FlattenIO` is very similar but the pass might better belong to Conversion.

### DIFF
--- a/include/circt/Dialect/HW/Passes.td
+++ b/include/circt/Dialect/HW/Passes.td
@@ -95,6 +95,10 @@ def HWAggregateToComb : Pass<"hw-aggregate-to-comb", "hw::HWModuleOp"> {
     This pass lowers aggregate *operations* to comb operations within modules.
     Note that this pass does not lower ports. Ports lowering is handled
     by FlattenIO.
+
+    This pass will change the behavior of out-of-bounds access of arrays,
+    specifically the last element of the array is used as a value for
+    out-of-bounds access.
   }];
   let dependentDialects = ["comb::CombDialect"];
 }
--- a/include/circt/Dialect/HW/Passes.td
+++ b/include/circt/Dialect/HW/Passes.td
@@ -93,8 +93,8 @@ def HWAggregateToComb : Pass<"hw-aggregate-to-comb", "hw::HWModuleOp"> {
 
   let description = [{
     This pass lowers aggregate *operations* to comb operations within modules.
-    This pass does not lower ports, as ports are handled by FlattenIO. This pass
-    will also change the behavior of out-of-bounds access of arrays.
+    Note that this pass does not lower ports. Ports lowering is handled
+    by FlattenIO.
   }];
   let dependentDialects = ["comb::CombDialect"];
 }
--- a/include/circt/Dialect/HW/Passes.td
+++ b/include/circt/Dialect/HW/Passes.td
@@ -87,4 +87,16 @@ def FooWires : Pass<"hw-foo-wires", "hw::HWModuleOp"> {
   let constructor = "circt::hw::createFooWiresPass()";
 }
 
+def HWAggregateToComb : Pass<"hw-aggregate-to-comb", "hw::HWModuleOp"> {
+  let summary = "Lower aggregate operations to comb operations";
+  let constructor = "circt::hw::createHWAggregateToCombPass()";
+
+  let description = [{
+    This pass lowers aggregate *operations* to comb operations within modules.
+    This pass does not lower ports, as ports are handled by FlattenIO. This pass
+    will also change the behavior of out-of-bounds access of arrays.
+  }];
+  let dependentDialects = ["comb::CombDialect"];
+}
+
 #endif // CIRCT_DIALECT_HW_PASSES_TD
--- a/integration_test/circt-synth/comb-lowering-lec.mlir
+++ b/integration_test/circt-synth/comb-lowering-lec.mlir
@@ -1,7 +1,7 @@
 // REQUIRES: libz3
 // REQUIRES: circt-lec-jit
 
-// RUN: circt-opt %s --convert-comb-to-aig --convert-aig-to-comb -o %t.mlir
+// RUN: circt-opt %s --hw-aggregate-to-comb --convert-comb-to-aig --convert-aig-to-comb -o %t.mlir
 // RUN: circt-lec %t.mlir %s -c1=bit_logical -c2=bit_logical --shared-libs=%libz3 | FileCheck %s --check-prefix=COMB_BIT_LOGICAL
 // COMB_BIT_LOGICAL: c1 == c2
 hw.module @bit_logical(in %arg0: i32, in %arg1: i32, in %arg2: i32, in %arg3: i32,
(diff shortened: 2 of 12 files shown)


maerhart:
Would it make sense to have aggregate to comb lowering in a separate pass? To me, it seems like it could also be useful outside of the AIG lowering path.

uenoku:
> Would it make sense to have aggregate to comb lowering in a separate pass? 

Generally yes, I agree. One concern is that the lowering pattern for `HWArrayGetOpConversion` might be too low-level for general use case. ArrayGet is lowered into mux-chains for each selector bits but some downstream pass might prefer chains of `comb.icmp`. Anyway I think having `HWAggregateToComb` looks great.

lib/Dialect/HW/Transforms/HWAggregateToComb.cpp
```diff
@@ -0,0 +1,187 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/HW/HWPasses.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "llvm/ADT/TypeSwitch.h"
```
uenoku, line 14: Unnecessary include

lib/Dialect/HW/Transforms/HWAggregateToComb.cpp
```diff
@@ -0,0 +1,186 @@
...
+    while (!worklist.empty()) {
+      auto current = worklist.pop_back_val();
+      if (auto innerArray = dyn_cast<ArrayAttr>(current)) {
+        for (auto elem : llvm::reverse(innerArray))
+          worklist.push_back(elem);
+        continue;
+      }
+
+      if (auto intAttr = dyn_cast<IntegerAttr>(current)) {
+        results.push_back(rewriter.create<hw::ConstantOp>(loc, intAttr));
```
maerhart, line 62: Wouldn't it be better to have an APInt and using its `concat` method rather than materializing all this IR just to fold it?

lib/Dialect/HW/Transforms/HWAggregateToComb.cpp
```diff
@@ -0,0 +1,186 @@
...
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Value> results;
+    auto arrayType = cast<hw::ArrayType>(op.getInput().getType());
+    auto elemType = arrayType.getElementType();
+    auto numElements = arrayType.getNumElements();
+    auto elemWidth = hw::getBitWidth(elemType);
+    if (elemWidth < 0)
+      return rewriter.notifyMatchFailure(op.getLoc(), "unknown element width");
+
+    auto lowered = rewriter.getRemappedValue(op.getInput());
```
maerhart, line 99: Why not use the adaptor? We wouldn't need the check below then.
```suggestion
    auto lowered = adaptor.getInput();
```

integration_test/circt-synth/comb-lowering-lec.mlir
```diff
@@ -78,3 +78,19 @@ hw.module @shift5(in %lhs: i5, in %rhs: i5, out out_shl: i5, out out_shr: i5, ou
...
+// RUN: circt-lec %t.mlir %s -c1=array -c2=array --shared-libs=%libz3 | FileCheck %s --check-prefix=COMB_ARRAY
+// COMB_ARRAY: c1 == c2
+hw.module @array(in %arg0: i2, in %arg1: i2, in %arg2: i2, in %arg3: i2, in %sel1: i2, in %sel2: i2, out out1: i2, out out2: i2) {
+  %0 = hw.array_create %arg0, %arg1, %arg2, %arg3 : i2
+  %1 = hw.array_get %0[%sel1] : !hw.array<4xi2>, i2
+  %2 = hw.array_create %arg0, %arg1, %arg2 : i2
+  %c3_i2 = hw.constant 3 : i2
+  // NOTE: If the index is out of bounds, the result value is undefined.
+  // In LEC such value is lowered into unbounded SMT variable and cause
+  // the LEC to fail. So just asssume that the index is in bounds.
```
maerhart, line 91: Yeah, for these kind of things, a translation validation tool would be quite nice 🙂

test/Dialect/HW/hw-aggregate-to-comb.mlir
```diff
@@ -0,0 +1,61 @@
...
+// CHECK-LABEL: @array_concat
+hw.module @array_concat(in %lhs: !hw.array<2xi4>, in %rhs: !hw.array<3xi4>, out out: i4) {
+  %0 = hw.array_concat %lhs, %rhs : !hw.array<2xi4>, !hw.array<3xi4>
+  %c_i2 = hw.constant 3 : i3
+  // CHECK-NEXT: %[[BITCAST_RHS:.+]] = hw.bitcast %rhs : (!hw.array<3xi4>) -> i12
+  // CHECK-NEXT: %[[BITCAST_LHS:.+]] = hw.bitcast %lhs : (!hw.array<2xi4>) -> i8
+  // CHECK-NEXT: %[[CONCAT:.+]] = comb.concat %[[BITCAST_LHS]], %[[BITCAST_RHS]] : i8, i12
+  // CHECK:      %[[EXTRACT:.+]] = comb.extract %[[CONCAT]] from 12 : (i20) -> i4
+  // CHECK:      hw.output %[[EXTRACT]] : i4
+  %1 = hw.array_get %0[%c_i2] : !hw.array<5xi4>, i3
```
maerhart, line 32: Maybe get rid of this array_get to make sure array outputs are also fine? array_get itself is already tested above.

maerhart (approved):
LGTM!

lib/Dialect/HW/Transforms/HWAggregateToComb.cpp
uenoku: Oh yeah. That works, thanks!

integration_test/circt-synth/comb-lowering-lec.mlir
uenoku: Yeah without circt-lec it has been almost impossible to implement lowering pattern with confidence for the correctness. Thank you @maerhart @frog-in-the-well @TaoBi22 for outstanding work!

include/circt/Dialect/HW/Passes.td
```diff
@@ -87,4 +87,16 @@ def FooWires : Pass<"hw-foo-wires", "hw::HWModuleOp"> {
...
 }
 
+def HWAggregateToComb : Pass<"hw-aggregate-to-comb", "hw::HWModuleOp"> {
+  let summary = "Lower aggregate operations to comb operations";
+  let constructor = "circt::hw::createHWAggregateToCombPass()";
+
+  let description = [{
+    This pass lowers aggregate *operations* to comb operations within modules.
+    This pass does not lower ports, as ports are handled by FlattenIO. This pass
+    will also change the behavior of out-of-bounds access of arrays.
```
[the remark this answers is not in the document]
uenoku: > Can we mention here how it changes the behavior? 

Yes. I'll do.

> If I understand correctly it's just a refinement

Yes, it's refinement so it should be fine from the semantics perspective but i would at least mention in the doc. Maybe in the future we might want to run post-synthesis simulation on arcilator but only pre-synthesis IR can catch out-of-bounds access in this case.

lib/Dialect/HW/Transforms/HWAggregateToComb.cpp
```diff
@@ -0,0 +1,186 @@
...
+  void runOnOperation() override;
+  using HWAggregateToCombBase<HWAggregateToCombPass>::HWAggregateToCombBase;
+};
+} // namespace
+
+void HWAggregateToCombPass::runOnOperation() {
+  ConversionTarget target(getContext());
+
+  // TODO: Add structs.
+  target.addIllegalOp<hw::ArrayGetOp, hw::ArrayCreateOp, hw::ArrayConcatOp,
```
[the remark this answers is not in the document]
uenoku: I'm planning on adding struct as well (even though all operations are fine but structs do block integer optimizations) so I'll mention both. Thanks!

PR #8071, opened 2025-01-12
[SMT][python] enable python bindings

makslevental:
This PR enables the python bindings for the `smt` dialect.

I'm not sure what y'alls approach is to testing these bindings since [PyCDE only tests the bindings it uses](https://github.com/llvm/circt/tree/main/frontends/PyCDE/test) and I don't see any other python bindings tests around...

### DIFF
--- a/include/circt-c/Dialect/SMT.h
+++ b/include/circt-c/Dialect/SMT.h
@@ -1,4 +1,4 @@
-//===- SMT.h - C interface for the SMT dialect ----------------*- C -*-===//
+//===- SMT.h - C interface for the SMT dialect --------------------*- C -*-===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
--- a/include/circt-c/Dialect/SMT.h
+++ b/include/circt-c/Dialect/SMT.h
@@ -1,0 +1,24 @@
+//===- SMT.h - C interface for the SMT dialect ----------------*- C -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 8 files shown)


uenoku:
Thanks! We test python bindings in `integration_test/Bindings/Python/dialects`.

makslevental:
@uenoku added a "smoke test"

include/circt-c/Dialect/SMT.h
```diff
@@ -0,0 +1,24 @@
+//===- SMT.h - C interface for the SMT dialect ----------------*- C -*-===//
```
maerhart, line 1: 80 cols

lib/Bindings/Python/dialects/SMTOps.td
```diff
@@ -0,0 +1,14 @@
+//===- SMTOps.td - Entry point for SMT bindings --------*- tablegen -*-===//
```
maerhart, line 1: 80 cols

lib/Bindings/Python/dialects/SMTOps.td
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef BINDINGS_PYTHON_SMT_OPS
+#define BINDINGS_PYTHON_SMT_OPS
+
+include "circt/Dialect/SMT/SMT.td"
+
+#endif
```
maerhart, line 14:
```suggestion
#endif // BINDINGS_PYTHON_SMT_OPS
```

lib/CAPI/Dialect/SMT.cpp
```diff
@@ -0,0 +1,14 @@
+//===- SMT.cpp - C interface for the SMT dialect ----------------------===//
```
maerhart, line 1: 80 cols

maerhart (approved):
Thanks for adding this!

PR #8074, opened 2025-01-13
[MemoryBanking] Bank memref::copy logically using memref::subView

jiahanxie353:
This patch banks `memref::copy` op logically when the operation is not used in parallel ops (which implies that we don't need to create new physical memories) by using `memref::subView` op.

### DIFF
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -409,6 +409,141 @@ private:
   DenseSet<Value> &oldMemRefVals;
 };
 
+struct BankMemrefCopyPattern : public OpRewritePattern<memref::CopyOp> {
+  BankMemrefCopyPattern(MLIRContext *ctx,
+                        uint64_t bankingFactor,
+                        unsigned bankingDimension,
+                        DenseMap<Value, SmallVector<Value>> &memoryToBanks)
+      : OpRewritePattern<memref::CopyOp>(ctx),
+        bankingFactor(bankingFactor),
+        bankingDimension(bankingDimension),
+        memoryToBanks(memoryToBanks) {}
+
+  LogicalResult matchAndRewrite(memref::CopyOp copyOp,
+                                PatternRewriter &rewriter) const override {
+    Value srcMem = copyOp.getSource();
+    Value dstMem = copyOp.getTarget();
+    
+    bool srcIsBanked = memoryToBanks.count(srcMem) != 0;
+    bool dstIsBanked = memoryToBanks.count(dstMem) != 0;
+    if (!srcIsBanked && !dstIsBanked) {
+      copyOp.emitError("At least one of the source and destination memory reference should be banked");
+      return failure();
+    }
+
+    Location loc = copyOp.getLoc();
+
+    if (srcIsBanked && dstIsBanked) {
+      auto &srcBanks = memoryToBanks.find(srcMem)->second;
+      auto &dstBanks = memoryToBanks.find(dstMem)->second;
+
+      rewriter.setInsertionPoint(copyOp);
+      for (size_t i = 0; i < bankingFactor; ++i)
+        rewriter.create<memref::CopyOp>(loc, srcBanks[i], dstBanks[i]);
+    } else if (srcIsBanked) {
+      auto &srcBanks = memoryToBanks.find(srcMem)->second;
+      if (failed(copyFromBanksToUnbanked(rewriter, loc,
+                                         srcBanks, dstMem,
+                                         bankingFactor, bankingDimension))) {
+        return failure();
+      }
+    } else {
+      auto &dstBanks = memoryToBanks.find(dstMem)->second;
+      if (failed(copyFromUnbankedToBanks(rewriter, loc, srcMem, dstBanks, bankingFactor, bankingDimension)))
+        return failure();
+    }
+
+    rewriter.eraseOp(copyOp);
+
+    return success();
+  }
+
+private:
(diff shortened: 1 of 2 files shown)


lib/Transforms/MemoryBanking.cpp
```diff
@@ -409,6 +409,151 @@ struct BankAffineStorePattern
...
+  LogicalResult copyFromBanksToUnbanked(PatternRewriter &rewriter, Location loc,
+                                        ArrayRef<Value> srcBanks, Value dstMem,
+                                        uint64_t bankingFactor,
+                                        unsigned bankingDimension) const {
+    MemRefType dstType = dyn_cast<MemRefType>(dstMem.getType());
+    if (!dstType || dstType.getRank() <= bankingDimension)
+      return failure();
+
+    ArrayRef<int64_t> dstShape = dstType.getShape();
+    int64_t chunkSize = dstShape[bankingDimension] / (int64_t)bankingFactor;
```
cgyurgyik, line 511: Please avoid implicit casts (here and above and below).

```suggestion
    int64_t chunkSize = dstShape[bankingDimension] / static_cast<int64_t>(bankingFactor);
```

lib/Transforms/MemoryBanking.cpp
```diff
@@ -409,6 +409,151 @@ struct BankAffineStorePattern
...
+                                        Value srcMem, ArrayRef<Value> dstBanks,
+                                        uint64_t bankingFactor,
+                                        unsigned bankingDimension) const {
+    auto srcType = dyn_cast<MemRefType>(srcMem.getType());
+    if (!srcType || srcType.getRank() <= bankingDimension)
+      return failure();
+
+    ArrayRef<int64_t> srcShape = srcType.getShape();
+    // We'll assume static shapes and exact divisibility.
+    int64_t chunkSize = srcShape[bankingDimension] / (int64_t)bankingFactor;
```
cgyurgyik, line 540: same thing here, prefer explicit cast vs implicit.

lib/Transforms/MemoryBanking.cpp
```diff
@@ -409,6 +409,151 @@ struct BankAffineStorePattern
...
+      rewriter.create<memref::CopyOp>(loc, srcBanks[i], subView);
+    }
+
+    return success();
+  }
+
+  /// Copy from unbanked source memref `srcMem` to destination memory references
+  /// `dstBanks`, where each bank has the same MemRefType, and `srcMem` has the
+  /// shape of the "original" unbanked array.
+  LogicalResult copyFromUnbankedToBanks(PatternRewriter &rewriter, Location loc,
```
cgyurgyik, line 530: This seems to be almost exactly the same as the function above. Can we add additional arguments or template parameters to avoid having two nearly duplicate functions? WDYT?

lib/Transforms/MemoryBanking.cpp
```diff
@@ -409,6 +409,151 @@ struct BankAffineStorePattern
...
+      : OpRewritePattern<memref::CopyOp>(ctx), bankingFactor(bankingFactor),
+        bankingDimension(bankingDimension), memoryToBanks(memoryToBanks) {}
+
+  LogicalResult matchAndRewrite(memref::CopyOp copyOp,
+                                PatternRewriter &rewriter) const override {
+    Value srcMem = copyOp.getSource();
+    Value dstMem = copyOp.getTarget();
+
+    bool srcIsBanked = memoryToBanks.count(srcMem) != 0;
+    bool dstIsBanked = memoryToBanks.count(dstMem) != 0;
```
cgyurgyik, line 425: Again, a good habit here to avoid double lookups is

```cpp
auto srcIt = memoryToBanks.find(srcMem);
auto dstIt = memoryToBanks.find(dstMem);
bool srcIsBanked = srcIt != memoryToBanks.end();
bool dstIsBanked = dstIt != memoryToBanks.end();

...

auto& srcBanks = srcIt->second;
```

lib/Transforms/MemoryBanking.cpp
```diff
@@ -409,6 +409,151 @@ struct BankAffineStorePattern
...
+    SmallVector<OpFoldResult> offsets(rank, rewriter.getIndexAttr(0));
+    SmallVector<OpFoldResult> sizes(rank);
+    SmallVector<OpFoldResult> strides(rank, rewriter.getIndexAttr(1));
+
+    for (int64_t d = 0; d < rank; ++d) {
+      int64_t dimSize = shape[d];
+      if (d == bankingDimension) {
+        // Round-robin partition: offset = bankIndex, size = chunkSize, stride =
+        // bankingFactor
+        int64_t chunkSize = dimSize / (int64_t)bankingFactor;
```
cgyurgyik, line 482: Same here

lib/Transforms/MemoryBanking.cpp
```diff
@@ -409,6 +409,151 @@ struct BankAffineStorePattern
   DenseSet<Value> &oldMemRefVals;
 };
 
+struct BankMemrefCopyPattern : public OpRewritePattern<memref::CopyOp> {
```
cgyurgyik, line 412: Add a comment here please describing the purpose/goal of this rewrite.

jiahanxie353:
Closing the PR because it's not a crucial pass for now.
Initially, I wanted to avoid returning a `memref` type in `func`, so I ran `--buffer-results-to-out-params` and it turned the returning `memref` result into `memref::copy` to a newly created function argument. But this can be avoided if we specify the `hoist-static-allocs` option.

PR #8075, opened 2025-01-13
[SCFToCalyx] Re-initialize IRMapping at the start of each loop

jiahanxie353:
This patch addresses the issue with `IRMapping` when unrolling the parallel loop during the SCFToCalyx pass. If there are region-based operations, such as `scf.for` inside `scf.parallel`, the block arguments weren't cloned correctly because if we keep reusing the old `IRMapping`, the `IRMapping` will already have the entries from previous iterations that overshadows new ones. To fix it, we need to re-initialize the `IRMapping` every iteration.


jiahanxie353:
Going to merge it since the CI has passed and this patch does not introduce significant change. cc @cgyurgyik @rachitnigam @andrewb1999

PR #8076, opened 2025-01-13
[Verif] disambiguate ctor

makslevental:
`clang-tidy` reports this ctor call as ambiguous and in fact it fails on Windows with `clang-cl` with 

```
  D:\a\circt\circt\lib\Conversion\VerifToSV\VerifToSV.cpp(48,22): error: call to constructor of 'APInt' is ambiguous
     48 |         op.getLoc(), APInt(32, 0x80000001, false));
        |                      ^     ~~~~~~~~~~~~~~~~~~~~~
  D:\a\circt\circt\llvm\llvm\include\llvm/ADT/APInt.h(157,3): note: candidate constructor
    157 |   APInt(unsigned numBits, unsigned numWords, const uint64_t bigVal[]);
        |   ^
  D:\a\circt\circt\llvm\llvm\include\llvm/ADT/APInt.h(111,3): note: candidate constructor
    111 |   APInt(unsigned numBits, uint64_t val, bool isSigned = false,
        |   ^
  1 error generated.
```

### DIFF
--- a/lib/Conversion/VerifToSV/VerifToSV.cpp
+++ b/lib/Conversion/VerifToSV/VerifToSV.cpp
@@ -44,8 +44,8 @@ struct PrintOpConversionPattern : public OpConversionPattern<PrintOp> {
                   ConversionPatternRewriter &rewriter) const override {
 
     // Printf's will be emitted to stdout (32'h8000_0001 in IEEE Std 1800-2012).
-    Value fdStdout = rewriter.create<hw::ConstantOp>(
-        op.getLoc(), APInt(32, 0x80000001, false));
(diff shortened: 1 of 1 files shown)


makslevental:
> Interesting that this doesn't work properly on windows, but since 'false' is the default anyway this LGTM.

Presumably it does work on `msvc.exe` because you guys have [tests](https://github.com/llvm/circt/actions/workflows/buildAndTestWindows.yml)? But yea not sure why `clang-cl.exe` complains (and then what's funnier/weirder is that it doesn't fail for `clang` on Linux/Mac :shrug:).

PR #8077, opened 2025-01-13
[CMake] enable install C API objects

makslevental:
I would like to build CIRCT Python bindings *downstream of CIRCT*. For this to work you need to have the so-called "aggregate objects" [installed](https://github.com/llvm/llvm-project/blob/3318a7248ae464af0abd0bea5515fa58c962b890/mlir/cmake/modules/AddMLIR.cmake#L424-L425).  

`add_circt_public_c_api_library` does `DISABLE_INSTALL` in order to prevent `add_mlir_public_c_api_library` from generating install targets that collide with those generated by `add_circt_library_install` but in doing so skips the install of the object file `obj.${name}`. So we just put that back.

EDIT: headers need to be installed too...


mikeurbach (approved):
The reasoning makes sense to me, and the cmake code matches what is upstream, so if this helps support your use case, this seems reasonable to me.

PR #8079, opened 2025-01-14
[SCFToCalyx] Lower `scf.if` op when it yields nothing

jiahanxie353:
This patch lowers `scf.ifOp` when it has an empty yield/does not return any value.

### DIFF
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -1117,7 +1117,7 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
                            rewriter, forOp.getLoc(), types);
 
       auto directions = addOp.portDirections();
-      // For an add operation, we expect two input ports and one output port
+      // For an add operation, we expect two input ports and one output port.
       SmallVector<Value, 2> opInputPorts;
       Value opOutputPort;
       for (auto dir : enumerate(directions)) {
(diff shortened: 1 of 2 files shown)


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1102,70 +1102,74 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+                                                              groupOp);
+      getState<ComponentLoweringState>().registerEvaluatingGroup(opOutputPort,
+                                                                 groupOp);
+      return success();
+    }
+    if (auto ifOp = dyn_cast<scf::IfOp>(yieldOp->getParentOp()))
+      // Empty yield inside ifOp, essentially a no-op
+      return success();
+    return yieldOp.getOperation()->emitError()
+           << "Unsupported empty yieldOp outside ForOp or IfOp.";
```
cgyurgyik, line 1172:
```suggestion
    return "Unsupported empty yieldOp: "
           << yieldOp.getOperation()->emitError();
```

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1102,70 +1102,74 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
...
+                                       inductionReg, opOutputPort);
+      // Set group as For Loop's "latch" group.
+      getState<ComponentLoweringState>().setForLoopLatchGroup(forOpInterface,
+                                                              groupOp);
+      getState<ComponentLoweringState>().registerEvaluatingGroup(opOutputPort,
+                                                                 groupOp);
+      return success();
+    }
+    if (auto ifOp = dyn_cast<scf::IfOp>(yieldOp->getParentOp()))
+      // Empty yield inside ifOp, essentially a no-op
```
cgyurgyik, line 1169: Nit: here and elsewhere, add periods please.

cgyurgyik (approved):
There's a lot of control flow going on in this method making it difficult to follow. I think it can be simplified / split apart in smaller functions, but that wasn't your doing so feel free to leave it. LGTM.

PR #8084, opened 2025-01-15
[circt-test] Allow tests to filter according to test runners

fabianschuiki:
Interpret the `require_runners` and `exclude_runners` attributes on formal tests as a filter for the potential runners that can execute a test. If the set of required runners is not empty, a runner from this set is picked. Tests for which no runner is available report as "unsupported", for example if none of the available runners match the filter.

### DIFF
--- a/tools/circt-test/circt-test.cpp
+++ b/tools/circt-test/circt-test.cpp
@@ -34,6 +34,7 @@
 #include "mlir/Parser/Parser.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Support/FileUtilities.h"
+#include "mlir/Support/ToolUtilities.h"
 #include "llvm/ADT/ScopeExit.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/FileSystem.h"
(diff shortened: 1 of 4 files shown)


tools/circt-test/circt-test.cpp
```diff
@@ -240,6 +246,21 @@ static StringRef toString(TestKind kind) {
...
+  if (!attr)
+    return;
+  if (auto arrayAttr = dyn_cast<ArrayAttr>(attr)) {
+    for (auto elementAttr : arrayAttr.getValue())
+      if (auto stringAttr = dyn_cast<StringAttr>(elementAttr))
+        names.insert(stringAttr);
+    return;
+  }
+  if (auto stringAttr = dyn_cast<StringAttr>(attr))
+    names.insert(stringAttr);
```
uenoku, line 261: Allowing null/array/string looks a bit too untyped to me. Can we at least reject StringAttr case? Also it would be nice to have a document for parameters.

uenoku (approved):
One nit regarding filter types but otherwise it looks good to me.

tools/circt-test/circt-test.cpp
fabianschuiki: Yeah that's a good point. I'll make this fallible and add some error reporting. I can add the attributes that have some specific meaning to the op definition, and then later on we can maybe move them to the docs where they're easier for users to find?

uenoku: Sounds great, thanks!

Issue #7970, opened 2024-12-11
[CAPI] Windows and Nightly Failure after #7960


seldridge:
I'm going to blamelessly revert #7960 (https://github.com/llvm/circt/commit/14622691f3c63793411a38b3fa874cb6866c9f43). Please re-land this @SpriteOvO, once you've fixed the issues with it.

Here's what I'm seeing on my end from `valgrind`:

```
# valgrind --track-origins=yes /home/schuylere/scratch/repos/github.com/llvm/circt/build-clang-clang++-RelWithDebInfo-on-vg/bin/circt-capi-firrtl-test
==3842745== Memcheck, a memory error detector
==3842745== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==3842745== Using Valgrind-3.21.0 and LibVEX; rerun with -h for copyright info
==3842745== Command: /home/schuylere/scratch/repos/github.com/llvm/circt/build-clang-clang++-RelWithDebInfo-on-vg/bin/circt-capi-firrtl-test
==3842745==
FIRRTL version 4.1.0
circuit ExportTestSimpleModule :
  public module ExportTestSimpleModule : @[- 2:3]
    input in_1 : UInt<32> @[- 2:44]
    input in_2 : UInt<32> @[- 3:44]
    output out : UInt<32> @[- 4:45]

    connect out, and(in_1, in_2) @[- 6:5]
module {
  firrtl.circuit "AnnoTest" attributes {rawAnnotations = [{class = "firrtl.transforms.DontTouchAnnotation", target = "~AnnoTest|AnnoTest>in"}]} {
    firrtl.module @AnnoTest(in %in: !firrtl.uint<32>) {
    }
  }
}
==3842745== Conditional jump or move depends on uninitialised value(s)
==3842745==    at 0x405B30: testTypeDiscriminantsAndQueries (test/CAPI/firrtl.c:300)
==3842745==    by 0x4065C4: main (test/CAPI/firrtl.c:368)
==3842745==  Uninitialised value was created by a stack allocation
==3842745==    at 0x4053BE: testTypeDiscriminantsAndQueries (test/CAPI/firrtl.c:192)
==3842745==
==3842745== Conditional jump or move depends on uninitialised value(s)
==3842745==    at 0x405B7D: testTypeDiscriminantsAndQueries (test/CAPI/firrtl.c:302)
==3842745==    by 0x4065C4: main (test/CAPI/firrtl.c:368)
[...]

PR #8085, opened 2025-01-15
[CI] Enable Windows builds for PRs

SpriteOvO:
As discussed in #7970, try enabling it to detect Windows-only bugs before merging.

### DIFF
--- a/.github/workflows/buildAndTestWindows.yml
+++ b/.github/workflows/buildAndTestWindows.yml
@@ -1,11 +1,11 @@
 name: Windows build and test
 
-# Run on main (after PRs land) to ensure that they aren't heinously breaking the
-# Windows build.
 on:
   push:
     branches:
       - main
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


SpriteOvO:
From the CI results of this PR, the Windows builds only took 3 more minutes than other jobs, which is acceptable IMO.

teqdruid (approved):
They're quick in the average case (thanks to @seldridge for the caching; but, when the LLVM submodule gets updated or the cache gets flushed it'll take _forever_. Still worth it I think, though the informal policy should be that if the build is taking forever that it should be considered optional since most changes do not break this build.

Scutua (approved):

SpriteOvO:
I noticed that almost all of Windows CI jobs rebuild LLVM in the last couple of days, and there seems to be some problem with caching - all these jobs show sccache "No cache found".

I checked https://github.com/llvm/circt/actions/caches, GitHub does warn about exceeding the free 10G quota, and it should evict earlier caches first, but there are multiple Windows caches on the page without any hits. I'm not sure why. 🤔

PR #8087, opened 2025-01-15
[SwitchToIf] Empty yielded result

jiahanxie353:
This patch adds support to transform index switch to nested if-else when the yielded result is empty.


cgyurgyik (commented):
LGTM

PR #8089, opened 2025-01-16
Bump LLVM to ebc7efbab5c58b46f7215d63be6d0208cb588192.

mikeurbach:
This is a relatively straightforward LLVM bump, but there were a few
things updated here.

https://github.com/llvm/llvm-project/commit/abba01a

This updated PointerUnion to deprecate get and is in favor of
llvm::cast and llvm::isa, so this PR updates the relevant callsites.

https://github.com/llvm/llvm-project/commit/f023da1

This updated FloatType to remove the helpers like getF64. FloatType
was previously turned into a TypeInterface, and the helpers are now
static get methods on the concrete FloatType implementations, so this
PR updates FloatType::getF64 to Float64Type::get.

https://github.com/llvm/llvm-project/commit/b39c5cb6

This updated InlinerInterface to add a new
allowSingleBlockOptimization method. We need to define that on our
PrefixInliner, or else we will hit the new assertion.

### DIFF
--- a/include/circt/Dialect/Arc/ArcOps.td
+++ b/include/circt/Dialect/Arc/ArcOps.td
@@ -235,7 +235,7 @@ def StateOp : ArcOp<"state", [
 
     /// Set the callee for this operation.
     void setCalleeFromCallable(mlir::CallInterfaceCallable callee) {
-      (*this)->setAttr(getArcAttrName(), callee.get<mlir::SymbolRefAttr>());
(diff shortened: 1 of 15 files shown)


teqdruid (approved):
Thanks!

mikeurbach:
Will re-run CI after https://github.com/llvm/circt/pull/8090 is confirmed and merged.

PR #8094, opened 2025-01-17
[FIRRTL] LowerLayers: Update innerrefs in extracted verbatim's

dtzSiFive:
This continues to not address lack of global updating, but fixes use of verbatims that are extracted along with their targets.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -478,6 +478,7 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
 
     SmallVector<hw::InnerSymAttr> innerSyms;
     SmallVector<RWProbeOp> rwprobes;
+    SmallVector<sv::VerbatimOp> verbatims;
     auto layerBlockWalkResult = layerBlock.walk([&](Operation *op) {
       // Record any operations inside the layer block which have inner symbols.
       // Theses may have symbol users which need to be updated.
@@ -606,6 +607,9 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
           createInputPort(operand, op->getLoc());
       }
 
+      if (auto verbatim = dyn_cast<sv::VerbatimOp>(op))
+        verbatims.push_back(verbatim);
+
       return WalkResult::advance();
     });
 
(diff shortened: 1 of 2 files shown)


test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -763,3 +763,24 @@ firrtl.circuit "RWTH" {
...
+      sv.verbatim "// node: {{0}}, {{1}}"(%n) : !firrtl.uint<10> {symbols = [#hw.innerNameRef<@Verbatim::@node>]}
+    }
+  }
+}
+// CHECK:        firrtl.module private @Verbatim_ViewLayer() {
+// CHECK-NEXT:     firrtl.constant 1
+// CHECK-NEXT:     %[[N:.+]] = firrtl.node sym @[[SYM:.+]] %
+// CHECK-NEXT:     sv.verbatim 
+// CHECK-SAME:     !firrtl.uint<10> {symbols = [#hw.innerNameRef<@Verbatim_ViewLayer::@[[SYM]]>]}
+// CHECK-NEXT:   }
```
seldridge, line 786: Nit: using a substitution for `@node` seems incorrect as this inner symbol should be preserved? Conversely, _not_ using a substitution block for the resulting `@Verbatim_ViewLayer` seems wrong. I'd invert these to not use a substitution for the former and use a substitution for the latter.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -699,6 +703,18 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
           hw::InnerRefAttr::get(mapped->second.second, targetRef.getName()));
     }
 
+    // Update verbatims that target operations extracted alongside.
+    mlir::AttrTypeReplacer replacer;
```
seldridge, line 707: I wasn't aware of this. Nice.

seldridge (approved):
LGTM

test/Dialect/FIRRTL/lower-layers.mlir
dtzSiFive: Good call, thanks!  I wasn't sure if it was /necessary/ the inner symbol was re-used as much as a happenstance of implementation for simplicity/efficiency.

Inverting sounds good to me, will update!

PR #8096, opened 2025-01-18
[ESI] Snoop op

teqdruid:
The 'snoop' operation allows something to spy on a channel. It simply exposes the internal signals combinationally.

### DIFF
--- a/include/circt/Dialect/ESI/ESIChannels.td
+++ b/include/circt/Dialect/ESI/ESIChannels.td
@@ -82,6 +82,52 @@ def ChannelTypeImpl : ESI_Type<"Channel"> {
                          ::circt::esi::ChannelSignaling::ValidReady, 0);
       }]>,
   ];
+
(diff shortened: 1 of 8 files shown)


teqdruid:
I'm sure I missed a few places where channel users are checked, but I can't make anything fail.

mortbopet:
Why have specialized snoop operations for each signaling standard (`snoop.vr`)?. As opposed to having a `snoop` op which returns an `esi.channel` that you unwrap. Then it's all in the lowerings to properly source/sink that `esi.channel` that is returned by the snoop op, s.t. it doesn't interact with the transactional signals of the snooped link - and, in `lower ESI to HW`, wrt. the `snoop <-> channel` interaction, you'd have only a single case to check.

teqdruid:
The unwrap operations take backpressure as an input signal (argument),
rather than the output (result) we want.

mortbopet (approved):

Issue #8086, opened 2025-01-15
[SCFToCalyx] Issue with lowering `scf::parallel` to Calyx when there is `scf::if` in it


jiahanxie353:
We lower `scf::parallel` in SCFToCalyx by assuming constant loop bounds and strides, which enables us to unroll the parallel loop manually: https://github.com/llvm/circt/pull/7830

The issue is that, if `scf::parallel` contains `scf::if`, whose condition variable depends on some constants and `scf::parallel`'s loop induction variables, those condition variables will be eventually evaluated to constant `true/false` automatically even if we never implement this sort of "canonicalization" pass manually in `SCFToCalyx.cpp` (I guess it's done by some underling MLIR mechanisms?).

For example, if we begin with:
```mlir
func.func() {
    %c0 = arith.constant 0 : index
    scf.parallel (%arg1, %arg2) = (%c0, %c0) to (%c2, %c2) step (%c1, %c1) {
      %1 = arith.remsi %arg2, %c2 : index
      %2 = arith.cmpi slt, %1, %c0 : index
      %3 = scf.if %2 -> (f32) {
        scf.yield ...
      }
    }
}
```
will turn into:
```mlir
func.func() {
    %c0 = arith.constant 0 : index
    scf.parallel (%arg1, %arg2) = (%c0, %c0) to (%c2, %c2) step (%c1, %c1) {
    ^bb0:
      %1 = arith.remsi "value 0", %c2 : index
      %2 = arith.cmpi slt, %1, %c0 : index
      %3 = scf.if %2 -> (f32) {
        scf.yield ...
      }
    ^bb1:
      %1 = arith.remsi "value 1", %c2 : index
      %2 = arith.cmpi slt, %1, %c0 : index
      %3 = scf.if %2 -> (f32) {
        scf.yield ...
      }
    ^bb2:
    ...
    }
}
```
And in each basic block:
```mlir
%1 = arith.remsi "value 0", %c2 : index
%2 = arith.cmpi slt, %1, %c0 : index
```
will be folded to constants automatically, turning the `scf::if` op to:
```mlir
%3 = scf.if true/false {
 ...
}
```

If they were to be lowered to Calyx, it'll become:
```mlir
calyx.if 1'd1/1'd0 {
...
}
```
but it's not allowed by Calyx' grammar: https://github.com/calyxir/calyx/blob/2e18d555ff00339f46c99cc9d8b698ee27ce25ad/calyx-frontend/src/syntax.pest#L348

[...]

PR #8098, opened 2025-01-20
[SCFToCalyx] Wrap blocks with scf::ExecuteRegion when creating the new scf::Parallel

jiahanxie353:
This patch adjusts the current approach to lower `scf::parallel` operation, which used to be illegal because it contains multiple blocks in the body region. Now we use a legal way by wrapping those blocks with `scf::execute_region` so that the newly create `scf::parallel` op contain exactly one block in the body region.


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1929,25 +1941,32 @@ class BuildParGroups : public calyx::FuncOpPartialLoweringPattern {
     SmallVector<int64_t> indices = lbVals;
 
     while (true) {
+      // Create an `scf.execute_region` to wrap each unrolled block since
```
cgyurgyik, line 1944: Should this comment be placed where the execute region is created?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -2043,14 +2062,19 @@ class BuildControl : public calyx::FuncOpPartialLoweringPattern {
...
-        for (auto &innerBlock : parOp.getRegion().getBlocks()) {
-          rewriter.setInsertionPointToEnd(calyxParOp.getBodyBlock());
-          auto seqOp = rewriter.create<calyx::SeqOp>(parOp.getLoc());
-          rewriter.setInsertionPointToEnd(seqOp.getBodyBlock());
-          if (LogicalResult res = scheduleBasicBlock(
-                  rewriter, path, seqOp.getBodyBlock(), &innerBlock);
-              res.failed())
-            return res;
+        for (auto &op : parOp.getBody()->getOperations()) {
+          if (auto execRegion = dyn_cast<scf::ExecuteRegionOp>(&op)) {
```
cgyurgyik, line 2066: Nit: There's 4 levels of nesting here. Can we do some early 'break' statements to reduce this?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -2043,14 +2062,19 @@ class BuildControl : public calyx::FuncOpPartialLoweringPattern {
...
         auto calyxParOp = rewriter.create<calyx::ParOp>(parOp.getLoc());
-        for (auto &innerBlock : parOp.getRegion().getBlocks()) {
-          rewriter.setInsertionPointToEnd(calyxParOp.getBodyBlock());
-          auto seqOp = rewriter.create<calyx::SeqOp>(parOp.getLoc());
-          rewriter.setInsertionPointToEnd(seqOp.getBodyBlock());
-          if (LogicalResult res = scheduleBasicBlock(
-                  rewriter, path, seqOp.getBodyBlock(), &innerBlock);
-              res.failed())
-            return res;
+        for (auto &op : parOp.getBody()->getOperations()) {
```
cgyurgyik, line 2065: Can't we use a walk here?

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: sounds good

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: yeah using `walk` makes more sense and it reduces the nested structure

jiahanxie353:
Thanks for the review, @cgyurgyik ! Merging it as the CI has passed.
Still need to insert an `scf::if` canonicalization pass to fully address https://github.com/llvm/circt/issues/8086

PR #8101, opened 2025-01-21
[RTG][Elaboration] Support fixed and virtual registers

maerhart:


lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -955,6 +1024,8 @@ class Elaborator : public RTGOpVisitor<Elaborator, FailureOr<DeletionKind>> {
 
   // A map from SSA values to a pointer of an interned elaborator value.
   DenseMap<Value, ElaboratorValue> state;
+
+  uint64_t virtualRegisterID = 0;
```
maerhart, line 1028: While not incorrect placing this here because a separate operation will be materialized in separate regions anyway, it might be better to move this to the shared state to be more robust against changes to this pass in the future.

maerhart: There actually are problems when the register is declared in the target instead of the test, so we should move to the shared state

youngar (approved):
LGTM. An idea for the future: we could malloc the storage for VirtualRegister and use the address as the ID.

PR #8103, opened 2025-01-22
[SCFToCalyx] Canonicalize `scf::if` at the end of `BuildParGroup`

jiahanxie353:
This patch canonicalizes `scf::if` op (if there's any) at the end of the `BuildParGroup` pass to fully address: https://github.com/llvm/circt/issues/8086

### DIFF
--- a/test/Conversion/SCFToCalyx/convert_simple.mlir
+++ b/test/Conversion/SCFToCalyx/convert_simple.mlir
@@ -581,3 +581,51 @@ module {
     return
   }
 }
+
+// Test lower scf.parallel when there is a nested scf.if that can be simplified/canonicalized.
+
+// -----
+
+// CHECK:           calyx.control {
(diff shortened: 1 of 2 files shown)


test/Conversion/SCFToCalyx/convert_simple.mlir
```diff
@@ -581,3 +581,51 @@ module {
     return
   }
 }
+
+// Test lower scf.parallel when there is a nested scf.if that can be simplified/canonicalized.
```
cgyurgyik, line 585:
```suggestion
// Test lower scf.parallel when there is a nested scf.if that can be
// canonicalized. See: https://github.com/llvm/circt/issues/8086
```

cgyurgyik (approved):
LGTM!

jiahanxie353:
Thanks for the review @cgyurgyik ! Merging it.

PR #8104, opened 2025-01-22
[Pipeline] Make `reset` signal optional

mortbopet:
In some usecases, having a reset signal to a pipeline may do more harm than good in terms of synthetis/routing overhead. Instead, users can reset pipelines in other ways - e.g. by holder `go` low for `N stages` cycles. This, in turn, will set the pipeline stage enable registers to a deterministic state.

Allow for this by making the top-level `reset` signal optional.

### DIFF
--- a/lib/Conversion/PipelineToHW/PipelineToHW.cpp
+++ b/lib/Conversion/PipelineToHW/PipelineToHW.cpp
@@ -361,12 +361,15 @@ public:
         vInput.replaceAllUsesWith(vArg);
     }
 
-    // Build stage enable register. The enable register is always reset to 0.
-    // The stage enable register takes the previous-stage combinational valid
-    // output and determines whether this stage is active or not in the next
-    // cycle.
-    // A non-stallable stage always registers the incoming enable signal,
-    // whereas other stages register based on the current stall state.
+    // Build stage enable register. The enable register is reset to 0 iff a
+    // reset signal is available. We here rely on the compreg builders, which
+    // accept reset signal/reset value mlir::Value's that are null.
+    //
+    // The stage enable register takes the
+    // previous-stage combinational valid output and determines whether this
+    // stage is active or not in the next cycle. A non-stallable stage always
+    // registers the incoming enable signal, whereas other stages register based
+    // on the current stall state.
     StageKind stageKind = pipeline.getStageKind(stageIndex);
     Value stageEnabled;
     if (stageIndex == 0) {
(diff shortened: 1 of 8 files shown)


lib/Conversion/PipelineToHW/PipelineToHW.cpp
```diff
@@ -361,21 +361,28 @@ class PipelineInlineLowering : public PipelineLowering {
...
     if (stageIndex == 0) {
       stageEnabled = args.enable;
     } else {
       auto stageRegPrefix = getStagePrefix(stageIndex);
       auto enableRegName = (stageRegPrefix.strref() + "_enable").str();
-      hw::ConstantOp enableRegResetVal =
-          builder.create<hw::ConstantOp>(loc, APInt(1, 0, false));
+
+      Value enableRegResetVal;
+      if (args.reset) {
```
teqdruid, line 382: nit: no curly braces needed here.

lib/Dialect/Pipeline/PipelineOps.cpp
```diff
@@ -196,9 +207,15 @@ static ParseResult parsePipelineOp(mlir::OpAsmParser &parser,
   }
 
   Type clkType = seq::ClockType::get(parser.getContext());
-  if (parser.resolveOperand(clockOperand, clkType, result.operands) ||
-      parser.resolveOperand(resetOperand, i1, result.operands) ||
-      parser.resolveOperand(goOperand, i1, result.operands))
+  if (parser.resolveOperand(clockOperand, clkType, result.operands))
+    return failure();
+
+  if (withReset) {
```
teqdruid, line 213: nit: consider merging these two `if` statements.

lib/Dialect/Pipeline/PipelineOps.cpp
```diff
@@ -246,16 +263,17 @@ static void printPipelineOp(OpAsmPrinter &p, TPipelineOp op) {
...
+    p << " ";
   }
 
   // Print the clock, reset, and go.
   printKeywordOperand(p, "clock", op.getClock());
   p << " ";
-  printKeywordOperand(p, "reset", op.getReset());
-  p << " ";
+  if (op.hasReset()) {
+    printKeywordOperand(p, "reset", op.getReset());
```
teqdruid, line 274: Have you tried round-tripping the IR? I suspect there is an asymmetry in printing vs parsing. I think there's a new flag in circt-opt to automatically round-trip the IR.

teqdruid (approved):

lib/Dialect/Pipeline/PipelineOps.cpp
teqdruid: this and the one below, that is.

blakep-msft (approved):

lib/Dialect/Pipeline/PipelineOps.cpp
mortbopet: Round tripping seems to work fine; the existing round-trip tests pass successfully, and they've been extended with the reset signal removal.

PR #8107, opened 2025-01-22
[RTG][Elaboration] Support labels

maerhart:


lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -105,10 +105,24 @@ struct VirtualRegister {
...
+  bool operator==(const LabelValue &other) const {
+    return name == other.name && id == other.id;
+  }
+
+  /// The label name. For unique labels, this is just the prefix.
+  StringAttr name;
+
+  /// Standard label declarations always have id=0
+  uint64_t id;
+};
```
maerhart, line 120: Missing a hash_value implementation

lib/Dialect/RTG/Transforms/ElaborationPass.cpp
```diff
@@ -1026,6 +1109,7 @@ class Elaborator : public RTGOpVisitor<Elaborator, FailureOr<DeletionKind>> {
   DenseMap<Value, ElaboratorValue> state;
 
   uint64_t virtualRegisterID = 0;
+  uint64_t uniqueLabelID = 1;
```
maerhart, line 1112: Move to shared state.

youngar: Similar to VirtualRegisters, UniqueLabels could rely on a malloc'd address instead giving each one a unique ID.

youngar (approved):

PR #8109, opened 2025-01-22
[OM] Preserve non-OM operations in OM LinkModules

uenoku:
This PR changes OM LinkModule pass to preserve non-OM operations. Orthogonally we shouldn't use builtin.module for the container of nested hw modules and instead use hw.design in the long term but it should be ok for now. 

h/t @dtzSiFive for original idea to keep non-OM operations in artifact.

### DIFF
--- a/lib/Dialect/OM/Transforms/LinkModules.cpp
+++ b/lib/Dialect/OM/Transforms/LinkModules.cpp
@@ -13,6 +13,7 @@
 #include "circt/Dialect/OM/OMOps.h"
 #include "circt/Dialect/OM/OMPasses.h"
 #include "circt/Support/Namespace.h"
+#include "mlir/IR/Block.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/IR/SymbolTable.h"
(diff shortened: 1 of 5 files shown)


mikeurbach (approved):
Thanks, I am good with this change. As we discussed, this is really about combining multiple builtin.module ops without (yet) getting into "linking" questions. This produces perfectly valid IR, and I've used this patch downstream successfully. This seems like a step towards the ultimate hw.design or whatever we call it, and I don't think doing this now makes it any harder to go there eventually.

dtzSiFive (commented):
Cool!

So we hoist + split the OM bits out of each design/module but keep the originals mostly intact otherwise, modulo a "namespace" attribute? (via shuffling into a buffer `Block`, but regardless).

uenoku:
Yes, `om.namespace` attribute won't be consumed by other than OM LinkModules so I'll just delete it in the follow-up. It is used only when giving a good name to private symbols. The produced IR is more of like just collections of post-emission IR and we could separately parse and concatenate multiple MLIR files in the downsteram tool if we want to do that.  

Probably we might want to rename `om-linker` to different name like `circt-artifact-generator` etc as we use the tool not ony for OM linking.

PR #8110, opened 2025-01-22
[Python] Switch CIRCT Python extension and dialects to nanobind.

mikeurbach:
This follows the general nanobind porting guide and the upstream MLIR changes in the same vein. For the most part, this is a very direct migration accomplished with sed.

There were a couple minor differences.

Nanobind is more strict about implicit casting, which popped up in a couple places:

* In MSFTModule, getNearestFreeInColumn was expecting a CIRCTMSFTPrimitiveType (int), but was actually being passed PrimitiveType enums at the call sites. Nanobind refused to automatically cast enum to int, so it was done manually.
* In SVModule, SVAttributeAttr.get was expecting a bool for emitAsComment, even though it was declared as having a default of None. Nanobind refused to automatically cast None to bool, so it was done manually, with a default of false.

Nanobind also prints enums slightly differently, so one FileCheck test had to be updated for ESI.

Otherwise, the changes were purely mechanical.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -514,11 +514,11 @@ option(CIRCT_BINDINGS_PYTHON_ENABLED "Enables CIRCT Python bindings." OFF)
 if(CIRCT_BINDINGS_PYTHON_ENABLED)
   message(STATUS "CIRCT Python bindings are enabled.")
   set(MLIR_DISABLE_CONFIGURE_PYTHON_DEV_PACKAGES 0)
-  mlir_detect_pybind11_install()
+  mlir_detect_nanobind_install()
   # Prime the search like mlir_configure_python_dev_modules
   find_package(Python3 3.8 COMPONENTS Interpreter Development)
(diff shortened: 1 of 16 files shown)


mikeurbach:
Kicked off wheels here: https://github.com/llvm/circt/actions/runs/12916014564

teqdruid (approved):
LGTM!

Will this work on Windows? We still don't have the integration tests running on the Windows builds (since they still fail). I'll have a go at checking on Windows, though it may take me a few days so feel free to land this. I'm probably the only one who has to care about Windows.

mikeurbach:
I don't know, and don't currently have a Windows environment, but I'm cautiously optimistic, because I think upstream does care about Windows, and it seems stable upstream.

mikeurbach:
Looks like wheels succeeded. Unfortunately, there was basically no change in the size: 20.9MB with nanobind, versus the previous wheels taking 21.0MB with pybind11. I'll take these wheels and see how the performance compares in a downstream tool.

mikeurbach:
I'm gonna go ahead and land this.

PR #8112, opened 2025-01-22
[FIRRTL] Move GC after AssignOutputDirs.

dtzSiFive:
This allows GC to set the output directory information of the SV operations it creates when lowering firrtl.view (instead of other way around).


seldridge (approved):
LGTM

This change scares me. Gotta trust the tests, though. 😬

rwy7 (approved):

PR #8113, opened 2025-01-23
[FIRRTL][Inliner] Don't remove modules with symbol uses

fabianschuiki:
Fix an issue in the FIRRTL inliner where modules would be removed even if some other top-level operations still had references to their symbol name.


youngar (approved):
LGTM!

PR #8114, opened 2025-01-23
[FIRRTL][IMDCE] Don't remove modules with symbol uses

fabianschuiki:
Mark any modules alive that appear as symbol references outside of hierarchical path and instance operations. Prevent modules from being deleted if they are explicitly marked as alive. This fixes an issue in IMDCE where references to modules in `firrtl.formal` or other unknown operations would become invalidated due to the pass deleting the modules despite these references.

The pattern of walking operations' attribute dictionary an looking for symbol references is inspired by what MLIR's `SymbolTable` does internally in `walkSymbolRefs` when looking for symbol uses.


uenoku (approved):
LGTM

PR #8115, opened 2025-01-23
[FIRRTL][IMCP] Overdefine ports of modules with unknown symbol uses

fabianschuiki:
If a module is referenced from an unknown top-level operation, i.e. an operation that is not an `hw.hierpath`, mark the module's inputs as overdefined. IMCP cannot reason about how the module is used by such an unknown operation, and therefore should assume that the operation might instantiate the module and apply arbitrary values to its input.

As an example, the `firrtl.formal` operation may refer to a private module as to be executed as a formal test, applying symbolic values to the module's inputs. While IMCP could simply special-case the `firrtl.formal` operation, it feels cleaner to make the pass defensive in the presence of _any_ operation which it does not explicitly know how to deal with.

### DIFF
--- a/include/circt/Support/InstanceGraph.h
+++ b/include/circt/Support/InstanceGraph.h
@@ -200,11 +200,30 @@ public:
   InstanceGraph(const InstanceGraph &) = delete;
   virtual ~InstanceGraph() = default;
 
-  /// Look up an InstanceGraphNode for a module.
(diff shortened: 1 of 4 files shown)


uenoku (approved):

fabianschuiki:
Landing this since Windows build failure is related to 26d80cf.

PR #8116, opened 2025-01-23
[Pass] Remove unnecessary OperationPass<mlir::ModuleOp>

uenoku:
This commit replaces `OperationPass<mlir::ModuleOp>` with `Pass`.

CombToArith, CombToSMT and LowerComb were restricted to run on `mlir::ModuleOp` but as far as I see they are all implemented with Conversion framework and only perform local mutations (no change in func.func or hw.module). Please correct me if I'm wrong but we can remove the restriction.


maerhart (approved):
LGTM!

dtzSiFive (approved):

PR #8118, opened 2025-01-23
[RTGTest] Add store instructions

maerhart:
Should we support labels as operands to store instructions? (and consequently also load instructions)

Assemblers typically only support them as replacements for both the immediate AND the base register. Otherwise, the label address needs to be small enough to fit in the immediate or the assembler inserts additional instructions to load the label in a register to compute the offset (I think most assemblers just error out here, e.g., GAS does). So `lw ra, data` is fine but `lw ra, data(zero)` is not.

Not supporting them would mean we must manually insert the additional instructions to move a label value into a register and add it to the base register. The front-end could provide a utility function for that. It wouldn't lead to performance-optimal binary but we don't care about runtime performance anyway.


youngar:
Without really knowing more, it sounds like we can't guarantee the label address is small enough for an immediate and I would rather ban something that has a "random" chance of not working.

youngar (approved):

maerhart:
We can guarantee that the label fits into immediate+base register (so the store instruction with label would only have one register plus the label; the assembler inserts additional instructions to first fill the base register and then use that plus the immediate as operands to the actual store.
The other case (where it cannot be guaranteed) isn't really supported by assembler anyway.

So, if we want to support the safe label operand option, we should probably add them as separate operations.

PR #8120, opened 2025-01-23
[FIRRTL] GrandCentral: set iface output_file based on parent.

dtzSiFive:
Simply propagate the "output_file" information entirely to any generated sv.interface created during lowering.

cc #8112 .

No change of behavior for annotation-based views.


rwy7 (approved):
😎

PR #8121, opened 2025-01-23
Add passes to strip OM and Emit dialect ops

fabianschuiki:
Several tools in CIRCT do not process the OM and Emit dialects at all, for example, arcilator, circt-bmc, and circt-lec. Since these tools do a full lowering to LLVM internally, they have to discard all OM and Emit ops at some point in the pipeline. Arcilator currently does this in its StripSV pass; circt-bmc and circt-lec simply error out.

This commit adds a `StripOM` pass to the OM dialect and a `StripEmit` pass to the Emit dialect which remove any OM and Emit ops from the IR, respectively. These passes are added to the arcilator, circt-bmc, and circt-lec pipelines in order to discard OM and Emit from the input.

The commit also includes a few boilerplate changes and alphabetical ordering of dialects and dependencies.

@uenoku Maybe this is useful for `circt-synth` as well?

### DIFF
--- a/lib/Dialect/Emit/Transforms/StripEmit.cpp
+++ b/lib/Dialect/Emit/Transforms/StripEmit.cpp
@@ -1,0 +1,36 @@
+//===- StripEmit.cpp ------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Emit/EmitOps.h"
+#include "circt/Dialect/Emit/EmitPasses.h"
+#include "mlir/Pass/Pass.h"
+#include "llvm/Support/Debug.h"
+
+namespace circt {
+namespace emit {
+#define GEN_PASS_DEF_STRIPEMITPASS
+#include "circt/Dialect/Emit/EmitPasses.h.inc"
+} // namespace emit
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+
+namespace {
+struct StripEmitPass : public emit::impl::StripEmitPassBase<StripEmitPass> {
+  void runOnOperation() override {
+    for (auto &op : llvm::make_early_inc_range(getOperation().getOps())) {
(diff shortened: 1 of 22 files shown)


lib/Dialect/Emit/Transforms/StripEmit.cpp
```diff
@@ -0,0 +1,32 @@
...
+} // namespace emit
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+
+namespace {
+struct StripEmitPass : public emit::impl::StripEmitPassBase<StripEmitPass> {
+  void runOnOperation() override {
+    for (auto &op : llvm::make_early_inc_range(getOperation().getOps()))
```
dtzSiFive, line 27: This strips at one level, would this make sense to be a `walk` instead?

Also, no need to remove attributes (since there aren't any, I think) is that right?

fabianschuiki: Yeah good point. I chose to go with a scan over the top-level since we don't (or even can't according to the HasParent traits) have Emit/OM ops nested within modules, to avoid having to do a full walk of the IR. We could switch this to a walk later if needed, with appropriate skips over modules and other things we know can never contain Emit/OM ops.

uenoku: A symbol on `emit.fragment` is captured in `emit.fragments` attribute in HWModuleOp though it doesn't cause a verification error as it's not proper SymbolUse. I ideally we should delete them but I don't know the best way to deal with this kind of unspecified use of symbols.

fabianschuiki: Oh nice, thanks for pointing that out. If it's a dialect attribute like `emit.fragments`, we could just generally delete them. Or to be more conservative, we could specifically walk `HWModuleOp`s and remove that specific op. WDYT?

uenoku: I think deleting dialect attributes attached as discardable is good thing to do.

fabianschuiki: Done! I specifically only delete `emit.fragments` on HWModuleOp right now to not walk the entire IR for now. Do you think that'll be enough?

mikeurbach (approved):

PR #8122, opened 2025-01-23
[ESI][Runtime] XRT support for host memory access

teqdruid:
Adds support for allocating, flushing, and getting the device pointer on XRT platforms.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/lib/backends/Xrt.cpp
+++ b/lib/Dialect/ESI/runtime/cpp/lib/backends/Xrt.cpp
@@ -51,7 +51,7 @@ XrtAccelerator::connect(Context &ctxt, std::string connectionString) {
 }
 
 struct esi::backends::xrt::XrtAccelerator::Impl {
-  constexpr static char kernel[] = "esi_kernel";
+  constexpr static char kernel_name[] = "esi_kernel";
 
   Impl(std::string xclbin, std::string device_id) {
     if (device_id.empty())
@@ -59,17 +59,32 @@ struct esi::backends::xrt::XrtAccelerator::Impl {
     else
       device = ::xrt::device(device_id);
 
-    auto uuid = device.load_xclbin(xclbin);
-    ip = ::xrt::ip(device, uuid, kernel);
+    // Find memory group for the host.
+    ::xrt::xclbin xcl(xclbin);
+    std::optional<::xrt::xclbin::mem> host_mem;
+    for (auto mem : xcl.get_mems())
+      // The host memory is tagged with "HOST[0]". Memory type is wrong --
+      // reports as DRAM rather than host memory so we can't filter on that.
+      if (mem.get_tag().starts_with("HOST"))
+        host_mem = mem;
+    if (!host_mem)
+      throw std::runtime_error("No host memory found in xclbin");
+    memoryGroup = host_mem->get_index();
+
+    // Load the xclbin and instantiate the IP.
+    auto uuid = device.load_xclbin(xcl);
+    ip = ::xrt::ip(device, uuid, kernel_name);
   }
 
   std::map<std::string, ChannelPort &> requestChannelsFor(AppIDPath,
                                                           const BundleType *) {
-    throw std::runtime_error("XRT does not support channel communication yet");
+    // TODO: Create plugin DMA engines. Instantiate elsewhere.
+    return {};
   }
 
   ::xrt::device device;
   ::xrt::ip ip;
+  int32_t memoryGroup;
 };
 
 /// Construct and connect to a cosim server.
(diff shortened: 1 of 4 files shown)


lib/Dialect/ESI/runtime/cpp/lib/backends/Xrt.cpp
```diff
@@ -51,25 +51,40 @@ XrtAccelerator::connect(Context &ctxt, std::string connectionString) {
...
-    auto uuid = device.load_xclbin(xclbin);
-    ip = ::xrt::ip(device, uuid, kernel);
+    // Find memory group for the host.
+    ::xrt::xclbin xcl(xclbin);
+    std::optional<::xrt::xclbin::mem> host_mem;
+    for (auto mem : xcl.get_mems())
+      // The host memory is tagged with "HOST[0]". Memory type is wrong --
+      // reports as DRAM rather than host memory so we can't filter on that.
+      if (mem.get_tag().starts_with("HOST"))
+        host_mem = mem;
```
mortbopet, line 69: I'd say this kind of multiline for with a comment inside almost requires brackets. ([ref](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements) case 2 and 4).

mortbopet, line 69: Missing `break`? or some kind of handling if multiple `"HOST#` memories are found?

lib/Dialect/ESI/runtime/cpp/tools/esitester.cpp
```diff
@@ -99,63 +104,80 @@ void registerCallbacks(AcceleratorConnection *conn, Accelerator *accel) {
...
-  for (int i = 0; i < 100; ++i) {
-    if (*dataPtr != 0)
-      break;
-    std::this_thread::sleep_for(std::chrono::microseconds(100));
+  if (write) {
+    auto *writeMem = acc->getPorts()
+                         .at(AppID("WriteMem"))
+                         .getAs<services::MMIO::MMIORegion>();
+    *dataPtr = 0;
+    writeMem->write(0, (uint64_t)dataPtr);
```
mortbopet, line 157: Don't use C-style casts.

lib/Dialect/ESI/runtime/cpp/lib/backends/Xrt.cpp
```diff
@@ -100,6 +115,47 @@ class XrtMMIO : public MMIO {
...
+    }
+    virtual void *getPtr() const override { return ptr; }
+    /// On XRT platforms, the pointer which the device sees is different from
+    /// the pointer the user application sees.
+    virtual void *getDevicePtr() const override { return (void *)bo.address(); }
+    virtual std::size_t getSize() const override { return bo.size(); }
+    /// It is recommended to use 'sync' to flush the caches before executing any
+    /// DMA.
+    virtual void flush() override { bo.sync(XCL_BO_SYNC_BO_TO_DEVICE); }
+
```
mortbopet, line 141: Is there any sense in adding 'dirty'-flags to `HostMemRegion` s.t. when a `write` has been performed to it, `HostMem` implementations can check/verify that flush has been called?

I'm just a bit sceptical about the "it's recommended" phrase (although that probably originates from xrt documentation and not you).  I can accept "it's recommended" if it's purely performance related, but if not, it either is required or it isn't.

mortbopet (approved):
Don't know much about the XRT intricacies, so someone else would be better suited to review that. A few other style comments... (and I'll never get over the use of raw pointers and `new` in the ESI runtime used anywhere outside of an immediate wrapping of a smart pointer, but you know that 🤷).

lib/Dialect/ESI/runtime/cpp/lib/backends/Xrt.cpp
teqdruid: Yeah, this probably deserves a check and exception.

lib/Dialect/ESI/runtime/cpp/lib/backends/Xrt.cpp
teqdruid: There isn't a 'write' method on HostMemRegion. Standard use case is that the application writes directly to the host pointer. Avoids a memcpy, which as you well know is a killer on the host. So I have no way of (quickly) determining whether the buffer is dirty.

As for it's recommended... certain platforms will require it, and certain platforms won't. In the interest of applications writing platform-independent code, I wrote that it's recommended. As for Xrt, I found some documentation saying:

```
XRT Native API: xrt::bo::sync()) should be used. Though these API will not do any DMA operation, they are used for Cache Invalidate/Flush as the application works on the Cache memory.
```

so it sounds like it's required. I'll update the comment.

teqdruid: Actually, I'm surprised that it's not required for the Catapult based platforms on x86. But I don't fully understand how this new user-pointer translation support works. It must go through the cache system.

teqdruid: Said documentation: https://xilinx.github.io/XRT/master/html/hm.html

PR #8123, opened 2025-01-23
[ESI][XRT] Hostmem service: read side for XRT platform

teqdruid:
Implements the read side of the hostmem -- AXI4 bridge.


mortbopet (approved):
Python LGTM, but this is obviously highly XRT and AXI specific, so it'd be better if you could get someone more knowledgeable about that to review. Rubber stamping for progress.

PR #8125, opened 2025-01-24
[MooreToCore] Add nested moore.conditional support

AndreyVV-100:
Hi! This code adds correct lowering for nested ternary operators like this:

```verilog
module Mod(input a, input b, output logic [1:0] c);
always_comb
    c = a ? b ? 2'd0 : 2'd1 : 2'd2;
endmodule
```


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1346,6 +1346,9 @@ struct ConditionalOpConversion : public OpConversionPattern<ConditionalOp> {
               !memOp.hasEffect<MemoryEffects::Free>())
             return WalkResult::advance();
 
+        if (operation->hasTrait<OpTrait::HasRecursiveMemoryEffects>())
+          return WalkResult::advance();
```
fabianschuiki, line 1350: Nice one!

test/Conversion/MooreToCore/basic.mlir
```diff
@@ -982,6 +982,51 @@ moore.module @StringConstant() {
...
+      %6 = moore.read %b_1 : <l1>
+      %7 = moore.conditional %6 : l1 -> l2 {
+        moore.yield %2 : l2
+      } {
+        moore.yield %1 : l2
+      }
+      moore.yield %7 : l2
+    } {
+      moore.yield %0 : l2
+    }
```
fabianschuiki, line 1020: This test uses quite a few other supporting operations (module, nets, variables, procedure, blocking assigns) to test the conditional lowering. Would it be possible to move just the conditionals and the corresponding checks into a `func.func`, similar to `@Conversions` just below? It might be possible to even move it inside of `@Conversions`. That would make the test less brittle if any of the module and procedure lowerings change, which are unrelated to the conditionals.

AndreyVV-100: Hi! I simplified test and rebased branch. Could you check it?

fabianschuiki (approved):
This looks great! Thanks for tackling this :+1:

AndreyVV-100:
If there are no remarks, could you merge it? I haven't commit access.

fabianschuiki:
Sure thing, thanks!

PR #8130, opened 2025-01-25
[CombToAIG] Add support for div/mod operations

uenoku:
This patch adds support for lowering comb div/mod operations to AIG:
* Support unsigned div/mod for power-of-2 constant divisors 
* Support for non-power-of-2 divisors by emulating the operation with a mux tree when the number of unknown bits is small (default threshold is 10 bits). This is not ideal and eventually we should lower to a proper div/mod. Practically this lowering pattern covers many cases since users rarely use div/mod because these operations are very expensive (users write their own div/mod in their frontend language).  
* LEC tests are added.

### DIFF
--- a/lib/Conversion/CombToAIG/CombToAIG.cpp
+++ b/lib/Conversion/CombToAIG/CombToAIG.cpp
@@ -16,6 +16,7 @@
 #include "circt/Dialect/HW/HWOps.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/DialectConversion.h"
+#include "llvm/ADT/PointerUnion.h"
 
 namespace circt {
 #define GEN_PASS_DEF_CONVERTCOMBTOAIG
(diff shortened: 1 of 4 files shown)


lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -90,6 +91,150 @@ static Value createShiftLogic(ConversionPatternRewriter &rewriter, Location loc,
...
+    return failure();
+
+  int64_t totalUnknownBits = numLhsUnknownBits + numRhsUnknownBits;
+  if (totalUnknownBits > maxEmulationUnknownBits)
+    return failure();
+
+  SmallVector<Value> emulatedResults;
+  emulatedResults.reserve(1 << totalUnknownBits);
+
+  // All zero value used for division by zero.
```
uenoku, line 188:
```suggestion
```

lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -360,6 +505,103 @@ struct CombMulOpConversion : OpConversionPattern<MulOp> {
...
+        Value lowerBits = rewriter.createOrFold<comb::ExtractOp>(
+            op.getLoc(), adaptor.getLhs(), 0, extractAmount);
+        Value constZero = rewriter.create<hw::ConstantOp>(
+            op.getLoc(), APInt::getZero(width - extractAmount));
+        rewriter.replaceOpWithNewOp<comb::ConcatOp>(
+            op, op.getType(), ArrayRef<Value>{constZero, lowerBits});
+        return success();
+      }
+
+    // Non-power-of-2 divisor is not trivial so currently not supported.
```
uenoku, line 569:
```suggestion
```

fabianschuiki (approved):
LGTM! I agree that people will mostly build their own dividers, since there are so many different architectures. In the long run there might be a "simple" long division that AIG could lower to.

PR #8132, opened 2025-01-25
[MemoryBanking] Adjust default dimension

jiahanxie353:
This patch adjusts the default dimension for banking. It used be the outer most dimension (=0), but I think it makes more sense to default to the innermost dimension. For example, if we have:
```mlir
memref.global @__constant_4x2_i32 [[1, 2], [3, 4], [5, 6], [7, 8]] : memref<4x2xi32>
```
if we bank with factor = 2 in a round-robin way, it makes more sense to be:
```mlir
memref.global @__constant_4x2_i32_bank0 [[1], [3], [5], [7]] : memref<4x1xi32>
memref.global @__constant_4x2_i32_bank1 [[2], [4], [6], [8]] : memref<4x1xi32>
```
since it'll preserve the same order as the banking result after flattening to a 1-d array.

Additionally, we further require it to be the innermost dimension that has _shape not equals to 1_. It's not uncommon to have memrefs like:
```mlir
memref.global @__constant_4x1_i32 [[1], [2], [3], [4]] : memref<4x1xi32>
```
if we have a case above, the banked results are:
```mlir
memref.global @__constant_2x1_i32_bank0 [[1], [3]] : memref<2x1xi32>
memref.global @__constant_2x1_i32_bank1 [[2], [4]] : memref<2x1xi32>
```

The story behind this requirement is we can only specify a single `banking-factor` and `dimension` via `circt-opt` command line option for all `memref`s across the program. It is possible that all multi-dim `memref`s have all of their dimensions' shapes > 1 but one or two with shapes like `memref<4x1>`. So instead of having those few special `memref`s disable the entire transformation pass by triggering the assertion failure: https://github.com/llvm/circt/blob/0cea50e9e7d28044485a0e9547de013049b70ecd/lib/Transforms/MemoryBanking.cpp#L82, we adjust the default bahavior.

I hope this understanding makes sense.

### DIFF
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -44,7 +44,7 @@ struct MemoryBankingPass
   MemoryBankingPass(const MemoryBankingPass &other) = default;
   explicit MemoryBankingPass(
       std::optional<unsigned> bankingFactor = std::nullopt,
-      std::optional<unsigned> bankingDimension = std::nullopt) {}
+      std::optional<int> bankingDimension = std::nullopt) {}
 
   void runOnOperation() override;
 
@@ -200,30 +200,41 @@ SmallVector<Value, 4> handleGetGlobalOp(memref::GetGlobalOp getGlobalOp,
   return banks;
 }
 
+std::optional<int> getBankingDimension(std::optional<int> bankingDimensionOpt,
+                                       int64_t rank, ArrayRef<int64_t> shape) {
+  // If the banking dimension is already specified, return it.
+  // Note, the banking dimension will always be nonempty because TableGen will
+  // assign it with a default value -1 if it's not specified by the user. Thus,
+  // -1 is the sentinel value to indicate the default behavior, which is the
+  // innermost dimension with shape greater than 1.
+  if (bankingDimensionOpt.has_value() && *bankingDimensionOpt >= 0) {
+    return bankingDimensionOpt;
+  }
+
+  // Otherwise, find the innermost dimension with size > 1.
+  // For example, [[1], [2], [3], [4]] with `bankingFactor`=2 will be banked to
+  // [[1], [3]] and [[2], [4]].
+  for (int dim = rank - 1; dim >= 0; --dim) {
+    if (shape[dim] > 1) {
+      return dim;
+    }
+  }
+
+  return std::nullopt;
+}
+
 SmallVector<Value, 4> createBanks(Value originalMem, uint64_t bankingFactor,
-                                  std::optional<unsigned> bankingDimensionOpt) {
+                                  std::optional<int> bankingDimensionOpt) {
   MemRefType originalMemRefType = cast<MemRefType>(originalMem.getType());
   unsigned rank = originalMemRefType.getRank();
   ArrayRef<int64_t> shape = originalMemRefType.getShape();
 
-  // Get the banking dimension if specified; use the innermost dimension that
-  // has size greater than one if not.
-  unsigned bankingDimension;
-  if (!bankingDimensionOpt.has_value()) {
-    bankingDimension = rank;
-    for (int dim = rank - 1; dim >= 0; --dim) {
-      if (shape[dim] > 1) {
-        bankingDimension = dim;
-        break;
-      }
-    }
-  } else {
-    bankingDimension = *bankingDimensionOpt;
-  }
-  assert(bankingDimension < rank && "No eligible dimension for banking");
+  auto bankingDimension = getBankingDimension(bankingDimensionOpt, rank, shape);
+  assert(bankingDimension.has_value() && bankingDimension >= 0 &&
+         "No eligible dimension for banking");
 
   MemRefType newMemRefType = computeBankedMemRefType(
-      originalMemRefType, bankingFactor, bankingDimension);
+      originalMemRefType, bankingFactor, *bankingDimension);
   SmallVector<Value, 4> banks;
   if (auto blockArgMem = dyn_cast<BlockArgument>(originalMem)) {
     Block *block = blockArgMem.getOwner();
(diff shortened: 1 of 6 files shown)


lib/Transforms/MemoryBanking.cpp
```diff
@@ -201,8 +201,27 @@ SmallVector<Value, 4> handleGetGlobalOp(memref::GetGlobalOp getGlobalOp,
...
 SmallVector<Value, 4> createBanks(Value originalMem, uint64_t bankingFactor,
-                                  unsigned bankingDimension) {
+                                  std::optional<unsigned> bankingDimensionOpt) {
   MemRefType originalMemRefType = cast<MemRefType>(originalMem.getType());
+  unsigned rank = originalMemRefType.getRank();
+  ArrayRef<int64_t> shape = originalMemRefType.getShape();
+
+  // Get the banking dimension if specified; use the innermost dimension that
+  // has size greater than one if not.
+  unsigned bankingDimension;
```
cgyurgyik, line 211: Its not obvious to me this will always be assigned a value. Can you provide it some default value, or make this optional and assert it has a value?

lib/Transforms/MemoryBanking.cpp
```diff
@@ -509,7 +528,11 @@ void MemoryBankingPass::runOnOperation() {
       auto [it, inserted] =
           memoryToBanks.insert(std::make_pair(memrefVal, SmallVector<Value>{}));
       if (inserted)
-        it->second = createBanks(memrefVal, bankingFactor, bankingDimension);
+        it->second = createBanks(
+            memrefVal, bankingFactor,
+            bankingDimension.hasValue()
+                ? std::optional<unsigned>(bankingDimension.getValue())
```
cgyurgyik, line 534: Why can't this just be `bankingDimension`?

cgyurgyik:
> since it'll preserve the same order as the banking result after flattening to a 1-d array.

Can you elaborate on this? I'm not following why one makes more sense than the other.

> Additionally, we further require it to be the innermost dimension that has shape not equals to 1. It's not uncommon to have memrefs like:

Can you document this somewhere please? This seems important for a user to know.

> The story behind this requirement is we can only specify a single banking-factor and dimension via circt-opt command line option for all memrefs across the program

IMO a single banking factor for all memrefs seems like it is going to be a pretty big limitation, but I guess if it works fine for your experiments currently, then we can leave it for now.

jiahanxie353:
> > since it'll preserve the same order as the banking result after flattening to a 1-d array.
> 
> Can you elaborate on this? I'm not following why one makes more sense than the other.

Sure! Suppose we have a 2d memory: `[[1, 2], [3, 4]]`. If `banking-dimension` is not specified by the user, it's default to 0, i.e., the outer most one. So the banking results are two `1x2` `memref`s: `[1, 2]` and `[3,4]`; and my argument is that it makes less sense than the banking results of this patch's proposal: two `2x1` `memref`s - `[[1], [3]]`, and `[[2], [4]]`.

Why did I mention "after flattening to a 1-d array"? Because we all know that at the end of the day, all multi-dim memories is just a disguise, `[[1, 2], [3, 4]]` is actually `[1, 2, 3, 4]` after "flattening". And if we were to bank `[1, 2, 3, 4]`, the results are two 1-d memories with rank=1 and shape=2: `[1, 3]` and, `[2, 4]`. Therefore, the banking result is this patch's proposal aligns better with what would actually happen if we view all memories as a long, contiguous array.


> > Additionally, we further require it to be the innermost dimension that has shape not equals to 1. It's not uncommon to have memrefs like:
> 
> Can you document this somewhere please? This seems important for a user to know.

Sure, sounds good!

> > The story behind this requirement is we can only specify a single banking-factor and dimension via circt-opt command line option for all memrefs across the program
> 
> IMO a single banking factor for all memrefs seems like it is going to be a pretty big limitation, but I guess if it works fine for your experiments currently, then we can leave it for now.

I agree it is. It is working fine now; in the future, I think a possible way to address this is to have the command line arguments as default values; moreover, we can let user attaching attributes to the operations; those attributes can overwrite the default values.

cgyurgyik:
Makes sense, thanks for the clarification.

lib/Transforms/MemoryBanking.cpp
jiahanxie353: sounds good, done!

lib/Transforms/MemoryBanking.cpp
jiahanxie353: indeed it can be!

lib/Transforms/MemoryBanking.cpp
```diff
@@ -200,11 +200,41 @@ SmallVector<Value, 4> handleGetGlobalOp(memref::GetGlobalOp getGlobalOp,
...
 SmallVector<Value, 4> createBanks(Value originalMem, uint64_t bankingFactor,
-                                  unsigned bankingDimension) {
+                                  std::optional<int> bankingDimensionOpt) {
   MemRefType originalMemRefType = cast<MemRefType>(originalMem.getType());
+  unsigned rank = originalMemRefType.getRank();
+  ArrayRef<int64_t> shape = originalMemRefType.getShape();
+
+  auto bankingDimension = getBankingDimension(bankingDimensionOpt, rank, shape);
+  assert(bankingDimension.has_value() && bankingDimension >= 0 &&
+         "No eligible dimension for banking");
```
cgyurgyik, line 234: Suggestion: Instead of repeating this 3 times, should we just place it directly in the `getBankingDimension` function?

cgyurgyik (approved):

lib/Transforms/MemoryBanking.cpp
jiahanxie353: indeed, thanks!

PR #8133, opened 2025-01-27
[MemoryBanking] Operation-granularity banking configurations by attaching attributes

jiahanxie353:
This patch supports operation-granularity banking configurations by attaching attributes to overwrite the command line argument.

### DIFF
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -71,16 +71,21 @@ DenseSet<Value> collectMemRefs(mlir::affine::AffineParallelOp parOp) {
   return memrefVals;
 }
 
-MemRefType computeBankedMemRefType(MemRefType originalType,
-                                   uint64_t bankingFactor,
-                                   unsigned bankingDimension) {
-  ArrayRef<int64_t> originalShape = originalType.getShape();
+// Verify the banking scheme with different conditions.
+void verifyBankingConditions(unsigned bankingDimension, unsigned bankingFactor,
+                             MemRefType originalType) {
+  auto originalShape = originalType.getShape();
   assert(!originalShape.empty() && "memref shape should not be empty");
-
   assert(bankingDimension < originalType.getRank() &&
          "dimension must be within the memref rank");
   assert(originalShape[bankingDimension] % bankingFactor == 0 &&
          "memref shape must be evenly divided by the banking factor");
+}
+
+MemRefType computeBankedMemRefType(MemRefType originalType,
+                                   uint64_t bankingFactor,
+                                   unsigned bankingDimension) {
+  ArrayRef<int64_t> originalShape = originalType.getShape();
   SmallVector<int64_t, 4> newShape(originalShape.begin(), originalShape.end());
   newShape[bankingDimension] /= bankingFactor;
   MemRefType newMemRefType =
@@ -200,8 +205,9 @@ SmallVector<Value, 4> handleGetGlobalOp(memref::GetGlobalOp getGlobalOp,
   return banks;
 }
 
-unsigned getBankingDimension(std::optional<int> bankingDimensionOpt,
-                             int64_t rank, ArrayRef<int64_t> shape) {
+unsigned getSpecifiedOrDefaultBankingDim(std::optional<int> bankingDimensionOpt,
+                                         int64_t rank,
+                                         ArrayRef<int64_t> shape) {
   // If the banking dimension is already specified, return it.
   // Note, the banking dimension will always be nonempty because TableGen will
   // assign it with a default value -1 if it's not specified by the user. Thus,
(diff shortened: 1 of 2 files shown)


test/Transforms/memory_banking_attrs.mlir
```diff
@@ -0,0 +1,123 @@
+// RUN: circt-opt %s -memory-banking="factor=3 dimension=1" | FileCheck %s --check-prefix ATTRIBUTE
```
jiahanxie353, line 1: Having multiple banking configurations in the same file just complicates everything and I had to make a new test file for it..

lib/Transforms/MemoryBanking.cpp
```diff
@@ -71,16 +71,21 @@ DenseSet<Value> collectMemRefs(mlir::affine::AffineParallelOp parOp) {
...
 }
 
-MemRefType computeBankedMemRefType(MemRefType originalType,
-                                   uint64_t bankingFactor,
-                                   unsigned bankingDimension) {
-  ArrayRef<int64_t> originalShape = originalType.getShape();
+// Verify the banking scheme with different conditions.
+void verifyBankingConditions(unsigned bankingDimension, unsigned bankingFactor,
+                             MemRefType originalType) {
+  auto originalShape = originalType.getShape();
```
cgyurgyik, line 78: Suppppper nit: any reason why you got rid of the type here
```suggestion
  ArrayRef<int64_t> originalShape = originalType.getShape();
```

lib/Transforms/MemoryBanking.cpp
```diff
@@ -226,13 +233,109 @@ unsigned getBankingDimension(std::optional<int> bankingDimensionOpt,
...
+  } else if (isa<BlockArgument>(originalMem)) {
+    auto blockArg = cast<BlockArgument>(originalMem);
+    auto *parentOp = blockArg.getOwner()->getParentOp();
+
+    auto funcOp = dyn_cast<func::FuncOp>(parentOp);
+    assert(funcOp &&
+           "Expected the original memory to be a FuncOp block argument!");
+
+    unsigned argIndex = blockArg.getArgNumber();
+
```
cgyurgyik, line 256:
```suggestion
```
Remove unnecessary empty line

lib/Transforms/MemoryBanking.cpp
```diff
@@ -226,13 +233,109 @@ unsigned getBankingDimension(std::optional<int> bankingDimensionOpt,
...
+void resolveBankingAttributes(Value originalMem, unsigned &bankingFactor,
+                              unsigned &bankingDimension) {
+  if (auto *originalDef = originalMem.getDefiningOp()) {
+    if (auto attrFactor = dyn_cast_if_present<IntegerAttr>(
+            originalDef->getAttr("banking.factor")))
+      bankingFactor = attrFactor.getInt();
+    if (auto attrDimension = dyn_cast_if_present<IntegerAttr>(
+            originalDef->getAttr("banking.dimension")))
+      bankingDimension = attrDimension.getInt();
+  } else if (isa<BlockArgument>(originalMem)) {
```
cgyurgyik, line 247: Suggestion: What about we just have a `return;` here instead of `else if`? 

if (auto* A = ...) {
  ...
  return;
}

if (auto* B = ...) {
  ...
  return;
}

Just reads slightly cleaner to me, but this is personal preference so do as you please.

lib/Transforms/MemoryBanking.cpp
```diff
@@ -355,8 +475,8 @@ struct BankAffineLoadPattern
   }
 
 private:
-  uint64_t bankingFactor;
-  std::optional<int> bankingDimensionOpt;
+  mutable unsigned bankingFactor;
```
cgyurgyik, line 478: Why is it necessary that these are mutable?

lib/Transforms/MemoryBanking.cpp
```diff
@@ -517,8 +647,22 @@ LogicalResult cleanUpOldMemRefs(DenseSet<Value> &oldMemRefVals,
...
   }
 
-  // Modify the function type accordingly
+  // Modify the function argument attributes and function type accordingly
   for (auto funcOp : funcsToModify) {
+    ArrayAttr existingArgAttrs = funcOp->getAttrOfType<ArrayAttr>("arg_attrs");
+    if (existingArgAttrs) {
+      SmallVector<Attribute, 4> updatedArgAttrs;
+      DenseSet<unsigned> indicesToErase(erasedArgIndices[funcOp].begin(),
+                                        erasedArgIndices[funcOp].end());
```
cgyurgyik, line 656:
```suggestion
      auto erasedIndices = erasedArgIndices[funcOp];
      DenseSet<unsigned> indicesToErase(erasedIndices.begin(),
                                        erasedIndices.end());
```

test/Transforms/memory_banking_attrs.mlir
cgyurgyik: Makes sense. I will note that this is a rather large test example. Can this be made into a smaller example? I'll leave this to your discretion.

cgyurgyik, line 1: Curiously, how come `--check-prefix ATTRIBUTE`?

lib/Transforms/MemoryBanking.cpp
jiahanxie353: no absolutely, sounds good to me as well!

lib/Transforms/MemoryBanking.cpp
jiahanxie353: Because `matchAndOverwrite` has `const` in its signature: https://github.com/llvm/circt/blob/335e50ce67b802b20ea960578e37f4950df233e5/lib/Transforms/MemoryBanking.cpp#L295

test/Transforms/memory_banking_attrs.mlir
jiahanxie353: ahh because I was trying to stuff everything in the `memory_banking_multi_dim` test file. It didn't work and I simply just copied from there. Now that we are only testing this example for now, I agree that we do not need prefix checking for splitted cases testing.

jiahanxie353: I decided to keep this large example just to demonstrate a full working example with different configurations

cgyurgyik (approved):

cgyurgyik:
Nice!

jiahanxie353:
Merging it as the CI has passed, thank you so much for the review! @cgyurgyik

PR #8134, opened 2025-01-28
[HW] Start of high imedence operations and types

darthscsi:
Separate high impedence from the rest of the system.  This will lead to a great simplification of inout and hw.

### DIFF
--- a/include/circt/Dialect/HW/HWMiscOps.td
+++ b/include/circt/Dialect/HW/HWMiscOps.td
@@ -189,4 +189,53 @@ def EnumCmpOp : HWOp<"enum.cmp", [Pure]> {
   }];
 }
 
+
+
+// Deal with HiZ types
+
+def ReadHiZOp
+ : HWOp<"hiZ.sample", [Pure, HiZTypeConstraint<"result", "input">]> {
+  let summary = "Get the value of a high impediance type (hiZ).";
+  let description = [{
+    Extract the value of a high impedance typed value.  This converts a hiZ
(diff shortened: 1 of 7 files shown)


include/circt/Dialect/HW/HWMiscOps.td
```diff
@@ -189,4 +189,53 @@ def EnumCmpOp : HWOp<"enum.cmp", [Pure]> {
...
+        [Pure, HiZTypeConstraint<"value", "input">]> {
+  let summary = "Drive a value on a high impediance type (hiZ).";
+  let description = [{
+    Conditionally drive a value on a high impedance bus.  This converts a value
+    to hiZ type and asserts that value on the bus.  This operation doesn't make 
+    guarantees about the bus at runtime.  It is possible for multiple drivers
+    to drive the bus.  This behavior is resolved by simulation or physics.
+  }];
+
+  let arguments = (ins HiZType:$input, HWValueType: $value, I1:$cond);
```
fzi-hielscher, line 222: From my experience, most tristate busses (QSPI, I2C, GPIO, etc.) will need independent direction control for each bit. Could we give `$cond` the same type as `$value`? This would make verilog lowering more complex though.
Alternatively, if we had some way of concatenating HiZ types, we also would not have to plumb all lines out individually.

fzi-hielscher (commented):
Nice!

I'd also suggest adding `HiZType` to the list of movable declarations in `PrepareForEmission.cpp:568`. Otherwise I expect crashes for non-dominated uses of `hw.hiZ.create`.

PR #8139, opened 2025-01-29
[RTG] Populate pipeline with new passes

maerhart:
Add the new passes to the pipeline and pass their options through to the frontend. Also fixes a bug with the existing unsupported instructions option.


fabianschuiki (approved):
I'm not fully up to speed on the RTG details, but this looks good as far as I can tell.

PR #8142, opened 2025-01-29
[RTGTest] Add the last remaining ops for RV32I

maerhart:


include/circt/Dialect/RTGTest/IR/RTGTestAttributes.td
```diff
@@ -71,6 +71,7 @@ class ImmediateAttrBase<int width> : RTGTestAttrDef<"Imm" # width, [
   let genVerifyDecl = 1;
 }
 
+def Imm5 : ImmediateAttrBase<5>;
```
fabianschuiki, line 74: Mega nit: 
```suggestion
def Imm5  : ImmediateAttrBase<5>;
```

include/circt/Dialect/RTGTest/IR/RTGTestOps.td
```diff
@@ -275,8 +278,192 @@ class InstFormatSOpBase<string mnemonic, int opcode7, int funct3>
...
+                                           FoldAdaptor adaptor) {
+      auto rd = cast<rtg::RegisterAttrInterface>(adaptor.getRd());
+      auto rs = cast<rtg::RegisterAttrInterface>(adaptor.getRs());
+      auto imm = cast<Imm12Attr>(adaptor.getImm()).getAPInt();
+
+      auto binary = imm
+      .concat(llvm::APInt(5, rs.getClassIndex()))
+      .concat(llvm::APInt(3, }] # funct3 # [{))
+      .concat(llvm::APInt(5, rd.getClassIndex()))
+      .concat(llvm::APInt(7, }] # opcode7 # [{));
```
fabianschuiki, line 394: This _might_ be easier to read with a TableGen `subst` call around the code snippet substituting something like `$funct3`. But then again, this is pretty clean, and there are only two values to be interpolated. The substitution might make it harder to read.

include/circt/Dialect/RTGTest/IR/RTGTestTypes.td
```diff
@@ -38,6 +38,7 @@ class ImmTypeBase<int width> : TypeDef<RTGTestDialect, "Imm" # width, []> {
   let mnemonic = "imm" # width;
 }
 
+def Imm5Type : ImmTypeBase<5>;
```
fabianschuiki, line 41: Bombastic nit: 
```suggestion
def Imm5Type  : ImmTypeBase<5>;
```

fabianschuiki (approved):
LGTM

include/circt/Dialect/RTGTest/IR/RTGTestOps.td
maerhart: Doing `!subst("$funct3", funct3, "...")` leads to a cryptic tablegen error saying that the initializer could not be resolved. Maybe because `funct3` is an integer and not a string? Working around all of that might make it more complicated.

PR #8143, opened 2025-01-29
[ExportVerilog] Treat verif.contract as no-op

fabianschuiki:
Ignore `verif.contract` operations during Verilog emission and simply pass their operands through to the results. This is always a valid lowering of a contract.


dtzSiFive (approved):
LGTM

Can't speak for the appropriateness of what this doing (just lack context) but seems good here!

PR #8144, opened 2025-01-29
[Verif] Add StripContracts pass

fabianschuiki:
Add a simple pass that removes all contracts from the IR, treating them as passthroughs. Run the pass in firtool's HW-to-SV pipeline early on to add additional optimization opportunities. ExportVerilog will eventually ignore contracts anyway.

### DIFF
--- a/lib/Dialect/Verif/Transforms/StripContracts.cpp
+++ b/lib/Dialect/Verif/Transforms/StripContracts.cpp
@@ -1,0 +1,44 @@
+//===- StripContracts.cpp -------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Verif/VerifOps.h"
+#include "circt/Dialect/Verif/VerifPasses.h"
+#include "mlir/Pass/Pass.h"
+
+namespace circt {
+namespace verif {
+#define GEN_PASS_DEF_STRIPCONTRACTSPASS
+#include "circt/Dialect/Verif/Passes.h.inc"
+} // namespace verif
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+using namespace verif;
+
+namespace {
+struct StripContractsPass
+    : public verif::impl::StripContractsPassBase<StripContractsPass> {
+  void runOnOperation() override {
+    getOperation()->walk([](ContractOp op) {
+      op->replaceUsesWithIf(op.getInputs(), [&](OpOperand &operand) {
+        return operand.getOwner() != op;
+      });
+      // Prevent removal of self-referential contracts that have other users.
+      // Removing them would result in invalid IR.
+      for (auto operand : op->getOperands())
+        if (operand.getDefiningOp() == op)
+          for (auto *user : operand.getUsers())
+            if (user != op)
+              return;
+      op->dropAllReferences();
+      op->erase();
+    });
+  }
+};
+} // namespace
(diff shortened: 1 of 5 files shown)


lib/Dialect/Verif/Transforms/StripContracts.cpp
```diff
@@ -0,0 +1,35 @@
...
+using namespace verif;
+
+namespace {
+struct StripContractsPass
+    : public verif::impl::StripContractsPassBase<StripContractsPass> {
+  void runOnOperation() override {
+    getOperation()->walk([](ContractOp op) {
+      op.replaceAllUsesWith(op.getInputs());
+      op.erase();
+    });
```
maerhart, line 32: Pretty degenerate case but does this work in the following case?
```
%0 = verif.contract %1 : i32 {}
%1 = verif.contract %0 : i32 {}
```

fabianschuiki: Good thinking. Should work. Let me add this to the sets.

fabianschuiki: Actually needed a tweak to avoid an assert in `replaceAllUsesWith` and to allow `op.erase` to work if the operation is cycle.

lib/Firtool/Firtool.cpp
```diff
@@ -304,6 +304,7 @@ LogicalResult firtool::populateLowFIRRTLToHW(mlir::PassManager &pm,
 
 LogicalResult firtool::populateHWToSV(mlir::PassManager &pm,
                                       const FirtoolOptions &opt) {
+  pm.addPass(verif::createStripContractsPass());
```
uenoku, line 307: Could you nest to any operation to reduce overhead?

lib/Dialect/Verif/Transforms/StripContracts.cpp
```diff
@@ -0,0 +1,38 @@
...
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Verif/VerifOps.h"
+#include "circt/Dialect/Verif/VerifPasses.h"
+#include "mlir/Pass/Pass.h"
+#include "llvm/Support/Debug.h"
```
uenoku, line 12: nit: Debug not used.

lib/Firtool/Firtool.cpp
fabianschuiki: Great point! Thanks :-)

fabianschuiki: Done!

test/Dialect/Verif/strip-contracts.mlir
```diff
@@ -0,0 +1,16 @@
...
+  %0 = verif.contract %arg0 : i42 {}
+  // CHECK: return %arg0
+  return %0 : i42
+}
+
+// CHECK-LABEL: hw.module @bar
+hw.module @bar() {
+  // CHECK-NOT: verif.contract
+  %0 = verif.contract %1 : i42 {}
+  %1 = verif.contract %0 : i42 {}
```
maerhart, line 15: What's actually the semantics if we return %1 as an output of the module? Also, the pass would still crash in that case, right? Not really important to fix here though. We have similar issues with comb.

maerhart (approved):

test/Dialect/Verif/strip-contracts.mlir
fabianschuiki: That's a great point, that would still crash. It would be better if it didn't, though. I think in that case we'd want to just not remove the contract. But the semantics are the same as in `%0 = comb.add %0, %0; hw.output %0` -- hard to define. Maybe some `undefined` or `poison` value, or `unknown`.

fabianschuiki: Fixed this for now by just not removing cyclical contracts.

PR #8146, opened 2025-01-30
[RTG] Custom parser/printer for sequence op and type for sequence families

maerhart:
Allow type parameters in the sequence type and attach it as a type attribute to the sequence op. That way ops referring to a sequence don't have to access the operation's body to verify the type.


fabianschuiki (approved):
LGTM! Nice that you no longer have to chase after ops to figure out the type signature.

PR #8147, opened 2025-01-30
[RTG] Separate operation for sequence randomization and randomized sequence type

maerhart:


fabianschuiki (approved):
This split looks good to me after a cursory glance!

PR #8148, opened 2025-01-30
[RTG] Support partial sequence substitutions

maerhart:


fabianschuiki (approved):
This is a neat way of materializing a symbol as an SSA value. I wonder what other uses this pattern might have 🤔

maerhart:
One use-case is the `context_switch` op where you have to pass a sequence with 3 arguments of specific types. If multiple switches have a similar sequence you can create one sequence that has > 3 arguments and partially substitute the first N-3 to specialize for one particular switch instead of copy pasting the whole sequence.

PR #8154, opened 2025-01-30
[FIRRTL] Add require and ensure intrinsics

fabianschuiki:
Add the `circt.verif.require` and `circt.verif.ensure` intrinsics as a way to produce the `verif.require` and `verif.ensure` operations from FIRRTL. At a later point we'll probably want to promote the five verif ops `assert`, `assume`, `cover`, `require`, and `ensure` to the FIRRTL spec and merge the functionality of the `circt.verif.assert` intrinsic with the `firrtl.assert` operation.

Note that this requires dropping of the clock operand from `verif.require` and `verif.ensure`. This operand was never used, so it feels like a reasonable change. At a later point we might want to merge `verif.clocked_assert` into `verif.assert` by adding clock operands to the verif ops in general.

### DIFF
--- a/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
+++ b/docs/Dialects/FIRRTL/FIRRTLIntrinsics.md
@@ -414,3 +414,42 @@ what the field in the bundle is.
 | elements | array  | List of AugmentedFields                                   |
 
 Creates a SystemVerilog interface for each bundle type.
+
+### circt_verif_assert
+
+Asserts that a property holds.
+The property may be an boolean, sequence, or property.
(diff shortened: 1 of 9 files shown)


dtzSiFive (approved):
LGTM! Thanks!

Consider documenting the intrinsics in `docs/Dialects/FIRRTL/FIRRTLIntrinsics.md` .

fabianschuiki:
Intrinsics have been documented 😏

mikeurbach:
Took a quick look, this is awesome.

dtzSiFive (commented):
LGTM!

PR #8155, opened 2025-01-30
[Verif] Add ContractOp to Visitor

fabianschuiki:
Add the `ContractOp` to the `verif::Visitor`. This requires a minor tweak to the HWToBTOR2 pass to handle unknown operations more generally.


maerhart (approved):
LGTM! (if CI passes)

PR #8156, opened 2025-01-30
[FIRRTL] Add contract declaration

fabianschuiki:
Add `contract` declaration support to the FIRRTL parser, and a corresponding operation to the FIRRTL dialect. A contract acts like a `node`, but allows for additional formal properties to be expressed in the contract body. The contract can then either be treated as a node for synthesis, where the formal properties are just discarded, or it can be used to subdivide a larger formal problem using a combination of the assume-guarantee and cutpointing techniques commonly used in formal verification tools.

This has not landed in the FIRRTL spec yet. As soon as that happens, I will replace the `missingSpecFIRVersion` use with the actual spec version.

Actual lowering follows in a separate commit.


mikeurbach (commented):
I haven't reviewed in depth yet, but very excited about this whole series of patches.

dtzSiFive (approved):
LGTM.

Sometimes we have parser check things like the required types (and things like "must be passive") but that's a bit of duplication re:verifier/ODS.

fabianschuiki:
Yeah great point about the checks. I'm pretty sure the contracts will get quite a few additional checks as we hash out the spec details. I'll start poking at stuff like ExpandWhens and friends, since you shouldn't be able to leak values out of a contract body through last connect semantics.

PR #8157, opened 2025-01-31
[FIRRTL] Fix hasDontTouch crash on non-module block args

fabianschuiki:
Fix an issue in `firrtl::hasDontTouch` where the cast to `FModuleOp` would crash if the function was called on a block argument of an operation that is not a module, for example, a `ContractOp`. Such block arguments are considered to not be marked as "don't touch".


uenoku (approved):
This looks reasonable to me.

PR #8158, opened 2025-01-31
[FIRRTL][IMDCE] Support operations with blocks

fabianschuiki:
Avoid a crash in IMDCE when encountering a block argument that is not defined by a module, but some other operation that has nested blocks, such as a contract. Additionally, when handling liveness of generic operations, conservatively mark any nested blocks executable.


uenoku (approved):
LGTM

PR #8159, opened 2025-01-31
[FIRRTL][LowerToHW] Lower contract ops

fabianschuiki:
Lower `firrtl.contract` to `verif.contract` ops. The lowering itself is pretty straightforward, since the ops are basically equivalent. The only exception are block arguments in FIRRTL that get replaced with op results in Verif due to the switch from an SSACFG region with dominance to a graph region.

The block arguments require a slight change to how nested operations are handled by the pass. Currently, a post-order walk over the operations is used. This is problematic since the parent operation does not get an opportunity to lower its block arguments before its child operations are lowered, which may need access to those block arguments. Switching to a pre-order walk does not work, since that wouldn't allow us to modify the operations during the walk, which we obviously do during the lowering. This commit therefore adds a worklist that tracks operation ranges that are yet to be lowered. The worklist allows parent operations to lower themselves and any of their block arguments, and then add the nested regions, blocks, or operations onto the worklist for lowering. It's basically a pre-walk with the ability to mutate the parent operation before children are lowered.

Block arguments are also no longer automatically assumed to be already lowered. This was only valid for modules, since we lower the module ports before we lower the module body. In case of ops like contracts, the block arguments still need to be lowered. To fix this, modules push a 1-1 mapping of their block arguments into the map of lowered values, which allows other ops to provide some other lowering for their block arguments.

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -1580,6 +1580,7 @@ struct FIRRTLLowering : public FIRRTLVisitor<FIRRTLLowering, LogicalResult> {
   LogicalResult visitDecl(MemOp op);
   LogicalResult visitDecl(InstanceOp oldInstance);
   LogicalResult visitDecl(VerbatimWireOp op);
+  LogicalResult visitDecl(ContractOp op);
 
   // Unary Ops.
   LogicalResult lowerNoopCast(Operation *op);
@@ -1837,6 +1838,19 @@ private:
   /// the LTL ops, which were necessary to go from the def-before-use FIRRTL
   /// dialect to the graph-like HW dialect.
   SetVector<Operation *> ltlOpFixupWorklist;
+
+  /// A worklist of operation ranges to be lowered. Parnet operations can push
+  /// their nested operations onto this worklist to be processed after the
+  /// parent operation has handled the region, blocks, and block arguments.
+  SmallVector<std::pair<Block::iterator, Block::iterator>> worklist;
+
+  void addToWorklist(Block &block) {
+    worklist.push_back({block.begin(), block.end()});
+  }
+  void addToWorklist(Region &region) {
+    for (auto &block : llvm::reverse(region))
+      addToWorklist(block);
+  }
 };
 } // end anonymous namespace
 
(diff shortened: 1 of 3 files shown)


lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1847,16 +1861,27 @@ LogicalResult FIRRTLModuleLowering::lowerModuleOperations(
 
 // This is the main entrypoint for the lowering pass.
 LogicalResult FIRRTLLowering::run() {
-  // FIRRTL FModule is a single block because FIRRTL ops are a DAG.  Walk
-  // through each operation, lowering each in turn if we can, introducing
-  // casts if we cannot.
-  auto &body = theModule.getBody();
+  // Mark the module's block arguments are already lowered. This will allow
+  // `getLoweredValue` to return the block arguments as they are.
+  for (auto arg : theModule.getBody().front().getArguments())
```
uenoku, line 1866: nit:does this wokr?
```suggestion
  for (auto arg : theModule.getBodyBlock()->getArguments())
```

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -2724,6 +2741,10 @@ void FIRRTLLowering::addIfProceduralBlock(Value cond,
 ///
 FIRRTLLowering::UnloweredOpResult
 FIRRTLLowering::handleUnloweredOp(Operation *op) {
+  // Push nested operations onto the worklist such that they are lowered.
+  for (auto &region : op->getRegions())
+    addToWorklist(region);
```
uenoku, line 2746: Can you put this under `if (!isa<FIRRTLDialect>(op->getDialect())) {` below and causes an error when there is unhandled FIRRTL operation with a region? I think it's more predictable not to silently lower unhandled FIRRTL operations with a region.

uenoku (approved):
Few nits otherwise looks good to me.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
fabianschuiki: Oh yeah that's a great idea! Let me change that.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
fabianschuiki: Sweet, thanks 😎

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
fabianschuiki: Done 👍

PR #8166, opened 2025-01-31
[circt-test] Add support for contracts

fabianschuiki:
Add the `LowerContracts` or `StripContracts` pass to the circt-test pipeline in order to leverage contracts during formal verification. To make this work, circ-test has to run a preparatory pass pipeline on the input MLIR, and pass the result as input to the MLIR runners instead of the original file.

Also add an `--ir` option to dump the IR after the preparatory passes.


dobios (approved):
LGTM!

PR #8168, opened 2025-02-01
[FIRRTL] LowerXMR: process all modules

youngar:
This changes LowerXMR to process all modules instead of just those that are reachable from the top level module.


seldridge (approved):
LGTM

llvm
youngar, line 1: 👀

dtzSiFive (approved):
LGTM!

Thanks for the fix!

PR #8169, opened 2025-02-01
[firtool] Add an option to emit HW MLIR into file

uenoku:
This PR adds `-output-hw-mlir` option to firtool which emits HW IR into file in a similar way to `-output-final-mlir`. 

This PR adds `DumpIR` pass to simplify the emission. It's currently not exposed to other tool as it's very specific to firtool pipeline (actually the pass indirectly uses cl options defined in firtool.cpp)

### DIFF
--- a/tools/firtool/firtool.cpp
+++ b/tools/firtool/firtool.cpp
@@ -495,7 +495,7 @@ static LogicalResult processBuffer(
                                             (*outputFile)->os())))
         return failure();
 
-    // If requseted, emit the HW IR to hwOutFile.
+    // If requested, emit the HW IR to hwOutFile.
     if (!hwOutFile.empty())
       pm.addPass(std::make_unique<DumpIRPass>(hwOutFile.getValue()));
 
--- a/tools/firtool/firtool.cpp
+++ b/tools/firtool/firtool.cpp
@@ -207,6 +207,12 @@ static cl::list<std::string> inputAnnotationFilenames(
     cl::CommaSeparated, cl::value_desc("filename"), cl::cat(mainCategory));
 
 static cl::opt<std::string>
+    hwOutFile("output-hw-mlir",
+              cl::desc("Optional file name to output the HW IR into, in "
+                       "addition to the output requested by -o"),
+              cl::init(""), cl::value_desc("filename"), cl::cat(mainCategory));
+
+static cl::opt<std::string>
     mlirOutFile("output-final-mlir",
                 cl::desc("Optional file name to output the final MLIR into, in "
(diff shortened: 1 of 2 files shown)


tools/firtool/firtool.cpp
```diff
@@ -456,6 +494,11 @@ static LogicalResult processBuffer(
       if (failed(firtool::populateHWToBTOR2(pm, firtoolOptions,
                                             (*outputFile)->os())))
         return failure();
+
+    // If requseted, emit the HW IR to hwOutFile.
```
dtzSiFive, line 498: *requested

tools/firtool/firtool.cpp
```diff
@@ -336,6 +342,38 @@ struct EmitSplitHGLDDPass
...
+struct DumpIRPass
+    : public PassWrapper<DumpIRPass, OperationPass<mlir::ModuleOp>> {
+  DumpIRPass(const std::string &outputFile)
+      : PassWrapper<DumpIRPass, OperationPass<mlir::ModuleOp>>() {
+    this->outputFile.setValue(outputFile);
+  }
+
+  DumpIRPass(const DumpIRPass &other) : PassWrapper(other) {
+    outputFile.setValue(other.outputFile.getValue());
+  }
```
mikeurbach, line 355: I'm all for re-use here, but I may be missing something. What is the benefit to using a small pass versus a helper function?

mikeurbach (commented):
Generally, excited for this, just one general question.

tools/firtool/firtool.cpp
uenoku: It's necessary to use a pass because it's not possible to get IR after LowerToHW outside of pass pipeline. To use helper function we have to separate pass pipeline but I don't think it's better.

fabianschuiki (approved):
Really excited for this! For any tool or analysis that wants to start from an HW design, this is going to be very valuable!

tools/firtool/firtool.cpp
mikeurbach: Ah I see, makes sense to me. It's a little weird to see a pass definition inside a tool cpp file... it seems fine for this use case, but I was wondering.

sequencer:
Is possible for adding a c-api for this?

uenoku:
> Is possible for adding a c-api for this?

It's a bit tricky to do that because the option is not pipeline option strictly speaking (The option is not added to FirtoolOption intentionally). In CAPI you can manually dump IR after HW lowering if you necessary, and I think it's more flexible. So let me first merge the PR and add CAPI if it turned it's really necessary.

PR #8171, opened 2025-02-01
[LLHD] Move maxPrimitives check before initializing the 'dontCare' table

maerhart:
Fixes a crash if the number of primitives is too big


fabianschuiki (approved):
LGTM

Issue #7611, opened 2024-09-19
[FIRRTL] InferResets FullReset (formerly FART) errors due to "different" reset domains when the reset value is identical 


jackkoenig:
This is related to https://github.com/llvm/circt/issues/4886 in that this is sort of a special case. Fixing #4886 would fix this issue so long as infer-resets is followed by dedup (as it currently is)

Consider the following:
```firrtl
FIRRTL version 4.0.0
; CHECK-LABEL: module test(
circuit top :%[[
  {
    "class":"circt.FullResetAnnotation",
    "target":"~top|parent_1>reset",
    "resetType":"async"
  },
  {
    "class":"circt.FullResetAnnotation",
    "target":"~top|parent_2>reset",
    "resetType":"async"
  }
]]
  module child :
    input clock : Clock
    input in : UInt<8>
    output out : UInt<8>

    reg r : UInt<8>, clock
    connect r, in
    connect out, r

  module parent_1 :
    input clock : Clock
    input reset : AsyncReset
    input in : UInt<8>
    output out : UInt<8>

    inst c of child
    connect c.clock, clock
    connect c.in, in
    connect out, c.out

  module parent_2 :
    input clock : Clock
    input reset : AsyncReset
    input in : UInt<8>
    output out : UInt<8>

    inst c of child
    connect c.clock, clock
    connect c.in, in
    connect out, c.out


  public module top :
    input clock : Clock
    input reset1 : AsyncReset
    input reset2 : AsyncReset
    input in : UInt<8>
    output out : UInt<8>

    inst p1 of parent_1
    connect p1.clock, clock
    connect p1.reset, reset1
    connect p1.in, in

    inst p2 of parent_2
    connect p2.clock, clock
    connect p2.reset, reset2
    connect p2.in, in

    connect out, and(p1.out, p2.out)
```

We have 2 full reset domains, but they are identical (same port name and type). This currently errors with:
```
test.fir:15:3: error: module 'child' instantiated in different reset domains
  module child :
  ^
test.fir:15:3: note: see current operation: 
"firrtl.module"() ({
^bb0(%arg0: !firrtl.clock, %arg1: !firrtl.uint<8>, %arg2: !firrtl.uint<8>):
  %0 = "firrtl.reg"(%arg0) {annotations = [], name = "r", nameKind = #firrtl<name_kind droppable_name>} : (!firrtl.clock) -> !firrtl.uint<8>                                                                       
[...]

Issue #7271, opened 2024-07-02
[FIRRTL] FART: logic to reuse existing port can result in the wrong reset


youngar:
Given this FIRRTL:
```firrtl
FIRRTL version 4.0.0
circuit Foo: %[[
  {"class":"sifive.enterprise.firrtl.FullAsyncResetAnnotation", "target":"~Foo|Foo>r"}
]]
  public module Foo:
    input p : UInt<1>
    input r : AsyncReset
    input c : Clock
    inst bar of Bar
    connect bar.r, asAsyncReset(UInt<1>(0))
    connect bar.c, c

  module Bar:
    input c : Clock
    input r : AsyncReset

    reg reg : UInt<8>, c
```
When compiled with `firtool` with no additional options, we see InferResets uses the port `Bar.r` as the reset domain for the internals of `Bar`. 
 ```mlir
firrtl.circuit "Foo" {
  firrtl.module @Foo(in %p: !firrtl.uint<1>, in %r: !firrtl.asyncreset [{class = "sifive.enterprise.firrt
l.FullAsyncResetAnnotation"}], in %c: !firrtl.clock) attributes {convention = #firrtl<convention scalariz
ed>} {
    %bar_c, %bar_r = firrtl.instance bar @Bar(in c: !firrtl.clock, in r: !firrtl.asyncreset)
    %c0_ui1 = firrtl.constant 0 : !firrtl.const.uint<1>
    %0 = firrtl.asAsyncReset %c0_ui1 : (!firrtl.const.uint<1>) -> !firrtl.const.asyncreset
    %1 = firrtl.constCast %0 : (!firrtl.const.asyncreset) -> !firrtl.asyncreset
    firrtl.matchingconnect %bar_r, %1 : !firrtl.asyncreset
    firrtl.matchingconnect %bar_c, %c : !firrtl.clock
  }
  firrtl.module private @Bar(in %c: !firrtl.clock, in %r: !firrtl.asyncreset) {
    %reg = firrtl.reg %c : !firrtl.clock, !firrtl.uint<8>
  }
}

// -----// IR Dump After InferResets (firrtl-infer-resets) //----- //
firrtl.circuit "Foo" {
  firrtl.module @Foo(in %p: !firrtl.uint<1>, in %r: !firrtl.asyncreset [{class = "sifive.enterprise.firrtl.FullAsyncResetAnnotation"}], in %c: !firrtl.clock) attributes {convention = #firrtl<convention scalarized>} {
[...]

PR #8172, opened 2025-02-01
[FIRRTL] FART: allow modules to be in multiple domains at once

youngar:
When a child module is underneath two different reset domains, the pass
would emit an error message.  This change makes it so that it is allowed
as long as the reset signal name and the reset signal type between the
two domains are the same.

This also changes the logic to re-use a preexisting reset port: if there
is a port with the correct name for a reset domain, but the type does
match, we used to pick a different port name.  This changes it to be an
error, which matches the old SFC behaviour.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
@@ -55,48 +55,65 @@ using namespace firrtl;
 // Utilities
 //===----------------------------------------------------------------------===//
 
-namespace {
-/// A reset domain.
-struct ResetDomain {
-  /// Whether this is the root of the reset domain.
-  bool isTop = false;
-  /// The reset signal for this domain. A null value indicates that this domain
-  /// explicitly has no reset.
-  Value reset;
-
-  // Implementation details for this domain.
-  Value existingValue;
-  std::optional<unsigned> existingPort;
-  StringAttr newPortName;
-
-  ResetDomain(Value reset) : reset(reset) {}
-};
-} // namespace
-
-inline bool operator==(const ResetDomain &a, const ResetDomain &b) {
-  return (a.isTop == b.isTop && a.reset == b.reset);
-}
-inline bool operator!=(const ResetDomain &a, const ResetDomain &b) {
-  return !(a == b);
-}
-
 /// Return the name and parent module of a reset. The reset value must either be
 /// a module port or a wire/node operation.
 static std::pair<StringAttr, FModuleOp> getResetNameAndModule(Value reset) {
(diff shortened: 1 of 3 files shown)


seldridge:
I think this fixes https://github.com/llvm/circt/issues/7611?

youngar:
Yeah, this fixes that issue.  This also fixes issue https://github.com/llvm/circt/issues/7271 to error if the pre-existing reset port has the wrong type. (will update the commit message to reflect this)

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -55,50 +55,67 @@ using namespace firrtl;
...
   /// Whether this is the root of the reset domain.
   bool isTop = false;
+
   /// The reset signal for this domain. A null value indicates that this domain
   /// explicitly has no reset.
-  Value reset;
-
-  // Implementation details for this domain.
-  Value existingValue;
+  Value rootReset;
```
seldridge, line 84: What is the current interaction between `isTop` and `rootReset`? Is the former now unnecessary?

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -1579,72 +1606,61 @@ void InferResetsPass::determineImpl() {
 ///   the existing port/wire/node as reset.
 /// - If the module already has a port with the reset's name:
 ///   - If the port has the same type as the reset domain, reuses that port.
```
seldridge, line 1608: Possibly missing a change indicating that the name must match too?

seldridge (approved):
LGTM

Have you tested this out internally? I'm curious if this change to require unique reset names is going to cause a breakage during the next CIRCT bump. 🧐

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
youngar: At this point, `rootReset` is really only there for error messages. I do think there is some redundant information in this struct, but I am not sure how to factor it.  In the end I decided to just keep it simple, but maybe we can tackle this more in the future? For example, I am not sure why `buildDomains` and `determineImpl` are separate phases.  If we squash them in to one, we don't have to hold on to so much information, and could delete the `isTop` field.  As well, something like the `existingPort` could be determined by casting the local reset to a block arg and asking for the port number.  There are probably some other things we could do.

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
youngar: Thanks!

PR #8174, opened 2025-02-03
[circt-verilog] Fix negative Rest timing

maerhart:
In the pass timing summary there is usually a negative amount for Rest which is the result of measuring the verilog parsing and preprocessing twice


fabianschuiki (approved):
Oops. LGTM!

Issue #8162, opened 2025-01-31
[Verif] `verif.symbolic_value` semantics are ambiguous


TaoBi22:
It looks to me like the semantics of the `verif.symbolic_value` are unclear right now - specifically, there's no information in the docs about how/when the value is allowed to change. We discussed it a little [here](https://github.com/llvm/circt/pull/7145#discussion_r1678878295) when the op was first added, but I'm not sure we really reached a conclusion before it was merged.

Ultimately the issue stems from the fact that there are two interpretations of what a symbolic value means. In the SMT solver world it means a value that can have one assignment, whereas from a hardware verification angle it could potentially refer to an input that can change value over time.

I had a chat with @maerhart and we think there are three options:

1. `symbolic_value` can have only one assignment. (This is my preferred option, as I think the name `symbolic_value` implies it only ever has one value. In this case I'd suggest we also add a `symbolic_input` op where the values can change)
2. `symbolic_value` can change value on a clock edge. I think this is what we assume in #7870 (and I believe is consistent with BTOR2), but the semantics are still pretty murky on this interpretation because we don't have an associated clock. If we pick this we think it the op needs a `!seq.clock` operand to determine when it can change.
3. `symbolic_value` can change value whenever. This wouldn't require any modification to the op, and I think it probably holds with the interpretation taken in #7870 since I guess the value would always be latched when it's not going into combinational ops, but I'm not sure I haven't missed a corner case to do with comparing the value in a property or assumption later on.

IMO any of these would be fine, I just think we need to clarify which is the case in the op description.

CC: @fabianschuiki @leonardt @dobios

dobios:
My original design of this was that it would directly mirror the `input` op in BTOR2 (which is why the original name of this op was `symbolic_input` before people decided that it was too confusing and should have a more generic name), the idea was to determine the clock they were associated to based on the clocked assertions they are used in, which is why I didn't include a clock in the original design. So the value changes every clock cycle.

I like the idea of distinguishing the two but I'm not sure how this would translate to any front-end as inputs are inferred as symbolic based on whether or not they are used in the definition of a formal test and never explicitly defined, again why I tried to keep the signature so simple. It's unclear to me if a newly distinguished set of symbolic would actually ever be used tbh

fabianschuiki:
[...]

PR #8177, opened 2025-02-03
[NFC][Verif] Clarify `symbolic_value` semantics

TaoBi22:
As discussed in #8162, currently the semantics of `verif.symbolic_value` are unclear. This clarifies that the value can change between timesteps.


fabianschuiki (approved):
Thanks for clarifying! 👍

PR #8178, opened 2025-02-03
[MooreToCore] Don't insert wait event checks if all detects are AnyChange

maerhart:
This is legal because the llhd.wait operation already makes sure that at least one of the observed values has changed, i.e., the disjunct of all not-equal comparisons will always evaluate to true.


fabianschuiki (approved):
LGTM! This is a very neat simplification 👍

PR #8179, opened 2025-02-03
[circt-verilog] Add LLHD lowering pipeline

maerhart:


fabianschuiki (approved):
LGTM! This is fantastic to have as part of the tool.

PR #8182, opened 2025-02-03
[MooreToCore] Properly handle OOB accesses of moore.extract

maerhart:


lib/Dialect/HW/HWOps.cpp
```diff
@@ -2110,6 +2110,9 @@ void ArrayConcatOp::build(OpBuilder &b, OperationState &state,
 }
 
 OpFoldResult ArrayConcatOp::fold(FoldAdaptor adaptor) {
+  if (getInputs().size() == 1)
+    return getInputs()[0];
+
```
fabianschuiki, line 2115: Nice catch!

fabianschuiki (approved):
This looks great! Thanks for adding support for this 🥳

PR #8183, opened 2025-02-03
[FIRRTL] Add Clock Switch intrinsic

darthscsi:
Add an intrnisic for clock muxes.  This is needed as clock-manipulation must be understood for various aggressive FPGA transformations.

See https://www.eetimes.com/techniques-to-make-clock-switching-glitch-free/

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
@@ -110,6 +110,31 @@ def ClockGateIntrinsicOp : FIRRTLOp<"int.clock_gate", [Pure]> {
   }];
 }
 
+def ClockSwitchIntrinsicOp : FIRRTLOp<"int.clock_switch", [Pure]> {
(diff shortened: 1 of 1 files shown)


include/circt/Dialect/FIRRTL/FIRRTLIntrinsics.td
```diff
@@ -110,6 +110,31 @@ def ClockGateIntrinsicOp : FIRRTLOp<"int.clock_gate", [Pure]> {
   }];
 }
 
+def ClockSwitchIntrinsicOp : FIRRTLOp<"int.clock_switch", [Pure]> {
+  let summary = "Safely switch a clock with an enable signal";
+  let description = [{
+    The `int.clock_switch` switches clocks safely, without glitches,
+    based on a boolean enable value. If the enable input is 1, the output clock
+    produced by the second input clock.  If the enable
```
fabianschuiki, line 118:
```suggestion
    based on a boolean switch value. If the switch input is 1, the output clock
    produced by the second input clock.  If the switch
```

PR #8187, opened 2025-02-04
[RTG] Add the PyRTG frontend

maerhart:
This commit adds the boiler plate and basic rtgtool with a minimal example to run.

### DIFF
--- a/frontends/PyRTG/test/lit.cfg.py
+++ b/frontends/PyRTG/test/lit.cfg.py
@@ -1,0 +1,82 @@
+# -*- Python -*-
+
+import os
+import platform
+import re
+import shutil
+import subprocess
+import tempfile
+import warnings
+
+import lit.formats
+import lit.util
+
+from lit.llvm import llvm_config
+from lit.llvm.subst import ToolSubst
+from lit.llvm.subst import FindTool
+
+# Configuration file for the 'lit' test runner.
+
+# name: The name of this test suite.
+config.name = 'PyRTG'
+
+config.test_format = lit.formats.ShTest(not llvm_config.use_lit_shell)
+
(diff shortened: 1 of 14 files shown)


maerhart:
If I run `yapf` on `lit.cfg.py` I get this error `IndexError: tuple index out of range`. Also, there doesn't seem to be an artifact with a diff attached for Python misformatting. Anyone a clue how to get this file formatted properly?

frontends/PyRTG/test/lit.cfg.py
```diff
@@ -0,0 +1,79 @@
...
+
+llvm_config.add_tool_substitutions(tools, tool_dirs)
+
+# Tweak the PYTHONPATH to include the binary dir.
+llvm_config.with_environment(
+    'PYTHONPATH', [os.path.join(config.circt_python_packages_dir, 'pyrtg')],
+    append_path=True)
+
+# rtgtool shortcut
+config.substitutions.append(('%rtgtool%', f'{config.python_executable} {os.path.join(config.circt_python_packages_dir, 'pyrtg', 'rtgtool', 'rtgtool.py')}'))
```
fabianschuiki, line 79: Could the yapf issue be related to the nested `'` in the format string? It's a bit ugly, but maybe splitting the format string out into variables and interpolating those could work, or using a `... + " " + ...` instead to pull the complex interpolations out into expressions outside the string literal.

maerhart: Good call! Interesting that yapf just crashes instead of printing a useful error message in such a case.

fabianschuiki: Yeah very strange indeed 🤔

PR #8190, opened 2025-02-04
[ImportVerilog] Fix bit slicing into variables declared with offset

maerhart:
The case where the index value's bitwidth is not big enough to hold any index into the accessed array is not properly handled yet. We need to extend the index value in that case. Currently the ConstantOp folder would just crash.


test/Conversion/ImportVerilog/basic.sv
```diff
@@ -890,7 +892,7 @@ module Expressions;
     // CHECK: moore.extract [[TMP1]] from 1 : l32 -> l3
     y = vec_1[3:1];
     // CHECK: [[TMP1:%.+]] = moore.read %vec_2 : <l32>
-    // CHECK: moore.extract [[TMP1]] from 2 : l32 -> l2
+    // CHECK: moore.extract [[TMP1]] from 29 : l32 -> l2
```
hailongSun2000, line 895: I notice this case invokes the `translateIndex()`, but I have a question. For example, assume
```
logic [0:7] vec_2 = 8'b0000 1111;
```
From my perspective, the original `vec_2[2:3]` will return `00`(from left to right, get 2th-3th bits), but after invoking `tranlateIndex()`, it will return `11`(from left to right, get 5th-6th bits). Please correct me if I'm wrong.
Hey, @MikePopoloski! Greatly appreciate your help if you could explain this to me.

MikePopoloski: translateIndex() converts from an offset that is relative to a declared type range, like in your example, to an offset that is zero-based little endian. The intended purpose is to be used during constant evaluation, where internally all integers are stored in that canonical little endian format.

maerhart: Yeah, these things are quite confusing. We want that `vec_2[2:3]` returns `00` because that's what the SystemVerilog standard says (as far as I understand). However, this is not what currently (i.e., before this PR) happens. If we select index 2 in SV we will get 1 because we don't translate big-endian to little-endian, so `moore.extract` will still have `lowBit=2` which is then lowered to a `comb.extract` with `lowBit=2` (in this second lowering we cannot change this anymore because we don't know anything about the original endianess anymore). So, to get 0 we must convert to little endian during the conversion to the Moore dialect (in principle it could also choose big endian, but it doesn't explicitly say so, so I take the straightforward route of using the same system as Comb).

Also, as far as I know assigning a vector/array to vector/array of the other endianess does not reverse the bit order (but is more like a `reinterpret_cast` in C++).

The following should illustrate that:

```
--------------------------
| MSB                LSB |
--------------------------
| Verilog index (big endian, your example)
| 0  1  2  3  4  5  6  7 |
--------------------------
| Verilog index (little endian)
| 7  6  5  4  3  2  1  0 |
--------------------------
| Comb Dialect index [1]
| 7  6  5  4  3  2  1  0 |
--------------------------
| Value
| 0  0  0  0  1  1  1  1 |
--------------------------
```
[1] https://circt.llvm.org/docs/Dialects/Comb/RationaleComb/#bitcasts

Please let me know if I overlooked something.

fabianschuiki (approved):
This looks great! It's so confusing that Verilog let's you assign new numerical indices to the MSB and LSB, and even do so in totally reversed order, but then still have the underlying packed representation be identical regardless of what you do. Great to get this out of the way right when we build up the Moore dialect, such that no one ever has to think about this mess anymore afterwards 😁

test/Conversion/ImportVerilog/basic.sv
hailongSun2000: Can I consider that even the declared variable with [0:7](big endian), but finally, it will store matching in-hardware layout(little endian)? Such as:
--------------------------
| MSB                LSB |
--------------------------
| Declared verilog order (big endian)
| 0  1  2  3  4  5  6  7 |
--------------------------
| Real layout (littel endian)
| 7  6  5  4  3  2  1  0 |
--------------------------
| Value
| 0  0  0  0  1  1  1  1 |
--------------------------
So when slicing [2:3], we can get `00`([7-2]th and [7-3]th) from the `Real layout` rather than `Declared verilog order`?

hailongSun2000: I got it. Thanks for your patient explanation :heart:!

PR #8191, opened 2025-02-05
[FIRRTL] Disallow reading from property ports

youngar:
According to the FIRRTL spec, the right hand side of a propassign must have source flow.  This means that it is illegal to read from a property output port, something that is allowed for regular HW ports.  This change enforces this property during flow checking in the operation verifiers.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLOps.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOps.cpp
@@ -3518,9 +3518,11 @@ static LogicalResult checkConnectFlow(Operation *connect) {
   // instance/memory input ports.
   auto srcFlow = foldFlow(src);
   if (!isValidSrc(srcFlow)) {
-    // A sink that is a port output or instance input used as a source is okay.
+    // A sink that is a port output or instance input used as a source is okay,
+    // as long as it is not a property.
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -3518,9 +3518,11 @@ static LogicalResult checkConnectFlow(Operation *connect) {
   // instance/memory input ports.
   auto srcFlow = foldFlow(src);
   if (!isValidSrc(srcFlow)) {
-    // A sink that is a port output or instance input used as a source is okay.
+    // A sink that is a port output or instance input used as a source is okay,
+    // as long as it is not a property.
     auto kind = getDeclarationKind(src);
-    if (kind != DeclKind::Port && kind != DeclKind::Instance) {
+    if (isa<PropertyType>(src.getType()) ||
+        (kind != DeclKind::Port && kind != DeclKind::Instance)) {
```
seldridge, line 3525: This looks like the right change and the right place to make it.

seldridge (approved):
LGTM

It's kind of strange that this is running through flow checking. However, given that it is, this is exactly the right change to make. 👍

PR #8020, opened 2024-12-29
[ImportVerilog] add real literal support, refine real type, and add real format support

chenbo-again:
This PR is aimed to support systemverilog real, shortreal, realtime, and there conversion and operation.

One thing I'm not so sure about is that

```mlir
class BinaryOpBase<string mnemonic, list<Trait> traits = []> :
    MooreOp<mnemonic, traits # [Pure, SameOperandsAndResultType]> {
  let arguments = (ins SimpleBitVectorType:$lhs, SimpleBitVectorType:$rhs);
  let results = (outs SimpleBitVectorType:$result);
  let assemblyFormat = [{
    $lhs `,` $rhs attr-dict `:` type($result)
  }];
}
```

why this opbase just support SimpleBitVectorType? is there any shortcoming to just convert it to a UnpackedType? I also want to use this OpBase support float point operation.

PR #8192, opened 2025-02-05
[ImportVerilog]Support real math functions.

hailongSun2000:


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -716,17 +716,61 @@ struct RvalueExprVisitor {
...
-      return context.convertRvalueExpression(*args[0]);
-
-    if (subroutine.name == "$clog2") {
-      auto value = context.convertToSimpleBitVector(
-          context.convertRvalueExpression(*args[0]));
-      if (!value)
-        return {};
-      return builder.create<moore::Clog2BIOp>(loc, value);
-    }
+    std::map<std::string, std::function<Value()>> subroutineMap = {
```
hailongSun2000, line 723: Hey, @fabianschuiki! I want to replace `std::map` with `llvm::DenseMap`(lack of DenseMapInfo now). As far as I know, the time complexity of the `DenseMap` is O(1), and the space complexity is O(n), right? However, I'm not clear about the `DenseMap` and `multi-if` statements which one is better(From my perspective, the time complexity of `multi-if` is O(n), but the space complexity is O(1)). I follow you :smiley:!

hailongSun2000: Maybe the `llvm::StringSwitch` is a better choice :thinking:?

hailongSun2000: Appendix:
```
auto realMathFunc =
        llvm::StringSwitch<std::function<Value()>>(subroutine.name)
            // Signed and unsigned system functions.
            .Case("$signed", [&]() { return value; })
            .Case("$unsigned", [&]() { return value; })

            // Math functions in SystemVerilog.
            .Case("$clog2",
                  [&]() -> Value {
                    value = context.convertToSimpleBitVector(value);
                    if (!value)
                      return {};
                    return builder.create<moore::Clog2BIOp>(loc, value);
                  })
            .Case("$ln",
                  [&]() { return builder.create<moore::LnBIOp>(loc, value); })
            .Case(
                "$log10",
                [&]() { return builder.create<moore::Log10BIOp>(loc, value); })
            .Case("$sin",
                  [&]() { return builder.create<moore::SinBIOp>(loc, value); })
            .Case("$cos",
                  [&]() { return builder.create<moore::CosBIOp>(loc, value); })
            .Case("$tan",
                  [&]() { return builder.create<moore::TanBIOp>(loc, value); })
            .Case("$exp",
                  [&]() { return builder.create<moore::ExpBIOp>(loc, value); })
            .Case("$sqrt",
                  [&]() { return builder.create<moore::SqrtBIOp>(loc, value); })
            // TODO: $pow(x, y)
            .Case(
                "$floor",
                [&]() { return builder.create<moore::FloorBIOp>(loc, value); })
            .Case("$ceil",
                  [&]() { return builder.create<moore::CeilBIOp>(loc, value); })
            .Case("$asin",
                  [&]() { return builder.create<moore::AsinBIOp>(loc, value); })
            .Case("$acos",
                  [&]() { return builder.create<moore::AcosBIOp>(loc, value); })
            .Case("$atan",
                  [&]() { return builder.create<moore::AtanBIOp>(loc, value); })
            // TODO: atan2(y, x) and hypot(x, y)
            .Case("$sinh",
                  [&]() { return builder.create<moore::SinhBIOp>(loc, value); })
            .Case("$cosh",
                  [&]() { return builder.create<moore::CoshBIOp>(loc, value); })
            .Case("$tanh",
                  [&]() { return builder.create<moore::TanhBIOp>(loc, value); })
            .Case(
                "$asinh",
                [&]() { return builder.create<moore::AsinhBIOp>(loc, value); })
            .Case(
                "$acosh",
                [&]() { return builder.create<moore::AcoshBIOp>(loc, value); })
            .Case(
                "$atanh",
                [&]() { return builder.create<moore::AtanhBIOp>(loc, value); })

    // Invoke realMathFunc() to select which builtin is.
    if (auto result = realMathFunc())
      return result;
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1626,4 +1657,76 @@ def Clog2BIOp : Builtin<"clog2", [SameOperandsAndResultType]> {
...
+}
+
+def AcoshBIOp : RealMathFunc<"acosh"> {
+  let summary = "Arc-hyperbolic cosine";
+}
+
+def AtanhBIOp : RealMathFunc<"atanh"> {
+  let summary = "Arc-hyperbolic tangent";
+}
+
```
fabianschuiki, line 1731: Very cool!

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: I like the `StringSwitch` approach! The problem with the map/DenseMap is that it is constructed dynamically every single time this function is called, which is going to be much more expensive than checking a few strings.

In the long run we could do something like FIRRTL's intrinsic lowering, which uses a DenseMap as well, but it populates the dense map once at startup. This could even be a lazily-initialized static, or a DenseMap in the ImportVerilog context which we populate once in the constructor. No need to do any of that yet though, your StringSwitch approach is great and gets the job done nicely :+1:

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -716,17 +716,61 @@ struct RvalueExprVisitor {
                   const slang::ast::CallExpression::SystemCallInfo &info) {
     const auto &subroutine = *info.subroutine;
     auto args = expr.arguments();
+    auto value = context.convertRvalueExpression(*args[0]);
+    if (!value)
+      return {};
```
fabianschuiki, line 721: We need to check whether `args.size() == 1` first. Maybe you could wrap the `convertRvalueExpression` and your entire StringSwitch in an `if (args.size() == 1) { ... }`? In the future, when we support different arities of system tasks/calls, we can expand this to call into some special handling functions like `convertSystemCall1`, `convertSystemCall2`, `convertSystemCallVariadic`, and similar. Not needed now though, since all we support are unary/arity-1 system calls that have a single argument.

test/Conversion/ImportVerilog/builtins.sv
```diff
@@ -191,9 +192,46 @@ endfunction
...
+  // CHECK:  moore.builtin.cosh [[R]] : real
+  dummyB($cosh(r));
+  // CHECK:  moore.builtin.tanh [[R]] : real
+  dummyB($tanh(r));
+  // CHECK:  moore.builtin.asinh [[R]] : real
+  dummyB($asinh(r));
+  // CHECK:  moore.builtin.acosh [[R]] : real
+  dummyB($acosh(r));
+  // CHECK:  moore.builtin.atanh [[R]] : real
+  dummyB($atanh(r));
```
fabianschuiki, line 236: Really nice!

fabianschuiki (approved):
This looks great! Thanks for adding all of this. Just a minor comment about first checking that the syscall has exactly 1 argument. Besides that, this looks good to go!

lib/Conversion/ImportVerilog/Expressions.cpp
hailongSun2000: > The problem with the map/DenseMap is that it is constructed dynamically every single time this function is called

Yeah, right! I got it.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -716,17 +716,101 @@ struct RvalueExprVisitor {
...
+                  })
+            .Case("$atanh",
+                  [&](Value value) {
+                    return builder.create<moore::AtanhBIOp>(loc, value);
+                  })
+            .Default([&](Value value) -> Value { return {}; });
+
+    // Invoke convertSystemCallArity1() to select which builtin is.
+    if (auto result = convertSystemCallArity1(value); args.size() == 1)
+      return result;
```
hailongSun2000, line 813: I changed the original `realMathFunc()` into `convertSystemCallArity1(value)` and moved it out of the `if` block. In the future, we can add `convertSystemCallArity2(x, y)` or `convertSystemCallArityN()` without args. What do @fabianschuiki think about this tweak :thinking:?

fabianschuiki: Nice! How about turning this into a proper function instead of a large inline lambda?

fabianschuiki: I think you still need to move all of this into a `if (args.size() == 1)`. Further up where you call `context.convertRvalueExpression(*args[0])`, this call will crash if `args.size() == 0`.

I'd suggest moving the `convertSystemCallArity1` into a proper function and doing something like this:
```c++
if (args.size() == 1) {
  auto value = context.convertRvalueExpression(*args[0]);
  if (!value)
    return {};
  auto result = convertSystemCallArity1(value);
  if (failed(result))
    return {};
  if (*result)
    return *result;
}
```

One thing you need to do in `convertSystemCallArity1` is you need to make it clear from the return value whether:
1. The conversion was successful and you return the result.
2. The conversion failed due to an error, e.g. that `convertToSimpleBitVector` fails.
3. The system call is unknown.
I'd change the return type to `FailureOr<Value>`, and then doing the following:
1. Return `result` of the created op.
2. Return `failure()` due to an error.
3. Return `Value{}` for unsupported system calls.

hailongSun2000: You're right! Regarding `if(args.size() == 1)`, I attempted the `$random()` and it will crash.

hailongSun2000: > I'd change the return type to FailureOr<Value>, and then doing the following:
> 1. Return result of the created op.
> 2. Return failure() due to an error.
> 3. Return Value{} for unsupported system calls.

I have a question about why we'll change the return type to `FailureOr<Value>`.  I notice `FailureOr<Value> convertFormatString`, but this function is used in **Statement.cpp**.  However, `convertToSimpleBitVector` returns `Value`, even if failed conversion, it just returns `Value{}` with an error message and not `LogicalResult`, so `if(failed(result))` is never to be triggered.

fabianschuiki: You're definitely right! The difference is that `convertToSimpleBitVector` only needs to distinguish two cases:
1. successful conversion (a non-null `Value` is returned)
2. failed conversion (a null `Value` is returned)

However, `convertSystemCallArity1` needs to distinguish three cases:
1. successful conversion (a non-null `Value` is returned)
2. function name recognized, but conversion failed (`failure()` is returned)
3. function name not recognized (a null `Value` is returned)

The `FailureOr<Value>` allows you to distinguish a failure from the case where the function didn't recognize the function name.

hailongSun2000: Thanks for your detailed explanation!

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1310,3 +1316,96 @@ Value Context::materializeConversion(Type type, Value value, bool isSigned,
...
+          .Case("$unsigned", [&]() { return value; })
+
+          // Math functions in SystemVerilog.
+          .Case("$clog2",
+                [&]() -> FailureOr<Value> {
+                  value = convertToSimpleBitVector(value);
+                  if (!value)
+                    return failure();
+                  return (Value)builder.create<moore::Clog2BIOp>(loc, value);
+                })
```
hailongSun2000, line 1336: Maybe this is reasonable? First time to use `std::function<FailureOr<Value>()>`. It just looks correct 😂.

fabianschuiki: Nice! 😃

hailongSun2000: Greatly appreciate your guidance!

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -508,6 +508,23 @@ def StringConstantOp : MooreOp<"string_constant", [Pure]> {
...
+    %0 = moore.real_constant 1.23
+    ```
+    The real type has fixed-point(1.2) and exponent(2.0e10) formats.
+    See IEEE 1800-2017 § 5.7.2 "Real literal constants".
+  }];
+  let arguments = (ins F64Attr:$value);
+  let results = (outs RealType:$result);
+  let assemblyFormat = "$value attr-dict `:` type($result)";
+}
+
```
hailongSun2000, line 527: Borrow from https://github.com/llvm/circt/pull/8020 .

PR #8196, opened 2025-02-06
[LLHD][TCM] Ignore processes with CFG loops within a TR

maerhart:
Ignore loops for now and basically assume full loop unrolling was performed before this pass. This was an implicit assumption until now. This PR explicitly checks for it and ignores the process in that case.


fabianschuiki (approved):
LGTM!

PR #8199, opened 2025-02-07
[RTG][InlineSequences] Support interleave_sequences

maerhart:


lib/Dialect/RTG/Transforms/InlineSequencesPass.cpp
```diff
@@ -35,62 +40,183 @@ struct InlineSequencesPass
...
+      }
+
+      auto *op = builder.clone(*iters[i], mapping);
+      if (isa<InstructionOpInterface>(op))
+        ++k;
+
+      ++iters[i];
+    }
+
+    i = (i + 1) % blocks.size();
```
youngar, line 178: could use a for loop here
```c++
  for (unsigned i = 0; !finishedBlocks.all(); i = (i + 1) % blocks.size()) {
    if (finishedBlocks[i])
      continue;
    for (unsigned k = 0; k < batchSize;) {
      if (iters[i] == blocks[i]->end()) {
        finishedBlocks.set(i);
        break;
      }
      auto *op = builder.clone(*iters[i], mapping);
      if (isa<InstructionOpInterface>(op))
        ++k;
      ++iters[i];
    }
  }
```

lib/Dialect/RTG/Transforms/InlineSequencesPass.cpp
```diff
@@ -35,62 +40,183 @@ struct InlineSequencesPass
...
+  size_t numSequencesInlined = 0;
+  size_t numSequencesInterleaved = 0;
+};
+
+} // namespace
+
+void SequenceInliner::materializeInterleavedSequence(Value value,
+                                                     ArrayRef<Block *> blocks,
+                                                     uint32_t batchSize) {
+  auto interleavedBlock = std::make_unique<Block>();
```
youngar, line 148: Maybe allocate the block right into the owning storage.
 ```c++
auto *interleavedBlock = blockStorage.emplace_back(std::make_unique<Block>()).get();
```

youngar (approved):

Issue #6696, opened 2024-02-14
[FIRRTL] Add folders for IntegerAddOp


mikeurbach:
This is a primitive Property operation, defined by FIRRTL spec section 25.1.1. General folds for signed arbitrary precision integer arithmetic can be applied, but none were defined in the initial PR. This could include things like:

* [ ] integer_add(a, 0) -> a
* [ ] integer_add(constantA, constantB) -> constantA + constantB
* [ ] etc...

sazimi13:
Hey I would like to take this issue up, after completing can hopefully take the other 2 folding issues as well.

sazimi13:
Can you please provide the FIRRTL spec that you are referring to?

PR #8200, opened 2025-02-07
[FIRRTL] Add a folder for `add` property op

mtsokol:
Hi @mikeurbach,

This PR addresses https://github.com/llvm/circt/issues/6696.

It's a first-time contribution for me and these issues remained open for quite some time. I hope I didn't take anyone's assignment.

In `canonicalization.mlir` file I got an error when tried to reformat `firrtl.class @PropertyArithmetic...` line into multiple lines. Does `firrtl.class` support same format as `firrtl.module` in `mlir` files?

In a follow-up PR I can add other folding rules that were listed in aforementioned issues. Does `x + x -> x * 2` fold also makes sense here?

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -935,8 +935,20 @@ LogicalResult NEQPrimOp::canonicalize(NEQPrimOp op, PatternRewriter &rewriter) {
 }
 
 OpFoldResult IntegerAddOp::fold(FoldAdaptor adaptor) {
-  // TODO: implement constant folding, etc.
-  // Tracked in https://github.com/llvm/circt/issues/6696.
+  if (auto rhsCst = getConstant(adaptor.getRhs())) {
+    // Constant folding
+    if (auto lhsCst = getConstant(adaptor.getLhs())) {
+      auto resultWidth =
+          std::max(lhsCst->getBitWidth(), rhsCst->getBitWidth()) + 1;
+      APSInt lhsCstExt = (*lhsCst).extend(resultWidth);
+      APSInt rhsCstExt = (*rhsCst).extend(resultWidth);
+      return IntegerAttr::get(IntegerType::get(getContext(), resultWidth),
+                              lhsCstExt + rhsCstExt);
+    }
+    // x + 0 -> x
+    if (rhsCst->isZero())
+      return getLhs();
+  }
   return {};
 }
 
--- a/test/Dialect/FIRRTL/canonicalization.mlir
+++ b/test/Dialect/FIRRTL/canonicalization.mlir
@@ -3576,19 +3576,28 @@ firrtl.module private @Issue7562(in %sel : !firrtl.uint<1>, in %a : !firrtl.cons
 }
 
 // CHECK-LABEL: firrtl.class @PropertyArithmetic
-firrtl.class @PropertyArithmetic(in %in: !firrtl.integer, out %out0: !firrtl.integer, out %out1: !firrtl.integer) {
+firrtl.class @PropertyArithmetic(in %in: !firrtl.integer, out %out0: !firrtl.integer, out %out1: !firrtl.integer, out %out2: !firrtl.integer, out %out3: !firrtl.integer) {
+  // CHECK: [[C3:%.+]] = firrtl.integer 3
   // CHECK: [[C4:%.+]] = firrtl.integer 4
   %0 = firrtl.integer 0
   %1 = firrtl.integer 1
   %2 = firrtl.integer 2
 
-  %3 = firrtl.integer.shl %1, %2 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
-  %4 = firrtl.integer.shl %in, %0 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
+  %res0 = firrtl.integer.shl %1, %2 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
+  %res1 = firrtl.integer.shl %in, %0 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
 
   // CHECK: firrtl.propassign %out0, [[C4]]
   // CHECK: firrtl.propassign %out1, %in
-  firrtl.propassign %out0, %3 : !firrtl.integer
-  firrtl.propassign %out1, %4 : !firrtl.integer
+  firrtl.propassign %out0, %res0 : !firrtl.integer
+  firrtl.propassign %out1, %res1 : !firrtl.integer
+
+  %res2 = firrtl.integer.add %1, %2 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
+  %res3 = firrtl.integer.add %in, %0 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
+
+  // CHECK: firrtl.propassign %out2, [[C3]]
+  // CHECK: firrtl.propassign %out3, %in
+  firrtl.propassign %out2, %res2 : !firrtl.integer
+  firrtl.propassign %out3, %res3 : !firrtl.integer
 }
 
 // CHECK-LABEL: firrtl.module private @LayerBlocks
--- a/include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
@@ -1393,6 +1393,8 @@ def IntegerAddOp : IntegerBinaryPrimOp<"integer.add", [Commutative]> {
                                          !firrtl.integer
     ```
   }];
+
+  let hasFolder = true;
 }
 
 def IntegerMulOp : IntegerBinaryPrimOp<"integer.mul", [Commutative]> {


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -935,14 +935,37 @@ LogicalResult NEQPrimOp::canonicalize(NEQPrimOp op, PatternRewriter &rewriter) {
 }
 
 OpFoldResult IntegerAddOp::fold(FoldAdaptor adaptor) {
-  // TODO: implement constant folding, etc.
-  // Tracked in https://github.com/llvm/circt/issues/6696.
+  if (auto rhsCst = getConstant(adaptor.getRhs())) {
+    // Constant folding
+    if (auto lhsCst = getConstant(adaptor.getLhs())) {
+      auto resultWidth = lhsCst->getBitWidth() + rhsCst->getBitWidth();
```
mikeurbach, line 941: I think the result logic might be flipped between add and multiply... Don't we normally add bitwidths for multiplication and take max width + 1 for addition?

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -935,14 +935,37 @@ LogicalResult NEQPrimOp::canonicalize(NEQPrimOp op, PatternRewriter &rewriter) {
 }
 
 OpFoldResult IntegerAddOp::fold(FoldAdaptor adaptor) {
-  // TODO: implement constant folding, etc.
-  // Tracked in https://github.com/llvm/circt/issues/6696.
+  if (auto rhsCst = getConstant(adaptor.getRhs())) {
+    // Constant folding
```
mikeurbach, line 939: Any chance this could re-use this helper? https://github.com/llvm/circt/blob/f83436b47ed2056fd9fd58be7f5fe0762f6d0bbc/lib/Dialect/FIRRTL/FIRRTLFolds.cpp#L204-L210

I assume not, because that helper wants to work on hardware types and we specifically made separate Property types for `!firrtl.integer`.

It might be worth coming up with a similar helper for Property... but I think it's fine to leave this as-is for the first two folders before trying to generalize. Eventually if we have 3+ folders with a lot of repetition, maybe we make a helper.

test/Dialect/FIRRTL/canonicalization.mlir
```diff
@@ -3576,19 +3576,38 @@ firrtl.module private @Issue7562(in %sel : !firrtl.uint<1>, in %a : !firrtl.cons
...
+  %res2 = firrtl.integer.mul %2, %5 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
+  %res3 = firrtl.integer.mul %in, %1 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
+
+  // CHECK: firrtl.propassign %out2, [[C10]]
+  // CHECK: firrtl.propassign %out3, %in
+  firrtl.propassign %out2, %res2 : !firrtl.integer
+  firrtl.propassign %out3, %res3 : !firrtl.integer
+
+  %res4 = firrtl.integer.shl %1, %2 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
+  %res5 = firrtl.integer.shl %in, %0 : (!firrtl.integer, !firrtl.integer) -> !firrtl.integer
```
mikeurbach, line 3605: nit: it made the diff more confusing to move these to the bottom and rename their SSA values. At first, I thought "where'd the shl folder tests go".

I do like the name better, but in this case I think we'd prefer a smaller diff. Specifically, I think a couple LLVM guidelines apply here:

https://llvm.org/docs/CodingStandards.html#introduction (the note about sticking to existing style)
https://llvm.org/docs/DeveloperPolicy.html#incremental-development (each change should be kept as small as possible)

In this case, it should be possible to update this test file by adding new outputs and operations, without changing existing lines besides the module signature.

mikeurbach (commented):
Thank you for working on this!

Left some comments inline. A couple general points...

This should really be split into two PRs. There is a reason we have separate issues for the separate operations: we generally try to make changes as small as possible, and prefer to send many many small changes. This is spelled out in the LLVM developer policy, which we try to follow in CIRCT: https://llvm.org/docs/DeveloperPolicy.html. Specifically, the section on incremental development describes some of the rationale here. I would prefer to see two PRs for these changes, one for each issue.

Keep in mind that we also follow the LLVM model for code review: https://llvm.org/docs/CodeReview.html. At some point, once you get commit access, you have discretion to send "obvious" patches without pre-merge review. This is a policy that works well with incremental development.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
mtsokol: Ahh! 😅 Exactly, they're swapped. I took both from [the FIRRTL spec](https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#add-operation) but the other way around. It just shows that my test case is too trivial!

mikeurbach: Also, keep in mind that these are not the operations you linked to in the FIRRTL spec. Those are the primitive _hardware_ operations for `add` and `mul`. What you are adding folders for is the primitive _property_ operations for `integer_add` and `integer_mul`: https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#integer-add-operation

These are already just defined as "arbitrary precision". Do we need the explicit extension, or can we just use `+` or `*` with the APInts to get the right result (i.e., does APInt extend for us?).

mtsokol: My first attempt was to implement these operations without calling `extend` but unfortunately this results in an incorrect outcome. For example `2` and `5` (which are `APSInt`s that wrap `APInt`) are passed with bit-width `3` and multiplying them directly doesn't extent the bit-width to `4` and the result is `-6` instead of `10`.

AFAIC `APSInt` class doesn't have an automatic bit-width adjustment mechanism. It can be done by calling [extend](https://github.com/llvm/llvm-project/blob/dd369c771e215ebf7bf2d09e874643788e49d334/llvm/include/llvm/ADT/APSInt.h#L112) or [trunc](https://github.com/llvm/llvm-project/blob/dd369c771e215ebf7bf2d09e874643788e49d334/llvm/include/llvm/ADT/APSInt.h#L108) and passing the new width explicitly only. One might wonder how `%5 = firrtl.integer 5` knows the required number of bits. It's done by `APSInt::APSInt(StringRef Str)`, [the constructor](https://github.com/llvm/llvm-project/blob/dd369c771e215ebf7bf2d09e874643788e49d334/llvm/lib/Support/APSInt.cpp#L21) - it overestimates the number of bits, then checks how many are actually used with `getSignificantBits` and `trunc`s it.

So IMO we need an explicit extension, which isn't troubling as we know the formula for output's bit-width.

mtsokol: Also, I see that hardware ops, like `add` and `mul`, inherit from `IntBinaryPrimOp` which has pretty flexible `inferType` and does this inference there (e.g. for `add` it's `impl::inferAddSubResult`).

OTOH Property operations inherit from `IntegerBinaryPrimOp` where the `inferType` is [hardcoded for all ops](https://github.com/llvm/circt/blob/76a1dc39a2947ad12cf2a96cd5dba12ea5e3747e/include/circt/Dialect/FIRRTL/FIRRTLExpressions.td#L1372)!

```td
// We use a static inferType that always returns FIntegerType.
let inferType = "getFIntegerType";

// Define the getFIntegerType inline in ODS.
let firrtlExtraClassDeclaration = [{
  static FIntegerType getFIntegerType(FIRRTLType lhs,
                                      FIRRTLType rhs,
                                      std::optional<Location> loc) {
    return FIntegerType::get(lhs.getContext());
  }
}];
```
As you can see inferred type is just the type of LHS of the op.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
mtsokol: I looked into it and `constFoldFIRRTLBinaryOp` (and all hardware types' folders) are pretty similar but I can't use them right away. 

As explained in the comment above, Property Ops inherit from `IntegerBinaryPrimOp` which hardcodes result type to the type of LHS (and also imposes the bitwidth of the result). `constFoldFIRRTLBinaryOp` calls `op->getResult(0).getType()` to obtain the result type so we can't use this helper for Property Ops.
(For hardware ops there are separate functions for inferring the result type).

For now I agree to keep it simple and infer directly in the folder, but what is your opinion on refactoring `IntegerBinaryPrimOp` (Property base class) so that `inferType` isn't hardcoded but rather overridable by ops? Then I think potentially Hardware ops' logic can be reused for Properties.

Is there a reason why `inferType` for Property Ops is hardcoded to just LHS type?

test/Dialect/FIRRTL/canonicalization.mlir
mtsokol: Right, there's no need to enforce any particular order of these tests. Done! And I only renames result values for `shl`.

mtsokol:
@mikeurbach Thank you for all first-PR guidance! I reduced this PR to `add` folder only - it's ready from my side. The follow-ups will contain `mul` folder (and any infer-type refactorings if we decide so).

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
mikeurbach: I see that both these comment threads converged to the same `inferType`, so I'll comment about that aspect here...

I think this has been required since I first added IntegerBinaryPrimOp: https://github.com/llvm/circt/pull/6701#issuecomment-1944989233. So that's why it seems like `inferType` is kind of useless and hardcoded. Also, note that unlike the hardware types, FIntegerType does not carry the width in the type, so returning LHS or RHS is equivalent, and equivalent to constructing a fresh FIntegerType as well.

Anyway, I think there may be merit in considering if we can better integrate with the usual folding infrastructure for hardware types, but if we don't carry widths in FIntegerType, I don't really see how we could. And I don't think we necessarily want that either. So, what is in this PR makes sense to me. Thanks for considering my offhand question so deeply.

mikeurbach (approved):
Thanks for the iterations on this, this looks good to me!

mtsokol:
Thanks! Does this PR need anything else to be merged?

mikeurbach:
Oops, nope, sorry I forgot about this. I kicked off CI but forgot to come back and click merge. Doing so now. Thanks for the contribution!

PR #8201, opened 2025-02-07
[MooreToCore] Properly deal with OOB access in dyn_extract

maerhart:
The array slice case is not fixed yet. I couldn't really find a part in the SV standard about signed/unsigned indexing. I assume always interpreting the index operand as unsigned is fine.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -825,12 +825,30 @@ struct DynExtractOpConversion : public OpConversionPattern<DynExtractOp> {
                                      op->getLoc());
 
       if (isa<hw::ArrayType>(resultType)) {
+        // FIXME: for packed arrays, the part that is OOB must be 0 or X and the
+        // part overlapping with the range must match the values of the range;
+        // for unpacked arrays, if part of the accessed slice is OOB the entire
+        // result is 0 or X
         rewriter.replaceOpWithNewOp<hw::ArraySliceOp>(op, resultType,
                                                       adaptor.getInput(), idx);
         return success();
       }
 
-      rewriter.replaceOpWithNewOp<hw::ArrayGetOp>(op, adaptor.getInput(), idx);
+      Value size = rewriter.create<hw::ConstantOp>(
+          op.getLoc(), adaptor.getLowBit().getType(), arrType.getNumElements());
+      Value isOOB = rewriter.create<comb::ICmpOp>(
+          op.getLoc(), ICmpPredicate::uge, adaptor.getLowBit(), size);
+      Value access =
+          rewriter.create<hw::ArrayGetOp>(op.getLoc(), adaptor.getInput(), idx);
+      int64_t bw = hw::getBitWidth(arrType.getElementType());
+      if (bw < 0)
+        return failure();
+
+      Value zeroValueRaw =
+          rewriter.create<hw::ConstantOp>(op.getLoc(), APInt(bw, 0));
+      Value zeroValue = rewriter.createOrFold<hw::BitcastOp>(
+          op.getLoc(), arrType.getElementType(), zeroValueRaw);
+      rewriter.replaceOpWithNewOp<comb::MuxOp>(op, isOOB, zeroValue, access);
       return success();
     }
 
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -1123,7 +1123,7 @@ def ExtractOp : MooreOp<"extract", [Pure]> {
 def DynExtractOp : MooreOp<"dyn_extract", [Pure]> {
   let description = [{
     It's similar with extract, but it's used to select from a value
-    with a dynamic low bit.
+    with a dynamic low bit. The `lowBit` operand is always interpreted as unsigned.
   }];
   let arguments = (ins UnpackedType:$input, UnpackedType:$lowBit);
   let results = (outs UnpackedType:$result);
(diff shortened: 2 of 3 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -825,12 +825,30 @@ struct DynExtractOpConversion : public OpConversionPattern<DynExtractOp> {
...
+        // for unpacked arrays, if part of the accessed slice is OOB the entire
+        // result is 0 or X
         rewriter.replaceOpWithNewOp<hw::ArraySliceOp>(op, resultType,
                                                       adaptor.getInput(), idx);
         return success();
       }
 
-      rewriter.replaceOpWithNewOp<hw::ArrayGetOp>(op, adaptor.getInput(), idx);
+      Value size = rewriter.create<hw::ConstantOp>(
+          op.getLoc(), adaptor.getLowBit().getType(), arrType.getNumElements());
```
maerhart, line 838: That can cause problems if the lowBit bitwidth is too small to hold the array size.

fabianschuiki: Nice catch. In that case we might be guaranteed to always access in bounds?

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -825,12 +825,30 @@ struct DynExtractOpConversion : public OpConversionPattern<DynExtractOp> {
...
+          rewriter.create<hw::ArrayGetOp>(op.getLoc(), adaptor.getInput(), idx);
+      int64_t bw = hw::getBitWidth(arrType.getElementType());
+      if (bw < 0)
+        return failure();
+
+      Value zeroValueRaw =
+          rewriter.create<hw::ConstantOp>(op.getLoc(), APInt(bw, 0));
+      Value zeroValue = rewriter.createOrFold<hw::BitcastOp>(
+          op.getLoc(), arrType.getElementType(), zeroValueRaw);
+      rewriter.replaceOpWithNewOp<comb::MuxOp>(op, isOOB, zeroValue, access);
```
fabianschuiki, line 851: This looks reasonable assuming that `hw::ArrayGetOp` has undefined behaviour (or undefined result) for out-of-bounds accesses. I don't remember off the top of my head whether out-of-bounds index in `ArrayGetOp` causes a potential segfault in the lowered LLVM IR… But it might be nice if `hw::ArrayGetOp` just produce an undefined result, but don't cause arbitrary undefined behaviour such as crashing the simulator. Then we can just do exactly what you did here: read the value and then discard it with a mux if it's out of bounds.

fabianschuiki (approved):
LGTM!

lib/Conversion/MooreToCore/MooreToCore.cpp
maerhart: Yes 👍 
But I need to make sure this IR is not built in that case because the pass would crash.

hailongSun2000:
I have a question about the **unsigned** index, for example:
```
logic [6:-1] arr;
arr[-1];
```
Maybe `-1` represents the max if we assume the index is unsigned, right 🤔?

fabianschuiki:
@hailongSun2000 I think in your example ImportVerilog would lower this to an access index 0, because -1 is the "address" of the least significant bit. During conversion, the accessed index gets remapped from the range `[a:b]` to `[a-b:0]`. 🙂

hailongSun2000:
> @hailongSun2000 I think in your example ImportVerilog would lower this to an access index 0, because -1 is the "address" of the least significant bit. During conversion, the accessed index gets remapped from the range `[a:b]` to `[a-b:0]`. 🙂

That sounds  great!  All bit rang are converted into little  endian,  and LSB is always 0. Could you tell  me where the  implementation is?  I'm interested to it. Thanks in advance! :smiley:

hailongSun2000 (approved):
LGTM! :partying_face:

fabianschuiki:
> @hailongSun2000 All bit rang are converted into little endian, and LSB is always 0. Could you tell me where the implementation is?

I think @maerhart added this `getSelectIndex` call here to deal with the index remapping in a PR a few days ago: 🙂

https://github.com/llvm/circt/blob/6e93a0e8400d2e2f17725af096d470584713eda1/lib/Conversion/ImportVerilog/Expressions.cpp#L480

lib/Conversion/MooreToCore/MooreToCore.cpp
maerhart: The documentation of the HW aggregate ops is generally a bit sparse, but I'd say a OOB `hw.array_get` produces an undefined result, but no immediate UB. The HWToLLVM lowering currently does not take this into account. I'll create an issue for that.

maerhart:
As a little side-note: the example above would be lowered to an `moore.extract`, not a `moore.dyn_extract`

Even though all ranges are converted to little endian starting at 0, the index could still be a negative signed value in SV.

```sv
bit [31:0] arr;
bit [31:0] idx = -1;
arr[$signed(idx)];
```

What index does this actually access? -1 or 4,294,967,295? I couldn't really find an answer to this question in the standard, so let me know if you know more 😄 
If both are out-of-bounds, it doesn't really matter, but could you have something like

```sv
bit [4294967295:0] arr; // let's say all bits are 1
bit [31:0] idx = -1;
arr[$signed(idx)];
```

Should this now return 0 or 1? Does SV even support such big arrays (if it doesn't the same still applies but with smaller integers)?

fabianschuiki:
Since SV allows you to have negative indices on array declarations, like `bit [3:-8] x`, I'd assume that it supports signed indices when accessing arrays. So `x[-2]` would actually be treated as accessing index `-2`, which then gets translated to actual bit positions through that index remapping.

Since this only involves addition/subtraction which are identical in signed/unsigned arithmetic, I think we can just ignore the issue altogether: if you subtract the lower index of the declaration (-8) from the accessed index (-2), you'll automatically remap all valid negative indices to positive bit positions. Anything below the lowest bit address would overflow and be out of bounds immediately.

If I recall correctly, array index expressions have a self-determined type, not a type inferred from context. That would suggest in your example that the index `$signed(idx)` would still be a 32 bit number, even though this would create an ambiguity when accessing that large array. We could think about sign-extending signed indices in bit/part-select operations by one bit, to remove any ambiguity, or to sign/zero-extend the index to the minimum required to unambiguously address the entire array.

PR #8202, opened 2025-02-07
[HW] Resolve ambiguous array_slice documentation

maerhart:
The documentation explicitly states that an OOB access is UB, but also says that it is equivalent to the verilog example. However, OOB in verilog would return the default value of the type (e.g., 0 for 2-state) AFAIU, so it has different semantics. Also, I think it's better to keep Verilog analogies out of the core dialect documentation in general


uenoku (approved):
This looks great to me!

PR #8203, opened 2025-02-07
[FIRRTL] Add optional yaml parameter to view intrinsic, op

dtzSiFive:
Per View indication of whether and where to emit YAML description.

Provides access to mechanism currently encoded via `sifive.enterprise.grandcentral.GrandCentralHierarchyFileAnnotation`.

(new) Views with this parameter will have YAML output to the specified path.
Otherwise, fallback to behavior for annotation-based views (using the value from that annotation, if specified).

This is needed to separate out which Views have YAML emitted, so that some may have no YAML or emit it elsewhere (and to allow dropping the above annotation eventually).


test/Dialect/FIRRTL/lower-intrinsics.mlir
```diff
@@ -228,5 +228,9 @@ firrtl.circuit "Foo" {
...
+
```
seldridge, line 233: Nit: can the PR change this to use a minimal `info` field instead of a verbose one which is ignored in the following `CHECK`?

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -1915,6 +1916,11 @@ void GrandCentralPass::runOnOperation() {
     llvm::dbgs() << "\n";
   });
 
+  // per-YAML output list of interfaces.
+  SmallDenseMap<StringAttr, SmallVector<sv::InterfaceOp, 2>> interfaceYAMLMap;
```
seldridge, line 1920: This `interfaceYAMLMap` is later iterated over. While this won't matter for file creation later, it can create non-deterministic IR creation (e.g., breaking tests or requiring use of `CHECK-DAG` to workaround it). Is there a data structure that will provide a stable iteration order that can be used?

seldridge (commented):
Generally looks good! One question about iteration order.

Second question: the PR text mentions that the fallback behavior is to use the YAML file provided by the annotation. It seems better to have the intrinsic only use its own YAML information so that we can compartmentalize the two overlapping features. WDYT?

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
dtzSiFive: Oh, great call! Yeah let's make this deterministic, thanks!

dtzSiFive:
> Second question: the PR text mentions that the fallback behavior is to use the YAML file provided by the annotation. It seems better to have the intrinsic only use its own YAML information so that we can compartmentalize the two overlapping features.

Sounds good! I went back and forth but honestly that sounds best to me as well, thanks!

seldridge (approved):

PR #8208, opened 2025-02-07
[Verif] Mark SymbolicValueOp result as MemAlloc

fabianschuiki:
Add the `MemAlloc` side effect to the result of `verif.symbolic_value`. This allows the canonicalizer to delete the op if it has no users while still preventing CSE from mergin multiple symbolic values.

Also add a few placeholders to the `lower-contracts.mlir` test to make it more robust.


dobios (approved):
Nice! This is a great idea!! Thanks!!!

Issue #8176, opened 2025-02-03
[MooreToCore] Crash when getting values to observe


maerhart:
The following crashes due to an unattached region when calling `getValuesToObserve`.

```mlir
moore.module @crash(in %in0: !moore.i32, in %in1: !moore.i32) {
  %var = moore.variable : <!moore.i32>
  moore.procedure always_comb {
    %0 = moore.pows %in0, %in1 : !moore.i32
    moore.blocking_assign %var, %0 : !moore.i32
    moore.return
  }
}
```

hailongSun2000:
As far as I analyzed preliminarily, `in0` and `in1` are **block arguments**, and then we'll get the **\<\<UNLINKED BLOCK\>\>** when executing `value->getParentRegion()`
Appendix:
```
/// llvm/mlir/lib/IR/Value.cpp:41

/// Return the Region in which this Value is defined.
Region *Value::getParentRegion() {
  if (auto *op = getDefiningOp())
    return op->getParentRegion();
  return llvm::cast<BlockArgument>(*this).getOwner()->getParent();
}
```
`llvm::cast<BlockArgument>(*this).getOwner()` returns UNLIKED BLOCK.

maerhart:
I'm not really sure if it's enough to just check if the region is attached. Can we be sure that if the region is detached the value is defined outside the process or not necessarily?

We may have to analyze and cache the list of values to observe at every `wait_event` before calling the dialect converter, and pass a reference to this cache to the conversion pattern to use it there. The rewriter has a `getRemappedValue` method to get the value in the target IR for each of the values in the cache. However, it may return a NULL value if the defining operation was not processed yet, so if the value is defined after the procedure that might be the case. Maybe use `materializeTargetConversion' instead?

If we can't get things working reliably inside a pattern we might have to move the procedure lowering outside the dialect conversion (lowering the procedure boundaries before the dialect conversion).

hailongSun2000:
Like this 🤔 ?
```
moore.module @crash(in %in0: !moore.i32, in %in1: !moore.i32) {
  %var = moore.variable : <!moore.i32>
  moore.procedure always_comb {
    %0 = moore.read %a : <!moore.i32>
    %1 = moore.read %in0_0 : <i32>
    %2 = moore.pows %0, %1 : !moore.i32
    moore.blocking_assign %var, %2 : !moore.i32
    moore.return
  }
  // variable declaration
  %a = moore.variable : <!moore.i32>
[...]

PR #8210, opened 2025-02-08
[MooreToCore]Fix a crash caused by block args as observed values for llhd.wait.

hailongSun2000:
Fix this issue: https://github.com/llvm/circt/issues/8176

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -195,6 +195,9 @@ static void getValuesToObserve(Region *region,
   region->getParentOp()->walk<WalkOrder::PreOrder, ForwardDominanceIterator<>>(
       [&](Operation *operation) {
         for (auto value : operation->getOperands()) {
+          if (isa<BlockArgument>(value))
+            value = rewriter.getRemappedValue(value);
+
           if (region->isAncestor(value.getParentRegion()))
             continue;
           if (auto *defOp = value.getDefiningOp();
--- a/test/Conversion/MooreToCore/basic.mlir
+++ b/test/Conversion/MooreToCore/basic.mlir
@@ -1097,3 +1097,22 @@ moore.module @scfInsideProcess(in %in0: !moore.i32, in %in1: !moore.i32) {
     moore.return
   }
 }
+
+// CHECK-LABEL: @blockArgAsObservedValue
+moore.module @blockArgAsObservedValue(in %in0: !moore.i32, in %in1: !moore.i32) {
+  %var = moore.variable : <!moore.i32>
+  // CHECK: [[PRB:%.+]] = llhd.prb %var : !hw.inout<i32>
+  // CHECK: llhd.process
+  moore.procedure always_comb {
+    // CHECK: ^bb1:  // 2 preds: ^bb0, ^bb5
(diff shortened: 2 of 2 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -195,6 +195,9 @@ static void getValuesToObserve(Region *region,
   region->getParentOp()->walk<WalkOrder::PreOrder, ForwardDominanceIterator<>>(
       [&](Operation *operation) {
         for (auto value : operation->getOperands()) {
+          if (llvm::dyn_cast<BlockArgument>(value))
```
maerhart, line 198:
```suggestion
          if (isa<BlockArgument>(value))
```

test/Conversion/MooreToCore/basic.mlir
```diff
@@ -1097,3 +1097,16 @@ moore.module @scfInsideProcess(in %in0: !moore.i32, in %in1: !moore.i32) {
...
   }
 }
+
+// CHECK-LABEL: @blockArgAsObservedValue
+moore.module @blockArgAsObservedValue(in %in0: !moore.i32, in %in1: !moore.i32) {
+  %var = moore.variable : <!moore.i32>
+  moore.procedure always_comb {
+    // CHECK: [[COND:%.+]] = comb.icmp slt %in1, %{{.*}} : i32
+    // CHECK: [[IN0:%.+]] = comb.mux [[COND]], %{{.*}}, %in0 : i32
+    // CHECK: [[IN1:%.+]] = comb.mux [[COND]], %{{.*}}, %in1 : i32
```
maerhart, line 1107: Nit: it might be good to check for the `llhd.wait` op and the values that got added to its `observed` list.

maerhart (approved):
Thanks a lot for digging into that!

lib/Conversion/MooreToCore/MooreToCore.cpp
hailongSun2000: :sweat_smile: I forgot to modify it! Thanks for checking.

test/Conversion/MooreToCore/basic.mlir
```diff
@@ -1097,3 +1097,22 @@ moore.module @scfInsideProcess(in %in0: !moore.i32, in %in1: !moore.i32) {
...
+  moore.procedure always_comb {
+    // CHECK: ^bb1:  // 2 preds: ^bb0, ^bb5
+      // CHECK: [[COND:%.+]] = comb.icmp slt %in1, %{{.*}} : i32
+      // CHECK: comb.mux [[COND]], %{{.*}}, %in0 : i32
+      // CHECK: comb.mux [[COND]], %{{.*}}, %in1 : i32
+    %0 = moore.pows %in0, %in1 : !moore.i32
+    moore.blocking_assign %var, %0 : !moore.i32
+    
+    // CHECK: ^bb5:  // pred: ^bb4
+    // CHECK:   llhd.wait (%in0, %in1, [[PRB]] : i32, i32, i32), ^bb1
```
hailongSun2000, line 1115: `[[PRB]]` shouldn't be observed, right?

maerhart: That's a great catch! It shouldn't. We need to exclude operands that have a memory write effect. We can create an issue and address it in another PR.
Thanks for adding this to the test!

PR #8213, opened 2025-02-10
[ImportVerilog] Fix assertion with constant folding

likeamahoney:
This PR fixes cases when selector maximum or minimum values do not fit into the range of the type for the folding `Moore` constant. 

CIRCT debug build fails with an [assertion](https://github.com/llvm/llvm-project/blob/aa580c2ec5eb4217c945a47a561181be7e7b1032/llvm/include/llvm/ADT/APInt.h#L120): 

```console
circt/llvm/llvm/include/llvm/ADT/APInt.h:120: llvm::APInt::APInt(unsigned int, uint64_t, bool, bool):
Assertion llvm::isIntN(BitWidth, val) && "Value is not an N-bit signed value"' failed.
```

With this test:
```verilog
module RangeElementSelection(
    input reg [3:0] a [0:2],
    output reg [3:0] b,
    input reg [1:0] c);
  
    always_comb begin
      b = a[c]; // Assertion here
      b[3:0] = b[c[0]-:2]; // Assertion here
    end
endmodule
```

This can be fixed by increasing `intType` size by `1` (which passed to `ConstantOp` builder) in order to include both maximum and minimum positive and negative numbers or making a builder sensitive to signedness.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -489,7 +489,7 @@ def ConstantOp : MooreOp<"constant", [Pure, ConstantLike]> {
   let builders = [
     OpBuilder<(ins "IntType":$type, "const FVInt &":$value)>,
     OpBuilder<(ins "IntType":$type, "const APInt &":$value)>,
-    OpBuilder<(ins "IntType":$type, "int64_t":$value)>,
+    OpBuilder<(ins "IntType":$type, "int64_t":$value, CArg<"bool", "true">:$isSigned)>,
   ];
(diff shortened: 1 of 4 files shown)


maerhart (approved):
Oops, seems like I didn't cover all cases when adding this.
Thanks for fixing it! 💯 

Let me know if you want me to merge it for you.

likeamahoney:
> Oops, seems like I didn't cover all cases when adding this. Thanks for fixing it! 💯
> 
> Let me know if you want me to merge it for you.

Thank you! PR ready to merge!

PR #8220, opened 2025-02-11
[Moore] Mark wait_event with side-effect even if it is empty

maerhart:
If an `always @(*)` procedure does not read any values, an empty wait_event is inserted by ImportVerilog. Retaining this is important because it means the procedure suspends indefinitely at that point.

Also, lower an empty wait_event directly to a 'halt' op because it's more efficient in simulation and code size, and makes no difference in behavior.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -390,7 +390,6 @@ def EdgeAttr: I32EnumAttr<"Edge", "Edge kind",
 }
 
 def WaitEventOp : MooreOp<"wait_event", [
-  RecursiveMemoryEffects,
   NoRegionArguments,
   SingleBlock,
   NoTerminator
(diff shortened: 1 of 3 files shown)


fabianschuiki (approved):
Very nice! LGTM

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -390,7 +390,6 @@ def EdgeAttr: I32EnumAttr<"Edge", "Edge kind",
 }
 
 def WaitEventOp : MooreOp<"wait_event", [
-  RecursiveMemoryEffects,
```
hailongSun2000, line 393: Just one question. I'm confused about this trait. I always thought if there are child operations with the side-effect in its region, it should be marked `RecursiveMemoryEffects`.

fabianschuiki: That is true if the outer operation itself has no side effect. For example, if an `scf.if` contains no side-effecting op, the `scf.if` itself has no side effect. Therefore `RecursiveMemoryEffects` are appropriate. However, the `WaitEventOp` itself has a side effect: it blocks execution and waits. That's why it doesn't need `RecursiveMemoryEffects`: it is side-effecting regardless of the operations it contains.

hailongSun2000: Appreciate your elaboration greatly! :+1:

Issue #8212, opened 2025-02-10
[arcilator] Install configuration missing for arcilator-runtime.h


maerhart:
The arcilator-runtime.h and python script are not installed and thus not included in pre-built releases.

For more information:
https://discourse.llvm.org/t/proper-way-to-obtain-arcilator-runtime-h-for-c-integration/84553

PR #8223, opened 2025-02-11
[arcilator] Add missing install configuration for arcilator-runtime.h and arcilator-header-cpp.py

FurryAcetylCoA:
See #8212 
>The arcilator-runtime.h and python script are not installed and thus not included in pre-built releases.

I'm not entirely sure about the correct location for arcilator-runtime.h, but it seems that the header generated by arcilator-header-cpp.py expects arcilator-runtime.h to be located at the root of the `include/` directory.

Additionally, the add_custom_target command is unnecessary and has been removed.


FurryAcetylCoA:
friendly ping @maerhart

PR #8224, opened 2025-02-11
[Verif] Add simulation op

fabianschuiki:
Add the `verif.simulation` op as an entry point for simulations. The op works in a fashion similar to `verif.formal`, but encodes a simulation run instead of a formal check.


include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -346,6 +349,105 @@ def FormalOp : VerifOp<"formal", [
...
+    the simulation. The simulation stops when the done signal is 1 during a
+    0-to-1 transition of the clock.
+
+    The second operand is an "exit code" signal which may be of any integer
+    type. It indicates success of a test as 0, or failure as any non-zero
+    integer. Only the zero versus non-zero value distinction is relevant to
+    indicate simulation success and must be preserved. The exact non-zero value
+    is for informational purposes only and may not be reproduced accurately by
+    simulators. Simulators may in the worst case map all non-zero values to 1,
+    reducing the exit code to a single bit indicating failure or success.
```
maerhart, line 388: If we can't give any guarantees about the exit code anyway, should we just require `i1`? Or is it quite rare that simulators don't preserve it?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -346,6 +349,105 @@ def FormalOp : VerifOp<"formal", [
...
+    | t0 (>=0s) | undef | undef | Clock and init may initially be undefined.
+    | t1 (>=t0) | 0     | 1     | Initialization code (e.g., `seq.initial`, Verilog `initial` procedures) may run before or after clock and init change to their initial value.
+    | t2 (>t1)  | 1     | 1     | Single rising clock edge occurs while init is high.
+    | t3 (>t2)  | 0     | 1     | Init may stay during the falling clock edge.
+    | t4 (>=t3) | 0     | 0     | Init goes to 0 before second rising clock edge.
+    | t5 (>t4)  | 1     | 0     | Clock toggles continue indefinitely.
+    | t6 (>t5)  | 0     | 0     |
+
+    Simulation termination occurs when the done signal is 1 during a 0-to-1
+    transition of the clock. The exit code is sampled at the same time as the
```
maerhart, line 406: This sentence seems to be a duplicate from above

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -346,6 +349,105 @@ def FormalOp : VerifOp<"formal", [
   }];
 }
 
+def SimulationOp : VerifOp<"simulation", [
+  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmBlockArgumentNames"]>,
+  IsolatedFromAbove,
+  NoTerminator,
```
maerhart, line 355: Why `NoTerminator` if we require a `verif.yield`? I'm actually a bit surprised that this didn't lead to any issues.

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -346,6 +349,105 @@ def FormalOp : VerifOp<"formal", [
   }];
 }
 
+def SimulationOp : VerifOp<"simulation", [
```
maerhart, line 352: What's the rationale of having this in the `verif` dialect vs `sim` dialect?

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -346,6 +349,105 @@ def FormalOp : VerifOp<"formal", [
...
+
+      // Check results and track failures.
+      %4 = comb.add %1, %2 : i42
+      %5 = comb.icmp ne %3, %4 : i42
+      %false = hw.constant false
+      %failure = seq.compreg %6, %clock reset %init, %false : i1
+      %6 = comb.or %failure, %5 : i1
+
+      verif.yield %done, %failure : i1, i1
+    }
```
maerhart, line 434: Is the body supposed to be a hardware implementation of a testbench or are you allowed to do any kind of software stuff? I guess not, because it's a graph region.

lib/Dialect/Verif/VerifOps.cpp
```diff
@@ -159,6 +159,37 @@ LogicalResult BoundedModelCheckingOp::verifyRegions() {
...
+//===----------------------------------------------------------------------===//
+// SimulationOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult SimulationOp::verifyRegions() {
+  if (getBody()->getNumArguments() != 2)
+    return emitOpError() << "must have two block arguments";
+  auto clockType = seq::ClockType::get(getContext());
+  auto i1Type = IntegerType::get(getContext(), 1);
+  if (getBody()->getArgumentTypes() != std::array<Type, 2>{clockType, i1Type})
```
maerhart, line 171: Nit: Maybe cheaper to do `isa<ClockType>(type)` and `type.isSignlessInteger(1)` checks rather than building new types?

include/circt/Dialect/Verif/VerifOps.td
fabianschuiki: Yeah agreed, the lines are somewhat blurry here. I think it would make sense to keep the verification intent operations in the same dialect, since they are similar and might share documentation and implementation bits in the future. That's why I opted for adding a `verif.simulation` instead of a `sim.simulation`. I agree though that we should use the `sim` dialect for all other simulation-specific ops. For example, if we ever add something like a stimulus / procedural test sequence op, it would probably be something like `sim.execute_region { ... }` or similar.

include/circt/Dialect/Verif/VerifOps.td
fabianschuiki: Yikes, my bad. Pretty interesting that this didn't cause an issue anywhere :stuck_out_tongue:

include/circt/Dialect/Verif/VerifOps.td
fabianschuiki: For most simulators we'll be able to preserve this pretty accurately. Arcilator and Verilator for example can definitely reproduce this, modulo some operating system limitations.

But you do have a point, if there's no guarantee here, then we might just want to make it an `i1` in general and not burden implementations with trying to funnel these codes through. WDYT? Should this be dropped? @seldridge any thoughts here?

include/circt/Dialect/Verif/VerifOps.td
fabianschuiki: Removed

include/circt/Dialect/Verif/VerifOps.td
fabianschuiki: It's basically the same as a `hw.module` or `verif.formal` op. Any software specifics would have to be packaged up in an op that can describe how a SSACFG region can be executed and can update some form of output values it produces, maybe something like `sim.execute_region` with multiple `sim.yield` resume points.

lib/Dialect/Verif/VerifOps.cpp
fabianschuiki: Nice idea, done

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -346,6 +349,104 @@ def FormalOp : VerifOp<"formal", [
...
+    clock, and then drops to 0 for the remainder of the simulation.
+
+    ### Outputs
+
+    The body must have a `verif.yield` terminator op with exactly two operands:
+
+    The first operand is a "done" signal of type `i1` which indicates the end of
+    the simulation. The simulation stops when the done signal is 1 during a
+    0-to-1 transition of the clock. No additional clock toggles occur once done
+    has been sampled as 1.
```
TaoBi22, line 380: super-nit for readability
```suggestion
    The first operand is a "done" signal of type `i1` which indicates the end of
    the simulation. The simulation stops when the done signal is 1 during a
    0-to-1 transition of the clock. No additional clock toggles occur once the
    done signal has been sampled as 1.
```

TaoBi22 (approved):
Exciting! :tada:

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -346,6 +349,104 @@ def FormalOp : VerifOp<"formal", [
...
+
+    The clock and init values adhere to the following schedule during
+    simulation:
+
+    | Time      | Clock | Init  | Comment
+    |-----------|-------|-------| ----
+    | t0 (>=0s) | undef | undef | Clock and init may initially be undefined.
+    | t1 (>=t0) | 0     | 1     | Initialization code (e.g., `seq.initial`, Verilog `initial` procedures) may run before or after clock and init change to their initial value.
+    | t2 (>t1)  | 1     | 1     | Single rising clock edge occurs while init is high.
+    | t3 (>t2)  | 0     | 1     | Init may stay during the falling clock edge.
```
TaoBi22, line 401:
```suggestion
    | t3 (>t2)  | 0     | 1     | Init may stay high during the falling clock edge.
```

fabianschuiki:
Landed this as is for now. We may want to revisit the exit code in the future and potentially change it to an `i1`, since it's not clear how useful an exit code is if the OS and simulators don't really guarantee how non-zero values are preserved.

PR #8225, opened 2025-02-12
Bump LLVM to 289b17635958d986b74683c932df6b1d12f37b70.

mikeurbach:
This picked up a few breaking changes.

https://github.com/llvm/llvm-project/commit/327d6270

This added new required methods to CallOpInterface and CallableOpInterface. These methods are related to arrays of dictionary attributes for each argument and result, and are normally implemented upstream by simply defining the right arg_attrs and res_attrs arguments on the operations. This allows some of the common utilities upstream to work with the attributes.

However, we seem to have actually moved away from arrays of per-argument or per-result attributes in most cases, and handle this ourselves when we want to. I opted to stub out the methods in the tablegen definitions of our operations, rather than adding optional attributes and setting them to null in builders and builder callsites like was done upstream.

There were also some accompanying name changes to a couple helper functions, which we were able to simply apply without needing to add new optional attributes everywhere.

https://github.com/llvm/llvm-project/commit/e84f6b6a

This updated several LLVM dialect helper functions related to lookupOrCreateFn to return FailureOr<LLVM::LLVMFuncOp>. In most cases, we simply check for the error case and return it or signal pass failure. There was one function in the SMT to Z3 conversion that assumes this always succeeds and in this function I made an assertion rather than changing the function to also potentially return failure.

https://github.com/llvm/llvm-project/commit/f4e3b878

Looks like upstream is simply moving from undef to poison in many cases, so this required us to update one test case.

### DIFF
--- a/include/circt/Dialect/Arc/ArcOps.td
+++ b/include/circt/Dialect/Arc/ArcOps.td
@@ -237,6 +237,16 @@ def StateOp : ArcOp<"state", [
     void setCalleeFromCallable(mlir::CallInterfaceCallable callee) {
       (*this)->setAttr(getArcAttrName(), llvm::cast<mlir::SymbolRefAttr>(callee));
     }
+
+    /// Stub implementations for ArgumentAttributesMethods. If desired,
+    /// implement these by defining arg_attrs and res_attrs as arguments to the
+    /// operation as OptionalAttr<DictArrayAttr>.
+    mlir::ArrayAttr getArgAttrsAttr() { return nullptr; }
+    mlir::ArrayAttr getResAttrsAttr() { return nullptr; }
+    void setArgAttrsAttr(mlir::ArrayAttr args) {}
+    void setResAttrsAttr(mlir::ArrayAttr args) {}
+    mlir::Attribute removeArgAttrsAttr() { return nullptr; }
+    mlir::Attribute removeResAttrsAttr() { return nullptr; }
   }];
 }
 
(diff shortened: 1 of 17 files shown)


mikeurbach:
I also have a version of this where I handled https://github.com/llvm/llvm-project/commit/327d627066e6452b081365b595657d17f2690a3b by adding `arg_attrs` and `res_attrs` like they did here, for example: https://github.com/llvm/llvm-project/commit/327d627066e6452b081365b595657d17f2690a3b#diff-55d842f400d430d940f6036970f462764765330d7a617750cd21e7d499b1c5da

But this seemed like a step back for us... I think we pretty actively moved away from the per-argument / per-result attributes the way upstream does it on our ModuleLikes at least. I don't know what we are currently gaining by using these CallOpInterface and CallableOpInterfaces, but in the past I remember that trying to shoehorn HWModuleOp into the old FunctionLike interface was a problem. It seems better to me to stub out the new required methods, since they explicitly say they can just return null. They're mainly for IR printing / parsing helpers from what I can tell, so I don't think we're missing anything by opting out. But I leave it to the various codeowners to implement these by defining the attribute arrays if they want in a follow up PR.

maerhart (approved):
Seems like there is one more use of `mlir-cpu-runner` that needs to be renamed. But otherwise LGTM! Completely agree with your approach about the arg/res attrs!

fabianschuiki:
Thanks for fixing all of these! :partying_face: :+1: We haven't really used arg/result attributes in the Arc dialect at all, so the stubs are perfect.

dtzSiFive:
Thanks for the great work and write-up! LGTM!

mikeurbach:
Will plan to merge this EOD today once CI is passing.

uenoku (approved):
SV/Sim change looks great to me!

maerhart:
I need https://github.com/llvm/llvm-project/commit/9a63a2c4babcaabaeb1a106345f3b037485efa4c
Should I do another bump within this PR or follow up with another one?

PR #8230, opened 2025-02-12
[LLHD][TCM] Fix CFG loops detection

AndreyVV-100:
Hi! This PR fixes CFG loops detection. In the current algorithm confluence block is pushed to workset twice, I fixed that by memorizing control flow.


AndreyVV-100:
For context: below sv code example crashes arcilator on the current main branch.
```verilog
module mod (input clk, input rst, output reg a, output reg b);
  always_ff @(posedge clk or posedge rst) begin
    if (rst)
        a <= 0;
    b <= 1;
  end
endmodule
```

```console
circt-verilog mod.sv | arcilator

<stdin>:13:5: error: 'llhd.process' op has regions; not supported by ConvertToArcs
```

PR #8233, opened 2025-02-12
[Arcilator] Allow running verif.simulation ops

fabianschuiki:
Add the `LowerVerifSimulations` pass to the Arc dialect. This pass converts any `verif.simulation` ops in the input into a test harness module and a `func.func` that instantiates the module and applies the necessary stimulus to it using `arc.sim.*` ops.

Also add the pass to the arcilator pipeline. This allows the user to run any `verif.simulation` ops in the input design by specifying the `--run --jit-entry=<simulation-name>` command line options. We may want to streamline this in the future.

Also add an integration test which verifies an adder circuit and prints the simulation results.

### DIFF
--- a/integration_test/arcilator/JIT/verif-simulation.mlir
+++ b/integration_test/arcilator/JIT/verif-simulation.mlir
@@ -1,0 +1,86 @@
+// RUN: arcilator --run --jit-entry=AdderTest %s | FileCheck %s
+// REQUIRES: arcilator-jit
+
+// CHECK: simulation started
+// CHECK: simulation passed
+
+verif.simulation @AdderTest {} {
+^bb0(%clock: !seq.clock, %init: i1):
+  // Count the first 9001 simulation cycles.
+  %c0_i19 = hw.constant 0 : i19
+  %c1_i19 = hw.constant 1 : i19
+  %c9001_i19 = hw.constant 9001 : i19
+  %count = seq.compreg %0, %clock reset %init, %c0_i19 : i19
+  %done = comb.icmp eq %count, %c9001_i19 : i19
+  %0 = comb.add %count, %c1_i19 : i19
+
+  // Generate inputs to the adder.
+  %1, %2 = sim.func.dpi.call @generateAdderInputs(%count) : (i19) -> (i42, i42)
+  %3 = hw.instance "dut" @Adder(a: %1: i42, b: %2: i42) -> (c: i42)
+
+  // Check results and track failures.
+  %4 = comb.add %1, %2 : i42
+  %5 = comb.icmp ne %3, %4 : i42
+  %false = hw.constant false
+  %failure = seq.compreg %6, %clock reset %init, %false : i1
+  %6 = comb.or %failure, %5 : i1
+
+  // Print some statistics about the test.
+  sim.func.dpi.call @printStart() clock %clock enable %init : () -> ()
+  sim.func.dpi.call @printFailure(%1, %2, %3, %4) clock %clock enable %failure : (i42, i42, i42, i42) -> ()
+  sim.func.dpi.call @printDone(%failure) clock %clock enable %done : (i1) -> ()
+
+  verif.yield %done, %failure : i1, i1
+}
+
+hw.module private @Adder(in %a: i42, in %b: i42, out c: i42) {
+  %0 = comb.add %a, %b : i42
+  hw.output %0 : i42
+}
+
+sim.func.dpi @generateAdderInputs(in %idx: i19, out a: i42, out b: i42) attributes {verilogName = "generateAdderInputs.impl"}
+func.func private @generateAdderInputs.impl(%arg0: i19, %arg1: !llvm.ptr, %arg2: !llvm.ptr) {
+  %c4224496855063_i42 = arith.constant 4224496855063 : i42
+  %c1_i42 = arith.constant 1 : i42
+  %0 = arith.extui %arg0 : i19 to i42
+  %1 = arith.addi %0, %c1_i42 : i42
+  %a = arith.muli %0, %c4224496855063_i42 : i42
+  %b = arith.muli %1, %c4224496855063_i42 : i42
+  llvm.store %a, %arg1 : i42, !llvm.ptr
+  llvm.store %b, %arg2 : i42, !llvm.ptr
+  return
+}
+
+sim.func.dpi @printStart() attributes {verilogName = "printStart.impl"}
+sim.func.dpi @printDone(in %failure: i1) attributes {verilogName = "printDone.impl"}
+sim.func.dpi @printFailure(in %a: i42, in %b: i42, in %c_act: i42, in %c_exp: i42) attributes {verilogName = "printFailure.impl"}
+
+func.func private @printStart.impl() {
+  %0 = llvm.mlir.addressof @str.start : !llvm.ptr
+  call @puts(%0) : (!llvm.ptr) -> ()
+  return
+}
+
+func.func private @printFailure.impl(%a: i42, %b: i42, %c_act: i42, %c_exp: i42) {
+  %0 = llvm.mlir.addressof @str.mismatch : !llvm.ptr
+  call @puts(%0) : (!llvm.ptr) -> ()
+  arc.sim.emit "a", %a : i42
+  arc.sim.emit "b", %b : i42
+  arc.sim.emit "c_act", %c_act : i42
+  arc.sim.emit "c_exp", %c_exp : i42
+  return
+}
+
+func.func private @printDone.impl(%failure: i1) {
+  %0 = llvm.mlir.addressof @str.pass : !llvm.ptr
+  %1 = llvm.mlir.addressof @str.fail : !llvm.ptr
+  %2 = llvm.select %failure, %1, %0 : i1, !llvm.ptr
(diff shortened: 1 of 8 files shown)


integration_test/arcilator/JIT/verif-simulation.mlir
```diff
@@ -0,0 +1,83 @@
...
+  %3 = hw.instance "dut" @Adder(a: %1: i42, b: %2: i42) -> (c: i42)
+
+  // Check results and track failures.
+  %4 = comb.add %1, %2 : i42
+  %5 = comb.icmp ne %3, %4 : i42
+  %false = hw.constant false
+  %failure = seq.compreg %6, %clock reset %init, %false : i1
+  %6 = comb.or %failure, %5 : i1
+
+  // Print some statistics about the test.
```
uenoku, line 28: Nice use of DPI!

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  // If done, exit the loop.
+  auto &exitBlock = execOp.getRegion().emplaceBlock();
+  builder.create<cf::CondBranchOp>(op.getLoc(), doneSample, &exitBlock,
+                                   &loopBlock);
+  builder.setInsertionPointToEnd(&exitBlock);
+
+  // Compute `exit_code | (exit_code != 0)` as a way of guaranteeing that
+  // the exit code will be non-zero even if bits are truncated by the operating
+  // system.
```
maerhart, line 178: Smart way to deal with the fact that exit typically ANDs with 0xFF. Maybe worth mentioning in the simulation op documentation that on non-zero exit code the least significant bit is always set? But I guess also not really, because we are very vague about that anyway.

test/Dialect/Arc/lower-verif-simulations.mlir
```diff
@@ -0,0 +1,83 @@
...
+// CHECK: arc.sim.instantiate @verif.simulation.impl.Foo as [[A:%.+]] {
+// CHECK:   scf.execute_region {
+// CHECK:     arc.sim.set_input [[A]], "clock" = [[C0]]
+// CHECK:     arc.sim.set_input [[A]], "init" = [[I1]]
+// CHECK:     arc.sim.step [[A]]
+// CHECK:     arc.sim.set_input [[A]], "clock" = [[C1]]
+// CHECK:     arc.sim.step [[A]]
+// CHECK:     arc.sim.set_input [[A]], "clock" = [[C0]]
+// CHECK:     arc.sim.set_input [[A]], "init" = [[I0]]
+// CHECK:     arc.sim.step [[A]]
```
maerhart, line 34: I think the simulation op description didn't mention that done is not checked when init=1. Which, of course, makes sense, but might be worth mentioning.

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
```diff
@@ -0,0 +1,205 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Arc/ArcOps.h"
+#include "circt/Dialect/Arc/ArcPasses.h"
+#include "circt/Dialect/Comb/CombOps.h"
```
maerhart, line 11: Unused?

integration_test/arcilator/JIT/verif-simulation.mlir
```diff
@@ -0,0 +1,83 @@
...
+  // Count the first 9001 simulation cycles.
+  %c0_i19 = hw.constant 0 : i19
+  %c1_i19 = hw.constant 1 : i19
+  %c9001_i19 = hw.constant 9001 : i19
+  %count = seq.compreg %0, %clock reset %init, %c0_i19 : i19
+  %done = comb.icmp eq %count, %c9001_i19 : i19
+  %0 = comb.add %count, %c1_i19 : i19
+
+  // Generate inputs to the adder.
+  %1, %2 = func.call @generateAdderInputs(%count) : (i19) -> (i42, i42)
```
maerhart, line 18: Strictly speaking this would need to be a module instantiation since the region describes hardware, right? And the function would need to use comb instead of arith.

maerhart (approved):
Really cool work! 😎

integration_test/arcilator/JIT/verif-simulation.mlir
fabianschuiki: Yeah that's a good point. This happens to survive the Arcilator pipeline unharmed. A DPI call would probably fit better here, to reflect what you would type up in some SV input.

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
```diff
@@ -0,0 +1,205 @@
...
+    auto builder = OpBuilder::atBlockBegin(getOperation().getBody());
+    auto func = builder.create<func::FuncOp>(
+        getOperation().getLoc(), builder.getStringAttr("exit"),
+        builder.getFunctionType({builder.getI32Type()}, {}));
+    SymbolTable::setSymbolVisibility(func, SymbolTable::Visibility::Private);
+  }
+
+  SmallVector<StringAttr> symbols;
+  getOperation().walk([&](verif::SimulationOp op) {
+    auto *symbolTableOp = SymbolTable::getNearestSymbolTable(op);
```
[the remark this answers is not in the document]
fabianschuiki: That would definitely be an option! I wasn't entirely sure though if we support `HWModule`s nested in other operations – I think we might, for stuff like `sv.ifdef` and similar. But we could definitely just force the `verif.*` tests to be in a `builtin.module`.

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
```diff
@@ -0,0 +1,205 @@
...
+    assert(symbolTableOp);
+    if (symbolTableOp == getOperation())
+      symbols.push_back(op.getSymNameAttr());
```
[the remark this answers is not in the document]
fabianschuiki: Ooops… originally thought about creating a "runAll" function that would execute all simulation tests, but that doesn't work well with the `exit` call in there, and it would also partially overlap with what `circt-test` does.

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
fabianschuiki: Removed

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  // Declare the `exit` function if it does not yet exist.
+  auto &symbolTable = symbolTables.getSymbolTable(getOperation());
+  if (!symbolTable.lookup("exit")) {
+    auto builder = OpBuilder::atBlockBegin(getOperation().getBody());
+    auto func = builder.create<func::FuncOp>(
+        getOperation().getLoc(), builder.getStringAttr("exit"),
+        builder.getFunctionType({builder.getI32Type()}, {}));
+    SymbolTable::setSymbolVisibility(func, SymbolTable::Visibility::Private);
+  }
```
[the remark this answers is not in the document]
fabianschuiki: Good point. Added a check for that.

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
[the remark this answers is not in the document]
fabianschuiki: Remove this now.

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
```diff
@@ -0,0 +1,205 @@
...
+      builder.create<hw::HWModuleOp>(op.getLoc(), implName, implPorts);
+  symbolTable.insert(implOp);
+  implOp.getBody().takeBody(op.getBodyRegion());
+
+  // Create a new function for the verification op.
+  auto funcType = builder.getFunctionType({}, {});
+  auto funcOp =
+      builder.create<func::FuncOp>(op.getLoc(), op.getSymNameAttr(), funcType);
+  auto &funcBody = funcOp.getBody().emplaceBlock();
+  builder.setInsertionPointToEnd(&funcBody);
```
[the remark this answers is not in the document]
fabianschuiki: Used that here

lib/Dialect/Arc/Transforms/LowerVerifSimulations.cpp
fabianschuiki: Yeah I'm less and less convinced that we should actually have a multi-bit exit code at all. Maybe a `success` line would be better…

test/Dialect/Arc/lower-verif-simulations.mlir
fabianschuiki: Good point. Added that.

PR #8234, opened 2025-02-13
[circt-verilog-lsp-server] Add Verilog Language Server

uenoku:
This is a first PR to add a Language Server Protocol (LSP) implementation for Verilog/SystemVerilog using the slang frontend. 

The server is built as circt-verilog-lsp-server and eventually integrates with CIRCT's existing Verilog import capabilities. It leverages MLIR's LSP server support libraries for the protocol implementation.

* VerilogServer.cpp - Core language service implementation that manages documents
   and interfaces with slang for parsing and analysis. Maintains document state
   and handles document synchronization events.

* LSPServer.cpp - Protocol layer that implements LSP message handling and converts
   between LSP types and internal representations. Built on MLIR's LSP server
   support libraries.

`lib/Tools/circt-verilog-lsp-server/{Utils, VerilogSeverImpl}` are added as workaround for RTTI build issue. The current version of slang requires RTTI but LLVM may not be built with RTTI. So if LLVM functions use RTTI in VerilogServer.cpp it causes compilation error  because there is no RTTI. As a workaround, I specialized utility functions in LSPUtils.cpp, and use it from VerilogServer.cpp.  
 
Not implemented in this initial PR but the goal is to provide modern LSP features such as hover/go to def/references integrated with CIRCT (e.g. hover source location).

Screenshot from 2025-02-12 16-14-32

Screenshot from 2025-02-12 16-14-45

### DIFF
--- a/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
+++ b/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
@@ -108,7 +108,7 @@ private:
   llvm::SourceMgr sourceMgr;
 
   // The URI of the document.
-  const mlir::lsp::URIForFile &uri;
+  mlir::lsp::URIForFile uri;
 };
 
 } // namespace
@@ -138,8 +138,7 @@ void LSPDiagnosticClient::report(const slang::ReportedDiagnostic &slangDiag) {
   // Show only the diagnostics in the current file.
   if (loc.uri != document.getURI())
     return;
-  diags.emplace_back();
-  auto &mlirDiag = diags.back();
+  auto &mlirDiag = diags.emplace_back();
   mlirDiag.severity = getSeverity(slangDiag.severity);
   mlirDiag.range = loc.range;
   mlirDiag.source = "slang";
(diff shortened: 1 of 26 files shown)


lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -0,0 +1,353 @@
...
+  std::vector<std::string> libDirs;
+  libDirs.push_back(uriDirectory.str().str());
+  libDirs.insert(libDirs.end(), context.options.libDirs.begin(),
+                 context.options.libDirs.end());
+
+  // Populate source managers.
+  const llvm::MemoryBuffer *memBuffer = sourceMgr.getMemoryBuffer(bufferId);
+
+  driver.options.libDirs = libDirs;
+  driver.sourceManager.setDisableProximatePaths(true);
```
uenoku, line 181: This is not necessary
```suggestion
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -0,0 +1,353 @@
...
+  FailureOr<std::unique_ptr<slang::ast::Compilation>> compilation;
+
+  // The slang driver.
+  slang::driver::Driver driver;
+
+  // The LLVM source manager.
+  llvm::SourceMgr sourceMgr;
+
+  // The URI of the document.
+  const mlir::lsp::URIForFile &uri;
```
uenoku, line 111: URI might be deallocated so it's necessary to copy here.
```suggestion
  mlir::lsp::URIForFile uri;
```

PR #8237, opened 2025-02-13
[verif.contract] add sby integration test

leonardt:


fabianschuiki (approved):
LGTM!

integration_test/lit.cfg.py
```diff
@@ -54,7 +54,7 @@
 # directories.
 config.excludes = [
     'Inputs', 'CMakeLists.txt', 'README.txt', 'LICENSE.txt', 'lit.cfg.py',
-    'lit.local.cfg.py'
+    'lit.local.cfg.py', '.circt-test'
```
uenoku, line 57: I think this is not necessary, or do we have a plan to consider `.circt-test` as special directory?

leonardt: Without this we get the following "unresolved" message, but I'm open to other ways to avoid it rather than this config change

image

fabianschuiki: It looks like llvm-lit is looking into `.*` directories when searching for tests. Ideally we'd be able to generally include `.*` directories in lit, but I'm not sure you can have a regex or glob pattern in `config.excludes`.

uenoku: I think one way is to pass a result directory properly (`-d %T`) to `circt-test` instead of using default `.circt-test`. I think using the default `.circt-test` may cause issues when multiple tests are ran parallely.

fabianschuiki: That's a great idea! `-d %T` sounds like a plan :smile:

PR #8241, opened 2025-02-13
[LLHD] Don't implement BranchOpInterface for WaitOp

maerhart:
BranchOpInterface requries the control-flow edges to be side-effect free. However, WaitOp suspends execution and once execution is resumed, values on signals are expected to have changed from outside the process


fabianschuiki (approved):
Definitely a great idea :grin:

PR #8243, opened 2025-02-14
[Python] Declare a separate sub-target for support sources

maerhart:
Makes it easier for sub-projects to only include parts of the python bindings


uenoku (approved):
LGTM

mikeurbach (approved):
Nice, the goal makes sense. Been a while since I looked at this CMake stuff, but I think this is exactly right.

PR #8248, opened 2025-02-17
[Calyx] Unroll Affine ParallelOp to multiple SCF ExecuteRegionOps

jiahanxie353:
This patch unrolls `affine.parallel` op to multiple `scf.execute_region` ops for easier lowering to `calyx.par` op so that we can do more higher-level optimizations and we can make the lowering of `scf.parallel` in the SCFToCalyx` pass trivial.

### DIFF
--- a/lib/Dialect/Calyx/Transforms/AffineParallelUnroll.cpp
+++ b/lib/Dialect/Calyx/Transforms/AffineParallelUnroll.cpp
@@ -1,0 +1,166 @@
+//===- AffineParallelUnroll.cpp - Unroll AffineParallelOp ------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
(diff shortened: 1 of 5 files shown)


lib/Dialect/Calyx/Transforms/AffineParallelUnroll.cpp
```diff
@@ -0,0 +1,166 @@
...
+using namespace mlir::arith;
+
+namespace {
+
+struct AffineParallelUnroll : public OpRewritePattern<AffineParallelOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AffineParallelOp affineParallelOp,
+                                PatternRewriter &rewriter) const override {
+    if (affineParallelOp->hasAttr("calyx.parallel"))
```
cgyurgyik, line 43: Nit: maybe `calyx.unroll` is better suited here?

cgyurgyik (approved):
LGTM. 

> This patch unrolls affine.parallel op to multiple scf.execute_region ops for easier lowering to calyx.par op so that we can do more higher-level optimizations and we can make the lowering of scf.parallel in the SCFToCalyx` pass trivial.

Can you also add the relevant Zulip conversation to this ^ please

jiahanxie353:
> Can you also add the relevant Zulip conversation to this ^ please

Sure no problem!

PR #8252, opened 2025-02-18
[Calyx] virtual dtor to properly destruct derived classes

dtzSiFive:
ASAN caught new/delete type mismatch error.


cgyurgyik (approved):
Thanks!

PR #8253, opened 2025-02-18
[Verif] Change simulation exit code to success boolean

fabianschuiki:
After some discussions on `verif.simulation`, the consensus seems to be that an arbitrary integer exit code that has no guarantee of being preserved properly by simulators, besides the zero vs non-zero distinction, is virtually useless in practice. Instead, simulations can be made simpler by generating a simple `i1` success result indicating whether the test passed or failed. Simulators can then map this to an appropriate exit code. The operating system's potential truncation of exit codes to fewer bits (e.g. 7 or 8 bits on Linux) is no longer an issue.


maerhart (approved):
LGTM!

Issue #2231, opened 2021-11-23
[hw] Revise the design of "doNotPrint"


lattner:
The current support for bind operations is based on an undeclared "doNotPrint" attribute on hw.instance.  This is a bit of a wormhole that isn't documented or really defined in the IR

I think it would be cleaner to:
1) Rename this to "emitAsBind" since that is what it does (it certainly prints!)
2) Declare this in ODS so it is a documented part of the contract that hw.instance provides, and so it gets type checked.
3) Mention this in the HW dialect rationale.

PR #8254, opened 2025-02-18
[Transforms][circt-synth] Add HierarchicalRunner pass

uenoku:
Add a new pass that allows running passes under hierarchy. This is useful when we want to run synthesis passes only on real designs.

This implementation is based on MLIR's upstream [CompositePass](https://github.com/llvm/llvm-project/blob/0afe2bd21b06bed4d48eb88a99d13a768426359c/mlir/lib/Transforms/CompositePass.cpp#L30-L52) and [Inliner utilities](https://github.com/llvm/llvm-project/blob/0afe2bd21b06bed4d48eb88a99d13a768426359c/mlir/lib/Transforms/Utils/Inliner.cpp#L541-L560). The pass takes a pipeline string and runs it on modules in the design hierarchy. It supports configurable options including the pipeline to run under hierarchy, the name of the top-level module, and whether to include bound instances in the hierarchy traversal.

The implementation updates circt-synth tool with the required library dependencies and necessary dialect and pass headers.

### DIFF
--- a/lib/Transforms/HierarchicalRunner.cpp
+++ b/lib/Transforms/HierarchicalRunner.cpp
@@ -86,8 +86,7 @@ void HierarchicalRunnerPass::runOnOperation() {
   auto am = getAnalysisManager();
   while (!worklist.empty()) {
     auto *node = worklist.pop_back_val();
-    if (!node)
-      continue;
+    assert(node && "node should not be null");
     auto op = node->getModule();
     if (!op || !visited.insert(op))
       continue;
--- a/lib/Transforms/HierarchicalRunner.cpp
+++ b/lib/Transforms/HierarchicalRunner.cpp
@@ -1,0 +1,146 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements the HierarchicalRunner pass which runs a pass pipeline
+// on specific hierarchichy.
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Support/InstanceGraph.h"
+#include "circt/Transforms/Passes.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/Threading.h"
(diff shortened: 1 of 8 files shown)


lib/Transforms/HierarchicalRunner.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  SmallVector<igraph::InstanceGraphNode *> worklist;
+  llvm::SetVector<Operation *> visited;
+  worklist.push_back(top);
+
+  auto am = getAnalysisManager();
+  while (!worklist.empty()) {
+    auto *node = worklist.pop_back_val();
+    if (!node)
+      continue;
```
Max-astro, line 90: I'm curious about the pop-out node. In what situation will it be nullptr?

lib/Transforms/HierarchicalRunner.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    // Ensure an analysis manager has been constructed for each of the nodes.
+    // This prevents thread races when running the nested pipelines.
+    am.nest(op);
+
+    for (auto *child : *node) {
+      auto childOp = child->getInstance();
+      if (!childOp ||
+          (!includeBoundInstances && childOp->hasAttr("doNotPrint")))
```
Max-astro, line 102: Does the "doNotPrint" attribute have the same semantics as "set_dont_touch" in SDC?

Max-astro (commented):
Running different passes by module hierarchy is a valuable feature, especially when a designer wants to ungroup (inline) certain kinds of modules by some rules. It's good to know you implemented this feature. Nice job!

lib/Transforms/HierarchicalRunner.cpp
uenoku: "doNotPrint" is used for instances instantiated by SystemVerilog bind. Normally they won't be instantiated in the synthesizable parts. It's not good that we are relying on this vaguely defined attribute, and we have wanted to fix (See https://github.com/llvm/circt/issues/2231) but no effort has been made so far.

fabianschuiki (approved):
LGTM! Nice idea to use the instance graph to only run a bunch of passes on a specific hierarchy :+1:

uenoku:
> Just out of curiosity: have you come across any uses of this in the wild?

It's generally necessary for circt-synth to skip verification code. Sometimes verification code has constructs that are hard to synthesize such as 32bit div/mod.

PR #8255, opened 2025-02-19
[MooreToCore]Remove unexpected observed values for WaitOp.

hailongSun2000:


maerhart:
Thanks @hailongSun2000 for working on that!

I think you also need to check for `extract_ref`, `dyn_extract_ref`, and `concat_ref` (not sure if there are more). But then you also need to check what the alias is used for and only skip the aliases that are only written to. For example:
```sv
module Test(output logic [31:0] out);
  always_comb begin
    out[30:0] = 'h0;
    out[31] = 'h1;
  end
endmodule
```

`out` will still be added because the assignments are to subelement aliases.

chenbo-again:
maybe a liveness check will help?

chenbo-again:
> I think you also need to check for `extract_ref`, `dyn_extract_ref`, and `concat_ref` (not sure if there are more). But then you also need to check what the alias is used for and only skip the aliases that are only written to. For example:我认为你还

maybe a liveness check can fix that? if there is more Operation in the future, it will help.

chenbo-again:
also this code will wait in1 and in2
```
module Test(output logic [31:0] out, input logic [31:0] in1, input logic [31:0] in2);
  always_comb begin
    out = in1;
    out = in2;
  end
endmodule
```
do you think we should refine this kind of problem in llhd?

maerhart:
@chenbo-again I think your example is expected to have `in1` and `in2` in `llhd.wait` (but not `out`) since both of them are read.

We could factor out this observed values interference into a separate pass that runs before MooreToCore if we want. And then assume that there is already an explicit `wait_event` in always_comb when running MooreToCore. We can't really do this at the LLHD level because the only difference between always_comb, always @(*), always_ff etc. is the position and operands of the wait op.

hailongSun2000:
> also this code will wait in1 and in2
> 
> ```
> module Test(output logic [31:0] out, input logic [31:0] in1, input logic [31:0] in2);
>   always_comb begin
>     out = in1;
>     out = in2;
>   end
> endmodule
> ```
> 
> do you think we should refine this kind of problem in llhd?

Maybe you can check the SystemVerilog IEEE Std 1800-2017 9.4.2.2 § Implicit event_expression list.

PR #8257, opened 2025-02-20
[circt-synth] Load dialects that firtool emits

uenoku:
This PR makes circt-synth load dialects that firtool could emit.


fabianschuiki (approved):
LGTM!

uenoku:
> Consider small test covering the unregistered dialects option?

Added, thank you for the review!

PR #8261, opened 2025-02-20
[LLHD][Sig2Reg] Support integer signal aliasing

maerhart:
Only handles integer signal slices for now. I think it makes sense to expand on that in future PRs (array, struct aliases, drives with enables, etc.)


fabianschuiki (approved):
This is really cool! Love the detailed interval analysis. LGTM 👍

PR #8262, opened 2025-02-21
Implemented AIG node balancing algorithm

Max-astro:
This implementation is based on a tree-balancing algorithm described in this paper: "https://ieeexplore.ieee.org/abstract/document/6105357". 

Here is an easy-to-understand example (Which has been added as a test case):
image

The reference algorithm has two stages, "Covering" and "Tree-balancing," divided into `maximum-and-cover` and `aig-balance-variadic` passes. 

A typical usage is: `circt-synth %s --maximum-and-cover --aig-balance-variadic --cse`. The `maximum-and-cover` pass should be always applied before the `aig-balance-variadic` pass.

### DIFF
--- a/include/circt/Analysis/OpDepthAnalysis.h
+++ b/include/circt/Analysis/OpDepthAnalysis.h
@@ -1,0 +1,68 @@
+//===- OpDepthAnalysis.h - operation depth analyses -----------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This header file defines AIG operation depth analysis.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_ANALYSIS_OPDEPTH_ANALYSIS_H
+#define CIRCT_ANALYSIS_OPDEPTH_ANALYSIS_H
+
+#include "circt/Support/LLVM.h"
+#include "mlir/IR/Operation.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SetVector.h"
+
+#include "circt/Dialect/AIG/AIGOps.h"
+#include "circt/Dialect/HW/HWOps.h"
+
+namespace mlir {
+class AnalysisManager;
+} // namespace mlir
+namespace circt {
+namespace aig {
+namespace analysis {
+
+class OpDepthAnalysis {
+public:
+  OpDepthAnalysis(hw::HWModuleOp moduleOp, mlir::AnalysisManager &am);
+
+  /// Get the depth of operations of a specific name
+  size_t getOpDepth(AndInverterOp op) const {
+    assert(opDepths.count(op));
+    return opDepths.at(op);
+  }
+
+  bool isOnCriticalPath(AndInverterOp op) const {
+    return criticalPath.count(op);
+  }
+
+  const DenseMap<AndInverterOp, size_t> &getOpDepthMap() const {
+    return opDepths;
+  }
+
+  size_t updateLevel(AndInverterOp op, bool isRoot = false);
+  void updateAllLevel();
+
+  SmallVector<AndInverterOp> getPOs();
+
+private:
+  void setCriticalPath(AndInverterOp op);
+
+private:
+  DenseMap<AndInverterOp, size_t> opDepths;
+  SetVector<AndInverterOp> criticalPath;
+  size_t currDepth = 0;
+  hw::HWModuleOp module;
+};
+
+} // namespace analysis
+} // namespace aig
+} // namespace circt
+
+#endif // CIRCT_ANALYSIS_OPDEPTH_ANALYSIS_H
--- a/include/circt/Dialect/AIG/AIGOps.td
+++ b/include/circt/Dialect/AIG/AIGOps.td
@@ -22,7 +22,7 @@ include "mlir/Interfaces/SideEffectInterfaces.td"
 class AIGOp<string mnemonic, list<Trait> traits = []> :
   Op<AIG_Dialect, mnemonic, traits>;
 
-def AndInverterOp : AIGOp<"and_inv", [SameOperandsAndResultType, Pure]> {
+def AndInverterOp : AIGOp<"and_inv", [SameOperandsAndResultType, Pure, Commutative]> {
   let summary = "AIG dialect AND operation";
   let description = [{
     The `aig.and_inv` operation represents an And-Inverter in the AIG dialect.
--- a/lib/Dialect/AIG/Transforms/BalanceVariadic.cpp
+++ b/lib/Dialect/AIG/Transforms/BalanceVariadic.cpp
@@ -1,0 +1,166 @@
+//===- BalanceVariadic.cpp - Lowering Variadic to Binary Ops ------*- C++
+//-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This pass lowers variadic AndInverter operations to balanced binary
+// AndInverter operations.
+//
+//===----------------------------------------------------------------------===//
+#include "llvm/ADT/PriorityQueue.h"
+
+#include "circt/Analysis/OpDepthAnalysis.h"
+#include "circt/Dialect/AIG/AIGOps.h"
+#include "circt/Dialect/AIG/AIGPasses.h"
+#include "circt/Dialect/HW/HWOps.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#include "mlir/IR/Iterators.h"
+#include "mlir/Transforms/WalkPatternRewriteDriver.h"
+
+#define DEBUG_TYPE "aig-balance-variadic"
+
+namespace circt {
+namespace aig {
+#define GEN_PASS_DEF_BALANCEVARIADIC
+#include "circt/Dialect/AIG/AIGPasses.h.inc"
+} // namespace aig
+} // namespace circt
+
+using namespace circt;
+using namespace aig;
+
+namespace {
+/// For wrapping Value and complement information into one object
+struct Signal {
+  Value value;
+  bool complement;
+
+  Signal() = default;
+  Signal(Value v, bool complement) : value(v), complement(complement) {}
+
+  bool isComplement() const { return complement; }
+  Value getValue() const { return value; }
+};
+
+struct BalanceVariadicDriver {
+  BalanceVariadicDriver(mlir::IRRewriter &rewriter,
+                        aig::analysis::OpDepthAnalysis *opDepthAnalysis)
+      : rewriter(rewriter), opDepthAnalysis(opDepthAnalysis) {}
+
+  struct PairSorter {
+    bool operator()(const std::pair<size_t, Signal> &lhs,
+                    const std::pair<size_t, Signal> &rhs) const {
+      return lhs.first > rhs.first;
+    }
+  };
+
+  using NodeLevelHeap =
+      llvm::PriorityQueue<std::pair<size_t, Signal>,
+                          std::vector<std::pair<size_t, Signal>>, PairSorter>;
+
+  void balanceVariadicAndInverterOp(AndInverterOp op) {
+    rewriter.setInsertionPoint(op);
+
+    NodeLevelHeap sortByLevel;
+    for (auto [fanin, inverted] :
+         llvm::zip(op.getOperands(), op.getInverted())) {
+      auto faninOp = fanin.getDefiningOp<AndInverterOp>();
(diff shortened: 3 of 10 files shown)


test/Dialect/AIG/balancing.mlir
```diff
@@ -0,0 +1,34 @@
...
+  %3 = aig.and_inv not %2, %f : i1
+  %4 = aig.and_inv not %c, %3 : i1
+  %5 = aig.and_inv %1, not %4 : i1
+  %6 = aig.and_inv %5, %g : i1
+
+  hw.output %6 : i1
+}
+
+// RUN: circt-opt %s --maximum-and-cover --aig-balance-variadic | FileCheck %s
+// CHECK-LABEL: @Tree2
```
uenoku, line 23: nit: you don't need second run

include/circt/Analysis/OpDepthAnalysis.h
```diff
@@ -0,0 +1,68 @@
...
+namespace aig {
+namespace analysis {
+
+class OpDepthAnalysis {
+public:
+  OpDepthAnalysis(hw::HWModuleOp moduleOp, mlir::AnalysisManager &am);
+
+  /// Get the depth of operations of a specific name
+  size_t getOpDepth(AndInverterOp op) const {
+    assert(opDepths.count(op));
```
uenoku, line 37: nit: `at` already assert this

include/circt/Analysis/OpDepthAnalysis.h
```diff
@@ -0,0 +1,68 @@
+//===- OpDepthAnalysis.h - operation depth analyses -----------------------===//
```
uenoku, line 1: Could you create `include/circt/Dialects/AIG/Analysis` and move the file?

include/circt/Dialect/AIG/AIGOps.td
```diff
@@ -22,7 +22,7 @@ include "mlir/Interfaces/SideEffectInterfaces.td"
 class AIGOp<string mnemonic, list<Trait> traits = []> :
   Op<AIG_Dialect, mnemonic, traits>;
 
-def AndInverterOp : AIGOp<"and_inv", [SameOperandsAndResultType, Pure]> {
+def AndInverterOp : AIGOp<"and_inv", [SameOperandsAndResultType, Pure, Commutative]> {
```
uenoku, line 25: Good point!

test/Dialect/AIG/balancing.mlir
```diff
@@ -0,0 +1,34 @@
+// RUN: circt-opt %s --maximum-and-cover --aig-balance-variadic | FileCheck %s
```
uenoku, line 1: Thank you for adding test! Having a test that runs both `maximum-and-cover` and `aig-balance-variadic` is great. Can we also add a test for each of them as well?

test/Dialect/AIG/balancing.mlir
```diff
@@ -0,0 +1,34 @@
...
+  // CHECK-NEXT: %0 = aig.and_inv %d, %c : i1
+  // CHECK-NEXT: %1 = aig.and_inv %b, %a : i1
+  // CHECK-NEXT: %2 = aig.and_inv %0, %1 : i1
+
+  %1 = aig.and_inv %a, %b : i1
+  %2 = aig.and_inv %c, %1 : i1
+  %3 = aig.and_inv %d, %2 : i1
+
+  hw.output %3 : i1
+}
```
uenoku, line 34: nit: please insert a new line

uenoku (commented):
Looks really great! Thank you for working on this!! It's really nice to separate the balancing into two separate passes. What do you think replacing the existing LowerVariadic pass with your BlanceVariadic pass? I think it makes sense to me always run if it's always beneficial

Max-astro:
I appreciate your feedback and suggestions. Thank you for taking the time to share your review!

> What do you think replacing the existing LowerVariadic pass with your BlanceVariadic pass? I think it makes sense to me always run if it's always beneficial

The BlanceVariadic pass needs to calculate each node's depth during lowering, it might have runtime issues when the input design has very deep combinational logic. In those kinds of cases, using the existing LowerVariadic will be a good solution to get a usable result in less time. So, I think we should keep the LowerVariadic pass in case anyone has this kind of need.

lib/Dialect/AIG/Analysis/OpDepthAnalysis.cpp
```diff
@@ -0,0 +1,84 @@
...
+        it != opDepths.end() && it->second == currDepth) {
+      setCriticalPath(op);
+      break;
+    }
+  }
+}
+
+size_t OpDepthAnalysis::updateLevel(AndInverterOp op, bool isRoot) {
+  if (auto it = opDepths.find(op); !isRoot && it != opDepths.end()) {
+    return it->second;
```
uenoku, line 59:
```suggestion
  if(!isRoot) {
    if (auto it = opDepths.find(op); it != opDepths.end())
      return it->second;
  }
```

lib/Dialect/AIG/Analysis/OpDepthAnalysis.cpp
```diff
@@ -0,0 +1,84 @@
...
+void OpDepthAnalysis::setCriticalPath(AndInverterOp op) {
+  size_t clevel = opDepths[op];
+  for (auto fanin : op.getOperands()) {
+    auto faninOp = fanin.getDefiningOp<AndInverterOp>();
+    if (faninOp && opDepths[faninOp] + 1 == clevel) {
+      setCriticalPath(faninOp);
+      // break; // TODO: Should break when there are multiple critical paths?
+    }
+  }
+}
```
uenoku, line 84: What is expected behavior of `setCriticalPath`?. As far as I see, it's does not modify anything. I believe you wanted to insert to `criticalPath`?

lib/Dialect/AIG/Transforms/MaximumAndCover.cpp
```diff
@@ -0,0 +1,92 @@
...
+    auto hasMultiFanout = [](AndInverterOp op) {
+      int fanouts = 0;
+      for (auto *user : op->getUsers()) {
+        (void)user;
+        fanouts++;
+        if (fanouts > 1) {
+          return true;
+        }
+      }
+      return false;
```
uenoku, line 46: nit:
```suggestion
     return !(op->hasOneUse() || op->use_empty());
```

lib/Dialect/AIG/Transforms/MaximumAndCover.cpp
```diff
@@ -0,0 +1,92 @@
+//===- Balancing.cpp -AIG Binary Ops depth balancing ------*- C++ -*-===//
```
uenoku, line 1: nit: 
```suggestion
//===----------------------------------------------------------------------===//
```

FYI https://github.com/llvm/llvm-project/pull/118553

test/Dialect/AIG/maximum-and-cover.mlir
```diff
@@ -0,0 +1,17 @@
+// RUN: circt-opt %s --maximum-and-cover | FileCheck %s
+// CHECK-LABEL: @Tree1
+hw.module @Tree1(in %a: i1, in %b: i1, in %c: i1, in %d: i1, in %e: i1, in %f: i1, in %g: i1, out o1: i1) {
+ // CHECK-NEXT:  %0 = aig.and_inv %d, not %e : i1
```
uenoku, line 4: nit: please use string substitution ( https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-string-substitution-blocks)
```suggestion
// CHECK-NEXT: %[[AND_INV0:.+]] = aig.and_inv %d, not %e : i1
// CHECK-NEXT: %[[AND_INV1:.+]] = aig.and_inv not %c, not %[[AND_INV0]], %f : i1
// CHECK-NEXT: %[[AND_INV2:.+]] = aig.and_inv %a, not %b, not %[[AND_INV1]], %g : i1
// CHECK-NEXT:  hw.output %[[AND_INV2]] : i1


```

lib/Dialect/AIG/Transforms/BalanceVariadic.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+      size_t level = std::max(llv, rlv) + 1;
+      sortByLevel.push({level, Signal(balanced, false)});
+    }
+
+    switch (sortByLevel.size()) {
+    case 0:
+      break;
+    case 1: {
+      auto signal = sortByLevel.top().second;
```
uenoku, line 112: I think it's necessary to check if the signal is negated.

lib/Dialect/AIG/Transforms/BalanceVariadic.cpp
```diff
@@ -0,0 +1,181 @@
...
+    for (auto [fanin, inverted] :
+         llvm::zip(op.getOperands(), op.getInverted())) {
+      auto faninOp = fanin.getDefiningOp<AndInverterOp>();
+      size_t level = faninOp ? opDepthAnalysis->updateLevel(faninOp, true) : 0;
+      sortByLevel.push({level, Signal(fanin, inverted)});
+    }
+
+    // extract the top two elements with minimum level
+    // and replace them with a new AndInverterOp
+    while (sortByLevel.size() > 2) {
```
uenoku, line 94: nit: Can you change the condition to `sortByLevel.size() > 1` and remove the default case in the following switch statement? I think it's a bit redundant.

include/circt/Dialect/AIG/Analysis/OpDepthAnalysis.h
```diff
@@ -0,0 +1,60 @@
...
+  bool isOnCriticalPath(AndInverterOp op) const {
+    return criticalPath.count(op);
+  }
+
+  const DenseMap<AndInverterOp, size_t> &getOpDepthMap() const {
+    return opDepths;
+  }
+
+  size_t updateLevel(AndInverterOp op, bool isRoot = false);
+  void updateAllLevel();
```
uenoku, line 42: Could you delete API If there is no user at the moment?

lib/Dialect/AIG/Analysis/OpDepthAnalysis.cpp
```diff
@@ -0,0 +1,84 @@
...
+  // updateAllLevel();
+}
+
+SmallVector<AndInverterOp> OpDepthAnalysis::getPOs() {
+  SmallVector<AndInverterOp> po;
+  for (auto op : module.getOps<AndInverterOp>()) {
+    bool isPO = true;
+    for (auto *user : op->getUsers()) {
+      if (isa<AndInverterOp>(user)) {
+        isPO = false;
```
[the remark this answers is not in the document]
Max-astro: You're right; I didn't consider this situation before.
Will change the for loop code like this:
``` cpp
    bool isPO = true;
    for (auto *user : op->getUsers()) {
      if (isa<hw::OutputOp>(user)) {
        isPO = true;
        break;
      }
      if (isa<AndInverterOp>(user)) {
        isPO = false;
      }
    }
    if (isPO)
      po.push_back(op);
```
Is this a bug-free solution?

lib/Dialect/AIG/Analysis/OpDepthAnalysis.cpp
Max-astro: This API has not been used yet and has been removed. removed.

lib/Dialect/AIG/Transforms/BalanceVariadic.cpp
Max-astro: Thanks for pointing it out. I now add an extra inverter node if the signal is negated. Is it better to push the inverter to its output if the uses are also AIG nodes? (By modifying all uses' inverted array)

uenoku (commented):
Looks great, let me test this on a larger design to see if it works!

lib/Dialect/AIG/Transforms/BalanceVariadic.cpp
uenoku: > Is it better to push the inverter to its output if the uses are also AIG nodes?

I think it would be nice optimization but I think we don't need for now.

uenoku:
I'm terribly sorry for updating so so late. I will rebase and merge MaxCover pass in this PR separately. I cherry-picked changes/test cases related to balance pass in https://github.com/llvm/circt/pull/9086 to merge existing LowerVariadic pass to basically do the balancing.

Max-astro:
> I'm terribly sorry for updating so so late. I will rebase and merge MaxCover pass in this PR separately. I cherry-picked changes/test cases related to balance pass in #9086 to merge existing LowerVariadic pass to basically do the balancing.

That's fine, and I'm glad to see my implementation can be integrated into the Circt project. Will see what I can do next.
And BTW, should I close this PR?

uenoku:
> Tat's fine, and I'm glad to see my implementation can be integrated into the Circt project. 

I really appreciate your help and sorry again for my delay. I opened https://github.com/llvm/circt/pull/9090 to check CI passes. I can just merge #9090 or I can force-push the commit (under your name) to this PR and merge from this PR if you prefer that.

Max-astro:
> > Tat's fine, and I'm glad to see my implementation can be integrated into the Circt project.
> 
> I really appreciate your help and sorry again for my delay. I opened #9090 to check CI passes. I can just merge #9090 or I can force-push the commit to this PR and merge from this PR if you prefer that.

Merging the new PR seems more straightforward. That's totally OK with me, and I'll close this PR later.

uenoku:
Thanks, merged by https://github.com/llvm/circt/commit/ed501a8f9d95a33a72b81e881a1e0100933c0c28.

PR #8264, opened 2025-02-21
Bump LLVM to 6bfedfa0ba31a8ac8fd7fcfd2d33afaa5eabe0e5.

mikeurbach:
This is mostly a straightforward bump.

One change we had to incorporate is MLIR is generally moving towards auto-generated pass constructor functions. For more info, see: https://github.com/llvm/llvm-project/commit/d25beca.

This shows up for us as a few pass constructors we use from upstream getting renamed. We just need to apply the same renames as upstream: https://github.com/llvm/llvm-project/commit/c1a2292.


seldridge (approved):
LGTM

PR #8265, opened 2025-02-21
[FlattenMemRef] Flatten MemRef ReshapeOp when its source memory reference is flattened in other passes

jiahanxie353:
This patch flattens `memref.reshape` when its source memory reference is flattened in other passes

### DIFF
--- a/lib/Transforms/FlattenMemRefs.cpp
+++ b/lib/Transforms/FlattenMemRefs.cpp
@@ -238,6 +238,25 @@ struct GetGlobalOpConversion : public OpConversionPattern<memref::GetGlobalOp> {
   }
 };
 
+struct ReshapeOpConversion : public OpConversionPattern<memref::ReshapeOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(memref::ReshapeOp op, OpAdaptor adaptor,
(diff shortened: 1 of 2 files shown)


lib/Transforms/FlattenMemRefs.cpp
```diff
@@ -238,6 +238,25 @@ struct GetGlobalOpConversion : public OpConversionPattern<memref::GetGlobalOp> {
...
+    Value flattenedSource = rewriter.getRemappedValue(op.getSource());
+    if (!flattenedSource)
+      return failure();
+
+    auto flattenedSrcType = cast<MemRefType>(flattenedSource.getType());
+    if (isUniDimensional(flattenedSrcType) ||
+        !flattenedSrcType.hasStaticShape())
+      rewriter.replaceOp(op, flattenedSource);
+
+    return success();
```
cgyurgyik, line 256: Shouldn't this only return success if the rewrite is successful, i.e.,
```suggestion
    if (isUniDimensional(flattenedSrcType) ||
        !flattenedSrcType.hasStaticShape()) {
      rewriter.replaceOp(op, flattenedSource);
      return success();
    }

    return failure();
```

jiahanxie353: yeah you are totally right

cgyurgyik (approved):
LGTM

PR #8270, opened 2025-02-24
[ImportVerilog] add unpacked array concatenation

chenbo-again:
This PR is aimed to provide unpacked array concatenation.
```
// unpacked array concatenation
string s[] = { "hello", "sad", "world" };
// string concatenation
string = { "hello", "sad", "world" };
```
these two operation do not act the same way, and the result of concat is depends on the Lhs.


chenbo-again:
This PR import UnpackedArrayConcatOp operation, This Operation is very similar to array_create operation, but can use unpacked array as operands.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -993,6 +1003,7 @@ module Expressions;
     // CHECK: moore.extract_ref %vec_1 from 11 : <l32> -> <l3>
     vec_1[15-:3] = y;
 
+    vec_s = { "hello", "sad", "world" };
```
hailongSun2000, line 1006: If there isn't `stringConcat`, you have to remove it. Otherwise, you should add `// CHECK: `s to test it.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -439,6 +439,12 @@ def AnyUnionType : MooreType<
   "packed or unpacked union type",
   "moore::UnpackedType">;
 
+def AnyIndexedUnpackedArrayType : MooreType<
+  Or<[UnpackedArrayType.predicate, OpenUnpackedArrayType.predicate, QueueType.predicate]>, 
+  "unpacked array type except associative array type",
+  "moore::UnpackedType">;
```
hailongSun2000, line 445: I noticed that `UnpackedArrayType`, `OpenUnpackedArrayType`, and `QueueType` are all `AnyIndexedUnpackedArrayType`, but you just tested the `OpenUnpackedArrayType`, the others you also have to test.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1176,8 +1186,43 @@ struct LvalueExprVisitor {
...
+      }
+      return builder.create<moore::UnpackedArrayConcatOp>(loc, requiredType,
+                                                          operands);
+    }
+
+    default:
+      return convertRvalueExpression(expr, requiredType);
+    }
+  }
+
```
hailongSun2000, line 1225: You should move this code snippet to a proper place, like `Value visit(const slang::ast::ConcatenationExpression &expr)`.

hailongSun2000: By the way, maybe you can check the following case(it's mentioned at SV IEEE Std 10.10.3):
```
typedef int T_QI[$];
T_QI jagged_array[$] = '{ {1}, T_QI'{2,3,4}, {5,6} };
```

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -439,6 +439,12 @@ def AnyUnionType : MooreType<
...
+
+
```
fabianschuiki, line 447:
```suggestion
```

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: I agree with @terapines-osc-1: you should be able to move this into the `visit` function for `ConcatenationExpression`. Slang has already performed the type-checking, so you should be able to just check the `expr.type` to see if is an unpacked array concatenation, and then create the new unpacked array concat op.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -707,6 +707,30 @@ module Expressions;
...
+  // CHECK: %queue_s = moore.variable [[TMP3]] : <queue<string, 0>>
+  string queue_s [$] = {vec_s, "hello"};
+  // CHECK: [[TMP0:%.+]] = moore.read %queue_s : <queue<string, 0>>
+  // CHECK: [[TMP1:%.+]] = moore.string_constant "hello" : i40
+  // CHECK: [[TMP2:%.+]] = moore.conversion [[TMP1]] : !moore.i40 -> !moore.string
+  // CHECK: [[TMP3:%.+]] = moore.string_constant "new" : i24
+  // CHECK: [[TMP4:%.+]] = moore.conversion [[TMP3]] : !moore.i24 -> !moore.string
+  // CHECK: [[TMP5:%.+]] = moore.unpacked_array_concat [[TMP0]], [[TMP2]], [[TMP4]] : (!moore.queue<string, 0>, !moore.string, !moore.string) -> uarray<3 x string>
+  // CHECK: %unpack_s = moore.variable [[TMP5]] : <uarray<3 x string>>
+  string unpack_s [3] = {queue_s, "hello", "new"};
```
fabianschuiki, line 733: Very very cool 😎

PR #8271, opened 2025-02-25
[ImportVerilog] Add option to lower always @* as always_comb

fabianschuiki:
Add the `--always-at-star-as-comb` option to circt-verilog and add a corresponding option to ImportVerilog. When this option is set, detect `always @*` in the Verilog AST and lower it as a `always_comb`. This is a common pattern in synthesizers and simulators, since the traiditonal `always @*` in Verilog does not accurately describe the behaviour of combinational logic.

Enable this option by default, which seems to be a common behaviour among existing tools.

### DIFF
--- a/include/circt/Conversion/ImportVerilog.h
+++ b/include/circt/Conversion/ImportVerilog.h
@@ -49,6 +49,9 @@ struct ImportVerilogOptions {
   /// Generate debug information in the form of debug dialect ops in the IR.
   bool debugInfo = false;
 
+  /// Interpret `always @(*)` as `always_comb`.
+  bool lowerAlwaysAtStarAsComb = true;
+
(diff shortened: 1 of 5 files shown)


maerhart (approved):
Does this also apply if there is a begin...end block and the first statment is an implicit event?  Should it?

Otherwise LGTM!

fabianschuiki:
I think we could detect that as well (`always begin @(*) ... end`), but it's a pretty uncommon thing as far as I know. You'd have to write a strange `always begin @(*) begin ... end end` whenever you'd want multiple statements in that body, since the `@(*)` infers accesses performed within its own sub-statement. I'd suggest we narrowly detect `always @(*) ...` initially, and add more patterns as we encounter them.

PR #8277, opened 2025-02-26
[MemoryBanking] Memory banking supports multiple banking factors/dimensions config

jiahanxie353:
This patch supports passing multiple `banking.factors/dimensions` to the command line options as well as attaching as attributes.

### DIFF
--- a/include/circt/Transforms/Passes.td
+++ b/include/circt/Transforms/Passes.td
@@ -132,7 +132,7 @@ def MemoryBanking : Pass<"memory-banking", "::mlir::func::FuncOp"> {
            "The elements specified in banking factors should be greater than 1;"
            "The elements specified in banking factors will be paired with the ones specified in banking dimensions."
            "In principle, the number of elements in banking factors should be equal to banking dimensions',"
-           "with the only exception that there is one banking factor with zero banking dimension.">,
+           "with a single exception case: there is one banking factor with zero banking dimensions.">,
     ListOption<"bankingDimensionsList", "dimensions", "unsigned",
            "The dimensions along which to bank the memory. If unspecified and"
            "there is only one factor, the innermost dimension with size > 1 is used.">
--- a/include/circt/Transforms/Passes.td
+++ b/include/circt/Transforms/Passes.td
@@ -127,10 +127,11 @@ def MemoryBanking : Pass<"memory-banking", "::mlir::func::FuncOp"> {
   let summary = "Partition the memories used in affine parallel loops into banks";
   let constructor = "circt::createMemoryBankingPass()";
   let options = [
-    Option<"bankingFactor", "factor", "unsigned", /*default=*/"1",
+    ListOption<"bankingFactorsList", "factors", "unsigned",
            "Use this banking factor for all memories being partitioned">,
-    Option<"bankingDimension", "dimension", "int", /*default=*/"-1",
-           "The dimension along which to bank the memory. If -1, the innermost dimension with size > 1 is used.">
+    ListOption<"bankingDimensionsList", "dimensions", "unsigned",
+           "The dimension along which to bank the memory. If unspecified and"
+           "there is only one factor, the innermost dimension with size > 1 is used.">
   ];
   let dependentDialects = ["mlir::memref::MemRefDialect, mlir::scf::SCFDialect, mlir::affine::AffineDialect"];
 }
--- a/include/circt/Transforms/Passes.td
+++ b/include/circt/Transforms/Passes.td
@@ -128,9 +128,13 @@ def MemoryBanking : Pass<"memory-banking", "::mlir::func::FuncOp"> {
   let constructor = "circt::createMemoryBankingPass()";
   let options = [
     ListOption<"bankingFactorsList", "factors", "unsigned",
-           "Use this banking factor for all memories being partitioned">,
+           "Use banking factors to partition all memories that don't have banking attributes."
+           "The elements specified in banking factors should be greater than 1;"
+           "The elements specified in banking factors will be paired with the ones specified in banking dimensions."
+           "In principle, the number of elements in banking factors should be equal to banking dimensions',"
+           "with the only exception that there is one banking factor with zero banking dimension.">,
     ListOption<"bankingDimensionsList", "dimensions", "unsigned",
-           "The dimension along which to bank the memory. If unspecified and"
+           "The dimensions along which to bank the memory. If unspecified and"
            "there is only one factor, the innermost dimension with size > 1 is used.">
   ];
   let dependentDialects = ["mlir::memref::MemRefDialect, mlir::scf::SCFDialect, mlir::affine::AffineDialect"];
--- a/lib/Transforms/MemoryBanking.cpp
+++ b/lib/Transforms/MemoryBanking.cpp
@@ -25,7 +25,6 @@
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/FormatVariadic.h"
-#include "llvm/Support/raw_ostream.h"
 #include <numeric>
 
 namespace circt {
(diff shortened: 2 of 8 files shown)


lib/Transforms/MemoryBanking.cpp
```diff
@@ -695,55 +765,168 @@ LogicalResult cleanUpOldMemRefs(DenseSet<Value> &oldMemRefVals,
...
+    });
+
+    if (failed(applyMemoryBanking(getOperation(), &getContext()))) {
+      signalPassFailure();
+    } else if (noBanksCreated)
+      // We stop the pass when IR stops changing:
+      // 1. there is no more banks are created;
+      // 2. `applyMemoryBanking` has reached its fixed point, which means every
+      // memory read/write operation has been rewritten to be using the newly
+      // created banks, and that the old memory references are erased.
```
cgyurgyik, line 897: Nit: this seems a bit redundant with (1), I'm not totally in love with my change but it'd be nice to avoid saying the same thing twice.

```suggestion
      // We run this pass until convergence, i.e., `applyMemoryBanking` has reached its fixed point, which means every
      // memory read/write operation has been rewritten to be using the newly
      // created banks, and that the old memory references are erased.
```

lib/Transforms/MemoryBanking.cpp
```diff
@@ -324,43 +320,118 @@ void updateFuncOpArgAttrs(func::FuncOp funcOp, unsigned argIndex,
...
+           "BankingConfig dimensions ArrayAttr should not be empty");
+    return arrayAttr.size() > 1
+               ? ArrayAttr::get(context, arrayAttr.getValue().take_back(
+                                             arrayAttr.size() - 1))
+               : nullptr;
+  }
+  auto intAttr = dyn_cast<IntegerAttr>(bankingConfigAttrs.dimensions);
+  assert(intAttr && "BankingConfig dimension must be an integer");
+  return nullptr;
+}
```
cgyurgyik, line 390: Nit this looks almost exactly the same; can we have a helper function here to handle this? Same with `getCurrBankingInfo`.

lib/Transforms/MemoryBanking.cpp
```diff
@@ -426,45 +494,48 @@ struct BankAffineLoadPattern
...
     int64_t memrefRank = originalMemRefType.getRank();
-    ArrayRef<int64_t> shape = originalMemRefType.getShape();
 
-    auto bankingDimension =
-        getSpecifiedOrDefaultBankingDim(bankingDimensionOpt, memrefRank, shape);
+    BankingConfigAttributes currBankingConfig =
+        getMemRefBankingConfig(originalMem);
+    if (!currBankingConfig.factors) {
+      assert(!currBankingConfig.dimensions &&
+             "A banking dimension must be paired with a factor");
```
cgyurgyik, line 502: This was dicey to me. Why aren't these passed as pairs already? Is it really easier to have these as two lists for the compiler / frontend user?

test/Transforms/memory_banking_multi_config.mlir
```diff
@@ -0,0 +1,168 @@
+// RUN: circt-opt %s -memory-banking="factors=2 dimensions=1" --canonicalize --split-input-file | FileCheck %s
```
cgyurgyik, line 1: What does `factors=2 dimensions=1` here really mean? Is there documentation for these attributes?

lib/Transforms/MemoryBanking.cpp
jiahanxie353: sure!

lib/Transforms/MemoryBanking.cpp
jiahanxie353: you are right, I'll try to refactor

test/Transforms/memory_banking_multi_config.mlir
jiahanxie353: `factors=2 dimensions=1` means we want to bank the memory references by a factor of 2 at dimension 1, right? Or am I misunderstanding the point you are confused about?

lib/Transforms/MemoryBanking.cpp
jiahanxie353: sometimes users might only pass in `factors` without `dimensions`;
and it also adheres to the existing HLS tool by Vitis: https://docs.amd.com/r/en-US/ug1399-vitis-hls/Array-Partitioning

cgyurgyik: Ok so to check my understanding, `factors=4` would apply it to every dimension? Regardless it'd be nice if this was documented

test/Transforms/memory_banking_multi_config.mlir
cgyurgyik: Ah, so this applies to every memory that does not have the factors already included.

Can a user also pass multiple factors/dimensions, e.g., `-memory-banking="factors=[2,4] dimensions=[0,1]"

jiahanxie353: > so this applies to every memory that does not have the factors already included.

yes!

> Can a user also pass multiple factors/dimensions, e.g., `-memory-banking="factors=[2,4] dimensions=[0,1]"

yep! and this is what this PR is meant for

lib/Transforms/MemoryBanking.cpp
jiahanxie353: Just refactored the logic and ready for another round of review!

include/circt/Transforms/Passes.td
```diff
@@ -127,10 +127,11 @@ def MemoryBanking : Pass<"memory-banking", "::mlir::func::FuncOp"> {
   let summary = "Partition the memories used in affine parallel loops into banks";
   let constructor = "circt::createMemoryBankingPass()";
   let options = [
-    Option<"bankingFactor", "factor", "unsigned", /*default=*/"1",
+    ListOption<"bankingFactorsList", "factors", "unsigned",
            "Use this banking factor for all memories being partitioned">,
```
cgyurgyik, line 131: This comment needs to be updated, what if there are multiple factors?

lib/Transforms/MemoryBanking.cpp
```diff
@@ -324,43 +320,111 @@ void updateFuncOpArgAttrs(func::FuncOp funcOp, unsigned argIndex,
...
+    assert(!arrayAttr.empty() &&
+           "BankingConfig dimensions ArrayAttr should not be empty");
+    return arrayAttr.size() > 1
+               ? ArrayAttr::get(context, arrayAttr.getValue().take_back(
+                                             arrayAttr.size() - 1))
+               : nullptr;
+  }
+  auto intAttr = dyn_cast<IntegerAttr>(bankingConfigAttrs.dimensions);
+  assert(intAttr && "BankingConfig dimension must be an integer");
+  return nullptr;
```
cgyurgyik, line 384: Can we refactor this too? It looks identical to me.

lib/Transforms/MemoryBanking.cpp
```diff
@@ -695,55 +757,208 @@ LogicalResult cleanUpOldMemRefs(DenseSet<Value> &oldMemRefVals,
...
+  // its fixed point, which means every memory read/write operation has been
+  // rewritten to be using the newly created banks, and that the old memory
+  // references are erased.
+  bool banksCreated = false;
+  do {
+    memoryToBanks.clear();
+    oldMemRefVals.clear();
+    opsToErase.clear();
+
+    banksCreated = false;
```
cgyurgyik, line 909: Can't we delete this line?

cgyurgyik (commented):
A few smaller comments, thanks! Looks great.

lib/Transforms/MemoryBanking.cpp
jiahanxie353: we should not right?
say a memory reference needs to be banked in multiple dimensions.
In the first pass to bank the 0'th dimension, `banksCreated` will be set to `true` after `createBanks`. Afterwards, `banksCreated` will remain `true`, so there's no way we can break out of the loop.

lib/Transforms/MemoryBanking.cpp
jiahanxie353: sure will do!

lib/Transforms/MemoryBanking.cpp
cgyurgyik: Ah yes, you're definitely right.

jiahanxie353:
> A few smaller comments, thanks! Looks great.

Done! Thank you for your time and review!

include/circt/Transforms/Passes.td
```diff
@@ -127,10 +127,15 @@ def MemoryBanking : Pass<"memory-banking", "::mlir::func::FuncOp"> {
...
-    Option<"bankingFactor", "factor", "unsigned", /*default=*/"1",
-           "Use this banking factor for all memories being partitioned">,
-    Option<"bankingDimension", "dimension", "int", /*default=*/"-1",
-           "The dimension along which to bank the memory. If -1, the innermost dimension with size > 1 is used.">
+    ListOption<"bankingFactorsList", "factors", "unsigned",
+           "Use banking factors to partition all memories that don't have banking attributes."
+           "The elements specified in banking factors should be greater than 1;"
+           "The elements specified in banking factors will be paired with the ones specified in banking dimensions."
+           "In principle, the number of elements in banking factors should be equal to banking dimensions',"
+           "with the only exception that there is one banking factor with zero banking dimension.">,
```
cgyurgyik, line 135:
```suggestion
           "with a single exception case: there is one banking factor with zero banking dimensions.">,
```

cgyurgyik (approved):
Nice!

PR #8278, opened 2025-02-26
[LLHD] Support conditional drives in mem2reg

fabianschuiki:
Extend the LLHD Mem2Reg pass to support conditional drives. This is pretty trivial since the pass already tracks drive conditions generated through control flow. All that's needed is to initialize the condition of a drive's reaching definition that gets propagated across the lattice.


maerhart (approved):
LGTM

PR #8279, opened 2025-02-26
[LLHD] Support drives with delta delay in mem2reg

fabianschuiki:
Extend the LLHD Mem2Reg pass to also support non-blocking drives. These have a delay value of `1d` instead of the `1e` of blocking drives. To do so, add an additional bit to the slots associated with reaching definitions, indicating whether the definition for a slot is blocking/immediate or applies only after a delta delay. Blocking drives can then appropriately kill earlier non-blocking drives, and probes can only consider earlier blocking drives when resolving their value.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
+++ b/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
@@ -40,11 +40,28 @@ static bool isEpsilonDelay(Value value) {
   return false;
 }
 
+/// Check whether a value is defined by `llhd.constant_time <0ns, 1d, 0e>`.
+static bool isDeltaDelay(Value value) {
+  if (auto timeOp = value.getDefiningOp<ConstantTimeOp>()) {
+    auto t = timeOp.getValue();
+    return t.getTime() == 0 && t.getDelta() == 1 && t.getEpsilon() == 0;
+  }
+  return false;
+}
+
 /// Check whether an operation is a `llhd.drive` with an epsilon delay. This
 /// corresponds to a blocking assignment in Verilog.
 static bool isBlockingDrive(Operation *op) {
   if (auto driveOp = dyn_cast<DrvOp>(op))
     return isEpsilonDelay(driveOp.getTime());
+  return false;
+}
+
+/// Check whether an operation is a `llhd.drive` with a delta delay. This
+/// corresponds to a non-blocking assignment in Verilog.
+static bool isDeltaDrive(Operation *op) {
+  if (auto driveOp = dyn_cast<DrvOp>(op))
+    return isDeltaDelay(driveOp.getTime());
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


test/Dialect/LLHD/Transforms/mem2reg.mlir
```diff
@@ -493,5 +493,61 @@ hw.module @ConditionalDrives(in %u: i42, in %v: i42, in %q: i1, in %r: i1) {
...
+  ^bb2:
+    // CHECK-NEXT: ^bb2([[ABLK:%.+]]: i42, [[ABLKCOND:%.+]]: i1, [[ADEL:%.+]]: i42, [[ADELCOND:%.+]]: i1):
+    // CHECK-NEXT: [[T:%.+]] = llhd.constant_time <0ns, 0d, 1e>
+    // CHECK-NEXT: llhd.drv %a, [[ABLK]] after [[T]] if [[ABLKCOND]]
+    // CHECK-NEXT: [[T:%.+]] = llhd.constant_time <0ns, 1d, 0e>
+    // CHECK-NEXT: llhd.drv %a, [[ADEL]] after [[T]] if [[ADELCOND]]
+    // CHECK-NEXT: llhd.halt
+    llhd.halt
+  }
+}
```
maerhart, line 550: If there are two delayed drives to the same signal some conflict resolution has to happen AFAIK, but in mem2reg we probably just want to leave it as is and not merge the drives. Can you add a test to check what mem2reg does in such a situation?

maerhart (approved):
I'm not very familiar with the semantics of blocking and delayed drives to the same signal, but otherwise LGTM.

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
```diff
@@ -884,10 +919,13 @@ void Promoter::propagateForward(LatticeNode *node, bool optimisticMerges) {
     return;
   }
 
-  // Drives propagate the driven value as a reaching def.
+  // Drives propagate the driven value as a reaching def. Blocking drives kill
+  // earlier non-blocking drives.
```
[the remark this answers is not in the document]
likeamahoney: "Blocking drives kill earlier non-blocking drives." is wrong statement due to that fact from 10.4.2 SystemVerilog LRM section:

> At the end of the time step means that the nonblocking assignments are the last assignments executed in a time step

Blocking drives values are need only for other blocking assignments in procedural flow and also for rvalues in right-hand sides of nonblocking assignemnt

test/Dialect/LLHD/Transforms/mem2reg.mlir
fabianschuiki: If I remember correctly, this only happens on wires, where you cannot have procedural drives assign to them. If you have two drives to the same variable, the latter one should win and override the earlier one. Let me go check on the details

fabianschuiki: 10.4.2 in IEEE the 1800-2007 standard indicates that driving the same variable from different processes will see one of the assignments win but no conflict resolution occurs:

Screen Shot 2025-02-27 at 08 53 23

In the same section, there's an example showing how delayed drives with a shorter delay override already scheduled delayed drives with a longer delay:

Screen Shot 2025-02-27 at 08 54 41

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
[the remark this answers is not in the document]
fabianschuiki: Let me tweak this comment a bit. Consider the following SV:
```systemverilog
a <= #10ns 42;  // A
a <= 43;        // B
a = 44;         // C
```
Since writes with less delay override/kill already scheduled writes with longer delay, (A) schedules a change to 42 after 10ns, but then (B) schedules a change to 43 after a delta delay / at the end of the time step, which overrides the earlier (A), but then (C) applies an immediate change to 44, which overrides the earlier (B). To preserve this, it's important that blocking drives kill the reaching definitions of non-blocking drives to not schedule a later signal change.

PR #8281, opened 2025-02-27
[circt-verilog-lsp][vscode] Add initial VS Code extension configuration for CIRCT Verilog LSP

uenoku:
This commit adds the basic configuration files needed for a VS Code extension
that provides Verilog language server protocol (LSP) support through CIRCT.

While it may seem unusual to include a VS Code extension in-tree, this follows
the precedent set by the MLIR VS Code extension
(https://github.com/llvm/llvm-project/tree/main/mlir/utils/vscode). Having the
extension in-tree ensures it stays in sync with the server implementation and
simplifies maintenance.

The extension will enable VS Code to communicate with the CIRCT Verilog LSP
server, providing features like syntax highlighting, error checking, and code
navigation for Verilog files (.v and .sv).


utils/vscode/LICENSE
```diff
@@ -0,0 +1,234 @@
+==============================================================================
```
uenoku, line 1: This is not necessary. Clone LICENSE file in the extension repository (circt/vscode-circt).

PR #8282, opened 2025-02-27
[AffineParallelUnroll] Fix empty map lookup; move canonicalization passes before `MemoryConflictResolver`

jiahanxie353:
This patch fixes the empty lookup issue when first looking up `constantWriteOpIndices`.

Also, I moved the canonicalization passes before `MemoryConflictResolver`. It's because the `AffineParallelUnroll` pass introduces a lot of constant values, and running the canonicalization passes in advance can simplify the analysis work in `MemoryConflictResolver`.

And I also added a test case for the changes.


cgyurgyik (approved):

jiahanxie353:
Thank you for the timely review! @cgyurgyik 
Going to merge it as the CI has passed.

PR #8284, opened 2025-03-01
[LLHD] Add pass to hoist probes out of processes

fabianschuiki:
Add the HoistSignals pass which is intended to hoist probes and drives out of processes. Currently only probe hoisting is implemented. Probes that immediately follow resumption points, i.e. the entry block and blocks after `llhd.wait`, can be hoisted out of the process if their result is not live across any wait ops. Almost all probes left after the Mem2Reg pass can be hoisted.

In the future, this pass should also hoist drives out of processes and promote the driven value to process results. This will remove all drives from combinational processes, such that only drives encoding register/latch behaviour remain.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/HoistSignals.cpp
+++ b/lib/Dialect/LLHD/Transforms/HoistSignals.cpp
@@ -1,0 +1,193 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/LLHD/IR/LLHDOps.h"
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h"
+#include "mlir/Analysis/Liveness.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "llhd-hoist-signals"
+
+namespace circt {
+namespace llhd {
+#define GEN_PASS_DEF_HOISTSIGNALSPASS
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h.inc"
+} // namespace llhd
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+using namespace llhd;
+
+//===----------------------------------------------------------------------===//
+// Probe Hoisting
+//===----------------------------------------------------------------------===//
+
+namespace {
+/// The struct performing the hoisting in a single region.
+struct Hoister {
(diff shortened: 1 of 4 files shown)


lib/Dialect/LLHD/Transforms/HoistSignals.cpp
```diff
@@ -0,0 +1,184 @@
...
+      // Move the probe out of the process, trying to reuse any previous probe
+      // that we've already hoisted.
+      auto &hoistedOp = hoistedProbes[probeOp.getSignal()];
+      if (hoistedOp) {
+        LLVM_DEBUG(llvm::dbgs() << "- Replacing " << probeOp << "\n");
+        probeOp.replaceAllUsesWith(hoistedOp.getResult());
+        probeOp.erase();
+      } else {
+        LLVM_DEBUG(llvm::dbgs() << "- Hoisting " << probeOp << "\n");
+        probeOp->moveBefore(region.getParentOp());
```
maerhart, line 152: What if the signal operand of the probe is not available at the destination, e.g., when the signal is defined inside the process?

test/Dialect/LLHD/Transforms/hoist-signals.mlir
```diff
@@ -0,0 +1,77 @@
...
+  ^bb1(%1: i42):
+    // CHECK: ^bb1([[A:%.+]]: i42):
+    llhd.wait ^bb2
+  ^bb2:
+    // CHECK: ^bb2:
+    // CHECK-NEXT: call @use_i42([[A]])
+    func.call @use_i42(%1) : (i42) -> ()
+    llhd.halt
+  }
+}
```
maerhart, line 74: There's no test to check the reuse of probes already hoisted.

maerhart (approved):
I think we need to check for the case where the signal is defined inside the process and add a test for that. Otherwise, looks good! 🎉

lib/Dialect/LLHD/Transforms/HoistSignals.cpp
```diff
@@ -0,0 +1,184 @@
...
+  /// The set of values that are alive across wait ops themselves, or that have
+  /// transitive users that are live across wait ops.
+  DenseSet<Value> liveAcrossWait;
+
+  /// A lookup table of probes we have already hoisted, for deduplication.
+  DenseMap<Value, PrbOp> hoistedProbes;
+};
+} // namespace
+
+LogicalResult Hoister::hoist() {
```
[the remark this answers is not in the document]
fabianschuiki: Good point! Turns out the pass can't really fail 😁

lib/Dialect/LLHD/Transforms/HoistSignals.cpp
fabianschuiki: Excellent point. I've added a check and a test for that.

test/Dialect/LLHD/Transforms/hoist-signals.mlir
fabianschuiki: Good point; added one.

PR #8290, opened 2025-03-03
[SCFToCalyx] Arith FPToSI lowering and its emit

jiahanxie353:
This patch supports lowering and emit of `arith.fptosi`, with the backend support in Calyx: https://github.com/calyxir/calyx/pull/2422.

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -516,7 +516,7 @@ def WireLibOp : UnaryLibraryOp<"wire", "std_", [SameTypeConstraint<"in", "out">]
 def ExtSILibOp  : UnaryLibraryOp<"signext", "std_"> {}
 
 class UnaryFloatingPointLibraryOp<string mnemonic, list<Trait> traits = []> :
-  UnaryLibraryOp<mnemonic, "", !listconcat(traits, [
+  UnaryLibraryOp<mnemonic, /*prefix=*/"", !listconcat(traits, [
     DeclareOpInterfaceMethods<FloatingPointOpInterface>
   ])> {}
 
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -520,14 +520,14 @@ class UnaryFloatingPointLibraryOp<string mnemonic, list<Trait> traits = []> :
     DeclareOpInterfaceMethods<FloatingPointOpInterface>
   ])> {}
 
+// This models the floating point to integer conversion interface in Berkeley HardFloat implementation. It rounds to the nearest (towards zero) integer, which is `round_minMag` in the Berkeley HardFloat library.
 def FpToIntOpIEEE754 : UnaryFloatingPointLibraryOp<"ieee754.fptoint"> {
   let results = (outs I1:$clk, I1:$reset, I1:$go,
     AnySignlessInteger:$in, I1: $signedOut, AnySignlessInteger:$out, I1:$done);
 
   let extraClassDefinition = [{
     SmallVector<StringRef> $cppClass::portNames() {
-      return {clkPort, resetPort, goPort, "in", "signedOut", "out", donePort
-      };
+      return {clkPort, resetPort, goPort, "in", "signedOut", "out", donePort};
     }
 
     SmallVector<Direction> $cppClass::portDirections() {
--- a/include/circt/Dialect/Calyx/CalyxPrimitives.td
+++ b/include/circt/Dialect/Calyx/CalyxPrimitives.td
@@ -488,22 +488,22 @@ def DivUPipeLibOp : ArithBinaryPipeLibraryOp<"divu"> {}
 def RemUPipeLibOp : ArithBinaryPipeLibraryOp<"remu"> {}
 def RemSPipeLibOp : ArithBinaryPipeLibraryOp<"rems"> {}
 
-class UnaryLibraryOp<string mnemonic, list<Trait> traits = []> :
-    CalyxLibraryOp<mnemonic, "std_", !listconcat(traits, [Combinational])> {
+class UnaryLibraryOp<string mnemonic, string prefix, list<Trait> traits = []> :
+    CalyxLibraryOp<mnemonic, prefix, !listconcat(traits, [Combinational])> {
   let results = (outs AnyInteger:$in, AnyInteger:$out);
 }
 
-def PadLibOp : UnaryLibraryOp<"pad"> {
+def PadLibOp : UnaryLibraryOp<"pad", "std_"> {
   let hasVerifier = 1;
 }
 
-def SliceLibOp : UnaryLibraryOp<"slice"> {
+def SliceLibOp : UnaryLibraryOp<"slice", "std_"> {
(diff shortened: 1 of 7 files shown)


include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -488,22 +488,22 @@ def DivUPipeLibOp : ArithBinaryPipeLibraryOp<"divu"> {}
 def RemUPipeLibOp : ArithBinaryPipeLibraryOp<"remu"> {}
 def RemSPipeLibOp : ArithBinaryPipeLibraryOp<"rems"> {}
 
-class UnaryLibraryOp<string mnemonic, list<Trait> traits = []> :
-    CalyxLibraryOp<mnemonic, "std_", !listconcat(traits, [Combinational])> {
+class UnaryLibraryOp<string mnemonic, string prefix, list<Trait> traits = []> :
```
cgyurgyik, line 491: Is there some place that requires this additional change? I'm not seeing it

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -513,4 +513,45 @@ def WireLibOp : UnaryLibraryOp<"wire", [SameTypeConstraint<"in", "out">]> {
...
+  ])> {}
+
+def FpToIntOpIEEE754 : UnaryFloatingPointLibraryOp<"ieee754.fptoint"> {
+  let results = (outs I1:$clk, I1:$reset, I1:$go,
+    AnySignlessInteger:$in, I1: $signedOut, AnySignlessInteger:$out, I1:$done);
+
+  let extraClassDefinition = [{
+    SmallVector<StringRef> $cppClass::portNames() {
+      return {clkPort, resetPort, goPort, "in", "signedOut", "out", donePort
+      };
```
cgyurgyik, line 530:
```suggestion
      return {clkPort, resetPort, goPort, "in", "signedOut", "out", donePort};
```

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -1001,7 +1006,7 @@ void Emitter::emitLibraryPrimTypedByFirstOutputPort(
 void Emitter::emitLibraryFloatingPoint(Operation *op) {
   auto cell = cast<CellInterface>(op);
   // magic number for the index of `left/right` input port
-  size_t inputPortIndex = cell.getInputPorts().size() - 3;
+  size_t inputPortIndex = cell.getInputPorts().size() - 2;
```
cgyurgyik, line 1009: Question: how come this just "works"? This kind of scares me :)

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -1034,10 +1039,17 @@ void Emitter::emitLibraryFloatingPoint(Operation *op) {
   if (auto fpOp = dyn_cast<calyx::FloatingPointOpInterface>(op)) {
     opName = fpOp.getCalyxLibraryName();
   }
+
   indent() << getAttributes(op, /*atFormat=*/true) << cell.instanceName()
            << space() << equals() << space() << opName << LParen() << expWidth
-           << comma() << sigWidth << comma() << bitWidth << RParen()
-           << semicolonEndL();
+           << comma() << sigWidth << comma() << bitWidth;
+  if (opName == "fpToInt") {
```
cgyurgyik, line 1046: Nit: is there something less brittle than the opName, e.g., perhaps a `Type` that we can dyn_cast?

include/circt/Dialect/Calyx/CalyxPrimitives.td
```diff
@@ -513,4 +513,45 @@ def WireLibOp : UnaryLibraryOp<"wire", [SameTypeConstraint<"in", "out">]> {
...
 
-def ExtSILibOp  : UnaryLibraryOp<"signext"> {}
+def ExtSILibOp  : UnaryLibraryOp<"signext", "std_"> {}
+
+class UnaryFloatingPointLibraryOp<string mnemonic, list<Trait> traits = []> :
+  UnaryLibraryOp<mnemonic, "", !listconcat(traits, [
+    DeclareOpInterfaceMethods<FloatingPointOpInterface>
+  ])> {}
+
+def FpToIntOpIEEE754 : UnaryFloatingPointLibraryOp<"ieee754.fptoint"> {
```
cgyurgyik, line 523: Probably be a good idea to list of what rounding mode this is.

cgyurgyik, line 523: Does this work only for 32-bit floating points, or any ieee754 format, i.e., f16, f32, f64, f128, and f256?

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
cgyurgyik: Got it. Can we add an assertion at the very least?

```cpp
auto inputPorts = cell.getInputPorts();
assert inputPortIndex < inputPorts.size();
```

jiahanxie353: sure sounds good!

include/circt/Dialect/Calyx/CalyxPrimitives.td
jiahanxie353: yes for everything you listed except f256: https://github.com/llvm/circt/blob/2ba6439ae3e4081978b289ceffc4a78f1abeee44/lib/Dialect/Calyx/Export/CalyxEmitter.cpp#L1034 - at least from IR's perspective

There will still be some exceptions at runtime, and we can refer to section 7 of: https://www.jhauser.us/arithmetic/HardFloat-1/doc/HardFloat-Verilog.html

jiahanxie353: yes for everything you listed except f256: https://github.com/llvm/circt/blob/2ba6439ae3e4081978b289ceffc4a78f1abeee44/lib/Dialect/Calyx/Export/CalyxEmitter.cpp#L1034 - at least from IR's perspective

There will still be some exceptions at runtime, and we can refer to section 7 of: https://www.jhauser.us/arithmetic/HardFloat-1/doc/HardFloat-Verilog.html

cgyurgyik: Got it!

cgyurgyik (approved):
The rest of my comments are mostly nit picks, but the change in general looks good to me.

lib/Dialect/Calyx/Export/CalyxEmitter.cpp
jiahanxie353: wait, if `size_t inputPortIndex = cell.getInputPorts().size() - 2`, then wouldn't `inputPortIndex` guaranteed to be `< inputPorts.size()`. Did you mean to check `inputPortIndex >= 2` or something like that?

cgyurgyik: Yep sorry typing from a phone, mostly just wanting a bounds assertion so we don't get out of bounds UB case

jiahanxie353: no worries, sure!

PR #8294, opened 2025-03-04
[SCFToCalyx] Lower Arith Bitcast 

jiahanxie353:
This is my current proposal on how to lower `arith.bitcast` should be lowered, not 100% sure.

Since Calyx treat everything as bit vectors, I just keep every bit unchanged.

### DIFF
--- a/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
+++ b/lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
@@ -309,7 +309,7 @@ public:
                              /// floating point
                              AddFOp, SubFOp, MulFOp, CmpFOp, FPToSIOp, SIToFPOp,
                              /// others
-                             SelectOp, IndexCastOp, CallOp>(
+                             SelectOp, IndexCastOp, BitcastOp, CallOp>(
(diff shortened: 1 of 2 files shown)


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -1500,6 +1501,12 @@ LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
   return res;
 }
 
+LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
```
cgyurgyik, line 1504:
```suggestion
// The Calyx language treats values as bit vectors, i.e., there is no type system,
// so this is essentially a no-op.
LogicalResult BuildOpGroups::buildOp(PatternRewriter &rewriter,
```

cgyurgyik (approved):

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
cgyurgyik, line 1504: Please add a small unit test as well

PR #8299, opened 2025-03-05
[SMT] fix the declaration function name in ArrayBroadcastOp

Clo91eaf:
Fixed a minor issue in the SMT description. The SMT dialect does not have an smt.declare operation. It should be smt.declare_fun.

### DIFF
--- a/include/circt/Dialect/SMT/SMTArrayOps.td
+++ b/include/circt/Dialect/SMT/SMTArrayOps.td
@@ -75,7 +75,7 @@ def ArrayBroadcastOp : SMTArrayOp<"broadcast", [
     This operation represents a broadcast of the 'value' operand to all indices
     of the array. It is equivalent to
     ```
-    %0 = smt.declare "array" : !smt.array<[!smt.int -> !smt.bool]>
+    %0 = smt.declare_fun "array" : !smt.array<[!smt.int -> !smt.bool]>
     %1 = smt.forall ["idx"] {
     ^bb0(%idx: !smt.int):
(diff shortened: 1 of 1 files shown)


include/circt/Dialect/SMT/SMTArrayOps.td
```diff
@@ -75,7 +75,7 @@ def ArrayBroadcastOp : SMTArrayOp<"broadcast", [
     This operation represents a broadcast of the 'value' operand to all indices
     of the array. It is equivalent to
     ```
-    %0 = smt.declare "array" : !smt.array<[!smt.int -> !smt.bool]>
+    %0 = smt.declare_func "array" : !smt.array<[!smt.int -> !smt.bool]>
```
maerhart, line 78:
```suggestion
    %0 = smt.declare_fun "array" : !smt.array<[!smt.int -> !smt.bool]>
```

maerhart (approved):
Thanks for spotting this!

include/circt/Dialect/SMT/SMTArrayOps.td
Clo91eaf: Thanks for pointing out my bug fixes, the naming conventions are always `fun`😄

makslevental:
FYI this needs to be fixed here now https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/SMT/IR/SMTArrayOps.td#L78

Clo91eaf:
@makslevental Thank you for porting smt to mlir! I will upstream this typo fix pr to llvm later.

PR #8301, opened 2025-03-06
[sim] Add emission for plusargs for UPF simulations 

youngar:
This change adds an alternative method of supporting plusargs in UPF
simulations by ensuring that the plusargs found variable is compared
with case-equals (`found === 1'h1`) to squash the `x` value to `0`.


surajvellengar (approved):
anilkumar.marella verified this fix. He ran 3 tests and all are passing.

PR #8302, opened 2025-03-06
[LLHD] Hoist drives out of processes

fabianschuiki:
Extend the `HoistSignals` pass and hoist drives before halt and wait terminators out of the parent process. All drives that have no side-effecting operations in between them and the block's suspending terminator can be hoisted. If the same signal is driven multiple times in the same block, only the last drive is hoisted.

To hoist drives, the pass collects all drives for each driven signal slot and creates new yield operands and process results to carry the drive operands out of the process. Special care is taken to not create unnecessary results for values that are uniform across all drives and available outside the processes.

### DIFF
--- a/include/circt/Dialect/LLHD/IR/LLHDTypes.h
+++ b/include/circt/Dialect/LLHD/IR/LLHDTypes.h
@@ -23,4 +23,14 @@
 #define GET_ATTRDEF_CLASSES
 #include "circt/Dialect/LLHD/IR/LLHDAttributes.h.inc"
 
+namespace llvm {
+template <>
+struct PointerLikeTypeTraits<circt::llhd::TimeAttr>
+    : public PointerLikeTypeTraits<mlir::Attribute> {
+  static inline circt::llhd::TimeAttr getFromVoidPointer(void *p) {
+    return circt::llhd::TimeAttr::getFromOpaquePointer(p);
+  }
+};
+} // namespace llvm
+
 #endif // CIRCT_DIALECT_LLHD_IR_LLHDTYPES_H
--- a/lib/Dialect/LLHD/Transforms/HoistSignals.cpp
+++ b/lib/Dialect/LLHD/Transforms/HoistSignals.cpp
@@ -10,6 +10,7 @@
 #include "circt/Dialect/LLHD/Transforms/LLHDPasses.h"
 #include "mlir/Analysis/Liveness.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/IR/Matchers.h"
 #include "llvm/Support/Debug.h"
 
 #define DEBUG_TYPE "llhd-hoist-signals"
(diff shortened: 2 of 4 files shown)


test/Dialect/LLHD/Transforms/hoist-signals.mlir
```diff
@@ -92,10 +92,131 @@ hw.module @DontHoistLocalSignals() {
...
+    llhd.drv %b, %v after %0 : !hw.inout<i42>
+    // CHECK-NOT: llhd.drv %a, %v
+    llhd.drv %a, %v after %0 : !hw.inout<i42>
+    // CHECK-NOT: llhd.drv %b, %u
+    llhd.drv %b, %u after %0 : !hw.inout<i42>
+    // CHECK-NEXT: llhd.halt
     llhd.halt
   }
+  // CHECK: llhd.drv %a, %v
+  // CHECK: llhd.drv %b, %u
```
fabianschuiki, line 217: @maerhart noticed an issue here: the hoisted drives can't directly use the module-level values `%u` and `%v`, since any changes to them would cause the updated value to be driven onto `%a` and `%b`. Instead, these values must still be process results, such that they only update once when hitting the `llhd.halt`, and never again. Same for `llhd.wait`.

lib/Dialect/LLHD/Transforms/HoistSignals.cpp
```diff
@@ -166,6 +169,428 @@ void Hoister::hoistProbes() {
...
+
+  auto materialize = [&](DriveValue driveValue) {
+    if (auto value = driveValue.getValueOrNull())
+      return value;
+    auto attr = driveValue.getAttr();
+    auto &slot = materializedConstants[attr];
+    if (!slot) {
+      auto *op = processOp->getDialect()->materializeConstant(
+          builder, attr, driveValue.getType(), processOp.getLoc());
+      assert(op && "constant could not be materialized");
```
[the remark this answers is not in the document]
maerhart: and the LLHD dialect only has materializers for integers and time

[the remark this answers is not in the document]
fabianschuiki: Good point, so we should probably only try to materialize a limited set of constants, and keep the rest as SSA values even if the defining op is a `ConstantLike`.

fabianschuiki:
Fixed the issues pointed out by @maerhart. The pass now creates results even for values that are available outside the process, since the process may not execute as often as those values update. The constant materialization is now more limited: it only supports integer and time attributes, which makes this a lot less brittle.

include/circt/Dialect/LLHD/IR/LLHDTypes.h
```diff
@@ -23,4 +23,14 @@
...
+namespace llvm {
+template <>
+struct PointerLikeTypeTraits<circt::llhd::TimeAttr>
+    : public PointerLikeTypeTraits<mlir::Attribute> {
+  static inline circt::llhd::TimeAttr getFromVoidPointer(void *p) {
+    return circt::llhd::TimeAttr::getFromOpaquePointer(p);
+  }
+};
+} // namespace llvm
+
```
maerhart, line 35: Is that necessary for PointerUnion?

maerhart (approved):
Thanks!

include/circt/Dialect/LLHD/IR/LLHDTypes.h
fabianschuiki: Yes, it needs to know the number of available bits. It's unfortunate that TableGen doesn't produce this -- I've copied this snippet over from `mlir/IR/BuiltingAttributes.h` :disappointed:

PR #8305, opened 2025-03-06
[ESI] Simple FromHost DMA engine

teqdruid:
A simple, non-performant DMA-based from host communication channel.

  1) Host sends address of buffer address via MMIO write to register 0x08.
  2) Host sends address of completion address via MMIO write to register 0x10.
  3) Device reads data from said buffer and sends down the channel.
  4) Device writes '1' to the first byte of the completion buffer to signal that
     the transfer is done.
  5) Goto 1.

### DIFF
--- a/frontends/PyCDE/integration_test/esitester.py
+++ b/frontends/PyCDE/integration_test/esitester.py
@@ -28,7 +28,7 @@
 import pycde
 from pycde import AppID, Clock, Module, Reset, generator, modparams
 from pycde.bsp import get_bsp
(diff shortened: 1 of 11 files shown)


teqdruid:
@mikeurbach @mortbopet Now that this has merged and the to host DMA engine is there, we've got fully function host-device capabilities on XRT / Azure NP systems! It's nowhere near performant, but we've got it!

mikeurbach:
Very cool!

PR #8306, opened 2025-03-07
Bump LLVM to d90423e310482bdbc731242fa25dcb3dd44e69de.

mikeurbach:
This was mostly a clean bump.

There was one breaking change: after years, and based on discussion on PR review, a decision was made to finally do away with separate `match` and `rewrite` member functions on rewrite patterns:

https://github.com/llvm/llvm-project/commit/a21cfca

This was actually really smooth for us; only one pass with two patterns still used the split match and rewrite approach, and it was trivial to convert to a matchAndRewrite.


mikeurbach:
Gonna land this now that the work week is wrapping up.

PR #8307, opened 2025-03-07
Add a few mux-of-const and reg-of-and/or canonicalizers.

rwy7:
This PR adds four new canonicalization patterns to FIRRTL. 

```
mux(c, 0, x) ==> ~c && x
mux(c, 1, x) ==>  c && x
mux(c, x, 0) ==>  c || x
mux(c, x, 1) ==> ~c || x
```

These canonicalizations are already present for the comb dialect, but we want to run these canonicalizers before lowering layers, which can obscure constant ops behind ports, so I am porting them to the firrtl dialect.

The problem with these canonicalizers is, they interfere with preexisting canonicalizers for registers of this style:

```
connect reg, mux(reset, constant, reg) --> reg ==> 0
``` 

To get this behaviour back, this PR adds four additional canonicalizations for registers:

```
connect reg, reg && x --> reg ==> 0
connect reg, x && reg --> reg ==> 0
connect reg, reg || x --> reg ==> 1
connect reg, x || reg --> reg ==> 1
```

... and for regresets:

```
reset(reg) = 0 --> connect reg, reg && x --> reg ==> 0
reset(reg) = 0 --> connect reg, x && reg --> reg ==> 0
reset(reg) = 1 --> connect reg, reg || x --> reg ==> 1
reset(reg) = 1 --> connect reg, x || reg --> reg ==> 1
```

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -1472,12 +1472,12 @@ public:
 
 void MuxPrimOp::getCanonicalizationPatterns(RewritePatternSet &results,
                                             MLIRContext *context) {
-  results
-      .add<MuxPad, MuxSharedCond, patterns::MuxEQOperands,
-           patterns::MuxEQOperandsSwapped, patterns::MuxNEQ, patterns::MuxNot,
-           patterns::MuxSameTrue, patterns::MuxSameFalse,
-           patterns::NarrowMuxLHS, patterns::NarrowMuxRHS, patterns::MuxPadSel>(
-          context);
+  results.add<MuxPad, MuxSharedCond, patterns::MuxEQOperands,
+              patterns::MuxEQOperandsSwapped, patterns::MuxNEQ,
+              patterns::MuxNot, patterns::MuxSameTrue, patterns::MuxSameFalse,
+              patterns::NarrowMuxLHS, patterns::NarrowMuxRHS,
+              patterns::MuxPadSel, patterns::MuxLHS0, patterns::MuxLHS1,
+              patterns::MuxRHS0, patterns::MuxRHS1>(context);
 }
 
 void Mux2CellIntrinsicOp::getCanonicalizationPatterns(
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -1476,8 +1476,8 @@ void MuxPrimOp::getCanonicalizationPatterns(RewritePatternSet &results,
               patterns::MuxEQOperandsSwapped, patterns::MuxNEQ,
               patterns::MuxNot, patterns::MuxSameTrue, patterns::MuxSameFalse,
               patterns::NarrowMuxLHS, patterns::NarrowMuxRHS,
-              patterns::MuxPadSel, patterns::MuxLHS0, patterns::MuxLHS1,
-              patterns::MuxRHS0, patterns::MuxRHS1>(context);
+              patterns::MuxPadSel, patterns::MuxLhsZero, patterns::MuxLhsOne,
+              patterns::MuxRhsZero, patterns::MuxRhsOne>(context);
 }
 
 void Mux2CellIntrinsicOp::getCanonicalizationPatterns(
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -2256,6 +2256,73 @@ struct FoldResetMux : public mlir::RewritePattern {
...
 
+namespace {
+// This canonicalizer provides the following patterns:
+//
+// - and(reg, x) -> 0 if reset is 0
+// - or(reg, x)  -> 1 if reset is 1
+//
+// Justification: if we assume the register currently has the value of its
+// reset, and if that means the next value of the register is unchanged, then we
+// can assume the register will always have its reset value.
```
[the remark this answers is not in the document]
rwy7: I can write that up! I actually deleted this note before I saw your comment, but I will add it back.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -2256,6 +2256,73 @@ struct FoldResetMux : public mlir::RewritePattern {
...
+          rewriter.eraseOp(connect);
+          replaceOpAndCopyName(rewriter, op, reset.getResult());
+          return success();
+        }
+      }
+    }
+
+    return failure();
+  }
+};
```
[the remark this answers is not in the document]
rwy7: I don't know if that makes sense, since this isn't so much a pattern on the register op, but more, a pattern on the connect driving the register. Maybe I could do it using some new constraints. I took a quick look around and it doesn't look like we do this kind of thing using ODS. I'm interested in suggestions though.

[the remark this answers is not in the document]
seldridge: I see, the problem isn't in generating the match, but the replacement. The match would be something like: `ConnectOp(RegResetOp(...), and(RegResetOp(...), _)` (where the registers are the same). It's that the replacement isn't a replacement of the `ConnectOp`, but of the `RegResetOp`. Yeah, I'm not sure exactly how to do this.

rwy7:
> The symmetry of these may motivate a canonical mux representation. I.e., this could be written as four canonicalizers for each of the four permutations of where a single binary constant could appear or as two canonicalizers for two of the permutations and then one canonicalizer that puts constants in "canonical" position. "Constants last" is the canonical form chosen below:
> 
> ```
> mux(cond, a, 0) -> and(cond, a)
> mux(cond, a, 1) -> or(not(cond), a)
> mux(cond, constant, nonconstant) -> mux(~cond, nonconstant, constant)
> ```
> 
> Do we already have a canonical form form muxes?
> 

This is a cool idea! I don't see anything like this, but I do see a [conflicting canonicalization](https://github.com/llvm/circt/blob/main/lib/Dialect/Comb/CombFolds.cpp#L2473-L2482) in comb:

```
mux(!p, x, y) -> mux(p, y, x)
```

(this canonicalization is missing in FIRRTL)

Usually our canonicalizers are reducing the cost of a computation, but negating the condition would increase the cost. Maybe, putting constants last only makes sense when the op is symmetric and can freely rearrange its operands.

rwy7:
> Have you gotten any data from internal designs?

For the most part, these canonicalizers just shuffle around the source-info comments and variable spilling in the verilog. There are a few places where the expressions are different but not meaningfully so, and I've only spotted one place where a mux actually was optimized away.

seldridge:
> Usually our canonicalizers are reducing the cost of a computation, but negating the condition would increase the cost. 

Yes, this would be bad. Given that, I think the right approach is to keep the four canonicalizers and, in a follow-on, add the `mux(!cond, a, b) -> mux(cond, b, a)`.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -2256,6 +2256,77 @@ struct FoldResetMux : public mlir::RewritePattern {
 };
 } // namespace
 
+namespace {
+/// This canonicalizer provides the following patterns:
+///   reset(reg) = 0 ==> connect(reg, and(reg, x)) ==> reg -> 0
+///   reset(reg) = 0 ==> connect(reg, and(x, reg)) ==> reg -> 0
+///   reset(reg) = 1 ==> connect(reg, or(reg, x))  ==> reg -> 1
+///   reset(reg) = 1 ==> connect(reg, or(x, reg))  ==> reg -> 1
```
seldridge, line 2264: Mega-nit: This syntax is bothering me and took me a while to figure it out. The `==>` is an "AND" in the first use and is "is converted to" (or "IMPLIES"?) in the second.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -2256,6 +2256,77 @@ struct FoldResetMux : public mlir::RewritePattern {
...
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(RegResetOp op,
+                                PatternRewriter &rewriter) const override {
+    // Do not fold the register away if it is important.
+    if (hasDontTouch(op.getOperation()) || !AnnotationSet(op).empty() ||
+        op.isForceable())
+      return failure();
+
+    // This canonicalization only applies when the register holds 1 bit.
+    auto type = dyn_cast<UIntType>(op.getResult().getType());
```
seldridge, line 2282: I think `SIntType` would work, too?

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -2256,6 +2256,77 @@ struct FoldResetMux : public mlir::RewritePattern {
...
+    if (!type || type.getWidthOrSentinel() != 1)
+      return failure();
+
+    // This canonicalization only applies when the reset is a constant.
+    auto reset =
+        dyn_cast_or_null<ConstantOp>(op.getResetValue().getDefiningOp());
+    if (!reset)
+      return failure();
```
seldridge, line 2290: Nit: please be exact. This is the _reset value_ and not the _reset_ (or exactly, the _reset signal_).

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -2256,6 +2256,77 @@ struct FoldResetMux : public mlir::RewritePattern {
...
+/// Justification: The initial value of a register is indeterminant, which means
+/// we are free to choose any initial value when optimizing the circuit. For the
+/// AND patterns, if the reset is zero, and we assume the initial value is zero,
+/// then the register will always be zero. For the OR patterns, if the reset is
+/// one, and we assume the initial value is one, then the register will always
+/// be one.
+struct RegResetAndOrOfSelf : public mlir::OpRewritePattern<RegResetOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(RegResetOp op,
+                                PatternRewriter &rewriter) const override {
```
seldridge, line 2275: Supreme nit: 
```suggestion
  LogicalResult matchAndRewrite(RegResetOp reg,
                                PatternRewriter &rewriter) const override {
```

When reading the later rewrite pattern, I found it clearer to use the specific "reg" instead of the generic "op".

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -3156,10 +3229,66 @@ static LogicalResult foldHiddenReset(RegOp reg, PatternRewriter &rewriter) {
...
+///   connect(reg, and(reg, x)) ==> reg -> 0
+///   connect(reg, and(x, reg)) ==> reg -> 0
+///   connect(reg, or(reg, x))  ==> reg -> 1
+///   connect(reg, or(x, reg))  ==> reg -> 1
+///
+/// Justification: The initial value of a register is indeterminant, which means
+/// We are free to choose any initial value when optimizing the circuit. For the
+/// AND patterns, if we assume the initial value is zero, then the register will
+/// always be zero. For the OR patterns, if we assume the initial value is one,
+/// then the register will always be one.
```
seldridge, line 3242: Great for including this on both rewrite patterns! Providing this as a comment makes it clear not only what the pattern is doing, but _why_ and with what justification in the spec we are using. 💯

seldridge (approved):
LGTM

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
rwy7: Ah OK I can delete these comments, I don't think they're helpful. FWIW `==>` is implication, `->` is canonicalization.

dtzSiFive:
Friendly ping -- what's the status of this? Does this need more review or anything else? Thanks!

PR #8309, opened 2025-03-10
add support for String literal operations

ccccccccccccccccvbnb:
https://chipsalliance.github.io/sv-tests-results/tests/chapter-11/11.10.1--string_compare.sv.html
add support like display(a) and $display(":assert:('%s' == 'testtesttesttest')", str);
string is available to be used in display function

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -1272,7 +1272,39 @@ Value Context::convertToSimpleBitVector(Value value) {
                                                  value);
     }
   }
-
+  if(llvm::isa<moore::StringType>(value.getType())){
+    auto readop = value.getDefiningOp();
+    if(auto declop = readop->getOperand(0).getDefiningOp()){
+      Operation *assignop = nullptr;
+      for(auto user: declop->getResults().getUsers()){
+          if(llvm::isa<moore::BlockingAssignOp>(user)){
+            assignop = user;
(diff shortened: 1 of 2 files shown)


hailongSun2000:
The `string` type will be transformed into the `bit` type starting at 0 in the slang AST if I remember correctly. And if you want to solve string assignment ops, I think we have completed that. Please check the related test cases (https://github.com/llvm/circt/blob/main/test/Conversion/ImportVerilog/basic.sv). 😃

ccccccccccccccccvbnb:
> The `string` type will be transformed into the `bit` type starting at 0 in the slang AST if I remember correctly. And if you want to solve string assignment ops, I think we have completed that. Please check the related test cases (https://github.com/llvm/circt/blob/main/test/Conversion/ImportVerilog/basic.sv). 😃

Maybe I didn't explain it clearly. Currently, when import-verilog processes strings, it converts the bit vector to the !moore.string type through convertOp. When display(str), because str is of string type, it does not match the input of display, so I add another convertOp to convert string to bit vector , as well as support for NamedValueExpression and ConcatenationExpression in display (previously, display function could only process stringliteral types).
for example:
```
string c = "hello";
```
will be translate as:
```
    %0 = moore.string_constant "hello" : i40
    %1 = moore.conversion %0 : !moore.i40 -> !moore.string
    %c = moore.variable %1 : <string>
```
I add correspond test case at the bottom of basic.sv

PR #8311, opened 2025-03-12
[FIRRTL] Add simulation op and parser support

fabianschuiki:
Add the `firrtl.simulation` op, which declares a simulation unit test and references a module to act as the top-level for the simulation. The target module must have exactly four ports, clock, init, done, and success, with corresponding directions and types. This new op is analogous to the `firrtl.formal` op.

Also add support for a `simulation` construct in the parser, also analogous to the `formal` construct.

A later commit will add a lowering to `verif.formal`.


lib/Dialect/FIRRTL/Export/FIREmitter.cpp
```diff
@@ -407,9 +410,8 @@ void Emitter::emitCircuit(CircuitOp op) {
             emitModule(op);
             ps << PP::newline;
           })
-          .Case<LayerOp>([&](auto op) { emitDeclaration(op); })
-          .Case<OptionOp>([&](auto op) { emitDeclaration(op); })
-          .Case<FormalOp>([&](auto op) { emitDeclaration(op); })
+          .Case<LayerOp, OptionOp, FormalOp, SimulationOp>(
+              [&](auto op) { emitDeclaration(op); })
```
seldridge, line 414: Nice simplification.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -3395,16 +3395,82 @@ void RegResetOp::getAsmResultNames(OpAsmSetValueNameFn setNameFn) {
...
+  auto isBool = [](Type type) {
+    if (auto uintType = dyn_cast<UIntType>(type))
+      return uintType.getWidth() == 1;
+    return false;
+  };
+  return success(checkPort(0, "clock", Direction::In, isClock, "clock") &&
+                 checkPort(1, "init", Direction::In, isBool, "uint<1>") &&
+                 checkPort(2, "done", Direction::Out, isBool, "uint<1>") &&
+                 checkPort(3, "success", Direction::Out, isBool, "uint<1>"));
+}
```
seldridge, line 3472: This verification is annoyingly duplicated for all simulation ops.

There are two ways to resolve this:

1. The verification is inverted. I.e., this port restriction is moved onto the module. The module then looks for symbol users and if it is used by a simulation op, then it checks its port structure. I'm not sure if this kind of verification is allowed.
2. The IR for this should be changed so that the simulation op is a list of tests (or something else which collapses this down).

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5481,6 +5499,10 @@ ParseResult FIRCircuitParser::parseToplevelDefinition(CircuitOp circuit,
     if (getToken().getKind() == FIRToken::kw_module)
       return parseModule(circuit, /*isPublic=*/true, indent);
     return emitError(getToken().getLoc(), "only modules may be public");
+  case FIRToken::kw_simulation:
+    if (requireFeature({4, 2, 0}, "simulation tests"))
```
seldridge, line 5503: I just released 4.2.0 with some minor fixes around `%c`. We're into 5.0.0 territory given that I have merged the `{{}}` work. If you want to get the simulation tests onto a 4.3.0, I can do that for you. Otherwise, I think this is going to be 5.0.0.

test/Dialect/FIRRTL/parse-basic.fir
```diff
@@ -1889,6 +1889,33 @@ circuit Foo:
...
+  ; CHECK-SAME: a_int = 42
+  ; CHECK-SAME: b_string = "hello"
+  ; CHECK-SAME: c_array = [42 : ui32, "hello", [9001 : ui32], {foo = 1337 : ui32}]
+  ; CHECK-SAME: d_map = {a = 42 : ui32, b = "hello", c = [9001 : ui32], d = {foo = 1337 : ui32}}
+  ; CHECK-SAME: }
+  simulation someNameB of Foo:
+    a_int = 42
+    b_string = "hello"
+    c_array = [42, "hello", [9001], {foo = 1337}]
+    d_map = {a = 42, b = "hello", c = [9001], d = {foo = 1337}}
```
seldridge, line 1917: Neat! I see that you have a syntax in mind for this, it's just not in the FIRRTL spec PR. I'm fine with this syntax.

seldridge (approved):
LGTM modulo some verification slowness I'm worried about.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
fabianschuiki: I think (1) might have worse performance than the existing solution, since the `verifSymbolUses` gets a SymbolTable to make the symbol lookup efficient. If we flipped that around, the verifier for `FModuleOp` would have to scan the entire IR for uses (can't use cached symbol user maps in a verifier). It feels wrong to have verification for `firrtl.simulation` split and put onto `firrtl.module`.

(2) could work! I'm pretty sure the performance overhead here is negligible, though. The input will have roughly 1 simulation op per module, maybe a handful in the worst case. I think that pales in comparison to the port list verification that frequent `func.call` and `hw.instance` ops would have to do.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5391,19 +5395,33 @@ ParseResult FIRCircuitParser::parseModule(CircuitOp circuit, bool isPublic,
...
 
-/// formal-old ::= 'formal' id 'of' id ',' 'bound' '=' int info?
-/// formal-new ::= 'formal' id 'of' id ':' info? INDENT (param NEWLINE)* DEDENT
+/// formal ::= 'formal' formal-like
 ParseResult FIRCircuitParser::parseFormal(CircuitOp circuit, unsigned indent) {
   consumeToken(FIRToken::kw_formal);
+  return parseFormalLike<FormalOp>(circuit, indent);
+}
+
+/// formal ::= 'formal' formal-like
```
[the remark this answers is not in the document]
fabianschuiki: Woops, nice catch thanks!

lib/Dialect/FIRRTL/Import/FIRParser.cpp
fabianschuiki: Does the `{{}}` break backwards compatibility? If that's the case, 5.0.0 sounds great! Otherwise, if existing 4.2.0 FIR would continue to work with `{{}}`, 4.3.0 might make more sense. I'll bump the `nextFIRVersion` to whatever we decide.

fabianschuiki: Should I switch the `{{}}` feature to `nextFIRVersion` as well for consistency? To be bumped/replaced once we cut the spec release?

PR #8312, opened 2025-03-12
[FIRRTL] Lower firrtl.simulation to verif.simulation

fabianschuiki:
Add a lowering from `firrtl.simulation` to `verif.simulation` in the FIRRTL-to-HW conversion. The lowering creates a `verif.simulation` op that simply instantiates the module pointed to by `firrtl.simulation`, connecting its `clock` and `init` ports to the simulation op's block arguments, and yielding the `done` and `success` ports back. We may want to inline the instance in the future, but that is purely cosmetic.

This now allows FIR files to contain `simulation` unit tests that can be lowered to HW through `firtool` and then executed using `arcilator`.


lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -678,6 +681,19 @@ void FIRRTLModuleLowering::runOnOperation() {
...
+              auto newSimulationOp = builder.create<verif::SimulationOp>(
+                  loc, oldSimulationOp.getNameAttr(),
+                  oldSimulationOp.getParametersAttr());
+              auto &body = newSimulationOp.getRegion().emplaceBlock();
+              body.addArgument(seq::ClockType::get(builder.getContext()), loc);
+              body.addArgument(builder.getI1Type(), loc);
+              state.recordModuleMapping(oldSimulationOp, newSimulationOp);
+              simulationOpsToProcess.push_back(newSimulationOp);
+              return success();
+            })
```
seldridge, line 696: Nit: I know you're mirroring the `FormalOp` handling above, but using `old`/`oldOp` and `newOp` would make this easier to read.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1447,6 +1470,35 @@ FIRRTLModuleLowering::lowerFormalBody(verif::FormalOp formalOp,
...
+  auto oldSimulationOp =
+      cast<SimulationOp>(loweringState.getOldModule(simulationOp));
+  auto moduleName = oldSimulationOp.getModuleNameAttr().getAttr();
+  auto oldModule = cast<FModuleLike>(
+      *loweringState.getInstanceGraph().lookup(moduleName)->getModule());
+  auto newModule =
+      dyn_cast_or_null<hw::HWModuleLike>(loweringState.getNewModule(oldModule));
+  if (!newModule)
+    return oldSimulationOp->emitOpError()
+           << "could not find module " << oldModule.getNameAttr();
```
seldridge, line 1490: Can this check be removed since I _think_ it is already guaranteed by a verifier?

seldridge (approved):
LGTM!

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
fabianschuiki: That's a good point! Let me change these to casts, and make things bit less verbose.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
fabianschuiki: Done :+1:

PR #8320, opened 2025-03-14
[FIRRTL] Canonicalize Assert(clk, p, e) when p == e

rwy7:
If the predicate and enable are the same expression, then the predicate must be true whenever the enable is true, so the assert cannot fire, and we can safetly remove the assert.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLCanonicalization.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLCanonicalization.td
@@ -16,6 +16,7 @@
 include "FIRRTLDeclarations.td"
 include "FIRRTLExpressions.td"
 include "FIRRTLStatements.td"
+include "FIRRTLIntrinsics.td"
 include "mlir/IR/OpBase.td"
 include "mlir/IR/PatternBase.td"
 
@@ -735,5 +736,12 @@ def CVTUnSigned : Pat<
       (PadPrimOp $x,
         (NativeCodeCall<"$_builder.getI32IntegerAttr(type_cast<SIntType>($0.getType()).getBitWidthOrSentinel())"> $old)))),
   [(UIntType $x), (KnownWidth $old)]>;
+
+def AssertXWhenX : Pattern<(AssertOp $_, $x, $x, $_, $_, $_, $_, $_), []>;
+
+def AssumeXWhenX : Pattern<(AssumeOp $_, $x, $x, $_, $_, $_, $_, $_), []>;
+
+def UnclockedAssumeIntrinsicXWhenX
+    : Pattern<(UnclockedAssumeIntrinsicOp $x, $x, $_, $_, $_), []>;
 
 #endif // CIRCT_DIALECT_FIRRTL_FIRRTLCANONICALIZATION_TD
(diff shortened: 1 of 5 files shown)


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -3194,6 +3194,24 @@ static LogicalResult eraseIfZeroOrNotZero(Operation *op, Value predicate,
...
+  LogicalResult matchAndRewrite(T op,
+                                PatternRewriter &rewriter) const override {
+    if (op.getPredicate() == op.getEnable()) {
+      rewriter.eraseOp(op);
+      return success();
+    }
+
+    return failure();
+  }
+};
```
seldridge, line 3212: Challenge 1: Can this be written using ODS or does that not allow op removal?

seldridge (commented):
Makes sense to me!

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
rwy7: I honestly didn't think it was possible, but turned out to be quite easy 👍

include/circt/Dialect/FIRRTL/FIRRTLCanonicalization.td
```diff
@@ -736,4 +737,16 @@ def CVTUnSigned : Pat<
...
+  []>;
+
+def AssumeXWhenX : Pattern<
+  (AssumeOp:$old $clk, $x, $x, $msg, $subs, $name, $conc, $ctrl),
+  []>;
+
+def UnclockedAssumeIntrinsicXWhenX : Pattern<
+  (UnclockedAssumeIntrinsicOp:$old $x, $x, $msg, $subs, $name),
+  []>;
+
```
seldridge, line 751: Very nice!

seldridge: Nit: I think you can use wildcards for much of these, e.g.,:

```
(AssertOp $_, $x, $x, $_, $_, $_, $_, $_)
```

include/circt/Dialect/FIRRTL/FIRRTLCanonicalization.td
```diff
@@ -736,4 +737,12 @@ def CVTUnSigned : Pat<
         (NativeCodeCall<"$_builder.getI32IntegerAttr(type_cast<SIntType>($0.getType()).getBitWidthOrSentinel())"> $old)))),
   [(UIntType $x), (KnownWidth $old)]>;
 
+def AssertXWhenX : Pattern<(AssertOp:$_ $_, $x, $x, $_, $_, $_, $_, $_), []>;
+
+def AssumeXWhenX : Pattern<(AssumeOp:$_ $_, $x, $x, $_, $_, $_, $_, $_), []>;
+
+def UnclockedAssumeIntrinsicXWhenX : Pattern<
+  (UnclockedAssumeIntrinsicOp:$_ $x, $x, $_, $_, $_),
+  []>;
```
seldridge, line 746: Super nit: The op result can be ignored. I think the following works:

```suggestion
def AssertXWhenX : Pattern<(AssertOp $_, $x, $x, $_, $_, $_, $_, $_), []>;

def AssumeXWhenX : Pattern<(AssumeOp $_, $x, $x, $_, $_, $_, $_, $_), []>;

def UnclockedAssumeIntrinsicXWhenX : Pattern<
  (UnclockedAssumeIntrinsicOp $x, $x, $_, $_, $_),
  []>;
```

seldridge (approved):
LGTM

PR #8321, opened 2025-03-17
[OM] Add missing CAPI omTypeIsAMapType

unlsycn:
This API exists in OM.h but is missing an implementation.

### DIFF
--- a/lib/CAPI/Dialect/OM.cpp
+++ b/lib/CAPI/Dialect/OM.cpp
@@ -87,6 +87,9 @@ MlirType omStringTypeGet(MlirContext ctx) {
   return wrap(StringType::get(unwrap(ctx)));
 }
 
+/// Is the Type a MapType.
+bool omTypeIsAMapType(MlirType type) { return isa<MapType>(unwrap(type)); }
+
 /// Return a key type of a map.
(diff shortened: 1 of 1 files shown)


mikeurbach (approved):
Thanks for adding this. The reason we never needed this is the people using this actually don't use Maps for anything...

There was some discussion in Chisel a while back, but basically, Map and Tuple are redundant with what you can express with Class, Object, and List.

I will open a ticket to track their removal, but basically, everything related to Map and Tuple are going to be removed.

unlsycn:
> Thanks for adding this. The reason we never needed this is the people using this actually don't use Maps for anything...
> 
> There was some discussion in Chisel a while back, but basically, Map and Tuple are redundant with what you can express with Class, Object, and List.
> 
> I will open a ticket to track their removal, but basically, everything related to Map and Tuple are going to be removed.

Thanks for your explanation! These two patches are just for our library in [zaozi](https://github.com/sequencer/zaozi), so feel free to leave it for now or land it if it's convenient.

mikeurbach:
Makes sense. Normally I wouldn't want to maintain too much based on IR constructs we know we're removing, but thse two patches should be fine.

PR #8322, opened 2025-03-17
[OM] Fix tuple_get operation in OMEvaluator

unlsycn:
The Evaluator will call `getOrCreateValue` when instantiating an Object, which this function lacks the evaluation of the `tuple_get' operation.

~~Further tests will be added later.~~
Done.


mikeurbach:
I could approve this if there were a test that failed before to demonstrate the problem.

But like I said on the other PR--we actually want to get rid of Map and Tuple code paths.

sequencer:
I'd like to remove the Tuple and Map for reducing the maintain burden.

mikeurbach (approved):
Agreed that we don't need to maintain this, but approving this PR if we want to fix this before we have time to rip it all out.

PR #8325, opened 2025-03-18
[OM] Fix finalization of nested ReferenceValue

unlsycn:
Further tests will be added later.


unlsycn:
cc @mikeurbach 
Would you like to review this PR when you have a moment?
Thanks a lot!

mikeurbach (commented):
I think this makes sense, but can you please include more info about what this is trying to resolve? This would be good info to have in the commit message when it merges. I think I can understand it from the test case, but in general, we request to have good commit messages: https://llvm.org/docs/DeveloperPolicy.html#commit-messages

@uenoku please take a look as well

mikeurbach (approved):
I think the actual change makes.

unlsycn:
> I think this makes sense, but can you please include more info about what this is trying to resolve? This would be good info to have in the commit message when it merges. I think I can understand it from the test case, but in general, we request to have good commit messages: https://llvm.org/docs/DeveloperPolicy.html#commit-messages
> 
> @uenoku please take a look as well

Sure!

mikeurbach:
Yes, I think so. I briefly mentioned it to @uenoku , but we can merge without his approval.

uenoku (approved):
Sorry for delay. It looks great to me.

PR #8326, opened 2025-03-18
[LowerToBMC] Add `comb` as a dialect dependency

TaoBi22:
`comb` is currently missing as a dependency of `LowerToBMC`, which means it's not loaded on designs with no comb ops in them and crashes. Regression test included.


maerhart (approved):
Thanks!

PR #8327, opened 2025-03-18
[LowerSMTToZ3LLVM] Change printf type provided to lookupOrCreateFn

TaoBi22:
Looks like the debug mode of `LowerSMTToZ3LLVM` (which is used when `--print-solver-output` is passed to circt-bmc) broke in https://github.com/llvm/circt/pull/8225 - there was [a commit in LLVM](https://github.com/llvm/llvm-project/pull/123378) that added some extra checks on the result type passed to `lookupOrCreateFn`, and it looks like it's expecting a void type rather than an i32. I don't know too much about this side of things so LMK if this isn't the preferred fix.

### DIFF
--- a/integration_test/Dialect/SMT/basic.mlir
+++ b/integration_test/Dialect/SMT/basic.mlir
@@ -247,26 +247,26 @@ func.func @check(%expr: !smt.bool) {
   smt.assert %expr
   %0 = smt.check sat {
     %1 = llvm.mlir.addressof @sat : !llvm.ptr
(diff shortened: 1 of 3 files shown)


maerhart (approved):
Thanks for fixing this! The more robust fix would be to use the more concrete llvm function helper where we don't have to provide the type of the function, but this is also fine.

TaoBi22:
Yeah good point, I'll merge this for now to fix the bug but agreed that it would make sense to eventually swap to using the helper

PR #8329, opened 2025-03-19
[SV] Add preprocessor include op

rwy7:
Add a backtick-include directive. The include target is just specified as a string attribute.

### DIFF
--- a/include/circt/Dialect/SV/SVStatements.td
+++ b/include/circt/Dialect/SV/SVStatements.td
@@ -913,6 +913,22 @@ def MacroDefOp : SVOp<"macro.def",
   }];
 }
 
+def IncludeStyle
+    : I32EnumAttr<
+          "IncludeStyle",
+          "Double-quoted local include, or angle bracketed system include",
+          [I32EnumAttrCase<"Local", 0, "local">,
+           I32EnumAttrCase<"System", 1, "system">]> {
+  let genSpecializedAttr = 0;
+}
+
+def IncludeStyleAttr : EnumAttr<SVDialect, IncludeStyle, "include_style">;
+
+def IncludeOp : SVOp<"include"> {
+  let summary = "preprocessor `include directive";
+  let arguments = (ins IncludeStyleAttr:$style, StrAttr:$target);
+  let assemblyFormat = [{ $style $target attr-dict }];
+}
 
 //===----------------------------------------------------------------------===//
 // Function/Call
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -4062,6 +4062,7 @@ private:
   LogicalResult visitSV(FuncCallProceduralOp op);
   LogicalResult visitSV(FuncCallOp op);
   LogicalResult visitSV(ReturnOp op);
+  LogicalResult visitSV(IncludeOp op);
 
 public:
   ModuleEmitter &emitter;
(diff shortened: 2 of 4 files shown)


include/circt/Dialect/SV/SVStatements.td
```diff
@@ -913,6 +913,11 @@ def MacroDefOp : SVOp<"macro.def",
   }];
 }
 
+def IncludeOp : SVOp<"include"> {
+  let summary = "preprocessor `include directive";
+  let arguments = (ins StrAttr:$target);
+  let assemblyFormat = [{ $target attr-dict }];
+}
```
seldridge, line 920: I was going to suggest that the parent should be an MLIR moduleOp. However, the SystemVerilog spec allows this to be used anywhere in the text. 😬 This is correct to allow it anywhere.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -4451,6 +4452,14 @@ LogicalResult StmtEmitter::visitSV(ReturnOp op) {
   return emitOutputLikeOp(op, ports);
 }
 
+LogicalResult StmtEmitter::visitSV(IncludeOp op) {
+  startStatement();
+  ps << "`include" << PP::nbsp << "\"" << op.getTarget() << "\"";
+  emitLocationInfo(op.getLoc());
+  setPendingNewline();
+  return success();
+}
```
seldridge, line 4461: There are two types of emission here. Quotes and angle brackets. They have different behavior. For completeness, can the PR add support for this in the ODS description of the op to indicate if it should be "double quote" or "angle brackets". See Section 22.4 of 1800-2023 for more information.

rwy7: Updated to include both types. I added an enum, and named the options "IncludeStyle::Local" and "IncludeStyle::System". I don't really want to call them "double-quoted" and "angle-bracketed", but I'm worried local/system are misnomers.

include/circt/Dialect/SV/SVStatements.td
rwy7: A related question I had was, what happens with an IncludeOp located in the top-level MLIR module when we are emitting multiple files. I followed the example of MacroDefOp, and have it replicated into each output file.

seldridge: I think replicating makes sense and aligns with other ops.

Ideally, we should move everything to use the emit ops which solves this problem as this is then explicit.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
seldridge: I was hoping that the spec provided guidance on this, but it's not clear and seems to actually refer to them as "double quote" and "angle brackets". So, the original suggestion was regrettably intentional. 🥲 The spec does talk about the double quotes being "relative to the compiler's current working directory" and the angle brackets being "implementation dependent". You could use "standard"/"normal" and "implementation dependent".

I won't hold this up over that, I just like to have something that points directly back to what it is in the spec. And, again regrettably, the spec hasn't really named these.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -6713,7 +6728,7 @@ void SharedEmitterState::gatherFiles(bool separateModules) {
           collectPorts(op);
           // External modules are _not_ emitted.
         })
-        .Case<VerbatimOp, IfDefOp, MacroDefOp, FuncDPIImportOp>(
+        .Case<VerbatimOp, IfDefOp, MacroDefOp, sv::IncludeOp, FuncDPIImportOp>(
```
seldridge, line 6731: Consistency nit: 
```suggestion
        .Case<VerbatimOp, IfDefOp, MacroDefOp, IncludeOp, FuncDPIImportOp>(
```

seldridge (approved):
LGTM

lib/Conversion/ExportVerilog/ExportVerilog.cpp
rwy7: I'm just going to stick with with local/system 🤷

PR #8331, opened 2025-03-19
[FIRRTL] Add doNotPrint flag to InstanceOp

rwy7:
This flag acts the same as doNotPrint in hardware. When lowering an instance to the hw dialect, copy the doNotPrint attribute over.

I am preparing to add a bind op to the firrtl dialect (not the frontend language), which will pair with this flag. My goal is to allow lower-layers to place the bind op for layerblocks into the body of a bind file, which is represented by an emit.file op.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
@@ -64,61 +64,57 @@ def InstanceOp : HardwareDeclOp<"instance", [
     ```
   }];
 
-  let arguments = (ins FlatSymbolRefAttr:$moduleName, StrAttr:$name, NameKindAttr:$nameKind,
-                       DenseBoolArrayAttr:$portDirections, StrArrayAttr:$portNames,
-                       AnnotationArrayAttr:$annotations,
-                       PortAnnotationsAttr:$portAnnotations,
-                       LayerArrayAttr:$layers,
-                       UnitAttr:$lowerToBind,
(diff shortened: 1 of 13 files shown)


seldridge (commented):
I think this is practical.

However, `doNotPrint` feels a little weak. Specifically, what does this mean if there is no bind op? Will this just cause the instance to not show up in Verilog and won't that allow for the creation of illegal circuits?

Would one of the following be a better way to model this?

- There is a dedicated bind instance op which takes a symbol/inner symbol that must resolve to op in an emit region. This lookup would likely hing on the ability to do nested symbol lookups into the `emit.file`:

``` mlir
emit.file "bind.sv" sym @Bind {
  firrtl.bind @a
}

firrtl.module @Foo {
  firrtl.instance.bind bar @Bar() {bindLocation = @Bind::a}
}
```

- The referencing in the above is inversed to allow for the `firrtl.bind` to refer to the `firrtl.instance.bind`. This may be hard as it looks like `emit.file` has a symbol table and would therefore need to opt into inner symbol tables to make this work.

Could you elaborate a bit on what the full solution would look like?

seldridge (approved):
Thanks for the offline discussion of where this is going. This will give us finer control over where the bind op is going to live which is what we need.

In the follow-on that adds a `firrtl.bind`, as long as this removes `lowerToBind`, this makes sense to me.

rwy7:
Opened an issue to track the work for "removing lowerToBind": https://github.com/llvm/circt/issues/8386
The PR to add a bind op is here: https://github.com/llvm/circt/pull/8384

I'll remove the lowerToBind in a later PR.

Issue #8231, opened 2025-02-12
Tests that require Z3 fail rather than being unsupported


mikeurbach:
I've been dealing with something in my circt integration tests for a while, and it has been an annoyance, but not a major issue. However, it actually masked a problem in https://github.com/llvm/circt/pull/8225. Basically, I haven't configured z3, and any test with `// REQUIRES: libz3` fails, rather than being unsupported. Does anyone else see this?

Separately, I need to just get z3 set up so I can run these tests during these LLVM bumps, but the current behavior seems wrong.

I didn't see any open issue about this, but please redirect me if this is a dupe.

For completeness, here is how I configure things:

```
cmake -B build -G Ninja llvm/llvm \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_C_COMPILER=clang \
    -DCMAKE_CXX_COMPILER=clang++ \
    -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
    -DLLVM_ENABLE_PROJECTS=mlir \
    -DLLVM_ENABLE_ASSERTIONS=ON \
    -DLLVM_EXTERNAL_PROJECTS=circt \
    -DLLVM_EXTERNAL_CIRCT_SOURCE_DIR=. \
    -DLLVM_TARGETS_TO_BUILD=host \
    -DLLVM_LIT_ARGS="-v --show-unsupported" \
    -DLLVM_USE_LINKER=gold \
    -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
    -DCIRCT_BINDINGS_PYTHON_ENABLED=ON
```

I will note that I see this in the output:

```
CMake Warning at cmake/modules/FindZ3.cmake:116 (message):
  Failed to determine Z3 library version, defaulting to 0.0.0.
Call Stack (most recent call first):
  /scratch/verif/mikeu/circt/CMakeLists.txt:474 (find_package)


-- Could NOT find Z3 (missing: Z3_LIBRARIES Z3_INCLUDE_DIR) (found version "0.0.0")
-- Did NOT find z3.
```

teqdruid:
I get that message at config time but I don't get a test failure as a
result.

TaoBi22:
Took a poke around and think I've figured out what the problem is here, will get a PR in shortly

mikeurbach:
Confirmed the fix does the trick; on my test machine without Z3, these tests are now correctly marked as unsupported. Thanks @TaoBi22 .

PR #8333, opened 2025-03-20
[SMT] expand Z3 library check in integration tests

TaoBi22:
I managed to recreate the issue reported in #8231, and upon some poking around found out that the issue was that, on my system at least, the Z3 library is set to `Z3_LIBRARIES-NOTFOUND` by the code that searches for a Z3 installation when there is no available Z3 library, rather than being left as an empty string. This expands the check to check if it's either an empty string or `Z3_LIBRARIES-NOTFOUND`. Hopefully this also fixes it for you @mikeurbach - if not then I'd be curious to know what is assigned to `config.z3_library` in your `build/integration_test/lit.site.cfg.py`

### DIFF
--- a/integration_test/lit.cfg.py
+++ b/integration_test/lit.cfg.py
@@ -213,7 +213,7 @@ if config.z3_path != "":
   config.available_features.add('z3')
 
 # Enable libz3 if it has been detected.
-if config.z3_library != "":
+if config.z3_library not in ("", "Z3_LIBRARIES-NOTFOUND"):
   tools.append(ToolSubst(f"%libz3", config.z3_library))
   config.available_features.add('libz3')
 


mikeurbach (approved):
I haven't tested myself but this makes sense--I've seen this NOTFOUND business in CMake find_package results. I think it might actually be a somewhat standard thing. I haven't tested this yet, but pre-emptively approving based on your testing.

mikeurbach:
I'm going to go ahead and merge this and try it out since I happen to be building on top of tree right now.

Max-astro:
@TaoBi22, thanks for working on this issue. I also encounter a problem with Z3. I specified the Z3 path by `cmake -DZ3_PATH=XXX/z3/build/z3`, and the cmake log like this:
```
CMake Warning at llvm/build/lib/cmake/llvm/FindZ3.cmake:116 (message):
  Failed to determine Z3 library version, defaulting to 0.0.0.
-- Found z3 at XXX/z3/build
-- Could NOT find Z3 (missing: Z3_LIBRARIES Z3_INCLUDE_DIR) (found version "0.0.0")
```
But the `XXX/z3/build/` folder does exist `libz3.dylib` file.
Do you have any clues about this warning?

So, for now, every time I run `check-circuit-integration,` I need to manually modify the `lit.cfg.py` file and set the Z3 library path in it.
Is there any way that I can set the Z3 library path as an environment variable so I don't need to change the .py file manually?

maerhart:
@Max-astro Have you also tried `Z3_DIR` instead of `Z3_PATH`? That's what's checked for in the CIRCT CMake config.

Max-astro:
> @Max-astro Have you also tried `Z3_DIR` instead of `Z3_PATH`? That's what's checked for in the CIRCT CMake config.

I tried your suggestion just now. The cmake warning message disappeared, but the CMAKE said it `Found Z3: Z3_LIB_LOCATION-NOTFOUND (found version "4.13.4.0")`.

And the `check-circuit-integration` still cannot find the correct z3 lib; the log shows that the `circt-bmc` used `--shared-libs=z3::libz3` as the library directly, which doesn't exist.

maerhart:
Actually just noticed that you need `Z3_DIR` in addition to `Z3_PATH` not instead of it. `Z3_PATH` is used to find the binary and `Z3_DIR` is used to find the z3 shared library. I'm not exactly sure what paths you can provide for `Z3_DIR` for CMake to find it, but the path to the directory in which the `libz3.so` or `libz3.dylib` is located should work (not the path to a file or binary directly).

PR #8337, opened 2025-03-21
[Python] Add binding for OutputFileAttr filename.

mikeurbach:
We already have a binding for the OutputFileAttr, but it didn't provide any useful accessor. This adds a safe, typed accessor to read out the filename as a string.


uenoku (approved):
LGTM

PR #8339, opened 2025-03-21
[FIRRTL] Add "special substitutions" to printfs

seldridge:
Implement parsing/emission of a new substitution format for FIRRTL. This allows for special double curly brace quoted strings to be used to mean "special" substitutions. Currently, the only one supported is `SimulationTime`. We plan to add `HierarchicalName`, too.

E.g., this will parse the following differently if the FIRRTL version is 4.2.0:

``` firrtl
FIRRTL version 4.2.0
circuit Foo :
  public module Foo :
    input clock: Clock

    printf(clock, UInt<1>(1), "[{{SimulationTime}}]: hello world")
```

This becomes:

``` mlir
module {
  firrtl.circuit "Foo" {
    firrtl.module @Foo(in %clock: !firrtl.clock) attributes {convention = #firrtl<convention scalarized>} {
      %c1_ui1 = firrtl.constant 1 : !firrtl.const.uint<1>
      firrtl.printf %clock, %c1_ui1, "[{{}}]: hello world" {specialSubstitutions = [#firrtl<printfSubstitution SimulationTime>]}  : !firrtl.clock, !firrtl.const.uint<1>
    }
  }
}
```

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
@@ -13,8 +13,10 @@
 #ifndef CIRCT_DIALECT_FIRRTL_FIRRTLATTRIBUTES_TD
 #define CIRCT_DIALECT_FIRRTL_FIRRTLATTRIBUTES_TD
 
+include "FIRRTLEnums.td"
 include "FIRRTLDialect.td"
 include "mlir/IR/BuiltinAttributeInterfaces.td"
+include "mlir/IR/EnumAttr.td"
 include "circt/Types.td"
 
 //===----------------------------------------------------------------------===//
@@ -186,6 +188,16 @@ def InternalPathAttr : AttrDef<FIRRTLDialect, "InternalPath"> {
 
 def InternalPathArrayAttr
   : TypedArrayAttrBase<InternalPathAttr, "InternalPath array attribute">;
+
+//===----------------------------------------------------------------------===//
+// Printf attributes
+//===----------------------------------------------------------------------===//
+
+def PrintfSubstitutionEnumAttr
+  : EnumAttr<FIRRTLDialect, PrintfSubstitutionEnum, "printfSubstitution">;
+
+def PrintfSubstitutionEnumArrayAttr
+  : TypedArrayAttrBase<PrintfSubstitutionEnumAttr, "Array of special printf substitutuions">;
 
 //===----------------------------------------------------------------------===//
 // Miscellaneous attributes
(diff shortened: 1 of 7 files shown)


sequencer:
cc @FanShupei. We can get rid of our simulation counter based on this pr

seldridge:
Note: I'm trying to go a slightly different route internally with this using https://github.com/llvm/circt/pull/8342 (and running into really weird issues...). The FIRRTL API will stay the same, though.

Once finished, this will lower to `%t` (or `%d` or configurable in some way?) and `$time`.

FanShupei:
> cc FanShupei. We can get rid of our simulation counter based on this pr

The feature looks promising. It will make our code a bit more simpler.

Though I welcome feature, I'm bit confused on the current approach. Since printf already takes a format string, may we just add more format specifier? e.g. Add "%t" to denote current time and "%m" to denote current hierarchical path. 

On Chisel side, we may just use `printf(cf"time = $currentTime")`, where `cureentTime` is a special marker type.

I'm afraid adding another level substitution may bring much complexity. e.g. How to escape if the user just intend to print double braces literally.

seldridge:
> Though I welcome feature, I'm bit confused on the current approach. Since printf already takes a format string, may we just add more format specifier? e.g. Add "%t" to denote current time and "%m" to denote current hierarchical path.

We may wind up doing that. However, changing to that later is easy once this is demonstrated end to end if we want that.

I'm trying to avoid having substitutions be the same as Verilog or nearly the same as Verilog. I don't like using `%t` as it is only a format specifier in Verilog and we want to use it to mean both format specifier _and_ `$time`. I don't like using `%m` as an in-IR representation as it's difficult for compilers to work with, e.g., FPGA synthesis. (This is more about a format like what this PR is doing where the "special" format specifiers are pulled out.) We also likely want to later support more than Verilog. E.g., it's not unreasonable to have a `{{ModuleName}}` different from Verilog's `%m`. We can use more letters, but it's easier to just use `{{SomeWord}}`.

Practically, though, you won't see this! The API you're proposing and we want to support via the `cf` substitutions: `printf(cf"time = $currentTime")` If that becomes `"time = {{CurrentTime}}"` or `"time = %m"`, it does not matter.

> I'm afraid adding another level substitution may bring much complexity. e.g. How to escape if the user just intend to print double braces literally.

Good point. Yes, the syntax needs to be defined. There would need to be an escaping mechanism for `{{`, just as there is for `%`. One benefit of unifying around `{{` is that it avoids needing to escape if you want to use Verilog format specifiers directly, e.g., `%m` can be used directly if desired.

seldridge:
After borrowing some ideas from how the `sim` dialect handles format string representations, I'd rather move in the direction of something like the following patch sequence instead of this PR:

1. https://github.com/llvm/circt/pull/8342
2. https://github.com/llvm/circt/pull/8343

PR #8340, opened 2025-03-21
[AffineParallelOpUnparallelize][Calyx] Affine parallel loop unparallelize to nested for and parallel loop

jiahanxie353:
This pass unparallelize `affine.parallel` op by `factor` (attached as the attribute).

We need this pass to coordinate with the memory banking pass for the Calyx dialect.
Suppose we have a memory `%alloc` banked to two: `%alloc_0` and `%alloc_1`, and suppose the program looks like:
```mlir
affine.parallel (%arg) = (0) to (8) {
  scf.index_switch %arg mod 2:
  case 1 {
    memref.store %cst, %alloc_0[%arg floordiv 2]
  }
  case 0 {
    memref.store %cst, %alloc_1[%arg floordiv 2]
  }
}
```

By running `--lower-scf-to-calyx` pass, the resulting program will have `calyx.par` with 8 arms; but there are only 2 banks. So the general issue is that, when `affine.parallel`'s total number of trip counts is greater than the number of banks, there will be memory access contention when we later lower to Calyx.

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxPasses.h
+++ b/include/circt/Dialect/Calyx/CalyxPasses.h
@@ -31,6 +31,7 @@ std::unique_ptr<mlir::Pass> createResetInsertionPass();
 std::unique_ptr<mlir::Pass> createGroupInvariantCodeMotionPass();
 std::unique_ptr<mlir::Pass> createAffineParallelUnrollPass();
 std::unique_ptr<mlir::Pass> createAffineToSCFPass();
+std::unique_ptr<mlir::Pass> createAffinePloopUnparallelizePass();
 
 /// Generate the code for registering passes.
 #define GEN_PASS_REGISTRATION
--- a/include/circt/Dialect/Calyx/CalyxPasses.td
+++ b/include/circt/Dialect/Calyx/CalyxPasses.td
@@ -283,7 +283,7 @@ def AffinePloopUnparallelize : Pass<"affine-ploop-unparallelize", "::mlir::func:
           }
           default {}
         }
-      }
+      } {unparallelized}
       ```
   }];
 }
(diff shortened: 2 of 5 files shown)


include/circt/Dialect/Calyx/CalyxPasses.h
```diff
@@ -31,6 +31,7 @@ std::unique_ptr<mlir::Pass> createResetInsertionPass();
 std::unique_ptr<mlir::Pass> createGroupInvariantCodeMotionPass();
 std::unique_ptr<mlir::Pass> createAffineParallelUnrollPass();
 std::unique_ptr<mlir::Pass> createAffineToSCFPass();
+std::unique_ptr<mlir::Pass> createAffinePloopUnparallelizePass();
```
cgyurgyik, line 34: Nit: why is there a `P`? aren't we just creating an `AffineFor`?

include/circt/Dialect/Calyx/CalyxPasses.td
```diff
@@ -245,4 +245,47 @@ def AffineToSCF : Pass<"calyx-affine-to-scf", "::mlir::func::FuncOp"> {
...
 }
 
+def AffinePloopUnparallelize : Pass<"affine-ploop-unparallelize", "::mlir::func::FuncOp"> {
+  let summary = "Unparallelize `affine.parallel` op to `affine.for`";
+  let dependentDialects = ["mlir::affine::AffineDialect"];
+  let constructor = "circt::calyx::createAffinePloopUnparallelizePass()";
+  let description = [{
+      Unparallelize `affine.parallel` op to `affine.for` by the factor
+      specified in the attribute.
+      An example:
```
cgyurgyik, line 255:
```suggestion
      specified in the attribute. For example:
```

cgyurgyik (approved):
LGTM, a few smaller comments.

include/circt/Dialect/Calyx/CalyxPasses.h
jiahanxie353: I made `Ploop` short for `ParallelLoop`, so `createAffineParallelLoopUnparallelizePass`

PR #8341, opened 2025-03-21
[SCFToCalyx] If op with sequential condition bug fix

jiahanxie353:
This patch fixes a logic bug that checks if an `scf.if` op's condition needs a register to hold the result of sequential computation

### DIFF
--- a/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
+++ b/include/circt/Dialect/Calyx/CalyxLoweringUtils.h
@@ -405,9 +405,13 @@ public:
   /// Put the name of the callee and the instance of the call into map.
   void addInstance(StringRef calleeName, InstanceOp instanceOp);
 
-  bool isSeqGuardCmpLibOp(Operation *);
+  /// Returns if `op` is a compare operator that requires a register to hold the
+  /// value of its sequential guard computation.
+  bool isSeqGuardCmpLibOp(Operation *op);
 
-  void addSeqGuardCmpLibOp(Operation *);
+  /// Add `op` if it's a compare operator that requires a register to hold the
+  /// value of its sequential guard computation.
+  void addSeqGuardCmpLibOp(Operation *op);
 
   /// Returns the evaluating group or None if not found.
   template <typename TGroupOp = calyx::GroupInterface>
@@ -540,6 +544,8 @@ private:
   /// https://docs.calyxir.org/lang/data-format.html?highlight=json#the-data-format
   llvm::json::Value extMemData;
 
+  /// A set of compare operators that require registers to hold their sequential
+  /// guard computation.
   DenseSet<Operation *> seqGuardCmpLibOps;
 };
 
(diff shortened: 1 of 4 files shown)


lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -473,25 +469,39 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
...
     getState<ComponentLoweringState>().addBlockScheduleable(cmpIOp->getBlock(),
                                                             groupOp);
 
-    buildAssignmentsForRegisterWrite(
-        rewriter, groupOp, getState<ComponentLoweringState>().getComponentOp(),
-        resReg, condReg.getOut());
+    rewriter.setInsertionPointToEnd(groupOp.getBodyBlock());
+    auto loc = cmpIOp.getLoc();
+    // `calyxOp` will be one of `EqLibOp`, `SleLibOp`, etc. The
+    // 2'th `result` corresponds to the output result of `calyxOp`.
```
cgyurgyik, line 485: I'd prefer this to be an `assert`ion rather than a comment, i.e., can we assert this type matches?

include/circt/Dialect/Calyx/CalyxLoweringUtils.h
```diff
@@ -535,6 +539,8 @@ class ComponentLoweringStateInterface {
   /// A json file to store external global memory data. See
   /// https://docs.calyxir.org/lang/data-format.html?highlight=json#the-data-format
   llvm::json::Value extMemData;
+
+  DenseSet<Operation *> seqGuardCmpLibOps;
```
cgyurgyik, line 543: Add a comment please.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
```diff
@@ -473,25 +469,42 @@ class BuildOpGroups : public calyx::FuncOpPartialLoweringPattern {
...
-    buildAssignmentsForRegisterWrite(
-        rewriter, groupOp, getState<ComponentLoweringState>().getComponentOp(),
-        resReg, condReg.getOut());
+    rewriter.setInsertionPointToEnd(groupOp.getBodyBlock());
+    auto loc = cmpIOp.getLoc();
+    assert(
+        (isa<calyx::EqLibOp, calyx::NeqLibOp, calyx::SleLibOp, calyx::SltLibOp,
+             calyx::LeLibOp, calyx::LtLibOp, calyx::GeLibOp, calyx::GtLibOp,
+             calyx::SgeLibOp, calyx::SgtLibOp>(calyxOp.getOperation())) &&
+        "Must be a Calyx comparison library operation.");
```
cgyurgyik, line 488: Ah, sorry I didn't realize we couldn't just do something like `isa<calyx::Comparison>(op)`.

cgyurgyik (approved):
LGTM, thanks.

lib/Conversion/SCFToCalyx/SCFToCalyx.cpp
jiahanxie353: yeah.. no it's fine :)

PR #8342, opened 2025-03-22
[FIRRTL] Add FormatStringType

seldridge:
Add a format string type.  This is intended to be used to improve the way we capture format strings and extend this to capturing more advanced concepts like "time" via attributes as opposed to resorting to enums and attributes.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLTypes.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLTypes.td
@@ -91,11 +91,11 @@ def PassiveType : FIRRTLDialectType<
 
 def SizedType : FIRRTLDialectType<CPred<"type_isa<FIRRTLBaseType>($_self) && "
     "!type_cast<FIRRTLBaseType>($_self).hasUninferredWidth()">,
-    "a sized type (contains no uninferred widths)", 
+    "a sized type (contains no uninferred widths)",
     "::circt::firrtl::FIRRTLBaseType">;
 def SizedOrForeignType : AnyTypeOf<[SizedType, ForeignType]>;
 def SizedPassiveType : FIRRTLDialectType<And<[SizedType.predicate,PassiveType.predicate]>,
-    "a sized passive base type (contains no uninferred widths, or flips)", 
+    "a sized passive base type (contains no uninferred widths, or flips)",
     "::circt::firrtl::FIRRTLBaseType">;
 def SizedPassiveOrForeignType : AnyTypeOf<[SizedPassiveType, ForeignType]>;
 
@@ -235,5 +235,19 @@ def DoubleType : FIRRTLDialectTypeHelper<"DoubleType", "double type">,
 
 def PathType : FIRRTLDialectTypeHelper<"PathType", "path type">,
   BuildableType<"::circt::firrtl::PathType::get($_builder.getContext())">;
+
+//===----------------------------------------------------------------------===//
+// Format String Types
+//===----------------------------------------------------------------------===//
+
+def FStringTimeType :
+  FIRRTLDialectTypeHelper<"FStringTimeType",
+    "a format string type that represents the current simulation time">,
+  BuildableType<"::circt::firrtl::FStringTimeType::get($_builder.getContext())">;
+
+def PrintfOperandType : FIRRTLDialectType<
+  CPred<"type_isa<FIRRTLBaseType, FStringTimeType>($_self)">,
+    "a printf operand type (a FIRRTL base type or a format string type)",
+    "::circt::firrtl::FIRRTLType">;
 
 #endif // CIRCT_DIALECT_FIRRTL_FIRRTLTYPES_TD
--- a/include/circt/Dialect/FIRRTL/FIRRTLTypes.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLTypes.td
@@ -240,13 +240,16 @@ def PathType : FIRRTLDialectTypeHelper<"PathType", "path type">,
 // Format String Types
 //===----------------------------------------------------------------------===//
 
-def FStringTimeType :
-  FIRRTLDialectTypeHelper<"FStringTimeType",
-    "a format string type that represents the current simulation time">,
-  BuildableType<"::circt::firrtl::FStringTimeType::get($_builder.getContext())">;
+def FStringType :
+  FIRRTLDialectTypeHelper<"FStringType",
+    "a format string type">,
+  BuildableType<"::circt::firrtl::FStringType::get($_builder.getContext())">;
 
+// TODO: Migrate off of this by making the operands for `PrintfOp` use only
+// `FStringType`.  This requires conversion ops that convert from
+// `FIRRTLBaseType` to `FStringType`.
 def PrintfOperandType : FIRRTLDialectType<
-  CPred<"type_isa<FIRRTLBaseType, FStringTimeType>($_self)">,
+  CPred<"type_isa<FIRRTLBaseType, FStringType>($_self)">,
     "a printf operand type (a FIRRTL base type or a format string type)",
     "::circt::firrtl::FIRRTLType">;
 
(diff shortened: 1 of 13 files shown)


include/circt/Dialect/FIRRTL/FIRRTLTypes.td
```diff
@@ -236,4 +236,18 @@ def DoubleType : FIRRTLDialectTypeHelper<"DoubleType", "double type">,
...
+def FStringTimeType :
+  FIRRTLDialectTypeHelper<"FStringTimeType",
+    "a format string type that represents the current simulation time">,
+  BuildableType<"::circt::firrtl::FStringTimeType::get($_builder.getContext())">;
+
+def PrintfOperandType : FIRRTLDialectType<
+  CPred<"type_isa<FIRRTLBaseType, FStringTimeType>($_self)">,
+    "a printf operand type (a FIRRTL base type or a format string type)",
+    "::circt::firrtl::FIRRTLType">;
+
```
fabianschuiki, line 252: I'm wondering if we need to distinguish between the exact nature of the format string snippets at all. We might be able to get away with just a generic `FStringType` which all interpolator ops produce. Something like this:

```mlir
%0 = firrtl.fstring.time ... : !firrtl.fstring
%1 = firrtl.fstring.literal ": got wrong value " : !firrtl.fstring
%2 = firrtl.fstring.int hex, width 16, padzero, ... : !firrtl.fstring
%3 = firrtl.fstring.concat %0, %1, %2 : !firrtl.fstring
firrtl.printf %clock, %c1_ui1, %3
```

WDYT?

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
```diff
@@ -578,4 +578,11 @@ def DoubleImpl : PropImplType<"Double"> {
   let typeName = "firrtl.propDouble";
 }
 
+def FStringTimeImpl : FIRRTLImplType<"FStringTime", [], "::circt::firrtl::FIRRTLType" > {
+  let summary = "A format string type that represents the current simulatoin time";
```
fabianschuiki, line 582:
```suggestion
  let summary = "A format string type that represents the current simulation time";
```

lib/Dialect/FIRRTL/Export/FIREmitter.cpp
```diff
@@ -834,8 +836,59 @@ void Emitter::emitStatement(PrintFOp op) {
...
+              });
+          formatString.append("}}");
+        }
+        i += 3;
+        break;
+      default:
+        formatString.push_back(c);
+      }
+    }
+    ps.writeQuotedEscaped(formatString);
```
fabianschuiki, line 889: Thanks for updating the emitter :raised_hands:

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6304,6 +6304,18 @@ LayerBlockOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// Format Sring operations
+//===----------------------------------------------------------------------===//
+
+// FIRRTLType TimeOp::inferReturnType(mlir::MLIRContext *context) {
+//   return FormatStringType::get(context);
+// }
+
```
fabianschuiki, line 6314:
```suggestion
```

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -2936,7 +2941,78 @@ ParseResult FIRStmtParser::parsePrintf() {
...
+          operands.push_back(specOperands[opIdx++]);
+          break;
+        // Let anything we don't know about through.  This allows for wildcat
+        // use of `%m` if a user wants.
+        default:
+          break;
+        }
+        validatedFormatString.push_back(c);
+        break;
+      }
```
fabianschuiki, line 2975: This is really cool! We could spit out dedicated `firrtl.fstring.int ...` ops for these, and then concatenate everything into a single `!firrtl.fstring` that we pass to `firrtl.printf`.

fabianschuiki (approved):
This is really really cool. I love the direction of this. Rust has had [some good results](https://doc.rust-lang.org/std/macro.format_args.html) with this approach of statically parsing format strings into a fixed data structure.

I'm wondering if we shouldn't go all in on this, and remove string interpolation from `printf` entirely and use dedicated formatting ops. That would allow us to define string interpolation once in a bunch of ops, and then reuse it for printf, fprintf, assert, assume, cover, and other stuff I forgot :grin:

include/circt/Dialect/FIRRTL/FIRRTLTypes.td
seldridge: I went back and forth on this. I originally coded it this way and then changed it for reasons that aren't entirely clear. I think I was originally wanting to have some information about the type so that the operand could know what it is without looking at the op. Though, that seems bad.

Let me rework this in that direction. I agree that you eventually just want a single `fstring` type that other ops produce, like concat, and having this blow up with many different per-op types is wrong.

seldridge: Changed to this in: 41e10163f

lib/Dialect/FIRRTL/Export/FIREmitter.cpp
seldridge: I use it and like to see it work. 🚀

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
seldridge: Accidentally fixed via elision in: https://github.com/llvm/circt/pull/8342/commits/3abc9fdcf762c827344667c13a6c258d533db2ca

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: Fixed in e9cd52cf3.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
seldridge: This is _exactly_ the direction I would like to move this in in the future. Thanks for the points to `sim` dialect and Rust's `format_args` that precipitated this push.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -2936,7 +2941,78 @@ ParseResult FIRStmtParser::parsePrintf() {
...
+  if (version < missingSpecFIRVersion) {
+    validatedFormatString = formatString;
+    operands.append(specOperands);
+  } else {
+    for (size_t i = 0, e = formatString.size(), opIdx = 0; i != e; ++i) {
+      auto c = formatString[i];
+      switch (c) {
+      // FIRRTL percent format strings.  If this is actually a format string,
+      // then grab one of the "spec" operands.
+      case '%': {
```
fzi-hielscher, line 2959: IIRC `%c` is not mentioned in the FIRRTL spec, but it has at least one prominent user:
https://github.com/chipsalliance/rocket-chip-blocks/blob/main/src/main/scala/devices/uart/UARTTx.scala#L69
Should that substitution still be allowed?

seldridge: Good catch. We likely need to let this through, too.

seldridge: Added in: https://github.com/llvm/circt/pull/8342/commits/db20bd87988c7d10a37b66fedcd3358f31fefdcb.

seldridge: Adding this to the FIRRTL spec here: https://github.com/chipsalliance/firrtl-spec/pull/287

Note: prior to this PR there was no checking whatsoever of the legality of the the format strings (and this PR still isn't strong enough as it lets through anything it doesn't understand). 😬 To be fair, this is already checked in Chisel, so it's of limited concern for unified Chisel/CIRCT usage.

PR #8343, opened 2025-03-23
[LowerToHW] Lower PrintfOp w/ TimeOp

seldridge:
Add a lowering for PrintfOps which use TimeOps for substitutions.  A TimeOp used by a PrintfOp will be lowered to an sv::TimeOp that is co-located with the Verilog FWriteOp.  (This helps to avoid problems where the system function `$time` could get spilled to a wire.  This is sunk into the same procedural region as the fwrite to avoid this.)

Practically, this is the last piece needed to compile FIRRTL that includes the `{{SimulationTime}}` special substitution all the way to Verilog.

This is stacked on #8342.

#### Example

The following FIRRTL includes the lone supported special substitution `{{SimulationTime}}`:

``` firrtl
FIRRTL version 4.2.0
circuit Foo :
  public module Foo :
    input clock: Clock
    input a: UInt<1>

    printf(clock, UInt<1>(1), "[{{SimulationTime}}]: %d", a)
```

Compiled to Verilog using `firtool Foo.fir`, this becomes:

``` verilog
// Generated by CIRCT firtool-1.109.0-23-ge68c0327e-dirty

// Users can define 'PRINTF_FD' to add a specified fd to prints.
`ifndef PRINTF_FD_
  `ifdef PRINTF_FD
    `define PRINTF_FD_ (`PRINTF_FD)
  `else  // PRINTF_FD
    `define PRINTF_FD_ 32'h80000002
  `endif // PRINTF_FD
`endif // not def PRINTF_FD_

`ifndef PRINTF_COND_
  `ifdef PRINTF_COND
  `else  // PRINTF_COND
    `define PRINTF_COND_ 1
  `endif // PRINTF_COND
module Foo(
  input clock,
        a
);

  `ifndef SYNTHESIS
    always @(posedge clock) begin
      if ((`PRINTF_COND_) & 1'h1)
        $fwrite(`PRINTF_FD_, "[%t]: %d", $time, a);
    end // always @(posedge)
  `endif // not def SYNTHESIS
endmodule
```

For completeness, this is parsed into the following FIRRTL Dialect:

``` mlir
module {
  firrtl.circuit "Foo" {
    firrtl.module @Foo(in %clock: !firrtl.clock, in %a: !firrtl.uint<1>) attributes {convention = #firrtl<convention scalarized>} {
      %c1_ui1 = firrtl.constant 1 : !firrtl.const.uint<1>
      %time = firrtl.fstring.time : !firrtl.fstring.time
      firrtl.printf %clock, %c1_ui1, "[{{}}]: %d" (%time, %a) : !firrtl.clock, !firrtl.const.uint<1>, !firrtl.fstring.time, !firrtl.uint<1>
    }
  }
}
```

And is lowered to the following by `LowerToHW`:

``` mlir
module {
  sv.macro.decl @SYNTHESIS
  sv.macro.decl @PRINTF_FD
  sv.macro.decl @PRINTF_FD_
  emit.fragment @PRINTF_FD_FRAGMENT {
    sv.verbatim "\0A// Users can define 'PRINTF_FD' to add a specified fd to prints."
    sv.ifdef  @PRINTF_FD_ {
    } else {
      sv.ifdef  @PRINTF_FD {
        sv.macro.def @PRINTF_FD_ "(`PRINTF_FD)"
      } else {
        sv.macro.def @PRINTF_FD_ "32'h80000002"
      }
    }
  }
  sv.macro.decl @PRINTF_COND
  sv.macro.decl @PRINTF_COND_
  emit.fragment @PRINTF_COND_FRAGMENT {
    sv.verbatim "\0A// Users can define 'PRINTF_COND' to add an extra gate to prints."
    sv.ifdef  @PRINTF_COND_ {
    } else {
      sv.ifdef  @PRINTF_COND {
        sv.macro.def @PRINTF_COND_ "(`PRINTF_COND)"
      } else {
        sv.macro.def @PRINTF_COND_ "1"
      }
    }
  }
  hw.module @Foo(in %clock : !seq.clock, in %a : i1) attributes {emit.fragments = [@PRINTF_FD_FRAGMENT, @PRINTF_COND_FRAGMENT]} {
    %true = hw.constant true
    %0 = seq.from_clock %clock
    sv.ifdef  @SYNTHESIS {
    } else {
      sv.always posedge %0 {
        %PRINTF_COND_ = sv.macro.ref.expr @PRINTF_COND_() : () -> i1
        %1 = comb.and bin %PRINTF_COND_, %true : i1
        sv.if %1 {
          %PRINTF_FD_ = sv.macro.ref.expr @PRINTF_FD_() : () -> i32
          %2 = sv.system.time : i64
          sv.fwrite %PRINTF_FD_, "[%t]: %d"(%2, %a) : i64, i1
        }
      }
    }
    hw.output
  }
  om.class @Foo_Class(%basepath: !om.basepath) {
    om.class.fields
  }
}
```

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -1726,6 +1726,9 @@ struct FIRRTLLowering : public FIRRTLVisitor<FIRRTLLowering, LogicalResult> {
   LogicalResult visitExpr(XMRRefOp op);
   LogicalResult visitExpr(XMRDerefOp op);
 
+  // Format String Operations
+  LogicalResult visitExpr(TimeOp op);
+
   // Statements
   LogicalResult lowerVerificationStatement(
       Operation *op, StringRef labelPrefix, Value clock, Value predicate,
@@ -2399,6 +2402,12 @@ Value FIRRTLLowering::getLoweredAndExtOrTruncValue(Value value, Type destType) {
 /// format strings. Zero bit operands are rewritten as one bit zeros and signed
 /// integers are wrapped in $signed().
 Value FIRRTLLowering::getLoweredFmtOperand(Value operand) {
+  // Handle special substitutions.
+  if (type_isa<FStringType>(operand.getType())) {
+    if (isa<TimeOp>(operand.getDefiningOp()))
+      return builder.create<sv::TimeOp>();
+  }
+
   auto loweredValue = getLoweredValue(operand);
   if (!loweredValue) {
     // If this is a zero bit operand, just pass a one bit zero.
@@ -4279,6 +4288,11 @@ LogicalResult FIRRTLLowering::visitExpr(XMRDerefOp op) {
   return setLoweringTo<seq::ToClockOp>(op, readXmr);
 }
 
+// Do nothing when lowering this operation.  We need to handle this at its usage
+// sites to guarantee creation of `sv::TimeOp` that is colocated with the
+// `sv::FWriteOp`.
+LogicalResult FIRRTLLowering::visitExpr(TimeOp op) { return success(); }
+
 //===----------------------------------------------------------------------===//
 // Statements
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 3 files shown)


lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -4279,6 +4288,11 @@ LogicalResult FIRRTLLowering::visitExpr(XMRDerefOp op) {
   return setLoweringTo<seq::ToClockOp>(op, readXmr);
 }
 
+// Do nothing when lowering this operation.  We need to handle this at its usage
+// sites to guarantee creation of `sv::TimeOp` that is colocated with the
+// `sv::FWriteOp`.
+LogicalResult FIRRTLLowering::visitExpr(TimeOp op) { return success(); }
+
```
fabianschuiki, line 4295: I'm wondering if this is strictly necessary :thinking:. Since `$time` should be essentially an integer value in Verilog, spilling might just work. I'm also totally fine with treating this as a special case here.

fabianschuiki (approved):
LGTM! Really neat :+1:

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
seldridge: What I was worried about was if this somehow got CSE'd very far away from the `printf`. As long as it is in the same `always` block as the `printf`, it should be completely fine, even if spilled. I was worried about it getting outside that `always` block to produce something like:

```
wire _GEN = $time;

always @posedge(clock)
  $dipslay("%t", _GEN);

always @posedge(clock)
  $dipslay("%t", _GEN);
```

Same issue for the more complicate multi-clock case, or similar.

PR #8344, opened 2025-03-24
[DOC] Add missing passes

fhunstock-semron:
The documentation for some of the "general" passes was missing.

AFAICT @mortbopet and @mikeurbach accidentally added it as `add_mlir_doc` instead of `add_circt_doc` in https://github.com/llvm/circt/commit/fe32c20d965dd2e61420a5a6d309015f9be6b0e6
everything else was already switched over in https://github.com/llvm/circt/commit/77f9520ea156b093e2de0a4ad73cbb6592484cb5

The same applies for the BMC and LEC passes.

(the changed arguments come from https://github.com/llvm/circt/commit/68280a14263239ad6a2001d33310decb1fc2f698)

### DIFF
--- a/docs/Passes.md
+++ b/docs/Passes.md
@@ -4,6 +4,10 @@ This document describes the available CIRCT passes and their contracts.
 
 [TOC]
 
+## General Passes
+
+[include "CIRCTGeneralPasses.md"]
+
 ## Conversion Passes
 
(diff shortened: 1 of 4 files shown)


seldridge (approved):
This looks right, thanks!

mikeurbach (approved):
Thank you, makes sense to me.

Issue #7773, opened 2024-11-05
[FIRRTL] Parser Accepts Duplicated Names Under Layerblock


seldridge:
The following should be illegal, but is currently parsed:

``` firrtl
FIRRTL version 4.0.0
circuit Foo:
  layer A, bind:
  public module Foo:

    layerblock A:
      wire a: UInt<1>

    wire a: UInt<1>
```

If the `layerblock` is moved _after_ the second wire definition, then the correct error is produced.

youngar:
Related to https://github.com/llvm/circt/issues/1537

sahil07-rg:
Hey!
The reason could be that the "wire a: UInt<1>" is defined twice.
OR we can check the FIRRTL grammar.

Issue #1537, opened 2021-08-10
[FIRRTL] SFC requires a unique name for everything


youngar:
SFC doesn't just disallow name shadowing, it doesn't allow any two things to have the same name:
```
circuit foo :
  module foo :
    input p: UInt<1>

    when p:
      wire x: UInt<1>
    when p:
      wire x: UInt<1>
 ```
 
 SFC produces an error for this:
 ```
 firrtl-1.5-SNAPSHOT -i ./test-names.fir -o test-names
Exception in thread "main" firrtl.passes.CheckHighFormLike$NotUniqueException: : [module foo] Reference x does not have a unique name
```

MFC has no problem with this code.

mmaloney-sf:
@youngar pointed me to this.

SFC is acting in accordance with the spec. MFC is actually the one not complying here:

> As mention in Section 14, circuit component declarations in a module must be unique within the module’s flat namespace; this means that shadowing a component in an enclosing scope with a component of the same name inside a conditional statement is not allowed.

(FIRRTL Spec version `v2.3.0`)

Ignore the commentary after the semicolon. The first part of the snippet clearly states that the module's namespace is flat. Nested scopes in FIRRTL only affect where references to circuit components are permitted. This is notably different from how scopes work in most programming languages.

This issue has been open for over a year, so it could be that the spec actually changed during this time. I tried to look back, but the spec repo was reorganized back in March of 2022 and I didn't try to look back beyond the history of the `spec.md` file.

PR #8345, opened 2025-03-24
[FIRRTL] Fix: Enforce flat namespace for modules

Jatin-exe:
Fix for #7773 and  #1537.
- Added a check for duplicates in the symbolTable. Added and removed tests for the same. 
- Made changes in parseContract to avoid re-declaration error 
  -  I'm deleting the symbol for contract body before adding the results to the symbolTable (I'm guessing it's no longer used, since its marked as invalid)
  - or is there a better way to do this, by updating the moduleContext with the results before its being given to ContextScope, so we can avoid calling addSymbolEntry twice ? (not sure about this)
 

Let me know if this works, or if this breaks something I don't understand.
Feedback is appreciated.  @fabianschuiki

### DIFF
--- a/lib/Dialect/FIRRTL/Import/FIRParser.cpp
+++ b/lib/Dialect/FIRRTL/Import/FIRParser.cpp
@@ -1429,6 +1429,9 @@ struct FIRModuleContext : public FIRParser {
                           insertNameIntoGlobalScope);
   }
 
+  // Removes a symbol from symbolTable (Workaround since symbolTable is private)
+  void removeSymbolEntry(StringRef name);
+
   /// Resolved a symbol table entry to a value.  Emission of error is optional.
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -4801,10 +4818,12 @@ ParseResult FIRStmtParser::parseContract(unsigned blockIndent) {
 
   // Declare the results.
   for (auto [id, loc, value, result] :
-       llvm::zip(ids, locs, values, contract.getResults()))
+       llvm::zip(ids, locs, values, contract.getResults())) {
+    // Remove previous symbol to avoid duplicates
+    moduleContext.removeSymbolEntry(id);
```
fabianschuiki, line 4823: LGTM!

darthscsi (approved):

PR #8347, opened 2025-03-25
[circt-reduce] Add bytecode import/export

seldridge:
Add the ability to read and write bytecode for `circt-reduce`.  This is intended to help with very large MLIR files to avoid wasting all of `circt-reduce`'s time on printing/parsing.

### DIFF
--- a/lib/Reduce/Tester.cpp
+++ b/lib/Reduce/Tester.cpp
@@ -11,6 +11,8 @@
 //===----------------------------------------------------------------------===//
 
 #include "circt/Reduce/Tester.h"
+#include "circt/Support/Version.h"
+#include "mlir/Bytecode/BytecodeWriter.h"
 #include "mlir/IR/Verifier.h"
 #include "llvm/Support/ToolOutputFile.h"
 
@@ -23,9 +25,9 @@ using namespace circt;
 //===----------------------------------------------------------------------===//
 
 Tester::Tester(StringRef testScript, ArrayRef<std::string> testScriptArgs,
-               bool testMustFail)
+               bool testMustFail, bool emitBytecode)
     : testScript(testScript), testScriptArgs(testScriptArgs),
-      testMustFail(testMustFail) {}
+      testMustFail(testMustFail), emitBytecode(emitBytecode) {}
 
 std::pair<bool, size_t> Tester::isInteresting(ModuleOp module) const {
   auto test = get(module);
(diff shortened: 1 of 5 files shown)


tools/circt-reduce/circt-reduce.cpp
```diff
@@ -142,7 +148,18 @@ static LogicalResult writeOutput(ModuleOp module) {
...
+    if (failed(writeBytecodeToFile(
+            module, output->os(),
+            mlir::BytecodeWriterConfig(getCirctVersion())))) {
+      mlir::emitError(UnknownLoc::get(module.getContext()),
+                      "failed to emit bytecode to file \"")
+          << outputFilename << "\n";
+      return failure();
+    }
+  } else
+    module.print(output->os());
```
maerhart, line 161: IIRC the coding guidelines want consistent use of braces, i.e., if the then-block has them the else-block should also have them.

lib/Reduce/Tester.cpp
```diff
@@ -131,12 +133,22 @@ void TestCase::ensureFileOnDisk() {
...
-                                   filepath + "`",
-                               false);
+    if (tester.emitBytecode) {
+      if (failed(writeBytecodeToFile(
+              module, file->os(),
+              mlir::BytecodeWriterConfig(getCirctVersion()))))
+        llvm::report_fatal_error(
+            llvm::Twine("Error emitting the IR to file `") + filepath + "`",
+            false);
+      file->os().close();
```
maerhart, line 143: Shouldn't this 'close' be before the abort because there it will not be closed by the destructor.

lib/Reduce/Tester.cpp
```diff
@@ -131,12 +133,22 @@ void TestCase::ensureFileOnDisk() {
...
+    } else {
+      module.print(file->os());
+      file->os().close();
+      if (file->os().has_error())
+        llvm::report_fatal_error(
+            llvm::Twine("Error emitting the IR to file `") + filepath + "`",
+            false);
```
maerhart, line 150: Isn't that check relevant to both branches of the if-stmt?

maerhart (approved):
LGTM!

tools/circt-reduce/circt-reduce.cpp
```diff
@@ -142,7 +148,18 @@ static LogicalResult writeOutput(ModuleOp module) {
...
-  module.print(output->os());
+  if (emitBytecode) {
+    if (failed(writeBytecodeToFile(
+            module, output->os(),
+            mlir::BytecodeWriterConfig(getCirctVersion())))) {
+      mlir::emitError(UnknownLoc::get(module.getContext()),
+                      "failed to emit bytecode to file \"")
+          << outputFilename << "\n";
+      return failure();
+    }
```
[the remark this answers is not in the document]
seldridge: I think this still needs the `return failure()` as `emitError` isn't enough to get it to terminate.

tools/circt-reduce/circt-reduce.cpp
seldridge: 👍 This will be fixed in commit 32cc70e41.

lib/Reduce/Tester.cpp
seldridge: Yeah, I think you're right. This will be fixed in commit: d9d3b095d

lib/Reduce/Tester.cpp
seldridge: Good catch. This will be fixed in commit: d9d3b095d

Issue #8348, opened 2025-03-26
[FIRRTL] Regreset canonicalizer replaces register with value of different type


rwy7:
If a regreset has a constantly-high reset, then we can replace the register with its reset value. However, if the reset value's type does
not match the type of the register, then we end up replacing the register with a value of a different type. When the register is driven by a matching connect, this is invalid IR.

Here is an example:
    
```mlir
firrtl.circuit "Scratch" {
  firrtl.module @Scratch(out %out : !firrtl.uint<2>) {

    // The reset value is 1-bit, but the register is 2-bits.
    %c0_ui1 = firrtl.constant 0 : !firrtl.uint<1>
    %c0_ui2 = firrtl.constant 0 : !firrtl.uint<2>
    %c0_clock = firrtl.specialconstant 0 : !firrtl.clock
    %c1_asyncreset = firrtl.specialconstant 1 : !firrtl.asyncreset
    %reg = firrtl.regreset %c0_clock, %c1_asyncreset, %c0_ui1 : !firrtl.clock, !firrtl.asyncreset, !firrtl.uint<1>, !firrtl.uint<2>

    // The canonicalizer ends up driving the 1-bit reset to the 2-bit output port using a matching connect.
    firrtl.matchingconnect %out, %reg : !firrtl.uint<2>
    firrtl.matchingconnect %reg, %c0_ui2 : !firrtl.uint<2>
  }
}
```

After canonicalization:
```mlir
// -----// IR Dump After Canonicalizer Failed (canonicalize) //----- //
"firrtl.module"() <{annotations = [], convention = #firrtl<convention internal>, layers = [], portAnnotations = [], portDirections = array<i1: true>, portLocations = [loc("./scratch2.mlir":2:30)], portNames = ["out"], portSymbols = [], portTypes = [!firrtl.uint<2>], sym_name = "Scratch"}> ({
^bb0(%arg0: !firrtl.uint<2>):
  %0 = "firrtl.constant"() <{value = 0 : ui1}> : () -> !firrtl.uint<1>
  "firrtl.matchingconnect"(%arg0, %0) : (!firrtl.uint<2>, !firrtl.uint<1>) -> ()
}) : () -> ()
```

Run using:
```
circt-opt --pass-pipeline="builtin.module(firrtl.circuit(firrtl.module(canonicalize{max-iterations=10 max-num-rewrites=-1 region-simplify=disabled test-convergence=false top-down=true})))" ./scratch2.mlir --mlir-print-ir-after-failure
```

PR #8349, opened 2025-03-26
[FIRRTL] Fold regreset to its reset value iff the type matches

rwy7:
If the regreset has a constantly-high reset, then we can replace the register with its reset value. However, if the reset value's type does not match the type of the register, we cannot do the replacement.

This commit adds a check to an old canonicalizer, to ensure the reset value's type matches, before doing the replacement.

Fixes: #8348

### DIFF
--- a/test/Dialect/FIRRTL/canonicalization.mlir
+++ b/test/Dialect/FIRRTL/canonicalization.mlir
@@ -2293,6 +2293,20 @@ firrtl.module @ForceableRegResetToNode(in %clock: !firrtl.clock, in %dummy : !fi
   firrtl.connect %foo, %reg: !firrtl.uint<1>, !firrtl.uint<1>
 }
 
+// https://github.com/llvm/circt/issues/8348
+// CHECK-LABEL: firrtl.module @RegResetInvalidResetValueType
+// We cannot replace a regreset with its reset value, when the reset value's type does not match.
+firrtl.module @RegResetInvalidResetValueType(in %c : !firrtl.clock, out %out : !firrtl.uint<2>) {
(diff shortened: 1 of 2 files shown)


seldridge (approved):
LGTM!

test/Dialect/FIRRTL/canonicalization.mlir
```diff
@@ -2293,6 +2293,21 @@ firrtl.module @ForceableRegResetToNode(in %clock: !firrtl.clock, in %dummy : !fi
   firrtl.connect %foo, %reg: !firrtl.uint<1>, !firrtl.uint<1>
 }
 
+  // https://github.com/llvm/circt/issues/8348
+  // CHECK-LABEL: firrtl.module @RegResetInvalidResetValueType
+  // We cannot replace a regreset with its reset value, when the reset value's type does not match.
+  firrtl.module @RegResetInvalidResetValueType(out %out : !firrtl.uint<2>) {
+    %c0_ui1 = firrtl.constant 0 : !firrtl.uint<1>
+    %c0_ui2 = firrtl.constant 0 : !firrtl.uint<2>
+    %c0_clock = firrtl.specialconstant 0 : !firrtl.clock
```
dtzSiFive, line 2302: Maaaaybe feed clock from a block argument?

I don't know but "a register that we know is never clocked" -> (??) might trigger a different canonicalizer that's not the one intended to be tested here.

dtzSiFive (approved):
LGTM, thanks!

test/Dialect/FIRRTL/canonicalization.mlir
rwy7: Done, thanks :+1:

PR #8354, opened 2025-03-27
[LowerToHW] Add '0' to format string width

seldridge:
Add a zero width to all format string substitutions when lowering from FIRRTL to HW/SV.  E.g., this will lower `%x` into `%0x`.  This is almost always preferred and avoids the need for (right now) adding width specifiers to Chisel and FIRRTL.  While we would like to do this, we don't want to do it as passthrough from Chisel where Chisel users write Verilog format substitutions.

As @jackkoenig and I got to bottom of, there are non-standard format specifiers that people might want (e.g., `%-` for left justified) that tools support differently and we would like to figure out how to actually lower these if they matter.

### DIFF
--- a/test/firtool/print.fir
+++ b/test/firtool/print.fir
@@ -7,7 +7,7 @@ circuit PrintTest:
     input clock : Clock
     input cond : UInt<1>
     input var : UInt<32>
-    printf(clock, cond, "test %d\n", var)
+    printf(clock, cond, "test %b %c %d %x\n", var, var, var, var)
 
     ; CHECK:      sv.ifdef  @SYNTHESIS {
     ; CHECK-NEXT: } else {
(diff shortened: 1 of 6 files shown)


rwy7 (approved):

test/firtool/print.fir
```diff
@@ -16,7 +16,7 @@ circuit PrintTest:
     ; CHECK-NEXT:     [[COND:%.+]] = comb.and bin [[PRINTF_COND]], %cond : i1
     ; CHECK-NEXT:     sv.if [[COND]] {
     ; CHECK-NEXT:       [[PRINTF_FD:%.+]] = sv.macro.ref.expr @PRINTF_FD_() : () -> i32
-    ; CHECK-NEXT:       sv.fwrite [[PRINTF_FD]], "test %d\0A"(%var) : i32
+    ; CHECK-NEXT:       sv.fwrite [[PRINTF_FD]], "test %0b %0c %0d %0x\0A"(%var, %var, %var, %var) : i32
```
jackkoenig, line 19: As far as I can tell, tools accept it, but IMO `%0c` is weird since `%c` is by definition a single character. I'm not sure if any lint tools would complain.

seldridge: I can't find evidence in the spec that it's allowed, so let's not do it. Good thinking. Changed to not do this for `%c` in: https://github.com/llvm/circt/pull/8354/commits/c667a3ba6c3d3cfa614f4c5e18e01a9f45992c63

PR #8360, opened 2025-03-31
[RTG] Add array create and extract operations

maerhart:


lib/Dialect/RTG/IR/RTGOps.cpp
```diff
@@ -552,6 +552,45 @@ LogicalResult MemoryBlockDeclareOp::inferReturnTypes(
...
+// ArrayCreateOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult ArrayCreateOp::verify() {
+  if (!getElements().empty() &&
+      getElements()[0].getType() != getType().getElementType())
+    return emitOpError(
+               "operand types must match array eleement type, expected ")
+           << getType().getElementType() << " but got "
+           << getElements()[0].getType();
```
rwy7, line 565: I think you need to check that all the inputs are instances of the element type.

lib/Dialect/RTG/IR/RTGOps.cpp
```diff
@@ -552,6 +552,45 @@ LogicalResult MemoryBlockDeclareOp::inferReturnTypes(
...
 
+//===----------------------------------------------------------------------===//
+// ArrayCreateOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult ArrayCreateOp::verify() {
+  if (!getElements().empty() &&
+      getElements()[0].getType() != getType().getElementType())
+    return emitOpError(
+               "operand types must match array eleement type, expected ")
```
rwy7, line 563: typo:
```suggestion
               "operand types must match array element type, expected ")
```

include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -455,6 +455,45 @@ def BagUniqueSizeOp : RTGOp<"bag_unique_size", [Pure]> {
...
+  }];
+
+  let arguments = (ins Variadic<AnyType>:$elements);
+  let results = (outs ArrayType:$result);
+
+  let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
+}
+
+def ArrayExtractOp : RTGOp<"array_extract", [
```
rwy7, line 477: If you want to align on firrtl terminology, the corresponding operation in firrtl is called "subaccess".

rwy7 (approved):

rwy7 (approved):

include/circt/Dialect/RTG/IR/RTGOps.td
maerhart: I aligned it to the HW aggregates, but called it extract like `hw.struct_extract` and `comb.extract` because I think `hw.array_get` is the inconsistent one there.

PR #8370, opened 2025-03-31
[RTG] Add tuple operations

maerhart:


lib/Dialect/RTG/IR/RTGOps.cpp
```diff
@@ -339,6 +339,52 @@ LogicalResult BagCreateOp::verify() {
...
+// TupleCreateOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult TupleCreateOp::inferReturnTypes(
+    MLIRContext *context, std::optional<Location> loc, ValueRange operands,
+    DictionaryAttr attributes, OpaqueProperties properties, RegionRange regions,
+    SmallVectorImpl<Type> &inferredReturnTypes) {
+  if (operands.empty()) {
+    if (loc)
+      return mlir::emitError(*loc) << "empty tuples not allowed";
```
rwy7, line 352: What's the reason for disallowing empty tuples?

youngar (approved):

lib/Dialect/RTG/IR/RTGOps.cpp
maerhart: The only reason is that the builtin tuple type does not support an empty list of element types. I could add a new RTG tuple type and allow this. My thought was that I can generalize this when we need it, but I'm also happy to add it right away if you prefer.

PR #8372, opened 2025-03-31
[LLHD] Make probes side-effect-free in graph regions

fabianschuiki:
Make `llhd.prb` ops only have read side effects if used in an SSACFG region. This works around an issue where CSE would not move or eliminate probes across drives and other processes, even though that would be legal in a graph region. In graph regions, a probe acts like a pure op dereferencing a signal and the exact order of operations, especially the order with respect to drives and other ops interacting with the signal do not matter.


maerhart (approved):
Bit unfortunate that we have to solve it this way. The probe actually still has a side-effect, it's just that the ordering of probes and drives in a graph region doesn't matter. For example, if we had a graph region nested in an SSACFG region containing a probe and the SSACFG region has a drive, then it might not be legal to reorder them. However, this is not something we expect to see in practice, so I think this is a fine workaround.

fabianschuiki:
Ugh yeah that's an excellent point :thinking: One option would be to see if it would make sense to ignore op order in graph regions upstream for CSE and friends. Or we could add our own custom CSE pass for probes, which is annoying but would also work. Maybe worth reconsidering if we ever see graph regions nested in SSACFG regions :thinking:

PR #8384, opened 2025-04-02
[FIRRTL] Add a BindOp

rwy7:
The PR adds a firrtl.bind op. This new bind op is lowered directly to an sv.bind op in LowerToHW. Today, we mark firrtl.instance ops as "lowerToBind", which cues LowerToHW to create an SV BindOp in the top-level scope.  The reason for a bind op in the FIRRTL dialect, is so that FIRRTL passes can choose the placement of the bind op, such as under an sv.ifdef or emit.file op (which is required for the lowering of bound-in layerblocks). The sv.bind op can only target hw.modules ops, while this new firrtl.bind op targets firrtl.module ops. There are no plans to add bind statements to the FIRRTL surface language. This is just an intermediate lowering for layers.


lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6394,6 +6394,39 @@ void HierarchicalModuleNameOp::getAsmResultNames(
...
+  if (!module)
+    return emitError("Referenced module doesn't exist ")
+           << getInstance().getModule() << "::" << getInstance().getName();
+  auto inst =
+      findInnerSym<InstanceOp>(getInstance().getName(), module.getBodyBlock());
+  if (!inst)
+    return emitError("Referenced instance doesn't exist ")
+           << getInstance().getModule() << "::" << getInstance().getName();
+  if (!inst.getDoNotPrint())
+    return emitError("Referenced instance isn't marked as doNotPrint");
```
seldridge, line 6426: Fantastic.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6394,6 +6394,39 @@ void HierarchicalModuleNameOp::getAsmResultNames(
...
+namespace {
+template <class Op>
+Op findInnerSym(StringAttr name, Block *body) {
+  for (auto &op : llvm::reverse(body->getOperations()))
+    if (auto target = dyn_cast<Op>(op))
+      if (auto innerSym = target.getInnerSym())
+        if (innerSym->getSymName() == name)
+          return target;
+  return {};
+}
```
seldridge, line 6410: This is _very_ expensive.

Can this be changed to use `verifyInnerRefs` so that this is done efficiently?

seldridge (commented):
I'm good with the approach. I have a question about efficient verification.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: Thanks, I didn't know about this trait/verification method. The [sv.bind op](https://github.com/llvm/circt/blob/main/include/circt/Dialect/SV/SVStatements.td#L576) is not using verifyInnerRefs either. I'm not sure if this is intentional or an oversight.

seldridge: The SV bind op is ancient and doesn't represent modern conventions. The fact that it doesn't implement this may stem from the problem of it being rooted in a `builtin.module` which restricts our ability to do verification.

include/circt/Dialect/FIRRTL/FIRRTLStatements.td
```diff
@@ -428,4 +429,26 @@ def LayerBlockOp : FIRRTLOp<
...
+
+def BindOp
+    : FIRRTLOp<"bind", [DeclareOpInterfaceMethods<InnerRefUserOpInterface>]> {
+  let summary = "Indirect instantiation statement";
+  let description = [{
+    Indirectly instantiate a module from the context of another module. BindOp
+    pairs with a `firrtl.instance` which tracks all information except the
+    emission point for the bind.
+
+    See `sv.bind` for the system verilog equivalent.
```
seldridge, line 444: Could the PR include a note that this is only used to aid in lowering and for allow for tracking of where a SystemVerilog bind op will be located (in what file). Essentially, include some of the great PR text here about the motivation for this.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6404,6 +6404,29 @@ void HierarchicalModuleNameOp::getAsmResultNames(
...
+
+  auto instance = dyn_cast<InstanceOp>(target.getOp());
+  if (!instance)
+    return emitError() << "target " << ref << " is not an instance";
+
+  if (!instance.getDoNotPrint())
+    return emitError() << "target " << ref << " is not marked doNotPrint";
+
+  return success();
+}
```
seldridge, line 6428: This looks great.

seldridge (approved):
LGTM

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: Yep that seems to be the case, the firrtl CircuitOp is an InnerRefNamespace and the MLIR ModuleOp cannot be.

PR #8387, opened 2025-04-03
[LLHD] Update LowerProcesses pass to work with process results

fabianschuiki:
Add an update version of the process lowering pass. This new version expects all probes and drives to have been hoisted out of processes. The pass then converts an `llhd.process` op to an `scf.execute_region` if the process

- has a single wait op;
- has a simple control flow loop through that single wait op;
- has no destination operands on the wait op; and
- has all operands that are used in the process but defined outside of the process observed by the wait op.

This essentially lowers any combinational LLHD processes with a CFG loop through a wait op which triggers when any of the inputs change into a simple SCF execute region without the CFG loop. The assumption here is that the execute region will be re-evaluated if any of the values used within it change.

Life would be a lot easier if we didn't have to chase through the observed values of the wait op for this lowering. Ideally LLHD would have a simplified version of a process that we can canonicalize to early on.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
+++ b/lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
@@ -1,0 +1,280 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/LLHD/IR/LLHDOps.h"
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h"
+#include "mlir/Analysis/Liveness.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Transforms/RegionUtils.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "llhd-lower-processes"
+
+namespace circt {
+namespace llhd {
+#define GEN_PASS_DEF_LOWERPROCESSESPASS
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h.inc"
+} // namespace llhd
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+using namespace llhd;
+using llvm::SmallDenseSet;
+using llvm::SmallSetVector;
+
+namespace {
+struct Lowering {
+  Lowering(ProcessOp processOp) : processOp(processOp) {}
+  void lower();
+  bool matchControlFlow();
+  void markObservedValues();
+  bool allOperandsObserved();
+  bool isObserved(Value value);
+
+  ProcessOp processOp;
+  WaitOp waitOp;
+  SmallDenseSet<Value> observedValues;
+};
+} // namespace
+
+void Lowering::lower() {
+  // Ensure that the process describes combinational logic.
+  if (!matchControlFlow())
+    return;
+  markObservedValues();
+  if (!allOperandsObserved())
+    return;
+  LLVM_DEBUG(llvm::dbgs() << "Lowering process " << processOp.getLoc() << "\n");
+
+  // Replace the process.
+  OpBuilder builder(processOp);
+  auto executeOp = builder.create<scf::ExecuteRegionOp>(
+      processOp.getLoc(), processOp.getResultTypes());
+  executeOp.getRegion().takeBody(processOp.getBody());
+  processOp.replaceAllUsesWith(executeOp);
+  processOp.erase();
+  processOp = {};
+
+  // Replace the `llhd.wait` with an `scf.yield`.
+  builder.setInsertionPoint(waitOp);
+  builder.create<scf::YieldOp>(waitOp.getLoc(), waitOp.getYieldOperands());
+  waitOp.erase();
+
+  // Simplify the execute op body region since disconnecting the control flow
+  // loop through the wait op has potentially created unreachable blocks.
+  IRRewriter rewriter(builder);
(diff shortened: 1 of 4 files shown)


lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+/// Ensure that any value defined outside the process that is used inside the
+/// process is derived entirely from an observed value.
+bool Lowering::allOperandsObserved() {
+  // Collect all ancestor regions such that we can easily check if a value is
+  // defined outside the process.
+  SmallPtrSet<Region *, 4> properAncestors;
+  for (auto *region = processOp->getParentRegion(); region;
+       region = region->getParentRegion())
+    properAncestors.insert(region);
```
maerhart, line 192: Is this just a performance optimization compared to using `isProperAncestor`?

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
```diff
@@ -0,0 +1,279 @@
...
+      continue;
+
+    // Look through probe ops to mark the probe signal as well, just in case
+    // there may be multiple probes of the same signal.
+    if (auto probeOp = dyn_cast<PrbOp>(op))
+      markObserved(probeOp.getSignal());
+
+    // Look through operations that simply reshape incoming values into an
+    // aggregate form from which any changes remain apparent.
+    if (isa<hw::ArrayCreateOp, hw::StructCreateOp, comb::ConcatOp>(op))
```
maerhart, line 178: `hw.bitcast` is another one I have seen a few times in the past. Mostly with a constant as input, however, so that case should already be covered I think.

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+  // Otherwise get the operation that defines it such that we can check if the
+  // value is derived from purely observed values. If it isn't define by an op,
+  // the value is unobserved.
+  auto *defOp = value.getDefiningOp();
+  if (!defOp)
+    return false;
+
+  // Otherwise visit all ops in the fan-in cone and ensure that they are
+  // observed. If anything value is unobserved, immediately return false.
```
maerhart, line 234:
```suggestion
  // observed. If any value is unobserved, immediately return false.
```

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+    // Otherwise check that all operands are observed. If we haven't seen an
+    // operand before, and it is not a signal, add it to the worklist to be
+    // checked.
+    for (auto operand : op->getOperands()) {
+      if (observedValues.contains(operand))
+        continue;
+      if (isa<hw::InOutType>(operand.getType()))
+        return false;
+      auto *defOp = operand.getDefiningOp();
```
maerhart, line 254: Do we conservatively need to check for no side-effects? Registers should be fine, but not sure about memories.

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
```diff
@@ -0,0 +1,279 @@
...
+    LLVM_DEBUG(llvm::dbgs() << "Skipping process " << processOp.getLoc()
+                            << ": wait op has destination operands\n");
+    return false;
+  }
+
+  // Helper function to skip across empty blocks with only a single successor.
+  auto skipToMergePoint = [&](Block *block) -> std::pair<Block *, ValueRange> {
+    ValueRange operands;
+    while (auto branchOp = dyn_cast<cf::BranchOp>(block->getTerminator())) {
+      if (!block->without_terminator().empty())
```
maerhart, line 107: In the past I've frequently seen constants and `llhd.sig.extract` ops in the entry block. It should still be possible to do the lowering in that case. Or can this not occur anymore with your new mem2reg+deseq pipeline?

maerhart (approved):
LGTM!

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
fabianschuiki: Yeah... Pretty sure this is a premature optimization. But the corresponding function in `RegionUtils.h` does the same thing, which is where I stole the code from :smirk:

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
fabianschuiki: Oh yeah good thinking, let me add that!

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
fabianschuiki: Yeah good point, we probably should stop on side-effecting ops. Let me add that.

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
fabianschuiki: I haven't seen constants and `llhd.sig.extract` in the entry block anymore. This might be due to the processes no longer being isolated from above. If you can pull an op into the entry block, chances are high that you can hoist it out of the process entirely. That might be the reason. But we should keep an eye out for these cases and add handling if they appear. My hope is that we can modify the process op (or create a separate flavor of processes) where you no longer have an entry block that you have to check for ops.

lib/Dialect/LLHD/Transforms/LowerProcesses.cpp
```diff
@@ -0,0 +1,279 @@
...
+  waitOp.erase();
+
+  // Simplify the execute op body region since disconnecting the control flow
+  // loop through the wait op has potentially created unreachable blocks.
+  IRRewriter rewriter(builder);
+  (void)simplifyRegions(rewriter, executeOp->getRegions());
+}
+
+/// Check that the process' entry block trivially joins a control flow loop
+/// immediately after the wait op.
```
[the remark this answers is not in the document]
fabianschuiki: I've seen both. I tend to go back and forth on this… usually I find it helpful if it's close to the implementation, as long as the declaration isn't in a separate header. But that's also pretty inconsistent 😞

PR #8389, opened 2025-04-03
[RTG][Elaboration] Treat randomized sequences as regular identity values

maerhart:
Currently we special-case randomized sequences. This PR changes them to be values with identity. This has the advantage that we don't need to generate a new, unique name for each randomizes sequence from the start, i.e., if we randomize 1000 sequences, put them in a set and only select 1 at random and embed it, then we only need to generate 1 name after this PR instead of 1000. This name uniquification made up a considerable chunk of runtime in some previous profiling. On the negative side, we now pass randomized sequence values as block arguments to nested sequences which makes life a bit more difficult for sequence inlining. Not sure if avoiding this is worth the special-casing.


fabianschuiki (approved):
LGTM

PR #8390, opened 2025-04-04
[FIRRTL] Add VersionAttr, put on Circuit, use to parse FIRRTL 6 Annotations

seldridge:
This adds a new `VersionAttr` which is made an optional attribute on FIRRTL circuits.  This is used to determine which version of the annotation format should be used in the `LowerAnnotations` pass.  This can additionally be used in the future if we ever get into a hard situation where we need to support a legacy behavior in the FIRRTL spec as opposed to just syntax differences or new features.

See individual commits for more information.


seldridge:
After out-of-band discussion, this is _not_ needed. Instead, we can augment the target syntax to optionally elide the circuit name (https://github.com/chipsalliance/firrtl-spec/commit/160bafef053f1fa28e8592ebf54400e81b90e210). This is already accepted by CIRCT and we _don't_ need to carry the version forward.

PR #8391, opened 2025-04-04
[RTG][Elaboration] Improve how tests are matched with targets

maerhart:
* Tests can now carry a string identifying the original test name because a test can be cloned to test multiple targets. This is useful for result reporting
* Tests can now specify a fixed target. Before elaboration it acts as a restriction to only be elaborated against that target, after elaboration it specifies for which target the test was elaborated.
* The body of a target is not copied into the tests anymore. This allows us to materialize operations that are only legal in targets and keep them until later in the pipeline.
* The test <-> target matching now supports a kind of 'subtyping'
* Remove the special casing for tests without requirements.

### DIFF
--- a/include/circt/Dialect/RTG/IR/RTGOps.td
+++ b/include/circt/Dialect/RTG/IR/RTGOps.td
@@ -697,7 +697,8 @@ def TestOp : RTGOp<"test", [
   SingleBlock,
   NoTerminator,
   DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmBlockArgumentNames"]>,
-  HasParent<"mlir::ModuleOp">
+  DeclareOpInterfaceMethods<SymbolUserOpInterface>,
+  HasParent<"mlir::ModuleOp">,
 ]> {
   let summary = "the root of a test";
   let description = [{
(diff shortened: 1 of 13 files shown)


include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -691,14 +691,12 @@ def VirtualRegisterOp : RTGOp<"virtual_reg", [
...
-  NoTerminator,
-  DeclareOpInterfaceMethods<OpAsmOpInterface, ["getAsmBlockArgumentNames"]>,
-  HasParent<"mlir::ModuleOp">
-]> {
+def TestOp
+    : RTGOp<"test", [IsolatedFromAbove, Symbol, SingleBlock, NoTerminator,
+                     DeclareOpInterfaceMethods<
+                         OpAsmOpInterface, ["getAsmBlockArgumentNames"]>,
+                     DeclareOpInterfaceMethods<SymbolUserOpInterface>,
+                     HasParent<"mlir::ModuleOp">]> {
```
fabianschuiki: Super-nit: the rustfmt-style original formatting feels a bit more readable and less noisy in diffs than the clang-format-style formatting of the new code :grin: :bike: :house:

fabianschuiki (approved):
LGTM! I can't judge the RTG-specifics, but code and tests look nice!

PR #8392, opened 2025-04-04
[ImportVerilog] Convert the unpacked array to a simple bit vector

AnnuCode:
With these changes, ImportVerilog can now convert unpacked arrays to simple bit vectors and pass this [test](https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+7.4.3+operations-on-unpacked-arrays-equality).

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -1272,6 +1272,14 @@ Value Context::convertToSimpleBitVector(Value value) {
                                                  value);
     }
   }
+  if (auto unpacked = dyn_cast<moore::UnpackedType>(value.getType())) {
+    if (auto bits = unpacked.getBitSize()) {
+      auto sbvType =
+          moore::IntType::get(value.getContext(), *bits, unpacked.getDomain());
+      return builder.create<moore::ConversionOp>(value.getLoc(), sbvType,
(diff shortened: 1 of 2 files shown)


fabianschuiki (commented):
I think converting unpacked types to a packed simple bit vector might not be legal according to the SV standard 🤔. The implicit cast to a simple bit vector only exists for packed types. This touches on a similar issue as we've seen in the `string` type: it's an unpacked type, but we still need to be able to compare strings for equality and order. I'm wondering if we should introduce dedicated ops for comparing unpacked arrays and strings, something like `moore.cmp_uarray eq %array1, %array2` and `moore.cmp_string eq %array1, %array2`. When we lower from Moore to the core dialects, we'll need to somehow implement these comparisons in a special way anyway, since we don't have core dialect equivalents for things like dynamic strings. So it would make sense to have a distinct op.

My suggestion would be: create a `moore.cmp_uarray` and `moore.cmp_string` operation that works like `arith.cmpi` and `comb.icmp`. It accepts two operands and a comparison predicate (eq, ne, lt, gt, etc.), and produces an `i1` result. We can then start to think about how we should lower these ops, and strings/unpacked arrays in general, in MooreToCore a bit later.

AnnuCode:
@fabianschuiki  Thanks for the correction and the idea 💯

PR #8401, opened 2025-04-07
[AffineParallelOpUnparallelize] Simplify nested SCF IndexSwitch

jiahanxie353:
This patch simplifies the nested `scf.index_switch` operations if their arguments only depend on `affine.parallel`'s loop induction variable and if they are the results of some modulo expressions. Use cases are when they come from the memory banking pass where it uses `scf.index_switch` to decide which bank to use.

### DIFF
--- a/lib/Dialect/Calyx/Transforms/AffinePloopUnparallelize.cpp
+++ b/lib/Dialect/Calyx/Transforms/AffinePloopUnparallelize.cpp
@@ -144,6 +144,9 @@ public:
   }
 
 private:
+  // Collect all simplifiable `scf.index_switch` ops in `affineParallelOp`. An
+  // `scf.index_switch` op is simpliiable if its argument only depends on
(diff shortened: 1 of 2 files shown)


lib/Dialect/Calyx/Transforms/AffinePloopUnparallelize.cpp
```diff
@@ -132,8 +136,44 @@ class AffinePloopUnparallelize
...
+    for (auto indexSwitchOp : simplifiableIndexSwitchOps) {
+      indexSwitchOp.setOperand(innerParallel.getIVs().front());
+    }
+
     return success();
   }
+
+private:
+  SmallVector<scf::IndexSwitchOp>
+  collectSimplifiableIndexSwitchOps(affine::AffineParallelOp affineParallelOp,
```
cgyurgyik, line 148: Nit: add a comment describing what "simplifiable" means here.

cgyurgyik (approved):

PR #8404, opened 2025-04-09
Bump LLVM to 9344b2196cbc36cdc577314bbb2b889606ba6820.

mikeurbach:
There were series of patches to llvm::EquivalenceClasses this week,
and we updated accordingly:

* There is no longer a template parameter for the comparator
* The idiom for iterating over EquivalenceClasses was updated
* The idiom for iterating over members was updated

There was a change to the InlinerUtils functions to now include a
required callback for ops that are cloned. To update our call sites,
this default constructs an InlinerConfig and uses the default
clone callback.

Finally, there was a change to the memref.subview verifier to actually
check bounds when statically possible. Since memref.subview is
technically supposed to "represent a reduced-size view of the
original memref", I think our previous lowering was technically
supposed to be interpreted as out of bound accesses.

See https://discourse.llvm.org/t/out-of-bounds-semantics-of-memref-subview/85293
for more info.

In any case, it seems this is the expected use for
memref.collapse_shape, so this was adopted instead of memref.subview
for this use case.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerXMR.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerXMR.cpp
@@ -130,7 +130,7 @@ class LowerXMRPass : public circt::firrtl::impl::LowerXMRBase<LowerXMRPass> {
     CircuitNamespace ns(getOperation());
     circuitNamespace = &ns;
 
-    llvm::EquivalenceClasses<Value, ValueComparator> eq;
+    llvm::EquivalenceClasses<Value> eq;
     dataFlowClasses = &eq;
 
     InstanceGraph &instanceGraph = getAnalysis<InstanceGraph>();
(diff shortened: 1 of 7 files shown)


mikeurbach:
@mortbopet is there anything depending on the usage of `memref.subview` in the FlattenMemRefs lowering? I didn't see anything in the codebase that expected `memref.subview`, but maybe I'm missing something.

lib/Dialect/FIRRTL/Transforms/LowerXMR.cpp
```diff
@@ -370,14 +370,12 @@ class LowerXMRPass : public circt::firrtl::impl::LowerXMRBase<LowerXMRPass> {
     }
 
     LLVM_DEBUG({
-      for (auto I = dataFlowClasses->begin(), E = dataFlowClasses->end();
-           I != E; ++I) { // Iterate over all of the equivalence sets.
+      for (const auto &I :
+           *dataFlowClasses) { // Iterate over all of the equivalence sets.
```
seldridge, line 374: Mega-nit: Moving the commend _above_ may get `clang-format` to put this all one one line.

seldridge (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/LowerXMR.cpp
makslevental: yup that's correct

mortbopet:
> @mortbopet is there anything depending on the usage of `memref.subview` in the FlattenMemRefs lowering? I didn't see anything in the codebase that expected `memref.subview`, but maybe I'm missing something.

Sorry for the late reply. No, memref lowerings don't handle `memref.subview`.

PR #8407, opened 2025-04-10
[tools] fix `Wdangling-assignment-gsl`

makslevental:
This PR fixes `Wdangling-assignment-gsl` in `circt-bmc`, `circt-lec`, and `arcilator`.

### DIFF
--- a/tools/arcilator/arcilator.cpp
+++ b/tools/arcilator/arcilator.cpp
@@ -461,9 +461,11 @@ static LogicalResult processBuffer(
 
     mlir::ExecutionEngineOptions engineOptions;
     engineOptions.jitCodeGenOptLevel = llvm::CodeGenOptLevel::Aggressive;
-    engineOptions.transformer = mlir::makeOptimizingTransformer(
-        /*optLevel=*/3, /*sizeLevel=*/0,
-        /*targetMachine=*/nullptr);
+    static std::function<llvm::Error(llvm::Module *)> transformer =
+        mlir::makeOptimizingTransformer(
+            /*optLevel=*/3, /*sizeLevel=*/0,
+            /*targetMachine=*/nullptr);
+    engineOptions.transformer = transformer;
     engineOptions.sharedLibPaths = sharedLibraries;
 
     auto executionEngine =
--- a/tools/arcilator/arcilator.cpp
+++ b/tools/arcilator/arcilator.cpp
@@ -461,7 +461,7 @@ static LogicalResult processBuffer(
 
     mlir::ExecutionEngineOptions engineOptions;
     engineOptions.jitCodeGenOptLevel = llvm::CodeGenOptLevel::Aggressive;
-    static std::function<llvm::Error(llvm::Module *)> transformer =
+    std::function<llvm::Error(llvm::Module *)> transformer =
         mlir::makeOptimizingTransformer(
             /*optLevel=*/3, /*sizeLevel=*/0,
             /*targetMachine=*/nullptr);
--- a/tools/circt-bmc/circt-bmc.cpp
+++ b/tools/circt-bmc/circt-bmc.cpp
@@ -264,8 +264,10 @@ static LogicalResult executeBMC(MLIRContext &context) {
     SmallVector<StringRef, 4> sharedLibraries(sharedLibs.begin(),
                                               sharedLibs.end());
     mlir::ExecutionEngineOptions engineOptions;
-    engineOptions.transformer = mlir::makeOptimizingTransformer(
-        /*optLevel*/ 3, /*sizeLevel=*/0, /*targetMachine=*/nullptr);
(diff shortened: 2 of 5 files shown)


makslevental:
This currently fails here

https://github.com/makslevental/circt/blob/37b7baff75754211a89875df6353dd12a9263c61/test/Transforms/flatten_memref_calls.mlir#L8

```mlir
// CHECK-LABEL:   func @main() {
// CHECK:           %[[VAL_0:.*]] = arith.constant 0 : index
// CHECK:           %[[VAL_1:.*]] = memref.alloca() : memref<30x30xi32>
// CHECK:           %[[VAL_2:.*]] = memref.subview %[[VAL_1]][0, 0] [1, 900] [1, 1] : memref<30x30xi32> to memref<900xi32>
// CHECK:           %[[VAL_3:.*]] = call @foo(%[[VAL_2]]) : (memref<900xi32>) -> i32
// CHECK:           return
// CHECK:         }
module  {
  func.func private @foo(memref<30x30xi32>) -> i32
  func.func @main() {
    %0 = memref.alloca() : memref<30x30xi32>
    %1 = call @foo(%0) : (memref<30x30xi32>) -> i32
    return
  }
}
```
because of https://github.com/llvm/llvm-project/commit/529ee3cf3bb97b7ff468b5046372d070a8667147 but I'm not sure how this ever worked- the subview is clearly out of bounds? I'm not familiar with this pass but is `memref.reshape` a better alternative here? Also btw seems like [`[-gpu-decompose-memrefs](https://mlir.llvm.org/docs/Passes/#-gpu-decompose-memrefs)`](https://github.com/llvm/llvm-project/blob/556a64507b3ae4ee3b998c47eb5451a60235bcb1/mlir/lib/Dialect/GPU/Transforms/DecomposeMemRefs.cpp) does the same thing?

tools/arcilator/arcilator.cpp
```diff
@@ -461,9 +461,11 @@ static LogicalResult processBuffer(
 
     mlir::ExecutionEngineOptions engineOptions;
     engineOptions.jitCodeGenOptLevel = llvm::CodeGenOptLevel::Aggressive;
-    engineOptions.transformer = mlir::makeOptimizingTransformer(
-        /*optLevel=*/3, /*sizeLevel=*/0,
-        /*targetMachine=*/nullptr);
+    static std::function<llvm::Error(llvm::Module *)> transformer =
```
makslevental, line 464: otherwise 

```
warning: object backing the pointer engineOptions.transformer will be destroyed at the end of the full-expression [-Wdangling-assignment-gsl]
```

because `engineOptions.transformer` is a `function_ref`.

mikeurbach:
@makslevental I had started an LLVM update here: https://github.com/llvm/circt/pull/8404. Should I get that one in before your SMT-focused changes?

tools/arcilator/arcilator.cpp
fabianschuiki: Is using a global variable the approved way of doing this upstream? I suspect a `auto transformer = ...; engineOptions.transformer = transformer;` might suffice 🤔

makslevental: it's locally allocated/scope - it'll die at the end of the body while the `executionEngine` is returned

mikeurbach:
Ok, I merged mine. Sorry for the delay and apparent merge conflicts. Hopefully simple to rebase and just deal with the function_ref thing.

makslevental:
Okay I rebased and renamed this PR to just be about `Wdangling-assignment-gsl`.

tools/arcilator/arcilator.cpp
maerhart: To me it looks like an `auto transformer = ...` would have the same lifetime as the execution engine. Where is the engine returned?

makslevental: oh um whoops - I skimmed the file too quickly (the return is `success`). Lemme change.

tools/circt-bmc/circt-bmc.cpp
```diff
@@ -264,8 +264,10 @@ static LogicalResult executeBMC(MLIRContext &context) {
     SmallVector<StringRef, 4> sharedLibraries(sharedLibs.begin(),
                                               sharedLibs.end());
     mlir::ExecutionEngineOptions engineOptions;
-    engineOptions.transformer = mlir::makeOptimizingTransformer(
-        /*optLevel*/ 3, /*sizeLevel=*/0, /*targetMachine=*/nullptr);
+    std::function<llvm::Error(llvm::Module *)> transformer =
```
maerhart, line 267: Don't you have to declare the variable outside this nested scope, because otherwise it'd be deallocated at the `engine->invokePacked` call, right? (also in `circt-lec.cpp`)

maerhart (approved):
LGTM!

PR #8408, opened 2025-04-10
[LLVM] integrate upstream SMT

makslevental:
This PR is a "trial balloon" - it bumps LLVM to the merge commit associated with https://github.com/llvm/llvm-project/pull/131480.

Note, it is also stacked on top of https://github.com/llvm/circt/pull/8407 which contains the necessary/unrelated LLVM integrations.


lib/Conversion/CombToSMT/CombToSMT.cpp
```diff
@@ -34,8 +34,8 @@ struct CombReplicateOpConversion : OpConversionPattern<ReplicateOp> {
   LogicalResult
   matchAndRewrite(ReplicateOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    rewriter.replaceOpWithNewOp<smt::RepeatOp>(op, op.getMultiple(),
-                                               adaptor.getInput());
+    rewriter.replaceOpWithNewOp<mlir::smt::RepeatOp>(op, op.getMultiple(),
```
TaoBi22, line 37: Is there a reason to not just `using namespace mlir` in this file?

TaoBi22 (approved):
This looks good to me, thanks!

lib/Conversion/CombToSMT/CombToSMT.cpp
makslevental: At first I did that but then I couldn't tell what the convention you guys are following - whether code outside of CIRCT should be fully qualified. If you say `using namespace mlir` is ok that's good enough for me!

TaoBi22: I'm definitely not an authority on our style conventions, but it [looks like](https://github.com/search?q=repo%3Allvm%2Fcirct+%22using+namespace+mlir%22+path%3A%2F%5Elib%5C%2FConversion%5C%2F%2F&type=code) plenty of conversion passes use it!

maerhart (approved):
LGTM, thanks for doing the integration!

PR #8415, opened 2025-04-14
[AffineParallelUnroll] A more general approach to hoist memory reads across `scf.execute_region`s

jiahanxie353:
This patch supports a more general approach (previously only considering memory accesses with constant indices) to hoist memory reads out from the `scf.execute_region`s under an `affine.parallel` to prevent memory read access conflict.

### DIFF
--- a/lib/Dialect/Calyx/Transforms/AffineParallelUnroll.cpp
+++ b/lib/Dialect/Calyx/Transforms/AffineParallelUnroll.cpp
@@ -15,6 +15,7 @@
 #include "mlir/Dialect/Affine/IR/AffineMemoryOpInterfaces.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/OperationSupport.h"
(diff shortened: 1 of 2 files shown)


lib/Dialect/Calyx/Transforms/AffineParallelUnroll.cpp
```diff
@@ -45,201 +83,151 @@ namespace {
...
+        allReadOps.insert(read);
 
-      SmallVector<int64_t, 4> evaluatedIndices, foldedResults;
-      bool hasPoison = false;
-      map.partialConstantFold(operandConsts, &foldedResults, &hasPoison);
-      if (!(hasPoison || foldedResults.empty()))
-        constantMemAccessIndices[op] = foldedResults;
+        AffineAccessExpr key{read.getMemRef(), read.getAffineMap(),
+                             read.getMapOperands()};
+        readOpCounts[key]++;
```
cgyurgyik, line 134:
```suggestion
        ++readOpCounts[key];
```

lib/Dialect/Calyx/Transforms/AffineParallelUnroll.cpp
```diff
@@ -45,201 +83,151 @@ namespace {
...
-    auto key = std::pair(readOp.getMemRef(), constIndices);
-    if (constantReadOpCounts[key] > 1) {
-      if (hoistedReads.find(key) == hoistedReads.end()) {
-        builder.setInsertionPoint(affineParallelOp);
-        Operation *clonedRead = builder.clone(*readOp.getOperation());
-        hoistedReads[key] = clonedRead->getOpResults();
-      }
-      readOp->replaceAllUsesWith(hoistedReads[key]);
+    // Only hoist once per unique access
+    if (!hoistedReads.contains(key)) {
```
cgyurgyik, line 207: Can we use `.find(key)` please to avoid triple lookup (`contains`, then `[key]` twice below)?

test/Dialect/Calyx/affine-parallel-unroll.mlir
```diff
@@ -360,3 +391,130 @@ module {
     return
   }
 }
+
+// -----
+
+// Test hosting read when the access indices depend on outer nested for-loops.
```
cgyurgyik, line 397:
```suggestion
// Test hoisting read when the access indices depend on outer nested for-loops.
```

cgyurgyik (approved):
LGTM, a few small nits.

jiahanxie353:
Thanks for the review! @cgyurgyik

Issue #8417, opened 2025-04-15
[FIRRTL] Cannot use Paths to target port of an extmodule


mikeurbach:
Minimal reproducer:

```firrtl
FIRRTL version 5.1.0

circuit Test:
  public module Test:
    input clock : Clock
    output foo_port : Path

    inst foo of Foo
    connect foo.clock, clock

    propassign foo_port, path("OMReferenceTarget:~|Foo>clock")

  extmodule Foo:
    input clock : Clock
```

```
firtool example.fir
```

Fails with:

```
firtool: /scratch/verif/mikeu/circt/lib/Dialect/FIRRTL/FIRRTLUtils.cpp:866: hw::InnerRefAttr circt::firrtl::getInnerRefTo(const hw::InnerSymTarget &, circt::firrtl::GetNamespaceCallback): Assertion `mod && "must be an operation inside an FModuleOp or port of FModuleOp"' \
failed.
PLEASE submit a bug report to https://github.com/llvm/circt and include the crash backtrace.
Stack dump:
0.      Program arguments: build/bin/firtool example.fir
 #0 0x0000000000d822ed llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /scratch/verif/mikeu/circt/llvm/llvm/lib/Support/Unix/Signals.inc:804:11
 #1 0x0000000000d827db PrintStackTraceSignalHandler(void*) /scratch/verif/mikeu/circt/llvm/llvm/lib/Support/Unix/Signals.inc:880:1
 #2 0x0000000000d80806 llvm::sys::RunSignalHandlers() /scratch/verif/mikeu/circt/llvm/llvm/lib/Support/Signals.cpp:105:5
 #3 0x0000000000d8301d SignalHandler(int, siginfo_t*, void*) /scratch/verif/mikeu/circt/llvm/llvm/lib/Support/Unix/Signals.inc:418:7
 #4 0x00007f7421bc8cf0 __restore_rt (/lib64/libpthread.so.0+0x12cf0)
 #5 0x00007f74205c6acf raise (/lib64/libc.so.6+0x4eacf)
 #6 0x00007f7420599ea5 abort (/lib64/libc.so.6+0x21ea5)
 #7 0x00007f7420599d79 _nl_load_domain.cold.0 (/lib64/libc.so.6+0x21d79)
 #8 0x00007f74205bf426 (/lib64/libc.so.6+0x47426)
[...]

PR #8418, opened 2025-04-15
[FIRRTL] Support Paths that target ports of extmodules.

mikeurbach:
Existing target support was built up asserting that only ports of FModuleOps are targeted. But there isn't anything really fundamental here, it looks like we have everything in we need in the FModuleLike interface and inner symbol infrastructure to put targets on ports of FExtModuleOps in exactly the same manner.

Fixes https://github.com/llvm/circt/issues/8417.


mikeurbach:
> Can the PR include a simpler test that just an annotation can be applied to an extmodule port? 

@seldridge were you looking for something like this one (which you dug up)? https://github.com/llvm/circt/blob/b26b4ae5a4b862982fcc77678c365c86598a3630/test/Dialect/FIRRTL/annotations.mlir#L497

Or should I add some other test here?

seldridge:
That test is sufficient. I had thought that we weren't testing this at all. However, there has always been a test that you can annotate extmodule ports.

PR #8419, opened 2025-04-15
[FIRRTL][SV] Support substitutions in fprintf output file name. Add sv.sformatf. 

uenoku:
Add a new SFormatFOp operation to handle string formatting in SystemVerilog.
This operation produces a constant string literal and supports variable
substitutions. Also enhance FIRRTL fprintf operation by:
- Support dynamic file names in FIRRTL fprintf operations
- Replace FStringType with PrintfOperandType for output file substitutions
- Add SystemVerilog sformatf lowering for formatted file names
- Update file descriptor handling in emitter

It implements micro-optimization regarding FD creation. If the file is known to be static, the fd is initialized in initial block but for dynamic file name it opens fd just before its uses. 

```scala
FIRRTL version 5.1.0

circuit PrintTest:
  public module PrintTest :
    input clock : Clock
    input cond : UInt<1>
    input a : UInt<8>

    fprintf(clock, cond, "test.txt", "static file name (w/o substitution)\n")

    node c = UInt<8>(97)
    fprintf(clock, cond, "{{HierarchicalModuleName}}%c.txt", c, "[{{SimulationTime}}]: static file name (w/ substitution)\n")

    fprintf(clock, cond, "{{SimulationTime}}%d.txt", a, "[{{SimulationTime}}]: dynamic file name\n")
```

```verilog
module PrintTest(
  input       clock,
              cond,
  input [7:0] a
);

  logic [31:0] ___circt_lib_logging3A3AFileDescriptor3A3Aget_0;
  `ifndef SYNTHESIS
    reg [31:0] fd_test_txt;
    reg [31:0] fd_25m25c_txt;
    initial begin
      fd_test_txt = __circt_lib_logging::FileDescriptor::get("test.txt");
      fd_25m25c_txt = __circt_lib_logging::FileDescriptor::get($sformatf("%m%c.txt",
                                                                         8'h61));
    end // initial
    always @(posedge clock) begin
      if ((`PRINTF_COND_) & cond) begin
        $fwrite(fd_test_txt, "static file name (w/o substitution)\n");
        $fwrite(fd_25m25c_txt, "[%0t]: static file name (w/ substitution)\n", $time);
        ___circt_lib_logging3A3AFileDescriptor3A3Aget_0 = __circt_lib_logging::FileDescriptor::get($sformatf("%0t%d.txt",
                                                                                                             $time,
                                                                                                             a));
        $fwrite(___circt_lib_logging3A3AFileDescriptor3A3Aget_0,
                "[%0t]: dynamic file name\n", $time);
      end
    end // always @(posedge)
  `endif // not def SYNTHESIS
endmodule
```

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLStatements.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStatements.td
@@ -122,7 +122,7 @@ def PrintFOp : FIRRTLOp<"printf"> {
   }];
 }
 
-def FPrintFOp : FIRRTLOp<"fprintf"> {
+def FPrintFOp : FIRRTLOp<"fprintf", [AttrSizedOperandSegments]> {
   let summary = "Formatted File Print Statement";
 
   let description = [{
@@ -131,12 +131,16 @@ def FPrintFOp : FIRRTLOp<"fprintf"> {
   }];
 
   let arguments = (ins ClockType:$clock, UInt1Type:$cond, StrAttr:$outputFile,
+                       // Use FStringType instead of PrintfOperandType to reject hardware types.
+                       Variadic<FStringType>:$outputFileSubstitutions,
                        StrAttr:$formatString, Variadic<PrintfOperandType>:$substitutions,
                        StrAttr:$name);
   let results = (outs);
 
+  let hasVerifier = 1;
+
   let assemblyFormat = [{
-    $clock `,` $cond `,` $outputFile `,` $formatString ``
+    $clock `,` $cond `,` $outputFile (` ` `(` $outputFileSubstitutions^ `)`)? `,` $formatString ``
     custom<FPrintfAttrs>(attr-dict) ` ` (`(` $substitutions^ `)`)? `:`
     type($clock) `,` type($cond) (`,` qualified(type($substitutions))^)?
   }];
--- a/include/circt/Dialect/FIRRTL/FIRRTLStatements.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStatements.td
@@ -131,19 +131,11 @@ def FPrintFOp : FIRRTLOp<"fprintf", [AttrSizedOperandSegments]> {
   }];
 
   let arguments = (ins ClockType:$clock, UInt1Type:$cond, StrAttr:$outputFile,
-                       // Use FStringType instead of PrintfOperandType to reject hardware types.
-                       Variadic<FStringType>:$outputFileSubstitutions,
+                       Variadic<PrintfOperandType>:$outputFileSubstitutions,
                        StrAttr:$formatString, Variadic<PrintfOperandType>:$substitutions,
                        StrAttr:$name);
   let results = (outs);
-
-  let hasVerifier = 1;
-
-  let assemblyFormat = [{
-    $clock `,` $cond `,` $outputFile (` ` `(` $outputFileSubstitutions^ `)`)? `,` $formatString ``
-    custom<FPrintfAttrs>(attr-dict) ` ` (`(` $substitutions^ `)`)? `:`
-    type($clock) `,` type($cond) (`,` qualified(type($substitutions))^)?
-  }];
+  let hasCustomAssemblyFormat = 1;
 }
 
 def SkipOp : FIRRTLOp<"skip", [Pure]> {
--- a/include/circt/Dialect/SV/SVExpressions.td
+++ b/include/circt/Dialect/SV/SVExpressions.td
@@ -201,6 +201,18 @@ def ConstantStrOp : SVOp<"constantStr", [Pure]> {
   let assemblyFormat = " $str attr-dict";
 }
 
+def SFormatFOp : SVOp<"sformatf", [Pure]> {
+  let summary = "sformatf";
+  let description = [{
+    This operation produces a constant string literal.
+  }];
+
+  let arguments = (ins StrAttr: $format_string, Variadic<AnyType>:$substitutions);
+  let results = (outs HWStringType:$result);
+  let assemblyFormat =
+    "$format_string attr-dict (`(` $substitutions^ `)` `:` qualified(type($substitutions)))?";
+}
+
 def LocalParamOp : SVOp<"localparam",
       [FirstAttrDerivedResultType, Pure, HasCustomSSAName]> {
   let summary = "Declare a localparam";
--- a/include/circt/Dialect/SV/SVExpressions.td
+++ b/include/circt/Dialect/SV/SVExpressions.td
@@ -202,9 +202,12 @@ def ConstantStrOp : SVOp<"constantStr", [Pure]> {
 }
 
 def SFormatFOp : SVOp<"sformatf", [Pure]> {
-  let summary = "sformatf";
+  let summary = "'$sformatf' task";
   let description = [{
-    This operation produces a constant string literal.
+    This operations represents `$sformatf` task that produces a string from
+    a format string and substitutions.
+
+    See section 21.3.3 of 1800-2023 for more details.
   }];
 
   let arguments = (ins StrAttr: $format_string, Variadic<AnyType>:$substitutions);
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -3235,7 +3235,7 @@ SubExprInfo ExprEmitter::visitSV(SFormatFOp op) {
     ps.writeQuotedEscaped(op.getFormatString());
     // TODO: if any of these breaks, it'd be "nice" to break
     // after the comma, instead of:
-    // $fwrite(5, "...", a + b,
+    // $sformatf("...", a + b,
     //         longexpr_goes
     //         + here, c);
     // (without forcing breaking between all elements, like braced list)
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -247,7 +247,7 @@ bool ExportVerilog::isVerilogExpression(Operation *op) {
   // These are SV dialect expressions.
   if (isa<ReadInOutOp, AggregateConstantOp, ArrayIndexInOutOp,
           IndexedPartSelectInOutOp, StructFieldInOutOp, IndexedPartSelectOp,
-          ParamValueOp, XMROp, XMRRefOp, SampledOp, EnumConstantOp,
+          ParamValueOp, XMROp, XMRRefOp, SampledOp, EnumConstantOp, SFormatFOp,
           SystemFunctionOp, STimeOp, TimeOp, UnpackedArrayCreateOp,
           UnpackedOpenArrayCastOp>(op))
     return true;
@@ -2315,6 +2315,7 @@ private:
   SubExprInfo visitSV(SystemFunctionOp op);
   SubExprInfo visitSV(ReadInterfaceSignalOp op);
   SubExprInfo visitSV(XMROp op);
+  SubExprInfo visitSV(SFormatFOp op);
   SubExprInfo visitSV(XMRRefOp op);
   SubExprInfo visitVerbatimExprOp(Operation *op, ArrayAttr symbols);
   SubExprInfo visitSV(VerbatimExprOp op) {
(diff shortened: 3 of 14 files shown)


include/circt/Dialect/FIRRTL/FIRRTLStatements.td
```diff
@@ -131,12 +131,16 @@ def FPrintFOp : FIRRTLOp<"fprintf"> {
   }];
 
   let arguments = (ins ClockType:$clock, UInt1Type:$cond, StrAttr:$outputFile,
+                       // Use FStringType instead of PrintfOperandType to reject hardware types.
+                       Variadic<FStringType>:$outputFileSubstitutions,
```
seldridge, line 135: This is a good use of `FStringType` until we want to (if we want to) expand it later.

include/circt/Dialect/FIRRTL/FIRRTLStatements.td
```diff
@@ -131,12 +131,16 @@ def FPrintFOp : FIRRTLOp<"fprintf"> {
...
                        StrAttr:$formatString, Variadic<PrintfOperandType>:$substitutions,
                        StrAttr:$name);
   let results = (outs);
 
+  let hasVerifier = 1;
+
   let assemblyFormat = [{
-    $clock `,` $cond `,` $outputFile `,` $formatString ``
+    $clock `,` $cond `,` $outputFile (` ` `(` $outputFileSubstitutions^ `)`)? `,` $formatString ``
```
seldridge, line 143: Mega-nit: the `formatString`, when serialized, doesn't have a space after it, but the `outputFile` format string does. It at least shows this in the `parse-basic.fir` test. Is there a way to make this the same?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -5852,7 +5852,18 @@ static ParseResult parseFPrintfAttrs(OpAsmParser &p,
...
+}
+
+LogicalResult FPrintFOp::verify() {
+  for (auto operand : getOutputFileSubstitutions())
+    if (!operand.getDefiningOp<firrtl::HierarchicalModuleNameOp>())
+      return mlir::emitError(operand.getLoc())
+             << "Only a hierarchical module name can be used as a substitution "
+                "for fprintf output file name";
+
+  return success();
```
seldridge, line 5866: Nice verifier. It didn't occur to me that the verifiers are much easier now given the fstring ops as this no longer requires re-parsing.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -4746,14 +4763,23 @@ LogicalResult FIRRTLLowering::visitPrintfLike(T op, StringAttr outputFile) {
...
 
         addToInitialBlock([&]() {
-          auto constant = builder.create<sv::ConstantStrOp>(outputFile);
+          Value fileNameVal = builder.create<sv::ConstantStrOp>(outputFile);
+          if (outputFileAndRequiresSubst.getInt()) {
+            fileNameVal = builder
+                              .create<sv::SystemFunctionOp>(
+                                  hw::StringType::get(builder.getContext()),
+                                  builder.getStringAttr("sformatf"),
+                                  ValueRange{fileNameVal})
```
seldridge, line 4772: Eventually we probably want to make this a first-class system function that CIRCT knows about.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -4724,8 +4721,28 @@ LogicalResult FIRRTLLowering::visitPrintfLike(T op, StringAttr outputFile) {
...
+  return success();
+}
+
+// Printf/FPrintf is a macro op that lowers to an sv.ifdef.procedural, an sv.if,
+// and an sv.fwrite all nested together.
+template <class T>
+LogicalResult FIRRTLLowering::visitPrintfLike(
+    T op, llvm::PointerIntPair<StringAttr, 1> outputFileAndRequiresSubst) {
+  auto clock = getLoweredNonClockValue(op.getClock());
+  auto cond = getLoweredValue(op.getCond());
```
seldridge, line 4734: This template is almost hinting that there is a missing interface for "command" statements. That is a discussion for a later time...

seldridge (commented):
Please update the `FIREmitter` as well.

include/circt/Dialect/FIRRTL/FIRRTLStatements.td
uenoku: This is replaced with PrintfOperand type as it's necessary to handle as well. A side effect of PrintfOperand not being buildable type is that assembly format got too hard to keep the consistency with other operations. So I end up implementing custom assembly format.

include/circt/Dialect/FIRRTL/FIRRTLStatements.td
uenoku: It should be fixed in hand-written printer.

include/circt/Dialect/FIRRTL/FIRRTLStatements.td
```diff
@@ -131,15 +131,11 @@ def FPrintFOp : FIRRTLOp<"fprintf"> {
...
                        StrAttr:$formatString, Variadic<PrintfOperandType>:$substitutions,
                        StrAttr:$name);
   let results = (outs);
-
-  let assemblyFormat = [{
-    $clock `,` $cond `,` $outputFile `,` $formatString ``
-    custom<FPrintfAttrs>(attr-dict) ` ` (`(` $substitutions^ `)`)? `:`
-    type($clock) `,` type($cond) (`,` qualified(type($substitutions))^)?
-  }];
+  let hasCustomAssemblyFormat = 1;
```
seldridge, line 138: Nit: does this _have_ to use a custom assembly format or is there some format that could work here? I realize that the double variadic may make this hard. I would've thought that the default format could handle this as long as there are appropriate anchors.

My motivation for using the default format is that it is just lower maintenance.

test/Dialect/FIRRTL/parse-basic.fir
```diff
@@ -2083,10 +2083,14 @@ FIRRTL version 5.1.0
...
-    ; CHECK{{LITERAL}}: firrtl.fprintf %clock, %c1_ui1, "test.txt", "[{{}}]: hello world {{}}"(%[[TIME]], %[[HIERARCHICALMODULENAME]])
-    fprintf(clock, UInt<1>(1), "test.txt", "[{{SimulationTime}}]: hello from {{HierarchicalModuleName}}")
+    input c: UInt<8>
+    ; CHECK:            %[[HIERARCHICALMODULENAME_0:.+]] = firrtl.fstring.hierarchicalmodulename : !firrtl.fstring
+    ; CHECK-NEXT:       %[[TIME_0:.+]] = firrtl.fstring.time : !firrtl.fstring
+    ; CHECK-NEXT:       %[[TIME_1:.+]] = firrtl.fstring.time : !firrtl.fstring
+    ; CHECK-NEXT:       %[[HIERARCHICALMODULENAME_1:.+]] = firrtl.fstring.hierarchicalmodulename : !firrtl.fstring
+    ; CHECK{{LITERAL}}:      firrtl.fprintf %clock, %c1_ui1, "{{}}{{}}%d.txt" (%[[HIERARCHICALMODULENAME_0]], %[[TIME_0]], %c),
+    ; CHECK-SAME{{LITERAL}}:   "[{{}}]: hello world {{}}"(%[[TIME_1]], %[[HIERARCHICALMODULENAME_1]])
+    fprintf(clock, UInt<1>(1), "{{HierarchicalModuleName}}{{SimulationTime}}%d.txt", c, "[{{SimulationTime}}]: hello from {{HierarchicalModuleName}}")
```
seldridge, line 2093: This syntax will eventually be problematic if we ever decide to support `%s` where a literal string can be substituted. We will likely need to migrate to something like:

```
fprintf(clock, cond, "%d"(a), "%d"(a))
fprintf(clock, cond, sprintf("%d", a), "%d", a)
fprintf(clock, cond, sprintf("%d", a), sprintf("%d", a))
```

Likely, the full `sprintf` is cleanest and we can just change this all at once for these.

This isn't blocking. We can have this discussion for FIRRTL 6.0.

include/circt/Dialect/SV/SVExpressions.td
```diff
@@ -201,6 +201,18 @@ def ConstantStrOp : SVOp<"constantStr", [Pure]> {
   let assemblyFormat = " $str attr-dict";
 }
 
+def SFormatFOp : SVOp<"sformatf", [Pure]> {
+  let summary = "sformatf";
+  let description = [{
+    This operation produces a constant string literal.
```
seldridge, line 207: Saying "constant" may not be correct.

Can you mention `Section 21.3.3 of 1800-2023`. 😉

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -3225,6 +3226,28 @@ SubExprInfo ExprEmitter::visitSV(SampledOp op) {
...
+SubExprInfo ExprEmitter::visitSV(SFormatFOp op) {
+  if (hasSVAttributes(op))
+    emitError(op, "SV attributes emission is unimplemented for the op");
+
+  ps << "$sformatf(";
+  ps.scopedBox(PP::ibox0, [&]() {
+    ps.writeQuotedEscaped(op.getFormatString());
+    // TODO: if any of these breaks, it'd be "nice" to break
+    // after the comma, instead of:
+    // $fwrite(5, "...", a + b,
```
seldridge, line 3238: Is this comment wrong?

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -199,6 +199,51 @@ static void tryCopyName(Operation *dst, Operation *src) {
       dst->setAttr("sv.namehint", attr);
 }
 
+namespace {
+
+// A helper strutc to hold information about output file descriptor.
```
seldridge, line 204:
```suggestion
// A helper struct to hold information about output file descriptor.
```

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -199,6 +199,51 @@ static void tryCopyName(Operation *dst, Operation *src) {
...
+struct FileDescriptorInfo {
```
seldridge, line 205: Nit: LLVM coding guidelines would say this should be a `class` since it has private members.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -199,6 +199,51 @@ static void tryCopyName(Operation *dst, Operation *src) {
...
+private:
+  // Set true if the file name is dynamic, i.e. $time or normal hardware value
+  // is substituted.
+  bool isDynamicFileName;
+
+  // "Verilog" format string for the output file.
+  StringAttr outputFileFormat = {};
+
+  // "FIRRTL" pre-lowered operands.
+  mlir::ValueRange substitutions;
```
seldridge, line 242: Nit: the private members are commented, but the public ones are not.

lib/Dialect/FIRRTL/Export/FIREmitter.cpp
```diff
@@ -945,10 +938,38 @@ void Emitter::emitPrintfLike(OpTy op, StringAttr fileName) {
...
+    emitExpression(op.getCond());
+    ps << "," << PP::space;
+
+    SmallVector<Value, 4> outputFileSubstitutions;
+    emitFormatString(op, op.getOutputFile(), op.getOutputFileSubstitutions(),
+                     outputFileSubstitutions);
+    for (auto operand : outputFileSubstitutions) {
+      ps << "," << PP::space;
+      emitExpression(operand);
+    }
```
seldridge, line 956: `interleaveComma`?

lib/Dialect/FIRRTL/Export/FIREmitter.cpp
```diff
@@ -945,10 +938,38 @@ void Emitter::emitPrintfLike(OpTy op, StringAttr fileName) {
...
+
+    ps << "," << PP::space;
+
+    SmallVector<Value, 4> substitutions;
+    emitFormatString(op, op.getFormatString(), op.getSubstitutions(),
+                     substitutions);
+    for (auto operand : substitutions) {
+      ps << "," << PP::space;
+      emitExpression(operand);
+    }
```
seldridge, line 966: `interleaveComma`?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge, line 1: I barely reviewed this.

seldridge (approved):
LGTM

Thank you for getting this together. I realize that this may be a massive footgun if users are not careful and use lots of dynamic file names. However, I think this is pragmatic.

Thank you for the thought that went into the CIRCT logging library. 💯

PR #8422, opened 2025-04-16
[Verif] Add pass to lower symbolic values

fabianschuiki:
Add the `LowerSymbolicValues` pass which converts `verif.symbolic_value` ops either to a black box instance or to a wire declaration with a `(* anyseq *)` attribute. The former is useful since most formal tools have a way of treating black boxes as internal port boundaries, such that black box output ports behave the same as top-level input ports. The latter is useful for interoperability with Yosys which has a custom Verilog attribute "anyseq" that marks a wire declaration as a symbolic value.

### DIFF
--- a/lib/Firtool/Firtool.cpp
+++ b/lib/Firtool/Firtool.cpp
@@ -306,6 +306,8 @@ LogicalResult firtool::populateHWToSV(mlir::PassManager &pm,
                                       const FirtoolOptions &opt) {
   pm.nestAny().addPass(verif::createStripContractsPass());
   pm.addPass(verif::createLowerFormalToHWPass());
+  pm.addPass(
+      verif::createLowerSymbolicValuesPass({opt.getSymbolicValueLowering()}));
 
   if (opt.shouldExtractTestCode())
     pm.addPass(sv::createSVExtractTestCodePass(
@@ -752,6 +754,12 @@ struct FirtoolCmdOptions {
       llvm::cl::desc(
           "Specialize instance choice to default, if no option selected"),
       llvm::cl::init(false)};
+
+  llvm::cl::opt<verif::SymbolicValueLowering> symbolicValueLowering{
+      "symbolic-values",
+      llvm::cl::desc("Control how symbolic values are lowered"),
+      llvm::cl::init(verif::SymbolicValueLowering::ExtModule),
+      verif::symbolicValueLoweringCLValues()};
 };
 } // namespace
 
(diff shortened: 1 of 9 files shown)


lib/Firtool/Firtool.cpp
```diff
@@ -306,6 +306,8 @@ LogicalResult firtool::populateHWToSV(mlir::PassManager &pm,
                                       const FirtoolOptions &opt) {
   pm.nestAny().addPass(verif::createStripContractsPass());
   pm.addPass(verif::createLowerFormalToHWPass());
+  pm.addPass(
```
dobios, line 309: Just making sure I understand this right, does `opt.shouldExtractTestCode()` state that test code should be lowered or should be removed? If it means that it should be lowered, shouldn't this pass only be run if we want to generate "formal verification" SV and symbolic values should be outright removed otherwise?

test/Dialect/Verif/lower-symbolic-values.mlir
```diff
@@ -0,0 +1,46 @@
...
+  // Reuse existing extmodule for same 42 bit types, cast to array<6 x i7>.
+  // CHECK-EXTMODULE: [[TMP:%.+]] = hw.instance {{.*}} @circt.symbolic_value.42<WIDTH: i32 = 42>
+  // CHECK-EXTMODULE: [[SYM:%.+]] = hw.bitcast [[TMP]] : (i42) -> !hw.array<6xi7>
+  // CHECK-YOSYS: [[TMP:%.+]] = sv.wire {sv.attributes = [#sv.attribute<"anyseq">]} : !hw.inout<array<6xi7>>
+  // CHECK-YOSYS: [[SYM:%.+]] = sv.read_inout [[TMP]]
+  // CHECK: dbg.variable "x3", [[SYM]] : !hw.array<6xi7>
+  %3 = verif.symbolic_value : !hw.array<6xi7>
+  dbg.variable "x3", %3 : !hw.array<6xi7>
+}
+
```
dobios, line 41: Is there also a test for when we should remove symbolic values when lowering to SV?

dobios (approved):
Looks great! This is super useful and hopefully enables these formal tests to be used in downstream SV tools! I just wanted to make sure I understand the tooling correctly, so I left some small comments about that.

lib/Firtool/Firtool.cpp
fabianschuiki: ExtractTestCode is a FIRRTL-specific pass that moves assert/assume/cover statements into a separate Verilog file/module that can then be bound into the original location. As long as we create any asserts/assumes/covers before that pass runs, everything should be good 👍

test/Dialect/Verif/lower-symbolic-values.mlir
fabianschuiki: Currently there's no mode for the pass that removes symbolic values. I'm not entirely sure how that would look like. Replacing with a constant feels wrong, and otherwise I don't know what we could put there instead of the symbolic value. Do you have any ideas?

PR #8423, opened 2025-04-17
[FIRRTL] Dedup: dedup wires when only one has a symbol

youngar:
[FIRRTL] Dedup: dedup wires when only one has a symbol
This change allows deduplication of wires when only one of them has an inner symbol.  Before, when hashing an operation we would include the index/id of the inner symbol in the wire's hash, which was done to make sure that local inner symbol references were targeting the same declaration between two modules.  Now, we preprocess the inner symbols inside a module, and with a stable identity of inner symbols we can confidently use the index of an innersymbol when hashing an inner symbol reference.  With this change, we don't have to include the inner symbol on the declaration operation, meaning that if there are no references to the symbol, the fact that an declaration has an inner symbol does not matter at all.

[FIRRTL] Dedup: properly merge inner symbols
When we deduplicate modules, we allow operations with different inner
symbol definitions to deduplicate with each other. Unfortunately the
operation merging procedure could only handle inner symbols with field
id 0, i.e. inner symbols which targeted the entire operation. If an
operation with a field-specific inner symbol was merged into another
operation, we would end up dropping the inner symbol. This change
properly handles inner symbols with many symbol definitions when merging
operations by properly merging the inner symbol property list.

### DIFF
--- a/test/Dialect/FIRRTL/dedup.mlir
+++ b/test/Dialect/FIRRTL/dedup.mlir
@@ -161,12 +161,12 @@ firrtl.circuit "Annotations" {
     
     // Should deduplicate wires where only one has a symbol
     // CHECK: %h = firrtl.wire sym @h : !firrtl.uint
-    // CHECK: %i = firrtl.wire sym @sym : !firrtl.uint
+    // CHECK: %i = firrtl.wire sym @p : !firrtl.uint
     %h = firrtl.wire sym @h : !firrtl.uint
     %i = firrtl.wire : !firrtl.uint
 
     // Should merge multiple different symbols.
-    // CHECK: %j = firrtl.wire sym [<@sym_0,0,public>, <@j,1,private>]
+    // CHECK: %j = firrtl.wire sym [<@q,0,private>, <@j,1,private>] : !firrtl.bundle<a: uint<1>>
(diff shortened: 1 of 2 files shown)


rwy7 (approved):

test/Dialect/FIRRTL/dedup.mlir
```diff
@@ -158,6 +158,16 @@ firrtl.circuit "Annotations" {
...
     %g = firrtl.wire {annotations = [{circt.fieldID = 1 : i32, class = "subanno"}]} : !firrtl.bundle<a: uint<1>>
+    
+    // Should deduplicate wires where only one has a symbol
+    // CHECK: %h = firrtl.wire sym @h : !firrtl.uint
+    // CHECK: %i = firrtl.wire sym @sym : !firrtl.uint
+    %h = firrtl.wire sym @h : !firrtl.uint
+    %i = firrtl.wire : !firrtl.uint
+
+    // Should merge multiple different symbols.
+    // CHECK: %j = firrtl.wire sym [<@sym_0,0,public>, <@j,1,private>]
```
youngar, line 169: 🐛 this symbol went from private to public

youngar:
I pushed a second commit to this to properly merge inner symbols with multiple properties.

PR #8425, opened 2025-04-17
Comb Interval Range Analysis and Comb Opt Narrowing pass 

cowardsa:
Building on the existing MLIR integer interval range analysis framework, build the interface for the comb dialect. Use the interval range analysis to develop a comb opt narrowing pass, that reduces a comb opt based on the interval range of the operation. Currently this is only supported for addition, subtraction and multiplication, but this could be extended in the future. Future work will leverage the interval analysis to validate the combination of consecutive addition operators into a multi-operand addition.


lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,256 @@
...
+  auto urange = ConstantIntRanges::fromUnsigned(umin, umax);
+  setResultRange(getResult(), urange);
+};
+
+//===----------------------------------------------------------------------===//
+// ReplicateOp
+//===----------------------------------------------------------------------===//
+void comb::ReplicateOp::inferResultRanges(ArrayRef<ConstantIntRanges> argRanges,
+                                          SetIntRangeFn setResultRange) {
+  // Compute concat as an unsigned integer of bits
```
TaoBi22, line 194:
```suggestion
  // Compute replicate as an unsigned integer of bits
```

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,256 @@
...
+  APInt max = APInt::getAllOnes(1);
+
+  std::optional<bool> truthValue = intrange::evaluatePred(pred, lhs, rhs);
+  if (truthValue.has_value() && *truthValue)
+    min = max;
+  else if (truthValue.has_value() && !(*truthValue))
+    max = min;
+
+  setResultRange(getResult(), ConstantIntRanges::fromUnsigned(min, max));
+}
```
TaoBi22, line 256: nit: missing EOF newline

lib/Dialect/Comb/Transforms/IntRangeOptimizations.cpp
```diff
@@ -0,0 +1,145 @@
+//===- RangeAnalysisComb.cpp - Lower some ops in comb -------------------*- C++
+//-*-===//
```
TaoBi22, line 2: Weird header line wrap happening here?

lib/Dialect/Comb/Transforms/IntRangeOptimizations.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace circt;
+using namespace circt::comb;
+using namespace mlir;
+using namespace mlir::dataflow;
+
+namespace circt {
+namespace comb {
+#define GEN_PASS_DEF_COMBINTRANGENARROWING
```
TaoBi22, line 30: Could this be changed to a name that matches the file name?

lib/Dialect/HW/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,30 @@
+//===- InferIntRangeInterfaceImpls.cpp - Integer range impls for arith -===//
```
TaoBi22, line 1: Should be padded to 80 chars as above

lib/Dialect/HW/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include "llvm/Support/Debug.h"
+#include <optional>
+
+#define DEBUG_TYPE "int-range-analysis"
+
+using namespace mlir;
+using namespace mlir::intrange;
+using namespace circt;
+using namespace circt::hw;
```
TaoBi22, line 21: Is this necessary? Doesn't look to me like you ever implicitly refer to anything in `hw`

lib/Dialect/HW/Transforms/CMakeLists.txt
```diff
@@ -20,4 +20,5 @@ add_circt_dialect_library(CIRCTHWTransforms
   MLIRIR
   MLIRPass
   MLIRTransformUtils
+  MLIRInferIntRangeCommon
```
TaoBi22, line 23: How come this is necessary? You don't seem to be touching anything else in `lib/Dialect/HW/Transforms`

test/Dialect/Comb/comb-int-range-narrowing.mlir
```diff
@@ -0,0 +1,113 @@
...
+  %0 = comb.concat %c0_i9, %a : i9, i8
+  %1 = comb.concat %c0_i8, %b : i8, i9
+  %2 = comb.and %0, %1 : i17
+  %3 = comb.concat %false, %2 : i1, i17
+  %4 = comb.concat %c0_i8, %c : i8, i10
+  %5 = comb.or %3, %4 : i18
+  %6 = comb.concat %c0_i2, %d : i2, i16
+  %7 = comb.add %5, %6 : i18
+  hw.output %7 : i18
+}
```
TaoBi22, line 113: nit: missing EOF newline as above

test/Dialect/Comb/comb-int-range-narrowing.mlir
```diff
@@ -0,0 +1,113 @@
+// RUN: circt-opt %s --comb-int-range-narrowing --canonicalize | FileCheck %s
+
+// CHECK-LABEL: @basic_csa
+hw.module @basic_csa(in %a : i1, in %b : i1, in %c : i1, out add_abc : i3) {
+// CHECK-NEXT: %false = hw.constant false
+// CHECK-NEXT: %0 = comb.concat %false, %a : i1, i1
```
TaoBi22, line 6: To make tests less fragile, we try to avoid hardcoding the SSA names of values into tests and use regex format and labelling instead - looking through existing tests should help figure this out, randomly picked example [here](https://github.com/llvm/circt/blob/main/test/Conversion/AIGToComb/aig-to-comb.mlir) - let me know if that's unclear at all!

test/Dialect/Comb/comb-int-range-narrowing.mlir
```diff
@@ -0,0 +1,113 @@
+// RUN: circt-opt %s --comb-int-range-narrowing --canonicalize | FileCheck %s
```
TaoBi22, line 1: Same point as above re: where this actually belongs

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,256 @@
+//===- InferIntRangeInterfaceImpls.cpp - Integer range impls for comb -===//
```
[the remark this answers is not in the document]
TaoBi22: Also, when creating an analysis we often create a test pass in [this file](https://github.com/llvm/circt/blob/main/lib/Analysis/TestPasses.cpp) that just runs the analysis and prints out the results in a checkable format, and then use this to create some FileCheck tests for the analysis. It would be nice to do that here too (and if this moves to Analysis then you could just add it in that file)

TaoBi22 (commented):
This looks pretty good to me! Mostly just nits, the only big thing is where this int range analysis should live (expanded on below). @uenoku since this is along the lines of synthesis, is there any chance you could take a quick look at it too?

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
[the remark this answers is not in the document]
cowardsa: @TaoBi22 - agree with this suggestion - there's currently no lib/Analysis/Comb hierarchy and given that I've got an implementation for both HW and Comb would it be sensible to introduce the directory structure in lib/Analysis?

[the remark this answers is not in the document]
cowardsa: Note that in MLIR these files live here say, lib/Dialect/Arith/IR/InferIntRangeInterfaceImpls.cpp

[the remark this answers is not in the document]
maerhart: I think the current location for these files is fine.
One could make an argument that this logic should be tied more to the transformation pass and less to the dialect definition in which case you could implement them as external models and move them to the `Transforms` folder (or a new `Analysis` folder). Both seem fine to me.

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,256 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Implemenation of the interval range analysis interface.
+// The overflow flags are not set for the comb operations since they is
+// no meaningful concept of overflow detection in comb.
+//
+//===----------------------------------------------------------------------===//
+#include "circt/Dialect/Comb/CombOps.h"
```
maerhart, line 14: Super-nit: to be consistent with other files in CIRCT (also in the other files)
```suggestion
//===----------------------------------------------------------------------===//

#include "circt/Dialect/Comb/CombOps.h"
```

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,256 @@
...
+};
+
+//===----------------------------------------------------------------------===//
+// ExtractOp
+//===----------------------------------------------------------------------===//
+
+void comb::ExtractOp::inferResultRanges(ArrayRef<ConstantIntRanges> argRanges,
+                                        SetIntRangeFn setResultRange) {
+  // Right-shift and truncate (trunaction implicitly handled)
+  // auto res_width = getResult().getType().getIntOrFloatBitWidth();
```
maerhart, line 181:
```suggestion
```

maerhart (commented):
Nice work! :tada:
Just added some drive-by comments.

lib/Dialect/Comb/Transforms/IntRangeOptimizations.cpp
```diff
@@ -0,0 +1,145 @@
...
+  solver.load<IntegerRangeAnalysis>();
+  if (failed(solver.initializeAndRun(op)))
+    return signalPassFailure();
+
+  RewritePatternSet patterns(ctx);
+  populateCombNarrowingPatterns(patterns, solver);
+  GreedyRewriteConfig config;
+  // We specifically need bottom-up traversal as cmpi pattern needs range
+  // data, attached to its original argument values.
+  config.useTopDownTraversal = false;
```
uenoku, line 135: HW module is a graph region there is no guarantee about traversal order (even though usually they are topo sorted).

uenoku (commented):
Thank you for interesting work! I think the interface is definitely worth to have 👍 
High-level comments:
* Is it possible to have a test for range interface decoupled from narrowing pass? I think it's a bit difficult to track the correctness of the analysis. i.e. we don't have a test for range interface tests of comb.mux etc. 
* Please use camelBack for variable names

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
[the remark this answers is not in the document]
TaoBi22: Cool, I don't have strong feelings on this so if it matches what's done upstream then feel free to ignore

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,256 @@
...
+//
+//===----------------------------------------------------------------------===//
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/Comb/CombPasses.h"
+#include "circt/Dialect/HW/HWOps.h"
+
+#include "mlir/Interfaces/InferIntRangeInterface.h"
+#include "mlir/Interfaces/Utils/InferIntRangeCommon.h"
+
+#include "llvm/Support/Debug.h"
```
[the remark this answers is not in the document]
cowardsa: Isn't InferIntRangeCommon.h needed to define inferAdd/inferSub/... - perhaps the comment is only referring to Debug.h?

include/circt/Dialect/Comb/Combinational.td
```diff
@@ -60,24 +61,24 @@ class UTVariadicOp<string mnemonic, list<Trait> traits = []> :
...
+    def SubOp : UTBinOp<"sub",   [DeclareOpInterfaceMethods<InferIntRangeInterface, ["inferResultRanges"]>]>;
   }
 }
 
-def AndOp : UTVariadicOp<"and", [Commutative]>;
-def OrOp  : UTVariadicOp<"or", [Commutative]>;
-def XorOp : UTVariadicOp<"xor", [Commutative]> {
+def AndOp : UTVariadicOp<"and", [Commutative, DeclareOpInterfaceMethods<InferIntRangeInterface, ["inferResultRanges"]>]>;
+def OrOp  : UTVariadicOp<"or",  [Commutative, DeclareOpInterfaceMethods<InferIntRangeInterface, ["inferResultRanges"]>]>;
+def XorOp : UTVariadicOp<"xor", [Commutative, DeclareOpInterfaceMethods<InferIntRangeInterface, ["inferResultRanges"]>]> {
```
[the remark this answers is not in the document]
cowardsa: Moved into several base classes - did not put in CombOp base class as not implemented for every operator - e.g. TruthTableOp is an  odd case, where we're not likely to get a lot from an interval analysis.

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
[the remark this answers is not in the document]
cowardsa: Ok will stick with as is for now - if the space of analyses grows then will perhaps make sense to separate as per Bea's suggestion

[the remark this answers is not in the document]
cowardsa: @TaoBi22  and @uenoku - have added a testing pass as per suggestion that appends the interval analysis data as an operation attribute which we can then check using Filecheck - let me know if this is suitable? This includes testing for the mux interval analysis.

lib/Dialect/Comb/Transforms/IntRangeOptimizations.cpp
TaoBi22: Marking resolved since you plan to have multiple passes in here.

lib/Dialect/Comb/Transforms/IntRangeOptimizations.cpp
cowardsa: Agreed this was legacy from a similar arith dialect pass - have removed the rewriter config

cowardsa:
@uenoku/@maerhart - have addressed all the comments above - would you have time to take another pass and approve if you feel its ready?

Major changes:

- Formatting, newlines, camelBack etc corrected throughout
- Added interval analysis test pass adding the analysis results as attributes to be checked
- Made all tests use regex throughout and made each test only stress a single pass

uenoku (approved):
Thank you for adding a test pass. Could you add a test for replicate, extract and icmp? There is a correctness issue in add/mul pattern and crash with the following case, once after the issue was fixed PR look good to me. 

```mlir
hw.module private @InitReg1(in %clock: !seq.clock, in %reset: i1, in %io_d: i32, in %io_en: i1, out io_q: i32) {
  %false = hw.constant false
  %c1_i32 = hw.constant 1 : i32
  %c0_i32 = hw.constant 0 : i32
  %reg = seq.firreg %4 clock %clock sym @__reg__ reset async %reset, %c0_i32 : i32
  %reg2 = seq.firreg %reg2 clock %clock sym @__reg2__ reset sync %reset, %c0_i32 : i32
  %reg3 = seq.firreg %reg3 clock %clock sym @__reg3__ reset async %reset, %c1_i32 : i32
  %0 = comb.concat %false, %reg : i1, i32
  %1 = comb.concat %false, %reg2 : i1, i32
  %2 = comb.add %0, %1 : i33
  %3 = comb.extract %2 from 1 : (i33) -> i32
  %4 = comb.mux bin %io_en, %io_d, %3 : i32
  hw.output %reg : i32
}
```

```
circt-opt: /scratch/hidetou/circt/llvm/llvm/lib/Support/APInt.cpp:285: int llvm::APInt::compare(const llvm::APInt &) const: Assertion `BitWidth == RHS.BitWidth && "Bit widths must be same for comparison"' failed.
PLEASE submit a bug report to https://github.com/llvm/circt and include the crash backtrace.
Stack dump:
0.      Program arguments: ./build/bin/circt-opt --comb-int-range-narrowing a.mlir
 #0 0x0000000000cdccc7 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /scratch/hidetou/circt/llvm/llvm/lib/Support/Unix/Signals.inc:804:13
 #1 0x0000000000cdac00 llvm::sys::RunSignalHandlers() /scratch/hidetou/circt/llvm/llvm/lib/Support/Signals.cpp:106:18
 #2 0x0000000000cdd5da SignalHandler(int, siginfo_t*, void*) /scratch/hidetou/circt/llvm/llvm/lib/Support/Unix/Signals.inc:0:3
 #3 0x00007ff6960e3cf0 __restore_rt (/lib64/libpthread.so.0+0x12cf0)
 #4 0x00007ff69519cacf raise (/lib64/libc.so.6+0x4eacf)
 #5 0x00007ff69516fea5 abort (/lib64/libc.so.6+0x21ea5)
 #6 0x00007ff69516fd79 _nl_load_domain.cold.0 (/lib64/libc.so.6+0x21d79)
 #7 0x00007ff695195426 (/lib64/libc.so.6+0x47426)
 #8 0x0000000000cf2ee6 (./build/bin/circt-opt+0xcf2ee6)
 #9 0x0000000002dfbf69 llvm::APInt::ult(llvm::APInt const&) const /scratch/hidetou/circt/llvm/llvm/include/llvm/ADT/APInt.h:1111:58
#10 0x0000000002dfbf69 mlir::ConstantIntRanges::rangeUnion(mlir::ConstantIntRanges const&) const /scratch/hidetou/circt/llvm/mlir/lib/Interfaces/InferIntRangeInterface.cpp:93:35
#11 0x0000000000f5907c llvm::APInt::APInt(llvm::APInt&&) /scratch/hidetou/circt/llvm/llvm/include/llvm/ADT/APInt.h:184:39
#12 0x0000000000f5907c mlir::ConstantIntRanges::ConstantIntRanges(mlir::ConstantIntRanges&&) /scratch/hidetou/circt/llvm/llvm/../mlir/include/mlir/Interfaces/InferIntRangeInterface.h:27:7
#13 0x0000000000f5907c std::_Optional_payload_base<mlir::ConstantIntRanges>::_Storage<mlir::ConstantIntRanges, false>::_Storage<mlir::ConstantIntRanges>(std::in_place_t, mlir::ConstantIntRanges&&) /sifive/tools/gcc/gcc/11.2.0/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../include/c++/11.2.0/optional:225:8
#14 0x0000000000f5907c std::_Optional_payload_base<mlir::ConstantIntRanges>::_Optional_payload_base<mlir::ConstantIntRanges>(std::in_place_t, mlir::ConstantIntRanges&&) /sifive/tools/gcc/gcc/11.2.0/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../include/c++/11.2.0/optional:116:4
#15 0x0000000000f5907c std::_Optional_payload<mlir::ConstantIntRanges, true, false, false>::_Optional_payload<mlir::ConstantIntRanges>(std::in_place_t, mlir::ConstantIntRanges&&) /sifive/tools/gcc/gcc/11.2.0/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../include/c++/11.2.0/optional:358:42
#16 0x0000000000f5907c std::_Optional_payload<mlir::ConstantIntRanges, false, false, false>::_Optional_payload<mlir::ConstantIntRanges>(std::in_place_t, mlir::ConstantIntRanges&&) /sifive/tools/gcc/gcc/11.2.0/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../include/c++/11.2.0/optional:392:57
#17 0x0000000000f5907c std::_Optional_base<mlir::ConstantIntRanges, false, false>::_Optional_base<mlir::ConstantIntRanges, false>(std::in_place_t, mlir::ConstantIntRanges&&) /sifive/tools/gcc/gcc/11.2.0/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../include/c++/11.2.0/optional:484:4
#18 0x0000000000f5907c std::optional<mlir::ConstantIntRanges>::optional<mlir::ConstantIntRanges, true>(mlir::ConstantIntRanges&&) /sifive/tools/gcc/gcc/11.2.0/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../include/c++/11.2.0/optional:707:4
#19 0x0000000000f5907c mlir::IntegerValueRange::IntegerValueRange(mlir::ConstantIntRanges) /scratch/hidetou/circt/llvm/llvm/../mlir/include/mlir/Interfaces/InferIntRangeInterface.h:117:48
#20 0x0000000000f5907c mlir::IntegerValueRange::join(mlir::IntegerValueRange const&, mlir::IntegerValueRange const&) /scratch/hidetou/circt/llvm/llvm/../mlir/include/mlir/Interfaces/InferIntRangeInterface.h:145:12
#21 0x0000000000f58f5e llvm::function_ref<void (mlir::Value, mlir::IntegerValueRange const&)>::operator()(mlir::Value, mlir::IntegerValueRange const&) const /scratch/hidetou/circt/llvm/llvm/include/llvm/ADT/STLFunctionalExtras.h:69:12
#22 0x0000000000f58f5e circt::comb::MuxOp::inferResultRangesFromOptional(llvm::ArrayRef<mlir::IntegerValueRange>, llvm::function_ref<void (mlir::Value, mlir::IntegerValueRange const&)>) /scratch/hidetou/circt/lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp:227:3
#23 0x0000000000f4d103 mlir::detail::InferIntRangeInterfaceInterfaceTraits::Model<circt::comb::MuxOp>::inferResultRangesFromOptional(mlir::detail::InferIntRangeInterfaceInterfaceTraits::Concept const*, mlir::Operation*, llvm::ArrayRef<mlir::IntegerValueRange>, llvm::function_ref<void (mlir::Value, mlir::IntegerValueRange const&)>) /scratch/hidetou/circt/build/tools/mlir/include/mlir/Interfaces/InferIntRangeInterface.h.inc:131:3
```

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+//===----------------------------------------------------------------------===//
+// ExtractOp
+//===----------------------------------------------------------------------===//
+
+void comb::ExtractOp::inferResultRanges(ArrayRef<ConstantIntRanges> argRanges,
+                                        SetIntRangeFn setResultRange) {
+  // Right-shift and truncate (trunaction implicitly handled)
+  auto lowBit = getLowBit();
+  auto umin = argRanges[0].umin().ushl_sat(lowBit);
```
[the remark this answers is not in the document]
cowardsa: Yes this is the cause of the InitReg1 bug above

lib/Dialect/Comb/Transforms/IntRangeOptimizations.cpp
```diff
@@ -0,0 +1,137 @@
...
+    if (removeWidth == 0)
+      return rewriter.notifyMatchFailure(op, "no bits to remove");
+    else if (removeWidth == opWidth)
+      return rewriter.notifyMatchFailure(
+          op, "all bits to remove - replace by zero");
+
+    // Replace operator by narrower version of itself
+    Value lhs = op.getOperand(0);
+    Value rhs = op.getOperand(1);
+
```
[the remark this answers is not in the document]
cowardsa: Have included a check for binary operators at time of writing - will extend to variadic operators in a subsequent PR, where I'll also add support for narrowing more operators. Good catch!

cowardsa:
@uenoku - have addressed your remaining comments with the following:

- Checks for binary operators in Narrowing
- Adding support in analysis for variadic ops
- Correcting the incorrect shift bug above
- Adding tests for cmp, extract and replicate

test/Analysis/comb-int-range-analysis.mlir
```diff
@@ -0,0 +1,174 @@
...
+  %2 = comb.icmp ugt %a, %b : i3
+  %3 = comb.icmp uge %a, %c0_i3 : i3
+  %4 = comb.icmp slt %a, %b : i3
+  %5 = comb.icmp sle %a, %b  : i3
+  %6 = comb.icmp sgt %a, %b : i3
+  %7 = comb.icmp sge %a, %c0_i3 : i3
+  %8 = comb.icmp eq %a, %b  : i3
+  %9 = comb.icmp ne %a, %b  : i3
+  hw.output
+}
```
uenoku, line 174:
```suggestion
}

```

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,316 @@
...
+    break;
+  case comb::ICmpPredicate::ugt:
+    pred = intrange::CmpPredicate::ugt;
+    break;
+  case comb::ICmpPredicate::uge:
+    pred = intrange::CmpPredicate::uge;
+    break;
+  default:
+    llvm_unreachable("Unknown comparison predicate");
+  }
```
uenoku, line 305: nit: Could you return in default? 
```suggestion
  
  intrange::CmpPredicate pred;
  switch (combPred) {
  case comb::ICmpPredicate::eq:
    pred = intrange::CmpPredicate::eq;
    break;
  case comb::ICmpPredicate::ne:
    pred = intrange::CmpPredicate::ne;
    break;
  case comb::ICmpPredicate::slt:
    pred = intrange::CmpPredicate::slt;
    break;
  case comb::ICmpPredicate::sle:
    pred = intrange::CmpPredicate::sle;
    break;
  case comb::ICmpPredicate::sgt:
    pred = intrange::CmpPredicate::sgt;
    break;
  case comb::ICmpPredicate::sge:
    pred = intrange::CmpPredicate::sge;
    break;
  case comb::ICmpPredicate::ult:
    pred = intrange::CmpPredicate::ult;
    break;
  case comb::ICmpPredicate::ule:
    pred = intrange::CmpPredicate::ule;
    break;
  case comb::ICmpPredicate::ugt:
    pred = intrange::CmpPredicate::ugt;
    break;
  case comb::ICmpPredicate::uge:
    pred = intrange::CmpPredicate::uge;
    break;
  default:
    // These predicates are not supported for integer range analysis
    setResultRange(getResult(), ConstantIntRanges::fromUnsigned(min, max));
    return;
  }
```

uenoku (approved):
LG, thank you for addressing comments.

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
cowardsa: Resolved!

test/Analysis/comb-int-range-analysis.mlir
cowardsa: Resolved

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+  if (combPred == comb::ICmpPredicate::ceq ||
+      combPred == comb::ICmpPredicate::cne ||
+      combPred == comb::ICmpPredicate::weq ||
+      combPred == comb::ICmpPredicate::wne) {
+    // These predicates are not supported for integer range analysis
+    setResultRange(getResult(), ConstantIntRanges::fromUnsigned(min, max));
+    return;
+  }
+
```
maerhart, line 270: This already seems to be covered in the default case below?
```suggestion
```

include/circt/Dialect/Comb/Passes.td
```diff
@@ -22,4 +22,14 @@ def LowerComb : Pass<"lower-comb"> {
...
 }
 
+def CombIntRangeNarrowing : Pass<"comb-int-range-narrowing"> {
+  let summary = "Reduce comb operations bitwidth based on integer range analysis.";
+  let description = [{
+    Compute a basic value range analysis, by propagating integer intervals through the domain.
+    The analysis is limited by a lack of sign-extension operator in the comb dialect, leading to 
+    an over-approximation. Particularly for signed arithmetic, a single interval
+    is often an over-approximation, a more precise analysis would require a union of intervals.
+  }];
```
maerhart, line 32: Pls fit to 80 cols

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
cowardsa: After the update suggested by Hideto - indeed now redundant - have removed.

cowardsa:
Any final comments @maerhart?

maerhart:
> Any final comments @maerhart?

LGTM. I only did a code style and structure review though. But Hideto gave you approval to merge. Let me know if I should merge it for you.

lib/Dialect/Comb/Transforms/IntRangeOptimizations.cpp
```diff
@@ -0,0 +1,140 @@
...
+    auto removeWidth = ranges[0].umax().countLeadingZeros();
+    for (const ConstantIntRanges &range : ranges) {
+      auto rangeCanRemove = range.umax().countLeadingZeros();
+      removeWidth = std::min(removeWidth, rangeCanRemove);
+    }
+    if (removeWidth == 0)
+      return rewriter.notifyMatchFailure(op, "no bits to remove");
+    else if (removeWidth == opWidth)
+      return rewriter.notifyMatchFailure(
+          op, "all bits to remove - replace by zero");
```
maerhart, line 80: Nit: unnecessary else after return
```suggestion
    if (removeWidth == 0)
      return rewriter.notifyMatchFailure(op, "no bits to remove");
    
    if (removeWidth == opWidth)
      return rewriter.notifyMatchFailure(
          op, "all bits to remove - replace by zero");
```

lib/Dialect/Comb/InferIntRangeInterfaceImpls.cpp
```diff
@@ -0,0 +1,309 @@
...
+//===----------------------------------------------------------------------===//
+// AddOp
+//===----------------------------------------------------------------------===//
+
+void comb::AddOp::inferResultRanges(ArrayRef<ConstantIntRanges> argRanges,
+                                    SetIntRangeFn setResultRange) {
+  auto resultRange = argRanges[0];
+  for (size_t i = 1; i < argRanges.size(); ++i)
+    resultRange =
+        inferAdd({resultRange, argRanges[i]}, intrange::OverflowFlags::None);
```
maerhart, line 33: There's a `drop_front` method that you can use for these kind of things. That's super nit picky, but just so you know.
```suggestion
  for (auto arg : argRanges.drop_front())
    resultRange =
        inferAdd({resultRange, arg}, intrange::OverflowFlags::None);
```

cowardsa: Integrated the change will close.

cowardsa:
@maerhart - all looking good and checks passing - please merge when you get the time to

PR #8428, opened 2025-04-18
[LowerToHW] [ExtractTestCode] Remove FD caching to pass ETC

uenoku:
Using initilaization for lowering file descriptor casuses probems for ETC so this commit removes initialization and call library function every time fd is necessary

### DIFF
--- a/test/firtool/print.fir
+++ b/test/firtool/print.fir
@@ -1,4 +1,5 @@
 ; RUN: firtool %s | FileCheck %s
+; RUN: firtool %s --extract-test-code | FileCheck --check-prefix=ETC %s
 
 FIRRTL version 5.1.0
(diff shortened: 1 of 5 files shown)


test/firtool/print.fir
```diff
@@ -1,4 +1,5 @@
 ; RUN: firtool %s | FileCheck %s
+; RUN: firtool %s --extract-test-code | FileCheck --check-prefix=ETC %s
```
seldridge, line 2: Great use of a second `RUN` line. 💯

seldridge (approved):
LGTM

It pains me that your elegant caching can't easily be used here. Once ETC is dead and replaced with layers, we can revisit this.

PR #8435, opened 2025-04-22
[StripDebugInfoWithPred] Use worklist instead of recursion

uenoku:
As it caused stack overflow it is changed to use worklist. The behavior is slightly different since it flattens fused loc and dedupluicates locations.


seldridge (commented):
Looks good. Thanks for the quick fix.

Do we have any tests of this already? (I couldn't find any.)

seldridge (approved):

uenoku:
Thanks, a test was added.

PR #8439, opened 2025-04-24
[OM] Add field_locs array attribute for ClassFieldsOp locations

leonardt:
This changes the `ClassFieldsOp` logic to use an inherent attribute to store an array mapping field to location by index.  This avoids the issue where `FusedLoc` merges non-unique locations (preventing us from retrieving a fields location by mapping it to an index).  It also avoids an issue where `metadata` is removed from the `FusedLoc` that was previously being used to store this information.

### DIFF
--- a/lib/Dialect/OM/OMOps.cpp
+++ b/lib/Dialect/OM/OMOps.cpp
@@ -77,18 +77,11 @@ static void printPathString(OpAsmPrinter &p, Operation *op, PathAttr path,
 static ParseResult parseFieldLocs(OpAsmParser &parser, ArrayAttr &fieldLocs) {
   if (parser.parseOptionalKeyword("field_locs"))
     return success();
-  uint count = 0;
-  auto parseElt = [&]() -> ParseResult {
-    if (count > 0) {
-      parser.emitError(parser.getCurrentLocation(),
-                       "found more than one field_locs array");
-      return failure();
-    }
-    count++;
-    return parser.parseAttribute(fieldLocs);
-  };
-  return parser.parseCommaSeparatedList(OpAsmParser::Delimiter::Paren,
-                                        parseElt);
+  if (parser.parseLParen() || parser.parseAttribute(fieldLocs) ||
+      parser.parseRParen()) {
+    return failure();
+  }
+  return success();
 }
 
 static void printFieldLocs(OpAsmPrinter &printer, Operation *op,
--- a/lib/Dialect/OM/OMOps.cpp
+++ b/lib/Dialect/OM/OMOps.cpp
@@ -447,9 +447,10 @@ void circt::om::ClassOp::addNewFieldsOp(mlir::OpBuilder &builder,
 }
 
 mlir::Location circt::om::ClassOp::getFieldLocByIndex(size_t i) {
-  auto fieldLocs = this->getFieldsOp().getFieldLocs();
+  auto fieldsOp = this->getFieldsOp();
+  auto fieldLocs = fieldsOp.getFieldLocs();
   if (!fieldLocs.has_value())
-    return UnknownLoc::get(this->getContext());
+    return fieldsOp.getLoc();
   assert(i < fieldLocs.value().size() &&
          "field index too large for location array");
   return cast<LocationAttr>(fieldLocs.value()[i]);
--- a/lib/Dialect/OM/OMOps.cpp
+++ b/lib/Dialect/OM/OMOps.cpp
@@ -498,7 +498,7 @@ void circt::om::ClassExternOp::replaceFieldTypes(AttrTypeReplacer replacer) {
 // ClassFieldsOp
 //===----------------------------------------------------------------------===//
 //
-LogicalResult circt::om::ClassFieldsOp::verify() { 
+LogicalResult circt::om::ClassFieldsOp::verify() {
   auto fieldLocs = this->getFieldLocs();
   if (fieldLocs.has_value()) {
     auto fieldLocsVal = fieldLocs.value();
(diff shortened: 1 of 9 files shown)


lib/Dialect/OM/OMOps.cpp
```diff
@@ -415,25 +424,15 @@ void circt::om::ClassOp::addNewFieldsOp(mlir::OpBuilder &builder,
...
-    assert(metadataArr && "Expected fused loc to store metadata array");
-    assert(i < metadataArr.size() &&
-           "expected index to be less than array size");
-    LocationAttr locAttr = dyn_cast<LocationAttr>(metadataArr[i]);
-    assert(locAttr && "expected metadataArr entry to be location attribute");
-    loc = Location(locAttr);
-  }
-  return loc;
+  auto fieldLocs = this->getFieldsOp().getFieldsLocs();
+  if (!fieldLocs)
```
mikeurbach, line 433: Is there a valid case where this can be None, or is this just being flexible? If I'm understanding correctly, this should be set for any valid user input (at least from FIRRTL), since there's really only one path into this IR. I guess what I'm getting at is if this should be an assertion that fieldLocs is not null, and has exactly the number of elements as there are fields, since that is the invariant we already expect the compiler implementation to maintain.

lib/Dialect/OM/OMOps.cpp
```diff
@@ -415,25 +424,15 @@ void circt::om::ClassOp::addNewFieldsOp(mlir::OpBuilder &builder,
...
+    return this->getFieldsOp().getLoc();
+  return cast<LocationAttr>(fieldLocs.value()[i]);
```
mikeurbach, line 435: This needs to check the array is the right size before doing this indexing. Should probably put that in a verifier on the operation, which checks the locations size matches the fields size, if the locations array is specified. Also see my other comment--I'm wondering if this should just be an inherent, non-optional attribute... but either way, we should probably verify the size in a verifier at least.

mikeurbach (commented):
Generally looks good, just a couple minor things from my end. I guess one question is if this should go on the ClassOp (like we do for HWModuleOp) or on the ClassFieldsOp. I don't have any particularly strong reason to go one way or another. Keeping it on ClassFieldsOp like this PR does seems reasonable, since existing code expects to get field locations from the ClassFieldsOp.

mikeurbach:
Also please update the description / commit message (really, the commit message when this merges), to explain why we want to make this change.

lib/Dialect/OM/OMOps.cpp
leonardt: We could maintain that assertion internally to the compiler, but it could be possible for some input `.mlir`, e.g. writing tests directly in the `om` dialect, to omit the optional attr.  If we implement a custom parser that fills that the attribute and make it non optional, then we could change this to an assertion/assumption.

lib/Dialect/OM/OMOps.cpp
leonardt: Makes sense, yea I think since we likely want to omit the location info unless the flag is set, implementing a custom parse/print will allow us to enforce the invariant that the array exists and is the right size.  We would want to verify that no one changes that in a pass.

leonardt:
Re: putting this in `ClassOp`, since we may need to implement a custom parser/printer for this, there could be motivation to just hook onto the existing parser/printer for `ClassOp`, but I think since it's information associated with the `fields` op it does make sense to store it with the op.

lib/Dialect/OM/OMOps.cpp
mikeurbach: Ok fair enough, I think it's up to you. IMHO it would be fine to say this is required, if we want it to be, and tests would have to specify it too. On the other hand, it really is optional information, and leaving it as such makes sense to me. Personally I'd prefer to keep things simple, but if we're adding a custom parser/printer to gate this behind the "should print debug info" flag, then I guess we could have it be required and let the parser fill in unknowns in the case we're taking in .mlir text in tests, etc.

lib/Dialect/OM/OMOps.cpp
mikeurbach: Fair enough, I'm mainly thinking we should have a check before we index into the array.

lib/Dialect/OM/OMOps.cpp
```diff
@@ -74,6 +74,33 @@ static void printPathString(OpAsmPrinter &p, Operation *op, PathAttr path,
...
+  if (parser.parseOptionalKeyword("field_locs"))
+    return success();
+  uint count = 0;
+  auto parseElt = [&]() -> ParseResult {
+    if (count > 0) {
+      parser.emitError(parser.getCurrentLocation(),
+                       "found more than one field_locs array");
+      return failure();
+    }
+    count++;
```
mikeurbach, line 87: nit: we generally prefer pre-increment: https://llvm.org/docs/CodingStandards.html#prefer-preincrement

lib/Dialect/OM/OMOps.cpp
```diff
@@ -475,6 +504,18 @@ void circt::om::ClassExternOp::replaceFieldTypes(AttrTypeReplacer replacer) {
...
 
+//===----------------------------------------------------------------------===//
+// ClassFieldsOp
+//===----------------------------------------------------------------------===//
+//
+LogicalResult circt::om::ClassFieldsOp::verify() { 
+  auto fieldLocs = this->getFieldLocs();
+  if (fieldLocs.has_value())
+    if (fieldLocs.value().size() != this->getFields().size())
+      return this->emitOpError("size of field_locs does not match number of fields");
```
mikeurbach, line 515: It might be helpful to print out the size of each, or even printing out the actual fieldLocs attributes and the field names, or something like that. Up to you.

lib/Dialect/OM/OMOps.cpp
mikeurbach: Also- this parseElt is really just so we can use parseCommaSeparatedList, but we actually just want something surrounded by parentheses right? It seems like we actually error if there are more than 1 value separated by commas, which seems a little weird. Maybe just something like:

```
if (parseLParen() || parseAttribute(fieldLocs) || parseRParen())
```

mikeurbach (approved):
A couple more minor comments. I think this would be fine to go in now, but a couple minor niceties that could help.

uenoku (approved):
LG

PR #8443, opened 2025-04-24
[circt-verilog] Add --verbose-pass-executions option

fabianschuiki:
Add the usual `--verbose-pass-executions` option to circt-verilog and add the corresponding pass manager instrumentation if the option is set.


seldridge (approved):
Nic.e

PR #8444, opened 2025-04-24
[SSP] Separate `ResourceType` from `OperatorType`

jiahanxie353:
This patch separates the concept of `ResourceType` from `OperatorType`, so that we can associate different kinds of resources and their constraints to a single operation.


jiahanxie353:
P.S. we have not supported adding multiple, heterogeneous `resource` to an `operation` yet - current patch only makes sure that the introduction of `resource` does not break the current test cases.

test/Dialect/SSP/Transforms/print.mlir
```diff
@@ -5,6 +5,7 @@ ssp.instance @Foo of "Problem" {
   library {
     operator_type @Bar [latency<1>]
   }
+  resource {}
```
andrewb1999, line 8: Can we change the printing so the resource section does not print if it is empty?

test/Dialect/SSP/roundtrip.mlir
```diff
@@ -142,64 +166,80 @@ ssp.instance @mco_outgoing_delays of "ChainingProblem" {
...
+// CHECK:    resource_type @slowAdder [limit<1>]
+// CHECK:    resource_type @fastAdder [limit<1>]
+// CHECK:  }
 // CHECK:   graph {
-// CHECK:     %[[op_0:.*]] = operation<@slowAdd>() [t<0>]
-// CHECK:     %[[op_1:.*]] = operation<@slowAdd>() [t<1>]
-// CHECK:     %[[op_2:.*]] = operation<@fastAdd>() [t<0>]
-// CHECK:     %[[op_3:.*]] = operation<@slowAdd>() [t<1>]
-// CHECK:     %[[op_4:.*]] = operation<@fastAdd>() [t<1>]
+// CHECK:     %[[op_0:.*]] = operation<@slowAdd>() [rsrc<@slowAdder>, t<0>]
```
andrewb1999, line 179: I think the used rsrc should be a first class property of the operation, similar to how the operator type is.

test/Dialect/SSP/roundtrip.mlir
```diff
@@ -142,64 +166,80 @@ ssp.instance @mco_outgoing_delays of "ChainingProblem" {
...
     operator_type @Implicit [latency<0>]
   }
+  resource {
+    resource_type @ReadPort [limit<1>]
+    resource_type @WritePort [limit<1>]
+  }
   graph {
-    %0 = operation<@MemPort> @load_A(@store_A [dist<1>]) [t<2>]
-    %1 = operation<@MemPort> @load_B() [t<0>]
+    %0 = operation<@MemAccess> uses<@ReadPort>@load_A(@store_A [dist<1>]) [t<2>]
```
andrewb1999, line 239: We should change uses to take an array of resources and for schedulers that do not support multiple resource types yet error out.

test/Dialect/SSP/roundtrip.mlir
```diff
@@ -142,64 +166,80 @@ ssp.instance @mco_outgoing_delays of "ChainingProblem" {
...
+    %1 = operation<@MemAccess> uses<@ReadPort>@load_B() [t<0>]
```
andrewb1999, line 240: There should be a space between the resource types and the symbol name (@load_B)

andrewb1999:
To provide some additional context to other reviewers, the ability to have multiple resource types per operation is absolutely essential for practical HLS.

test/Dialect/SSP/Transforms/print.mlir
jiahanxie353: But the empty `library`/`graph` are still being printed:
https://github.com/llvm/circt/blob/c07d31b33912b2b33d36f1f8f6312b0a4936137a/test/Dialect/SSP/roundtrip.mlir#L8-L11

test/Dialect/SSP/roundtrip.mlir
jiahanxie353: makes sense!

jiahanxie353:
Thanks for the review, @andrewb1999 ! I have addressed the feedback you gave.

Moreover, we can now associate multiple resource types to a single operation now!

include/circt/Dialect/SSP/SSPOps.td
```diff
@@ -24,25 +24,30 @@ def InstanceOp : SSPOp<"instance",
...
+      }
       graph {
-        %0 = operation<@MemPort> @load_A(@store_A [dist<1>]) [t<2>]
-        %1 = operation<@MemPort> @load_B() [t<0>]
-        %2 = operation<@Add> @add(%0, %1) [t<3>]
-        operation<@MemPort> @store_A(%2) [t<4>]
+        %0 = operation<@Memory> @load_A uses[@ReadPort](@store_A [dist<1>]) [t<2>]
+        %1 = operation<@Memory> @load_B uses[@ReadPort]() [t<0>]
+        %2 = operation<@Add> @add(%0, %1) [t<3>]  // no `resource_type` needed
+        operation<@Memory> @store_A uses[@WritePort](%2) [t<4>]
```
andrewb1999, line 50: I think the "uses" syntax should go after the arguments. So something like: `operation<@Memory> @store_A(%2) uses[@WritePort] [t<4>]`

jopperm:
> I think it would be good if someone who has worked on the scheduling codebase more recently could give some feedback as well before merging.

@7FM maybe?

I'll have a look on Tuesday (not at my desk currently).

jiahanxie353:
> One more minor comment, but overall LGTM. I think it would be good if someone who has worked on the scheduling codebase more recently could give some feedback as well before merging.

Thank you for your review @andrewb1999 ! I will change it accordingly

> I'll have a look on Tuesday (not at my desk currently).

If you could take a look, that'll be awesome @jopperm I'd really appreciate your input :)

7FM:
> > I think it would be good if someone who has worked on the scheduling codebase more recently could give some feedback as well before merging.
> 
> @7FM maybe?

I just skimmed over the changes, LGTM so far. I am a huge fan of a clear separation between operator types and resources (especially being able to assign multiple resources to an operation)! This will certainly clean up my downstream resource sharing code base.

The only thing I was wondering about while skimming the change was the purpose of adding wrapper structs for StringAttr as Operator/ResourceType: https://github.com/llvm/circt/pull/8444/files#diff-0db8e1c62706227aae263d37b3263f7c010ca77640a1329f3980bf83e1379f3aR98-R140

Right now, this does not seem to add functionality, but adds some [boilerplate](https://github.com/llvm/circt/pull/8444/files#diff-0db8e1c62706227aae263d37b3263f7c010ca77640a1329f3980bf83e1379f3aR577-R626). Could you give a rationale for adding them? Is it to provide a common interface for Resource- and OperatorTypes to simplify changing the underlying classes for downstream users? (I don't think I have tried/needed this yet, but maybe you could give an example use case if that was indeed the intent)

jiahanxie353:
> I just skimmed over the changes, LGTM so far. I am a huge fan of a clear separation between operator types and resources (especially being able to assign multiple resources to an operation)! This will certainly clean up my downstream resource sharing code base.
> 
> The only thing I was wondering about while skimming the change was the purpose of adding wrapper structs for StringAttr as Operator/ResourceType: https://github.com/llvm/circt/pull/8444/files#diff-0db8e1c62706227aae263d37b3263f7c010ca77640a1329f3980bf83e1379f3aR98-R140
> 
> Right now, this does not seem to add functionality, but adds some [boilerplate](https://github.com/llvm/circt/pull/8444/files#diff-0db8e1c62706227aae263d37b3263f7c010ca77640a1329f3980bf83e1379f3aR577-R626). Could you give a rationale for adding them? Is it to provide a common interface for Resource- and OperatorTypes to simplify changing the underlying classes for downstream users? (I don't think I have tried/needed this yet, but maybe you could give an example use case if that was indeed the intent)

Re.:
> Right now, this does not seem to add functionality, but adds some [boilerplate](https://github.com/llvm/circt/pull/8444/files#diff-0db8e1c62706227aae263d37b3263f7c010ca77640a1329f3980bf83e1379f3aR577-R626).

You are right, right now, my focus is to make sure the smooth transition from using `OperatorType` soley to separating to `OperatorType` and `ResourceType`, so I did was to just to make sure the separation design makes sense and it should not break the current CI. We will add more interesting features later.

Re.:
> Could you give a rationale for adding them?

Yes, there is a rationale:
1. It enables type-level distinction between conceptually different things
  - Even though both types are backed by a `StringAttr`, operator types and resource types represent different scheduling concepts, as we all agree.
  - Without wrappers, we'd risk using a `StringAttr` intended for one concept in place of another.
2. And it provides a foundation for future extensions
  - We will have more interesting scheduling cases and `ResourceType` will then actually be different from `OperatorType`, probably by adding different utility functions to different types, attach richer metadata etc.

Please let me know if you have any question. Thanks for your review and feedback, it's really helpful :)

jopperm (commented):
Thanks for working on this -- I agree this will be super useful and looks very neat already.

Have you considered tying the resources to the operator types instead of the operations?

jiahanxie353:
> Thanks for working on this -- I agree this will be super useful and looks very neat already.
> 
> Have you considered tying the resources to the operator types instead of the operations?

Sorry for the late reply, I got seriously blocked by something over the week.

I did not consider until you ask, and I think this is a really good question. Here is what I think after some pondering, and I will answer the question below together since I think you are trying to probe the same thing?
> Have you considered using just one shared container for operator types and resources?

My understanding is I feel like separating them both conception-wise and implementation-wise/container-wise is more neat, because operator types represents _what_ an operation _does_ and resource types tells us _where_ or _how_ the operation is being run.

Since there will be the same operator using different resource types, or the same resource type is shared across operators, combing the operator type and resource type container altogether will lead to duplicate resource declaration in different containers.

And correct me if I'm wrong, if we tie resource types to the operator type (by which I understand your proposal as instead of putting operation in the first hierarchy and both resource/operator type on the second hierarchy, tying resource types to operator types means that resource type will be on the third hierarchy?), then resource types cannot be shared across problems.

Thanks again for your review and valuable insight @jopperm and please let me know if there's any further concern :)

include/circt/Scheduling/Problems.h
```diff
@@ -256,14 +328,16 @@ class Problem {
   virtual PropertyStringVector getProperties(OperatorType opr);
   virtual PropertyStringVector getProperties();
 
+  virtual PropertyStringVector getResourceProperties(ResourceType rsrc);
+  virtual PropertyStringVector getResourceProperties();
```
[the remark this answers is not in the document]
jiahanxie353: Ahh it's useless, thanks for catching!

include/circt/Scheduling/Problems.h
```diff
@@ -256,14 +328,16 @@ class Problem {
   virtual PropertyStringVector getProperties(OperatorType opr);
   virtual PropertyStringVector getProperties();
 
+  virtual PropertyStringVector getResourceProperties(ResourceType rsrc);
```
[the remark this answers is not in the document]
jiahanxie353: Ohhh initially I made `ResourceType` also an alias to `StringAttr`, which makes `ResourceType` to have the same type as `OperatorType`, so I had to make the `getProperties` distinguished by different names otherwise the compiler will complain.
Now since we are using different `struct` types, we can just overload it.

jopperm:
Thanks for clarifying!

> Since there will be the same operator using different resource types [...]

I think that's the point I didn't consider initially. Then it makes sense to set operator type and resources separately per operation. What's your canonical example for this? Something like "memory read port" as the operator type, and "memory_XYZ" as the resource?

jiahanxie353:
> I think that's the point I didn't consider initially. Then it makes sense to set operator type and resources separately per operation. What's your canonical example for this? Something like "memory read port" as the operator type, and "memory_XYZ" as the resource?

Yeah something like that, I was thinking to have `memory_read` as the operator type since "read" is a verb, just like an operator to perform; and `memory_read_port` as the resource since "ports" are resources.

lib/Scheduling/SimplexSchedulers.cpp
```diff
@@ -923,21 +928,28 @@ LogicalResult SharedOperatorsSimplexScheduler::schedule() {
...
 
   for (auto *op : limitedOps) {
-    auto opr = *prob.getLinkedOperatorType(op);
-    unsigned limit = prob.getLimit(opr).value_or(0);
+    auto maybeRsrcs = prob.getLinkedResourceTypes(op);
+    assert(maybeRsrcs && "Limited operation must have linked resource types");
+
+    auto &rsrcs = *maybeRsrcs;
+    assert(rsrcs.size() == 1 &&
+           "Multi-resource operations are not yet supported by this scheduler");
```
jopperm, line 942: Please address this in a follow-up PR. We should have at least one complete, reference scheduler implementation for each problem defined in-tree. 🙏

jopperm (approved):
LGTM!

Can you please update the [documentation](https://github.com/llvm/circt/blob/main/docs/Scheduling.md) in a follow-up PR as well, please?

lib/Scheduling/SimplexSchedulers.cpp
jiahanxie353: Yes, no problem - this is actually an already ongoing patch!

PR #8445, opened 2025-04-25
[ci] Bump CI images

seldridge:
Bump the CI images to use Ubuntu 24.04.


fabianschuiki:
I think we might have bumped the CI images since this PR was opened, so it might no longer be necessary?

PR #8447, opened 2025-04-25
[Moore] Adding `moore.string_cmp` Op

AnnuCode:
Adding a new Op to handle operations involving `!moore.string` operands per the rules mentioned in IEEE Std 1800-2012 (page 108).
With the proposed code changes, when one operand is of `!moore.string` type and the other is a string literal, the string literal gets implicitly converted to `!moore.string` type for the comparison. Thus, all the scenarios are covered:
- Both operands are of `!moore.string` type
- One operand is of the `!moore.string` type, and the other is a string literal
- Both operands are string literals (the operator is the same comparison operator as for integral types)

Screenshot 2025-04-26 043055

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -626,7 +626,7 @@ def UArrayCmpPredicateAttr : I64EnumAttr<
 def StringCmpPredicateAttr : I64EnumAttr<
     "StringCmpPredicate", "",
     [
-     I64EnumAttrCase<"eq", 0>,
+      I64EnumAttrCase<"eq", 0>,
       I64EnumAttrCase<"ne", 1>,
       I64EnumAttrCase<"lt", 2>,
       I64EnumAttrCase<"le", 3>,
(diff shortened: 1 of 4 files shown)


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -623,6 +623,19 @@ def UArrayCmpPredicateAttr : I64EnumAttr<
   let cppNamespace = "circt::moore";
 }
 
+def StringCmpPredicateAttr : I64EnumAttr<
+    "StringCmpPredicate", "",
+    [
+     I64EnumAttrCase<"eq", 0>,
```
fabianschuiki, line 629: Small indentation tweak:
```suggestion
      I64EnumAttrCase<"eq", 0>,
```

fabianschuiki (approved):
Very nice, thanks for adding this! :partying_face:

fabianschuiki:
Thanks a lot @AnnuCode! Going to land this now :+1:

AnnuCode:
Thank you, @fabianschuiki! 🙂

PR #8448, opened 2025-04-27
[SCFToCalyx] Should not write to file when json data is empty

jiahanxie353:
We should not write to file when json data is empty.


jiahanxie353:
Merging it since the CI has passed and this is a very simple change.

PR #8452, opened 2025-04-29
[OM] Add test for nested lists in ListConcatOp evaluator.

mikeurbach:
This is just adding a test that we can concatenate lists of lists and
return the right result in the Evaluator.

### DIFF
--- a/unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
+++ b/unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
@@ -1194,6 +1194,73 @@ TEST(EvaluatorTests, ListConcatField) {
                    .getValue());
 }
 
+TEST(EvaluatorTests, ListOfListConcat) {
+  StringRef mod = "om.class @ListOfListConcat()  -> (result: "
+                  "    !om.list<!om.list<!om.string>>) {"
+                  "  %0 = om.constant \"foo\" : !om.string"
+                  "  %1 = om.constant \"bar\" : !om.string"
+                  "  %2 = om.constant \"baz\" : !om.string"
+                  "  %3 = om.constant \"qux\" : !om.string"
+                  "  %4 = om.list_create %0, %1 : !om.string"
+                  "  %5 = om.list_create %4 : !om.list<!om.string>"
+                  "  %6 = om.list_create %2, %3 : !om.string"
+                  "  %7 = om.list_create %6 : !om.list<!om.string>"
+                  "  %8 = om.list_concat %5, %7 : <!om.list<!om.string>>"
+                  "  om.class.fields %8 : !om.list<!om.list<!om.string>> "
+                  "}";
+
+  DialectRegistry registry;
+  registry.insert<OMDialect>();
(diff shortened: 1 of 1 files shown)


mikeurbach:
@uenoku do you mind taking a look at this one? It was an oversight in the initial list concat roll out.

unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
```diff
@@ -1194,6 +1194,67 @@ TEST(EvaluatorTests, ListConcatField) {
...
+                  "  %0 = om.constant \"foo\" : !om.string"
+                  "  %1 = om.constant \"bar\" : !om.string"
+                  "  %2 = om.constant \"baz\" : !om.string"
+                  "  %3 = om.constant \"qux\" : !om.string"
+                  "  %4 = om.list_create %0, %1 : !om.string"
+                  "  %5 = om.list_create %4 : !om.list<!om.string>"
+                  "  %6 = om.list_create %2, %3 : !om.string"
+                  "  %7 = om.list_create %6 : !om.list<!om.string>"
+                  "  %8 = om.list_concat %5, %7 : <!om.list<!om.string>>"
+                  "  om.class.fields %8 : !om.list<!om.list<!om.string>> "
```
leonardt, line 1209: It seems like `%8` is a a list of lists of strings, but below grabbing the result via the evaluator returns a flat list.  Is this the expected behavior for the semantics of the evaluator? My intuition would've expected a nested list value, so curious why it's being flattened.

lib/Dialect/OM/Evaluator/Evaluator.cpp
```diff
@@ -621,8 +647,7 @@ circt::om::Evaluator::evaluateListConcat(ListConcatOp op,
 
     // Append each EvaluatorValue from the sublist.
     evaluator::ListValue *subList = extractList(result.value().get());
-    for (auto subValue : subList->getElements())
-      values.push_back(subValue);
+    extractElements(subList);
```
leonardt, line 650: Does this assume that the concat argument is always a list of sublists? So this doesn't support concatenating two "depth 1" lists?

unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
mikeurbach: oh hmm you are right, i was trying to write a test for concatenating lists of lists, but i do not want to actually flatten away that structure. let me take another look at this.

mikeurbach: Thanks for the review, I was way off--this was actually working correctly, and I was mixing up what the test expectation should be. I've update the test, and I guess I will just merge it since it doesn't hurt to have more tests.

lib/Dialect/OM/Evaluator/Evaluator.cpp
mikeurbach: yeah, this was just wrong.

PR #8453, opened 2025-04-29
[Arc] Make `isInput` in `ImportedValue` (in SplitLoops) a bool

TaoBi22:
`isInput` is defined as an unsigned, but is documented as if it's a bool and is only ever written or read as a bool. Guessing this is not deliberate, but let me know if it is!


maerhart (approved):
LGTM

PR #8454, opened 2025-04-29
[Comb] Drop redundant concat verifier

### LINKED ISSUE #8394: [Comb] Redundant concat op verifier
The `comb.concat` operation implements the `InferTypeOpInterface` which specifies a verifier based on the `inferReturnType` method. However, we also have a custom verifier that checks the same thing. We should remove the custom verifier for performance reasons.

Thanks @Hatsunespica for making us aware of this in https://github.com/llvm/llvm-project/pull/131480#issuecomment-2779551689

TaoBi22:
Closes #8394. Also adds a test to check inference is as expected.

### DIFF
--- a/test/Dialect/Comb/errors.mlir
+++ b/test/Dialect/Comb/errors.mlir
@@ -13,9 +13,20 @@ hw.module @err(in %a: i1, in %b: i1) {
 }
 
 // -----
+
 hw.module @err(in %a: i4, out out: i7) {
   // expected-note @+1 {{prior use here}}
   %0 = comb.concat %a, %a : i4, i4
   // expected-error @+1 {{use of value '%0' expects different type than prior uses: 'i7' vs 'i8'}}
+  hw.output %0 : i7
+}
+
+// -----
+
+// Check return type is still inferred correctly with generic parser
+hw.module @err(in %a: i4, out out: i7) {
+  // expected-error @+2 {{'comb.concat' op inferred type(s) 'i8' are incompatible with return type(s) of operation 'i7'}}
+  // expected-error @+1 {{'comb.concat' op failed to infer returned types}}
+  %0 = "comb.concat"(%a, %a) : (i4, i4) -> i7
   hw.output %0 : i7
 }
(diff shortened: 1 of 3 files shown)


test/Dialect/Comb/errors.mlir
```diff
@@ -11,3 +11,11 @@ hw.module @err(in %a: i1, in %b: i1) {
...
   %0 = comb.truth_table %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b, %a, %b -> [false]
 }
+
+// -----
+hw.module @err(in %a: i4, in %b: i7) {
+  // expected-note @+1 {{prior use here}}
+  %0 = comb.concat %a, %a : i4, i4
+  // expected-error @+1 {{use of value '%0' expects different type than prior uses: 'i7' vs 'i8'}}
+  %1 = comb.and %0, %b : i7
+}
```
maerhart, line 21: Super-Nit: could specify b as an output port and do `hw.output %0` to not involve any other comb ops than concat.

I think what we also want to test is that the return type is not just set by the parser, but the property is also verified when using the generic format or C++ builder function.

maerhart (approved):
LGTM. Thanks for doing this!

test/Dialect/Comb/errors.mlir
TaoBi22: > Super-Nit: could specify b as an output port and do `hw.output %0` to not involve any other comb ops than concat.

smart :brain:

TaoBi22: > I think what we also want to test is that the return type is not just set by the parser, but the property is also verified when using the generic format or C++ builder function.

Added a test in generic syntax!

PR #8455, opened 2025-04-30
[CalyxEmitter] Negative, infinite, and NaN floating point values emitted as bitvectors

jiahanxie353:
This patch emits negative, infinite, and NaN floating point values as bitvectors because Calyx native compiler does not support parsing negative values like `std_float_const(32, -4.2)` nor any non-digits like `std_float_const(32, inf)`.

### DIFF
--- a/lib/Dialect/Calyx/Export/CalyxEmitter.cpp
+++ b/lib/Dialect/Calyx/Export/CalyxEmitter.cpp
@@ -1104,6 +1104,21 @@ void Emitter::emitAssignment(AssignOp op) {
     emitValue(op.getGuard(), /*isIndented=*/false);
     os << questionMark();
   }
+  if (auto constantOp =
+          dyn_cast_or_null<calyx::ConstantOp>(op.getSrc().getDefiningOp())) {
+    TypedAttr attr = constantOp.getValueAttr();
+    assert(isa<FloatAttr>(attr) && "must be a floating point constant");
+    auto fltAttr = cast<FloatAttr>(attr);
+    APFloat value = fltAttr.getValue();
+    if (value.isInfinity() || value.isNaN() || value.isNegative()) {
+      SmallString<8> str;
+      auto intValue = value.bitcastToAPInt();
+      intValue.toStringUnsigned(str, /*Radix=*/2);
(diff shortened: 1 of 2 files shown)


lib/Dialect/Calyx/Export/CalyxEmitter.cpp
```diff
@@ -1104,6 +1104,21 @@ void Emitter::emitAssignment(AssignOp op) {
     emitValue(op.getGuard(), /*isIndented=*/false);
     os << questionMark();
   }
+  if (auto constantOp =
+          dyn_cast_or_null<calyx::ConstantOp>(op.getSrc().getDefiningOp())) {
+    TypedAttr attr = constantOp.getValueAttr();
+    assert(isa<FloatAttr>(attr) && "must be a floating point constant");
+    auto fltAttr = cast<FloatAttr>(attr);
```
cgyurgyik, line 1111: Nit: use dyn_cast
```suggestion
    auto fltAttr = dyn_cast<FloatAttr>(attr);
    assert(attr != nullptr && "must be a floating point constant");
```

test/Dialect/Calyx/emit.mlir
```diff
@@ -614,3 +614,33 @@ module attributes {calyx.entrypoint = "main"} {
     }
   } {toplevel}
 }
+
+// -----
+
+// Emit negative floating point values
+
+module attributes {calyx.entrypoint = "main"} {
+  calyx.component @main(%clk: i1 {clk}, %reset: i1 {reset}, %go: i1 {go}) -> (%done: i1 {done}) {
```
cgyurgyik, line 623: Nit: Can we make this test just be 3 constants (one for a negative number, infinity, and a NaN) and as little control flow as possible?

cgyurgyik (approved):

test/Dialect/Calyx/emit.mlir
jiahanxie353: Sure, and done!

jiahanxie353:
Merging it as the CI has passed, thanks for your review @cgyurgyik !

PR #8458, opened 2025-05-01
[FIRRTL] Emit a bindfile for each public module

rwy7:
According to the FIRRTL ABI, we should be emitting a bindfile for each layer × public-module combination. Currently, we only emit one bindfile per layer, which enables the layer for the entire circuit.

This PR updates lower layers to correctly implement the FIRRTL ABI by emitting bindfiles for each public module.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -9,6 +9,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "circt/Dialect/Emit/EmitOps.h"
+#include "circt/Dialect/FIRRTL/FIRRTLInstanceGraph.h"
 #include "circt/Dialect/FIRRTL/FIRRTLOps.h"
 #include "circt/Dialect/FIRRTL/FIRRTLUtils.h"
 #include "circt/Dialect/FIRRTL/Namespace.h"
@@ -18,6 +20,7 @@
 #include "circt/Dialect/SV/SVOps.h"
 #include "circt/Support/Utils.h"
 #include "mlir/Pass/Pass.h"
+#include "llvm/ADT/PostOrderIterator.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/Mutex.h"
 
@@ -57,7 +60,6 @@ enum class Delimiter { BindModule = '_', BindFile = '-', InlineMacro = '$' };
 /// use to create names in the global namespace.  This is allocated per-layer
 /// block.
 struct LayerBlockGlobals {
-
   /// If the layer needs to create a module, use this name.
   StringRef moduleName;
 
@@ -130,15 +132,26 @@ static SmallString<32> instanceNameForLayer(SymbolRefAttr layerName) {
   return result;
 }
 
-/// For all layerblocks `@A::@B::@C` in a circuit called Circuit,
-/// the output filename is `layers-Circuit-A-B-C.sv`.
-static SmallString<32> fileNameForLayer(StringRef circuitName,
+/// For all layerblocks `@A::@B::@C` in a module called Module,
+/// the output filename is `layers-Module-A-B-C.sv`.
+static SmallString<32> fileNameForLayer(StringRef moduleName,
                                         SymbolRefAttr layerName) {
   SmallString<32> result;
   result.append("layers");
-  appendName(circuitName, result);
+  appendName(moduleName, result);
   appendName(layerName, result);
   result.append(".sv");
+  return result;
+}
+
+/// For all layerblocks `@A::@B::@C` in a module called Module,
+/// the include-guard macro is `layers_Module_A_B_C`.
+static SmallString<32> guardMacroNameForLayer(StringRef moduleName,
+                                              SymbolRefAttr layerName) {
+  SmallString<32> result;
+  result.append("layers");
+  appendName(moduleName, result, false, Delimiter::BindModule);
+  appendName(layerName, result, false, Delimiter::BindModule);
   return result;
 }
 
@@ -158,6 +171,19 @@ macroNameForLayer(StringRef circuitName,
 // LowerLayersPass
 //===----------------------------------------------------------------------===//
 
+namespace {
+/// Information about each bind file we are emitting. During a prepass, we walk
+/// the modules to find layerblocks, creating an emit::FileOp for each bound
+/// layer used under each module. As we do this, we build a table of these info
+/// objects for quick lookup later.
+struct BindFileInfo {
+  /// The path of the bind file.
+  StringAttr path;
+  /// Where to insert bind statements into the bind file.
+  Block *body;
+};
+} // namespace
+
 class LowerLayersPass
     : public circt::firrtl::impl::LowerLayersBase<LowerLayersPass> {
   hw::OutputFileAttr getOutputFile(SymbolRefAttr layerName) {
@@ -167,15 +193,15 @@ class LowerLayersPass
     return layer->getAttrOfType<hw::OutputFileAttr>("output_file");
   }
 
-  hw::OutputFileAttr outputFileForLayer(StringRef circuitName,
+  hw::OutputFileAttr outputFileForLayer(StringRef moduleName,
                                         SymbolRefAttr layerName) {
     if (auto file = getOutputFile(layerName))
       return hw::OutputFileAttr::getFromDirectoryAndFilename(
           &getContext(), file.getDirectory(),
-          fileNameForLayer(circuitName, layerName),
+          fileNameForLayer(moduleName, layerName),
           /*excludeFromFileList=*/true);
     return hw::OutputFileAttr::getFromFilename(
-        &getContext(), fileNameForLayer(circuitName, layerName),
+        &getContext(), fileNameForLayer(moduleName, layerName),
         /*excludeFromFileList=*/true);
   }
 
@@ -199,13 +225,22 @@ class LowerLayersPass
   /// Lower an inline layerblock to an ifdef block.
   void lowerInlineLayerBlock(LayerOp layer, LayerBlockOp layerBlock);
 
-  /// Preprocess layers to build macro declarations and cache information about
-  /// the layers so that this can be quired later.
+  /// Build macro declarations and cache information about the layers.
   void preprocessLayers(CircuitNamespace &ns, OpBuilder &b, LayerOp layer,
                         StringRef circuitName,
                         SmallVector<FlatSymbolRefAttr> &stack);
-  void preprocessLayers(CircuitNamespace &ns, LayerOp layer,
-                        StringRef circuitName);
+  void preprocessLayers(CircuitNamespace &ns);
+
+  /// For each module, build a bindfile for each bound-layer, if needed.
+  void preprocessModules(CircuitNamespace &ns, InstanceGraph &ig);
+
+  /// Build the bindfile skeletons for each module. Set up a table which tells
+  /// us for each module/layer pair, where to insert the bind operations.
+  void preprocessModule(CircuitNamespace &ns, InstanceGraphNode *node);
+
+  /// Build a bindfile skeleton for a particular module and layer.
+  void buildBindFile(CircuitNamespace &ns, InstanceGraphNode *node,
+                     OpBuilder &b, SymbolRefAttr layerName, LayerOp layer);
 
   /// Entry point for the function.
   void runOnOperation() override;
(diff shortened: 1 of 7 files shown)


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -734,46 +783,189 @@ void LowerLayersPass::preprocessLayers(CircuitNamespace &ns, LayerOp layer,
...
+    globals.hierPathName = ns.newName(layerBlockHierPathName);
+    layerBlockGlobals.insert({layerBlock, globals});
+  });
+
+  // Create a bindfile for layers used indirectly under this module.
+  for (auto *record : *node) {
+    auto *child = record->getTarget()->getModule().getOperation();
+    for (auto [layer, _] : bindFiles[child])
+      todo.insert(layer);
+  }
```
seldridge, line 930: Nice use of post-order and caching the bind files in the child modules to avoid exponential behavior. 💯

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -734,46 +783,189 @@ void LowerLayersPass::preprocessLayers(CircuitNamespace &ns, LayerOp layer,
...
+
+  // Build the bindfiles for any layer seen under this module.
+  for (auto [sym, layer] : symbolToLayer)
+    if (todo.contains(layer))
+      buildBindFile(ns, node, b, sym, layer);
+}
```
seldridge, line 936: This could use a `SetVector` to avoid iteration over all layers seeing if we get a hit. However, this would make the bind file order be layer block order instead of layer order (unless insertion order was even more complicated).

This is probably fine given that the number of layers is relatively small. If this changes, we may need to revisit this.

seldridge: Assuming the intention is to keep the bind files in layer declaration order, can you include a comment to this effect?

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -831,87 +1023,6 @@ void LowerLayersPass::runOnOperation() {
...
-                                                 "`endif // " + prefixMacro);
-    footer->setAttr("output_file", bindFile);
-
-    if (!layerOp.getBody().getOps<LayerOp>().empty())
-      layers.emplace_back(
-          layerOp,
-          builder.getStringAttr("`include \"" +
-                                bindFile.getFilename().getValue() + "\""));
-  });
-
```
seldridge, line 914: It is great to see this go away. 💯

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -649,24 +692,30 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
-        /*doNotPrint=*/false,
-        /*innerSym=*/
-        (innerSyms.empty() ? hw::InnerSymAttr{}
-                           : hw::InnerSymAttr::get(builder.getStringAttr(
-                                 ns.newName(instanceName)))));
-
-    auto outputFile =
-        outputFileForLayer(circuitName, layerBlock.getLayerName());
+        /*portAnnotations=*/ArrayRef<Attribute>{}, /*lowerToBind=*/false,
+        /*doNotPrint=*/true, innerSym);
```
seldridge, line 704: I'm still a little hazy on the story around `lowerToBind` vs. `doNotPrint`. This works because `lowerToBind` is replaced with bind ops in the bind files, correct?

Regardless, having both of these is not ideal and would be good to clean up.

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -629,3 +633,44 @@ firrtl.circuit "Verbatim" {
 // CHECK-NEXT:     sv.verbatim
 // CHECK-SAME:     !firrtl.uint<10> {symbols = [#hw.innerNameRef<@[[VL]]::@node>]}
 // CHECK-NEXT:   }
+
+
+// -----
+
+// Check that for public modules, bindfiles are generated for all known layers/
```
seldridge, line 640: Nit: 
```suggestion
// Check that for public modules, bindfiles are generated for all known layers.
```

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -629,3 +633,44 @@ firrtl.circuit "Verbatim" {
...
+firrtl.circuit "Test" {
+  firrtl.layer @A  bind { }
```
seldridge, line 642: Supreme nit (throughout) of double space:

```suggestion
  firrtl.layer @A bind { }
```

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -629,3 +633,44 @@ firrtl.circuit "Verbatim" {
...
+  firrtl.module public @Test2() {}
+}
+// CHECK: emit.file "layers-Test-A.sv"
+// CHECK: emit.file "layers-Test-B.sv"
+// CHECK: emit.file "layers-Test2-A.sv"
+// CHECK: emit.file "layers-Test2-B.sv"
+
+// -----
+
+// Check that for private modules, bindfiles files are only generated if they are used.
```
seldridge, line 654: Nice test!

test/firtool/layers.fir
```diff
@@ -51,15 +51,15 @@ circuit Foo: %[[
...
+; CHECK-NEXT:    `define layers_Foo_A
+; CHECK-NEXT:     bind Foo Foo_A a ();
+; CHECK-NEXT:  `endif // not def layers_Foo_A
+
+; CHECK-LABEL: FILE "layers-Foo-A-B.sv"
+; CHECK:       `ifndef layers_Foo_A_B
+; CHECK-NEXT:    `define layers_Foo_A_B
+; CHECK-NEXT:    `include "layers-Foo-A.sv"
+; CHECK-NEXT:    bind Foo Foo_A_B a_b ();
+; CHECK-NEXT:  `endif // not def layers_Foo_A_B
```
seldridge, line 65: This re-ordering to layer order is quite nice.

seldridge (commented):
This all looks great!

One test I was thinking of is around the behavior of the includes and how that composes with directories. Consider the following:

``` firrtl
FIRRTL version 5.1.0
circuit Foo: %[[
  {"class": "firrtl.transforms.DontTouchAnnotation", "target": "~|Foo>b"},
  {"class": "firrtl.transforms.DontTouchAnnotation", "target": "~|Bar>a"}
]]

  layer A, bind, "A":

  layer B, bind, "B":

  public module Bar enablelayer A:

    layerblock A:
      node a = UInt<1>(0)

  public module Foo enablelayer A:

    inst bar of Bar

    layerblock B:
      node b = UInt<1>(0)
```

With this PR, this produces the following bind files:

``` verilog
// ----- 8< ----- FILE "A/layers-Bar-A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.116.0-4-g651c02833
`ifndef layers_Bar_A
  `define layers_Bar_A
  bind Bar Bar_A a ();
`endif // not def layers_Bar_A

// ----- 8< ----- FILE "B/layers-Bar-B.sv" ----- 8< -----

// Generated by CIRCT firtool-1.116.0-4-g651c02833
`ifndef layers_Bar_B
  `define layers_Bar_B
`endif // not def layers_Bar_B

// ----- 8< ----- FILE "A/layers-Foo-A.sv" ----- 8< -----

// Generated by CIRCT firtool-1.116.0-4-g651c02833
`ifndef layers_Foo_A
  `define layers_Foo_A
  `include "layers-Bar-A.sv"
`endif // not def layers_Foo_A

// ----- 8< ----- FILE "B/layers-Foo-B.sv" ----- 8< -----

// Generated by CIRCT firtool-1.116.0-4-g651c02833
`ifndef layers_Foo_B
  `define layers_Foo_B
  `include "layers-Bar-B.sv"
  bind Foo Foo_B b ();
`endif // not def layers_Foo_B
```

It would be better if any `include` was relative to the output directory, i.e.:

```
`include "A/layers-Bar-A.sv"
`include "B/layers-Bar-B.sv"
```

My reasoning here is that keeping this all relative to one directory (the output directory) simplifies Verilog tooling as you only have to tell the tool about _one_ include directory. If, instead, this does what it is doing today, then you have to tell the tool about _all_ layer directories in order for this to work. 

Specifically, we're using item one of the Verilog spec here and it's convenient to keep this to either "the compiler's current working directory" (the user has to do nothing) or exactly one "user-specified location". Most simulators support the `+incdir` option which sets additional user-specified locations.
 
Screenshot 2025-05-06 at 13 46 38

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -649,24 +692,30 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
     builder.setInsertionPointAfter(layerBlock);
     auto instanceName = instanceNameForLayer(layerBlock.getLayerName());
+    auto innerSym =
+        hw::InnerSymAttr::get(builder.getStringAttr(ns.newName(instanceName)));
+
     auto instanceOp = builder.create<InstanceOp>(
         layerBlock.getLoc(), /*moduleName=*/newModule,
         /*name=*/
-        instanceName, NameKindEnum::DroppableName,
+        instanceName, NameKindEnum::InterestingName,
```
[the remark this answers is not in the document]
rwy7: I've dropped it! I'm not sure why I did it this way, I don't need it to be an interesting name. Thanks.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -199,13 +223,30 @@ class LowerLayersPass
...
 
-  /// Preprocess layers to build macro declarations and cache information about
-  /// the layers so that this can be quired later.
+  /// Build macro declarations and cache information about the layers.
   void preprocessLayers(CircuitNamespace &ns, OpBuilder &b, LayerOp layer,
                         StringRef circuitName,
                         SmallVector<FlatSymbolRefAttr> &stack);
   void preprocessLayers(CircuitNamespace &ns, LayerOp layer,
                         StringRef circuitName);
+  void preprocessLayers(CircuitNamespace &ns);
```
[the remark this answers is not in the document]
rwy7: Done, thanks.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
rwy7: Yep, that's right.

rwy7:
I've updated this PR so that, when a bindfile includes another, the inclusion uses the includee's output path. There is a test, too. Here is the output:

```verilog
// Generated by CIRCT unknown git version
module Bar();
endmodule

module Foo();
  Bar bar ();
endmodule


// ----- 8< ----- FILE "A/Bar_A.sv" ----- 8< -----

// Generated by CIRCT unknown git version
module Bar_A();
  wire a = 1'h0;
endmodule


// ----- 8< ----- FILE "A/layers-Bar-A.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`ifndef layers_Bar_A
  `define layers_Bar_A
  bind Bar Bar_A a ();
`endif // not def layers_Bar_A

// ----- 8< ----- FILE "B/layers-Bar-B.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`ifndef layers_Bar_B
  `define layers_Bar_B
`endif // not def layers_Bar_B

// ----- 8< ----- FILE "B/Foo_B.sv" ----- 8< -----

// Generated by CIRCT unknown git version
module Foo_B();
  wire b = 1'h0;
endmodule


// ----- 8< ----- FILE "A/layers-Foo-A.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`ifndef layers_Foo_A
  `define layers_Foo_A
  `include "A/layers-Bar-A.sv"
`endif // not def layers_Foo_A

// ----- 8< ----- FILE "B/layers-Foo-B.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`ifndef layers_Foo_B
  `define layers_Foo_B
  `include "B/layers-Bar-B.sv"
  bind Foo Foo_B b ();
`endif // not def layers_Foo_B
```

seldridge (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -734,11 +777,161 @@ void LowerLayersPass::preprocessLayers(CircuitNamespace &ns, OpBuilder &b,
...
+  OpBuilder::InsertionGuard _(b);
+  b.setInsertionPointToEnd(bindFile.getBody());
+
+  // Create the #ifndef for the include guard.
+  auto guard = b.create<sv::IfDefOp>(loc, macroSymbolAttr);
+  b.createBlock(&guard.getElseRegion());
+
+  // Create the #define for the include guard.
+  b.create<sv::MacroDefOp>(loc, macroSymbolAttr,
+                           StringAttr::get(&getContext()));
```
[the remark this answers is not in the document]
rwy7: I added a new builder for the sv::MacroDefOp which does not take a format_string argument, it uses an empty string as a default value. I think this is a common enough case that it warrants a builder.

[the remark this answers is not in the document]
seldridge: Nice idea. 👍

PR #8460, opened 2025-05-01
[OM] Ensure every sublist is finished in ListConcatOp evaluation.

mikeurbach:
We would previously check if the value corresponding to each operand of the ListConcatOp was finished evaluating, but in the case where the operand is a ReferenceValue, the reference may be done evaluating before while the underlying ListValue is not.

To handle this, a check was added that the result of extractList is fully evaluated, and the evaluation of the ListConcatOp doesn't proceed until that is true.

A small unit test shows this scenario, where there are object references in lists passed as field references to a ListConcatOp.

### DIFF
--- a/unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
+++ b/unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
@@ -1267,21 +1267,23 @@ TEST(EvaluatorTests, ListConcatPartialCycle) {
       "  %1 = om.list_create %field_in : !om.any"
       "  om.class.fields %1 : !om.list<!om.any>"
       "}"
-      "om.class @Leaf() -> () {"
-      "  om.class.fields"
+      "om.class @Leaf(%id_in: !om.integer) -> (id: !om.integer) {"
+      "  om.class.fields %id_in : !om.integer"
       "}"
       "om.class @ListConcatPartialCycle() -> (result: !om.list<!om.any>){"
-      "  %0 = om.object @Child(%6) : (!om.any) -> !om.class.type<@Child>"
+      "  %0 = om.object @Child(%7) : (!om.any) -> !om.class.type<@Child>"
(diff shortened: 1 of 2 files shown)


unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
```diff
@@ -1261,6 +1261,56 @@ TEST(EvaluatorTests, ListOfListConcat) {
...
+  ASSERT_TRUE(succeeded(result));
+
+  auto fieldValue = llvm::cast<evaluator::ObjectValue>(result.value().get())
+                        ->getField("result")
+                        .value();
+
+  auto finalList =
+      llvm::cast<evaluator::ListValue>(fieldValue.get())->getElements();
+
+  ASSERT_EQ(2U, finalList.size());
```
leonardt, line 1311: Perhaps it's worth checking the list contents are what we expect?

leonardt (approved):
LGTM, one comment about the test.

Although I can't see it based on the code shown here, I'm guessing the surrounding context will evaluate these values using a worklist style algorithm until everything is fully evaluated? So that makes returning the unevaluated list work since it will come back around and try again once the dependent values have been evaluated.

mikeurbach:
> I'm guessing the surrounding context will evaluate these values using a worklist style algorithm until everything is fully evaluated? So that makes returning the unevaluated list work since it will come back around and try again once the dependent values have been evaluated.

Yep that's exactly how it works. You can return a value with the "is fully evaluated" flag set false to indicate that this needs to go back at the end of the worklist.

unittests/Dialect/OM/Evaluator/EvaluatorTests.cpp
mikeurbach: Sure yeah let me do something with the object references to test that we got the two expected objects.

mikeurbach: Added a new field to each object so I could check the value of that field in the test.

PR #8463, opened 2025-05-05
[tests] drop removed SMTLIB CAPI test from lit config

TaoBi22:
I was getting `llvm-lit: /home/bea/Research/circt-fork/llvm/llvm/utils/lit/lit/llvm/subst.py:126: note: Did not find circt-capi-smtlib-test in /home/bea/Research/circt-fork/build/bin::/home/bea/Research/circt-fork/llvm/build/./bin:/usr/bin` when running check-circt. Looks like that's because there's a test still in the lit config that was removed in #8424. This just drops that from the config.


fabianschuiki (approved):
LGTM! Thanks :smiley:

PR #8464, opened 2025-05-05
Place black boxes with the design, if instantiated there

rwy7:
The BlackBoxReaderPass has its own logic for deciding if the output file of a black box module should be under the design, or testbench, based on how it is instantiated. This PR changes this logic, so that if a black-box module is instantiated at all under the effective design, then it will be placed in the design directory. Otherwise, it will be placed in the test bench directory. This reflects the principle that, the testbench may depend on the design, but the design must not depend on the testbench. Before this PR, the logic was: if there are any instances outside of the design, then the black-box module is placed in the testbench.

### DIFF
--- a/test/firtool/blackbox-directories.fir
+++ b/test/firtool/blackbox-directories.fir
@@ -347,3 +347,38 @@ circuit BlackBoxLCA: %[[
     layerblock B:
       inst bar of Bar
       connect bar.a, a
+
+; // -----
+
+; When a black box module is instantiated under both the design and the
+; testbench, the black box must be placed in the design (the testbench can refer
+; to modules in the design, but the design must not depend on the testbench).
+
+FIRRTL version 4.0.0
+
+circuit TestHarness: %[[
+  {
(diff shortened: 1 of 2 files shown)


seldridge (approved):
LGTM

test/firtool/blackbox-directories.fir
```diff
@@ -347,3 +347,38 @@ circuit BlackBoxLCA: %[[
...
+  {
+    "class": "sifive.enterprise.firrtl.MarkDUTAnnotation",
+    "target": "~|Component"
+  },
+  {
+    "class": "sifive.enterprise.firrtl.TestBenchDirAnnotation",
+    "dirname": "testbench"
+  }
+]]
+  ;; CHECK: FILE "./BlackBox.sv"
```
seldridge, line 375: Windows compatibility needed:

```suggestion
  ;; CHECK: FILE ".{{[/\]}}BlackBox.sv"
```

fabianschuiki (approved):
LGTM!

PR #8466, opened 2025-05-05
[firrtl] Improve LowerLayer XMR naming

### LINKED ISSUE #8426: [firrtl] Improve LowerLayer Optimizations
In https://github.com/llvm/circt/pull/8405, I introduced changes where an operand captured from outside a layer block will cause the operand to be replaced with an XMR to _a node_ driven by the operand's value with a prefix of `_layer_probe`.

This is sub-optimal as it leaves dead code in the design. However, it was empirically necessary to allow for heavier optimization of circuits. E.g., the following circuit is a reduction of circuit that proved problematic:

``` mlir
firrtl.circuit "Foo" {
  firrtl.layer @A bind {}
  firrtl.module @Foo(
    in %clk: !firrtl.clock,
    in %a: !firrtl.uint<1>,
    in %b: !firrtl.uint<8>,
    out %c: !firrtl.uint<8>
  ) {
    %c1_ui1 = firrtl.constant 1 : !firrtl.uint<1>
    %reg = firrtl.reg %clk : !firrtl.clock, !firrtl.uint<8>
    %x = firrtl.wire : !firrtl.uint<1>
    %0 = firrtl.or %a, %x : (!firrtl.uint<1>, !firrtl.uint<1>) -> !firrtl.uint<1>
    %1 = firrtl.mux(%0, %b, %reg) : (!firrtl.uint<1>, !firrtl.uint<8>, !firrtl.uint<8>) -> !firrtl.uint<8>
    %2 = firrtl.not %a : (!firrtl.uint<1>) -> !firrtl.uint<1>
    firrtl.matchingconnect %x, %2 : !firrtl.uint<1>
    firrtl.matchingconnect %reg, %1 : !firrtl.uint<8>
    firrtl.matchingconnect %c, %reg : !firrtl.uint<8>
    firrtl.layerblock @A {
      firrtl.assert %clk, %x, %c1_ui1, "" : !firrtl.clock, !firrtl.uint<1>, !firrtl.uint<1>
    }
  }
}
```

Without this dead node, I would get Verilog that looked like:

```verilog
 reg  [7:0] reg_0;
  wire       x = ~a;
  always @(posedge clk) begin
    if (a | x)
      reg_0 <= b;
```

The `if` condition is `a | ~a`. However, introducing a symbol on the wire `%x` will block later optimizations.

Figure out a way to untangle this and remove the `_layer_probe` nodes. This may be able to be handled with improved optimizations in the FIRRTL pipeline.

### LINKED ISSUE #8465: [FIRRTL] Regression in SystemVerilog signal naming with layers
cc @seldridge 

After #8405 was merged, there was a regression in SystemVerilog emission from FIRRTL for designs containing layers. Many intermediate signals, which previously had meaningful names, were replaced with `_layer_probe_*` signals.

I came up with a somewhat-minimal design which generates some interesting intermediate signals, which get nicely named in the output SystemVerilog. Then when you introduce a layer, the naming is degraded. Here is a Chisel design for recreating that condition:

```scala
//> using scala 2.13.16
//> using repository https://s01.oss.sonatype.org/content/repositories/snapshots
//> using dep org.chipsalliance::chisel:7.0.0-RC1+8-81480282-SNAPSHOT
//> using plugin org.chipsalliance:::chisel-plugin:7.0.0-RC1+8-81480282-SNAPSHOT
//> using options -unchecked -deprecation -language:reflectiveCalls -feature -Xcheckinit -Xfatal-warnings -Ywarn-dead-code -Ywarn-unused -Ymacro-annotations

// Replace the SNAPSHOT above with these to use firtool-1.114.1, which doesn't include #8405
// //> using dep org.chipsalliance::chisel:7.0.0-RC1+4-fe547683-SNAPSHOT
// //> using plugin org.chipsalliance:::chisel-plugin:7.0.0-RC1+4-fe547683-SNAPSHOT

import chisel3._
import chisel3.util._

class Foo extends Module {
  val in = IO(Flipped(Decoupled(UInt(4.W))))
  val out = IO(Decoupled(UInt(4.W)))

  val foo = RegInit(0.U(8.W))
  foo :<= foo + 1.U

  val qs = Seq.fill(2)(Module(new Queue(UInt(4.W), 8)))

  val allQsReady = qs.map(_.io.enq.ready).reduce(_ && _)
  val enqFire    = in.valid && allQsReady && (foo === 0xff.U)

  in.ready :<= enqFire
  qs.foreach(_.io.enq.valid :<= enqFire)
  qs.foreach(_.io.enq.bits :<= in.bits)

  val allQsValid = qs.map(_.io.deq.valid).reduce(_ && _)
  val deqFire = out.ready && allQsValid && (foo === 0.U)

  out.valid :<= deqFire
  qs.foreach(_.io.deq.ready :<= deqFire)

  out.bits :<= qs.map(_.io.deq.bits).reduce(_ | _)

  // Uncomment to experiment with how `enqFire` gets named
  // when(enqFire) {
  //   printf("enqFire\n")
  // }
}

object Main extends App {
  import _root_.circt.stage.ChiselStage
  println(
    ChiselStage.emitSystemVerilog(
      new Foo,
      firtoolOpts = Array("-disable-all-randomization", "-strip-debug-info")
    )
  )
}
```

When you run this, you get this output, which notably has the signal named `enqFire`:

```verilog
// Generated by CIRCT firtool-1.116.0
// VCS coverage exclude_file

// RAM and queue modules ...

module Foo(
  input        clock,
               reset,
  output       in_ready,
  input        in_valid,
  input  [3:0] in_bits,
  input        out_ready,
  output       out_valid,
  output [3:0] out_bits
);

  wire       _qs_1_io_enq_ready;
  wire       _qs_1_io_deq_valid;
  wire [3:0] _qs_1_io_deq_bits;
  wire       _qs_0_io_enq_ready;
  wire       _qs_0_io_deq_valid;
  wire [3:0] _qs_0_io_deq_bits;
  reg  [7:0] foo;
  wire       enqFire = in_valid & _qs_0_io_enq_ready & _qs_1_io_enq_ready & (&foo);
  wire       deqFire = out_ready & _qs_0_io_deq_valid & _qs_1_io_deq_valid & foo == 8'h0;
  always @(posedge clock) begin
    if (reset)
      foo <= 8'h0;
    else
      foo <= foo + 8'h1;
  end // always @(posedge)
  Queue8_UInt4 qs_0 (
    .clock        (clock),
    .reset        (reset),
    .io_enq_ready (_qs_0_io_enq_ready),
    .io_enq_valid (enqFire),
    .io_enq_bits  (in_bits),
    .io_deq_ready (deqFire),
    .io_deq_valid (_qs_0_io_deq_valid),
    .io_deq_bits  (_qs_0_io_deq_bits)
  );
  Queue8_UInt4 qs_1 (
    .clock        (clock),
    .reset        (reset),
    .io_enq_ready (_qs_1_io_enq_ready),
    .io_enq_valid (enqFire),
    .io_enq_bits  (in_bits),
    .io_deq_ready (deqFire),
    .io_deq_valid (_qs_1_io_deq_valid),
    .io_deq_bits  (_qs_1_io_deq_bits)
  );
  assign in_ready = enqFire;
  assign out_valid = deqFire;
  assign out_bits = _qs_0_io_deq_bits | _qs_1_io_deq_bits;
endmodule
```

If you uncomment that `printf` statement, you get this layer module:

```verilog
// ----- 8< ----- FILE "verification/Foo_Verification.sv" ----- 8< -----

// Generated by CIRCT firtool-1.116.0

`ifndef PRINTF_COND_
  `ifdef PRINTF_COND
  `else  // PRINTF_COND
    `define PRINTF_COND_ 1
  `endif // PRINTF_COND
module Foo_Verification();
  `ifndef SYNTHESIS
    always @(posedge Foo.clock) begin
      if ((`PRINTF_COND_) & Foo._layer_probe & ~Foo.reset)
        $fwrite(32'h80000002, "enqFire\n");
    end // always @(posedge)
  `endif // not def SYNTHESIS
endmodule
```

And you get the following diff in the `Foo` module:

```diff
--- Foo.without-layer.v 2025-05-05 11:14:44
+++ Foo.with-layer.v    2025-05-05 11:15:13
@@ -16,7 +16,7 @@
   wire       _qs_0_io_deq_valid;
   wire [3:0] _qs_0_io_deq_bits;
   reg  [7:0] foo;
-  wire       enqFire = in_valid & _qs_0_io_enq_ready & _qs_1_io_enq_ready & (&foo);
+  wire       _layer_probe = in_valid & _qs_0_io_enq_ready & _qs_1_io_enq_ready & (&foo);
   wire       deqFire = out_ready & _qs_0_io_deq_valid & _qs_1_io_deq_valid & foo == 8'h0;
   always @(posedge clock) begin
     if (reset)
@@ -28,7 +28,7 @@
     .clock        (clock),
     .reset        (reset),
     .io_enq_ready (_qs_0_io_enq_ready),
-    .io_enq_valid (enqFire),
+    .io_enq_valid (_layer_probe),
     .io_enq_bits  (in_bits),
     .io_deq_ready (deqFire),
     .io_deq_valid (_qs_0_io_deq_valid),
@@ -38,13 +38,13 @@
     .clock        (clock),
     .reset        (reset),
     .io_enq_ready (_qs_1_io_enq_ready),
-    .io_enq_valid (enqFire),
+    .io_enq_valid (_layer_probe),
     .io_enq_bits  (in_bits),
     .io_deq_ready (deqFire),
     .io_deq_valid (_qs_1_io_deq_valid),
     .io_deq_bits  (_qs_1_io_deq_bits)
   );
-  assign in_ready = enqFire;
+  assign in_ready = _layer_probe;
   assign out_valid = deqFire;
   assign out_bits = _qs_0_io_deq_bits | _qs_1_io_deq_bits;
 endmodule
```

So the `enqFire` signal gets replaced with `_layer_probe`. If you run this design with `firtool-1.114.1` (see comments in the sample above to perform that experiment), then `enqFire`'s name is retained in the design that includes the `printf`.

seldridge:
Improve the naming of values captured by the `LowerLayers` pass.  Change
the logic used when a node is created.  Try to use the underlying value if
possible.  If unable to do this, then create a node, but use a saner name
for it.

Fixes #8465.
Fixes #8426.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -333,8 +333,8 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
   DenseMap<Value, Operation *> nameableCache;
 
   // Get or create a node op for a value captured by a layer block.
-  auto getOrCreateNodeOp =
-      [&](Value operand, ImplicitLocOpBuilder &builder) -> Operation * {
+  auto getOrCreateNodeOp = [&](Value operand,
+                               ImplicitLocOpBuilder &builder) -> Operation * {
     // Use the cache hit.
     auto it = nameableCache.find(operand);
     if (it != nameableCache.end())
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -332,26 +332,13 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
   // A cache of values to nameable ops that can be used
   DenseMap<Value, Operation *> nameableCache;
 
-  // Get or create a nameable op (a node) for some value.  If value is a result
-  // of an operation that can support a name, then use that, except for
-  // instances and memories.  If the operation can't support a name, but
-  // provides a name hint, then use that.
-  auto getOrCreateNameableOp =
+  // Get or create a node op for a value captured by a layer block.
+  auto getOrCreateNodeOp =
       [&](Value operand, ImplicitLocOpBuilder &builder) -> Operation * {
     // Use the cache hit.
     auto it = nameableCache.find(operand);
     if (it != nameableCache.end())
       return it->getSecond();
-
-    // This _is_ a nameable op with a simple result.  Put that in the cache and
-    // use it.  For more complex operations, e.g., instances, do not do this.
-    // If we supported inner symbols on instance ports, then we could avoid
-    // needing to create a node.
-    auto *definingOp = operand.getDefiningOp();
-    if (isa_and_present<NodeOp, RegOp, RegResetOp, WireOp>(definingOp)) {
-      nameableCache.insert({operand, definingOp});
-      return definingOp;
-    }
 
     // Create a new node.  Put it in the cache and use it.
     OpBuilder::InsertionGuard guard(builder);
@@ -367,8 +354,6 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
       } else if (auto opName = definingOp->getAttrOfType<StringAttr>("name")) {
         nameHint.append(opName);
       }
-      if (!nameHint.empty())
-        nameHint.append("_probe");
     }
     return nameableCache
         .insert({operand, builder.create<NodeOp>(operand.getLoc(), operand,
@@ -433,17 +418,13 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
       auto *definingOp = value.getDefiningOp();
       hw::InnerRefAttr innerRef;
       if (definingOp) {
-        // TODO: Always create a node.  This is a trade-off between
-        // optimizations and dead code.  By adding the node, this allows the
-        // original path to be better optimized, but will leave dead code in the
-        // design.  If the node is not created, then the output is less
-        // optimized.  Err on the side of dead code.  This _should_ be able to
-        // be removed eventually [1].  A node should only be necessary for
-        // operations which cannot have a symbol attached to them, i.e.,
-        // expressions or instance ports.
-        //
-        // [1]: https://github.com/llvm/circt/issues/8426
-        definingOp = getOrCreateNameableOp(value, localBuilder);
+        // Always create a node.  This is a trade-off between optimizations and
+        // dead code.  By adding the node, this allows the original path to be
+        // better optimized, but will leave dead code in the design.  If the
+        // node is not created, then the output is less optimized.  Err on the
+        // side of dead code.  This dead node _may_ be eventually inlined by
+        // `ExportVerilog`.  However, this is not guaranteed.
+        definingOp = getOrCreateNodeOp(value, localBuilder);
         innerRef = getInnerRefTo(
             definingOp, [&](auto) -> hw::InnerSymbolNamespace & { return ns; });
       } else {
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -329,23 +329,52 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
   StringRef circuitName = circuitOp.getName();
   hw::InnerSymbolNamespace ns(moduleOp);
 
-  // A cache of nodes we've created for values which are captured, but cannot
-  // have a symbol attached to them.  This is done to avoid creating the same
-  // node multiple times.
-  DenseMap<Value, NodeOp> nodeCache;
+  // A cache of values to nameable ops that can be used
+  DenseMap<Value, Operation *> nameableCache;
 
-  // Get or create a node within this module.  This is used when we need to
-  // create an XMRDerefOp to an expression.
-  auto getOrCreateNodeOp = [&](Value operand, ImplicitLocOpBuilder &builder,
-                               StringRef nameHint) -> NodeOp {
-    auto it = nodeCache.find(operand);
-    if (it != nodeCache.end())
+  // Get or create a nameable op (a node) for some value.  If value is a result
+  // of an operation that can support a name, then use that, except for
+  // instances and memories.  If the operation can't support a name, but
+  // provides a name hint, then use that.
+  auto getOrCreateNameableOp =
+      [&](Value operand, ImplicitLocOpBuilder &builder) -> Operation * {
+    // Use the cache hit.
+    auto it = nameableCache.find(operand);
+    if (it != nameableCache.end())
       return it->getSecond();
+
+    // This _is_ a nameable op with a simple result.  Put that in the cache and
+    // use it.  For more complex operations, e.g., instances, do not do this.
+    // If we supported inner symbols on instance ports, then we could avoid
+    // needing to create a node.
+    auto *definingOp = operand.getDefiningOp();
+    if (isa_and_present<NodeOp, RegOp, RegResetOp, WireOp>(definingOp)) {
+      nameableCache.insert({operand, definingOp});
+      return definingOp;
+    }
+
+    // Create a new node.  Put it in the cache and use it.
     OpBuilder::InsertionGuard guard(builder);
     builder.setInsertionPointAfter(operand.getDefiningOp());
-    return nodeCache
-        .insert({operand,
-                 builder.create<NodeOp>(operand.getLoc(), operand, nameHint)})
+    SmallString<16> nameHint;
+    // Try to generate a "good" name hint to use for the node.
+    if (auto *definingOp = operand.getDefiningOp()) {
+      if (auto instanceOp = dyn_cast<InstanceOp>(definingOp)) {
+        nameHint.append(instanceOp.getName());
+        nameHint.push_back('_');
+        nameHint.append(
+            instanceOp.getPortName(cast<OpResult>(operand).getResultNumber()));
+      } else if (auto opName = definingOp->getAttrOfType<StringAttr>("name")) {
+        nameHint.append(opName);
+      }
+      if (!nameHint.empty())
+        nameHint.append("_probe");
+    }
+    return nameableCache
+        .insert({operand, builder.create<NodeOp>(operand.getLoc(), operand,
+                                                 nameHint.empty()
+                                                     ? "_layer_probe"
+                                                     : StringRef(nameHint))})
         .first->getSecond();
   };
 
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -329,23 +329,52 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+    // needing to create a node.
+    auto *definingOp = operand.getDefiningOp();
+    if (isa_and_present<NodeOp, RegOp, RegResetOp, WireOp>(definingOp)) {
+      nameableCache.insert({operand, definingOp});
+      return definingOp;
+    }
+
+    // Create a new node.  Put it in the cache and use it.
     OpBuilder::InsertionGuard guard(builder);
     builder.setInsertionPointAfter(operand.getDefiningOp());
```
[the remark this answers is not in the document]
dtzSiFive: (I did not confirm if that's reachable -- but if it is then bad, if it isn't the `and_present` is unnecessary/misleading)

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -414,7 +443,7 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
         // expressions or instance ports.
```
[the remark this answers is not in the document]
dtzSiFive: This also appears addressed, marking resolved.

Comment is still misusing the term "dead" and other details but not deal-breakers.

test/firtool/layers.fir
```diff
@@ -38,17 +38,17 @@ circuit Foo: %[[
             assert(clock, cond, enable, "Test")
 
 ; CHECK-LABEL: module Foo_A_B();
-; CHECK:         wire y = Foo.a._layer_probe;
-; CHECK:         wire z = Foo.a._layer_probe;
+; CHECK:         wire y = Foo.a.x_0;
```
dtzSiFive, line 41: I believe you've done this a few ways and compared output on various designs, but I'm surprised this provides the best output.  Do the `_0`'s generally make it to the output?

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -329,23 +329,37 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
-    auto it = nodeCache.find(operand);
-    if (it != nodeCache.end())
+  // A cache of values to nameable ops that can be used
+  DenseMap<Value, Operation *> nameableCache;
+
+  // Get or create a node op for a value captured by a layer block.
+  auto getOrCreateNodeOp = [&](Value operand,
+                               ImplicitLocOpBuilder &builder) -> Operation * {
+    // Use the cache hit.
+    auto it = nameableCache.find(operand);
```
dtzSiFive, line 339: Consider a variant that does a since lookup, and update the value if cache doesn't have this operand already.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -329,23 +329,37 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+
+    // Create a new node.  Put it in the cache and use it.
     OpBuilder::InsertionGuard guard(builder);
     builder.setInsertionPointAfter(operand.getDefiningOp());
-    return nodeCache
-        .insert({operand,
-                 builder.create<NodeOp>(operand.getLoc(), operand, nameHint)})
+    SmallString<16> nameHint;
+    // Try to generate a "good" name hint to use for the node.
+    if (auto *definingOp = operand.getDefiningOp()) {
```
dtzSiFive, line 348: Nope, resolved earlier comment too soon.  This check is inconsistent with the code 1 line above the `nameHint`; declaration -- if there is no defining op, `setInsertionPointAfter` looks like it will crash.

Generally I'd rather have defensive (but tight) code that also covers the natural things one would expect it to -- vs over-eagerly depending on the current guarantees made by the calling context / surroundings -- but this code is inconsistent across 3 lines.

Consider using `setInsertionPointAfterValue` (possibly) or go all-in on assuming this is never used on a block argument.

dtzSiFive (approved):
There's work to be done to improve the situation here, but "we'll fix that later" 🙃 .

Added some minor comments, have at it.

test/firtool/layers.fir
seldridge: The output from the previous iteration of this PR was better for this one, somewhat contrived test. The use of `dontTouch` is making this problematic, though it does indicate a deficiency in the algorithm (though it is still better than without this patch of `_layer_probe`).

The `dontTouch` will preserve nodes late into the pipeline. The proposed algorithm of _always create a node_ will not be optimal in this situation as it creates a double node. We could probably special case this situation, but it would be helpful if this part of the pipeline would lower `DontTouchAnnotation` into `innerSym` before this pass as opposed to needing to handle both situations (as at this point, we still have annotations that will be converted to inner symbols during `LowerToHW`).

This _could_ make it to the output if you capture a node marked `dontTouch`.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: I like the `setInsertionPointAfterValue` approach. I think I was originally leaning towards the "this is a lambda and it only exists to make some later, extremely restrictive code easier to read". From that perspective, not checking that this isn't a blockarg (or at least asserting) seemed fine. However, there's little cost to making this more sensible.

Nice suggestion.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: This is interesting... For some reason this pattern never occurred to me to do. Just for clarity, you're talking about something like:

``` c++
auto *op = foo.lookup(...);
if (op)
  return op;

/* body */

return op = builder.create<T>(...);
```

I normally don't consider this as I didn't think there was a default constructor here, but I guess the default would be a `nullptr` `Operation *` so this whole thing slots together nicely.

Do I understand you correctly?

Originally, I thought you were advising a hypothetical `lookupOrElse(key, function_ref)` which I didn't think existed (though I'm used to seeing in other languages).

seldridge: Changed in: https://github.com/llvm/circt/pull/8466/commits/1028e647fdc95b8dede529e503cfa190e35f4526

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
seldridge: Changed to `setInsertionPointAfterValue` in: https://github.com/llvm/circt/pull/8466/commits/9e867a2ac2e4b0daa7062faa1753adf779b4025f

PR #8467, opened 2025-05-06
Bump LLVM to c6c2e21028cadef854cf22f6ecaa5eb9d224b76d.

mikeurbach:
The only change was to adapt to new GreedyRewriteConfig APIs.

This is a straightforward mapping after GreedyRewriteConfig was given
a more fluent API, with some minor naming changes applied in:

https://github.com/llvm/llvm-project/commit/0c61b24


maerhart (approved):
LGTM

PR #8468, opened 2025-05-07
Make sv::IfDefOp's regions graph-regions

rwy7:
Make the sv::IfDefOp use graph regions. There is a distinct "procedural" ifdef op which uses ssa regions. [GraphRegionNoTerminator](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/RegionKindInterface.td#L57-L62) is a trait-list which implies single-block, no-terminator, and graph regions.

Co-authored-by: @dtzSiFive


dtzSiFive (approved):
LGTM, thanks!

PR #8470, opened 2025-05-08
[Arc] Don't let LowerState delete instantiated external modules

TaoBi22:
There's a part of LowerState that deletes any `hw.module.extern` operations - however, this produces invalid IR if there's an external module whose name is still referenced by e.g. an instance. This just adds a check to made sure that the symbol containing the module's name has no remaining references.

P.S. how come this pass needs to erase all external modules? CC: @fabianschuiki as per git blame

### DIFF
--- a/lib/Dialect/Arc/Transforms/LowerState.cpp
+++ b/lib/Dialect/Arc/Transforms/LowerState.cpp
@@ -1189,7 +1189,18 @@ void LowerStatePass::runOnOperation() {
       return signalPassFailure();
     moduleOp.erase();
   }
+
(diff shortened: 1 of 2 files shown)


fabianschuiki (approved):
LGTM! Thanks for adding a check for this :+1: 

LowerState converts instances of external modules to internal input/output ports that the user can control from the code which uses the simulation model. This deletes the instances, and LowerState then deletes the modules as well to not leave unused modules around that later break the LLVM lowering.

TaoBi22:
Ah I see, that makes sense. Thanks for the explanation!

PR #8473, opened 2025-05-09
[ci] Build static binaries w/ 8MiB thread stack

seldridge:
Set an 8MiB thread stack size when building static binaries with Alpine Linux, i.e., muslc.  The default thread stack size for muslc is 128KiB [[1]] which can cause failures for pretty normal Chisel-generated FIRRTL files [[2]].

This was observed after the transition to Alpine Linux for static builds in 1.114.0.

Testing here: https://github.com/llvm/circt/actions/runs/14921300803/job/41917042585

[1]: https://wiki.musl-libc.org/functional-differences-from-glibc.html
[2]: https://discourse.llvm.org/t/firtool-v1-114-1-segmentation-fault-with-chisel-7-0-0-rc1/86237


teqdruid:
I never understood why in this age of virtual memory (and plenty of physical memory) Linux is so stingy with stack space.

PR #8475, opened 2025-05-11
[circt-opt][Calyx] Integrate the Math to `circt-opt` and lower `math.sqrt` to Calyx

jiahanxie353:
After discussing with @fabianschuiki over the CIRCT Discord, we both agree that integrating the Math dialect can be helpful to both Calyx and Moore dialect. So I just initiate the integration by bundling with the Calyx use case.


cgyurgyik (approved):

fabianschuiki:
Very nice! :partying_face:

PR #8477, opened 2025-05-12
[Seq] Remove redundant enable mux in front of compreg.ce

fabianschuiki:
Add a canonicalizer that removes `comb.mux` and `arith.select` ops on the data input of the register if those ops use the register's enable as select operand. For example:

    %0 = comb.mux %enable, %a, %b
    seq.compreg.ce %0, %clock, %enable

gets canonicalized to

    seq.compreg.ce %a, %clock, %enable


uenoku (approved):
LGTM

teqdruid (approved):

PR #8478, opened 2025-05-12
SimToSV: Keep plusargs non-synth === under !SYNTHESIS block

youngar:
This changes the lowering of plusargs to make sure that the X-squashing === is underneath the !SYNTHESIS block, which fixes a warning produced by synthesis tools.


ig-raiderler:
Very clean code with good test coverage. This is a great example of best practices.

PR #8483, opened 2025-05-15
[arcilator] Remove JIT Runtime Library

### LINKED ISSUE #8430: [arcillator] Arcillator Can't be built Fully Static
I ran into this when trying to build a fully static firtool while also running tests. The problem that I hit was that arcillator wants to statically link a .so:

```
/usr/lib/gcc/x86_64-alpine-linux-musl/14.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: attempted static link of dynamic object `lib/libarc-jit-env.so'
```

Full build log:

https://github.com/llvm/circt/actions/runs/14545811276/job/40811100200#step:9:4754

fzi-hielscher:
This reverts commit 30ed753d539e981e5914801231e7a2d1c3710806.

Remove the arcilator runtime library for JIT runs for now as the dynamically linked library gets in the way of static builds. Should fix #8430. I do not expect this change to affect any users. 

The purpose of the library in its current form is to allow prints to `stderr` (which is the default for FIRRTL prints) from the compiled hardware model. This cannot be done directly from LLVM IR in a way that is portable across libc implementations. But since there is currently no lowering path for FIRRTL prints, we don't need this at the moment.

I'll open a tracking issue on getting this functionality back in later.


fabianschuiki (approved):
LGTM! Thanks :+1:

seldridge:
Thanks, @fzi-hielscher! I'm expecting to see this now work: https://github.com/llvm/circt/actions/runs/15051167585/job/42306080185 💯

PR #8487, opened 2025-05-17
[ESI][Runtime] Improvements / fixes to utilities/build

teqdruid:
- Updates the IMPORTED_LOCATION for the ESICppRuntime shared library.
- esi-cosim.py:
  - Adds logging of compile commands and introduces warning suppression flags in compile commands.
  - Slow down the simulation to one tick per millisecond if the DEBUG_PERIOD environment variable is not set if debug mode is used.
- Adds the inclusion of the Engines header to the list of runtime headers.


teqdruid:
- Updates the IMPORTED_LOCATION for the ESICppRuntime shared library.
- esi-cosim.py:
  - Adds logging of compile commands and introduces warning suppression flags in compile commands.
  - Slow down the simulation to one tick per millisecond if the DEBUG_PERIOD environment variable is not set if debug mode is used.
- Adds the inclusion of the Engines header to the list of runtime headers.

PR #8489, opened 2025-05-17
[Support] Add relative path caching to InstancePathCache

uenoku:
This commit adds infrastructure for caching relative instance paths between modules in the InstanceGraph. It adds a relativePathsCache field to InstancePathCache to store paths between arbitrary module pairs.

### DIFF
--- a/include/circt/Support/InstanceGraph.h
+++ b/include/circt/Support/InstanceGraph.h
@@ -343,7 +343,7 @@ inline llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
   return os;
 }
 
-/// A data structure that caches and provides absolute paths to module
+/// A data structure that caches and provides paths to module
 /// instances in the IR.
 struct InstancePathCache {
   /// The instance graph of the IR.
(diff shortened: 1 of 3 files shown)


fabianschuiki (approved):
LGTM!

include/circt/Support/InstanceGraph.h
```diff
@@ -343,16 +343,20 @@ inline llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
...
   InstanceGraph &instanceGraph;
 
   explicit InstancePathCache(InstanceGraph &instanceGraph)
       : instanceGraph(instanceGraph) {}
+
+  // Return all absolute paths from the top-level module to the given module.
   ArrayRef<InstancePath> getAbsolutePaths(ModuleOpInterface op);
-  ArrayRef<InstancePath> getAbsolutePaths(ModuleOpInterface op,
+
+  // Return all relative paths from the given module to the given top node.
```
dtzSiFive, line 358: These comments describe paths that go in different directions -- but AFAICT they use the same code and same orientation, is this right?

uenoku: Good point, thank you for catching it!

PR #8492, opened 2025-05-19
[SCFToCalyx] Lower Math AbsFOp

jiahanxie353:
This patch lowers `math.absFOp` by using masks to mask out the sign bit (most significant bit) of a floating point value.


mikeurbach:
Sorry I haven't been helping at all on the latest Calyx developments. Seems most of the review now comes from @cgyurgyik and @rachitnigam. Should we update the CODEOWNERS?

cgyurgyik (approved):

PR #8497, opened 2025-05-20
[circt-lec] Implement emit-smtlib functionality for circt-lec

cowardsa:
To reuse the translate to smtlib MLIR backend and enable benchmark generation for alternative smt solvers, need to lower to and `smt.solver` which does not return any results. To handle this make the LLVM reporting code insertion optional and disable when emitting smtlib. Also in verif to smt lowering only return a result from the solver if the result will be used.

### DIFF
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -93,10 +93,12 @@ struct LogicEquivalenceCheckingOpConversion
       return success();
     }
 
+    auto unusedResult = op.use_empty();
+
     // Solver will only return a result when it is used to check the returned
     // value.
     smt::SolverOp solver;
-    if (op.use_empty())
+    if (unusedResult)
       solver = rewriter.create<smt::SolverOp>(loc, TypeRange{}, ValueRange{});
     else
       solver = rewriter.create<smt::SolverOp>(loc, rewriter.getI1Type(),
@@ -157,10 +159,11 @@ struct LogicEquivalenceCheckingOpConversion
     rewriter.create<smt::AssertOp>(loc, toAssert);
 
     // Fifth, check for satisfiablility and report the result back.
-    // If no operation uses the result of this solver, we leave our check operations empty.
-    // If the result is used, we create a check operation with the result type of
-    // the operation and yield the result of the check operation.
-    if (op.use_empty()) {
+    // If no operation uses the result of this solver, we leave our check
+    // operations empty. If the result is used, we create a check operation with
+    // the result type of the operation and yield the result of the check
+    // operation.
+    if (unusedResult) {
       auto checkOp = rewriter.create<smt::CheckOp>(loc, TypeRange{});
       rewriter.createBlock(&checkOp.getSatRegion());
       rewriter.create<smt::YieldOp>(loc);
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -157,8 +157,8 @@ struct LogicEquivalenceCheckingOpConversion
     rewriter.create<smt::AssertOp>(loc, toAssert);
 
     // Fifth, check for satisfiablility and report the result back.
-    // For unused results, we create a check operation with empty regions.
-    // For used results, we create a check operation with the result type of
+    // If no operation uses the result of this solver, we leave our check operations empty.
+    // If the result is used, we create a check operation with the result type of
     // the operation and yield the result of the check operation.
     if (op.use_empty()) {
       auto checkOp = rewriter.create<smt::CheckOp>(loc, TypeRange{});
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -93,8 +93,15 @@ struct LogicEquivalenceCheckingOpConversion
       return success();
     }
 
-    smt::SolverOp solver =
-        rewriter.create<smt::SolverOp>(loc, rewriter.getI1Type(), ValueRange{});
+    // Solver will only return a result when it is used to check the returned
+    // value.
+    smt::SolverOp solver;
+    if (op.use_empty())
+      solver = rewriter.create<smt::SolverOp>(loc, TypeRange{}, ValueRange{});
+    else
+      solver = rewriter.create<smt::SolverOp>(loc, rewriter.getI1Type(),
+                                              ValueRange{});
+
     rewriter.createBlock(&solver.getBodyRegion());
 
     // First, convert the block arguments of the miter bodies.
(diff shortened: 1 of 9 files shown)


TaoBi22 (changes requested):
Could we also add some tests for this new behaviour? Maybe a filecheck test that the output is as expected with the new pass flags and a test that the expected SMTLIB is emitted for a given LEC problem

lib/Tools/circt-lec/ConstructLEC.cpp
```diff
@@ -69,16 +69,9 @@ void ConstructLECPass::runOnOperation() {
...
-  auto voidTy = LLVM::LLVMVoidType::get(&getContext());
 
-  // Lookup or declare printf function.
-  auto printfFunc = LLVM::lookupOrCreateFn(builder, getOperation(), "printf",
-                                           ptrTy, voidTy, true);
-  if (failed(printfFunc)) {
-    getOperation()->emitError("failed to lookup or create printf");
-    return signalPassFailure();
-  }
+  assert(!(insertMainFunc && (!insertReporting)) &&
```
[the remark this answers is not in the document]
cowardsa: @TaoBi22 - should this be an "getOperation()->emitError( "error message" )", as this isn't really an error with the operation itself - more an error with the pass configurations?

tools/circt-lec/circt-lec.cpp
```diff
@@ -227,6 +228,8 @@ static LogicalResult executeLEC(MLIRContext &context) {
     ConstructLECOptions opts;
     opts.firstModule = firstModuleName;
     opts.secondModule = secondModuleName;
+    if (outputFormat == OutputSMTLIB)
+      opts.insertReporting = false;
```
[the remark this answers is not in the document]
cowardsa: @TaoBi22 - circt-lec does not expose the insertReporting configuration to the user, by default the pass will insert the reporting code, but just for SMTLIB outputs it'll prevent that option - perhaps I'm misunderstanding your suggestion here though? Or I need to add a comment?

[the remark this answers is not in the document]
TaoBi22: Ah no that's my mistake sorry, must have lost track of what file I was looking at! In that case this looks fine

lib/Tools/circt-lec/ConstructLEC.cpp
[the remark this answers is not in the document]
TaoBi22: There's an `mlir::emitError` function that doesn't take an operation - it still requires a location but I would probably just give it the loc of the top-level `builtin.module`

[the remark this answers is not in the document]
cowardsa: Actioned!

tools/circt-lec/circt-lec.cpp
```diff
@@ -254,8 +257,12 @@ static LogicalResult executeLEC(MLIRContext &context) {
 
   if (outputFormat == OutputSMTLIB) {
     auto timer = ts.nest("Print SMT-LIB output");
-    llvm::errs() << "Printing SMT-LIB not yet supported!\n";
-    return failure();
+    auto smtModule =
```
maerhart, line 260: Super-nit: this name is a little misleading, it might suggest that a ModuleOp is returned here, but it's just a LogicalResult. We could just inline this variable.

lib/Tools/circt-lec/ConstructLEC.cpp
```diff
@@ -69,14 +69,9 @@ void ConstructLECPass::runOnOperation() {
...
   Location loc = getOperation()->getLoc();
-  auto ptrTy = LLVM::LLVMPointerType::get(builder.getContext());
-  auto voidTy = LLVM::LLVMVoidType::get(&getContext());
 
-  // Lookup or declare printf function.
-  auto printfFunc = LLVM::lookupOrCreateFn(builder, getOperation(), "printf",
-                                           ptrTy, voidTy, true);
-  if (failed(printfFunc)) {
-    getOperation()->emitError("failed to lookup or create printf");
+  if (insertMainFunc && (!insertReporting)) {
```
maerhart, line 73: Maybe `insertMainFunc` and `insertReporting` should be merged into one enum-option with 3 enumeration cases (no insertion, reporting, reporting+main)? That way the user isn't able to specify such an invalid combination in the first place. There are examples of such enum options in `ArcPasses.td` and `FIRRTL/Passes.td`.

lib/Tools/circt-lec/ConstructLEC.cpp
```diff
@@ -142,14 +150,16 @@ void ConstructLECPass::runOnOperation() {
...
+        lookupOrCreateStringGlobal(builder, getOperation(), "c1 == c2\n");
+    Value neqFormatString =
+        lookupOrCreateStringGlobal(builder, getOperation(), "c1 != c2\n");
+    Value formatString = builder.create<LLVM::SelectOp>(
+        loc, areEquivalent, eqFormatString, neqFormatString);
+    builder.create<LLVM::CallOp>(loc, printfFunc.value(),
+                                 ValueRange{formatString});
+
+    builder.create<func::ReturnOp>(loc, ValueRange{});
+  }
```
maerhart, line 164: Nit: from a code-structure pov it might be better to factor out the common code between the two `insertReporting` if-statements into a separate ConstructLECPass member function and do something like
```
if (!insertReporting) {
  thatHelperFunction(args);
  return;
}

<code currently in first if-statement>
thatHelperFunction(args);
<code currently in second if-statement>
```

test/Tools/circt-lec/construct-lec.mlir
```diff
@@ -57,3 +57,15 @@ hw.module @modB1() {
...
+
+// CHECK2:   [[V0:%.+]] = verif.lec first {
+// CHECK2:   ^bb0([[ARG0:%.+]]: i32, [[ARG1:%.+]]: i32):
+// CHECK2:     [[V1:%.+]] = comb.add [[ARG0]], [[ARG1]]
+// CHECK2:     verif.yield [[V1]]
+// CHECK2:   } second {
+// CHECK2:   ^bb0([[ARG0:%.+]]: i32, [[ARG1:%.+]]: i32):
+// CHECK2:     [[V2:%.+]] = comb.mul [[ARG0]], [[ARG1]]
+// CHECK2:     verif.yield [[V2]]
+// CHECK2:   }
```
maerhart, line 71: AFAICT you are testing the portion of code that is common for both insert-report flag cases. You should probably do some `CHECK2-NOT` checks to make sure the LLVM operations that flag is skipping are not present.

lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -150,21 +159,41 @@ struct LogicEquivalenceCheckingOpConversion
...
+      rewriter.createBlock(&checkOp.getUnknownRegion());
+      rewriter.create<smt::YieldOp>(loc, falseVal);
+      rewriter.createBlock(&checkOp.getUnsatRegion());
+      rewriter.create<smt::YieldOp>(loc, trueVal);
+      rewriter.setInsertionPointAfter(checkOp);
+      rewriter.create<smt::YieldOp>(loc, checkOp->getResults());
+
+      rewriter.replaceOp(op, solver->getResults());
+    }
+
```
maerhart, line 196: It seems like there's no unit test for the changes in this file (only the integration test, thus not super critical but still good to have).

maerhart (approved):
Thanks a lot for working on that! 🎉

cowardsa:
If satisfied with these changes then perhaps someone can merge the changes when time permits

test/Tools/circt-lec/construct-lec.mlir
```diff
@@ -57,3 +57,18 @@ hw.module @modB1() {
...
+
+// CHECK2-NOT: llvm.mlir.global private constant @"c1 == c2\0A"("c1 == c2\0A\00") {addr_space = 0 : i32}
+// CHECK2-NOT: llvm.mlir.global private constant @"c1 != c2\0A"("c1 != c2\0A\00") {addr_space = 0 : i32}
```
TaoBi22, line 74: Nit: missing EOF newline

test/Conversion/VerifToSMT/verif-to-smt.mlir
```diff
@@ -83,8 +83,26 @@ func.func @test_lec(%arg0: !smt.bv<1>) -> (i1, i1, i1) {
...
+  // CHECK:   smt.assert [[EQ3]]
+  // CHECK:   smt.check sat {
+  // CHECK:   } unknown {
+  // CHECK:   } unsat {
+  // CHECK:   }
+  // CHECK: }
+
   // CHECK: return [[EQ]], [[EQ2]], %true
   return %1, %2, %3 : i1, i1, i1
+  // return %3, %3, %3 : i1, i1, i1
```
TaoBi22, line 105: Nit: Think this is here unintentionally

lib/Tools/circt-lec/ConstructLEC.cpp
```diff
@@ -65,20 +67,43 @@ hw::HWModuleOp ConstructLECPass::lookupModule(StringRef name) {
   return cast<hw::HWModuleOp>(expectedModule);
 }
 
+Value ConstructLECPass::constructMiter(OpBuilder builder, Location loc,
+                                       hw::HWModuleOp moduleA,
+                                       hw::HWModuleOp moduleB) {
+  // Create the miter circuit that return equivalence result.
```
TaoBi22, line 73: nit:
```suggestion
  // Create the miter circuit that returns equivalence result.
```

lib/Tools/circt-lec/ConstructLEC.cpp
```diff
@@ -94,13 +119,30 @@ void ConstructLECPass::runOnOperation() {
...
+  }
+
+  // Reuse the name of the first module for the entry function, so we don't
+  // have to do any uniquing and the LEC driver also already knows this name.
   FunctionType functionType = FunctionType::get(&getContext(), {}, {});
   func::FuncOp entryFunc =
       builder.create<func::FuncOp>(loc, firstModule, functionType);
 
-  if (insertMainFunc) {
+  if (insertMode == lec::InsertAdditionalModeEnum::Main) {
```
TaoBi22, line 145: Probably needs to wrap L164-173

TaoBi22 (approved):
Couple of tiny nits then happy to hit merge!

test/Tools/circt-lec/construct-lec.mlir
maerhart, line 74: This should be enough for what this test is supposed to do and is more maintainable:
```suggestion
// CHECK2-NOT: func.func
// CHECK2-NOT: llvm.mlir.global
```

maerhart (approved):

lib/Tools/circt-lec/ConstructLEC.cpp
maerhart: I don't think so, because the lines you mentioned should be present in 'reporting' mode

TaoBi22: Oops this was an accidental comment, was a reply to another one I deleted!

PR #8498, opened 2025-05-21
[LLHD] Fix handling of conditional drives

### LINKED ISSUE #8494: [LLHD] Mem2Reg does not properly combine enables of successive drives
Consider the following input:
```mlir
hw.module @Foo(in %u: i42, in %v: i42, in %w: i42, in %q: i1, in %r: i1) {
  %a = llhd.sig %u : i42
  llhd.process {
    llhd.drv %a, %u after %0 : !hw.inout<i42>
    llhd.drv %a, %v after %0 if %q : !hw.inout<i42>
    llhd.drv %a, %w after %0 if %r : !hw.inout<i42>
    llhd.halt
  }
}
```
Running this through `circt-opt --llhd-mem2reg` produces this result:
```mlir
hw.module @Foo(in %u : i42, in %v : i42, in %w : i42, in %q : i1, in %r : i1) {
  %a = llhd.sig %u : i42
  llhd.process {
    llhd.drv %a, %w after %0 if %r : !hw.inout<i42>
    llhd.halt
  }
}
```
Only the final drive is retained in the output. Instead, the final value should be `w` if `r`, or `v` if `q&!r`, or `u` otherwise.

fabianschuiki:
The Mem2Reg pass in LLHD does not correctly handle multiple conditional drives and mixes of unconditional and conditional drives to the same signal. In its current implementation, a conditional drive completely replaces any previous drive as if it were unconditional. This is not correct, since a conditional drive only supersedes a prior drive if it is enabled. Similarly, a conditional drive to a signal that is later probed has to insert a mux to either forward the new value if the drive is enabled, or the current value of the signal if it is disabled.

This commit extends the Mem2Reg pass to handle these situations properly:

The backward propagation across the lattice is updated to forward the need for a definition in case the drive is conditional. This is required since the drive being disabled means that a previous drive or probe must be able to provide a definition.

The forward propagation across the lattice is updated to only kill incoming reaching definitions when a drive is unconditional. If it is conditional, the drive has to insert a multiplexer between the incoming definition if the drive is disabled, or the driven value if the drive is enabled.

New `resolveDefinitions` methods on `DriveNode`s insert multiplexers as necessary and aggregates drive conditions using OR ops. This ensures that multiple conditional drives in sequence collapse to a single drive where the driven value is selected with a multiplexer and the enable condition is derived as the OR of all conditional drives before the collapse.

An additional lit test covers the problematic cases that were not handled properly before.

Fixes #8494


lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
```diff
@@ -1255,12 +1313,96 @@ void Promoter::resolveDefinitions(ProbeNode *node) {
...
+       node->def->conditionIsPlaceholder))
+    resolveDefinitionCondition(node);
+}
+
+void Promoter::resolveDefinitionValue(DriveNode *node) {
+  // Get the slot definition reaching the drive. This is the value the drive has
+  // to update.
+  auto slot = getSlot(node->slot);
+  auto *inDef = node->valueBefore->reachingDefs.lookup(blockingSlot(slot));
+  assert(inDef && "no definition reaches drive");
```
maerhart, line 1347: This is guarded by the `if (!slotOrder.contains(getSlot(node->slot))) return;` above to not be reached by non-blocking drives, right?

maerhart (approved):
LGTM

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
fabianschuiki: Oh excellent catch. This should actually be `node->slot`, not `blockingSlot(slot)`. I'll add a test for that.

I think the `!slotOrder.contains(...)` can actually be dropped, because the pass only creates lattice nodes for drives that target promotable slots. So this check will always pass.

PR #8502, opened 2025-05-21
[HW] Merge basic round-trip tests; NFC

fabianschuiki:
The HW dialect has quite a few of its basic round-tripping tests split out across multiple files. Merge the `bitwise.mlir` and `round-trip.mlir` files into `basic.mlir`, and make use of the new `--verify-roundtrip` flag which also removes then need for explicit FileChecks.


seldridge (approved):
LGTM

Feel free to land this type of thing whenever you find it. I think there is some similar scattering of tests in FIRRTL dialect as well.

fabianschuiki:
👍 Sounds good, I can do that! The new roundtrip verifier also allows us to get rid of all FileCheck lines in these files. I've been reluctant to delete that much code silently, but it might be worth doing. Otherwise people feel like they still have to add FileCheck comments when they add roundtrip tests for new ops.

PR #8503, opened 2025-05-21
[HW] Add array element injection op

fabianschuiki:
Add a new `hw.array_inject` operation which updates a single element in an array value and returns the updated array. This fills in a gap where structs had a `struct_extract` and `struct_inject` op available, but arrays only had `array_get` without the complementary `array_inject`.

This commit also adds a `HWCanonicalization.td` file for any canonicalization rewrite patterns on HW ops. It is overkill for just the new `array_inject` operation, but my hope is that we'll fill this in as time progresses and we may even be able to migrate some existing patterns from C++ to the declarative format. The
`ArithCanonicalization.td` file has a lot of nice inspiration.

In the future we may want to add patterns that can fold injects directly into an `array_create` operation, and patterns that reorder injects and collapse adjacent injects into an `array_create`, `array_slice`, and `array_concat`.

### DIFF
--- a/include/circt/Dialect/HW/HWAggregates.td
+++ b/include/circt/Dialect/HW/HWAggregates.td
@@ -181,6 +181,33 @@ def ArrayGetOp : HWOp<"array_get", [
   let hasCanonicalizeMethod = 1;
 }
 
+def ArrayInjectOp : HWOp<"array_inject", [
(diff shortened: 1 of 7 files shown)


fabianschuiki:
Closing this because of the CI failure. Have hit this in the past. Re-opening the PR usually helps. Thanks Microsoft.

PR #8504, opened 2025-05-22
[HW] Add array element injection op

fabianschuiki:
Add a new `hw.array_inject` operation which updates a single element in an array value and returns the updated array. This fills in a gap where structs had a `struct_extract` and `struct_inject` op available, but arrays only had `array_get` without the complementary `array_inject`.

This commit also adds a `HWCanonicalization.td` file for any canonicalization rewrite patterns on HW ops. It is overkill for just the new `array_inject` operation, but my hope is that we'll fill this in as time progresses and we may even be able to migrate some existing patterns from C++ to the declarative format. The `ArithCanonicalization.td` file has a lot of nice inspiration.

In the future we may want to add patterns that can fold injects directly into an `array_create` operation, and patterns that reorder injects and collapse adjacent injects into an `array_create`, `array_slice`, and `array_concat`.

### DIFF
--- a/lib/Dialect/HW/HWOps.cpp
+++ b/lib/Dialect/HW/HWOps.cpp
@@ -255,6 +255,14 @@ void HWModulePortAccessor::setOutput(StringRef name, Value v) {
 }
 
 //===----------------------------------------------------------------------===//
+// Declarative Canonicalization Patterns
+//===----------------------------------------------------------------------===//
+
+namespace {
+#include "circt/Dialect/HW/HWCanonicalization.cpp.inc"
+} // namespace
+
+//===----------------------------------------------------------------------===//
 // ConstantOp
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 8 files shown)


fabianschuiki:
Ugh, no idea why this sends the compiler into an infinite recursion. I've had this randomly pop up in some other PR a while ago. The fix there was to wait for someone else to push something onto main and then re-open the PR. Doesn't seem to work here.

On top, I'm debugging an issue where that canonicalization pattern puts the canonicalizer into an infinite loop -- no idea how that comes about.

lib/Dialect/HW/HWOps.cpp
```diff
@@ -2855,6 +2863,11 @@ OpFoldResult ArrayGetOp::fold(FoldAdaptor adaptor) {
                                        intTy.getIntOrFloatBitWidth()));
   }
 
+  // array_get(array_inject(_, index, element), index) -> element
+  if (auto inject = getInput().getDefiningOp<ArrayInjectOp>())
+    if (getIndex() == inject.getIndex())
+      return inject.getElement();
+
```
uenoku, line 2870: Do we have a test for this?

uenoku (approved):
LGTM. Since this is not standard SV operation I think we will need legalization for hw.array_inject but it can be done in a follow-up.

lib/Dialect/HW/HWOps.cpp
fabianschuiki: Yes sir :saluting_face: https://github.com/llvm/circt/pull/8504/files#diff-5515faaa7267def62dc25befc3df06e49ea0af65cbc2cc34e75687998276a973R1085-R1091

Issue #7412, opened 2024-07-31
[PyCDE] Enable client code to type check properly


teqdruid:
As of today, it is impossible to use the pycde API in a way which pleases python type checkers (e.g. mypy). Make this possible. Will require API changes. Not sure that this is even possible given that we parameterize classes/functions so much.

Different from #3581 in that it doesn't require generating pybind11 stubs. That was only for internal development.

PR #8507, opened 2025-05-22
[PyCDE] Fix invalid `inputs` function

mortbopet:
This might just have been some outdated code.

The previous `inputs` function had an incorrect type annotation (i.e. it defers to the following code:
```python
  @property
  def inputs(self) -> List[Input]:
    return [p for p in self.ports if isinstance(p, Input)]
```
Might as well return that list instead of reducing it to a dictionary.

Secondly, also adds a (presumably) missing `outputs` method.

### DIFF
--- a/frontends/PyCDE/src/pycde/module.py
+++ b/frontends/PyCDE/src/pycde/module.py
@@ -661,11 +661,12 @@ class Module(_PyProxy, metaclass=ModuleLikeType):
     cls._builder.print(out)
 
   @classmethod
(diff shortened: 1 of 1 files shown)


teqdruid:
> Secondly, also adds a (presumably) missing `outputs` method.

I didn't add it since it wasn't needed and is thus untested.

teqdruid (approved):
Yeah, there are a bunch of incorrect type hints and missing type hints all over PyCDE. See #7412

PR #8509, opened 2025-05-22
[LLHD] Add support for `llhd.sig.array_get` projections to Mem2Reg

fabianschuiki:
Extend the Mem2Reg pass to support drives to `llhd.sig.array_get` projections into a promotable signal. This requires a few changes:

- When checking whether a signal is promotable, we have to recursively check its users and look through `llhd.sig.array_get` ops to see if all uses are safe for promotion.

- The lattice used to propagate needed and reaching definitions has to track the projection being driven separately from the underlying root slot being driven.

- A new set of helpers to unpack and pack stacks of projections help probes and drives take a promotable slot and its value, descend into subfields as indicated by a set of projection ops ("unpack"), and then optionally mutate the value of a subfield and zip the updated value back up to the promotable slot ("pack").

- During backward and forward propagation across the lattice, drives to projections have to create the need for a definition of the slot to be available because they have to take the slot's current value in order to mutate the projected field.

- The functions that resolve definitions have to unpack a slot's value according to the projections between the probe and the slot. In case of a drive they also have to mutate the projected field value after unpacking and then pack everything back op in order to create a new reaching definition for the entire slot to be forwarded to readers.

This commit also drops `slotOrder` in favor of a `promotable` set of values. The actual order of slots was never used.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
+++ b/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
@@ -207,8 +207,11 @@ static DefSlot blockingSlot(Value slot) { return {slot, 0}; }
 static DefSlot delayedSlot(Value slot) { return {slot, 1}; }
 static Value getSlot(DefSlot slot) { return slot.getPointer(); }
 static bool isDelayed(DefSlot slot) { return slot.getInt(); }
+static Type getStoredType(Value slot) {
+  return cast<hw::InOutType>(slot.getType()).getElementType();
+}
 static Type getStoredType(DefSlot slot) {
-  return cast<hw::InOutType>(slot.getPointer().getType()).getElementType();
+  return getStoredType(slot.getPointer());
 }
 static Location getLoc(DefSlot slot) { return slot.getPointer().getLoc(); }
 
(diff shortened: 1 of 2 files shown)


test/Dialect/LLHD/Transforms/mem2reg.mlir
```diff
@@ -645,5 +645,132 @@ hw.module @DelayedConditionalDrives(in %u: i42, in %v: i42, in %w: i42, in %q: i
...
+    %1 = llhd.sig.array_get %a[%i] : !hw.inout<array<4xi42>>
+    // CHECK-NEXT: [[TMP:%.+]] = hw.array_inject [[A]][%i], %v
+    llhd.drv %1, %v after %0 : !hw.inout<i42>
+    // CHECK-NEXT: llhd.constant_time
+    // CHECK-NEXT: llhd.drv %a, [[TMP]] after {{%.+}}
+    // CHECK-NEXT: llhd.halt
+    llhd.halt
+  }
+}
+
```
maerhart, line 773: Maybe add tests for the following scenarios:
* Projected signal value is passed via block argument within a process (make sure it doesn't crash)
* Multiple projection ops refer to the same part of the original signal (make sure multiple drives to these different projections that refer to the same part update to the correct value)

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
```diff
@@ -669,32 +758,76 @@ LogicalResult Promoter::promote() {
...
+        if (region.isProperAncestor(user->getParentRegion()))
+          return false;
+        // Ignore uses outside of the region.
+        if (user->getParentRegion() != &region)
+          return true;
+        // Projection operations are okay as long as they are in the same block
+        // as any of their users.
+        if (isa<SigArrayGetOp>(user)) {
+          for (auto *projectionUser : user->getUsers()) {
+            if (projectionUser->getBlock() != user->getBlock())
```
maerhart, line 787: Isn't that too restrictive for many practical cases due to CSE of projection ops? (not blocking)

maerhart (approved):

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
fabianschuiki: Yes that's a good point 👍. We should be able to relax this by only requiring that the projection op dominates its users. That would also allow projections in dominating predecessor blocks and in the module to be covered, but still allow us to not reason about signals references through block arguments (that'll be a mighty headache).

PR #8510, opened 2025-05-22
[Arc] Fix arc inliner deleting arcs that still have calls

TaoBi22:
Hit a very niche breaking case in InlineArcs in my optimization travels - currently if the inliner inlines `arcA` which calls `arcB`, it will only acknowledge the additional user of `arcB` if `arcA` is inlined into another define. Most of the time this flies under the radar, however, if `arcB` is prevented from being inlined by another user that is eventually deleted, the inliner will think that `arcB` has no users even though it still has a call where `arcA` was inlined. Breaking example added as a filecheck test.


fabianschuiki (approved):
Yikes, that's an excellent catch. Thanks a lot :partying_face:

maerhart (approved):
Thanks!

PR #8511, opened 2025-05-22
[LowerToHW] Emit an error when $fopen failed for FD lowering

uenoku:
Some simulators silently return 0 when $fopen failed to open a file. This PR changes the FileDescriptor::get to emit $error explicitly.


seldridge (approved):
LGTM

PR #8518, opened 2025-05-29
[SV] Enhanced name validation: support IEEE standard escape name checking

RonxBulld:
# Change Overview
This PR updates the name validation mechanism in the SV dialect, mainly modifying the `isNameValid` function in the `SVDialect.h` and `LegalizeNames.cpp` files to support escaped name checking in accordance with the IEEE standard.

# Change Reason

1. **IEEE Standard Compliance**: The original name validation function lacks the correct handling of escaped identifiers in the IEEE Verilog standard. The IEEE 1800 standard allows backslash-escaped identifiers (such as `\escaped_name`), which are common in Verilog code generation.

2. **Function Enhancement**: By adding a new `allowEscapedName` parameter, the `isNameValid` function can decide whether to allow escaped names based on the context, providing a more flexible name validation strategy.

3. **Backward compatibility**: The modification maintains compatibility with existing APIs while expanding functionality so that callers can choose whether to enable escape name support as needed.

4. **Code generation quality**: Correct handling of escape names can avoid illegal identifiers during Verilog code generation, improving the quality and standard compliance of generated code.

# Major changes
- Modify `isNameValid` function signature and add `allowEscapedName` parameter
- Update related call sites to pass new parameters
- Enhance name validation logic to support IEEE standard escape identifier format

### DIFF
--- a/include/circt/Dialect/SV/SVDialect.h
+++ b/include/circt/Dialect/SV/SVDialect.h
@@ -45,7 +45,10 @@ StringRef legalizeName(llvm::StringRef name,
 ///
 /// Call \c legalizeName() to obtain a legal version of the name.  If \p
 /// caseInsensitive is true, then the check will be done case insensitively.
-bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords);
+/// If \p allowEscapedName is true, IEEE-compliant escaped names are allowed to
+/// pass the check.
+bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords,
+                 bool allowEscapedName = false);
 
 } // namespace sv
 } // namespace circt
--- a/include/circt/Dialect/SV/SVDialect.h
+++ b/include/circt/Dialect/SV/SVDialect.h
@@ -48,7 +48,7 @@ StringRef legalizeName(llvm::StringRef name,
 /// If \p allowEscapedName is true, IEEE-compliant escaped names are allowed to
 /// pass the check.
 bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords,
-                 bool allowEscapedName = false);
+                 bool allowEscapedName = true);
 
 } // namespace sv
 } // namespace circt
--- a/include/circt/Dialect/SV/SVDialect.h
+++ b/include/circt/Dialect/SV/SVDialect.h
@@ -47,8 +47,7 @@ StringRef legalizeName(llvm::StringRef name,
 /// caseInsensitive is true, then the check will be done case insensitively.
 /// If \p allowEscapedName is true, IEEE-compliant escaped names are allowed to
 /// pass the check.
-bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords,
-                 bool allowEscapedName = true);
+bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords);
 
 } // namespace sv
 } // namespace circt
--- a/lib/Dialect/SV/SVDialect.cpp
+++ b/lib/Dialect/SV/SVDialect.cpp
@@ -166,8 +166,8 @@ bool circt::sv::isNameValid(StringRef name, bool caseInsensitiveKeywords,
     // Check if the name is a valid escaped name.
     if (name.size() < 2)
       return false;
-    for (size_t i = 1; i < name.size(); ++i) {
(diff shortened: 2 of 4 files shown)


include/circt/Dialect/SV/SVDialect.h
```diff
@@ -45,7 +45,10 @@ StringRef legalizeName(llvm::StringRef name,
 ///
 /// Call \c legalizeName() to obtain a legal version of the name.  If \p
 /// caseInsensitive is true, then the check will be done case insensitively.
-bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords);
+/// If \p allowEscapedName is true, IEEE-compliant escaped names are allowed to
+/// pass the check.
+bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords,
+                 bool allowEscapedName = false);
```
seldridge, line 51: Escaped names are legal in both Verilog (1364-1995 2.7.1) and SystemVerilog (1800-2023 5.6.1). I don't see a good reason to make this optionally allowed given that (I think) all Verilog versions support escaped identifiers. The only difference is on what the identifiers are.

seldridge (commented):
Thanks for starting this. I've actually wanted to use this in a couple places and thought it would require way more effort to get working.

Can you include at least one test of this behavior? E.g., this might be a good place: https://github.com/llvm/circt/blob/main/test/Conversion/ExportVerilog/name-legalize.mlir

include/circt/Dialect/SV/SVDialect.h
RonxBulld: > Escaped names are legal in both Verilog (1364-1995 2.7.1) and SystemVerilog (1800-2023 5.6.1). I don't see a good reason to make this optionally allowed given that (I think) all Verilog versions support escaped identifiers. The only difference is on what the identifiers are.

Yes, I concur with your viewpoint and believe that this feature should be enabled by default. Given that my understanding of CIRCT is not particularly in-depth, I am not entirely certain whether this modification might cause any side effects. Out of caution, I have temporarily set it as an optional feature, hoping that it will at least meet my needs. After extensive validation, you, I, or anyone else can set it as a default feature, which I consider to be a prudent choice.

RonxBulld:
> Thanks for starting this. I've actually wanted to use this in a couple places and thought it would require way more effort to get working.
> 
> Can you include at least one test of this behavior? E.g., this might be a good place: https://github.com/llvm/circt/blob/main/test/Conversion/ExportVerilog/name-legalize.mlir

Thank you for your reminder. I will add some tests at a later time.

include/circt/Dialect/SV/SVDialect.h
uenoku: +1, I think default is better.

include/circt/Dialect/SV/SVDialect.h
```diff
@@ -45,7 +45,10 @@ StringRef legalizeName(llvm::StringRef name,
 ///
 /// Call \c legalizeName() to obtain a legal version of the name.  If \p
 /// caseInsensitive is true, then the check will be done case insensitively.
-bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords);
+/// If \p allowEscapedName is true, IEEE-compliant escaped names are allowed to
+/// pass the check.
+bool isNameValid(llvm::StringRef name, bool caseInsensitiveKeywords,
+                 bool allowEscapedName = true);
```
uenoku: I think it's always true, correct? In that case please drop the argument.

lib/Dialect/SV/SVDialect.cpp
```diff
@@ -158,9 +158,21 @@ StringRef circt::sv::legalizeName(StringRef name,
...
     return false;
+  if (allowEscapedName && name.front() == '\\') {
+    // Check if the name is a valid escaped name.
+    if (name.size() < 2)
+      return false;
+    for (size_t i = 1; i < name.size(); ++i) {
+      bool isValidChar = llvm::isPrint(name[i]) && !llvm::isSpace(name[i]);
+      if (!isValidChar)
+        return false;
+    }
```
uenoku: super nit:
```suggestion
    for (auto ch: name.drop_front()) {
      bool isValidChar = llvm::isPrint(ch) && !llvm::isSpace(ch);
      if (!isValidChar)
        return false;
    }
```

uenoku (approved):
Few nits, but LGTM otherwise.

uenoku:
Thank you for the PR!

PR #8522, opened 2025-05-30
[Comb][AIG] Move rewriter helpers to support and use it for AIG

uenoku:
Move name propagation helpers from CombFolds to Support/Naming.h and rename them to better reflect their purpose of propagating "sv.namehint" attributes. Update all callers. No functional change for Comb.


uenoku:
This is NFC for Comb so I'm going to  merge the PR.

PR #8523, opened 2025-05-30
[CombToAIG] Implement Wallace Tree Multiplier

uenoku:
Replace the naive multiplier implementation with an efficient Wallace Tree multiplier (https://en.wikipedia.org/wiki/Wallace_tree). This implementation organizes partial products by bit position and uses a proper Wallace tree reduction algorithm that processes all columns in parallel at each stage.

The Wallace tree algorithm reduces the height of the entire tree in each stage using full adders for groups of 3 bits, resulting in a multiplier with linear log delay.

Correctness is verified by LEC.


sequencer:
I am considering whether it is feasible to integrate synthesis with a hardware generator callout.
Implementing different arithmetic hardware generators using Chisel should significantly simplify the generator writing experience.
See https://github.com/sequencer/arithmetic/tree/master/arithmetic/src

On the other hand, AIG algorithm for different FPGA's might have different optimized circuit due to the existence of "LUT-LUT cascade paths", see https://docs.amd.com/r/en-US/am005-versal-clb/Differences-from-Previous-Generations

Anyway, What I'm thinking is the `CombToAIG` pass should be optimized for different FPGA backend(and even ASIC backend!) but it should be device and physical aware.

uenoku:
> I am considering whether it is feasible to integrate synthesis with a hardware generator callout.
Implementing different arithmetic hardware generators using Chisel should significantly simplify the generator writing experience.

Interesting idea but it's fairly challenging and complicated to use a different language frontend in CIRCT pass. 
If a chisel user wants to use a specific multiplier (booth, wallace etc) I think it's much better to just switch at Chisel level. 

I agree that we do need to switch lowering patterns based on targets, but we don't have a good way to describe target yet. My current focus is to improve the baseline for target-independent lowering.

teqdruid:
> Interesting idea but it's fairly challenging and complicated to use a different language frontend in CIRCT pass.

It's actually not _too_ complicated to use PyCDE for this assuming that PyCDE is driving the flow. In fact I do exactly this for ESI Service Implementation generators:

https://github.com/llvm/circt/blob/c96f9a852dae411899b53bf7cdbc508c3932d414/lib/Bindings/Python/ESIModule.cpp#L63
https://github.com/llvm/circt/blob/c96f9a852dae411899b53bf7cdbc508c3932d414/frontends/PyCDE/src/pycde/esi.py#L411
https://github.com/llvm/circt/blob/c96f9a852dae411899b53bf7cdbc508c3932d414/lib/Dialect/ESI/ESIServices.cpp#L333

I've been meaning to add this capability to PyCDE so that users can override "default" lowerings.

uenoku:
> It's actually not too complicated to use PyCDE for this assuming that PyCDE is driving the flow. In fact I do exactly this for ESI Service Implementation generators:

Interesting, thank you for the reference! That looks like something we want to do eventually in circt-synth flow.

lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -674,40 +691,109 @@ struct CombMulOpConversion : OpConversionPattern<MulOp> {
...
-          op.getLoc(), andBit, 0, width - i);
-      if (i == 0) {
-        results.push_back(upperBits);
-        continue;
-      }
+    // Skip a zero width value.
+    if (width == 0) {
+      rewriter.replaceOpWithNewOp<hw::ConstantOp>(op, op.getType(), 0);
+      return success();
+    }
```
TaoBi22, line 704: Could we have a test for this case?

lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -674,40 +691,109 @@ struct CombMulOpConversion : OpConversionPattern<MulOp> {
...
+
+      partialProducts.push_back(row);
     }
 
-    rewriter.replaceOpWithNewOp<comb::AddOp>(op, results, true);
+    // If the width is 1, we are done.
+    if (width == 1) {
+      rewriter.replaceOp(op, partialProducts[0][0]);
+      return success();
+    }
```
TaoBi22, line 730: I don't think we have a test for this case either.

test/Conversion/CombToAIG/comb-to-aig-arith.mlir
```diff
@@ -60,23 +60,31 @@ hw.module @sub(in %lhs: i4, in %rhs: i4, out out: i4) {
...
   %0 = comb.mul %lhs, %rhs : i2
   hw.output %0 : i2
 }
 
+// CHECK-LABEL: @mul_17
+hw.module @mul_17(in %lhs: i17, in %rhs: i17, out out: i17) {
+  %0 = comb.mul %lhs, %rhs : i17
+  hw.output %0 : i17
+}
+
```
TaoBi22: What's the purpose of this test?

lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -668,6 +668,23 @@ struct CombSubOpConversion : OpConversionPattern<SubOp> {
...
+      rewriter.createOrFold<comb::XorOp>(loc, ArrayRef<Value>{a, b, c}, true);
+
+  auto carry = rewriter.createOrFold<comb::OrOp>(
+      loc,
+      ArrayRef<Value>{rewriter.createOrFold<comb::AndOp>(loc, a, b),
+                      rewriter.createOrFold<comb::AndOp>(loc, b, c),
+                      rewriter.createOrFold<comb::AndOp>(loc, c, a)},
+      true);
+
+  return {sum, carry};
```
fabianschuiki: I've seen alternate forms that use an intermediate XOR result to get rid of one of the AND gates:
```
tmp = a ^ b
sum = tmp ^ c
carry = (a & b) | (tmp & c)
```
I don't know if this actually simplifies anything in practice. Maybe it's useful to have only 2-input gates? No idea, just a random thought 😃

test/Conversion/CombToAIG/comb-to-aig-arith.mlir
uenoku: It checks it at least doesn't crash for larger width

uenoku (commented):
Thank you for the review @TaoBi22 !

lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -677,40 +694,110 @@ struct CombMulOpConversion : OpConversionPattern<MulOp> {
...
+
+    // Wallace tree reduction
+    replaceOpAndCopyNamehint(
+        rewriter, op,
+        wallaceReduction(falseValue, width, rewriter, loc, partialProducts));
     return success();
   }
+
+private:
+  // Perform Wallace tree reduction on partial products
```
fabianschuiki: Might be useful for readers to link to the [Wikipedia article](https://en.wikipedia.org/wiki/Wallace_tree) here. This is a nice compression scheme!

test/Conversion/CombToAIG/comb-to-aig-arith.mlir
```diff
@@ -60,23 +60,47 @@ hw.module @sub(in %lhs: i4, in %rhs: i4, out out: i4) {
...
+// ALLOW_ADD-NEXT:    %[[P_0_1:.+]] = comb.and %[[LHS_0]], %[[RHS_1]] : i1
+// ALLOW_ADD-NEXT:    %[[PARTIAL_SUM_0:.+]] = comb.concat %[[P_1_0]], %[[P_0_0]] : i1, i1
+// ALLOW_ADD-NEXT:    %[[PARTIAL_SUM_1:.+]] = comb.concat %[[P_0_1]], %[[FALSE]] : i1, i1
+// ALLOW_ADD-NEXT:    %[[RESULT:.+]] = comb.add bin %[[PARTIAL_SUM_0]], %[[PARTIAL_SUM_1]] : i2
+// ALLOW_ADD-NEXT:    hw.output %[[RESULT]] : i2
+// ALLOW_ADD-NEXT:  }
 hw.module @mul(in %lhs: i2, in %rhs: i2, out out: i2) {
   %0 = comb.mul %lhs, %rhs : i2
   hw.output %0 : i2
 }
```
fabianschuiki, line 93: I'm wondering if it makes sense to have a test for a multiplication that is complex enough to trigger the full adder logic. Probably a bit overkill, given that you run this through LEC. But it might be useful to ensure that the pass doesn't implement any multiplication, but specifically that Wallace tree reduction scheme. Although, thinking about it, _not_ testing it specifically might be a good thing too 😬

lib/Conversion/CombToAIG/CombToAIG.cpp
uenoku: Good point!

test/Conversion/CombToAIG/comb-to-aig-arith.mlir
uenoku: Yeah that's good point. I changed to use i3 which uses full-adder.

lib/Conversion/CombToAIG/CombToAIG.cpp
cowardsa: Most modern standard cell libraries include custom "majority gates" to compute the carry-out and may also include complete full-adder cells - so depends how you want to attack the tech-mapping problem

PR #8527, opened 2025-06-02
[ESI] Add get_cmake_dir to package

mortbopet:
Just as with PyBind, this helps massively with making it easier to locate the CMake files for the ESI runtime, when we're distributing it as a pip package.


teqdruid:
Are you looking for https://github.com/llvm/circt/blob/a1d66ba1349904ca867f957f3246ab86080c1d44/lib/Dialect/ESI/runtime/python/esiaccel/utils.py#L35 ?

Feel free to refactor this.

PR #8529, opened 2025-06-03
[AIG] Add LongestPathAnalysis implementation

uenoku:
This patch implements the LongestPathAnalysis for the AIG dialect, which calculates the maximum delay through combinational paths in a circuit where each AIG operation has a unit delay. The implementation:

- Adds the core analysis infrastructure with three main components:
  - Context: Thread-safe interface to access analysis results across modules
  - LocalVisitor: Per-module visitor that computes paths within module boundaries
  - Impl: Top-level implementation that orchestrates analysis across hierarchies
- Implements hierarchical analysis that works across module boundaries
- Supports parallel execution for improved performance on large designs
- Provides both closed paths (register-to-register) and open paths analysis
- Includes debug point tracing for detailed path information
- Adds a PrintLongestPathAnalysis pass for outputting results

The analysis is designed to be scalable for large designs through techniques like redundant path elimination, graph parallelism, and early pruning of paths. It's particularly useful for identifying critical paths, estimating circuit performance, and guiding optimization passes.

### DIFF
--- a/lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
+++ b/lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
@@ -64,6 +64,7 @@
 #include "llvm/Support/Mutex.h"
 #include "llvm/Support/ToolOutputFile.h"
 #include "llvm/Support/raw_ostream.h"
+#include <condition_variable>
 #include <cstdint>
 #include <memory>
 #include <mutex>
@@ -267,7 +268,8 @@ class LocalVisitor;
 //===----------------------------------------------------------------------===//
 // Context
 //===----------------------------------------------------------------------===//
-// This class provides a thread-safe interface to access the analysis results.
+/// This class provides a thread-safe interface to access the analysis results.
+
 class Context {
 public:
   Context(igraph::InstanceGraph *instanceGraph,
@@ -460,6 +462,8 @@ private:
 
   // This is set to true when the thread is done.
   std::atomic_bool done;
+  mutable std::condition_variable cv;
+  mutable std::mutex mutex;
 
   // A flag to indicate the module is top-level.
   bool topLevel = false;
@@ -482,8 +486,7 @@ ArrayRef<OpenPath> LocalVisitor::getResults(Value value, size_t bitPos) const {
   if (leader != ec.member_end()) {
     if (*leader != valueAndBitPos) {
       // If this is not the leader, then use the leader.
-      auto root = ecMap.at(*leader);
-      return getResults(root.first, root.second);
+      return getResults(leader->first, leader->second);
     }
   }
 
@@ -505,8 +508,8 @@ void LocalVisitor::putUnclosedResult(const Object &object, int64_t delay,
 
 void LocalVisitor::waitUntilDone() const {
   // Wait for the thread to finish.
-  while (!done.load())
-    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  std::unique_lock<std::mutex> lock(mutex);
+  cv.wait(lock, [this] { return done.load(); });
 }
 
 LogicalResult LocalVisitor::markRegFanOut(Value fanOut, Value start,
@@ -557,9 +560,7 @@ LogicalResult LocalVisitor::markEquivalent(Value from, size_t fromBitPos,
                                            Value to, size_t toBitPos,
                                            SmallVectorImpl<OpenPath> &results) {
   auto leader = ec.getOrInsertLeaderValue({to, toBitPos});
-  auto &slot = ecMap[leader];
-  if (!slot.first)
-    slot = {to, toBitPos};
+  (void)leader;
   // Merge classes, and visit the leader.
   auto newLeader = ec.unionSets({to, toBitPos}, {from, fromBitPos});
   assert(leader == *newLeader);
@@ -646,13 +647,9 @@ LogicalResult LocalVisitor::visit(hw::InstanceOp op, size_t bitPos,
   if (!ctx->instanceGraph)
     return markFanIn(value, bitPos, results);
 
-  // If an instance graph is available, then we can look up the module. It's an
-  // error when the module is not found.
+  // If an instance graph is available, then we can look up the module.
   auto *node = ctx->instanceGraph->lookup(moduleName);
-  if (!node || !node->getModule()) {
-    op.emitError() << "module " << moduleName << " not found";
-    return failure();
-  }
+  assert(node && "module not found");
 
   // Otherwise, if the module is not a HWModuleOp, then we should treat it as a
   // fanIn.
@@ -778,8 +775,7 @@ FailureOr<ArrayRef<OpenPath>> LocalVisitor::getOrComputeResults(Value value,
     auto leader = ec.findLeader({value, bitPos});
     // If this is not the leader, then use the leader.
     if (*leader != std::pair(value, bitPos)) {
-      auto root = ecMap.at(*leader);
-      return getOrComputeResults(root.first, root.second);
+      return getOrComputeResults(leader->first, leader->second);
     }
   }
 
@@ -974,7 +970,11 @@ LogicalResult LocalVisitor::initializeAndRun() {
     return WalkResult::advance();
   });
 
-  done.store(true);
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    done.store(true);
+    cv.notify_all();
+  }
   LLVM_DEBUG({ ctx->notifyEnd(module.getModuleNameAttr()); });
   return failure(walkResult.wasInterrupted());
 }
--- a/lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
+++ b/lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
@@ -29,26 +29,115 @@
 //===----------------------------------------------------------------------===//
 
 #include "circt/Dialect/AIG/Analysis/LongestPathAnalysis.h"
+#include "circt/Dialect/AIG/AIGOps.h"
 #include "circt/Dialect/AIG/AIGPasses.h"
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/HW/HWInstanceGraph.h"
 #include "circt/Dialect/HW/HWOps.h"
 #include "circt/Dialect/Seq/SeqOps.h"
+#include "circt/Dialect/Seq/SeqTypes.h"
 #include "circt/Support/InstanceGraph.h"
 #include "circt/Support/LLVM.h"
 #include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/Diagnostics.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/Threading.h"
 #include "mlir/IR/Value.h"
 #include "mlir/IR/Visitors.h"
 #include "mlir/Pass/AnalysisManager.h"
+#include "mlir/Support/FileUtilities.h"
 #include "mlir/Support/LLVM.h"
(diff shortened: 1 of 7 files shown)


test/Dialect/AIG/longest-paths.mlir
```diff
@@ -0,0 +1,99 @@
...
+  %q = comb.and %p, %a : i1  // q[0] := max(p[0], a[0]) + 1 = 2
+  %r = comb.or %q, %a : i1  // r[0] := max(q[0], a[0]) + 1 = 3
+  %s = comb.xor %r, %a : i1 // s[0] := max(r[0], a[0]) + 1 = 4
+  hw.output %s : i1
+}
+
+// expected-remark-re @+4 {{fanOut=Object($root.x[0]), fanIn=Object($root.a[0], delay=1, history=[{{.+}}])}}
+// expected-remark-re @+3 {{fanOut=Object($root.x[0]), fanIn=Object($root.b[0], delay=1, history=[{{.+}}])}}
+// expected-remark-re @+2 {{fanOut=Object($root.x[1]), fanIn=Object($root.a[1], delay=1, history=[{{.+}}])}}
+// expected-remark-re @+1 {{fanOut=Object($root.x[1]), fanIn=Object($root.b[1], delay=1, history=[{{.+}}])}}
```
fabianschuiki: Could `@below` work for these, instead of counting the number of lines?

test/Dialect/AIG/longest-paths.mlir
```diff
@@ -0,0 +1,99 @@
...
+
+hw.module private @child(in %a : i1, in %b : i1, out x : i1) {
+  %r = aig.and_inv %a, %b : i1 // r[0] := max(a[0], b[0]) + 1 = 1
+  hw.instance "bound" @bound(a: %a: i1) -> () {doNotPrint}
+  hw.output %r : i1
+}
+
+// Check history.
+// expected-remark @+2 {{fanOut=Object($root.x[0]), fanIn=Object($root.a[0], delay=2, history=[Object($root.c2.x[0], delay=2, comment="output port"), Object($root/c2:child.a[0], delay=1, comment="input port"), Object($root.c1.x[0], delay=1, comment="output port"), Object($root/c1:child.a[0], delay=0, comment="input port"), Object($root.a[0], delay=0, comment="input port")])}}
+// expected-remark @+1 {{fanOut=Object($root.x[0]), fanIn=Object($root.b[0], delay=2, history=[Object($root.c2.x[0], delay=2, comment="output port"), Object($root/c2:child.a[0], delay=1, comment="input port"), Object($root.c1.x[0], delay=1, comment="output port"), Object($root/c1:child.b[0], delay=0, comment="input port"), Object($root.b[0], delay=0, comment="input port")])}}
```
fabianschuiki: This is really really cool :partying_face:

unittests/Dialect/AIG/LongestPathTest.cpp
```diff
@@ -0,0 +1,154 @@
...
+      %r = hw.instance "inst" @child(a: %p: i2, b: %b: i2) -> (x: i2)
+      %s = aig.and_inv not %p, %q, %r : i2
+      %dummy = comb.concat %s, %a : i2, i2
+      hw.output %s, %dummy : i2, i4
+    } 
+    hw.module private @child(in %a : i2, in %b : i2, out x : i2) {
+      %r = aig.and_inv not %a, %b {sv.namehint = "child.r"} : i2
+      hw.output %r : i2
+    }
+    )MLIR";
```
fabianschuiki, line 42: Do you know if there is some precedence for having MLIR blobs encoded in the C++ source as strings? This feels sensible since you are trying to go through a lot of the details provided by the analysis. On the other hand, you could argue that maybe the printer should have a verbose option where it prints all the interesting info you'd want to check, which would then allow you to check this with FileCheck.

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -172,47 +262,1100 @@ struct llvm::DenseMapInfo<Object> {
   }
 };
 
+class LocalVisitor;
+
+//===----------------------------------------------------------------------===//
+// Context
+//===----------------------------------------------------------------------===//
+// This class provides a thread-safe interface to access the analysis results.
```
fabianschuiki:
```suggestion

/// This class provides a thread-safe interface to access the analysis results.
```

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -172,47 +262,1100 @@ struct llvm::DenseMapInfo<Object> {
...
+  LogicalResult visit(comb::ConcatOp op, size_t bitPos,
+                      SmallVectorImpl<OpenPath> &results);
+  LogicalResult visit(comb::ExtractOp op, size_t bitPos,
+                      SmallVectorImpl<OpenPath> &results);
+  LogicalResult visit(comb::ReplicateOp op, size_t bitPos,
+                      SmallVectorImpl<OpenPath> &results);
+  // Track equivalence classes for data movement ops
+  // Treat output as equivalent to input for pure data movement operations
+  llvm::EquivalenceClasses<std::pair<Value, size_t>> ec;
+  DenseMap<std::pair<Value, size_t>, std::pair<Value, size_t>> ecMap;
```
fabianschuiki, line 388: Nice use of `EquivalenceClasses`! :partying_face: I'm wondering if you still need to track that separate `ecMap` :thinking: It looks like you use it to map from an entry in the class to the leader, which I think the `ec` can do for you directly :thinking: Might be mistaken though.

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -172,47 +262,1100 @@ struct llvm::DenseMapInfo<Object> {
...
+  auto &slot = objectToMaxDistance[object];
+  if (slot.first >= delay && delay != 0)
+    return;
+  slot = {delay, history};
+}
+
+void LocalVisitor::waitUntilDone() const {
+  // Wait for the thread to finish.
+  while (!done.load())
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
fabianschuiki: Might be worth using a condition variable here to avoid sleeping. Not sure how much of a performance implication this has in practice :slightly_smiling_face:

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -172,47 +262,1100 @@ struct llvm::DenseMapInfo<Object> {
...
+                                           Value to, size_t toBitPos,
+                                           SmallVectorImpl<OpenPath> &results) {
+  auto leader = ec.getOrInsertLeaderValue({to, toBitPos});
+  auto &slot = ecMap[leader];
+  if (!slot.first)
+    slot = {to, toBitPos};
+  // Merge classes, and visit the leader.
+  auto newLeader = ec.unionSets({to, toBitPos}, {from, fromBitPos});
+  assert(leader == *newLeader);
+  return visitValue(to, toBitPos, results);
```
fabianschuiki: Really nice idea to get rid of aliases in the IR and boil things down to a single computed value regardless of how many wires and bit slices you have of that value :100:

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -172,47 +262,1100 @@ struct llvm::DenseMapInfo<Object> {
...
+  if (!ctx->instanceGraph)
+    return markFanIn(value, bitPos, results);
+
+  // If an instance graph is available, then we can look up the module. It's an
+  // error when the module is not found.
+  auto *node = ctx->instanceGraph->lookup(moduleName);
+  if (!node || !node->getModule()) {
+    op.emitError() << "module " << moduleName << " not found";
+    return failure();
+  }
```
fabianschuiki: Do you think we could turn this into an assert? I think the `hw.instance` verifier would already error if the module did not exist :thinking:

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -172,47 +262,1100 @@ struct llvm::DenseMapInfo<Object> {
...
+}
+
+LogicalResult LocalVisitor::addLogicOp(Operation *op, size_t bitPos,
+                                       SmallVectorImpl<OpenPath> &results) {
+  auto size = op->getNumOperands();
+  auto cost = std::max(1u, llvm::Log2_64_Ceil(size));
+  // Create edges each operand with cost ceil(log(size)).
+  for (auto operand : op->getOperands())
+    if (failed(addEdge(operand, bitPos, cost, results)))
+      return failure();
```
fabianschuiki, line 747: Using the logarithm as an approximation for the ideal work depth of a logic op is a nice idea! If I remember correctly there was even some theoretical work by Torsten Hoefler showing that this holds in many cases.

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -172,47 +262,1100 @@ struct llvm::DenseMapInfo<Object> {
...
+  }
+  return maxDelay;
+}
+
 //===----------------------------------------------------------------------===//
 // LongestPathAnalysis
 //===----------------------------------------------------------------------===//
 
-LongestPathAnalysis::~LongestPathAnalysis() {}
+LongestPathAnalysis::~LongestPathAnalysis() { delete impl; }
```
fabianschuiki, line 1324: Could impl be an `std::unique_ptr` to avoid having to deal with naked pointers and allocation/deallocation?

fabianschuiki (approved):
This is really really cool! I love the split into a thread-safe global and a multi-threaded local part of the analysis to make things scale. And using equivalence classes to deal with aliasing operations is very clever :partying_face: :100: Can't wait to see what we can do with this in CIRCT :rocket:

test/Dialect/AIG/longest-paths.mlir
uenoku: Yeah, thanks!

unittests/Dialect/AIG/LongestPathTest.cpp
uenoku: MLIR unittest use this pattern so I think it's ok to use it. 
https://github.com/llvm/llvm-project/blob/c6c2e21028cadef854cf22f6ecaa5eb9d224b76d/mlir/unittests/ExecutionEngine/Invoke.cpp#L74

> On the other hand, you could argue that maybe the printer should have a verbose option where it prints all the interesting info you'd want to check, which would then allow you to check this with FileCheck.

That's a good point but I think there are several things easier to test with C++ unittests, e.g. analysis has a debug point flag (analysis printer pass uses a debug flag because it's necessary to print traces). Also for retiming pass I have a plan to add an API to override visitors from analysis constructor so it will be necessary to write a test for that as well. So please let me keep the test for now.

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: Yeah totally!

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: Yeah I tried but it didn't work. std::unique_ptr requires size to be determined in template instantiation so it's necessary to move Impl definition to the header. But we need to move implementation details such as Context/LocalVisitor so I think it's ok to use raw alloc/dealloc 😞

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: interesting! I'll check 👍

PR #8531, opened 2025-06-03
[Support] Factor out unused op pruning from LLHD Mem2Reg; NFC

fabianschuiki:
Factor out the tracking and pruning of unused, side effect free ops from the Mem2Reg pass into a new `UnusedOpPruner` utility. This just wraps a set of ops and provides helper functions to mark ops, values, and value ranges as to be pruned if they end up being unused at a later point.

This helps passes clean up after themselves if they remove ops that may render other parts of the IR unused. I'm working on an LLHD pass that will benefit from this utility.

### DIFF
--- a/include/circt/Support/UnusedOpPruner.h
+++ b/include/circt/Support/UnusedOpPruner.h
@@ -1,0 +1,73 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_SUPPORT_UNUSEDOPPRUNER_H
+#define CIRCT_SUPPORT_UNUSEDOPPRUNER_H
+
+#include "circt/Support/LLVM.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
+
+namespace circt {
+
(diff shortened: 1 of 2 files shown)


include/circt/Support/UnusedOpPruner.h
```diff
@@ -0,0 +1,72 @@
...
+      auto *op = *it;
+      opsToEraseIfUnused.erase(it);
+      if (!isOpTriviallyDead(op))
+        continue;
+      eraseLaterIfUnused(op->getOperands());
+      op->erase();
+    }
+  }
+
+  /// The set of operations that may have become unused through the IR changes
```
uenoku, line 66: nit: Please make `opsToEraseIfUnused` private

include/circt/Support/UnusedOpPruner.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace circt {
+
+/// Utility that tracks operations that have potentially become unused and
+/// allows them to be cleaned up at a later point. Useful to make passes clean
+/// up dead code.
+struct UnusedOpPruner {
+  /// Mark an op the be erased later if it is unused at that point.
+  void eraseLaterIfUnused(Operation *op) {
+    if (op)
```
uenoku: nit: Maybe `assert(op)` instead of null guard?

include/circt/Support/UnusedOpPruner.h
fabianschuiki: Great idea :+1:

include/circt/Support/UnusedOpPruner.h
fabianschuiki: Yeah that's a good idea :+1:

PR #8533, opened 2025-06-03
[FIRRTL] Drop ExcludeMemFromMemToRegOfVec, unused

seldridge:
Drop an annotation-backed feature that had no users from the FIRRTL:
ExcludeMemFromMemToRegOfVec.  I deleted this internally as it had no
users.  This appears to have been created when the original pass was
ported from the SFC.


uenoku (approved):
LGTM

PR #8534, opened 2025-06-04
[LLHD] Add pass to combine drives to aggregate signals

fabianschuiki:
Add the *CombineDrives* LLHD pass. If multiple drives assign all fields of an aggregate signal, merge the drives into a single drive of the whole aggregate value. This is equivalent to replacing

    a[0] = u;
    a[1] = v;

 with

    a = {v, u};

Performing this replacement is an important step to eliminate signal projection operations like `llhd.sig.extract`, `llhd.sig.array_slice`, `llhd.sig.array_get`, and `llhd.sig.struct_extract`.

Real-world Verilog inputs will often declare struct, array, or bit vector signals and then separately assign its fields, elements, or bit slices in multiple concurrent assign statements. To simplify working with the IR in optimizations and later passes, we would much rather have a single SSA value be driven onto a signal. This makes it clear that the signal is uniquely driven and fully defined. In most cases we can then remove the signal altogether and use that aggregate SSA value instead.

The pass is pretty conservative at the moment. It does not support dynamic slice offsets and element indices since the semantics of driving such projections implies that individual bits are only driven under certain non-trivial conditions. The pass also makes no attempt at finding out if it can see all drives to a signal and whether it can use the signal's default value to fill in undriven portions of the signal. This might be a future optimization worth investigating.

### DIFF
--- a/test/Dialect/LLHD/Transforms/combine-drives.mlir
+++ b/test/Dialect/LLHD/Transforms/combine-drives.mlir
@@ -1,0 +1,211 @@
+// RUN: circt-opt --llhd-combine-drives %s | FileCheck %s
+
+// Basic aggregation of drives through multiple layers of aggregates.
+// CHECK-LABEL: @Basic
+hw.module @Basic(in %u0: i42, in %u1: i42, in %u2: i42, in %u3: i20, in %u4: i22) {
+  %c0_i2 = hw.constant 0 : i2
+  %c1_i2 = hw.constant 1 : i2
+  %false = hw.constant false
+  %true = hw.constant true
+  %c0_i6 = hw.constant 0 : i6
+  %c20_i6 = hw.constant 20 : i6
+  %0 = builtin.unrealized_conversion_cast to !hw.struct<a: !hw.array<3xi42>, b: i42>
+  %1 = llhd.constant_time <0ns, 0d, 1e>
+  // CHECK: %s = llhd.sig
+  // CHECK-NEXT: [[STRUCT:%.+]] = hw.struct_create ([[A:%.+]], [[B:%.+]])
+  // CHECK-NEXT: llhd.drv %s, [[STRUCT]] after {{%.+}}
+  %s = llhd.sig %0 : !hw.struct<a: !hw.array<3xi42>, b: i42>
+  // CHECK-NEXT: [[A]] = hw.array_create %u2, %u1, %u0
+  %a = llhd.sig.struct_extract %s["a"] : !hw.inout<struct<a: !hw.array<3xi42>, b: i42>>
+  %a12 = llhd.sig.array_slice %a at %c1_i2 : (!hw.inout<array<3xi42>>) -> !hw.inout<array<2xi42>>
+  %a0 = llhd.sig.array_get %a[%c0_i2] : !hw.inout<array<3xi42>>
+  %a1 = llhd.sig.array_get %a12[%false] : !hw.inout<array<2xi42>>
+  %a2 = llhd.sig.array_get %a12[%true] : !hw.inout<array<2xi42>>
+  llhd.drv %a0, %u0 after %1 : !hw.inout<i42>  // s.a[0] = u0
+  llhd.drv %a1, %u1 after %1 : !hw.inout<i42>  // s.a[2:1][0] = u1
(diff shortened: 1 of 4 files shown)


test/Dialect/LLHD/Transforms/combine-drives.mlir
```diff
@@ -0,0 +1,147 @@
...
+  // CHECK-NEXT: %a = llhd.sig
+  %a = llhd.sig %c0_i42 : i42
+  // CHECK-NEXT: [[TMP:%.+]] = comb.concat %v, %u
+  // CHECK-NEXT: llhd.drv %a, [[TMP]] after [[T0]]
+  // CHECK-NEXT: [[TMP:%.+]] = comb.concat %u, %v
+  // CHECK-NEXT: llhd.drv %a, [[TMP]] after [[T0]]
+  // CHECK-NEXT: [[TMP:%.+]] = comb.concat %u, %u
+  // CHECK-NEXT: llhd.drv %a, [[TMP]] after [[T1]]
+  // CHECK-NEXT: [[TMP:%.+]] = comb.concat %v, %v
+  // CHECK-NEXT: llhd.drv %a, [[TMP]] after [[T1]]
```
maerhart: Nit: maybe also check for the drive enable operand?

test/Dialect/LLHD/Transforms/combine-drives.mlir
```diff
@@ -0,0 +1,147 @@
...
+// CHECK-LABEL: @SkipDynamicArrayGet
+hw.module @SkipDynamicArrayGet(in %u: i6, in %v: i42) {
+  %0 = llhd.constant_time <0ns, 0d, 1e>
+  %1 = builtin.unrealized_conversion_cast to !hw.array<42xi42>
+  %a = llhd.sig %1 : !hw.array<42xi42>
+  // CHECK: llhd.sig.array_get
+  %2 = llhd.sig.array_get %a[%u] : !hw.inout<array<42xi42>>
+  // CHECK: llhd.drv
+  llhd.drv %2, %v after %0 : !hw.inout<i42>
+}
```
maerhart, line 128: Nit: we might want to check the drive signal operand because here it could match on an arbitrary drive the pass inserts

test/Dialect/LLHD/Transforms/combine-drives.mlir
```diff
@@ -0,0 +1,147 @@
...
+  %c42_i8 = hw.constant 42 : i8
+  %0 = llhd.constant_time <0ns, 0d, 1e>
+  %a = llhd.sig %u : i153
+  // CHECK: llhd.drv %a, %u
+  llhd.drv %a, %u after %0 : !hw.inout<i153>
+  // CHECK: [[TMP:%.+]] = llhd.sig.extract
+  // CHECK: llhd.drv [[TMP]], %v
+  %1 = llhd.sig.extract %a from %c42_i8 : (!hw.inout<i153>) -> !hw.inout<i5>
+  llhd.drv %1, %v after %0 : !hw.inout<i5>
+}
```
maerhart, line 177: Maybe add tests with signal users in nested regions (such as `llhd.process`)?

maerhart: There's also no test for the case when a signal is coming from a module port

maerhart (approved):
LGTM! Should we add this pass to the circt-verilog pipeline?

test/Dialect/LLHD/Transforms/combine-drives.mlir
fabianschuiki: :man_facepalming: That's what I wanted to do, and forgot. Thanks :grin:

test/Dialect/LLHD/Transforms/combine-drives.mlir
fabianschuiki: Good point. Let me tweak these!

test/Dialect/LLHD/Transforms/combine-drives.mlir
fabianschuiki: Excellent points, let me add those :+1:

fabianschuiki:
I have a branch that contains a bunch of tweaks to the circt-verilog pipeline. My plan was to open a PR with that once the reworked LLHD passes cover the functionality of the current pipeline.

PR #8535, opened 2025-06-04
[FIRRTL] Add moveDut field for inject DUT hier

seldridge:
Add a new field to the `InjectDUTHierarchyAnnotation` called `moveDut`.
This is used to control the placement of the `MarkDUTAnnotation` after the
pass runs.  This is done becasue we are exploring an internal flow which
wants to change the behavior of this pass to work this way.  We would like
to make this the default behavior, but will initially guard it with this
annotaiton option.

This patch is backwards compatible.  It will default to the old behavior
if it does not see the new field in the annotation.

This makes no changes to the placement of _other_ annotations.  I.e., if
there are annotations that were originally on the DUT, then those will
still be moved.  I went back and forth on this.  However, it shouldn't
really matter where _other_ annotations are nor should it matter where
NLAs are rooted so long as they are information preserving.  This
additionally makes the pass easier to maintain as it doesn't require
having different NLA movement logic for different annotation settings.

CC: @azidar

### DIFF
--- a/docs/Dialects/FIRRTL/FIRRTLAnnotations.md
+++ b/docs/Dialects/FIRRTL/FIRRTLAnnotations.md
@@ -595,10 +595,11 @@ Example:
 
 ### InjectDUTHierarchyAnnotation
 
-| Property | Type   | Description                                             |
-|----------|--------|---------------------------------------------------------|
-| class    | string | `sifive.enterprise.firrtl.InjectDUTHierarchyAnnotation` |
-| name     | string | The name of the module containing original DUT logic    |
+| Property | Type    | Description                                                                                       |
+|----------|---------|---------------------------------------------------------------------------------------------------|
+| class    | string  | `sifive.enterprise.firrtl.InjectDUTHierarchyAnnotation`                                           |
+| name     | string  | The name of the module containing original DUT logic                                              |
+| moveDut  | boolean | Optional: if true, then treat the newly created module as the DUT. If unset, then treat as false. |
 
 This annotation can be used to add an extra level of hierarchy in the design
 under the DUT (indicated with a `MarkDUTAnnotation`).  All logic in the original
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
```diff
@@ -165,10 +171,22 @@ void InjectDUTHierarchy::runOnOperation() {
...
+        return !moveDut;
+      return true;
+    });
+
+    // Finish setting up the DUT.  Strip the `MarkDUTAnnotation` is we are in
+    // "moveDut" mode.
+    if (moveDut)
+      AnnotationSet::removeAnnotations(dut, [&](Annotation anno) {
+        return anno.isClass(dutAnnoClass) && moveDut;
+      });
```
uenoku: nit: I think moveDut is always true in lambda, so simply
```suggestion
    if (moveDut)
      AnnotationSet::removeAnnotations(dut, dutAnnoClass);
```

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
```diff
@@ -607,13 +608,19 @@ used in combination with `ExtractBlackBoxAnnotation` (or with passes that add
...
+
 This annotation should only appear zero or once.
 
 Example:
 ``` json
 {
   "class": "sifive.enterprise.firrtl.InjectDUTHierarchyAnnotation",
-  "name": "Logic"
+  "name": "Logic",
+  "moveDutAnno": false
```
uenoku:
```suggestion
  "moveDut": false
```

uenoku (approved):
LGTM

lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
seldridge: Fixed in: adc0df3ca

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
seldridge: Fixed in: f661ca879

PR #8536, opened 2025-06-04
[LLHD] Add combinational process op

fabianschuiki:
Add a dedicated op for processes that should trigger when any of the values used as operands in its body region change. This reflects the behavior of `always_comb` in SystemVerilog and will allow passes such as desequentialization and process lowering to convert `llhd.process`es to this more streamlined and specialized representation.

A later commit will replace uses of `scf.execute_region` in the LLHD pass pipeline with `llhd.combinational`.


include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
```diff
@@ -86,6 +86,46 @@ def FinalOp : LLHDOp<"final", [
...
+  RecursiveMemoryEffects,
+  HasParent<"hw::HWModuleOp">,
+]> {
+  let summary = "A process that runs when any of its operand values change";
+  let description = [{
+    An `llhd.combinational` op encapsulates a region of IR that executes
+    whenever any of the values used in its body change. Control flow must
+    eventually end in an `llhd.halt` terminator. The process may have results,
+    in which case the `llhd.halt` terminators must provide a list of values to
+    yield for the process results.
```
maerhart: I think it's worth mentioning that the halt operation here does not permanently halt the process like it does in an `llhd.process` but just suspends it until the next value change is detected. It is essentially just like an `llhd.wait` in an `llhd.process` where the sensitivity list implicitly contains all values used in the region that are defined outside of it.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
```diff
@@ -86,6 +86,46 @@ def FinalOp : LLHDOp<"final", [
...
+
+    This op is commonly used to embed a control-flow description of some
+    combinational logic inside the surrounding module's graph region.
```
maerhart, line 110: Maybe mention that it is equivalent to an `llhd.process` that has a `wait` at the end, i.e., it executes once in the beginning regardless of whether a value changed.
Might also be worth mentioning that `llhd.wait` terminators are not allowed in the body of this op. Probably also worth adding a verifier for that.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
```diff
@@ -86,6 +86,46 @@ def FinalOp : LLHDOp<"final", [
...
+    ```mlir
+    hw.module @Foo() {
+      %0, %1 = llhd.combinational -> i42, i9001 {
+        cf.cond_br %2, ^bb1(%3, %4 : i42, i9001), ^bb1(%5, %6 : i42, i9001)
+      ^bb1(%7: i42, %8: i9001):
+        llhd.halt %7, %8 : i42, i9001
+      }
+    }
+    ```
+  }];
```
maerhart, line 122: Are there any restrictions around drives in this op? Multiple drives to the same signal (resulting in conflicts if naively inlined)? Conditional drives? Essentially, whose responsibility is it to check for such things, the pass that creates this op or the pass that analyses/lowers this op?

test/Dialect/LLHD/IR/basic.mlir
```diff
@@ -218,3 +218,14 @@ hw.module @ProcessWithResults(in %arg0: i42, in %arg1: i9001) {
...
+
+hw.module @CombinationalProcess(in %arg0: i1, in %arg1: i42, in %arg2: i9001, in %arg3: i42, in %arg4: i9001) {
+  llhd.combinational {
+    llhd.halt
+  }
+  %0:2 = llhd.combinational -> i42, i9001 {
+    cf.cond_br %arg0, ^bb1(%arg1, %arg2 : i42, i9001), ^bb1(%arg3, %arg4 : i42, i9001)
+  ^bb1(%1: i42, %2: i9001):
+    llhd.halt %1, %2 : i42, i9001
+  }
```
maerhart, line 230: No CHECK directives to verify detail of the printer that the parser doesn't necessarily care about (e.g. whitespace)?

fabianschuiki: I was hoping for `verify-roundtrip` to do a reasonable check, maybe minus the whitespace :innocent:

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
fabianschuiki: I think you can have anything you'd want in here. Might sometimes even be desirable, for example, if we want to directly map from Verilog's `always_comb` to `llhd.combinational`. That would then still contain weird drives. I think the difference is that a drive in a process/combinational op has different semantics from a drive in the module -- the former is like a store, overwriting previous values, while the latter is a permanent driver that produces drive conflicts. We might want to split these into separate ops in the future, to make it clear that moving drives from processes into modules is dangerous. It would then be the pass lowering/inlining the combinational op that has to check if the inlining is allowed. For example, if it contains any procedural drives it can't be inlined into a graph-region parent.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
fabianschuiki: Great idea. I've tweaked this to use an `llhd.yield` terminator, and reworded the description.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
fabianschuiki: Great idea. I've tweaked this to use an `llhd.yield` terminator, and reworded the description.

lib/Dialect/LLHD/IR/LLHDOps.cpp
```diff
@@ -494,6 +494,30 @@ LogicalResult ProcessOp::canonicalize(ProcessOp op, PatternRewriter &rewriter) {
...
+//===----------------------------------------------------------------------===//
+
+LogicalResult CombinationalOp::canonicalize(CombinationalOp op,
+                                            PatternRewriter &rewriter) {
+  // Inline the combinational region if it consists of a single block and
+  // contains no side-effecting operations.
+  if (op.getBody().hasOneBlock()) {
+    auto &block = op.getBody().front();
+    auto noSideEffects = llvm::all_of(block.without_terminator(), [](auto &op) {
+      return isMemoryEffectFree(&op);
```
maerhart: Isn't it possible to just call this function on the combinational op directly and the recursive side-effects trait does the rest? All the terminators we may use in this op are now pure.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
maerhart: That sounds like a reasonable decision. Properly verifying any of these restrictions/invariants would be almost impossible in practice anyway. `llhd.wait` terminators are not allowed in this op though, right?

test/Dialect/LLHD/IR/basic.mlir
maerhart: I had messed up whitespace in the assembly format before (especially around parentheses). It didn't break functionality but was weird to look at. That's why I brought it up.

include/circt/Dialect/LLHD/IR/LLHDStructureOps.td
fabianschuiki: Yes exactly! With the `llhd.yield`, you can only return a bunch of values as the process results. No suspension of execution is possible. So I guess most of the weirdness of processes disappears through this :thinking:

lib/Dialect/LLHD/IR/LLHDOps.cpp
fabianschuiki: That's a great point! Let me do that.

PR #8537, opened 2025-06-04
[LLHD] Use new combinational process instead of execute regions

fabianschuiki:
Make deseq and process lowering use the new `llhd.combinational` op instead of the existing `scf.execute_region`. The latter is too eager to inline into the parent region, even if the region contains ops with side effects such as probes and drives. These no longer have a clear order in the parent module. The specialized LLHD op is less eager about inlining and can deal with the details of the SSACFG-to-graph region transition in more detail.

This commit is basically just a whole bunch of renaming and changing comments. No change intended besides the different op.


maerhart (approved):

lib/Dialect/LLHD/Transforms/Deseq.cpp
```diff
@@ -1222,14 +1223,14 @@ ValueRange Deseq::specializeProcess(FixedValues fixedValues) {
       auto [oldBlock, newBlock] = worklist.pop_back_val();
       builder.setInsertionPointToEnd(newBlock);
       for (auto &oldOp : *oldBlock) {
-        // Convert `llhd.wait` into `scf.yield`.
+        // Convert `llhd.wait` into `llhd.halt`.
```
uenoku: nit: is this comment correct?

fabianschuiki: No you're right, this should be `llhd.yield`. Thanks! 🙌

PR #8542, opened 2025-06-05
[LLHD] Add llhd.sig.extract support to Mem2Reg

fabianschuiki:
Extend LLHD's Mem2Reg pass to support `llhd.sig.extract` projections. To implement the extraction and injection of bit ranges at dynamic offsets, add a few helper functions to `CombOps.h` that assemble the necessary operations. We may want to add a `comb.inject` op in the future to simplify this.

### DIFF
--- a/lib/Dialect/Comb/CombOps.cpp
+++ b/lib/Dialect/Comb/CombOps.cpp
@@ -14,11 +14,28 @@
 #include "circt/Dialect/HW/HWOps.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/ImplicitLocOpBuilder.h"
+#include "mlir/IR/Matchers.h"
 #include "mlir/IR/PatternMatch.h"
 #include "llvm/Support/FormatVariadic.h"
 
 using namespace circt;
 using namespace comb;
+
+Value comb::createZExt(OpBuilder &builder, Location loc, Value value,
+                       unsigned targetWidth) {
+  assert(value.getType().isInteger());
+  auto inputWidth = value.getType().getIntOrFloatBitWidth();
+  assert(inputWidth <= targetWidth);
+
+  // Nothing to do if the width already matches.
+  if (inputWidth == targetWidth)
+    return value;
+
+  // Create a zero constant for the upper bits.
+  auto zeros = builder.create<hw::ConstantOp>(
+      loc, builder.getIntegerType(targetWidth - inputWidth), 0);
+  return builder.createOrFold<ConcatOp>(loc, zeros, value);
+}
 
 /// Create a sign extension operation from a value of integer type to an equal
 /// or larger integer type.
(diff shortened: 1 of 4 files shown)


lib/Dialect/Comb/CombOps.cpp
```diff
@@ -14,12 +14,29 @@
...
+#include "mlir/IR/Matchers.h"
 #include "mlir/IR/PatternMatch.h"
 #include "llvm/Support/FormatVariadic.h"
 
 using namespace circt;
 using namespace comb;
 
+Value comb::createZExt(OpBuilder &builder, Location loc, Value value,
+                       unsigned targetWidth) {
+  assert(value.getType().isInteger());
```
maerhart: Nit: could also use `isSignlessInteger()` as comb only uses signless

lib/Dialect/Comb/CombOps.cpp
```diff
@@ -111,6 +128,90 @@ Value comb::constructMuxTree(OpBuilder &builder, Location loc,
...
+  value = builder.createOrFold<comb::AndOp>(loc, value, mask, twoState);
+
+  // Zero-extend the replacement value, shift it up to the offset, and merge it
+  // with the value that has the corresponding bits cleared.
+  replacement = createZExt(builder, loc, replacement, largeWidth);
+  replacement = builder.createOrFold<comb::ShlOp>(loc, replacement, offset);
+  return builder.createOrFold<comb::OrOp>(loc, value, replacement, twoState);
+}
+
+Value comb::createInject(OpBuilder &builder, Location loc, Value value,
```
maerhart, line 186: This function doesn't have a test or am I missing something?

lib/Dialect/Comb/CombOps.cpp
```diff
@@ -111,6 +128,90 @@ Value comb::constructMuxTree(OpBuilder &builder, Location loc,
...
+    return value;
+
+  // Assemble the pieces of the injection as everything below the offset, the
+  // replacement value, and everything above the replacement value.
+  SmallVector<Value, 3> fragments;
+  auto end = offset + smallWidth;
+  if (end < largeWidth)
+    fragments.push_back(
+        builder.create<comb::ExtractOp>(loc, value, end, largeWidth - end));
+  fragments.push_back(replacement);
```
maerhart: What if `end > largeWidth`? Should we extract from `replacement` and ignore the non-overlapping part?

fabianschuiki: Nice catch, thanks! I have added handling for that case.

lib/Dialect/Comb/CombOps.cpp
fabianschuiki: Good point. Let me trigger these through the mem2reg tests.

lib/Dialect/Comb/CombOps.cpp
fabianschuiki: :+1: Good point!

PR #8543, opened 2025-06-06
Add vectorization pass for Moore dialect

mafeguimaraes:
This patch introduces a new transformation pass for the Moore dialect that performs simple vectorization on scalar assignments.

The pass detects sequences of bit-level assignments from input to output signals and replaces them with a single vector assignment, simplifying the IR and generating cleaner Verilog.

**Example**

Given the Verilog design:
```Systemverilog
module simple_vectorization(output wire [3:0] out, input wire [3:0] in);
  assign out[3] = in[3];
  assign out[2] = in[2];
  assign out[1] = in[1];
  assign out[0] = in[0];
endmodule
```

The corresponding Moore IR before vectorization looks like:
```MLIR
module {
  moore.module @simple_vectorization(out out : !moore.l4, in %in : !moore.l4) {
    %out = moore.net wire : <l4>
    %0 = moore.extract_ref %out from 3 : <l4> -> <l1>
    %1 = moore.extract %in from 3 : l4 -> l1
    moore.assign %0, %1 : l1
    %2 = moore.extract_ref %out from 2 : <l4> -> <l1>
    %3 = moore.extract %in from 2 : l4 -> l1
    moore.assign %2, %3 : l1
    %4 = moore.extract_ref %out from 1 : <l4> -> <l1>
    %5 = moore.extract %in from 1 : l4 -> l1
    moore.assign %4, %5 : l1
    %6 = moore.extract_ref %out from 0 : <l4> -> <l1>
    %7 = moore.extract %in from 0 : l4 -> l1
    moore.assign %6, %7 : l1
    %8 = moore.read %out : <l4>
    moore.output %8 : !moore.l4
  }
}
```
After running the VectorizationPass, the IR is simplified to:
```MLIR
module {
  moore.module @simple_vectorization(out out : !moore.l4, in %in : !moore.l4) {
    %out = moore.net wire : <l4>
    moore.assign %out, %in : l4
    %0 = moore.read %out : <l4>
    moore.output %0 : !moore.l4
  }
}
```

Which generates the optimized Verilog:
```Systemverilog
module simple_vectorization(output wire [3:0] out, input wire [3:0] in);
  assign out = in;
endmodule
```

This pass currently handles straightforward, contiguous, and ordered cases only.
It matches extract and connect operations of the form out[i] = in[i] for a range of i.

This is the first vectorization case handled by the pass, future work will extend support to more complex and interleaved patterns.


IanTrudel:
Would you care to take a look at this PR @fabianschuiki ?

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+        PassPipelineRegistration<>(
+            "simple-vec", "Simple Vectorization Pass",
+            [](OpPassManager &pm) {
+              pm.addPass(std::make_unique<SimpleVectorizationPass>());
+            });
+      }};
+}
+
+MLIR_DECLARE_EXPLICIT_TYPE_ID(SimpleVectorizationPass)
+MLIR_DEFINE_EXPLICIT_TYPE_ID(SimpleVectorizationPass)
```
maerhart: I guess this was added to use this pass with `circt-opt`. For upstreaming, we should remove this and if a plugin is still needed for some reason we should add it for the entire Moore dialect and register all its passes (in a separate file).

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+          vectorizeGroup(group);
+      }
+    }
+  }
+
+  StringRef getArgument() const override { return "simple-vec"; }
+
+  StringRef getDescription() const override {
+    return "Simple Vectorization Pass using tree structure";
+  }
```
maerhart: You can declare your pass in `MoorePasses.td` (there are already 2 examples) which will generate a base class that already contains these and also all the pass registration hooks.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+    g.extractRef.erase();
+    g.extract.erase();
+  }
+}
+
+struct SimpleVectorizationPass
+    : public mlir::PassWrapper<SimpleVectorizationPass, mlir::OperationPass<mlir::ModuleOp>> {
+
+  void runOnOperation() override {
+    auto module = getOperation();
```
maerhart: Nit: `module` is a keyword from C++20 onward, so it's probably better to avoid it as a variable name

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+  moore::ExtractOp extract;
+  moore::ContinuousAssignOp assign;
+  int index;
+};
+
+struct ValueComparator {
+  bool operator()(mlir::Value lhs, mlir::Value rhs) const {
+    return lhs.getAsOpaquePointer() < rhs.getAsOpaquePointer();
+  }
+};
```
maerhart: I'd recommend to not compare values like that because it's not deterministic.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+using IndexedGroupMap = std::map<int, ScalarAssignGroup>;
+using SourceGroupMap = std::map<mlir::Value, IndexedGroupMap, ValueComparator>;
+using AssignTree = std::map<mlir::Value, SourceGroupMap, ValueComparator>;
```
maerhart: We usually don't use `std::map` in LLVM/CIRCT. I'd recommend to use `DenseMap` for an unordered hash-map and `MapVector` if you need deterministic iteration.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+void vectorizeGroup(std::vector<ScalarAssignGroup> &group) {
```
maerhart: It seems like the function argument is only read, in that case it's better to declare it as `ArrayRef<ScalarAssignGroup> group`.

maerhart: The LLVM coding standard document also recommends to use the `static` keyword for functions instead of wrapping them in an anonymous namespace.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  void runOnOperation() override {
+    auto module = getOperation();
+    AssignTree assignTree;
+
+    module.walk([&](moore::ContinuousAssignOp assign) {
+      auto lhs = assign.getDst();
+      auto rhs = assign.getSrc();
+
+      auto extractRef = dyn_cast_or_null<moore::ExtractRefOp>(lhs.getDefiningOp());
```
maerhart: Nit: there's a shortcut for exactly that
```suggestion
      auto extractRef = lhs.getDefiningOp<moore::ExtractRefOp>();
```

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+      auto extract = dyn_cast_or_null<moore::ExtractOp>(rhs.getDefiningOp());
+      if (!extractRef || !extract)
+        return;
+
+      auto indexRefAttr = extractRef->getAttrOfType<mlir::IntegerAttr>("lowBit");
+      auto indexAttr = extract->getAttrOfType<mlir::IntegerAttr>("lowBit");
```
maerhart: Why not use the `getLowBit` or `getLowBitAttr` getter directly? When changing the name of that attribute on the operation (in its table-gen declaration) this approach would just silently do something incorrect while the getter based approach would lead to a compile-time error.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+      if (!indexRefAttr || !indexAttr)
+        return;
```
maerhart: When doing the above this check won't be necessary anymore because because the attributes are verified.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+      if (index != indexAttr.getInt())
+        return;
+
+      assignTree[extractRef.getOperand()][extract.getOperand()][index] =
+          {extractRef, extract, assign, index};
+    });
+
+    for (auto &[dst, srcMap] : assignTree) {
+      for (auto &[src, indexMap] : srcMap) {
+        std::vector<int> sortedIndices;
```
maerhart: Nit: We typically use the LLVM internal `SmallVector` datastructure because it allocates small vectors on the stack for better performance and `std::vector` only in rare cases.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
+#include "mlir/IR/Builders.h"
```
maerhart: There should be a license header.

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "llvm/Support/raw_ostream.h"
+#include "circt/Dialect/Moore/MooreOps.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Pass/PassRegistry.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Support/IndentedOstream.h"
+#include "mlir/Tools/Plugins/PassPlugin.h"
+#include "llvm/Config/llvm-config.h"
```
maerhart: There seem to be duplicate and unused includes. Can we prune them to only the ones actually used?

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+  if (group.empty())
+    return;
+
+  auto builder = OpBuilder(group.front().assign.getContext());
+  builder.setInsertionPoint(group.front().assign);
+
+  auto dst = group.front().extractRef.getOperand();
+  auto src = group.front().extract.getOperand();
+
+  builder.create<moore::ContinuousAssignOp>(group.front().assign.getLoc(), dst, src);
```
maerhart: I'm missing some check to make sure all the extract operation extract from the same SSA value (on both rhs and lhs).

lib/Dialect/Moore/Transforms/VectorizationPass.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  for (auto &g : group) {
+    g.assign.erase();
+    g.extractRef.erase();
+    g.extract.erase();
+  }
```
maerhart, line 112: Before erasing the extract ops we have to check whether there are other users of the SSA value they define. Otherwise, this would lead to a crash.

mafeguimaraes:
All feedback has been addressed. The code is now refactored, formatted, and the new regression tests have been added and are passing. It should be ready for a final review. Thanks again for all the guidance, @maerhart!

fabianschuiki:
Hey @mafeguimaraes, thanks for coding up a pass and opening a PR 🥳! I'm wondering if this specific use case is already handled by the CombineDrives pass in the LLHD dialect. (I've only now gotten around to enabling it in circt-verilog in #8579 -- sorry that this hasn't been part of circt-verilog before by default 😢.)

For example, if I run your input Verilog through `circt-verilog input.sv | circt-opt --export-verilog`, I get the following:
```systemverilog
// Generated by CIRCT unknown git version
module simple_vectorization(	// <stdin>:2:3
  input  [3:0] in,	// <stdin>:2:38
  output [3:0] out	// <stdin>:2:52
);

  assign out = in;	// <stdin>:3:5
endmodule
```
This might be very close to what you are going for. Do you have any further ideas or plans on what vectorization should do? I was talking to @IanTrudel today, and he mentioned that detecting bit reversals and coming up with a concise representation would be cool. An optimization like that would be very convenient and compact to do at the Comb dialect level, and then have ExportVerilog emit a streaming operator in SV to implement the reversal 😃

IanTrudel:
> Do you have any further ideas or plans on what vectorization should do? I was talking to @IanTrudel today, and he mentioned that detecting bit reversals and coming up with a concise representation would be cool.

Here are a few select desirable outcomes:

```systemverilog
module simple_vectorization(output wire [3:0] out, input wire [3:0] in);
   assign out[3] = in[3];
   assign out[2] = in[2];
   assign out[1] = in[3];
   assign out[0] = in[0];

   // assign out = in;
endmodule
```

```systemverilog
module intermodule_vectorization(output wire [3:0] out, input wire [3:0] in);
   mybuf buf_inst1(out[3], in[3]);
   mybuf buf_inst2(out[2], in[2]);
   mybuf buf_inst3(out[1], in[1]);
   mybuf buf_inst4(out[0], in[0]);

   // assign out = in;
endmodule

module mybuf(output wire o, input wire i);
   assign o = i;
endmodule
```

```systemverilog
module reverse_endianness_vectorization(output wire [3:0] out, input wire [3:0] in);
   assign out[3] = in[0];
   assign out[2] = in[1];
   assign out[1] = in[2];
   assign out[0] = in[3];

   // assign out = {<<{in}}; // preferred
   // assign out = {in[0], in[1], in[2], in[3]}; // memory hungry on Jasper
endmodule
```

```systemverilog
module bit_mixing_vectorization(output wire [3:0] out, input wire [3:0] in);
   assign out[3] = in[0];
   assign out[1] = in[2];
   assign out[2] = in[3];
   assign out[0] = in[1];

   // assign out = {in[0], in[3:1]};
endmodule
```

```systemverilog
module pattern_recognition(output wire [3:0] result, input wire [3:0] a, b, input wire sel);
   assign result[3] = (a[3] & sel) | (b[3] & ~sel);
   assign result[2] = (a[2] & sel) | (b[2] & ~sel);
   assign result[1] = (a[1] & sel) | (b[1] & ~sel);
   assign result[0] = (a[0] & sel) | (b[0] & ~sel);

   // assign result = sel ? a : b;
endmodule
```

```systemverilog
module md4(output wire [7:0] out, input wire [7:0] in);
    assign out[7] = in[7];
    assign out[6] = in[6];
    assign out[5] = in[4];
    assign out[4] = in[5];
    assign out[3] = in[0];
    assign out[2] = in[3];
    assign out[1] = in[2];
    assign out[0] = in[1];

    //assign out = {in[7:6], in[4], in[5], in[0], in[3:1]};
endmodule
```

fabianschuiki:
I think some of these already work, or mostly work. You might have to tweak a few things in ExportVerilog to improve the emission quality -- we sometimes insert redundant wires where they are not strictly necessary. Here's what I get:

Simple vectorization works.
```systemverilog
// Generated by CIRCT unknown git version
module simple_vectorization(	// <stdin>:2:3
  input  [3:0] in,	// <stdin>:2:38
  output [3:0] out	// <stdin>:2:52
);

  assign out = in;	// <stdin>:3:5
endmodule
```

No change for `intermodule_vectorization`. That one is going to be more tricky. Cross-module optimizations can be challenging because users may expect a module instance `buf_inst1` to be present and they might want to force a port on it. Optimizations can't just inline or remove modules in that case. Unfortunately Verilog has no mechanism to annotate whether a module is private/local/static or public, which would allow you to do such optimizations.

Bit-reversal just creats a concat. This needs tweaks in the Comb dialect or pattern matching in ExportVerilog to emit a streaming operator instead.
```systemverilog
// Generated by CIRCT unknown git version
module reverse_endianness_vectorization(	// <stdin>:2:3
  input  [3:0] in,	// <stdin>:2:50
  output [3:0] out	// <stdin>:2:64
);

  wire _in_3;	// <stdin>:7:10
  wire _in_2;	// <stdin>:6:10
  wire _in_1;	// <stdin>:5:10
  wire _in_0;	// <stdin>:4:10
  assign _in_0 = in[0];	// <stdin>:4:10
  assign _in_1 = in[1];	// <stdin>:5:10
  assign _in_2 = in[2];	// <stdin>:6:10
  assign _in_3 = in[3];	// <stdin>:7:10
  assign out = {_in_0, _in_1, _in_2, _in_3};	// <stdin>:3:10, :4:10, :5:10, :6:10, :7:10, :8:5
endmodule
```

Bit mixing works, but it creates a redundant spill wire.
```systemverilog
// Generated by CIRCT unknown git version
module bit_mixing_vectorization(	// <stdin>:2:3
  input  [3:0] in,	// <stdin>:2:42
  output [3:0] out	// <stdin>:2:56
);

  wire _in_0;	// <stdin>:5:10
  assign _in_0 = in[0];	// <stdin>:5:10
  assign out = {_in_0, in[3:1]};	// <stdin>:3:10, :4:10, :5:10, :6:5
endmodule
```

The `pattern_recognition` example remains unchanged. This would be _very_ cool to support. To me this feels like it'd need an implementation of SLP (Superword-level Parallelism):
- https://groups.csail.mit.edu/cag/slp/SLP-PLDI-2000.pdf
- https://www.irisa.fr/alf/downloads/PMA/12.pdf
Since modules are a graph region, i.e., no dominance is enforced, the SLP algorithm probably becomes a lot easier than in something like LLVM.

Concatenating subranges also works, modulo redundant spills.
```systemverilog
// Generated by CIRCT unknown git version
module md4(	// <stdin>:2:3
  input  [7:0] in,	// <stdin>:2:21
  output [7:0] out	// <stdin>:2:35
);

  wire _in_0;	// <stdin>:8:10
  wire _in_5;	// <stdin>:7:10
  wire _in_4;	// <stdin>:6:10
  assign _in_4 = in[4];	// <stdin>:6:10
  assign _in_5 = in[5];	// <stdin>:7:10
  assign _in_0 = in[0];	// <stdin>:8:10
  assign out = {in[7:6], _in_4, _in_5, _in_0, in[3:1]};	// <stdin>:3:10, :4:10, :5:10, :6:10, :7:10, :8:10, :9:5
endmodule
```

PR #8545, opened 2025-06-08
[ImportVerilog] Support packed structs in `inside` operator

### LINKED ISSUE #8440: [ImportVerilog] `inside` expression only support simple bit vectors
The BlackParrot core in the sv-tests repository is currently failing on the `inside` expression only supporting simple bit vectors. I suspect that this is due to some enum/struct/union being passed to `inside`, which should be converted to a simple bit vector:

https://chipsalliance.github.io/sv-tests-results/?v=circt_verilog+black-parrot+bp_default

```
third_party/cores/black-parrot/bp_me/src/v/cce/bp_uce.sv:290:43: error: only simple bit vectors supported in 'inside' expressions
  wire miss_load_v_li   = cache_req_v_i & cache_req_cast_i.msg_type inside {e_miss_load};
```

AnnuCode:
Fixes #8440.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -637,12 +637,7 @@ struct RvalueExprVisitor {
         cond = builder.create<moore::AndOp>(loc, leftValue, rightValue);
       } else {
         // Handle expressions.
-        auto ty = listExpr->type;
-        bool isUnPackedStruct = ty->isUnpackedStruct();
-        bool isUnPackedArray = ty->isUnpackedArray();
-        bool isUnPackedUnion = ty->isUnpackedUnion();
-
-        if (isUnPackedArray || isUnPackedStruct || isUnPackedUnion) {
+        if (!listExpr->type->isIntegral()) {
           if (listExpr->type->isUnpackedArray()) {
             mlir::emitError(
                 loc, "unpacked arrays in 'inside' expressions not supported");
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -637,14 +637,9 @@ struct RvalueExprVisitor {
         cond = builder.create<moore::AndOp>(loc, leftValue, rightValue);
       } else {
         // Handle expressions.
-        if (!listExpr->type->isSimpleBitVector()) {
-          if (listExpr->type->isUnpackedArray()) {
-            mlir::emitError(
-                loc, "unpacked arrays in 'inside' expressions not supported");
-            return {};
-          }
+        if (listExpr->type->isUnpackedArray()) {
           mlir::emitError(
-              loc, "only simple bit vectors supported in 'inside' expressions");
+              loc, "unpacked arrays in 'inside' expressions not supported");
           return {};
         }
         auto value = context.convertToSimpleBitVector(
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -637,11 +637,22 @@ struct RvalueExprVisitor {
         cond = builder.create<moore::AndOp>(loc, leftValue, rightValue);
       } else {
         // Handle expressions.
-        if (listExpr->type->isUnpackedArray()) {
+        auto ty = listExpr->type;
+        bool isUnPackedStruct = ty->isUnpackedStruct();
+        bool isUnPackedArray = ty->isUnpackedArray();
+        bool isUnPackedUnion = ty->isUnpackedUnion();
+
+        if (isUnPackedArray || isUnPackedStruct || isUnPackedUnion) {
+          if (listExpr->type->isUnpackedArray()) {
+            mlir::emitError(
+                loc, "unpacked arrays in 'inside' expressions not supported");
+            return {};
+          }
           mlir::emitError(
-              loc, "unpacked arrays in 'inside' expressions not supported");
+              loc, "only simple bit vectors supported in 'inside' expressions");
           return {};
         }
+
         auto value = context.convertToSimpleBitVector(
             context.convertRvalueExpression(*listExpr));
         if (!value)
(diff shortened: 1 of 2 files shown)


test/Conversion/ImportVerilog/basic.sv
```diff
@@ -1545,6 +1545,22 @@ module Expressions;
     // CHECK: moore.struct_create [[TMP0]], [[TMP1]] : !moore.i32, !moore.i32 -> struct<{a: i32, b: i32}>
     struct0 = '{43, 9002};
 
+    // CHECK: [[TMP0:%.+]] = moore.constant 2
+    // CHECK: [[TMP1:%.+]] = moore.constant 3
+    // CHECK: moore.struct_create [[TMP0]], [[TMP1]] : !moore.i32, !moore.i32 -> struct<{a: i32, b: i32}>
+    struct0 = '{2,3};
+
```
fabianschuiki: I think you don't really need this reassignment of `struct0`, since it creates more or less the same IR as the test just above it.
```suggestion
```

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -1545,6 +1545,22 @@ module Expressions;
...
+    // CHECK: [[TMP0:%.+]] = moore.constant 2 : i32
+    // CHECK: [[TMP1:%.+]] = moore.sext [[TMP0]] : i32 -> i64
+    // CHECK: [[TMP2:%.+]] = moore.read %struct0 : <struct<{a: i32, b: i32}>>
+    // CHECK: [[TMP3:%.+]] = moore.conversion [[TMP2]] : !moore.struct<{a: i32, b: i32}> -> !moore.i64
+    // CHECK: [[TMP4:%.+]] = moore.wildcard_eq [[TMP1]], [[TMP3]] : i64 -> i1
+    // CHECK: [[TMP5:%.+]] = moore.zext [[TMP4]] : i1 -> i32
+    // CHECK: [[TMP6:%.+]] = moore.conversion [[TMP5]] : !moore.i32 -> !moore.l32
+    // CHECK: [[TMP7:%.+]] = moore.conversion [[TMP6]] : !moore.l32 -> !moore.i32
+    // CHECK: moore.blocking_assign %c, [[TMP7]] : i32
+    c = 2 inside {struct0};
```
fabianschuiki: I'm wondering what the `inside` operator is supposed to do according to the SystemVerilog spec if there's a struct on its right-hand side. The current behavior just flattens the struct into an equivalent vector of raw bits, and then performs an equality check with the left-hand side. In this case this would essentially check `2 == 8589934595 (2<<32 | 3)`.

AnnuCode: I see, that's interesting. Am I right in assuming that it might not be legal to use packed types on the right-hand side of the `inside` operator? While reading the standard, I couldn't find any mention or example in which rhs is a a packed struct/array. There is a mention of unpacked arrays and the `circt-verilog` tool already handles it. What do you think?

fabianschuiki: Hmmm I think your implementation is correct. I only find the following in the standard, which you also mentioned:
image
This suggests that *unpacked arrays* get special treatment, but all other types behave as if they would be on the right-hand-side of a comparison (`==` or `==?` depending on whether it's an integral type or not). So packed types would basically have to be converted to a simple bit vector before comparison, and any other types would just be compared directly.

The current implementation we have only works for packed types. So it might make sens to check `!listExpr->type->isPacked()` as a check instead of `!listExpr->type->isSimpleBitVector()`, to allow any packed type to be used and converted to a simple bit vector. Then we'd get the two error messages: "unpacked array unsupported" for unpacked arrays, and the "only simple bit vectors supported in 'inside' expressions" otherwise.

AnnuCode: Thanks! I'll do that.

AnnuCode:
@fabianschuiki Thanks for the suggestions. The `else` block now:

- emits a general error if rhs is an unpacked struct/union
- emits a special error if rhs is an unpacked array
- converts all packed types to a simple bit vector

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -637,7 +637,12 @@ struct RvalueExprVisitor {
         cond = builder.create<moore::AndOp>(loc, leftValue, rightValue);
       } else {
         // Handle expressions.
-        if (!listExpr->type->isSimpleBitVector()) {
+        auto ty = listExpr->type;
+        bool isUnPackedStruct = ty->isUnpackedStruct();
+        bool isUnPackedArray = ty->isUnpackedArray();
+        bool isUnPackedUnion = ty->isUnpackedUnion();
+
+        if (isUnPackedArray || isUnPackedStruct || isUnPackedUnion) {
```
fabianschuiki: I would use something like `!ty->isPacked()` here, instead of listing concrete unpacked types. For example, things like a `string` is also unpacked, which would be missed here. It's often easier to use a whitelist, i.e. error out if the type isn't strictly what you support, instead of using a blacklist, i.e. making a list of all the things that are not supported.

AnnuCode: I agree. I first tried to use `listExpr->type->ispacked()`, but couldn't as it says: class "slang::ast::Type" has no member "isPacked". Should I also add a `isString()` check?

fabianschuiki: Oh yeah, I hadn't realized that `isPacked()` doesn't exist. Browsing through `slang/ast/types/Type.h` a bit, it looks like `isIntegral()` is a way to determine if the type is essentially packed:
```c++
    /// Indicates whether this is an integral type, which includes all scalar types, predefined
    /// integer types, packed arrays, packed structures, packed unions, and enum types.
    bool isIntegral() const;
```
Maybe something like 
```c++
if (!ty->isIntegral()) {
  // ...
}
```
could work in your code?

AnnuCode: That will work, thanks! 💯

fabianschuiki (approved):
Looks really nice! Thanks a lot for adding this :100:. (Let me know if you want me to merge this for you.)

AnnuCode:
Thanks! Yes, please merge this.

PR #8546, opened 2025-06-09
[FIRRTL] Dedup: MustDedup support merging of pub+priv modules

youngar:
We upgraded dedup to be able to merge modules where one is public, as long as the other modules were private.  The MustDedup error message was not updated to handle this change in behaviour, and still assumed that it was impossible to merge any public module with other modules.  This changes MustDedup to only detect this as a difference if neither module is marked as private.

### DIFF
--- a/test/Dialect/FIRRTL/dedup-errors.mlir
+++ b/test/Dialect/FIRRTL/dedup-errors.mlir
@@ -666,9 +666,9 @@ firrtl.circuit "InstOfEquivPublic" attributes {annotations = [{
 // expected-error @below {{module "Test1" not deduplicated with "Test0"}}
 firrtl.circuit "VisibilityDoesNotBlockDedup" attributes {annotations = [{
       class = "firrtl.transforms.MustDeduplicateAnnotation",
-      modules = ["~VisibilityDoesNotBlockDedup|Test0", "~VisibilityDoesNotBlockDedup|Test1"]
+      modules = ["~|Test0", "~|Test1"]
     }]} {
-  firrtl.module private @Test0() {
+  firrtl.module @Test0() {
     // expected-note @below {{operation result types don't match, first type is '!firrtl.uint<1>'}}
     %wire = firrtl.wire : !firrtl.uint<1>
   }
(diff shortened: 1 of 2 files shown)


test/Dialect/FIRRTL/dedup-errors.mlir
```diff
@@ -656,3 +657,27 @@ firrtl.circuit "InstOfEquivPublic" attributes {annotations = [{
...
+
+// -----
+
+// When comparing two modules, if one is private and the other public, we should
+// not report that as the difference, we should examine the bodies of the 
+// modules.
+// expected-error @below {{module "Test1" not deduplicated with "Test0"}}
+firrtl.circuit "VisibilityDoesNotBlockDedup" attributes {annotations = [{
+      class = "firrtl.transforms.MustDeduplicateAnnotation",
+      modules = ["~VisibilityDoesNotBlockDedup|Test0", "~VisibilityDoesNotBlockDedup|Test1"]
```
seldridge: Terser target syntax may work here:

```suggestion
      modules = ["~|Test0", "~|Test1"]
```

test/Dialect/FIRRTL/dedup-errors.mlir
```diff
@@ -656,3 +657,27 @@ firrtl.circuit "InstOfEquivPublic" attributes {annotations = [{
...
+    }]} {
+  firrtl.module private @Test0() {
+    // expected-note @below {{operation result types don't match, first type is '!firrtl.uint<1>'}}
+    %wire = firrtl.wire : !firrtl.uint<1>
+  }
```
seldridge: Forgot to mark `public`?

```suggestion
  firrtl.module @Test0() {
    // expected-note @below {{operation result types don't match, first type is '!firrtl.uint<1>'}}
    %wire = firrtl.wire : !firrtl.uint<1>
  }
```

Does this test fail with the patch removed?

youngar: 🤦  thanks. Yeah, it still works.

test/Dialect/FIRRTL/dedup-errors.mlir
youngar: Yeah, this works.

mwachs5:
thanks!

seldridge (approved):
LGTM

dtzSiFive (approved):

PR #8550, opened 2025-06-10
[PyRTG] Add type wrappers

maerhart:
Add python classes to represent the type of values such that the construction of MLIR types can be delayed until an MLIR context is created.

### DIFF
--- a/frontends/PyRTG/src/pyrtg/support.py
+++ b/frontends/PyRTG/src/pyrtg/support.py
@@ -3,7 +3,9 @@
 #  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from .base import support, ir
-from .core import Value
+from .core import Value, Type
+
+from typing import Union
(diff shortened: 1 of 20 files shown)


frontends/PyRTG/src/pyrtg/support.py
```diff
@@ -55,6 +57,61 @@ def _FromCirctValue(value: ir.Value) -> Value:
...
+  if isinstance(type, Type):
+    return type
+
+  type = support.type_to_pytype(type)
+
+  from .rtg import rtg
+  from .rtgtest import rtgtest
+  if isinstance(type, rtg.ArrayType):
+    from .arrays import ArrayType
+    return ArrayType(_FromCirctType(type.element_type))
```
fabianschuiki, line 70: That's a nice pattern!

fabianschuiki (approved):
LGTM!

PR #8551, opened 2025-06-10
[ESI] Extend instead of override CMAKE_CXX_FLAGS

mortbopet:


teqdruid:
The problem with this approach is that a flag disabling exceptions may be
in the flags already. So appending isn't enough.

I think I looked into it awhile back, but cmake should really have a
function to deal with this.

teqdruid:
> The problem with this approach is that a flag disabling exceptions may be in the flags already. So appending isn't enough.

Which is the situation when the runtime is compiled in a CIRCT context.

mortbopet:
Sure, but that's not a blank check to override all CXX flags. In that case, the ESI runtime should scan the existing flags and remove any flags that it definitely does _not_ want to build with.
An example - in an internal build, it is required that `/FS` is set on _all_ targets, when using sccache.  
This, unfortunately, is impossible when ESI overrides all flags - and we do _not_ want to manually add `/FS` to the one-to-many targets which the ESI runtime builds (i.e. the runtime itself, perhaps zlib if it's also building that, ...).

teqdruid:
Oh sure. What it currently does is pretty bad. I'm not arguing against
changing it.

There's some cmake code somewhere else in CIRCT (or at least used to be)
which did a string replace based removal... I never liked that approach but
we can use it here since it's better than what we're currently doing. Can
you think of any other way? I'm surprised that cmake doesn't have
functionality for exactly this purpose. Is there something I'm missing?

mortbopet:
@teqdruid unfortunately, i am not aware of anything better... added a regex replace for exception-related flags, _hopefully_ that is a sufficient enough catch-all.

PR #8555, opened 2025-06-10
[FIRRTL] Extract instances above DUT w/ moveDut

seldridge:
Preserve the controlling annotation for FIRRTL's hierarchy injection pass.
This will be used to augment the behavior of FIRRTL's instance extraction
pass.  Note: this removes the idempotent property of hierarchy injection.

Modify newly added behavior related to the `moveDut` parameter added to
`InjectDUTHierarchyAnnotation` [[1]].  Change the behavior so, when the
`moveDut` parameter in the newly preserved annotation, instances will be
extracted one level above the newly _moved_ DUT.

This adds tech debt as it repurposes instance extraction to allow for both
hierarchy injection and extraction beyond the DUT.  Previously, these were
intended to be mutually exclusive.

The tech debt is acceptable because @mwachs5 is planning to remove this
flow entirely.

[1]: 91d94336afb65330a3b9f265aa2ad3d463a19a07

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
@@ -166,6 +166,11 @@ struct ExtractInstancesPass
   /// create a path to the instance
   DenseMap<InnerRefAttr, InstanceOp> innerRefToInstances;
   Type stringType, pathType;
+
+  /// If set, this indicates that the `InjectDUTHierarchy` pass ran with the
+  /// `moveDut` parameter enabled.  If true, then extraction continues outside
+  /// the DUT even when extraction annotations have a wrapper module specified.
+  bool moveDut;
 };
 } // end anonymous namespace
 
@@ -190,6 +195,7 @@ void ExtractInstancesPass::runOnOperation() {
   auto *context = circuitOp->getContext();
   stringType = StringType::get(context);
   pathType = PathType::get(context);
+  moveDut = false;
 
   // Walk the IR and gather all the annotations relevant for extraction that
   // appear on instances and the instantiated modules.
@@ -232,6 +238,23 @@ static bool isAnnoInteresting(Annotation anno) {
 /// populates the corresponding lists and maps of the pass.
 void ExtractInstancesPass::collectAnnos() {
   CircuitOp circuit = getOperation();
+
+  // Find an optional `InjectDUTHierarchyAnnotation`.  If it exists, inspect the
+  // `moveDut` field.  If this is `true`, then we moved the DUT from the
+  // original DUT to the wrapper.  If this occurred, then this affecst the
+  // behavior of whether or not we stop at the DUT (now the wrapper) when we
+  // extract instances.
+  //
+  // TODO: This is tech debt.  This was accepted on condition that work is done
+  // to remove this pass.
+  AnnotationSet::removeAnnotations(circuit, [&](Annotation anno) {
+    if (!anno.isClass(injectDUTHierarchyAnnoClass))
+      return false;
+
+    if (auto moveDutAnnoAttr = anno.getMember<BoolAttr>("moveDut"))
+      moveDut = moveDutAnnoAttr.getValue();
+    return true;
+  });
 
   // Grab the clock gate extraction annotation on the circuit.
   StringRef clkgateFileName;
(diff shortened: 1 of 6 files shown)


lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -166,6 +166,16 @@ struct ExtractInstancesPass
   /// create a path to the instance
   DenseMap<InnerRefAttr, InstanceOp> innerRefToInstances;
   Type stringType, pathType;
+
+  /// If set, this indicates that the `InjectDUTHierarchy` pass ran with the
+  /// `moveDut` parameter enabled.  If true, then extraction continues outside
+  /// the DUT even when extraction annotations have a wrapper module specified.
+  bool moveDut = false;
```
uenoku: Please initialize `moveDut` at the beginning of `runOnOperation` since this is a member of a pass.

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
```diff
@@ -166,6 +166,16 @@ struct ExtractInstancesPass
...
+
+  /// Set the `stopAtDut` field correctly based on the `moveDut` settings.
+  void setStopAtDUT(ExtractionInfo &info) {
+    info.stopAtDUT = !moveDut && !info.wrapperModule.empty();
+  }
```
uenoku: Could you delete `stopAtDut` in `ExtractInfo` and define `getStopAtDut(bool moveDut) { return moveDut && wrapperModule.empty();`}` instead? I think with that change you don't have to call `setStopAtDut` every ExtractInfo creation.

lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
```diff
@@ -163,7 +163,24 @@ void InjectDUTHierarchy::runOnOperation() {
...
+    // already outside the specification, this workflow is allowed even though
+    // it violates the FIRRTL ABI.  The mid-term plan is to remove this pass to
+    // avoid the tech debt that it creates.
+    if (moveDut) {
+      SymbolTable::setSymbolVisibility(newDUT,
+                                       SymbolTable::Visibility::Private);
+    } else {
+      SymbolTable::setSymbolVisibility(newDUT, dut.getVisibility());
+      SymbolTable::setSymbolVisibility(dut, SymbolTable::Visibility::Private);
+    }
```
uenoku: nit:
```suggestion
    if (moveDut) {
    newDut.setPrivate();
    } else {
     newDut.setVisibility(dut.getVisibility())
     ...
    }
```

uenoku (approved):
LGTM. Could you mention that ExtractInstance must be ran after InjectDut pass in pipeline generation Firtool.cpp?

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
seldridge: Yeah, this is better. Nice thought.

lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
seldridge: Nice! I didn't know about this!

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
seldridge: Fixed in: https://github.com/llvm/circt/pull/8555/commits/b0d161d4da223ccc71b80e68f77fd92198cecab6

lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
seldridge: Fixed in: https://github.com/llvm/circt/pull/8555/commits/b0d161d4da223ccc71b80e68f77fd92198cecab6

lib/Dialect/FIRRTL/Transforms/ExtractInstances.cpp
seldridge: Fixed in: https://github.com/llvm/circt/pull/8555/commits/b4c53873755bd1696767e89da7fef14c17b3e1c5

PR #8556, opened 2025-06-11
[OM][Python] Support string and list as evaluator arguments

uenoku:
Add support for passing Python strings and lists directly as arguments to the OM evaluator's instantiate method:

* Fix list_concat evaluation to work with list constants (previously operand must be ListValue, but changed to allow AttributeValue as well).
* Add StringType C API to construct typed strings in Python bridge.
* Implement conversion between Python lists and OM ListAttr

### DIFF
--- a/lib/Dialect/OM/Evaluator/Evaluator.cpp
+++ b/lib/Dialect/OM/Evaluator/Evaluator.cpp
@@ -957,6 +957,8 @@ circt::om::evaluator::AttributeValue::get(Attribute attr, LocationAttr loc) {
   if (!loc)
     loc = UnknownLoc::get(context);
 
+  // Special handling for ListType to create proper ListValue objects instead of
+  // AttributeValue objects.
   if (auto listType = dyn_cast<circt::om::ListType>(type)) {
     SmallVector<EvaluatorValuePtr> elements;
     auto listAttr = cast<om::ListAttr>(attr);
@@ -976,6 +978,9 @@ circt::om::evaluator::AttributeValue::get(Type type, LocationAttr loc) {
   auto *context = type.getContext();
   if (!loc)
     loc = UnknownLoc::get(context);
+
+  // Special handling for ListType to create proper ListValue objects instead of
+  // AttributeValue objects.
   if (auto listType = dyn_cast<circt::om::ListType>(type))
     return std::make_shared<evaluator::ListValue>(listType, loc);
   // Create the AttributeValue with the private tag


lib/Bindings/Python/OMModule.cpp
```diff
@@ -445,13 +445,36 @@ static MlirAttribute omPythonValueToPrimitive(PythonPrimitive value,
...
 
   if (auto *attr = std::get_if<nb::bool_>(&value)) {
     return mlirBoolAttrGet(ctx, nb::cast<bool>(*attr));
   }
 
+  // For a python list try constructing OM list attribute. The element
+  // type must be uniform.
+  if (auto *attr = std::get_if<nb::list>(&value)) {
+    if (attr->size() == 0)
+      throw nb::type_error("Empty list is prohibited now");
```
mikeurbach, line 460: This restriction is because we need at least one element to infer the result type, right?

lib/Dialect/OM/Evaluator/Evaluator.cpp
```diff
@@ -619,6 +619,15 @@ circt::om::Evaluator::evaluateListConcat(ListConcatOp op,
...
       return list;
 
+    if (evaluator::AttributeValue *attr =
+            llvm::dyn_cast<evaluator::AttributeValue>(result.value().get())) {
+      auto attributes = getEvaluatorValuesFromAttributes(
+          attr->getAttr().getContext(),
+          attr->getAs<om::ListAttr>().getElements().getValue());
+      values.append(attributes.begin(), attributes.end());
+      continue;
+    }
```
mikeurbach: This part makes sense, but I'm wondering if it can be handled more centrally. This is the code that handles the operands of list concat... what happens as we add other kinds of list operations, do they also need to handle list attributes? I'm not sure the best solution here, just wanted to bring it up.

mikeurbach (approved):
A couple questions but I think this looks good. Thanks for extending this support.

lib/Bindings/Python/OMModule.cpp
uenoku: Yes. We could use !om.any as an element type (or maybe just using om.any regardless of element type might be better) but I think we can modify it later.

lib/Dialect/OM/Evaluator/Evaluator.cpp
uenoku: Good point. I think alternatively we can initialize LIstAttr as ListValue of individual fields.. probably that's better :)

uenoku: I changed to generate ListValue in the ListAttr initialization so this part can be removed.

lib/Dialect/OM/Evaluator/Evaluator.cpp
```diff
@@ -949,3 +949,36 @@ LogicalResult circt::om::evaluator::AttributeValue::finalizeImpl() {
...
 }
+
+std::shared_ptr<evaluator::EvaluatorValue>
+circt::om::evaluator::AttributeValue::get(Attribute attr, LocationAttr loc) {
+  auto type = cast<TypedAttr>(attr).getType();
+  auto *context = type.getContext();
+  if (!loc)
+    loc = UnknownLoc::get(context);
+
+  if (auto listType = dyn_cast<circt::om::ListType>(type)) {
```
uenoku, line 962: Add comments.

sequencer:
Just reminder, it may break the downstream, see:
https://github.com/sequencer/zaozi/actions/runs/15766803098/job/44444692938?pr=62#step:4:3268
we will follow this PR for change our downstream.

PR #8557, opened 2025-06-11
[FIRRTL] Make CatPrimOp variadic

uenoku:
This patch converts CatPrimOp from a binary operation to a variadic operation that can accept multiple operands. The change is mostly mechanical:
- Updating the fold method to handle multiple operands
- Modifying canonicalization patterns to work with the new variadic form (https://mlir.llvm.org/docs/DeclarativeRewrites/#supporting-variadic-ops)
- Updating the FIRRTL to HW lowering to handle variadic concatenation
- Adjusting the FIREmitter FIRRTLVistior


lib/Dialect/FIRRTL/Export/FIREmitter.cpp
seldridge, line 1: Does the emitter continue to work?

uenoku: Yes, emitter is robust about the number of operands. [test/Dialect/FIRRTL/emit-basic.mlir](https://github.com/llvm/circt/pull/8557/files#diff-9c1dce56be0bc780a1a0040e6fd66aff955e4038a1ca670f11e018ccaeabcd61)

seldridge: While this emits something, it's not legal FIRRTL: 

```
Foo.fir:6:22: error: expected ')'
    node b = cat(a, a, a)
```

uenoku: Ok. I thought it's fine because after the PR that IR will be parsed but I can legalize it binary operation.

seldridge: And I'm totally fine with moving in the variadic direction on upstream FIRRTL. There's not a great reason to _not_ have variadic ops only that it was never done this way before. I think all the reasons for using variadic in HW still hold for FIRRTL.

uenoku: bca6c3900939020c7477c070e95fd6bf502faf25 adds a support for emitting older version of firrtl. I'd like to include variadic cat in fir 5.1.0. WDYT?

seldridge: Yep, I'm all in favor of changing this on the spec, and for any commutative operation.

PR #8558, opened 2025-06-11
[ESI] Set mode before `connectImpl`

mortbopet:
`connectImpl` is shared across all connect functions. However, if `mode` is only set _after_ `connectImpl` returns, `connectImpl` doesn't have a direct way to check the connect mode (we could look for whether or not the callback has been assigned, but that's then also placing an assumption on the ordering of things in `Ports.cpp`).

Obviously we could also override the callback and non-callback connect functions, respectively, but i feel like that may be unnecessary complexity, in case `impl` functions share most of the code.


teqdruid:
The 'mode' variable indicates the state of the connection (Disconnected, Callback, Polling). Therefore, it cannot be set to non-disconnected until after it's connected. I seem to recall that I used to have it like this but there was a race condition wherein some other thread would call read and read would try to do something because it saw that the port was not-disconnected.

Why do you need this information? The expectation is that backends could be oblivious to the mode if they just use `callback` to deliver new data.

PR #8559, opened 2025-06-11
[ImportVerilog] Add support for concurrent assertions

towoe:
Start of support for SystemVerilog SVAs.
Make use of existing LTL operations.

@fabianschuiki would be great if you could have a look at this.
I am not so sure about the handling of the expression types.
Right now I allow 'moore.l1' to 'i1'. It allows one bit wide types, and sets it all to 'i1'. I don't think this is a problem, as the SV output is also always using `i1`, I think. If the conversion results in 'x' and 'z' to evaluate to false, should be correct.

I based the implementation of the operations on [LTL dialect](https://circt.llvm.org/docs/Dialects/LTL/#representing-svas)

As the concurrent operations rely on a clock, further down the road there is a problem with `llhd.process` and `cf.br`, but this might also be related to the passes I use in my local example flow. For now I would concentrate on getting the parsing done and worry later about the output.

I am also not so sure about the implementation of the visitors, would be great to get some feedback here.

If the slang bump goes through, I will do a rebase and remove the WIP.

### DIFF
--- a/lib/Conversion/ImportVerilog/AssertionExpr.cpp
+++ b/lib/Conversion/ImportVerilog/AssertionExpr.cpp
@@ -86,8 +86,6 @@ struct AssertionExprVisitor {
     // Convert first to a moore type and then add a conversion to `i1`.
     auto value = context.convertRvalueExpression(expr.expr);
     auto loc = context.convertLocation(expr.expr.sourceRange);
-    if (!value)
-      return {};
     value = context.convertToI1(value);
     if (!value)
       return {};
@@ -99,10 +97,7 @@ struct AssertionExprVisitor {
 
     // There is a repetition, embed the expression into the kind of given
     // repetition
-    Value repeatedValue = createRepetition(loc, expr.repetition.value(), value);
-    if (!repeatedValue)
-      return {};
-    return repeatedValue;
+    return createRepetition(loc, expr.repetition.value(), value);
   }
 
   Value visit(const slang::ast::SequenceConcatExpr &expr) {
@@ -270,10 +265,7 @@ struct AssertionExprVisitor {
     auto assertionExpr = context.convertAssertionExpression(expr.expr, loc);
     if (!assertionExpr)
       return {};
-    auto result = context.convertLTLTimingControl(expr.clocking, assertionExpr);
-    if (!result)
-      return {};
-    return result;
+    return context.convertLTLTimingControl(expr.clocking, assertionExpr);
   }
 
   /// Emit an error for all other expressions.


lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+    // - Expression
+    // - Optional repetition
+
+    // The expression needs to have the type `i1` for the LTL operations.
+    // Convert first to a moore type and then add a conversion to `i1`.
+    auto value = context.convertRvalueExpression(expr.expr);
+    auto loc = context.convertLocation(expr.expr.sourceRange);
+    if (!value) {
+      mlir::emitError(loc,
+                      "Failed to convert expression in SimpleAssertionExpr");
```
fabianschuiki: No need to emit an error here; `convertRvalueExpression` should already do the reporting for you.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+      return {};
+    }
+    value = context.convertToI1(value);
+    if (!value) {
+      mlir::emitError(
+          loc,
+          "Failed to convert expression to i1 type in SimpleAssertionExpr - "
+          "expression may be too wide for boolean conversion");
```
fabianschuiki: Same; `convertToI1` should already emit a diagnostic if conversion fails.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+    if (!expr.repetition.has_value()) {
+      return value;
+    }
+
+    // There is a repetition, embed the expression into the kind of given
+    // repetition
+    Value repeatedValue = createRepetition(loc, expr.repetition.value(), value);
+    if (!repeatedValue) {
+      mlir::emitError(loc,
+                      "Failed to create repetition for SimpleAssertionExpr");
```
fabianschuiki: Same as a above: your `createRepetition` function already nicely reports why the repetition cannot be created, so no need to emit an error again here :+1:

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+      if (!sequenceValue)
+        continue;
+
+      Type valueType = sequenceValue.getType();
+      if (!valueType.isInteger(1) && !mlir::isa<ltl::SequenceType>(valueType)) {
+        mlir::emitError(
+            loc,
+            "SequenceConcatExpr element must be of type i1 or ltl.sequence");
+        return {};
+      }
```
fabianschuiki: I think this is guaranteed to be the case by the Slang frontend. During type checking it already emits an error if you are trying to e.g. concatenate two properties. In a case like this where we rely on some property of the frontend which isn't developed as part of CIRCT, it would make sense to move this into an assert:
```suggestion
      assert(valueType.isInteger(1) || mlir::isa<ltl::SequenceType>(valueType));
```

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+    if (expr.elements.empty())
+      return {};
+
+    SmallVector<Value> sequenceElements;
+
+    for (const auto &concatElement : expr.elements) {
+      Value sequenceValue =
+          context.convertAssertionExpression(*concatElement.sequence, loc);
+      if (!sequenceValue)
+        continue;
```
fabianschuiki: I think you'll want to return immediately here, otherwise errors in assert exprs get reported but we might still construct some half-correct IR where the problematic elements are missing.
```suggestion
      if (!sequenceValue)
        return {};
```
Man I miss Rust's `?` operator :cry:

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+                      "Failed to create repetition for SimpleAssertionExpr");
+      return {};
+    }
+    return repeatedValue;
+  }
+
+  Value visit(const slang::ast::SequenceConcatExpr &expr) {
+    // Create a sequence of delayed operations, combined with a concat operation
+    if (expr.elements.empty())
+      return {};
```
fabianschuiki: I think it's impossible to create an empty concat expression, since they are an infix operator in SV. We should be able to just assumpe that there's at least one element in the concatenation, and maybe assert that the assumption holds:
```suggestion
    // Create a sequence of delayed operations, combined with a concat operation
    assert(!expr.elements.empty());
```

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+          convertRangeToAttrs(concatElement.delay.min, concatElement.delay.max);
+      auto delayedSequence = builder.create<ltl::DelayOp>(
+          loc, sequenceValue, delayMin, delayRange);
+      sequenceElements.push_back(delayedSequence);
+    }
+
+    if (!sequenceElements.empty())
+      return builder.createOrFold<ltl::ConcatOp>(loc, sequenceElements);
+
+    return {};
```
fabianschuiki: I think you can just directly return the concat op you've created. There should never be a situation where we have a different number of items in `sequenceElements` than the original AST expression. Might even be worth asserting that (assert might be overkill though):
```suggestion
    assert(sequenceElements.size() == expr.elements.size());
    return builder.createOrFold<ltl::ConcatOp>(loc, sequenceElements);
```

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+                                             repetitionRange);
+      } else {
+        mlir::IntegerAttr repetitionRange;
+        return builder.create<ltl::RepeatOp>(
+            loc, value, builder.getI64IntegerAttr(0), repetitionRange);
+      }
+    case UnaryAssertionOperator::NextTime:
+      if (expr.range.has_value()) {
+        auto minRepetitions = mlir::IntegerAttr::get(builder.getIntegerType(64),
+                                                     expr.range.value().min);
```
fabianschuiki: The builder has a nice convenience function for this (not sure if this is the exact name :see_no_evil:):
```suggestion
        auto minRepetitions = builder.getI64IntegerAttr(expr.range.value().min);
```

fabianschuiki: Haha stupid me just realized that that's exactly what you've been using below :grin:. Might be worth using here as well.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+        return builder.create<ltl::DelayOp>(loc, value,
+                                            builder.getI64IntegerAttr(1),
+                                            builder.getI64IntegerAttr(0));
+      }
+    case UnaryAssertionOperator::Eventually:
+    case UnaryAssertionOperator::SNextTime:
+    case UnaryAssertionOperator::SAlways:
+      mlir::emitError(loc, "unsupported unary operator: ")
+          << slang::ast::toString(expr.op);
+      return {};
```
fabianschuiki, line 165: Nice! Thanks for covering the unsupported stuff with a diagnostic :1st_place_medal:

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+  Value visit(const slang::ast::ClockingAssertionExpr &expr) {
+    auto assertionExpr = context.convertAssertionExpression(expr.expr, loc);
+    if (!assertionExpr)
+      return {};
+    auto result = context.convertLTLTimingControl(expr.clocking, assertionExpr);
+    if (!result) {
+      mlir::emitError(loc, "Failed to convert clocking assertion expression");
+      return {};
+    }
```
fabianschuiki:
```suggestion
    if (!result)
      return {};
```

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+/// Helper function to convert a value to an i1 value.
+Value Context::convertToI1(Value value) {
+  if (!value)
+    return {};
+  if (auto type = dyn_cast_or_null<moore::IntType>(value.getType()))
+    if (type.getBitSize() != 1) {
+      mlir::emitError(value.getLoc(), "expected a 1-bit integer");
+      return {};
+    }
```
fabianschuiki: LLVM likes to use `{...}` on either all or none of the nested ifs. You should also be able to use `dyn_cast`, since `value.getType()` will always return a type. The user might also do something weird where the value isn't a Moore `IntType` at all -- should be caught by the frontend, but it might be worth emitting an error then as well:
```suggestion
  auto type = dyn_cast<moore::IntType>(value.getType());
  if (!type || type.getBitSize() != 1) {
    mlir::emitError(value.getLoc(), "expected a 1-bit integer");
    return {};
  }
```

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -642,6 +642,36 @@ struct StmtVisitor {
...
+      }
+      mlir::emitError(loc) << "unsupported concurrent assertion kind: "
+                           << slang::ast::toString(stmt.assertionKind);
+      return failure();
+    }
+
+    mlir::emitError(loc)
+        << "concurrent assertion statements with action blocks "
+           "are not supported yet";
+    return failure();
```
fabianschuiki, line 672: Excellent idea to leave this for a later PR :+1:

lib/Conversion/ImportVerilog/TimingControls.cpp
```diff
@@ -87,6 +104,41 @@ struct DelayControlVisitor {
...
+      return Value{};
+    }
+    Value condition;
+    if (ctrl.iffCondition) {
+      condition = context.convertRvalueExpression(*ctrl.iffCondition);
+      condition = context.convertToBool(condition, Domain::TwoValued);
+      if (!condition) {
+        mlir::emitError(loc, "failed to convert iff condition");
+        return Value{};
+      }
```
fabianschuiki:
```suggestion
      if (!condition)
        return {};
```

lib/Conversion/ImportVerilog/TimingControls.cpp
```diff
@@ -87,6 +104,41 @@ struct DelayControlVisitor {
...
+  OpBuilder &builder;
+  Value seqOrPro;
+
+  Value visit(const slang::ast::SignalEventControl &ctrl) {
+    auto edge = convertEdgeKindLTL(ctrl.edge);
+    auto expr = context.convertRvalueExpression(ctrl.expr);
+    if (!expr) {
+      mlir::emitError(loc, "failed to convert expression");
+      return Value{};
+    }
```
fabianschuiki:
```suggestion
    if (!expr)
      return {};
```

lib/Conversion/ImportVerilog/TimingControls.cpp
```diff
@@ -87,6 +104,41 @@ struct DelayControlVisitor {
...
+    if (ctrl.iffCondition) {
+      condition = context.convertRvalueExpression(*ctrl.iffCondition);
+      condition = context.convertToBool(condition, Domain::TwoValued);
+      if (!condition) {
+        mlir::emitError(loc, "failed to convert iff condition");
+        return Value{};
+      }
+    }
+    expr = context.convertToBool(expr, Domain::TwoValued);
+    expr.setType(builder.getI1Type());
```
fabianschuiki: Forcing the type does not always work. It implicitly relies on the value being produced by an operation that can have its return type trivially changed. This works for `moore.conversion`, but `convertToBool` might not insert a conversion op at all if it isn't necessary. Your `convertToI1` helper would be an excellent thing to use here!

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -2315,6 +2315,31 @@ module ImmediateAssertiWithActionBlock;
...
+    // CHECK: [[CONV_B:%.+]] = moore.conversion [[READ_B]] : !moore.l1 -> i1
+    // CHECK: [[DELAY_B:%.+]] = ltl.delay [[CONV_B]], 1, 0 : i1
+    // CHECK: [[CONCAT_AB:%.+]] = ltl.concat [[DELAY_A]], [[DELAY_B]] : !ltl.sequence, !ltl.sequence
+    // CHECK: [[READ_CLK:%.+]] = moore.read [[CLK]] : <l1>
+    // CHECK: [[CONV_CLK:%.+]] = moore.conversion [[READ_CLK]] : !moore.l1 -> i1
+    // CHECK: [[CLK_OP:%.+]] = ltl.clock [[CONCAT_AB]], posedge [[CONV_CLK]] : !ltl.sequence
+    // CHECK: verif.assert [[CLK_OP:%.+]] : !ltl.sequence
+    // CHECK:   moore.return
+  // CHECK: }
+  assert property (@(posedge clk) (a ##1 b));
```
fabianschuiki: It's interesting that Slang wraps `assert property` up in a procedure :thinking: That might be worth investigating further in a later PR, when we sort out smaller details of the lowering :+1:

fabianschuiki (commented):
Hey @towoe, this is some really really excellent work! :raised_hands: :100:. Thanks a lot for working on this so thoroughly. I've added a few minor nitpicks on some CIRCT and ImportVerilog conventions.

The only thing missing in my opinion are tests for each SVA construct you've added support for. I'd keep this pretty simple and create a small `assert property` for each:
```systemverilog
assert property (not a);
assert property (s_eventually a);
assert property (always a);
assert property (always[n] a);
assert property (nexttime a);
assert property (nexttime[n] a);
// ...
```
Basically a test for every separate code path through your assertion expression visitor that converts some pattern of the AST into a bit of IR.

Besides that, this PR has the perfect size to land initial SVA support in ImportVerilog! I would recommend getting this landed in CIRCT as is before working on other ops or features.

Really cool stuff :partying_face:. Does this rely on the Slang bump?

fabianschuiki:
The failing tests look like you'll have to link the `CIRCTImportVerilog` library against `CIRCTLTL` and `CIRCTVerif`.

towoe:
> Hey @towoe, this is some really really excellent work! 🙌 💯. Thanks a lot for working on this so thoroughly. I've added a few minor nitpicks on some CIRCT and ImportVerilog conventions.

Thanks a lot @fabianschuiki for taking the time to review and also help with getting this right.

> The only thing missing in my opinion are tests for each SVA construct you've added support for. I'd keep this pretty simple and create a small `assert property` for each:
> 
> ```systemverilog
> assert property (not a);
> assert property (s_eventually a);
> assert property (always a);
> assert property (always[n] a);
> assert property (nexttime a);
> assert property (nexttime[n] a);
> // ...
> ```
> 
> Basically a test for every separate code path through your assertion expression visitor that converts some pattern of the AST into a bit of IR.
> 
> Besides that, this PR has the perfect size to land initial SVA support in ImportVerilog! I would recommend getting this landed in CIRCT as is before working on other ops or features.

Makes sense, it will also help me add some more features iteratively.

> Really cool stuff 🥳. Does this rely on the Slang bump? If it doesn't _need_ the bump, I'd not bother waiting for that to land, to keep things happening in parallel.

For this PR it all works fine for the old Slang version.
I will address your comments and add more tests and update this PR soon.
Thanks again for the support!

towoe:
> The failing tests look like you'll have to link the `CIRCTImportVerilog` library against `CIRCTLTL` and `CIRCTVerif`.

Thanks for pointing this out. I never tested it with `BUILD_SHARED_LIBS=ON`, but with the dependencies you mentioned I saw it succeed in building locally.

lib/Conversion/ImportVerilog/TimingControls.cpp
towoe: Yes, thanks for catching this! I used to use setType a few times before I realized what it actually does. Now it all seems so simple with the `convertToI1` helper :)

lib/Conversion/ImportVerilog/TimingControls.cpp
towoe: And here I was thinking one could never have enough error messages ;)
Thanks for helping clean this up!

lib/Conversion/ImportVerilog/AssertionExpr.cpp
towoe: Makes sense since we now also return early if there is an invalid value in loop.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
towoe: Only a selfish reason to make it easier for myself in the future :)

lib/Conversion/ImportVerilog/AssertionExpr.cpp
towoe: Thanks for the detailed explanation!

lib/Conversion/ImportVerilog/Statements.cpp
towoe: Somehow I think this should not be too much work to add, but just not sure how to do it.

test/Conversion/ImportVerilog/basic.sv
towoe: It might be related to the `ifTrue` part of the assertion, which I do not yet handle, so it would have a place for the actions.

Is that also where the `llhd.process` is coming from?

towoe:
@fabianschuiki quick question on writing the tests. Is there a tool which generates the `// CHECK: ....` lines or is it all a manual adoption?

lib/Conversion/ImportVerilog/Statements.cpp
fabianschuiki: I think the big question is going to be where you would place the IR ops for the stuff in the action blocks. You can have basically arbitrary SV code in the action blocks, so these would probably have to get lowered into some procedural region. But `verif.assert` has no such procedural region, and there's also no Verif op that can check an LTL property and then execute a region of code. So this might actually require a new IR operation to handle -- which isn't a bad thing.

test/Conversion/ImportVerilog/basic.sv
fabianschuiki: Yes exactly! The procedure seems to get inserted by Slang implicitly, and ImportVerilog then converts it to an `llhd.process`. It's not entirely clear to me how the concurrent asserts should behave in a procedure, or how they interact with a procedure. I remember that this is allowed according to the SV spec:
```
always begin
  assert property (@(posedge clk) (a ## b));
end
```
But it also has some potentially strange interactions with clocks on the always block, or if the `assert property` is placed inside an `if`. We'll need to figure out what exactly to do there later down the road. We can probably also not worry too much about this and check if there's an "artificial" `always` procedure that got inserted by Slang, and simply skip it and inline the asserts into the module directly. No need to do any of this for a first shot at SVAs 😃

fabianschuiki:
@towoe It's usually a manual thing. I remember there being a tool for LLVM that did some of this, but the problem is always that you don't want to check every single operation and the _exact_ `%42` name. Instead, you'll likely just want to check the few operations you are interested in, and use patterns like `[[TMP:%.+]] = my.op ...` to match any SSA value names in the output.

lib/Conversion/ImportVerilog/TimingControls.cpp
fabianschuiki: https://www.youtube.com/watch?v=jr0JaXfKj68 😏

towoe:
Added the tests now. There is an update script in llvm/utils/, but it seems like this is not compatible with the test file anyway. I might have been a bit too specific with the naming of the variables. Let me know if I should use `TMP` predominantly.

test/Conversion/ImportVerilog/basic.sv
towoe: Thanks for pointing this out. At first I thought this might not be correct as my understanding of the 'concurrent' part was that they exists mostly next to the procedures, but you are right, IEEE 1800 16.14: "concurrent assertion statement ... in any ...: an always procedure or initial procedure as a statement".
Will keep your comment in mind for the future!

lib/Conversion/ImportVerilog/Statements.cpp
towoe: Somehow related to this I did not want to handle `SequenceWithMatchExpr` as I think this will also involve more IR ops.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  Value visit(const slang::ast::ClockingAssertionExpr &expr) {
+    auto assertionExpr = context.convertAssertionExpression(expr.expr, loc);
+    if (!assertionExpr)
+      return {};
+    auto result = context.convertLTLTimingControl(expr.clocking, assertionExpr);
+    if (!result)
+      return {};
+    return result;
```
towoe: Return the result directly.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,313 @@
...
+    // - Expression
+    // - Optional repetition
+
+    // The expression needs to have the type `i1` for the LTL operations.
+    // Convert first to a moore type and then add a conversion to `i1`.
+    auto value = context.convertRvalueExpression(expr.expr);
+    auto loc = context.convertLocation(expr.expr.sourceRange);
+    if (!value)
+      return {};
+    value = context.convertToI1(value);
```
towoe: `convertToI1` checks for an empty input`value` and we also check again after the call, so this check here might be unnecessary.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -0,0 +1,313 @@
...
+    if (!expr.repetition.has_value()) {
+      return value;
+    }
+
+    // There is a repetition, embed the expression into the kind of given
+    // repetition
+    Value repeatedValue = createRepetition(loc, expr.repetition.value(), value);
+    if (!repeatedValue)
+      return {};
+    return repeatedValue;
```
towoe: Return the value directly.

lib/Conversion/ImportVerilog/Statements.cpp
fabianschuiki: Yeah good idea!

fabianschuiki (approved):
This is very very cool, LGTM! Thanks for spearheading this :+1:. Do you have commit access, or do you want me to push merge?

towoe:
Thanks a lot @fabianschuiki for the support! I don't have commit access, so please push it :)

Issue #8530, opened 2025-06-03
Support linking multiple MLIR files in firtool


unlsycn:
We propose a mechanism for linking multiple MLIR files in firtool. Specifically, this involves combining multiple circuits into a single one.

All modules instantiated within another module (i.e., all `firrtl.instance` operations) must have corresponding declarations in the same circuit, which can be either a firrtl.module or a `firrtl.extmodule` serving as an interface. Notably, a module can have both an interface (as an `extmodule`) and an implementation (as a `module`) within the same circuit.

The linking process works as follows: First, we verify that each instance has a corresponding declaration. Then, all circuits are concatenated into one. Next, we iterate through all extmodules to find matching public module implementations, replacing them when possible. Finally, any remaining extmodules are treated as blackboxes.

seldridge:
There's a couple of other things that need to be considered here:

1. Mangling of private modules (private symbols). If the two circuits both have a `module Queue`, then these should be renamed, ideally deterministically (though I would not mandate this). The simple thing to do here is to use the circuit name as a prefix.
2. Layer-colored probes need to work. There are complicated ways to do this, but the simplest is to require have the layers of layer-colored probe ports on external modules use the exact same layers that are used in the circuit.

sequencer:
> Mangling of private modules (private symbols). If the two circuits both have a module Queue, then these should be renamed, ideally deterministically (though I would not mandate this). The simple thing to do here is to use the circuit name as a prefix.

I personally suggest just prefix the original circuit name to the private module Queue. However this also means we need to find all usage of Queue in the local circuit.

> Layer-colored probes need to work. There are complicated ways to do this, but the simplest is to require have the layers of layer-colored probe ports on external modules use the exact same layers that are used in the circuit.

I strongly suggest the simplest way, the fundamental issue here is that, layers are also a interface to a module, we should check it at link time.

sequencer:
At the first pass of linking(I'm not sure if it is a `Pass`) is mangling:
[...]

PR #8561, opened 2025-06-13
[firld] Add firld to link FIRRTL circuits

unlsycn:
Implement https://github.com/llvm/circt/issues/8530


lib/Dialect/FIRRTL/Transforms/LinkCircuits.cpp
```diff
@@ -0,0 +1,232 @@
...
+      module.getLoc(), StringAttr::get(&getContext(), baseCircuitName));
+  SmallVector<Attribute> mergedAnnotations;
+
+  for (auto circuit : circuits) {
+    if (!noMangle)
+      if (failed(mangleCircuitSymbols(circuit)))
+        return circuit->emitError("failed to mangle private symbol");
+
+    // TODO: other circuit attributes (such as enable_layers...)
+    append_range(mergedAnnotations, circuit.getAnnotations().getValue());
```
sequencer: Although enable_layers is from chisel time, I think think we should take care of this in the linker...

unlsycn:
I've added some tests for this change. Would someone mind reviewing when you have a moment? CC @sequencer @seldridge

PR #8562, opened 2025-06-13
[Seq] Add canonicalization for FirReg with mux-based constant drivers

uenoku:
Implement a canonicalization pattern for seq.firreg operations that optimizes registers with mux-based self-loops to constant values [exactly same way](https://github.com/llvm/circt/blob/5dc12802ef2b8264a00b70f236beae886db2c0da/lib/Dialect/FIRRTL/FIRRTLFolds.cpp#L3095) in FIRRTL dialect. 

The pattern matches registers where the next value is a mux with one branch being the register itself (creating a self-loop) and the other branch being a constant:
```mlir
  %mux = comb.mux %cond, %reg, %constant  // or (%constant, %reg)
  %reg = seq.firreg %mux clock %clk
```
In such cases, the register effectively holds a constant value and can be replaced with that constant, eliminating the register entirely.

The optimization is only applied when:
- One mux branch is the register output (self-loop)
- The other mux branch is a constant
- If the register has a reset value, it must be a constant matching the mux constant

### DIFF
--- a/lib/Dialect/Seq/SeqOps.cpp
+++ b/lib/Dialect/Seq/SeqOps.cpp
@@ -646,24 +646,26 @@ LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
       replacedValue = nextMux.getTrueValue();
     }
 
+    if (!replacedValue)
+      return failure();
+
     // Verify reset value compatibility: if register has reset, it must be
     // a constant that matches the mux constant
     if (op.getResetValue()) {
       Attribute resetConst;
       if (matchPattern(op.getResetValue(), m_Constant(&resetConst))) {
         if (resetConst != value)
-          replacedValue = {};
+          return failure();
       } else {
         // Non-constant reset value prevents optimization
-        replacedValue = {};
+        return failure();
       }
     }
 
+    assert(replacedValue);
     // Apply the optimization if all conditions are met
-    if (replacedValue) {
-      rewriter.replaceOp(op, replacedValue);
-      return success();
-    }
+    rewriter.replaceOp(op, replacedValue);
+    return success();
   }
 
   // For reset-less 1d array registers, replace an uninitialized element with


test/Dialect/Seq/canonicalization.mlir
```diff
@@ -307,3 +307,77 @@ hw.module @CompRegClockEnabled(in %clock: !seq.clock, in %d: i42, in %x: i42, in
...
+
+// CHECK-LABEL: @FirRegMuxConstantWithPreset
+hw.module @FirRegMuxConstantWithPreset(in %clk : !seq.clock, in %cond : i1) {
+  %c42_i32 = hw.constant 42 : i32
+  %c0_i32 = hw.constant 0 : i32
+
+  // Register with preset should not be optimized
+  %mux1 = comb.mux %cond, %reg1, %c42_i32 : i32
+  %reg1 = seq.firreg %mux1 clock %clk preset 42: i32
+  // CHECK: hw.instance "reg1" @Observe(x: %reg1: i32) -> ()
```
seldridge, line 348: it seems like we could optimize this to the constant given that the preset value is the same? This can be handled in a follow-on. However, because you wrote this explicitly, there may be a reason for _not_ doing this?

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -622,6 +622,50 @@ LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
...
+      replacedValue = nextMux.getTrueValue();
+    }
+
+    // Verify reset value compatibility: if register has reset, it must be
+    // a constant that matches the mux constant
+    if (op.getResetValue()) {
+      Attribute resetConst;
+      if (matchPattern(op.getResetValue(), m_Constant(&resetConst))) {
+        if (resetConst != value)
+          replacedValue = {};
```
seldridge: Consider direct returns to avoid the check on line 663:

```suggestion
          return failure();
```

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -622,6 +622,50 @@ LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
...
+      } else {
+        // Non-constant reset value prevents optimization
+        replacedValue = {};
```
seldridge:
```suggestion
        return failure();
```

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -622,6 +622,50 @@ LogicalResult FirRegOp::canonicalize(FirRegOp op, PatternRewriter &rewriter) {
...
+      }
+    }
+
+    // Apply the optimization if all conditions are met
+    if (replacedValue) {
+      rewriter.replaceOp(op, replacedValue);
+      return success();
+    }
```
seldridge: With the above changes to early exit:

```suggestion
    rewriter.replaceOp(op, replacedValue);
    return success();
```

test/Dialect/Seq/canonicalization.mlir
uenoku: Ah, I originally handled preset in the canonicalization but removed afterwards since I thought it's too complicated to do in a single PR. I'll do that in a follow up.

PR #8563, opened 2025-06-13
[ImportVerilog][CAPI] Add interfaces to allow external use

SpriteOvO:
This PR adds C-API for ImportVerilog stuff to allow external use.

### DIFF
--- a/include/circt-c/ImportVerilog.h
+++ b/include/circt-c/ImportVerilog.h
@@ -1,0 +1,84 @@
+//===-- circt-c/ImportVerilog.h - C API for importing Verilog -----*- C -*-===//
+//
+// This header declares the C interface for importing Verilog from Verilog
+// source files.
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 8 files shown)


uenoku (commented):
I wonder there is better direction. I understand we did the same thing for firtool but the maintainability has not been good.  Instead I'm curios we can use some string serialization (similar to what we are doing for ExportVerilog lowering option) or MLIR attribute as a way to pass option .

SpriteOvO:
@uenoku I do agree that the maintainability of such options structs is poor.

@sequencer If we change the options structs to something like a CLI arguments string, would that be acceptable to you guys?

sequencer:
> If we change the options structs to something like a CLI arguments string, would that be acceptable to you guys?

I think we only need to expose these options to C-API to allow external program to use import Verilog as library.

For C-API about `MlirImportVerilogOptions`, it can be safe and looks more like what https://github.com/llvm/circt/blob/firtool-1.122.0/include/circt-c/Firtool/Firtool.h does.

PR #8564, opened 2025-06-13
[Seq] Add a canonicalization to remove read-only memory

uenoku:
This commit implements FirMemOp canonicalization that eliminates dead memories:
- Read-only memories are replaced with constant 0

Previously only write-only memory was handled. Also it fixes an issue that write port was eliminated even when the memory has an inner symbol.

### DIFF
--- a/lib/Dialect/Seq/SeqOps.cpp
+++ b/lib/Dialect/Seq/SeqOps.cpp
@@ -874,9 +874,9 @@ LogicalResult FirMemOp::canonicalize(FirMemOp op, PatternRewriter &rewriter) {
     return success();
   }
 
-  if (readOnly) {
+  if (readOnly && !op.getInit()) {
     // Replace all read ports with a constant 0.
-    for (auto *user : op->getUsers()) {
+    for (auto *user : llvm::make_early_inc_range(op->getUsers())) {
       auto readOp = cast<FirMemReadOp>(user);
       Value zero = rewriter.create<hw::ConstantOp>(
           readOp.getLoc(), APInt::getZero(hw::getBitWidth(readOp.getType())));
--- a/lib/Dialect/Seq/SeqOps.cpp
+++ b/lib/Dialect/Seq/SeqOps.cpp
@@ -807,18 +807,42 @@ LogicalResult FirMemOp::canonicalize(FirMemOp op, PatternRewriter &rewriter) {
   if (op.getInnerSymAttr())
     return failure();
 
+  bool readOnly = true, writeOnly = true;
+
   // If the memory has no read ports, erase it.
   for (auto *user : op->getUsers()) {
-    if (isa<FirMemReadOp, FirMemReadWriteOp>(user))
-      return failure();
-    assert(isa<FirMemWriteOp>(user) && "invalid seq.firmem user");
+    if (isa<FirMemReadOp, FirMemReadWriteOp>(user)) {
+      writeOnly = false;
+    }
+    if (isa<FirMemWriteOp, FirMemReadWriteOp>(user)) {
+      readOnly = false;
+    }
+    assert((isa<FirMemReadOp, FirMemWriteOp, FirMemReadWriteOp>(user)) &&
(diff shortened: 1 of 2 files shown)


lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -853,18 +853,42 @@ LogicalResult FirMemOp::canonicalize(FirMemOp op, PatternRewriter &rewriter) {
...
 
-  for (auto *user : llvm::make_early_inc_range(op->getUsers()))
-    rewriter.eraseOp(user);
+    rewriter.eraseOp(op);
+    return success();
+  }
 
-  rewriter.eraseOp(op);
-  return success();
+  if (readOnly) {
```
fzi-hielscher:
```suggestion
  if (readOnly && !op.getInit()) {
```

We're frequently using FirMem ROMs with initialization files in our designs. I'd prefer them not to be removed. 😉

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -853,18 +853,42 @@ LogicalResult FirMemOp::canonicalize(FirMemOp op, PatternRewriter &rewriter) {
...
+    // Replace all read ports with a constant 0.
+    for (auto *user : op->getUsers()) {
```
fzi-hielscher:
```suggestion
    for (auto *user : llvm::make_early_inc_range(op->getUsers())) {
```

lib/Dialect/Seq/SeqOps.cpp
uenoku: Totally! Thank you for pointing it out!

test/Dialect/Seq/canonicalization.mlir
```diff
@@ -228,8 +228,11 @@ hw.module @FirMem(in %addr : i4, in %clock : !seq.clock, in %data : i42, out out
   // CHECK-NEXT: seq.firmem.read_port [[MEM]][%addr], clock [[CLK_FALSE]] {rw6}
   %8 = seq.firmem.read_write_port %0[%addr] = %data if %true, clock %clk_false {rw6} : <12 x 42, mask 3>
 
-  // CHECK: seq.firmem
+  // CHECK: %has_symbol = seq.firmem
   %has_symbol = seq.firmem sym @someMem 0, 1, undefined, undefined : <12 x 42, mask 3>
+  // CHECK-NEXT: seq.firmem.write_port %has_symbol
+  seq.firmem.write_port %has_symbol[%addr] = %data, clock %clock enable %false {w0} : <12 x 42, mask 3>
+
```
fabianschuiki, line 236: Nice catch!

lib/Dialect/Seq/SeqOps.cpp
```diff
@@ -853,18 +853,42 @@ LogicalResult FirMemOp::canonicalize(FirMemOp op, PatternRewriter &rewriter) {
...
-    rewriter.eraseOp(user);
+    rewriter.eraseOp(op);
+    return success();
+  }
 
-  rewriter.eraseOp(op);
-  return success();
+  if (readOnly && !op.getInit()) {
+    // Replace all read ports with a constant 0.
+    for (auto *user : llvm::make_early_inc_range(op->getUsers())) {
```
fabianschuiki, line 879: Does returning a zero here match FIRRTL semantics? Or does this need some kind of X constant?

uenoku: Yeah good catch. [FIRRTL folds](https://github.com/llvm/circt/blob/main/lib/Dialect/FIRRTL/FIRRTLFolds.cpp#L2397-L2399) replace it with uninitialized register which will be [canonicalized to zero](https://github.com/llvm/circt/blob/6a85b3b8f4b58451d727b675213c8f4f7e361132/lib/Dialect/Seq/SeqOps.cpp#L779).  So I think canonicalizing it to 0 matches FIRRTL semantics

Issue #8560, opened 2025-06-11
Probed constants moved out of layerblocks


rwy7:
The following simple example has bad output. A node, defined and probed under a layerblock, is lifted out of the layerblock and into the body of the module:.

```firrtl
FIRRTL version 5.0.0

circuit Example:
  layer Verification, bind:

  public module Example:
    output p: Probe<UInt<1>, Verification>
    layerblock Verification:
      node n = UInt<1>(0)
      define p = probe(n)
```

Goes to:

```verilog
// Generated by CIRCT unknown git version
module Example_Verification();
endmodule

module Example();
  wire _GEN = 1'h0;
endmodule


// ----- 8< ----- FILE "ref_Example.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`define ref_Example_p _GEN

// ----- 8< ----- FILE "layers-Example-Verification.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`ifndef layers_Example_Verification
  `define layers_Example_Verification
  bind Example Example_Verification verification ();
`endif // not def layers_Example_Verificatio
```

If I "donttouch" the node, we get much better output:
```firrtl
FIRRTL version 5.0.0

circuit Example: %[[
  {
    "class": "firrtl.transforms.DontTouchAnnotation",
    "target": "~Example|Example>n"
  }
]]
  layer Verification, bind:

  public module Example:
    output p: Probe<UInt<1>, Verification>
    layerblock Verification:
      node n = UInt<1>(0)
      define p = probe(n)
```

Goes to:
```verilog
// Generated by CIRCT unknown git version
module Example_Verification();
  wire n = 1'h0;
  wire n_probe = n;
endmodule

module Example();
endmodule


// ----- 8< ----- FILE "ref_Example.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`define ref_Example_p verification.n_probe

// ----- 8< ----- FILE "layers-Example-Verification.sv" ----- 8< -----

// Generated by CIRCT unknown git version
`ifndef layers_Example_Verification
  `define layers_Example_Verification
  bind Example Example_Verification verification ();
`endif // not def layers_Example_Verification
```

The cause of this seems to be that the initial node, located under the layerblock, is canonicalized away, and references to the node are replaced with references to the constant. Due to constant-pooling, the constant is located in the front of the body module. When the probes are lowered to XMRs, a new node is created for the XMR, which is co-located with the constant in the module body.

[...]

PR #8570, opened 2025-06-16
[FIRRTL] Make XMRRef and Deref pure

rwy7:
- Make XMRRefOp and XMRDerefOp pure
- Add descriptions to these ops
- Implement the SymbolUserOpInterface for these ops.
- Add some tests for advanced layer sink

Making the XMRRefOp and XMRDerefOp **pure** allows advanced layer sink to sink these operations into layers, which helps us clean up FIRRTL designs which have unguarded reads of probes.

This PR also add the SymbolUserOpInterface to the XMR ops. Now that we are verifying symbol references, this PR had to fix up a few layer related tests. The verification code just checks that the target exists, and is a hier path op.

This PR also adds a description to the XMR ops.

This PR also adds a test for advanced-layer-sink that checks if XMRs are sunk, and also an end-to-end test which, before this PR, did not sink probes into layers correctly.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLOps.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOps.cpp
@@ -6269,6 +6269,30 @@ LogicalResult RWProbeOp::verifyInnerRefs(hw::InnerRefNamespace &ns) {
   return checks(symOp.getTargetResult().getType(), symOp.getLoc());
 }
 
+LogicalResult XMRRefOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  auto *target = symbolTable.lookupNearestSymbolFrom(*this, getRefAttr());
+  if (!target)
+    return emitOpError("has an invalid symbol reference");
+
+  if (!isa<hw::HierPathOp>(target))
+    return emitOpError("does not target a hierpath op");
+
+  // TODO: Verify that the target's type matches the type of this op.
+  return success();
+}
+
+LogicalResult XMRDerefOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  auto *target = symbolTable.lookupNearestSymbolFrom(*this, getRefAttr());
+  if (!target)
+    return emitOpError("has an invalid symbol reference");
+
+  if (!isa<hw::HierPathOp>(target))
+    return emitOpError("does not target a hierpath op");
+
+  // TODO: Verify that the target's type matches the type of this op.
+  return success();
+}
+
 //===----------------------------------------------------------------------===//
 // Layer Block Operations
 //===----------------------------------------------------------------------===//
--- a/test/Dialect/FIRRTL/advanced-layer-sink.mlir
+++ b/test/Dialect/FIRRTL/advanced-layer-sink.mlir
@@ -769,3 +769,23 @@ firrtl.circuit "DoNotSinkInstanceOfModuleWithPortAnno" {
     }
   }
 }
+
+// CHECK-LABEL: firrtl.circuit "SinkXMRs"
+firrtl.circuit "SinkXMRs" {
+  firrtl.layer @A bind {}
+
+  hw.hierpath @xmr [@SinkXMRs::@target]
+
+  firrtl.module public @SinkXMRs() {
+    %target = firrtl.wire sym @target : !firrtl.uint<1>
+    %0 = firrtl.xmr.deref @xmr : !firrtl.uint<1>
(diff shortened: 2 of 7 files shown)


uenoku (approved):
LGTM, could you add a test for canonicalization.mlir to check dead XMR*Ops are deleted?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6269,6 +6269,24 @@ LogicalResult RWProbeOp::verifyInnerRefs(hw::InnerRefNamespace &ns) {
...
+    return emitOpError("invalid symbol reference");
+
+  // TODO: Verify that the target's type matches the type of this op.
+  return success();
+}
+
+LogicalResult XMRDerefOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  auto *target = symbolTable.lookupNearestSymbolFrom(*this, getRefAttr());
+  if (!target)
+    return emitOpError("invalid symbol reference");
```
dtzSiFive: Your idea, but yeah maybe good to check these resolve to HierPathOp's 👍 .  And some quick error tests in `errors.mlir` (again your idea).

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -310,13 +311,13 @@ firrtl.circuit "Test" {
   // XMR Ref ops used by force_initial are cloned.
   //
   // CHECK:      firrtl.module private @XmrRef_A()
-  // CHECK-NEXT:   %0 = firrtl.xmr.ref @RefXmrRef_path : !firrtl.rwprobe<uint<1>, @A>
+  // CHECK-NEXT:   %0 = firrtl.xmr.ref @XmrRef_path : !firrtl.rwprobe<uint<1>, @A>
```
dtzSiFive, line 314: Wow, wrong symbol reference! Thanks for fixing!

test/firtool/sink-probes-into-layers.fir
```diff
@@ -0,0 +1,40 @@
...
+FIRRTL version 4.0.0
+
+circuit Foo:
+  layer A, bind:
+
+  public module Foo:
+    input clock: Clock
+
+    inst bar of Bar
+    node cond = read(bar.probe)
```
dtzSiFive, line 37: Thanks for the great end-to-end test, especially given the pipeline-sensitive nature of the interplay between layers and probes (mostly historical, but we may yet want to shuffle them a little).

dtzSiFive: Maybe including a rwprobe (+force inside the layerblock) to ensure the rwprobe is sunk too? It's not the same since it's not "dereferenced" outside the layer like the bare `read(..)` here, so up to you.

dtzSiFive (approved):
Thanks for fixing this!

These operations only exist temporarily between LowerXMR until LowerToHW, to help with preserving all the needed FIRRTL types while letting LowerXMR lower its. portion of the IR (replacing probe operands with resolve hierarchical path to the resolved target).  They weren't given much attention, this was likely wrong from inception!

Thanks for noticing and fixing! 🎉 💯

dtzSiFive:
Does this fix https://github.com/llvm/circt/issues/8560 ?

rwy7:
> Does this fix https://github.com/llvm/circt/issues/8560 ?

Unfortunately not; the situation there could be improved by running lower-layers before lower-xmr, or maybe there is something else we can do.

test/Dialect/FIRRTL/advanced-layer-sink.mlir
```diff
@@ -769,3 +769,23 @@ firrtl.circuit "DoNotSinkInstanceOfModuleWithPortAnno" {
...
+  firrtl.module public @SinkXMRs() {
+    %target = firrtl.wire sym @target : !firrtl.uint<1>
+    %0 = firrtl.xmr.deref @xmr : !firrtl.uint<1>
+    %1 = firrtl.xmr.ref @xmr : !firrtl.ref<uint<1>>
+  
+    // CHECK: firrtl.layerblock @A
+    firrtl.layerblock @A {
+      // CHECK-NEXT: %0 = firrtl.xmr.deref @xmr : !firrtl.uint<1>
+      // CHECK-NEXT: %1 = firrtl.xmr.ref @xmr : !firrtl.probe<uint<1>>
+      "unknown"(%0, %1) : (!firrtl.uint<1>, !firrtl.ref<uint<1>>) -> ()
```
seldridge, line 788: Nice test and nice use of an unknown operation.

Issue #8548, opened 2025-06-09
[circt-bmc] Functionality to turn off assertion failure for specific clock period


uenoku:
When verifying following IR with circt-bmc (`circt-bmc  -b 10 --module Compare`) it fails because in the first clock register values are not stabilized. After the first clock, it must be same (actually if I set appropriate initial values to these registers it was successfully verified). It would be fairly useful for me if bmc has an option to turn off assertions for specific period of clock. Alternative solution would be an option to turn off assertion until first reset. 

```mlir
hw.module @before(in %clock: !seq.clock, in %reset: i1, in %a: i1, in %b: i1, in %c: i1, out out: i1) {
  %0 = comb.and %a, %b : i1
  %1 = comb.and %0, %c : i1
  %false = hw.constant false
  %pipe1 = seq.compreg %1, %clock : i1
  hw.output %pipe1 : i1
}

hw.module @after(in %clock : !seq.clock, in %reset : i1, in %a : i1, in %b : i1, in %c : i1, out out : i1) {
  %false = hw.constant false
  %0 = comb.and %a, %b : i1
  %retimed = seq.compreg %0, %clock : i1
  %retimed_c = seq.compreg %c, %clock : i1

  %1 = comb.and %retimed, %retimed_c : i1
  hw.output %1 : i1
}

hw.module @Compare(in %clk: !seq.clock, in %reset: i1, in %a: i1, in %b: i1, in %c: i1) {
  %before_out = hw.instance "before" @before(clock: %clk: !seq.clock, reset: %reset: i1, a: %a: i1, b: %b: i1, c: %c: i1) -> (out: i1) 
  %after_out = hw.instance "after" @after(clock: %clk: !seq.clock, reset: %reset: i1, a: %a: i1, b: %b: i1, c: %c: i1) -> (out: i1)
  %2 = comb.icmp eq %before_out, %after_out : i1
  verif.assert %2 : i1
}
```

TaoBi22:
This shouldn't be difficult, I'll try and get it done at some point soon!

TaoBi22:
Closed in #8573

PR #8573, opened 2025-06-17
[circt-bmc] Add feature to ignore asserts on some initial cycles

TaoBi22:
This adds a new feature to circt-bmc where assertions can be ignored for the first n cycles, as requested in #8548


uenoku (approved):
LGTM, thank you for adding this so quickly! I checked an example in https://github.com/llvm/circt/issues/8548 and it worked well!

Issue #8476, opened 2025-05-12
[MooreToCore] Lower exponentiation to `math.ipowi`


fabianschuiki:
@jiahanxie353 has added the Math dialect to circt-opt in #8475. This is a great precedent to start using that dialect in the MooreToCore conversion. For example, the `moore.pow` operation could be lowered to `math.ipowi` and `math.powf` depending on its type.

liamslj13:
Hi Fabian, I am a newcomer here and looking to contribute to this project. Admittedly, I am a bit lost navigating the codebase and wondering if you have any guidance on where to start working towards clearing this issue.

I have tried looking through the MooreToCore.h file and lib/Dialect/Moore directory, but maybe I am not looking in the correct places. After reviewing the linked PR it is still a bit fuzzy to me.

fabianschuiki:
Hey @liamslj13, great to have you on board! And don't worry, the codebase definitely takes a while to get used to. Once you've seen and edited a few files in different corners it becomes a lot easier 😃. The `MooreToCore` pass is a conversion, which lives in `Conversion/MooreToCore` instead of the usual dialect passes like `Dialect/Moore/Transforms`:

https://github.com/llvm/circt/blob/main/lib/Conversion/MooreToCore/MooreToCore.cpp

This file contains all the conversion patterns from the Moore dialect to HW/Comb/Seq. I think there are even existing ones for unsigned and signed "pow" operations in the Moore dialect. It would be really nice if those could call into `math.ipowi` instead 😃

liamslj13:
Thank you for the prompt response and the kind explanation! 🙇
I am looking at it now and trying to grasp what is going on the code. I will probably need to continue tomorrow morning. 
In the meantime is it possible to get assigned to this issue? I may come back with some questions later though.

liamslj13:
Hi again, I've been looking through the file that we need to change and would just like to clarify a few things that I am still a bit confused about.
```cpp
struct PowUOpConversion : public OpConversionPattern<PowUOp> {
  using OpConversionPattern::OpConversionPattern;

  LogicalResult
  matchAndRewrite(PowUOp op, OpAdaptor adaptor,
                  ConversionPatternRewriter &rewriter) const override {
    Type resultType = typeConverter->convertType(op.getResult().getType());

    Location loc = op.getLoc();
    auto intType = cast<IntType>(op.getRhs().getType());

    // transform a ** b into scf.for 0 to b step 1 { init *= a }, init = 1
[...]

PR #8574, opened 2025-06-17
[MooreToCore] Lower exponentiation to math.ipowi (PowSOpConversion)

liamslj13:
This pull request addresses part of #8476. I have removed the for-loop logic from the signed integer conversion and replaced it directly with `math.ipowi` from MLIR's Math dialect, which handles the exponentiation logic for us. I also included the `mlir.math` dialect header in the MooreToCore.cpp conversion file.

I have some concerns, however this might be related to my lack of experience with the codebase (and admittedly MLIR as a whole).
- The code builds, however the tests (test/Conversion/MooreToCore/basic.mlir) fail. This may be due to a type discrepancy in what the original logic returns and what `math.ipowi` returns.
-  If there is a type discrepancy, I am unsure if I was supposed to maybe convert the result back to the original type, and also not sure of how to change the tests to reflect the changes I have made here.

Any feedback helps, thank you for your time!

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -23,6 +23,7 @@
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Math/IR/Math.h"
 #include "mlir/IR/BuiltinDialect.h"
 #include "mlir/IR/Iterators.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -22,8 +22,8 @@
 #include "mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
-#include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Math/IR/Math.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/BuiltinDialect.h"
 #include "mlir/IR/Iterators.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
@@ -1359,8 +1359,10 @@ struct PowSOpConversion : public OpConversionPattern<PowSOp> {
     Type resultType = typeConverter->convertType(op.getResult().getType());
 
     Location loc = op.getLoc();
-    // utilize MLIR math dialect's math.ipowi to handle the exponentiation of expression
-    Value result = rewriter.create<mlir::math::IPowIOp>(loc, adaptor.getLhs(), adaptor.getRhs());
+    // utilize MLIR math dialect's math.ipowi to handle the exponentiation of
+    // expression
+    Value result = rewriter.create<mlir::math::IPowIOp>(loc, adaptor.getLhs(),
+                                                        adaptor.getRhs());
 
     rewriter.replaceOp(op, result);
 
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -1358,35 +1358,10 @@ struct PowSOpConversion : public OpConversionPattern<PowSOp> {
     Type resultType = typeConverter->convertType(op.getResult().getType());
 
     Location loc = op.getLoc();
-    auto intType = cast<IntType>(op.getRhs().getType());
-    // transform a ** b into scf.for 0 to b step 1 { init *= a }, init = 1
-    Type integerType = rewriter.getIntegerType(intType.getWidth());
-    Value lhsVal = rewriter.create<ConversionOp>(loc, resultType, op.getLhs());
-    Value rhsVal = rewriter.create<ConversionOp>(loc, integerType, op.getRhs());
-    Value constZero = rewriter.create<hw::ConstantOp>(loc, integerType, 0);
-    Value constZeroResult = rewriter.create<hw::ConstantOp>(loc, resultType, 0);
-    Value isNegative = rewriter.create<comb::ICmpOp>(loc, ICmpPredicate::slt,
-                                                     rhsVal, constZero);
+    // utilize MLIR math dialect's math.ipowi to handle the exponentiation of expression
+    Value result = rewriter.create<mlir::math::IPowIOp>(loc, adaptor.getLhs(), adaptor.getRhs());
 
-    // if the exponent is negative, return 0
-    lhsVal =
-        rewriter.create<comb::MuxOp>(loc, isNegative, constZeroResult, lhsVal);
-    Value upperBound =
-        rewriter.create<comb::MuxOp>(loc, isNegative, constZero, rhsVal);
-
-    Value lowerBound = constZero;
-    Value step = rewriter.create<hw::ConstantOp>(loc, integerType, 1);
-    Value initVal = rewriter.create<hw::ConstantOp>(loc, resultType, 1);
-
-    auto forOp = rewriter.create<scf::ForOp>(
-        loc, lowerBound, upperBound, step, ValueRange(initVal),
-        [&](OpBuilder &builder, Location loc, Value i, ValueRange iterArgs) {
-          auto loopVar = iterArgs.front();
-          auto mul = rewriter.create<comb::MulOp>(loc, lhsVal, loopVar);
-          rewriter.create<scf::YieldOp>(loc, ValueRange(mul));
-        });
-
-    rewriter.replaceOp(op, forOp.getResult(0));
+    rewriter.replaceOp(op, result);
 
     return success();
   }
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -1361,13 +1361,8 @@ struct PowSOpConversion : public OpConversionPattern<PowSOp> {
     Location loc = op.getLoc();
     // utilize MLIR math dialect's math.ipowi to handle the exponentiation of
     // expression
-    Value result = rewriter.create<mlir::math::IPowIOp>(loc, adaptor.getLhs(),
-                                                        adaptor.getRhs());
-    // cast the result back to MooreToCore i32 type; result is originally
-    // math.ipowi regular i32 so we need to revert it back
-    Value castedResult = rewriter.create<ConversionOp>(loc, resultType, result);
-
-    rewriter.replaceOp(op, castedResult);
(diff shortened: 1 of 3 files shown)


fabianschuiki:
Hey @liamslj13, thanks for working on this! Looks like the test fails because the Math dialect was never loaded:
```
LLVM ERROR: Building op `math.ipowi` but it isn't known in this MLIRContext: the dialect may
not be loaded or this operation hasn't been added by the dialect. See also
https://mlir.llvm.org/getting_started/Faq/#registered-loaded-dependent-whats-up-with-dialects-management
```
This usually happens if the dialect is not listed in `dependentDialects` of the pass. Take a look at the TableGen definition of the MooreToCore pass: https://github.com/llvm/circt/blob/main/include/circt/Conversion/Passes.td#L524-L539. You should be able to add the `mlir::math::MathDialect` to that list and get the test to pass that way.

The other CI failure is due to code formatting. You can usually fix this by running clang-format locally on the file you've edited. Most text editors support that out of the box. CI will usually also upload a diff showing you what needed reformatting, but it doesn't seem to have worked here :thinking:

liamslj13:
Thank you for pointing these things out! I probably should have tried harder to look into the error messages instead of panicking right away, I will try to do better next time. 

I have added the changes and will test the build/tests on my device here too. Hopefully all goes well. Will keep an eye on whether I need to reformat anything again.

fabianschuiki:
Formatting looks great 👍. Compilation also seems to work. The tests fail on the `moore.pows` op, which isn't surprising since that's the thing you're tweaking:
```
/__w/circt/circt/test/Conversion/MooreToCore/basic.mlir:1130:8: error: unexpected error: failed to legalize operation 'moore.pows'
  %0 = moore.pows %arg0, %arg1 : i32
       ^
```
To debug this, I would recommend copying the surrounding `func.func` into a new MLIR file and then running `circt-opt --convert-moore-to-core YourFile.mlir --debug`. The `--debug` option should spit out a ton of additional details. Most importantly, the dialect conversion framework of MLIR prints a lot of messages detailing why some operation could not be legalized. It's usually some op verifier failing, a type mismatch, or something similar. That could help you debug the issue.

liamslj13:
Thank you for the feedback again 🙇‍♂️ It is a little bit late so I will work on the adjustments that you have suggested tomorrow. If I think that I may have fixed the test, is it okay to edit the `basic.mlir` file directly? I don't want to touch/break anything I am not supposed to, and I feel that this might require me to change the `moore.pows` type to something else in the test due to `result` being dependant on the Math dialect instead. Does this sound alright?

fabianschuiki:
Yeah feel free to edit any files that are necessary! If you need to tweak `basic.mlir` that might indicate that the conversion pattern doesn't work fully, but modifying `basic.mlir` can be a good stepping stone to find some IR where the lowering works, and then work backwards and figure out why the original one didn't work 😄

liamslj13:
I've been iterating through and getting a feel how things work for a bit of the day. My current change in the test has landed on 
```mlir
// CHECK-LABEL: func.func @PowSOp
func.func @PowSOp(%arg0: !moore.i32, %arg1: !moore.i32) {
  // CHECK: %[[RES:.*]] = math.ipowi %arg0, %arg1 : i32
  %0 = moore.pows %arg0, %arg1 : i32
  return
}
```
and I have also adjusted the function (not pushed to my branch yet) to convert the result back to the `moore.i32`type, and also learned that I need to register `mlir::math` as a legal dialect for the conversion pass. This test I have shown above and a few others I have tried will work in isolation like you suggested to do, but when tested in `basic.mlir` they fail instantly and give an error like this
```
/Users/liam/circt/test/Conversion/MooreToCore/basic.mlir:1146:12: error: CHECK: expected string not found in input
 // CHECK: ^bb1: // 2 preds: ^bb0, ^bb5
           ^
<stdin>:710:14: note: scanning from here
 llhd.process {
             ^
<stdin>:712:2: note: possible intended match here
 ^bb1: // 2 preds: ^bb0, ^bb2
 ^

Input file: <stdin>
Check file: /Users/liam/circt/test/Conversion/MooreToCore/basic.mlir

-dump-input=help explains the following input dump.

Input was:
<<<<<<
              .
              .
              .
            705:  } 
            706:  hw.module @blockArgAsObservedValue(in %in0 : i32, in %in1 : i32) { 
            707:  %c0_i32 = hw.constant 0 : i32 
            708:  %var = llhd.sig %c0_i32 : i32 
            709:  %3 = llhd.prb %var : !hw.inout<i32> 
            710:  llhd.process { 
check:1146'0                  X~~ error: no match found
            711:  cf.br ^bb1 
check:1146'0     ~~~~~~~~~~~~
            712:  ^bb1: // 2 preds: ^bb0, ^bb2 
check:1146'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
check:1146'1      ?                             possible intended match
            713:  %4 = math.ipowi %in0, %in1 : i32 
check:1146'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            714:  %5 = llhd.constant_time <0ns, 0d, 1e> 
check:1146'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            715:  llhd.drv %var, %4 after %5 : !hw.inout<i32> 
check:1146'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            716:  cf.br ^bb2 
check:1146'0     ~~~~~~~~~~~~
            717:  ^bb2: // pred: ^bb1 
check:1146'0     ~~~~~~~~~~~~~~~~~~~~~
              .
              .
              .
>>>>>>
```
I am a bit lost with this error, and I am struggling to find information on how the predecessors here work in the tests.

fabianschuiki:
That error message comes from the following FileCheck line in your `basic.mlir` test file:
```
// CHECK: ^bb1: // 2 preds: ^bb0, ^bb5
```
It also shows you the line it _did_ find in the output, but which doesn't match exactly:
```
<stdin>:712:2: note: possible intended match here
 ^bb1: // 2 preds: ^bb0, ^bb2
 ^
```
So basically this happens:
```
^bb1: // 2 preds: ^bb0, ^bb5  <--- what the CHECK lines is looking for
^bb1: // 2 preds: ^bb0, ^bb2  <--- what is actually in the output
```
To make tests more robust under minor renaming of ops in the IR, you'd usually only match the parts of the output that are relevant, in your case `// CHECK: ^bb1` without the `: // 2 preds: ^bb0, ^bb5`. And you'd use `[[XYZ:%.+]]` or similar patterns to match any name -- which you're already doing anyway :smiley:.

Feel free to push all your changes and tweaks to `basic.mlir` as well -- that makes it easier to help debug stuff :smiley:

liamslj13:
Ah okay that makes a bit more sense then. I should maybe try and find the actual output and take a look at it myself instead of just leaving it all to the test until I get it sorted out. I probably should have pushed my new changes originally, I will do it now.

liamslj13:
I've given this a few looks over the week and I still can't figure out how to write a workaround for the tests. How important is it that the preds match exactly? Could it be fixed if I just check for any two?

liamslj13:
I tried it out on my build and the tests finally passed (for the first time in a while)! Hopefully it all goes smoothly this time. I will need to look more into MLIR so I can be a bit more efficient next time.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1358,36 +1359,15 @@ struct PowSOpConversion : public OpConversionPattern<PowSOp> {
...
-
+    // utilize MLIR math dialect's math.ipowi to handle the exponentiation of
+    // expression
+    Value result = rewriter.create<mlir::math::IPowIOp>(loc, adaptor.getLhs(),
+                                                        adaptor.getRhs());
+    // cast the result back to MooreToCore i32 type; result is originally
+    // math.ipowi regular i32 so we need to revert it back
+    Value castedResult = rewriter.create<ConversionOp>(loc, resultType, result);
+
+    rewriter.replaceOp(op, castedResult);
```
fabianschuiki: You might be able to do this without the conversion op, in one go (I stole this from the `ShrUOp` case above):
```suggestion
    // utilize MLIR math dialect's math.ipowi to handle the exponentiation of
    // expression
    rewriter.replaceOpWithNewOp<mlir::math::IPowIOp>(
        op, resultType, adaptor.getLhs(), adaptor.getRhs());
```
Not 100% sure this will work, but might be worth a try. Most other ops didn't have to do create a `ConversionOp`, so `math.ipowi` might also be able to get away without :smiley:

fabianschuiki (commented):
Very cool! Thanks for working through this so diligently, that's very much appreciated. I promise it gets less annoying over time :smirk:

liamslj13:
Thank you for giving me the opportunity to work on this issue! Apologies if I was slow, I tend to get lost pretty easily with new topics. I will hopefully be trying my hand at some other issues as well!

As for the change you suggested, I tried it locally and everything worked fine with regression tests so I have committed that too. This might be an unneeded nitpick or completely wrong, but I'm not sure we need the line `Location loc = op.getLoc();` anymore as the variable isn't used. It could just ensure there is no warning during build and make the function cleaner. Unless we use it later of course!

fabianschuiki:
Excellent point, the `loc` would no longer be needed :smiley: :rocket:. Thanks for getting this all the way through! Let me know if you want to tweak that `loc` line before landing this. Otherwise I'm happy to merge this for you :+1:

liamslj13:
Ah I might change it really quick before we merge then. It feels a bit weird to me that I addressed it but never fixed it. Sorry for all of the setbacks on this tiny change lol. Thank you for your patience as well.

liamslj13:
I made a mistake trying to add back the variable because it wouldn't pass the clang-format, but didn't change anything when I re-ran clang-format locally. Somehow I pushed the wrong thing. 

Edit 1:
Is it possible to merge just the approved changes or should I figure out how to get this back on track?

Edit 2:
I figured out how to revert this back to how it was when initially approved. If this works out maybe it's best to just merge it as is. I feel I have made some really weird mistakes while working on this change.

fabianschuiki:
Hey @liamslj13, thanks a lot for all the tweaks and updates :partying_face: :100:! I'll squash and merge this as it is right now, since CI passes and it implements the nice lowering to `math.ipowi` :smiley:.

If you feel like tackling more related things, there's also the `PowUOp` lowering pattern that would probably benefit from using `math.ipowi`. That one is interesting because `moore.powu` is an unsigned op, while `math.ipowi` is a signed operation. You might have to zero-extend the inputs with a single zero bit, to ensure they are not treated as a negative number, and then chop the top bit off the result of `ipowi` again.

liamslj13:
Hi Fabian, thank you for the support along the way! This is my first actual contribution to any open source, so I very much appreciate the patience and encouragement. It was very enjoyable to learn about the codebase. 

As for the conversion of the unsigned operation, I would love to give it a go and I actually made a branch for this change back when I started working on the signed conversion. I never got around to starting it though. Should I create an issue for this, or are we okay to keep the original issue as it mentions both?

fabianschuiki:
I'd keep the original issue as it covers both as you say. Feel free to open a PR (or draft PR) as soon as you have a bit of code :slightly_smiling_face:. The GitHub PRs work best if you have concrete code, even if it's broken, to talk about :+1:

liamslj13:
Sounds great then! As soon as I get some code down I will do as you suggested 🙌

PR #8575, opened 2025-06-17
[Arcilator] Don't try to run JIT if only part of the pipeline is run

TaoBi22:
Currently if you run, for example, `arcilator --run --until-before=state-lowering`, Arcilator still tries to JIT the file and obviously doesn't know how without the full pipeline running. This just adds an error that throws if both JIT and a non--end until-before/after are specified.


fabianschuiki (approved):
Excellent idea, thanks! There's an inadvertent LLVM bump in the commit -- once that's gone CI should go through :smiley:

TaoBi22:
> Excellent idea, thanks! There's an inadvertent LLVM bump in the commit -- once that's gone CI should go through 😃

Oops :facepalm: thanks for spotting that!

PR #8576, opened 2025-06-17
[FIRRTL] AdvancedLayerSink: clone ref ops into layerblocks

rwy7:
- Run advanced layer sink (ALS) before lower-XMR
  - Lower-XMR will place symbols on probed objects, which will prevent ALS from sinking these objects into layers
  - By moving ALS before lower-XMR, ALS is more able to sink objects into layers.
  - With this change, ALS will encounter ref ops, not XMR ops.
- In ALS, clone ref ops into layerblocks
  - Ensures that we will sink ref resolve ops as deep as possible, even if there exists no LCA layerblock.
  - Add ALS test for this behaviour
- Add end-to-end test checking that a probed node does not result in "residue" in the module body.

Fixes: https://github.com/llvm/circt/issues/8560

### DIFF
--- a/test/firtool/lower-layers-clone-refs.fir
+++ b/test/firtool/lower-layers-clone-refs.fir
@@ -1,0 +1,30 @@
+; RUN: firtool --disable-all-randomization --advanced-layer-sink %s | FileCheck %s
+
+; Check that there is no probe wire in the output example.
+
+; CHECK:      module Example_A();
+; CHECK-NEXT:   wire _GEN = 1'h0;
+; CHECK-NEXT: endmodule
+
+; CHECK:      module Example_B();
+; CHECK-NEXT:   wire _GEN = 1'h0;
+; CHECK-NEXT: endmodule
+
+; CHECK:      module Example();
+; CHECK-NEXT: endmodule
+
+FIRRTL version 5.0.0
+
+circuit Example:
+  layer A, bind:
+  layer B, bind:
+
+  public module Example:
+    output pa: Probe<UInt<1>, A>
+    output pb: Probe<UInt<1>, B>
+
+    node n = UInt<1>(0)
+    layerblock A:
+      define pa = probe(n)
+    layerblock B:
+      define pb = probe(n)


include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
```diff
@@ -1489,8 +1489,8 @@ def RefCastOp : FIRRTLOp<"ref.cast",
      "$input attr-dict `:` functional-type($input, $result)";
 }
 
-def RefResolveOp: FIRRTLExprOp<"ref.resolve",
-                              [RefTypeConstraint<"ref","result">]> {
+def RefResolveOp
+    : FIRRTLExprOp<"ref.resolve", [Pure, RefTypeConstraint<"ref", "result">]> {
```
dtzSiFive: FIRRTLExprOp adds a base set of traits to those specified here.

HasCustomSSAName
InferTypeOpInterface
Pure

---> these should already be pure! (phew 👍 )

test/Dialect/FIRRTL/canonicalization.mlir
```diff
@@ -3648,4 +3648,21 @@ firrtl.module private @XMRTest() {
...
+// CHECK-LABEL: firrtl.module private @RefTest
+firrtl.module private @RefTest() {
+  %target = firrtl.wire : !firrtl.bundle<a: uint<1>>
+
+  // CHECK-NOT: firrtl.ref.send
+  %0 = firrtl.ref.send %target : !firrtl.bundle<a: uint<1>>
+
+  // CHECK-NOT: firrtl.ref.resolve
+  %1 = firrtl.ref.resolve %0 : !firrtl.probe<bundle<a: uint<1>>>
+
```
dtzSiFive: (This test passes without this PR)

Nice to have the test though, don't see it directly tested at least in a quick bounce through our tests, maybe separate PR or just push to main? (WDYT?)

dtzSiFive (approved):
* Moving LayerSink portion of pipeline immediately before LowerXMR: ✅ 
* Cloning ref.{send,resolve,sub,cast}: ✅ 

Nice work! These operations are already 'pure', see comments.

include/circt/Dialect/FIRRTL/FIRRTLExpressions.td
rwy7: Derp thanks!

test/Dialect/FIRRTL/canonicalization.mlir
rwy7: https://github.com/llvm/circt/pull/8577

test/firtool/lower-layers-clone-refs.fir
```diff
@@ -0,0 +1,30 @@
...
+
+  public module Example:
+    output pa: Probe<UInt<1>, A>
+    output pb: Probe<UInt<1>, B>
+
+    node n = UInt<1>(0)
+    layerblock A:
+      define pa = probe(n)
+    layerblock B:
+      define pb = probe(n)
```
rwy7, line 30: This test is based on the example in https://github.com/llvm/circt/issues/8560, but uses two layerblocks to also check that the references are cloned, by having two layerblocks with no common parent layerblock, and although the node is outside the layerblocks, the node is canonicalized away by the time we hit advanced layer sink anyways.

Issue #8544, opened 2025-06-07
[FIRRTL] Failed to optimize `bits(cat(cat(..))`.


uenoku:
```mlir
  firrtl.circuit "A" {
    firrtl.module @A(in %clock: !firrtl.clock, in %u: !firrtl.uint<1>, in %cond: !firrtl.uint<1>, out %a: !firrtl.uint<2>) attributes {convention = #firrtl<convention scalarized>} {
      %c0_ui2 = firrtl.constant 0 : !firrtl.uint<2>
      %cat_1 = firrtl.cat %u, %c0_ui2 {name = "cat_1"} : (!firrtl.uint<1>, !firrtl.uint<2>) -> !firrtl.uint<3>
      %cat_2 = firrtl.cat %cat_1, %u {name = "cat_2"} : (!firrtl.uint<3>, !firrtl.uint<1>) -> !firrtl.uint<4>
      %0 = firrtl.bits %cat_2 2 to 1 : (!firrtl.uint<4>) -> !firrtl.uint<2>
      firrtl.matchingconnect %a, %0 : !firrtl.uint<2>
    }
  }
```

Even though %0 is constant 0 it's not optimized at FIRRTL. At core dialect operands are flattened to comb.concat, and it's obviously canonicalized. 

This issue causes this kind of unstable behavior for register optimizations:

```scala
FIRRTL version 5.0.0

circuit A:
  public module A :
    input clock : Clock
    input u: UInt<1>
    input cond : UInt<1>
    output a : UInt<2>
    output b: UInt<2>
    reg r_a : UInt<2>, clock
    reg r_b : UInt<2>, clock
    connect a, r_a
    connect b, r_b
    node cat_1 = cat(u, UInt<2>(0))
    node cat_2 = cat(cat_1, u)
    node next_a = bits(cat_2, 3, 1)
    when cond :
      connect r_a, next_a
      connect r_b, UInt<1>(0)
```
Output:
```verilog
 reg [1:0] r_a;
  always @(posedge clock) begin
    if (cond)
      r_a <= 2'h0;
  end // always @(posedge)
  assign a = r_a;
  assign b = 2'h0;
```

[...]

PR #8578, opened 2025-06-18
[FIRRTL] Improve canonicalization patterns for variadic cat and reduction operations

uenoku:
This commit enhances FIRRTL canonicalization for variadic cat flatten. Several canonicalizations are added to pass the existing tests.  

**Reduction-Cat Patterns:**
- Replace table-driven patterns for AndR/OrR/XorR with cat operands with C++
- AndRCat: Optimizes andr(cat(...)) by handling zero constants (-> 0) and all-ones constants (identity)
- OrRCat: Optimizes orr(cat(...)) by handling non-zero constants (-> 1) and zero constants (identity)
- XorRCat: Optimizes xorr(cat(...)) by preserving non-zero constants and eliminating zero constants

**Cat Optimization Patterns:**
- FlattenCat: Flattens nested cat operations into single cat operations for better optimization opportunities
- CatOfConstant: Folds successive constant operands in cat operations (e.g., cat(x, 3, 5, y) -> cat(x, 29, y))
- BitsOfCat: Optimizes bits extraction from cat operations by extracting directly from the appropriate operand when possible

The new patterns replace the previous table-driven AndRCat*, OrRCat*, and XorRCat* patterns since it's too complicated to handle the condition in tablegen
 
Close https://github.com/llvm/circt/issues/8544

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -1407,8 +1407,7 @@ public:
     while (!worklist.empty()) {
       auto value = worklist.pop_back_val();
       auto catOp = value.getDefiningOp<CatPrimOp>();
-      // Make sure not to flatten if the cat is used by other ops.
-      if (!catOp || !catOp->hasOneUse()) {
+      if (!catOp) {
         operands.push_back(value);
         continue;
       }
--- a/test/Dialect/FIRRTL/canonicalization.mlir
+++ b/test/Dialect/FIRRTL/canonicalization.mlir
@@ -1042,9 +1042,7 @@ firrtl.module @FlattenCat(in %a: !firrtl.uint<2>, in %b: !firrtl.uint<3>, in %c:
   %outer = firrtl.cat %middle1, %middle2 : (!firrtl.uint<10>, !firrtl.uint<5>) -> !firrtl.uint<15>
   firrtl.matchingconnect %out2, %outer : !firrtl.uint<15>
 
-  // Multiply used cat must not be flatten to prevent exponential blowup
-  // CHECK: %[[AB:.+]] = firrtl.cat %a, %b
-  // CHECK: %[[ABAB:.+]] = firrtl.cat %[[AB]], %[[AB]]
+  // CHECK: %[[ABAB:.+]] = firrtl.cat %a, %b, %a, %b
   // CHECK: firrtl.matchingconnect %out3, %[[ABAB]]
   %cat_a_b = firrtl.cat %a, %b : (!firrtl.uint<2>, !firrtl.uint<3>) -> !firrtl.uint<5>
   %cat_a_b_a_b = firrtl.cat %cat_a_b, %cat_a_b : (!firrtl.uint<5>, !firrtl.uint<5>) -> !firrtl.uint<10>


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1111,6 +1111,122 @@ void NotPrimOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+  /// Returns true if the entire reduction can be replaced with a constant.
+  /// May add non-zero constants to the remaining operands list.
+  virtual bool handleConstant(mlir::PatternRewriter &rewriter, Operation *op,
+                              ConstantOp constantOp,
+                              SmallVectorImpl<Value> &remaining) const = 0;
+
+  /// Return the unit value for this reduction operation:
+  /// - OrR: false (0) - OR with 0 is identity
+  /// - AndR: true (1) - AND with 1 is identity
+  /// - XorR: false (0) - XOR with 0 is identity
```
seldridge: Mega-nit: there's no need to forward-document the behavior of the classes that inherit from this.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1111,6 +1111,122 @@ void NotPrimOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+
+  LogicalResult
+  matchAndRewrite(Operation *op,
+                  mlir::PatternRewriter &rewriter) const override {
+    // Check if the operand is a cat operation
+    auto catOp = op->getOperand(0).getDefiningOp<CatPrimOp>();
+    if (!catOp)
+      return failure();
+
+    auto resultType = cast<IntType>(op->getResult(0).getType());
```
seldridge: Nit: this is single use in the `nonConstantOperands.empty()` case. Consider moving inside that `if` block.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1265,10 +1379,117 @@ void DShrPrimOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+        operands.push_back(cat.getInputs()[i]);
+        continue;
+      }
+      APSInt value = cst.getValue();
+      size_t j = i + 1;
+      for (; j < cat->getNumOperands(); ++j) {
+        auto nextCst = cat.getInputs()[j].getDefiningOp<ConstantOp>();
+        if (!nextCst)
+          break;
+        value = value.concat(nextCst.getValue());
```
seldridge, line 1459: It's super convenient that `APSInt` has a `concat` method. I didn't know about this.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1302,11 +1523,45 @@ OpFoldResult BitsPrimOp::fold(FoldAdaptor adaptor) {
...
+      auto operandWidth =
+          type_cast<IntType>(operand.getType()).getWidthOrSentinel();
+      if (operandWidth < 0)
+        return failure();
+      if (bitPos < operandWidth) {
+        if (bitPos + resultWidth <= operandWidth) {
+          rewriter.replaceOpWithNewOp<BitsPrimOp>(
+              op, operand, bitPos + resultWidth - 1, bitPos);
+          return success();
+        }
```
seldridge: This always creates the `BitsPrimOp` even if it is full size because that will be picked up by another canonicalizer/folder?

test/Dialect/FIRRTL/canonicalization.mlir
```diff
@@ -969,6 +969,178 @@ firrtl.module @Reduce(in %a: !firrtl.uint<1>, out %b: !firrtl.uint<1>,
...
+  %xorr1 = firrtl.xorr %cat5 : (!firrtl.uint<6>) -> !firrtl.uint<1>
+  firrtl.matchingconnect %out_xorr, %xorr1 : !firrtl.uint<1>
+
+  // Test XorR with cat containing non-zero constants - constants should remain
+  // CHECK: %[[CAT_XOR:.+]] = firrtl.cat %c3_ui2, %a
+  // CHECK: %[[XORR_CAT:.+]] = firrtl.xorr %[[CAT_XOR]]
+  // CHECK: firrtl.matchingconnect %out_xorr, %[[XORR_CAT]]
+  %cat6 = firrtl.cat %c3_ui2, %a, %c0_ui2 : (!firrtl.uint<2>, !firrtl.uint<4>, !firrtl.uint<2>) -> !firrtl.uint<8>
+  %xorr2 = firrtl.xorr %cat6 : (!firrtl.uint<8>) -> !firrtl.uint<1>
+  firrtl.matchingconnect %out_xorr, %xorr2 : !firrtl.uint<1>
```
seldridge, line 1020: Looking at this, this could _actually_ reduce by looking at the total number of one bits. To do this properly, it should really look across operands and not just within an operand.

This kind of thing would be easier if there was a separate canonicalizer for `cat` which moved constants forwards or backwards.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1265,10 +1379,117 @@ void DShrPrimOp::getCanonicalizationPatterns(RewritePatternSet &results,
   results.insert<patterns::DShrOfConstant>(context);
 }
 
+namespace {
+
+/// Canonicalize a cat tree into single cat operation.
+class FlattenCat : public mlir::RewritePattern {
```
seldridge, line 1382: This seems fine.

It would be nice if there was a way that the cat rely on the recursive canonicalization to make this simpler. E.g., if this could only work at the level of inlining a cat into another cat and then this would naturally flatten the tree structure. However, I think this would require multiple explicit recursive canonicalizations as opposed to it being automatic.

Any idea if this just works with the simpler version? This can be a follow-on discussion.

test/Dialect/FIRRTL/canonicalization.mlir
```diff
@@ -969,6 +969,178 @@ firrtl.module @Reduce(in %a: !firrtl.uint<1>, out %b: !firrtl.uint<1>,
...
+  %middle2 = firrtl.cat %e, %f : (!firrtl.uint<2>, !firrtl.uint<3>) -> !firrtl.uint<5>
+  %outer = firrtl.cat %middle1, %middle2 : (!firrtl.uint<10>, !firrtl.uint<5>) -> !firrtl.uint<15>
+  firrtl.matchingconnect %out2, %outer : !firrtl.uint<15>
+
+  // Multiply used cat must not be flatten to prevent exponential blowup
+  // CHECK: %[[AB:.+]] = firrtl.cat %a, %b
+  // CHECK: %[[ABAB:.+]] = firrtl.cat %[[AB]], %[[AB]]
+  // CHECK: firrtl.matchingconnect %out3, %[[ABAB]]
+  %cat_a_b = firrtl.cat %a, %b : (!firrtl.uint<2>, !firrtl.uint<3>) -> !firrtl.uint<5>
+  %cat_a_b_a_b = firrtl.cat %cat_a_b, %cat_a_b : (!firrtl.uint<5>, !firrtl.uint<5>) -> !firrtl.uint<10>
```
seldridge: I think we maybe do want to do this? Did you observe problems in internal designs with this on?

Alternatively, maybe we need a replicate operator for this.

seldridge (approved):
LGTM

This generated some discussion/thoughts that are all non-blocking.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
uenoku: BitsPrimOp folder will take care of that.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
uenoku: Each flatten takes O(length of operands) so it's important to flatten all at once to prevent O(N^2) regression.

test/Dialect/FIRRTL/canonicalization.mlir
uenoku: It's necessary to duplicate all non-constant operands of concat if we want to canonicalize further so I think for now I'd like to keep it as is.

lib/Dialect/FIRRTL/FIRRTLFolds.cpp
seldridge: Good point. This makes sense to me.

test/Dialect/FIRRTL/canonicalization.mlir
uenoku: Hmm, yeah. I thought it was necessary not to flatten multiply used concat but seems like comb performs flattening for multiply used concat. I'll just remove the condition.

PR #8579, opened 2025-06-18
[circt-verilog] Add updated LLHD passes to the pipeline

fabianschuiki:
Enable the LLHD passes that have been updated in the past few months in circt-verilog's pass pipeline. This contains revised versions of the Mem2Reg, Deseq, and LowerProcesses passes, alongside a new HoistSignals pass and the CombineDrives pass factored out of Sig2Reg.

This also disables the SROA pass on LLHD, moving it to only operate on Moore modules. @maerhart has been looking into fixing an upstream bug in SROA when it's used on operations in graph regions. Until that is fixed, we can just skip SROA on HW modules for now.

Also add an end-to-end test for circt-verilog to ensure that the basic forms of registers get detected and lowered properly.

### DIFF
--- a/test/circt-verilog/registers.sv
+++ b/test/circt-verilog/registers.sv
@@ -1,0 +1,64 @@
+// RUN: circt-verilog %s | FileCheck %s
+// REQUIRES: slang
+
+// CHECK-LABEL: hw.module @ClockPosEdgeAlwaysFF(
+module ClockPosEdgeAlwaysFF(input logic clock, input int d, output int q);
+  // CHECK: [[CLK:%.+]] = seq.to_clock %clock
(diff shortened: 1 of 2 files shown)


uenoku (approved):
LG. Does this mean ProcessLowering pass is deprecated?

fabianschuiki:
@IanTrudel This PR enables the passes in the circt-verilog pipeline that would help cover some of the use cases you were thinking of.

fabianschuiki:
@uenoku Yeah, a handful of passes is now deprecated. I've generally created new passes and moved a lot of the existing code over and adapted it to work on a few corner cases I've found. Once this lands we should be able to remove the original code.

PR #8580, opened 2025-06-19
[CI] Add smarter cache cleanup with size-based filtering

uenoku:
Updated cache cleanup system that removes small caches (likely from failed builds) and maintains only the most recent successful builds.

This implementation uses a dynamic size threshold of 1/2 of maximum cache size as the cleanup threshold. It employs regex pattern matching to find caches with `.*PATTERN.*` for flexible matching. The system keeps the top 1 cache to preserve the most recent successful build and is safe by default with dry-run mode enabled to prevent accidental deletions.

The changes add a reusable cleanup workflow at .github/workflows/cleanupCache.yml and integrate cache cleanup into unified build workflows. Cache cleanup has been added to short integration tests with cache key pattern generation and reuse implemented.

The cache cleanup logic calculates the threshold as 50% of largest cache size, then deletes all caches smaller than the threshold (failed builds) and deletes caches beyond top 1 that exceed the threshold (old builds). It uses regex matching for flexible cache key patterns.

The workflows updated include unifiedBuildTestAndInstall.yml, unifiedBuildTestAndInstallStatic.yml, and shortIntegrationTests.yml and windowsCI.


.github/workflows/cleanupCache.yml
```diff
@@ -0,0 +1,75 @@
...
+          # 1. All caches smaller than threshold (failed builds)
+          # 2. Caches beyond top 1 that are >= threshold (old successful builds)
+          if [ "$MAX_SIZE" -gt 0 ]; then
+            echo "$MATCHING_CACHES_IN_CURRENT_BRANCH" | \
+              jq --argjson threshold "$SIZE_THRESHOLD" \
+              '[.[] | select(.sizeInBytes < $threshold)] + (.[1:] | map(select(.sizeInBytes >= $threshold))) | .[].id' | \
+              ./utils/delete-cache.sh
+          else
+            echo "No caches found matching pattern: .*${CACHE_KEY_PATTERN}.*"
+          fi
```
seldridge: This will be more maintainable/readable as a shell script inside `.github/bin/` (or `utils/`). This can also then be a bit more of a real "tool" with option parsing, error handling, help text, etc. It's admittedly overkill. At minimum, please break this out, as-is (changing it to communicate the github `${{ ... }}` as arguments) to a shell script.

.github/workflows/shortIntegrationTests.yml
```diff
@@ -102,3 +111,12 @@ jobs:
...
           ninja -C build check-circt-integration -j$(nproc)
+  # Clean up cache after build completes
+  cleanup-cache:
+    needs: build-circt
+    if: always() # Run even if build fails
+    uses: ./.github/workflows/cleanupCache.yml
+    permissions:
+      actions: write
+    with:
+      cache_key_pattern: ${{ needs.build-circt.outputs.cache-key-pattern }}
```
seldridge, line 122: It's kind of annoying that this can't prevent the write into the cache and will instead cleanup the cache after the cache has been updated. However, this is way better than what we have now.

.github/workflows/cleanupCache.yml
```diff
@@ -0,0 +1,75 @@
...
+          # Get all caches and filter with regex pattern
+          ALL_CACHES=$(gh cache list -L 200 --json id,key,sizeInBytes)
+          MATCHING_CACHES=$(echo "$ALL_CACHES" | jq --arg pattern ".*${CACHE_KEY_PATTERN}.*" '[.[] | select(.key | test($pattern))]')
+
+          echo "Found $(echo "$MATCHING_CACHES" | jq 'length') matching caches"
+          echo "$MATCHING_CACHES" | jq -r '.[] | "\(.key) - \(.sizeInBytes) bytes"'
+
+          # Calculate the size threshold as 1/2 of the maximum cache size
+          MAX_SIZE=$(echo "$MATCHING_CACHES" | jq 'map(.sizeInBytes) | max // 0')
+          SIZE_THRESHOLD=$((MAX_SIZE / 2))
```
seldridge: I'm surprised this just works? I thought you would need to use `dc` or `bc` to do math in a shell script?

utils/delete-cache.sh
```diff
@@ -1,6 +1,36 @@
...
 
+# Enhanced cache deletion script with logging
+# Reads cache IDs from stdin and deletes them
+# Usage: ./delete-cache.sh [--dry-run]
+
+DRY_RUN=false
+if [[ "$1" == "--dry-run" ]]; then
+  DRY_RUN=true
+  echo "DRY RUN MODE: No caches will actually be deleted"
+fi
```
seldridge: This would be better if written to use `getopts`. It's admittedly overkill.

utils/delete-cache.sh
```diff
@@ -1,6 +1,36 @@
...
+      echo "Deleting cache ID: $id"
+      if gh cache delete "$id"; then
+        echo "Successfully deleted cache ID: $id"
+        ((count++))
+      else
+        echo "Failed to delete cache ID: $id" >&2
+      fi
+      sleep 2
+    fi
+  fi
```
seldridge: You can save on indentation here by inverting the match and using a continue.

utils/delete-cache.sh
```diff
@@ -1,6 +1,36 @@
...
+      ((count++))
+    else
+      echo "Deleting cache ID: $id"
+      if gh cache delete "$id"; then
+        echo "Successfully deleted cache ID: $id"
+        ((count++))
+      else
+        echo "Failed to delete cache ID: $id" >&2
+      fi
+      sleep 2
```
seldridge: What's the sleep for?

utils/delete-cache.sh
seldridge: Also, this addition of a dry run option is great.

.github/workflows/cleanupCache.yml
uenoku: I created a shell script under utils

.github/workflows/cleanupCache.yml
uenoku: I think posix shell supports (( )) for calculating single arithmetic expression( 2.6.4 https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html)

utils/delete-cache.sh
uenoku: There is rate limit for github API so just making sure it doesn't call overwhelming number of API calls. Though it wouldn't be a problem practically.

PR #8584, opened 2025-06-19
[Arc] don't sink ops with nested writes in MergeIfs

TaoBi22:
Currently MergeIfs makes sure it doesn't sink a write op, but it doesn't check whether it's sinking an op that _contains_ a write op. The current version sinks the first IfOp in the integration test into the second, which is invalid as it adds a new condition to the write inside. Instead, this PR just tracks whether the op currently being sunk contains or is a write, and uses that to decide whether to sink.


fabianschuiki (approved):
Ugh, nasty issue! Awesome, thanks a lot for finding and fixing this :partying_face:

PR #8588, opened 2025-06-20
[AIG][ImportAIGER] Fix incorrect tokenization and simplify parser

uenoku:
- Fix lexer tokenization by removing whitespace skipping at the end. This could be issue when whitespace is actually a part of binary encoding of AND gate definition 
- Add location-aware error reporting with SMLoc tracking throughout parser


fabianschuiki (approved):
Nice catch! LGTM

PR #8590, opened 2025-06-20
[LLHD] Embed procedural ops within modules into llhd.combinational ops

fabianschuiki:
Add the WrapProceduralOps pass that takes operations such as `func.call` or `scf.if` in HW module body and wraps them inside `llhd.combinational` ops. This allows those operations to be inlined or lowered to the CF dialect since the `llhd.combinational` op provides an SSACFG region.

Add the pass to the circt-verilog pipeline, alongside a lowering from SCF to CF. This makes the core dialect output of circt-verilog more regular, and will allow us to fully inline function calls and fully unroll loops in the future.


uenoku (approved):
LGTM

PR #8594, opened 2025-06-23
[firtool] Enable WireElimination pass

uenoku:
This adds EliminateWiresPass before IMCP

### DIFF
--- a/test/Dialect/FIRRTL/SFCTests/data-taps.fir
+++ b/test/Dialect/FIRRTL/SFCTests/data-taps.fir
@@ -128,7 +128,7 @@ circuit Top : %[[
     connect in, or(tap_0, tap_2[0])
     connect out, or(tap_1, tap_2[1])
 
-; CHECK:      module Top
+; CHECK-LABEL:      module Top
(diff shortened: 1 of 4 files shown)


test/Dialect/FIRRTL/SFCTests/data-taps.fir
```diff
@@ -128,7 +128,7 @@ circuit Top : %[[
     connect in, or(tap_0, tap_2[0])
     connect out, or(tap_1, tap_2[1])
 
-; CHECK:      module Top
+; CHECK-LABEL:      module Top
```
seldridge: Is this actually necessary?

If it is, please keep this stuff aligned. Though, I think that would require adding one extra space to everything.

Ditto throughout.

seldridge (approved):
LGTM

Have you checked the Verilog diff on any internal designs?

uenoku:
Diff is massive because of naming changes and more optimization opportunities. Verified LEC on small core (but probably LEC wouldn't be verified in general because of register optimizations), and manually looked at diff on large core and the change looks reasonable. Created a flag to disable the pass so we can turn off the pass when there is an issue.

PR #8596, opened 2025-06-23
[FIRRTL] Move all annotations in inject-dut-hier

seldridge:
Change FIRRTL's inject-dut-hier pass to move all annotations onto the
wrapper [^1] when in `moveDut=true` mode.  Previously, this would only move
the `MarkDUTAnnotation`.  This fixes an internal issue where Object Model
paths were not being updated when using this pass in `moveDut=true` mode.

[^1]: This is using the terminology of the pass which views this as taking
all the logic inside the original DUT and shoving it into a wrapper
instantiated inside the DUT.  When in `moveDut=true` mode, it is more
natural to think of the pass as creating a wrapper around the original
DUT.  However, to adopt this alternative terminology would be modal.
I.e., "wrapper" would mean something different when in `movedDut=true`
mode which would be confusing.

Other NFC chagnes:

1. Refactor the InjectDUTHierarchy pass to not introduce the notion of a
   `newDUT` variable.  This only confuses things.  Instead setup the DUT and its
   internal wrapper immediately and then reuse these terms in all places.
2. Rename a test which is now the main test of non-local annotation (NLA)
   behavior inside FIRRTL's inject-dut-hier pass.  Previously this was the "new"
   test which duplicated a now deleted test.
3. Remove a test of non-local annotation (NLA) handling in the inject-dut-hier
   FIRRTL pass.  This test started out as the original test, then a copy of it
   was created that used a different NLA format, then the test became a true
   duplicate.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
@@ -157,51 +157,42 @@ void InjectDUTHierarchy::runOnOperation() {
   // dut and wrapper do not match the logical definition.
   OpBuilder b(circuit.getContext());
   CircuitNamespace circuitNS(circuit);
-  {
-    b.setInsertionPointAfter(dut);
-    auto newDUT = b.create<FModuleOp>(dut.getLoc(), dut.getNameAttr(),
-                                      dut.getConventionAttr(), dut.getPorts(),
-                                      dut.getAnnotationsAttr());
+  b.setInsertionPointAfter(dut);
 
-    // This pass shouldn't create new public modules.  It should only preserve
-    // the existing public modules.  In "moveDut" mode, then the wrapper is the
-    // new DUT and we should move the publicness from the old DUT to the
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/Transforms/InjectDUTHierarchy.cpp
```diff
@@ -157,50 +209,59 @@ void InjectDUTHierarchy::runOnOperation() {
...
-    // avoid the tech debt that it creates.
-    if (moveDut) {
-      newDUT.setPrivate();
-    } else {
-      newDUT.setVisibility(dut.getVisibility());
-      dut.setPrivate();
+  b.setInsertionPointAfter(dut);
+
+  // After this, he original DUT module is now the "wrapper".  The new module we
+  // just created becomse the "DUT".
```
uenoku:
```suggestion
  // just created becomes the "DUT".
```

seldridge: Fixed in the rebase.

PR #8597, opened 2025-06-24
[LLHD] Update function call inlining pass; add to circt-verilog

fabianschuiki:
Update the LLHD pass intended to inline all function calls.

This commit renames the pass from "FunctionElimination" to "InlineCalls", since the updated pass no longer deletes the functions. Instead, MLIR's SymbolDCE pass should be used after call inlining to get rid of any unused function definitions.

The pass logic is also adjusted to support nested function calls. This allows deeply nested call graphs to be completely inlined, which is a prerequisite for many synthesis-related passes that may follow in the pipeline. A stack of called functions is used to track function nesting and prevent recursive calls from infinitely expanding.

As before, the pass produces an error diagnostic for any call op nested within an HW module that cannot be inlined.

This commit also adds the call inlining pass to the circt-verilog pipeline, alongside an additional SymbolDCE run to clean up the unused functions afterwards.


lib/Dialect/LLHD/Transforms/InlineCalls.cpp
```diff
@@ -0,0 +1,164 @@
...
+};
+} // namespace
+
+void InlineCallsPass::runOnOperation() {
+  auto &symbolTable = getAnalysis<SymbolTable>();
+  parallelForEach(&getContext(), getOperation().getOps<hw::HWModuleOp>(),
+                  [&](auto module) {
+                    CallStack callStack;
+                    runOnRegion(module.getBody(), symbolTable, callStack);
+                  });
```
uenoku: Can we change runOnRegion to return LogicalResult and use `parallelForEachFailable` here?

lib/Dialect/LLHD/Transforms/InlineCalls.cpp
```diff
@@ -0,0 +1,164 @@
...
+    // Only inline `func.func` ops.
+    if (!isa<func::FuncOp>(callable))
+      return false;
+
+    // Only inline into SSACFG regions embedded within LLHD processes.
+    if (!mayHaveSSADominance(*call->getParentRegion()))
+      return false;
+    auto *op = call;
+    while ((op = op->getParentOp()))
+      if (isa<ProcessOp, FinalOp, CombinationalOp>(op))
```
uenoku: Not blocking but can we create a trait or interface for these 3 operations similar to ProceduralOp in SV?

include/circt/Dialect/LLHD/Transforms/LLHDPasses.td
```diff
@@ -74,17 +74,24 @@ def MemoryToBlockArgument : Pass<"llhd-memory-to-block-argument",
...
+  let summary = "Inline all function calls in HW modules";
   let description = [{
-    Deletes all functions in the module. In case there is still a function
-    call in an entity or process, it fails.
-    This pass is intended as a post-inlining pass to check if all functions
-    could be successfully inlined and remove the inlined functions. This
-    is necessary because Structural LLHD does not allow functions. Fails in
-    the case that there is still a function call left in a `llhd.process` or
-    `hw.module`.
+    Inlines all `func.call` operations nested within `llhd.combinational` ops
```
uenoku: nit: `llhd.process` and `llhd.final` as well

lib/Dialect/LLHD/Transforms/InlineCalls.cpp
fabianschuiki: Good point. I don't know why I changed this from the old pattern with LogicResult 🙈. Let me switch this back over 👍

lib/Dialect/LLHD/Transforms/InlineCalls.cpp
fabianschuiki: Excellent idea! I've added an LLHD `ProceduralRegion` trait for these ops 👍

PR #8598, opened 2025-06-24
[FIRRTL] Allow duplicate tracker annotations in LowerClasses.

mikeurbach:
Historically, we would assert that every tracker was used exactly once. However, with certain combinations of multiple instantiation and dedup, we can run into duplicate trackers. Specifically, like in the test case, if there are two modules that dedup, but they are instantiated multiple times in modules that do not dedup, we can end up with duplicate trackers that point to the same thing. When dedup is working, the paths are distinct, and it is only at the end that we realize they are pointing to the same thing.

This handles that case, by actually checking the path associated with duplicated trackers. If the trackers ultimately ended up pointing to the same thing, this is now allowed. It is still an error to have the same tracker pointing to multiple different paths through the hierarchy.

Finally, this removed the sentinel PathInfoTableEntry objects we would create simply to check uniqueness. Now that we are actually using the paths to confirm uniqueness or lack thereof, we don't need the sigil, and putting in a PathInfoTableEntry without a path means we can't perform the new check. These only existed for unused paths, which the situation described above in dedup can create. Since they're dead, we can now safely just skip them.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerClasses.cpp
@@ -594,14 +594,8 @@ PathTracker::processPathTrackers(const AnnoTarget &target) {
     // Get the owning module. If there is no owning module, then this
     // declaration does not have a use, and we can return early.
     auto owningModule = owningModules.lookup(id);
-    if (!owningModule) {
-      PathInfoTableEntry entry;
-      // This entry is used for checking id uniquness.
-      entry.op = op;
-      entry.id = id;
-      entries.push_back(entry);
+    if (!owningModule)
       return true;
-    }
 
     // Copy the middle part from the annotation's NLA.
     if (hierPathOp) {
@@ -711,10 +705,26 @@ LogicalResult PathTracker::updatePathInfoTable(PathInfoTable &pathInfoTable,
     pathInfoTable.addAltBasePathRoot(root);
 
   for (const auto &entry : entries) {
+    assert(entry.pathAttr &&
+           "expected all PathInfoTableEntries to have a pathAttr");
+
     // Record the path operation associated with the path op.
     auto [it, inserted] = pathInfoTable.table.try_emplace(entry.id);
     auto &pathInfo = it->second;
     if (!inserted) {
+      // If this DistinctAttr has been seen before, check if it actually points
+      // to the same thing. This can happen in Dedup, where multiple trackers
+      // end up getting created for different paths, which ultimately coalesce
+      // into the same path after Dedup completes. Unfortunately, we can't
+      // easily detect this in the middle of Dedup, so allow duplicate
+      // DistinctAttrs here.
+      assert(pathInfo.symRef && "expected all PathInfos to have a symRef");
+      auto existingHierpath =
+          symbolTable.lookup<hw::HierPathOp>(pathInfo.symRef.getValue());
+      auto existingPathAttr = existingHierpath.getNamepath();
+      if (existingPathAttr == entry.pathAttr)
+        continue;
+
       assert(pathInfo.loc.has_value() && "all PathInfo should have a Location");
       auto diag = emitError(pathInfo.loc.value(),
                             "path identifier already found, paths must resolve "
@@ -727,13 +737,8 @@ LogicalResult PathTracker::updatePathInfoTable(PathInfoTable &pathInfoTable,
     // be invalidated later, so this is the last time we want to access it here.
     bool canBeInstanceTarget = isa<InstanceOp, FModuleLike>(entry.op);
 
-    if (entry.pathAttr) {
-      pathInfo = {entry.op->getLoc(), canBeInstanceTarget,
-                  cache.getRefFor(entry.pathAttr), entry.altBasePathModule};
-    } else {
-      pathInfo.loc = entry.op->getLoc();
-      pathInfo.canBeInstanceTarget = canBeInstanceTarget;
-    }
+    pathInfo = {entry.op->getLoc(), canBeInstanceTarget,
+                cache.getRefFor(entry.pathAttr), entry.altBasePathModule};
   }
   return success();
 }
--- a/test/Dialect/FIRRTL/lower-classes-errors.mlir
+++ b/test/Dialect/FIRRTL/lower-classes-errors.mlir
@@ -32,6 +32,9 @@ firrtl.circuit "PathIllegalHierpath" {
 
 firrtl.circuit "PathDuplicateID" {
   firrtl.module @PathDuplicateID() {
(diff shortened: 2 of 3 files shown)


mikeurbach:
@youngar I know this is ancient history at this point, but I think you originally worked on this. What do you think about relaxing this? For more context on top of the PR description, I was originally trying to _not_ create duplicate tracker annotations in dedup. It comes up here: https://github.com/llvm/circt/blob/0ab5e06471c947d77f080b7f254dcb903ad6a504/lib/Dialect/FIRRTL/Transforms/Dedup.cpp#L1046

In debugging the exact case I saw, I noticed that we had two parents, say, Foo1 and Foo2, which _do not_ dedup. They each instantiate two modules, say, Bar1 and Bar2, which have path tracker annotations, and _do_ dedup. So in `createNLAs`, we have 4 distinct paths through the hierarchy-- Bar1 > Foo1, Bar1 > Foo2, Bar2 > Foo1, Bar2 > Foo2. In `createNLAs`, the `arrayAttr` is distinct for each of those, and we do create four distinct NLAs. In the end, after dedup, the instantiations are Bar1 > Foo1, Bar1 > Foo1, Bar2 > Foo1, Bar2 > Foo1. That results in 2 duplicate tracker IDs, and 2 dead tracker IDs. I couldn't see a nice way to handle this in dedup without adding even more crazy logic, so instead I didn't touch dedup, but allowed LowerClasses to handle this case by: a) not worrying about the dead tracker IDs, and b) handling duplicate tracker IDs by allowing them, if the NLAs actually referred to the same hierarchical path. I _think_ this is sound.

mikeurbach:
I realize now that the test case I checked in is the simplest possible thing to demonstrate the feature... I can write a more complex one that illustrates the exact situation I described, but the IR after dedup looked like this:

```mlir
firrtl.circuit "Test" {
  hw.hierpath private @nla_2 [@Bar1::@sym_0, @Foo1]
  hw.hierpath private @nla_1 [@Bar2::@sym_0, @Foo1]
  hw.hierpath private @nla_0 [@Bar1::@sym, @Foo1]
  hw.hierpath private @nla [@Bar2::@sym, @Foo1]
  firrtl.class private @SomeClass(in %path_in: !firrtl.path, out %path: !firrtl.path) attributes {portAnnotations = [[], []]} {
    firrtl.propassign %path, %path_in : !firrtl.path
  }
  firrtl.module @Test(in %in: !firrtl.uint<1>, out %out: !firrtl.uint<1>) attributes {convention = #firrtl<convention scalarized>} {
    %bar1_in, %bar1_out = firrtl.instance bar1 @Bar1(in in: !firrtl.uint<1>, out out: !firrtl.uint<1>)
    %bar2_in, %bar2_out = firrtl.instance bar2 @Bar2(in in: !firrtl.uint<1>, out out: !firrtl.uint<1>)
    firrtl.matchingconnect %bar1_in, %in : !firrtl.uint<1>
    firrtl.matchingconnect %bar2_in, %bar1_out : !firrtl.uint<1>
    firrtl.matchingconnect %out, %bar2_out : !firrtl.uint<1>
  }
  firrtl.module private @Bar1(in %in: !firrtl.uint<1>, out %out: !firrtl.uint<1>) {
    %0 = firrtl.or %in, %in : (!firrtl.uint<1>, !firrtl.uint<1>) -> !firrtl.uint<1>
    %n = firrtl.node %0 : !firrtl.uint<1>
    %foo1_in, %foo1_out, %foo1_ref = firrtl.instance foo1 sym @sym @Foo1(in in: !firrtl.uint<1>, out out: !firrtl.uint<1>, out ref: !firrtl.class<@SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)>)
    %foo2_in, %foo2_out, %foo2_ref = firrtl.instance foo2 sym @sym_0 @Foo1(in in: !firrtl.uint<1>, out out: !firrtl.uint<1>, out ref: !firrtl.class<@SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)>)
    firrtl.matchingconnect %foo1_in, %n : !firrtl.uint<1>
    firrtl.matchingconnect %foo2_in, %foo1_out : !firrtl.uint<1>
    firrtl.matchingconnect %out, %foo2_out : !firrtl.uint<1>
  }
  firrtl.module private @Bar2(in %in: !firrtl.uint<1>, out %out: !firrtl.uint<1>) {
    %foo1_in, %foo1_out, %foo1_ref = firrtl.instance foo1 sym @sym @Foo1(in in: !firrtl.uint<1>, out out: !firrtl.uint<1>, out ref: !firrtl.class<@SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)>)
    %foo2_in, %foo2_out, %foo2_ref = firrtl.instance foo2 sym @sym_0 @Foo1(in in: !firrtl.uint<1>, out out: !firrtl.uint<1>, out ref: !firrtl.class<@SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)>)
    firrtl.matchingconnect %foo1_in, %in : !firrtl.uint<1>
    firrtl.matchingconnect %foo2_in, %foo1_out : !firrtl.uint<1>
    firrtl.matchingconnect %out, %foo2_out : !firrtl.uint<1>
  }
  firrtl.module private @Foo1(in %in: !firrtl.uint<1> [{circt.nonlocal = @nla, class = "circt.tracker", id = distinct[0]<>}, {circt.nonlocal = @nla_0, class = "circt.tracker", id = distinct[0]<>}, {circt.nonlocal = @nla_1, class = "circt.tracker", id = distinct[1]<>}, {c\
irct.nonlocal = @nla_2, class = "circt.tracker", id = distinct[1]<>}], out %out: !firrtl.uint<1>, out %ref: !firrtl.class<@SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)>) {
    %obj = firrtl.object @SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)
    %0 = firrtl.object.subfield %obj[path_in] : !firrtl.class<@SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)>
    %1 = firrtl.path reference distinct[0]<>
    firrtl.propassign %0, %1 : !firrtl.path
    firrtl.propassign %ref, %obj : !firrtl.class<@SomeClass(in path_in: !firrtl.path, out path: !firrtl.path)>
    firrtl.matchingconnect %out, %in : !firrtl.uint<1>
  }
}
```

Note that in the end, there is only one `firrtl.path`, even though the NLAs begin from both Bar1 and Bar2. It would seem like we are losing information, but because of the way paths get prefixed down the hierarchy in LowerClasses, by the time the final paths are frozen, we do get the hierarchical paths we'd expect.

mikeurbach:
Thanks for the review @youngar.

PR #8600, opened 2025-06-25
[LLHD] Add pass to lower acyclic control flow to mux ops

fabianschuiki:
Add the *RemoveControlFlow* pass to the LLHD dialect. It operates on all `llhd.combinational` ops in an `hw.module` if the op has no side-effects and the control flow is acyclic. The pass moves all operations in the `lhld.combinational` op body into its entry block and inserts `comb.mux` ops to resolve block arguments. All blocks besides the entry block are then removed.

A future loop unrolling pass would get rid of any cycles in the control flow, such that this pass can then pick up the remaining control flow lowering.

This is an important pass in lowering behavioral circuit descriptions, such as `always` blocks in Verilog or `process`es in VHDL, to a set of equivalent Comb dialect ops in the module body.

Shoutout to @maerhart for doing all the heavy lifting for this pass on his experimental branch!

### DIFF
--- a/lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
+++ b/lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
@@ -1,0 +1,384 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/LLHD/IR/LLHDOps.h"
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/IR/Dominance.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/Pass/Pass.h"
+#include "llvm/ADT/PostOrderIterator.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "llhd-remove-control-flow"
+
+namespace circt {
+namespace llhd {
+#define GEN_PASS_DEF_REMOVECONTROLFLOWPASS
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h.inc"
+} // namespace llhd
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+using namespace llhd;
+
+//===----------------------------------------------------------------------===//
+// Utilities
+//===----------------------------------------------------------------------===//
+
+namespace {
+/// A helper struct that tracks a boolean condition as either a constant false,
+/// constant true, or an SSA value.
+struct Condition {
+  Condition() {}
+  Condition(Value value) : pair(value, 0) {
+    if (value) {
+      if (matchPattern(value, m_One()))
+        *this = Condition(true);
+      if (matchPattern(value, m_Zero()))
+        *this = Condition(false);
+    }
+  }
+  Condition(bool konst) : pair(nullptr, konst ? 1 : 2) {}
+
+  explicit operator bool() const {
+    return pair.getPointer() != nullptr || pair.getInt() != 0;
+  }
+
+  bool isTrue() const { return !pair.getPointer() && pair.getInt() == 1; }
+  bool isFalse() const { return !pair.getPointer() && pair.getInt() == 2; }
+  Value getValue() const { return pair.getPointer(); }
+
+  /// Turn this condition into an SSA value, creating an `hw.constant` if the
+  /// condition is a constant.
+  Value materialize(OpBuilder &builder, Location loc) const {
+    if (isTrue())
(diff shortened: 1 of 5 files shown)


lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
```diff
@@ -0,0 +1,322 @@
...
+    auto *yieldBlock = builder.createBlock(&region, region.end(),
+                                           yieldOps[0].getOperandTypes(), locs);
+    sortedBlocks.push_back(yieldBlock);
+    yieldOp =
+        builder.create<YieldOp>(region.getLoc(), yieldBlock->getArguments());
+    for (auto yieldOp : yieldOps) {
+      builder.setInsertionPoint(yieldOp);
+      builder.create<cf::BranchOp>(yieldOp.getLoc(), yieldBlock,
+                                   yieldOp.getOperands());
+      yieldOp.erase();
```
uenoku, line 231: Please add a test for multiple yield.

lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
```diff
@@ -0,0 +1,322 @@
...
+          LLVM_DEBUG(llvm::dbgs() << "  - Both from " << condBrOp << "\n");
+          SmallVector<Value> mergedOperands;
+          mergedOperands.reserve(block->getNumArguments());
+          for (auto [trueArg, falseArg] :
+               llvm::zip(condBrOp.getTrueDestOperands(),
+                         condBrOp.getFalseDestOperands())) {
+            mergedOperands.push_back(builder.createOrFold<comb::MuxOp>(
+                trueArg.getLoc(), condBrOp.getCondition(), trueArg, falseArg));
+          }
+          mergeArgs(mergedOperands, Value{}, false);
```
uenoku, line 322: I think test is missing for this as well.

lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
```diff
@@ -0,0 +1,322 @@
...
+      // Helper function to create a multiplexer between the current
+      // `mergedArgs` and a new set of `args`, where the new args are picked if
+      // `cond` is true and control flows from `domBlock` to `pred`. The
+      // condition may be null, in which case the mux will directly use the
+      // branch decisions that lead from `domBlock` to `pred`.
+      auto mergeArgs = [&](ValueRange args, Value cond, bool invCond) {
+        if (mergedArgs.empty()) {
+          mergedArgs = args;
+          return;
+        }
```
uenoku, line 287: What is expected behavior when `mergedArgs` is empty and cond is nonnull? I think that means there is control flow path that these arguments are uninitialized? Is it correct to drop condition here, or MLIR ensures there is no such control path somehow?

lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
```diff
@@ -0,0 +1,322 @@
...
+  // can combine them into a single yield block later.
+  SmallVector<YieldOp, 2> yieldOps;
+  SmallPtrSet<Block *, 8> visitedBlocks, ipoSet;
+  for (auto &block : region) {
+    for (auto *ipoBlock : llvm::inverse_post_order_ext(&block, ipoSet)) {
+      if (!llvm::all_of(ipoBlock->getPredecessors(), [&](auto *pred) {
+            return visitedBlocks.contains(pred);
+          })) {
+        LLVM_DEBUG(llvm::dbgs() << "- Loop detected, giving up\n");
+        return success();
```
uenoku: Shouldn't it return failure? Control flow loop in combinatorial op feels fine to reject. Or is there well defined semantics for that?

uenoku (approved):
I have few question but otherwise LGTM. Using MLIR dominance is really smart.

lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
fabianschuiki: Nice catch, thanks

lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
fabianschuiki: Yeah, this is missing. Thanks!

lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
fabianschuiki: That happens if this is the very first control flow edge reaching the block. Since there was no other earlier control flow edge with block arguments, we don't need the condition to pick between those earlier args or the new args. This is basically like an "else" branch in the mux. For one CF edge, the first one in this case, we simply discard the condition and just use the value if all other conditions did not match.

lib/Dialect/LLHD/Transforms/RemoveControlFlow.cpp
fabianschuiki: I think it's okay for the pass to not transform combinational ops that don't meet the control flow removal criteria. These can then coexist in the IR. For synthesis we'll then want a pass that either removes or errors out if there are any such ops left, but for simulation you'll likely have a few combinational ops left over with some weird control flow that you can still simulate.

fabianschuiki: I think I might as well remove the `LogicalResult` return, since there aren't any ways for the removal to fail. Either it doesn't happen at all, or it works.

PR #8603, opened 2025-06-26
[FIRRTL] Allow GC Companion multi-instantiation

seldridge:
Change the Grand Central View lowering pass to now allow for Grand Central
Companions to be multiply instantiated.  This should have technically been
possible for a long time [[1]] as long as there was a guarantee that there
was never going to be a leaf that was _outside_ the companion.

This is a pretty simple change to just loop over all instances and extract
(add a `lowerToBind` attribute) for the ones which are in the design.  The
other, subtle blocker is that without the `InstanceInfo`, doing the check
of "in the design" was tedious.  Now this is pretty straightforward.

[1]: https://github.com/llvm/circt/commit/2d3a040f43053d10d195980626ed321a653a89f4

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
@@ -27,6 +27,7 @@
 #include "circt/Support/Debug.h"
 #include "mlir/Pass/Pass.h"
 #include "llvm/ADT/DepthFirstIterator.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/YAMLTraits.h"
@@ -2093,7 +2094,8 @@ void GrandCentralPass::runOnOperation() {
               }
 
               // Extract all instances in the design.
-              for (auto *i : instancePaths->instanceGraph.lookup(op)->uses()) {
+              for (auto *i : llvm::make_early_inc_range(
+                       instancePaths->instanceGraph.lookup(op)->uses())) {
                 auto instance = cast<InstanceOp>(i->getInstance());
 
                 // Skip extraction if the instance is not in the design.
@@ -2112,6 +2114,7 @@ void GrandCentralPass::runOnOperation() {
                         builder.create<WireOp>(port.getLoc(), port.getType());
                     port.replaceAllUsesWith(wire.getResult());
                   }
+                  i->erase();
                   instance->erase();
                 } else {
                   // Lower the companion to a bind unless the user told us
@@ -2228,7 +2231,9 @@ void GrandCentralPass::runOnOperation() {
         if (nla.root() == name)
           nla.erase();
       }
-
+      auto *node = instancePaths->instanceGraph.lookup(
+          cast<igraph::ModuleOpInterface>(*mod));
+      instancePaths->instanceGraph.erase(node);
       mod->erase();
     }
 


lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -2113,47 +2085,47 @@ void GrandCentralPass::runOnOperation() {
...
+                if (companionMode == CompanionMode::Drop) {
+                  // Delete the instance if companions are disabled.
+                  OpBuilder builder(&getContext());
+                  for (auto port : instance->getResults()) {
+                    builder.setInsertionPointAfterValue(port);
+                    auto wire =
+                        builder.create<WireOp>(port.getLoc(), port.getType());
+                    port.replaceAllUsesWith(wire.getResult());
+                  }
+                  instance->erase();
```
uenoku, line 2118: I think it's necessary to erase through InstanceGraph.

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -2113,47 +2085,47 @@ void GrandCentralPass::runOnOperation() {
...
               }
 
-              // If the companion is instantiated above the DUT, then don't
-              // extract it.
-              if (!instanceInfo->allInstancesUnderEffectiveDut(op)) {
-                ++numAnnosRemoved;
-                return true;
-              }
+              // Extract all instances in the design.
+              for (auto *i : instancePaths->instanceGraph.lookup(op)->uses()) {
```
uenoku: Early increment would be necessary
```suggestion
              for (auto *i : make_early_inc_range(..)) {
```

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
```diff
@@ -2113,47 +2085,47 @@ void GrandCentralPass::runOnOperation() {
...
+                auto instance = cast<InstanceOp>(i->getInstance());
+
+                // Skip extraction if the instance is not in the design.
+                if (instance->getParentOfType<LayerBlockOp>())
+                  continue;
+                if (!instanceInfo->allInstancesUnderEffectiveDut(
+                        i->getParent()->getModule()))
```
uenoku, line 2105: I guess if there is an instance under effective dut and one is not under dut, companion module will be kept under dut. Does that happen practically?

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: Yeah, this isn't keeping the instance graph up to date with its deletions. Same problem later when the module is deleted when in "drop" mode. This code was like this before. I will still fix it here. Good catch.

seldridge: Added this in aed1af726. I think that looks right?

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: Change in 7c1a1d72c. 

I think this, as written, should be fine as the instance graph was never updated. However, now that the instance graph is updated, this can definitely be a problem.

lib/Dialect/FIRRTL/Transforms/GrandCentral.cpp
seldridge: No, that would be very rare. Practically, this should never happen because of the dedup group annotations that Chisel emits. It is expected that the test and the design are kept separate to avoid behavior like this. The behavior here is more a stop gap of needing to do something sane if this ever occurs.

Clarifying the behavior: a parent instantiates the companion. We will _not_ extract the companion if the parent is _never_ instantiated in the effective design. Any instantiation in the effective design will force us to extract this.

There is potentially a better (though surprising) behavior of doing module duplication in this situation. However, we've generally tried to avoid this. The surprising/unpredictable nature has been viewed as unacceptable.

PR #8604, opened 2025-06-26
[FIRRTL][CreateSiFiveMetadata] Add memory read-under-write behavior to emitted metadata

fzi-hielscher:
This PR adds information on the read under write behavior of  memories, as specified by the FIRRTL/CHIRRTL IR, to the emitted memory metadata records. This includes the textual `repl-seq-mem-file`, the `seq_mem.json`, and the object model IR. It should allow external tools to generate memories with the appropriate behavior. 

- In the JSON file, there is a new string attribute `ruw_behavior` that can take one of the following values : `Undefined`, `New`, `Old`.
- In the textual file, `Undefined` behavior is elided to keep the format unchanged for the common case. Otherwise `ruw New` or `ruw Old` is appended to the memory's line.
- In the object model it is added to the `MemorySchema` as a string-typed subfield called `ruwBehavior`. I've never dealt with the object model before and I don't really know how enums should be handled here. So, please tell me if I should change my approach.

To make this information available at the point of metadata emission, I had to add the `ruw` attribute to `FMemModuleOp`. Previously, this attribute was just an `I32EnumAttr`. When adding it to the `FMemModuleOp`'s attribute dictionary this would become a generic `i32` typed attribute, which did not seem desirable to me. So I replaced the `RUWAttr` with a `RUWBehaviorAttr` dialect attribute to give it an enum specific serialization (`ruw = #firrtl<ruwbehavior Undefined>`). Unfortunately, this creates a fair amount of change noise. I can break it out as a separate PR if desired.


include/circt-c/Dialect/FIRRTL.h
```diff
@@ -292,7 +292,7 @@ MLIR_CAPI_EXPORTED MlirAttribute firrtlAttrGetParamDecl(MlirContext ctx,
 MLIR_CAPI_EXPORTED MlirAttribute firrtlAttrGetNameKind(MlirContext ctx,
                                                        FIRRTLNameKind nameKind);
 
-/// Creates a RUWAttr with the specified Read-Under-Write Behaviour.
+/// Creates a RUWBehaviorAttr with the specified Read-Under-Write Behavior.
```
fzi-hielscher, line 295: I apologise to the British, but since the FIRRTL spec is written in AE I chose 'behavior'.

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -288,35 +288,29 @@ def FMemModuleOp : FIRRTLModuleLike<"memmodule"> {
...
-           DefaultValuedAttr<LayerArrayAttr, "{}">:$layers
-      );
+  let arguments = (ins SymbolNameAttr:$sym_name, UI32Attr:$numReadPorts,
+      UI32Attr:$numWritePorts, UI32Attr:$numReadWritePorts, UI32Attr:$dataWidth,
+      UI32Attr:$maskBits, UI32Attr:$readLatency, UI32Attr:$writeLatency,
+      UI64Attr:$depth, ArrayAttr:$extraPorts,
+      DenseBoolArrayAttr:$portDirections, ArrayRefAttr:$portLocations,
+      ArrayRefAttr:$portAnnotations, ArrayRefAttr:$portSymbols,
+      ArrayRefAttr:$portNames, ArrayRefAttr:$portTypes,
+      DefaultValuedAttr<RUWBehaviorAttr, "firrtl::RUWBehavior::Undefined">:$ruw,
```
fzi-hielscher: Making `RUWBehavior::Undefined` the default reduces the impact of this change but might increase the chances of accidentally dropping specified behavior.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -566,9 +567,14 @@ CreateSiFiveMetadataPass::emitMemoryMetadata(ObjectModelIR &omir) {
...
         !isMasked ? "" : " mask_gran " + std::to_string(maskGran);
+
     seqMemConfStr = (StringRef(seqMemConfStr) + "name {{" + Twine(symId) +
                      "}} depth " + Twine(mem.getDepth()) + " width " +
-                     Twine(width) + " ports " + portStr + maskGranStr + "\n")
+                     Twine(width) + " ports " + portStr + maskGranStr +
+                     (mem.getRuw() == RUWBehavior::Undefined
+                          ? ""
+                          : " ruw " + stringifyRUWBehavior(mem.getRuw())) +
+                     "\n")
```
fzi-hielscher, line 608: This is getting a bit out of hand. When I'll also add the `MemoryInitAttr` to the metadata, I'll probably just append the string instead of concatenating twines.

lib/Dialect/FIRRTL/Transforms/LowerMemory.cpp
```diff
@@ -206,7 +206,8 @@ LowerMemoryPass::emitMemoryModule(MemOp op, const FirMemory &mem,
   auto moduleOp = b.create<FMemModuleOp>(
       mem.loc, moduleName, ports, mem.numReadPorts, mem.numWritePorts,
       mem.numReadWritePorts, mem.dataWidth, mem.maskBits, mem.readLatency,
-      mem.writeLatency, mem.depth);
+      mem.writeLatency, mem.depth,
+      *symbolizeRUWBehavior(uint32_t(mem.readUnderWrite)));
```
fzi-hielscher: Not a huge fan of this myself, but it is just the inverse of line 92: `*seq::symbolizeRUW(unsigned(op.getRuw())),`

test/firtool/memoryMetadataRenameFail.fir
```diff
@@ -116,13 +116,13 @@ circuit test:
 ; CHECK:      FILE "metadata/tb_seq_mems.json"
 ; CHECK:      [{"module_name":"memory","depth":16,"width":8,"masked":"true",
 ; CHECK-SAME: "read":"true","write":"true","readwrite":"false",
-; CHECK-SAME: "mask_granularity":8,"extra_ports":[],
+; CHECK-SAME: "ruw_behavior": "Undefined","mask_granularity":8,"extra_ports":[],
```
fzi-hielscher, line 119: I'm not really sure what this test is supposed to do. Does it just check that the metatdata is not emitted without the  --repl-seq-mem option?
At the moment it should always expectedly fail as the formatting has not been updated.

include/circt-c/Dialect/FIRRTL.h
seldridge: It's usually the Canadians you have to watch out for...

prithayan:
Thanks @fzi-hielscher for the PR, I'm also taking a look at it, will have the review by Monday.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -314,7 +315,7 @@ struct ObjectModelIR {
       extraPortsList.push_back(extraPortsObj);
     }
     auto extraPorts = builderOM.create<ListCreateOp>(
-        memorySchemaClass.getPortType(22), extraPortsList);
+        memorySchemaClass.getPortType(24), extraPortsList);
```
seldridge: Why did this bump by +2 when others only bumped by +1?

lib/Dialect/FIRRTL/Transforms/LowerMemory.cpp
seldridge: Nit?


```suggestion
      *symbolizeRUWBehavior(static_cast(mem.readUnderWrite)));
```

test/Dialect/FIRRTL/SFCTests/directories.fir
```diff
@@ -152,9 +152,8 @@ circuit TestHarness:
...
+; MLIR_OUT:    om.class.fields %name_in, %depth_in, %width_in, %maskBits_in, %readPorts_in, %writePorts_in, %readwritePorts_in, %writeLatency_in, %readLatency_in, %ruwBehavior_in, %hierarchy_in, %inDut_in, %extraPorts_in, %preExtInstName_in : !om.string, !om.integer, !om.integer, !om.integer, !om.integer, !om.integer, !om.integer, !om.integer, !om.integer, !om.string, !om.list<!om.path>, i1, !om.list<!om.class.type<@ExtraPortsMemorySchema>>, !om.list<!om.string>
```
seldridge, line 159: Ugh... this is not great to work with. Somebody (not you, likely me) needs to go in and convert all of this to `*-SAME` checks to make this more manageable.

test/Dialect/FIRRTL/emit-metadata.mlir
```diff
@@ -1086,3 +1091,92 @@ firrtl.circuit "Foo" {
...
+
+//-------------------------------------------------------------- (2) Memory JSON
+// CHECK-LABEL:         emit.file "metadata{{/|\\\\}}seq_mems.json"
+// CHECK-NEXT:            sv.verbatim "[
+// CHECK:                  \22ruw_behavior\22: \22Old\22
+
+//------------------------------------------------------- (3) Configuration File
+// CHECK-LABEL:         emit.file "mems.conf"
+// CHECK-NEXT:            sv.verbatim
+// CHECK-SAME:            ruw Old
```
seldridge, line 1226: Very nice test organization. 💯

test/firtool/memoryLowering.fir
```diff
@@ -59,4 +60,95 @@ circuit test: %[[
...
+; CHECK:     name memory_ext depth 16 width 32 ports mwrite,read mask_gran 8
+
+;// -----
+
+FIRRTL version 4.0.0
+circuit test: %[[
+  {
+    "class":"sifive.enterprise.firrtl.MarkDUTAnnotation",
+    "target": "test.test"
+  }
```
seldridge: Nit: is the `MarkDUTAnnotation` necessary?

Ditto for the second split test in this file.

test/firtool/memoryMetadataRenameFail.fir
seldridge: I'm not sure... @prithayan do you have any idea?

fzi-hielscher:
Thanks, @seldridge. It's not a problem for me to delay this a bit more to make sure all known users are prepared. I've also lined up the PR for `MemInitAttr` on top of this one. So if something needs changing, ideally that should be supported too:
https://github.com/fzi-hielscher/circt/pull/1

The script we are using internally has been forked many years ago from the rocket-chip repo:
https://github.com/chipsalliance/rocket-chip/blob/master/scripts/vlsi_mem_gen
I do not know if this is actively being used anymore.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
fzi-hielscher: Because it crashed after I initially just increased it by one. 😉 

The builder for the `ClassOp` creates both an input and an output port for each field. So the `extraPorts` port is shifted by two positions.

test/Dialect/FIRRTL/SFCTests/directories.fir
prithayan, line 159: Most likely I'm the culprit, I'll format this properly in another PR.

test/firtool/memoryMetadataRenameFail.fir
prithayan: Unfortunately, I've no recollection of why this was added. I'll get rid of it in another PR.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
prithayan: That's right, FIRRTL class, doesn't have the `fields` representation. It uses input and output ports to represent fields, hence the +2

prithayan (approved):
Looks good. Hopefully will "just work" with outer internal flow. :)

Issue #8461, opened 2025-05-02
[FIRRTL][LowerToHW] use-before-def with inline layer


seldridge:
The following fails to compile due to a use-before-def created by `LowerToHW`:

``` firrtl
FIRRTL version 5.1.0
circuit Foo :
  layer Verif, inline :

  public module Foo :
    input clock : Clock
    input reset : UInt<1>
    input enq : { flip ready : UInt<1>, valid : UInt<1>, bits : UInt<8>}
    output deq : { flip ready : UInt<1>, valid : UInt<1>, bits : UInt<8>}

    connect deq, enq
    wire valid : UInt<1>
    wire valid_rwprobe : RWProbe<UInt<1>, Verif>
    define valid_rwprobe = rwprobe(valid)
    connect valid, enq.valid
    connect deq.valid, valid
    skip
    layerblock Verif :
      regreset go_c_value : UInt<2>, clock, reset, UInt<2>(0h0)
      wire go : UInt<1>
      connect go, UInt<1>(0h0)
      when UInt<1>(0h0) :
        connect go_c_value, UInt<1>(0h0)
      else :
        when UInt<1>(0h1) :
          node go_wrap_wrap = eq(go_c_value, UInt<2>(0h3))
          node _go_wrap_value_T = add(go_c_value, UInt<1>(0h1))
          node _go_wrap_value_T_1 = tail(_go_wrap_value_T, 1)
          connect go_c_value, _go_wrap_value_T_1
          connect go, go_wrap_wrap
      when go :
        node has_been_reset = intrinsic(circt_has_been_reset : UInt<1>, clock, reset)
        node disable = eq(has_been_reset, UInt<1>(0h0))
        node _T = eq(disable, UInt<1>(0h0))
        force(clock, _T, valid_rwprobe, UInt<1>(0h0))
```

This is producing:

```
Foo.fir:19:7: error: operand #0 does not dominate this use
      regreset go_c_value : UInt<2>, clock, reset, UInt<2>(0h0)
      ^
Foo.fir:19:7: note: see current operation: %8 = "seq.firreg"(%14, %arg0, %arg1, %2) <{name = "go_c_value"}> {firrtl.random_init_start = 0 : ui64} : (i2, !seq.clock, i1, i2) -> i2
JackRWProbe.fir:28:37: note: operand defined here (op in the same block)
          node _go_wrap_value_T_1 = tail(_go_wrap_value_T, 1)
```

MLIR after failure:

``` mlir
[...]

PR #8605, opened 2025-06-26
[FIRRTL] Lower registers under ifdefs

rwy7:
In seq-to-sv, we transform seq FirRegOps to sv RegOps. This is done using a helper called FirRegLowering. This PR changes FirRegLowering to correctly lower registers under ifdef blocks.

The initialization of registers is placed in an initial block at the footer of the enclosing HWModuleOp. In order to initialize registers buried under ifdefs, we have to refer to these registers by a local XMR, because the buried registers will not dominate the initial block. We build these in a prepass over the MLIR module, returning a "PathTable" sending buried FirRegOps to their HierPathOps.

The main FirRegLowering procedure operates in parallel on each HWModuleOp. Each invocation is now handed a const reference to the path table, which allows us to produce an XMRRefOp as needed while emitting the register initialization code at the footer of the HWModule.

The initialization code for a register has to be guarded under the same ifdef conditions as the register itself. To do this, while lowering registers, we record the conditions under which the register is defined, and recreate them to guard the initialization code.

When we lower a register, we attempt to transform any input mux tree into an SV if-else tree of connects. This PR modifies the if-else generation to co-locate the if-else with the register. So if a register is guarded under an ifdef, then the if-else structure driving it will also be under the same ifdef.

Fixes: #8461

### DIFF
--- a/lib/Conversion/SeqToSV/FirRegLowering.cpp
+++ b/lib/Conversion/SeqToSV/FirRegLowering.cpp
@@ -8,6 +8,7 @@
 
 #include "FirRegLowering.h"
 #include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Support/Utils.h"
 #include "mlir/IR/Threading.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "llvm/ADT/DenseSet.h"
--- a/lib/Conversion/SeqToSV/FirRegLowering.cpp
+++ b/lib/Conversion/SeqToSV/FirRegLowering.cpp
@@ -23,6 +23,13 @@ using llvm::MapVector;
 
 #define DEBUG_TYPE "lower-seq-firreg"
 
+/// Immediately before the terminator, if present. Otherwise, the block's end.
+static Block::iterator getBlockEnd(Block *block) {
+  if (block->mightHaveTerminator())
+    return Block::iterator(block->getTerminator());
+  return block->end();
+}
+
 std::function<bool(const Operation *op)> OpUserInfo::opAllowsReachability =
     [](const Operation *op) -> bool {
   return (isa<comb::MuxOp, ArrayGetOp, ArrayCreateOp>(op));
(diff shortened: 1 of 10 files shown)


lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -107,53 +115,273 @@ void FirRegLowering::addToIfBlock(OpBuilder &builder, Value cond,
...
+      return sym;
+
+  // Otherwise, we have to create a new inner sym.
+  auto *context = reg->getContext();
+
+  auto hint = reg.getName();
+
+  // Create our new property for field 0.
+  auto sym = StringAttr::get(context, innerSymNS.newName(hint));
+  auto visibility = StringAttr::get(context, "private");
```
dtzSiFive: FWIW I believe this might be first use of visibility re:inner symbols (first that isn't default of 'public').

Probably we should remove this field until it has meaning.

Looks like some tests make use of it, but I don't think we create any that aren't "public".

Anyway, since it doesn't mean anything presently they aren't /not/ private either so... 👍 .

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -107,53 +115,273 @@ void FirRegLowering::addToIfBlock(OpBuilder &builder, Value cond,
...
+  auto property = hw::InnerSymPropertiesAttr::get(context, sym, 0, visibility);
+
+  // Build the new list of inner sym properties.
+  SmallVector<hw::InnerSymPropertiesAttr> properties;
+  if (attr)
+    llvm::append_range(properties, attr.getProps());
```
dtzSiFive, line 143: Since the new inner symbol is on lowest fieldID anyway, maybe can just append existing props after it and be done?

If there's any doubt, this is fine too 👍 .

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -107,53 +115,273 @@ void FirRegLowering::addToIfBlock(OpBuilder &builder, Value cond,
   }
 }
 
+/// Attach an inner-sym to field-id 0 of the given register, or use an existing
+/// inner-sym, if present.
+static StringAttr getInnerSymFor(InnerSymbolNamespace &innerSymNS,
```
dtzSiFive, line 120: Sorry you had to roll your own thing for this! 🙃

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "FirRegLowering.h"
 #include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/FIRRTL/FIRRTLUtils.h"
```
dtzSiFive: Is this an okay dependency? Do these conversions already depend on the FIRRTL dialect? If not, there may be places that need to be updated for that.

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -107,53 +115,273 @@ void FirRegLowering::addToIfBlock(OpBuilder &builder, Value cond,
...
+/// Locate the registers under the given HW module, which are not at the
+/// top-level of the module body. These registers will be initialized through an
+/// NLA. Put an inner symbol on each, and return a list of the buried registers
+/// and their inner-symbols.
+static std::vector<BuriedFirReg> getBuriedRegs(HWModuleOp module) {
+  auto name = SymbolTable::getSymbolName(module);
+  InnerSymbolNamespace isns(module);
+  std::vector<BuriedFirReg> result;
+  for (auto &op : *module.getBodyBlock()) {
+    // Look under one layer of operations for firregs.
```
dtzSiFive: Maybe it'll be clear later, but why a single layer?  Walk will reach down recursively, maybe we only care about the outer-most region it belongs to?

dtzSiFive: I see later we're tracking guarding conditions while recursively visiting in order to lower. 👍 .  Still curious, what's this about?

dtzSiFive (commented):
Great work!

Overall strategy seems right, thanks for putting this together!

I didn't finish going through it but left some comments mostly re:inner symbols 👍 .

lib/Conversion/SeqToSV/FirRegLowering.cpp
rwy7: Fixed, thanks!

lib/Conversion/SeqToSV/FirRegLowering.cpp
rwy7: Made it public :+1:

lib/Conversion/SeqToSV/FirRegLowering.cpp
rwy7: I included this header for the parallel transform/reduce utility defined there, not for anything related to the FIRRTL dialect. I've moved the transform/reduce helper to a common circt utility header and included that instead.

lib/Conversion/SeqToSV/FirRegLowering.cpp
rwy7: This code is finding registers that are not defined at the root of the HWModule body, but are "buried" in subregions, and attaching inner symbols to them.  These registers are initialized through an XMR reference, because they will not be in scope for the initial block at the back of the enclosing HWModule.

rwy7: I removed the comment in case it was confusing. By "look under one layer" I mean, look for ops with regions, and then look for registers under those ops.

dtzSiFive: I misread the code a bit, whoops (discussed offline in addition to your comments above), thanks!

PR #8619, opened 2025-06-28
[FIRRTL] Avoid interface cast in inject-dut-hier

seldridge:
Fix a nightly failure observed with gcc release builds due to a cast of an
interface to an operation.  This is known to be problematic.

Testing via manual trigger here: https://github.com/llvm/circt/actions/runs/15947527423

This is intended to fix this failure: https://github.com/llvm/circt/actions/runs/15943909376/job/44975510792#step:6:6782


seldridge:
This fixes nightly as shown here: https://github.com/llvm/circt/actions/runs/15948359036

PR #8620, opened 2025-06-28
[LLHD] Add loop unrolling pass

fabianschuiki:
Add an LLHD pass that unrolls loops with statically-known bounds.

The pass computes the dominance info for `llhd.combinational` ops, finds loop back-edges by looking for blocks that branch back to blocks that dominate them, which yields initial *loop head* and *back-edge* pairs. Based on these the pass populates a data structure with loop metadata, including the blocks in the loop body, exit control flow edges, the exit condition, induction variable, and loop bounds. After gathering this information, the loop body is simply stamped out muliple times, the few relevant control flow edges are adjusted, and a rough cleanup of unused blocks and values is performed.

The current implementation of the pass is *very* conservative. The loops are required to have very simple control flow, and the induction variable must go from 0 to a positive upper bound in increments of 1. We'll want to relax these constraints in the future and allow other increments/decrements, comparison predicates, and more.

Thanks again to @maerhart for doing a lot of initial work for this on his experimental branch.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/UnrollLoops.cpp
+++ b/lib/Dialect/LLHD/Transforms/UnrollLoops.cpp
@@ -1,0 +1,480 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/LLHD/IR/LLHDOps.h"
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h"
+#include "circt/Support/UnusedOpPruner.h"
+#include "mlir/Analysis/CFGLoopInfo.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/IR/Dominance.h"
+#include "mlir/IR/IRMapping.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/Pass/Pass.h"
+#include "llvm/ADT/PostOrderIterator.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "llhd-unroll-loops"
+
+namespace circt {
+namespace llhd {
+#define GEN_PASS_DEF_UNROLLLOOPSPASS
+#include "circt/Dialect/LLHD/Transforms/LLHDPasses.h.inc"
+} // namespace llhd
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+using namespace llhd;
+using llvm::SmallDenseSet;
+using llvm::SmallSetVector;
+
+//===----------------------------------------------------------------------===//
+// Utilities
+//===----------------------------------------------------------------------===//
+
+/// Clone a list of blocks into a region before the given block.
+///
+/// See `Region::cloneInto` for the original code that clones an entire region.
+static void cloneBlocks(ArrayRef<Block *> blocks, Region &region,
+                        Region::iterator before, IRMapping &mapper) {
+  // If the list is empty there is nothing to clone.
+  if (blocks.empty())
+    return;
+
+  // First clone all the blocks and block arguments and map them, but don't yet
+  // clone the operations, as they may otherwise add a use to a block that has
+  // not yet been mapped
+  SmallVector<Block *> newBlocks;
+  newBlocks.reserve(blocks.size());
+  for (auto *block : blocks) {
+    auto *newBlock = new Block();
+    mapper.map(block, newBlock);
+    for (auto arg : block->getArguments())
+      mapper.map(arg, newBlock->addArgument(arg.getType(), arg.getLoc()));
+    region.getBlocks().insert(before, newBlock);
+    newBlocks.push_back(newBlock);
+  }
+
+  // Now follow up with creating the operations, but don't yet clone their
+  // regions, nor set their operands. Setting the successors is safe as all have
+  // already been mapped. We are essentially just creating the operation results
+  // to be able to map them. Cloning the operands and region as well would lead
+  // to uses of operations not yet mapped.
+  auto cloneOptions =
+      Operation::CloneOptions::all().cloneRegions(false).cloneOperands(false);
+  for (auto [oldBlock, newBlock] : llvm::zip(blocks, newBlocks))
+    for (auto &op : *oldBlock)
+      newBlock->push_back(op.clone(mapper, cloneOptions));
+
+  // Finally now that all operation results have been mapped, set the operands
+  // and clone the regions.
+  SmallVector<Value> operands;
+  for (auto [oldBlock, newBlock] : llvm::zip(blocks, newBlocks)) {
+    for (auto [oldOp, newOp] : llvm::zip(*oldBlock, *newBlock)) {
+      operands.resize(oldOp.getNumOperands());
+      llvm::transform(
+          oldOp.getOperands(), operands.begin(),
+          [&](Value operand) { return mapper.lookupOrDefault(operand); });
+      newOp.setOperands(operands);
+      for (auto [oldRegion, newRegion] :
+           llvm::zip(oldOp.getRegions(), newOp.getRegions()))
+        oldRegion.cloneInto(&newRegion, mapper);
+    }
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// Loop Unroller
+//===----------------------------------------------------------------------===//
+
+namespace {
+/// A data structure tracking information on a single loop.
+struct Loop {
+  Loop(unsigned loopId, CFGLoop &cfgLoop) : loopId(loopId), cfgLoop(cfgLoop) {}
+  bool failMatch(const Twine &msg) const;
+  bool match();
+  void unroll(CFGLoopInfo &cfgLoopInfo);
+
+  /// A numeric identifier for debugging purposes.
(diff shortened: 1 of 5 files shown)


uenoku (commented):
Looks awesome! I'll review details Monday but I have a high-level question. Is there specific reason to unroll loops after SCF lowering? I'm wondering if we can simply unroll constant loops at SCF dialect which is substantially simpler (and there are many helpers in MLIR upstream). Reconstructing Loop from CF looks something we should avoid if possible.

fabianschuiki:
I remember @maerhart experimenting with that a little bit on his experimental branch. The problem is that Verilog has `continue`, `break`, and `return` which can exit early out of loops. So we can't directly lower to SCF in the frontend. The only alternative then would be to try to hoist CF into SCF to unroll the loops. AFAIK, that mainly creates `scf.while` loops, and loops with an early exit still wouldn't be supported :thinking:. That's why I thought about trying to unroll in CF directly.

maerhart:
@uenoku We originally retained the `for` loops in the moore dialect, but changed it to lower directly to CF to support `continue`, `break`, `return` as @fabianschuiki mentioned. The alternative would be to create our own version of SCF or change the upstream SCF to support these constructs (in that case we couldn't support frontends that have `goto` constructs, which I think is fine). That would make unrolling a lot easier (except for `while` loops), but it might add additional complexity to other passes such as deseq as it would have to take all these `for`, `break`, `continue`, `yield` ops into account instead of just `br`, and `cond_br` which we would still have to support unless we also use scf for `if` and `while`. The RegionBranchOpInterface could in theory unify all the special casing but when I used it last time a 2-3 years back it didn't work very well and was a lot more complicated than regular basic block branching, but I think there have been efforts to improve that in the meantime. Also, at the LLHD level it doesn't compose very well with the current design of the `wait` op. Of course, nothing is set in stone and we could consider moving to a more SCF based approach in the future including a redesign of the `wait` op. We could also think about retaining `for` loops just at the moore level again by adding above mentioned features and doing unrolling before lowering to LLHD.

I had experimented with the lift-control-flow pass upstream which converts any CF region to SCF, but all the operations end up in the condition region of an `scf.while` which doesn't help unrolling at all. Upstream only as utilities to unroll `scf.for` AFAIK.

In general, I think a CF unrolling pass would be a good addition to MLIR upstream, but I agree that it is relatively complicated to implement.

uenoku:
That's fair point. Unfortunate that MLIR SCF doesn't have generic control flow yet.

uenoku (commented):
Any chance we can use mlir::CFGLoopInfo for finding out loop candidates? I think some of the logic like finding a single exist and parent/child relation should work out-of-box. 

```bash
./build/bin/circt-opt foo.mlir --mlir-print-op-generic | ./build/bin/mlir-opt --allow-unregistered-dialect --pass-pipeline='builtin.module(func.func(test-cfg-loop-info))' 
Testing : "TwoNestedLoops"
Blocks : ^bb0, ^bb0, ^bb1, ^bb2, ^bb3, ^bb4, ^bb5
Loop at depth 1 containing: ^bb1<header><exiting>,^bb2,^bb4<latch>,^bb3
    Loop at depth 2 containing: ^bb2<header><exiting>,^bb3<latch>
module {
  func.func @TwoNestedLoops() -> i42 {
    %0 = "hw.constant"() <{value = 0 : i42}> : () -> i42
    %1 = "hw.constant"() <{value = 1 : i42}> : () -> i42
    %2 = "hw.constant"() <{value = 2 : i42}> : () -> i42
    %3 = "hw.constant"() <{value = 3 : i42}> : () -> i42
    %4 = "hw.constant"() <{value = 42 : i42}> : () -> i42
    cf.br ^bb1(%0, %0 : i42, i42)
  ^bb1(%5: i42, %6: i42):  // 2 preds: ^bb0, ^bb4
    %7 = "comb.icmp"(%5, %2) <{predicate = 2 : i64}> : (i42, i42) -> i1
    cf.cond_br %7, ^bb2(%0, %6 : i42, i42), ^bb5
  ^bb2(%8: i42, %9: i42):  // 2 preds: ^bb1, ^bb3
    %10 = "comb.icmp"(%8, %3) <{predicate = 2 : i64}> : (i42, i42) -> i1
    cf.cond_br %10, ^bb3, ^bb4
  ^bb3:  // pred: ^bb2
    %11 = "comb.add"(%9, %4) : (i42, i42) -> i42
    %12 = "comb.add"(%8, %1) : (i42, i42) -> i42
    cf.br ^bb2(%12, %11 : i42, i42)
  ^bb4:  // pred: ^bb2
    %13 = "comb.add"(%5, %1) : (i42, i42) -> i42
    cf.br ^bb1(%13, %9 : i42, i42)
  ^bb5:  // pred: ^bb1
    return %6 : i42
  }
}
```
Loop::match and Loop::unroll would be necessary to implement our own but other wise I think we can utilize LLVM LoopInfo.

fabianschuiki:
@uenoku That's a fantastic idea. Thanks for the pointer -- I wasn't aware of `mlir::CFGLoopInfo`. Let me try to swap that in.

fabianschuiki:
I have reworked the pass to use the `mlir::CFGLoopInfo` analysis. Works like a charm! Thanks for the pointer @uenoku 💯. We'll probably be able to relax some of the constraints in the future -- the unrolling is quite conservative right now, to simplify the first implementation.

lib/Dialect/LLHD/Transforms/UnrollLoops.cpp
```diff
@@ -0,0 +1,478 @@
...
+  beginBound = beginBoundAttr.getValue();
+
+  // Pattern match the increment operation on the induction variable.
+  if (auto addOp = indVarNext.getDefiningOp<comb::AddOp>();
+      addOp && addOp.getNumOperands() == 2) {
+    IntegerAttr incAttr;
+    if (!matchPattern(addOp.getOperand(1), m_Constant(&incAttr)))
+      return failMatch("non-constant increment");
+    indVarIncrement = incAttr.getValue();
+  } else {
```
uenoku: Is `addOp.getOperand(0) == invVar` guaranteed here?

fabianschuiki: The canonicalizer for commutative ops should move the constant to the end of the operand list. 👍

uenoku: I think currently the check is missing. This example derived from the test and `%ip` is modified. The unroll pass currently generates invalid IR.
```mlir
// CHECK-LABEL: @SimpleLoop
hw.module @SimpleLoop(in %input: i42, out x : i42) {
  %c0_i42 = hw.constant 0 : i42
  %c1_i42 = hw.constant 1 : i42
  %c3_i42 = hw.constant 3 : i42
  %c42_i42 = hw.constant 42 : i42
  // Loop of the form:
  //   x = 0
  //   for (i = 0; i < 3; ++i)
  //     x += 42
  // CHECK: llhd.combinational
  %0 = llhd.combinational -> i42 {
    // CHECK-NEXT:   cf.br [[ENTRY:\^.+]](%c0_i42 : i42)
    cf.br ^header(%c0_i42, %c0_i42 : i42, i42)
  ^header(%i: i42, %x: i42):  // 2 preds: ^bb0, ^body
    %1 = comb.icmp slt %i, %c3_i42 : i42
    cf.cond_br %1, ^body, ^exit
  ^body:  // pred: ^header
    // CHECK-NEXT: [[ENTRY]]([[X0:%.+]]: i42):
    // CHECK-NEXT:   [[X1:%.+]] = comb.add [[X0]], %c42_i42
    // CHECK-NEXT:   [[X2:%.+]] = comb.add [[X1]], %c42_i42
    // CHECK-NEXT:   [[X3:%.+]] = comb.add [[X2]], %c42_i42
    // CHECK-NEXT:   cf.br [[EXIT:\^.+]]
    %2 = comb.add %x, %c42_i42 : i42
    // %ip is not addition of indvar and constant. 
    %ip = comb.add %input, %c1_i42 : i42
    cf.br ^header(%ip, %2 : i42, i42)
  ^exit:  // pred: ^header
    // CHECK-NEXT: [[EXIT]]:
    // CHECK-NEXT:   llhd.yield [[X3]]
    llhd.yield %x : i42
  }
  hw.output %0 : i42
}
```

```
bar.mlir:2:1: error: 'hw.module' op entry block must have1 arguments to match module signature
hw.module @SimpleLoop(in %input: i42, out x : i42) {
^
bar.mlir:2:1: note: see current operation: 
"hw.module"() <{module_type = !hw.modty<input input : i42, output x : i42>, parameters = [], result_locs = [loc("bar.mlir":2:43)], sym_name = "SimpleLoop"}> ({
  %0 = "hw.constant"() <{value = 0 : i42}> : () -> i42
  %1 = "hw.constant"() <{value = 42 : i42}> : () -> i42
  %2 = "llhd.combinational"() ({
    "cf.br"(%0)[^bb1] : (i42) -> ()
  ^bb1(%3: i42):  // pred: ^bb0
    %4 = "comb.add"(%3, %1) : (i42, i42) -> i42
    %5 = "comb.add"(%4, %1) : (i42, i42) -> i42
    %6 = "comb.add"(%5, %1) : (i42, i42) -> i42
    "cf.br"()[^bb2] : () -> ()
  ^bb2:  // pred: ^bb1
    "llhd.yield"(%6) : (i42) -> ()
  }) : () -> i42
  "hw.output"(%2) : (i42) -> ()
}) : () -> ()
```

fabianschuiki: Oh yikes, you're totally right. Fantastic catch! Thanks 🚀🙇

fabianschuiki: Fixed, thanks @uenoku! :raised_hands:

uenoku (approved):
LGTM is it intentional that you removed CFGLoop?

fabianschuiki:
Yikes no, that was an accident. Let me get that back :+1:

PR #8621, opened 2025-06-28
[FIRRTL] Fix cat canonicalization to handle mixed signed/unsigned operands

### LINKED ISSUE #8618: [FIRRTL] Chisel generates FIRRTL which fails new Cat signedness check
I was trying the new `firtool-1.123.0` release, and am running into an issue with the FIRRTL generated by RocketChip, and its interaction with a new check added in https://github.com/llvm/circt/pull/8557. The code in question is here: https://github.com/llvm/circt/blob/211c3faba6aad6e2d57efe7a76d78678ab426184/lib/Dialect/FIRRTL/FIRRTLOps.cpp#L5099-L5116

I'm getting this error message:

```
generators/rocket-chip/src/main/scala/rocket/RocketCore.scala:1385:8: error: all operands must have same signedness
generators/rocket-chip/src/main/scala/rocket/RocketCore.scala:1385:8: error: 'firrtl.cat' op failed to infer returned types
generators/rocket-chip/src/main/scala/rocket/RocketCore.scala:1385:8: note: see current operation: %1737 = "firrtl.cat"(%1697, %1702, %1709, %1719, %1722, %1731, %1736) {name = "_ex_imm_T"} : (!firrtl.sint<1>, !firrtl.sint<11>, !firrtl.sint<8>, !firrtl.sint<1>, !firrtl.uint<6>, !firrtl.uint<4>, !firrtl.uint<1>) -> !firrtl.uint<32>
```

The Chisel code which causes this issue can be found here: https://github.com/chipsalliance/rocket-chip/blob/c357a0f32ecec9ee046ca1d6378c21ec7fec179b/src/main/scala/rocket/RocketCore.scala#L1369-L1387

I'm on Chisel 7.0.0-RC1+37-f8a2dec3-SNAPSHOT.

Is this a situation where Chisel is generating illegal FIRRTL? Or a situation where the new variadic form of `firrtl.cat` is interacting with a weird edge-case in FIRRTL? [The FIRRTL Spec says](https://github.com/chipsalliance/firrtl-spec/blob/b9ca6a6d1ba6946ed71663facaf0177f27088107/spec.md#L3964-L3971) that the `cat` function has two legal signatures: `(UInt, UInt) -> SInt` or `(SInt, SInt) -> UInt`. I can imagine that, since each two-operand `cat` generates a `UInt`, so long as you create the appropriate tree of `cat` operators, you can get a legal `cat` result from many arguments which have mixed signedness. For example: `cat(cat(SInt, SInt), UInt) -> cat(UInt, UInt) -> UInt`. But if you fold this into a 3-input `cat`, you get `cat(SInt, SInt, UInt)` which fails this check.

Any idea how to resolve this?

uenoku:
This commit enhances the cat operation canonicalization in FIRRTLFolds.cpp to properly handle cases where operands have mixed signed and unsigned types.

The implementation now tracks whether signed and unsigned operands exist during cat flattening. When mixed types are present, it adds explicit casting of signed values to unsigned using AsUIntPrimOp. Single operand cases are also properly cast to unsigned if needed.

Fix #8618

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLFolds.cpp
@@ -1404,24 +1404,42 @@ public:
         worklist.push_back(operand);
     };
     pushOperands(cat);
+    bool existsSigned = false, existsUnsigned = false;
     while (!worklist.empty()) {
       auto value = worklist.pop_back_val();
       auto catOp = value.getDefiningOp<CatPrimOp>();
       if (!catOp) {
         operands.push_back(value);
+        (type_isa<UIntType>(value.getType()) ? existsUnsigned : existsSigned) =
+            true;
         continue;
       }
 
       pushOperands(catOp);
     }
 
+    // Helper function to cast signed values to unsigned. CatPrimOp converts
+    // signed values to unsigned so we need to do it explicitly here.
+    auto castToUIntIfSigned = [&](Value value) -> Value {
+      if (type_isa<UIntType>(value.getType()))
+        return value;
+      return rewriter.create<AsUIntPrimOp>(value.getLoc(), value);
+    };
+
+    assert(operands.size() >= 1 && "zero width cast must be rejected");
+
     if (operands.size() == 1) {
-      rewriter.replaceOp(op, operands[0]);
+      rewriter.replaceOp(op, castToUIntIfSigned(operands[0]));
       return success();
     }
 
     if (operands.size() == cat->getNumOperands())
       return failure();
+
+    // If types are mixed, cast all operands to unsigned.
+    if (existsSigned && existsUnsigned)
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/FIRRTLFolds.cpp
```diff
@@ -1404,25 +1404,43 @@ class FlattenCat : public mlir::RewritePattern {
         worklist.push_back(operand);
     };
     pushOperands(cat);
+    bool existsSigned = false, existsUnsigned = false;
```
seldridge: Mega-nit: Using `hasSigned` and `hasUnsigned` may save enough characters that the LHS ternary assignment on line 1413 may fie on one line. 🫠

PR #8622, opened 2025-06-30
[circt-bmc] Support registers with reset signals

liuyic00:
This adds support for registers with reset signals in `ExternalizeRegistersPass`.

### DIFF
--- a/integration_test/circt-bmc/seq.mlir
+++ b/integration_test/circt-bmc/seq.mlir
@@ -52,3 +52,21 @@ hw.module @Counter(in %clk: !seq.clock, out count: i3) {
   verif.assert %lt : i1
   hw.output %reg : i3
 }
+
+// Check that reset can be triggered
+// RUN: circt-bmc %s -b 4 --module Counter_with_reset --shared-libs=%libz3 | FileCheck %s --check-prefix=REGRESET
+// REGRESET: Assertion can be violated!
+hw.module @Counter_with_reset(in %clk: !seq.clock, in %rst: i1, out count: i3) {
+  %init = seq.initial () {
+    %c1_i3 = hw.constant 1 : i3
+    seq.yield %c1_i3 : i3
+  } : () -> !seq.immutable<i3>
+  %c0_i3 = hw.constant 0 : i3
+  %c1_i3 = hw.constant 1 : i3
+  %regPlusOne = comb.add %reg, %c1_i3 : i3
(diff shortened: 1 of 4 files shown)


TaoBi22 (approved):
Hey @liuyic00, thanks for adding this!! It looks good to me - would you be able to also add a circt-bmc integration test for this functionality? Maybe something like a register with an initial value of 1 and a reset value of 0, and you check that it's possible for the register value to be 0 (obviously you'll have to make sure the time bound isn't high enough for the register to wrap round back to 0). I think after that this should be ready to merge!

liuyic00:
Thanks for the review!! The integration test has been added.

By the way, I will keep working on `circt-bmc` until it can accept Chisel-generated circuits. Please let me know if you have any suggestions since I'm new to this tool.

TaoBi22 (approved):
Integration test looks good to me, thanks! 

Very exciting that you'll be pushing for Chisel ingestion - what features do you still need that circt-bmc is currently missing? Very happy to give pointers where I can, and feel free to request reviews from me etc.!

liuyic00:
For a small example, we still need the support of:
- `seq.firreg`
- `verif.has_been_reset`
- `verif.clocked_assume`
- `verif.clocked_assert`

I will open another PR for `seq.firreg` after this one is merged.
As for the `verif.has_been_reset`, I'm considering adding a registor to record the circuit has_been_reset or not.
And then maybe adding an option to let bmc assume the reset is true at the first step (or replace initialValues). In this case, we wouldn’t need an extra register.

TaoBi22:
Cool!

Do you need async resets for the designs with firregs that you're looking at? I guess with synchronous resets you don't need anything new over compreg, but I can imagine async might be a bit more complex (that said, I could also believe that all you need to support them is another multiplexer like you do in this PR that additionally multiplexes between the register output and the reset value. The only unusual case I can think of off the top of my head is one where an async reset starts low, goes high between clock edges, then goes low again, but I think we cover that anyway in circt-bmc's default mode since it evaluates the negedge too. I suspect there'll be a corner case or two that I'm not thinking of anyway).

 A register to track whether the circuit's been reset or not sounds sensible to me.

Do you have merge permissions, or should I hit merge on this PR for you?

liuyic00:
I don’t need async resets, but you’re right! I will handle it as you said.
Processing async reset on both the posedge and negedge (in default mode) is good enough for BMC. The faster and shorter reset signal transitions are beyond the scope of step-by-step modeling method of BMC.

Yes please merge this! I dont have merge permissions.

PR #8623, opened 2025-06-30
[FIRRTL] Add "knownlayers" specifications to ExtModules

rwy7:
In a circuit, we only know about the layers which have been declared. If we
have an extmodule, it may have been defined under a different set of layers
than what is in the current circuit. In particular, if there is a layer which
we know about, but the extmodule does not, then the extmodule will not have a
bind file for us to include.

This commit adds a "knownlayer" declaration to extmodules, which is an explicit
specification of the layers which an extmodule was defined with.  For each
bound-in layer that an extmodule knows about, we can assume that there will be
a bindfile to include.

Changes:

FExtModule: Add a knownLayers property. Update the verifier so that, if a layer
is mentioned by an enablelayer specification or port type, then that layer must
be known by the exmodule. Bonus: implement verifySymbolUses for extmodules.

FIRParser: add parsing for knownlayer declarations on extmodules. For both knownlayers and enablelayers, update the parser to accept a comma-separated list of layer names, rather than a single layer name.

LowerLayersPass: Use the known-layers of extmodules while building the
bindfiles for modules in the current circuit. If a module in the current
circuit instantiates an extmodule, then that module's bindfiles will include
the bindfiles for the known-layers of the extmodule.

SpecializeLayers: Clean up known-layers when a layer is specialized away.

FirEmitter: Emit knownlayer declarations.


seldridge:
High level nit: for most internal flows, every extmodule is going to include all layers. Having to tack on 4+ `knowlayer` keywords is kind of annoying. Is there an alternative syntax that could be closer to the underlying representation, e.g., an array?

I realized that this _isn't_ the direction that `enablelayer`s went. However, maybe we want it for that as well?

``` firrtl
extmodule Foo :
  knownlayers: A, B, C.D
  ;; maybe:
  enablelayers: A, B
```

rwy7:
> Having to tack on 4+ knowlayer keywords is kind of annoying

Agreed. I'm going to make it so enablelayer/knownlayer specifications will accept multiple layer names at a time, which would be a backwards compatible change.   Although I like your suggested syntax more, I want the two syntaxes to be similar, and I don't think it is worth making a breaking change to the language.

test/Dialect/FIRRTL/parse-basic.fir
```diff
@@ -1719,6 +1723,35 @@ circuit LayerEnabledModule:
...
+  extmodule ExtModuleWithKnownLayers knownlayer A knownlayer B knownlayer B.C:
+
+  ; CHECK: firrtl.extmodule private @ExtModuleWithKnownLayers2
+  ; CHECK-SAME: knownLayers = [@A, @B, @B::@C]}
+  extmodule ExtModuleWithKnownLayers2 knownlayer A, B, B.C:
+
+  ; CHECK: firrtl.extmodule private @ExtModuleWithKnownLayerAndEnabledLayers()
+  ; CHECK-SAME: knownLayers = [@A, @B::@C]
+  ; CHECK-SAME: layers = [@A, @B]
+  extmodule ExtModuleWithKnownLayerAndEnabledLayers enablelayer A knownlayer A enablelayer B knownlayer B.C:
```
seldridge, line 1745: Good test.

test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -710,3 +710,63 @@ firrtl.circuit "Top" {
...
+// -----
+
+// Check that only known bound-in layers are included.
+firrtl.circuit "Top" {
+  firrtl.layer @Layer bind {}
+
+  firrtl.extmodule @ComponentA() attributes {knownLayers=[@Layer]}
+  firrtl.extmodule @ComponentB() attributes {knownLayers=[]}
+
+
```
seldridge: Mega-nit:


```suggestion

```

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -1621,8 +1631,6 @@ LogicalResult FExtModuleOp::verify() {
     return failure();
 
   auto params = getParameters();
-  if (params.empty())
-    return success();
```
seldridge, line 1625: What is this changing?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -1637,6 +1645,28 @@ LogicalResult FExtModuleOp::verify() {
...
+  }
+
+  SmallVector<SymbolRefAttr> missing;
+  if (!isLayerSetCompatibleWith(referenced, known, missing)) {
+    auto diag = emitOpError("references unknown layers");
+    auto &note = diag.attachNote();
+    note << "unknown layers: ";
+    interleaveComma(missing, note);
+    return failure();
+  }
```
seldridge, line 1668: This error looks great. Please include a test.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -1734,7 +1764,22 @@ LogicalResult FModuleOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
+
+  auto circuitOp = (*this)->getParentOfType<CircuitOp>();
+  for (auto layer : getKnownLayersAttr().getAsRange<SymbolRefAttr>()) {
+    if (!symbolTable.lookupSymbolIn(circuitOp, layer))
+      return emitOpError() << "knows of unknown layer '" << layer << "'";
+  }
+  for (auto layer : getLayersAttr().getAsRange<SymbolRefAttr>()) {
+    if (!symbolTable.lookupSymbolIn(circuitOp, layer))
+      return emitOpError() << "enables unknown layer '" << layer << "'";
+  }
```
seldridge: Nit: the language of this error message is a bit confusing to me. "Layer knows of unknown" and "enables unknown" feels very metaphysical. I'd opt for a more direct: "layer {layer} is undefined" or "references undefined layer {layer}".

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5220,28 +5226,93 @@ ParseResult FIRCircuitParser::parseLayerName(SymbolRefAttr &result) {
...
+    auto tokenKind = getToken().getKind();
+    // Parse an enablelayer spec.
+    if (tokenKind == FIRToken::kw_enablelayer) {
+      if (parseEnableLayerSpec(enabledLayersBuffer))
+        return failure();
+      continue;
+    }
+    // Didn't parse a layer spec.
+    break;
   }
```
seldridge, line 5241: Nit: it seems like there is a factoring of this that would allow it to be used by `parseOptionExtModuleLayerSpec`. However, it's a big complicated because that potential caller only has to deal with both enabled and optional layers and is handling the looping itself.

Not critical.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5220,28 +5226,93 @@ ParseResult FIRCircuitParser::parseLayerName(SymbolRefAttr &result) {
...
+
+  enabledLayers = ArrayAttr::get(getContext(), enabledLayersBuffer);
+  return success();
+}
+
+ParseResult
+FIRCircuitParser::parseOptionalExtModuleLayerSpec(ArrayAttr &enabledLayers,
+                                                  ArrayAttr &knownLayers) {
+  enabledLayers = nullptr;
+  knownLayers = nullptr;
```
seldridge: Would `clear` be better?

This API of I delete your input references is kind of odd. It would be more obvious to have this function only append. Though, I can't speak to the contract that the PR is trying to expose here.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5220,28 +5226,93 @@ ParseResult FIRCircuitParser::parseLayerName(SymbolRefAttr &result) {
...
+    break;
+  }
 
-  SmallVector<Attribute> layers;
+  if (enabledLayersBuffer.size() != 0)
+    if (requireFeature({4, 0, 0}, "extmodules with layers enabled"))
+      return failure();
+
+  if (knownLayersBuffer.size() != 0)
+    if (requireFeature({5, 0, 0}, "extmodules with known layers"))
```
seldridge: 5.0.0 is already released. The next unreleased minor version is: 5.2.0.

seldridge (commented):
I got most of the way through this. This is looking good!

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: If the params were empty, then it would early-return, but I want the verifier to keep going and execute the "known layer" verifier I've added below.

This early return was really unnecessary, you can see the code it was skipping right below--it's just `llvm::all_of(params, checkParmValue)`, which has negligible cost when the parameters are empty, anyways.

lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
```diff
@@ -574,6 +586,12 @@ struct SpecializeLayers {
     return nullptr;
   }
 
+  /// Specialize the known layers of an extmodule,.
```
rwy7:
```suggestion
  /// Specialize the known layers of an extmodule.
```

lib/Dialect/FIRRTL/Import/FIRParser.cpp
rwy7: Changed to 5, 2, 0; thanks!

lib/Dialect/FIRRTL/Import/FIRParser.cpp
rwy7: I'll just delete these lines

lib/Dialect/FIRRTL/Import/FIRParser.cpp
rwy7: Actually... should I be using 5.1.0, which is the nextFIRVersion, and also the exportFIRVersion? In order to test the FIRExporter changes I have made, the exportFIRVersion should allow for knownlayer specifications. Or, do we need to bump the nextFIRVersion?

https://github.com/llvm/circt/blob/main/include/circt/Dialect/FIRRTL/FIRParser.h#L127

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: Updated to "references unknown layer".

lib/Dialect/FIRRTL/Import/FIRParser.cpp
seldridge: Yeah, sorry "nextFIRVersion` is the right way to handle this (and then we lock that in as 5.2.0 once this hits the FIRRTL spec and is slated for a release).

rwy7: Using nextFIRVersion!

PR #8625, opened 2025-06-30
[FIRRTL] Fix Grand Central companion not-in-design

seldridge:
Fix a regression introduced in 88f5d5a56 where a Grand Central companion
instantiated only under the test harness would cause the companion to be
instantiated in the Grand Central directory and not under the test harness
directory.  An early exit if no instances are in the design was
incorrectly dropped.


uenoku (approved):
LGTM

PR #8628, opened 2025-07-01
[ImportVerilog] Detect and fixup two-state-exhaustive case statements

### LINKED ISSUE #8626: [ImportVerilog] Fixup case statements that look exhaustive but are not
I've seen quite a bit of code like the following out in the wild:

```systemverilog
module Foo (
  input  logic [1:0] a,
  output logic [3:0] z
);
  always_comb begin
    case (a)
      2'd0: z = 4'b0001;
      2'd1: z = 4'b0010;
      2'd2: z = 4'b0100;
      2'd3: z = 4'b1000;
    endcase
  end
endmodule
```

The intention clearly is to list all distinct values the input `a` can assume, and provide a unique output assignment for each. This case statement _looks_ exhaustive, and it also _is_ exhaustive in 2-state logic. But in SystemVerilog's 4-state `logic` type, this case-statement is non-exhaustive. If the input `a` contains an X or Z bit, none of the case items match and `z` remains unassigned, essentially modelling a latch.

Synthesis tools will often work around this by mapping the 4-state logic to 2-state logic, which makes case statements like this exhaustive. We need a similar workaround in CIRCT. For now, since the core dialects don't support 4-state logic, we can just assume that case statements that are exhaustive in 2-state logic need no else/default branch when we lower to IR.

The above produces the following IR:
```mlir
hw.module @Foo(in %a : i2, out z : i4) {
  %c0_i3 = hw.constant 0 : i3
  %true = hw.constant true
  %c0_i4 = hw.constant 0 : i4
  %c-1_i2 = hw.constant -1 : i2
  %c4_i4 = hw.constant 4 : i4
  %c-2_i2 = hw.constant -2 : i2
  %c2_i4 = hw.constant 2 : i4
  %c1_i2 = hw.constant 1 : i2
  %c1_i4 = hw.constant 1 : i4
  %c0_i2 = hw.constant 0 : i2
  %z = llhd.sig %c0_i4 : i4
  %1 = llhd.prb %z : !hw.inout<i4>
  %2 = comb.icmp ceq %a, %c0_i2 : i2
  %3 = comb.icmp ceq %a, %c1_i2 : i2
  %4 = comb.icmp ceq %a, %c-2_i2 : i2
  %5 = comb.icmp ceq %a, %c-1_i2 : i2
  %6 = comb.concat %5, %c0_i3 : i1, i3
  %7 = comb.xor %2, %true : i1
  %8 = comb.xor %3, %true : i1
  %9 = comb.and %8, %7, %4 : i1
  %10 = comb.mux %9, %c4_i4, %6 : i4
  %11 = comb.and %7, %3 : i1
  %12 = comb.mux %11, %c2_i4, %10 : i4
  %13 = comb.mux %2, %c1_i4, %12 : i4
  %14 = comb.or %2, %11, %9, %5 : i1
  llhd.drv %z, %13 after %0 if %14 : !hw.inout<i4>
  hw.output %1 : i4
}
```
The problematic part is the `if %14` on the `llhd.drv`: if none of the four case items match, the drive does not take effect. This needs to be fixed.

fabianschuiki:
Case statements often occur in the following form in real-world Verilog:

```
module Foo(input logic [1:0] a, output logic [3:0] z);
  always_comb begin
    case (a)
      2'd0: z = 4'b0001;
      2'd1: z = 4'b0010;
      2'd2: z = 4'b0100;
      2'd3: z = 4'b1000;
    endcase
  end
endmodule
```

The intention being that the case statement exhaustively covers all possible `a` input values, and assigns a corresponding `z` output value. This case statement _is_ exhaustive assuming two-state logic. However, the Verilog `logic` type is four-state, where the bits in `a` can also have the value X and Z. If `a` contains X or Z, none of the case statements would execute and `z` would not be assigned, essentially creating a latch.

In practice, logic synthesizers fold X and Z values to 0 or 1 however they see fit, often in an attempt to minimize logic. For this case statement the obvious assumption is that the input is two-state and the match is exhaustive. This does not work in simulation, though.

This commit makes ImportVerilog detect if a case statement is exhaustive assuming two-state logic. If it is, the last case item is treated as the default, being executed regardless of its case value. As a result, the above example generates a CFG where all possible paths through the case statement assign the output `z`. This then ensures that the case statement can be lowered to properly in circt-verilog, without generating conditional drives or latches.

The core dialects of CIRCT currently don't represent X and Z well, and most of ImportVerilog and the circt-verilog pipeline assumes all values are two-state. Once we add X and Z handling to the core dialects, we will have to revisit this behavior and make it opt-out.

Fixes #8626.

### DIFF
--- a/lib/Conversion/ImportVerilog/Statements.cpp
+++ b/lib/Conversion/ImportVerilog/Statements.cpp
@@ -256,7 +256,7 @@ struct StmtVisitor {
     return success();
   }
 
-  // Handle case statements.
+  /// Handle case statements.
   LogicalResult visit(const slang::ast::CaseStatement &caseStmt) {
     using slang::ast::CaseStatementCondition;
     auto caseExpr = context.convertRvalueExpression(caseStmt.expr);
@@ -267,11 +267,14 @@ struct StmtVisitor {
     // `unique0`, and `priority` modifiers which would allow for additional
     // optimizations.
     auto &exitBlock = createBlock();
+    Block *lastMatchBlock = nullptr;
+    SmallVector<moore::FVIntegerAttr> itemConsts;
 
     for (const auto &item : caseStmt.items) {
       // Create the block that will contain the main body of the expression.
       // This is where any of the comparisons will branch to if they match.
       auto &matchBlock = createBlock();
+      lastMatchBlock = &matchBlock;
 
       // The SV standard requires expressions to be checked in the order
       // specified by the user, and for the evaluation to stop as soon as the
@@ -281,6 +284,13 @@ struct StmtVisitor {
         if (!value)
           return failure();
         auto itemLoc = value.getLoc();
+
+        // Take note if the expression is a constant.
+        auto maybeConst = value;
+        if (auto defOp = maybeConst.getDefiningOp<moore::ConversionOp>())
+          maybeConst = defOp.getInput();
+        if (auto defOp = maybeConst.getDefiningOp<moore::ConstantOp>())
+          itemConsts.push_back(defOp.getValueAttr());
 
         // Generate the appropriate equality operator.
         Value cond;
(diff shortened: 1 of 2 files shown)


test/Conversion/ImportVerilog/basic.sv
```diff
@@ -2722,3 +2722,44 @@ module RangeElementSelection(
...
+    // CHECK: cf.cond_br {{%.+}}, [[CASE3:\^.+]], [[ELSE3:\^.+]]
+    // CHECK: moore.constant -8 : i4
+    // CHECK: cf.br [[EXIT]]
+    2'd3: z = 4'b1000;
+    // Instead of a default statement, branch to the final item.
+    // CHECK: [[ELSE3]]:
+    // CHECK-NOT: moore.constant -1 : i4
+    // CHECK-NOT: cf.br [[EXIT]]
+    // CHECK-NEXT: cf.br [[CASE3]]
+    default: z = 4'b1111;
```
maerhart, line 2763: Why a default case if the test is supposed to check that the exhaustive case is detected and the default is inserted automatically?

maerhart (approved):
LGTM. Kinda sad that it is so common place to apply such kind of hacks in Verilog tools/compilers.

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -325,12 +335,51 @@ struct StmtVisitor {
...
+    // case expression, but forgets to deal with cases involving X and Z bits in
+    // the input.
+    //
+    // Once the core dialects start supporting four-state values we may want to
+    // tuck this behind an import option that is on by default, since it does
+    // not preserve semantics.
+    auto twoStateExhaustive = false;
+    if (auto intType = dyn_cast<moore::IntType>(caseExpr.getType());
+        intType && intType.getWidth() < 32 &&
+        itemConsts.size() == (1 << intType.getWidth())) {
```
uenoku, line 350: Is `itemConsts` guaranteed to be unique?

fabianschuiki: I think not necessarily. If I recall correctly, you can have multiple case items match in Verilog, but only the first match as listed in the source text is executed. In that case this change wouldn't treat that case statement as exhaustive and instead just generate the regular default/no-match case in the control flow.

We might want a dedicated `moore.switch` control flow statement in the long run, especially once we need to handle X/Z bits properly. `cf.switch` doesn't really work too well there, since it uses plain old integers. There's also this batshit crazy Verilog pattern to express a one-hot match (constant in the case expr, dynamic values in the individual case items):
```systemverilog
unique case (1)
  a[0]: doStuff();
  a[1]: doStuff();
  a[2]: doStuff();
  a[3]: doStuff();
endcase
```
No idea if we'd want to capture this in the IR somehow. Verilog makes me sad.

fabianschuiki:
@maerhart Yeah this makes me sad, too :disappointed:. Especially because Verilog goes out of its way so much just to look and feel like C/C++, without realizing that its variable mutation is definitely _not_ a good model to describe hardware. This `case` thing _really_ wants to be an expression, like Rust's `match`.

PR #8629, opened 2025-07-01
[LLHD] Remove drive-only signals in Sig2Reg pass

fabianschuiki:
Extend the Sig2Reg LLHD pass to completely remove a signal and all its associated drives if the signal is only driven, but never probe or used in any other way.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/Sig2RegPass.cpp
+++ b/lib/Dialect/LLHD/Transforms/Sig2RegPass.cpp
@@ -341,6 +341,20 @@ void Sig2RegPass::runOnOperation() {
 
   for (auto sigOp :
        llvm::make_early_inc_range(moduleOp.getOps<llhd::SignalOp>())) {
+    // If the signal is only driven, but never read or used otherwise, remove
+    // it.
+    if (llvm::all_of(sigOp->getUses(), [](auto &use) {
+          return isa<llhd::DrvOp>(use.getOwner()) &&
+                 use.getOperandNumber() == 0;
+        })) {
+      LLVM_DEBUG(llvm::dbgs() << "  - Removing drive-only signal "
+                              << sigOp.getName() << "\n");
+      for (auto *user : llvm::make_early_inc_range(sigOp->getUsers()))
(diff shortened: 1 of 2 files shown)


uenoku (approved):
LGTM. Could you add this as canonicalization as well in a follow-up?

lib/Dialect/LLHD/Transforms/Sig2RegPass.cpp
```diff
@@ -341,6 +341,20 @@ void Sig2RegPass::runOnOperation() {
...
        llvm::make_early_inc_range(moduleOp.getOps<llhd::SignalOp>())) {
+    // If the signal is only driven, but never read or used otherwise, remove
+    // it.
+    if (llvm::all_of(sigOp->getUses(), [](auto &use) {
+          return isa<llhd::DrvOp>(use.getOwner()) &&
+                 use.getOperandNumber() == 0;
+        })) {
+      LLVM_DEBUG(llvm::dbgs() << "  - Removing drive-only signal "
+                              << sigOp.getName() << "\n");
+      for (auto *user : sigOp->getUsers())
```
uenoku: I think early increment is necessary

maerhart (approved):
LGTM!

> Could you add this as canonicalization as well in a follow-up?

This pass currently assumes that you don't care about tracking signals by their name afterward (e.g., in simulation). I don't think we want this by default in canonicalization. We might even want to add an option to this pass to insert `hw.wire` or similar operations to the promoted SSA values to keep the name references in the future.

lib/Dialect/LLHD/Transforms/Sig2RegPass.cpp
fabianschuiki: Nice catch, thanks! :+1:

fabianschuiki:
@uenoku @maerhart Do you remember what our stance was on canonicalizer complexity? Replacing signals with the driven value would be nice to have as a canonicalization, but I remember discussing that walking all users of an op in a canonicalizer might be too costly -- is that a "policy" we settled on?

PR #8631, opened 2025-07-01
[RTG] Add validate operation

maerhart:


include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -776,6 +776,43 @@ def YieldOp : RTGOp<"yield", [Pure, Terminator]> {
...
+    The value returned by this operation is not known during elaboration and
+    is thus treated like a value with identity (even though it might just be
+    a simple integer). Therefore, it is strongly recommended to not use the
+    result value of this operation in situations that expect structural
+    equivalence checks such as adding it to sets or bags.
+  }];
+
+  let arguments = (ins AnyType:$ref,
+                       AnyType:$defaultValue,
+                       OptionalAttr<StrAttr>:$id);
```
uenoku, line 808: I'm wondering if we can use distinct attr for id if the uniqueness matters?

maerhart: We could, but I'm not sure if there's really a benefit in doing so. I use this attribute to insert an assembly label that includes this ID which an ISA simulator can use to print a mapping for the form `<id>=<value>` to a file. In a second compilation run an can then parse this file and match the IDs with the validate ops to hardcode the simulated value. Between those two compilation runs I just store the IR in the state before lowering the validate ops to labels in a file. Using the distinct attribute would unique things while in in-memory representation, but it doesn't seem to really help me generating a unique string that I can give the ISA simulator.

In a compilation flow where the simulator can be run within a compiler pass and doesn't need the two separate compilation runs, we don't need an ID at all.

Furthermore, during elaboration a validate op may be duplicated (e.g. because it's in an `scf.for` body). In that case, we'd have to create and assign a new distinct attr. I currently keep the IDs set to nullopt until just before lowering the validate ops and do the unique name generation there.

I think there would be a benefit in using it if multiple validate ops are supposed to refer to the same ID and the user is allowed to provide a prefix/suffix for the ID. Which I'm currently not doing.

Let me know if I'm missing something.

PR #8634, opened 2025-07-01
Initialising the Datapath Dialect for Lowering Arithmetic Operators to Gates

cowardsa:
The datapath dialect will be used as part of the growing synthesis capabilities, where arithmetic operators such as comb.mul and multi-input comb.add can be lowered to gates via the datapath dialect. The datapath dialect allows us to directly express efficient circuits such as compressor trees, in the IR itself, allowing us to manipulate and combine arithmetic operations.

The key idea is to view datapath operators as generators of circuits that satisfy some contract, for example in the case of the `datapath.compress` summing it's results is equivalent to summing it's inputs. This allows us to defer implementing these critical circuits until later in the synthesis flow. 

In a simple example, we can fold a*b+c using the datapath dialect to remove a carry-propagate adder:
```mlir
%0 = comb.mul %a, %b : i4
%1 = comb.add %0, %c : i4
```
Which is equivalent to:
```mlir
%0:4 = datapath.pp %a, %b : 4 x i4
%1:2 = datapath.compress %0#0, %0#1, %0#2, %0#3, %c : 5 x i4 -> (i4, i4)
%2 = comb.add %1#0, %1#1 : i4
```

This is the first in a series of PRs to add datapath synthesis capabilities that will include the following (implemented but not merged):
1. Comb to Datapath pass
2. Datapath to Comb pass - for lowering datapath ops to comb gates
3. Datapath to SMT - for contract verification
4. Incorporating datapath passes into circt-synth
5. Replicate adds pass - to enable unsharing for better delay performance


teqdruid:
Drive-by comments:

1) Please break this PR up into two PRs: one for the boilerplate (dialect definition, etc) and one for the ops.

2) This dialect (thus far) seems to be only low-level combinational ops. Based entirely on the name, I had expected some sort of replacement for the pipeline dialect. Is this the eventual intent? If not, the name is slightly unintuitive to me. If so, you'll need to justify its existence given the pipeline dialect.

cowardsa:
Thanks for the comments @teqdruid - will split up as per suggestion.

Regarding the second question, the dialect is certainly not a replacement for the pipeline dialect. Potentially datapath and dataflow get used in many different ways. I've typically seen datapath synthesis refer to the challenge of converting arithmetic into efficient circuits. Synopsys also refers to this as a [datapath problem.](https://www.synopsys.com/dw/doc.php/wp/coding_guidelines_aug09.pdf) Particularly in an ASIC setting we can do a lot of arithmetic within a pipeline stage - opening up a lot of optimisation opportunities (like the opportunity to remove a carry-propagate adder from the critical path in the example). Happy to take other naming suggestions but the intent is to really optimise arithmetic circuits particularly leveraging efficient ASIC circuits.

cowardsa:
Will break up into separate PRs and resubmit tomorrow

teqdruid:
Great! The naming is a minor thing. If "datapath" is the standard name, I'm ok-ish with it.

PR #8636, opened 2025-07-01
[LLHD] Use initial signal value to fill gaps in CombineDrives

fabianschuiki:
If the CombineDrives pass can see and reason about all uses of a signal, use the signal's initial value to fill in any undriven parts of the signal. This works around a somewhat common pattern in input Verilog where an array signal has some of its elements completely unused -- neither read nor written:

```systemverilog
logic [1:0][31:0] a;
assign a[0] = ...;
assign ... = a[0];
// a[1] is completely unused
```

Driving previously undriven parts of the signal with their initial value is valid if we can guarantee that there are truly no drivers to those parts. In that case, adding a driver does not cause any drive conflicts, and the driven value is the same as the value already on the signal, namely the initial value.

One caveat is that this transform is only valid if the initial value of the signal is a constant. This is not always guaranteed, with the initial value potentially being provided by a constant passed in through a port. The Deseq pass faces a similar issue where real-world Verilog usage makes use of somewhat unclear semantics in initialization order. We should revisit the `llhd.sig` op at a later stage and consider using @uenoku's `seq.initial` mechanism to ensure that the value is static during runtime, or make the signal initial value an attribute.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/CombineDrives.cpp
+++ b/lib/Dialect/LLHD/Transforms/CombineDrives.cpp
@@ -28,6 +28,7 @@ using namespace circt;
 using namespace llhd;
 using hw::HWModuleOp;
 using llvm::SmallMapVector;
+using llvm::SmallSetVector;
 using llvm::SpecificBumpPtrAllocator;
 
 /// Determine the number of elements in a type. This returns the number of bits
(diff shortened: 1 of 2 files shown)


lib/Dialect/LLHD/Transforms/CombineDrives.cpp
```diff
@@ -340,6 +382,113 @@ void ModuleContext::aggregateDrives(Signal &signal) {
...
+      gapSlices.push_back(DriveSlice{DrvOp{}, Value{}, from, to - from});
+    }
+  };
+
+  // Go through the slices and keep track of the offset at which we expect the
+  // slice to start. If a slice starts beyond that offset, there is a gap which
+  // we can fill with a chunk of the signal's default value.
+  unsigned expectedOffset = 0;
+  for (auto slice : slices) {
+    fillGap(expectedOffset, slice.offset);
```
uenoku, line 421: Please early return when there is overlap.

uenoku (approved):
LGTM

lib/Dialect/LLHD/Transforms/CombineDrives.cpp
fabianschuiki: Good idea! Added 👍

PR #8639, opened 2025-07-02
[RTG] Add ValidationTypeInterface

maerhart:
This interface provides information necessary for additional static verification of the validate op and about how validate ops can be lowered for the types implementing this interface.

Two of the interface methods implemented by the integer register type are not yet tested, but will be in two future PRs that add passes making use of them (splitting it up in 3 PRs for easier reviewing).


include/circt/Dialect/RTG/IR/RTGInterfaces.td
```diff
@@ -45,4 +45,40 @@ def ContextResourceAttrInterface : AttrInterface<
...
+    as the type of the 'ref' operands of the 'rtg.validate' operation.
+  }];
+  let cppNamespace = "::circt::rtg";
+
+  let methods = [
+    InterfaceMethod<[{
+      Returns whether the specified type is a valid content type, e.g., if this
+      type is an integer register it is a handle to a bit-vector stored in that
+      register, thus an immediate of the width of the register is a valid
+      content tyep.
```
fabianschuiki:
```suggestion
      content type.
```

Issue #8640, opened 2025-07-02
[FIRRTL] Advanced layer sink will sink RWProbeOps and cause verifier errors


rwy7:
The verifier for the RWProbe op ensure that the target is under at-least the same layers as the probe op itself. The verifier provided part of the guarantee that, we cannot use a RWProbe to something in the design, from within a layer. Now, two issues:

1) this is no longer a restriction we want to enforce, so this verifier is outdated
2) advanced layer sink is sinking RWProbeOps into layer blocks, and causing verification errors

For (2), here is an example:

```firrtl
FIRRTL version 5.0.0

circuit Top:
  layer A, bind:

  public module Top:
    input  i: UInt<1>
    output o: UInt<1>
    output p: RWProbe<UInt<1>, A>

    wire   w: UInt<1>
    connect o, w
    connect w, i

    wire x : RWProbe<UInt<1>>
    define x = rwprobe(w)

    layerblock A:
      define p = x
```

IR before advanced layer sink:

```mlir
// -----// IR Dump After Vectorization (vectorization) //----- //
firrtl.module @Top(in %i: !firrtl.uint<1>, out %o: !firrtl.uint<1>, out %p: !firrtl.rwprobe<uint<1>, @A>) attributes {convention = #firrtl<convention scalarized>} {
  %w = firrtl.wire sym @sym : !firrtl.uint<1>
  firrtl.matchingconnect %o, %w : !firrtl.uint<1>
  firrtl.matchingconnect %w, %i : !firrtl.uint<1>
  %x = firrtl.wire : !firrtl.rwprobe<uint<1>>
  %0 = firrtl.ref.rwprobe <@Top::@sym> : !firrtl.rwprobe<uint<1>>
  firrtl.ref.define %x, %0 : !firrtl.rwprobe<uint<1>>
  firrtl.layerblock @A {
    %1 = firrtl.ref.cast %x : (!firrtl.rwprobe<uint<1>>) -> !firrtl.rwprobe<uint<1>, @A>
    firrtl.ref.define %p, %1 : !firrtl.rwprobe<uint<1>, @A>
  }
}
```

The error:

```text
../scratch.fir:16:16: error: 'firrtl.ref.rwprobe' op target has insufficient layer requirements
    define x = rwprobe(w)
               ^
[...]

PR #8641, opened 2025-07-02
[FIRRTL] Allow layers to RWProbe into the design

rwy7:
An old verifier for the RWPRobe op asserted that the target was under at least the same conditions as the probe op itself, ensuring that we could not create a forceable probe from within a layer, to something outside of that layer. This verifier helped ensure that "enabling a layer cannot change the behaviour of the design". We no longer have nor want this restriction, so this PR is removing the verifier.

This fixes an issue with advanced layer sink, where ALS will sink rwprobe ops into a layer and trigger this verification check.

Fixes: #8640

### DIFF
--- a/test/Dialect/FIRRTL/advanced-layer-sink.mlir
+++ b/test/Dialect/FIRRTL/advanced-layer-sink.mlir
@@ -817,3 +817,22 @@ firrtl.circuit "SinkRefs" {
     }
   }
 }
+
+firrtl.circuit "SinkRWProbe" {
+  firrtl.layer @A bind {}
+
+  firrtl.module public @SinkRWProbe(out %p: !firrtl.rwprobe<uint<1>, @A>) {
+    %w = firrtl.wire sym @sym : !firrtl.uint<1>
+    %0 = firrtl.ref.rwprobe <@SinkRWProbe::@sym> : !firrtl.rwprobe<uint<1>>
+    %1 = firrtl.ref.cast %0 : (!firrtl.rwprobe<uint<1>>) -> !firrtl.rwprobe<uint<1>, @A>
+    // CHECK:      %w = firrtl.wire sym @sym : !firrtl.uint<1>
+    // CHECK-NEXT: firrtl.layerblock @A {
+    // CHECK-NEXT:   %0 = firrtl.ref.rwprobe <@SinkRWProbe::@sym> : !firrtl.rwprobe<uint<1>>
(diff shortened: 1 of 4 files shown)


test/Dialect/FIRRTL/advanced-layer-sink.mlir
```diff
@@ -817,3 +817,22 @@ firrtl.circuit "SinkRefs" {
...
+  firrtl.module public @SinkRWProbe(out %p: !firrtl.rwprobe<uint<1>, @A>) {
+    %w = firrtl.wire sym @sym : !firrtl.uint<1>
+    %0 = firrtl.ref.rwprobe <@SinkRWProbe::@sym> : !firrtl.rwprobe<uint<1>>
+    %1 = firrtl.ref.cast %0 : (!firrtl.rwprobe<uint<1>>) -> !firrtl.rwprobe<uint<1>, @A>
+    // CHECK:      %w = firrtl.wire sym @sym : !firrtl.uint<1>
+    // CHECK-NEXT: firrtl.layerblock @A {
+    // CHECK-NEXT:   %0 = firrtl.ref.rwprobe <@SinkRWProbe::@sym> : !firrtl.rwprobe<uint<1>>
+    // CHECK-NEXT:   %1 = firrtl.ref.cast %0 : (!firrtl.rwprobe<uint<1>>) -> !firrtl.rwprobe<uint<1>, @A>
+    // CHECK-NEXT:   firrtl.ref.define %p, %1 : !firrtl.rwprobe<uint<1>, @A>
+    // CHECK-NEXT: }
```
seldridge, line 833: Not for this PR: it seems like _not_ sinking `%w` is a missed optimization here given that the only user of the wire is (now) inside the layer block. This is probably worth looking into in a follow-on.

seldridge (approved):
LGTM

test/Dialect/FIRRTL/advanced-layer-sink.mlir
rwy7: Agreed. It's the symbol from the rwprobe preventing advanced-layer-sink from sinking the op. Maybe we can revisit the use of symbols for rwprobes, or teach ALS how to find and interpret inner symbol users.

dtzSiFive: Forceable would help with this, if we aren't removing it (it's presently unused, see https://github.com/llvm/circt/pull/7138 which removes it).

PR #8643, opened 2025-07-02
[circt-verilog] Replace arith.select with comb.mux

fabianschuiki:
Add a small pass to the LLHD dialect to convert `arith.select` ops to `comb.mux` operations. (Also `arith.constant` to `hw.constant`.) The canonicalizers in the CF dialect produce `arith.select` operations, but we would like the circt-verilog pipeline to produce HW/Comb/Seq output where possible.

In the future we might want to create a full Arith-to-Comb conversion pass, similar to the Comb-to-Arith we already have. This feels like a bit of overkill to do just for the select ops, but we migth run into more of these ops at a later point.


lib/Dialect/LLHD/Transforms/SelectToMux.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace {
+struct SelectToMuxPass
+    : public llhd::impl::SelectToMuxPassBase<SelectToMuxPass> {
+  void runOnOperation() override;
+};
+} // namespace
+
+void SelectToMuxPass::runOnOperation() {
+  for (auto &op : llvm::make_early_inc_range(getOperation().getOps())) {
```
uenoku: I think it's necessary to use walk? Note that it's safe to erase the current operation in walk.

uenoku (approved):
LGTM

lib/Dialect/LLHD/Transforms/SelectToMux.cpp
fabianschuiki: Oh good point, we'll want to cover the nested stuff as well. Thanks!

maerhart:
Never used it, but is the `MapArithToComb` pass we already have not suitable for this?

fabianschuiki:
@maerhart That looks like it's doing *exactly* what I need. I hadn't realized that there are conversion passes in `lib/Transforms`. Now that I'm looking at it, it might be worthwhile moving `MapArithToComb` to `lib/Conversions` like all the other dialect translations :thinking:

fabianschuiki:
Closing this in favor of #8649.

PR #8645, opened 2025-07-02
[ExportVerilog] Add support for hw.array_inject

fabianschuiki:
Extend the PrepareForEmission part of ExportVerilog to lower `hw.array_inject` into a bit of procedural Verilog code that overwrites the injected array index:

```systemverilog
// %b = hw.array_inject %a[%i], %v
logic [N-1:0] b;
always_comb begin
  b = a;
  b[i] = v;
end
```

Unfortunately assignment patterns like `'{0: x, 1: y, 2: z}` only support constant expressions for the array indices. The only options to mutate a single array entry that I am aware of are the procedural style implement in this commit, and creating a separate mux/ternary op for each index separately, which feels like total overkill.

### DIFF
--- a/include/circt/Dialect/HW/HWVisitors.h
+++ b/include/circt/Dialect/HW/HWVisitors.h
@@ -30,6 +30,7 @@ public:
         .template Case<ConstantOp, AggregateConstantOp,
                        // Array operations
                        ArraySliceOp, ArrayCreateOp, ArrayConcatOp, ArrayGetOp,
+                       ArrayInjectOp,
                        // Struct operations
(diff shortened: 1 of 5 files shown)


uenoku (commented):
Could you add a few test cases especially legalization in procedural region with disallowLocalVariable lowering option?

fabianschuiki:
Excellent point @uenoku! I've added a few tests to make sure the expressions get pulled out of procedural regions as expected when local variables are disallowed, and that the lowering works within procedural regions as well.

uenoku (approved):
LGTM, thank you for adding tests.

PR #8646, opened 2025-07-03
[Python] Add Python bindings for AIG dialect

uenoku:
This commit adds boilerplates for Python bindings for the AIG dialect.

The implementation includes an AIG C API header with dialect registration and an AIG C API implementation with pass registration. It adds an AIG Python dialect TableGen file and Python dialect module. The AIG Python module C++ bindings are implemented using nanobind.

The implementation follows the standard pattern used by other CIRCT dialects


lib/Bindings/Python/AIGModule.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include "NanobindUtils.h"
+#include <nanobind/nanobind.h>
+namespace nb = nanobind;
+
+using namespace circt;
+using namespace mlir::python::nanobind_adaptors;
+
+/// Populate the aig python module.
+void circt::python::populateDialectAIGSubmodule(nb::module_ &m) {
+  m.doc() = "AIG dialect Python native extension";
```
maerhart, line 24: I guess you are planning to add some type/attr definitions here in a future PR?

maerhart (approved):
LGTM

lib/Bindings/Python/AIGModule.cpp
uenoku: Yes, my plan is to add a binding for LongestPathAnalysis.

PR #8647, opened 2025-07-03
[Datapath] Operator definitions and canonicalization patterns

cowardsa:
Building on the boiler plate definition adding two operators
* `datapath.compress` - compressor tree circuit
* `datapath.pp` - partial product generation circuit

The key idea is to view datapath operators as generators of circuits that satisfy some contract, for example in the case of the `datapath.compress` summing it's results is equivalent to summing it's inputs. This allows us to defer implementing these critical circuits until later in the synthesis flow.

In a simple example, we can fold a*b+c using the datapath dialect to remove a carry-propagate adder:
```mlir
%0 = comb.mul %a, %b : i4
%1 = comb.add %0, %c : i4
```
Which is equivalent to:
```mlir
%0:4 = datapath.pp %a, %b : 4 x i4
%1:2 = datapath.compress %0#0, %0#1, %0#2, %0#3, %c : 5 x i4 -> (i4, i4)
%2 = comb.add %1#0, %1#1 : i4
```
This is the first in a series of PRs to add datapath synthesis capabilities that will include the following (implemented but not merged):

* Comb to Datapath pass
* Datapath to Comb pass - for lowering datapath ops to comb gates
* Datapath to SMT - for contract verification
* Incorporating datapath passes into circt-synth
* Replicate adds pass - to enable unsharing for better delay performance


include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,72 @@ include "circt/Dialect/HW/HWTypes.td"
...
+    Example:
+    ```mlir
+    %0:2 = datapath.compress %a, %b, %c : 3 x i16 -> (i16, i16)
+    ```
+  }];
+
+  let arguments = (ins Variadic<HWIntegerType>:$inputs);
+  let results   = (outs Variadic<HWIntegerType>:$results);
+
+  let hasCustomAssemblyFormat = true;
```
uenoku: Can we use normal assembly format if possible for maintainability? I think functional-type is not that bad or you can also use [custom-directives](https://mlir.llvm.org/docs/DefiningDialects/Operations/#custom-directives) to print operand types nicely.

lib/Dialect/Datapath/DatapathFolds.cpp
```diff
@@ -0,0 +1,266 @@
...
+struct FoldCompressIntoCompress
+    : public OpRewritePattern<datapath::CompressOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  // compress(compress(a,b,c), add(e,f)) -> compress(a,b,c,e,f)
+  LogicalResult matchAndRewrite(datapath::CompressOp compOp,
+                                PatternRewriter &rewriter) const override {
+    // Get operands of the AddOp
+    auto operands = compOp.getOperands();
+    SmallVector<Value, 8> processedCompressorResults;
```
uenoku: Could you use SmallSetVector and replace `llvm::is_contained` with `SmallSetVector::contains` ? That should be as efficient as SmallVector + llvm::is_contained for small sizes and robust for large sets.

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,72 @@ include "circt/Dialect/HW/HWTypes.td"
...
+    OpBuilder<(ins "ValueRange":$lhs, "int32_t":$targetRows), [{
+      auto inputType = lhs.front().getType();
+      SmallVector<Type> resultTypes(targetRows, inputType);
+      return build($_builder, $_state, resultTypes, lhs);
+    }]>
+  ];
+}
+
+// Construct partial product array from two operands
+def PartialProductOp : DatapathOp<"pp",
```
uenoku: super nit: can we use more verbose name than `pp` such as `partial_prod` or `partial_product`?

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,72 @@ include "circt/Dialect/HW/HWTypes.td"
...
+// Construct partial product array from two operands
+def PartialProductOp : DatapathOp<"pp", 
+                                    [Pure, SameTypeOperands, 
+                                     SameOperandsAndResultType, Commutative]> {
+  let summary = "Generate partial products from multiplying the operands";
+  let description = [{
+    The first step in a multiplication is to generate partial products, which 
+    when summed, yield the product of the two operands. The partial
+    product operator does not specify an implementation, only that summing the 
+    results will yield the product of the two operands.
```
uenoku: Could you briefly describe the semantics of op here? Especially correspondence of number/bit position, column/row and operands.

lib/Dialect/Datapath/DatapathFolds.cpp
```diff
@@ -0,0 +1,266 @@
...
+    auto inputType = operands[0].getType();
+    unsigned inputWidth = inputType.getIntOrFloatBitWidth();
+
+    // TODO: implement a constant multiplication for the PartialProductOp
+
+    size_t maxNonZeroBits = 0;
+    // pp(concat(0,a), concat(0,b)) -> reduce number of results
+    for (Value operand : operands) {
+      // If the extracted bits are all known, then return the result.
+      auto knownBits = comb::computeKnownBits(operand);
```
uenoku, line 225: Not blocking but computeKnownBits is very expensive API (https://github.com/llvm/circt/pull/4690) to use in canonicalizer so we might need to revisit this later.

include/circt/Dialect/Datapath/DatapathOps.td
cowardsa: I found it really valuable to see <num-inputs> <input-type> to make it easy to count how many rows were in my compressor - will revisit if custom directives can achieve this - using functional-type involved a lot of counting (if I recall)

cowardsa:
Have now updated the datapath operator formatting to avoid custom assembly formats

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
 class DatapathOp<string mnemonic, list<Trait> traits = []> :
     Op<DatapathDialect, mnemonic, traits>;
 
+//===----------------------------------------------------------------------===//
+
+// Compress an array of bitvectors to a smaller set of bitvectors (at least 2).
```
maerhart: Nit: why not add this to the description (thus also the documentation on the website) instead?

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+def CompressOp : DatapathOp<"compress", 
+                                   [Pure, SameTypeOperands,
```
maerhart: Super-nit: this indentation looks a bit weird

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+    ```mlir
+    %0:2 = datapath.compress %a, %b, %c : 3 x i16 -> (i16, i16)
+    ```
+  }];
+
+  let arguments = (ins Variadic<HWIntegerType>:$inputs);
+  let results   = (outs Variadic<HWIntegerType>:$results);
+
+  let assemblyFormat = 
+    "$inputs attr-dict `:` custom<CompressFormat>(type($inputs), type($results))";
```
maerhart: 80 cols?

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+  let builders = [
+    OpBuilder<(ins "ValueRange":$lhs, "int32_t":$targetRows), [{
+      auto inputType = lhs.front().getType();
+      SmallVector<Type> resultTypes(targetRows, inputType);
+      return build($_builder, $_state, resultTypes, lhs);
+    }]>
+  ];
+}
+
+// Construct partial product array from two operands
```
maerhart: Nit: comment seems redundant with the `summary` and could thus be removed

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+def PartialProductOp : DatapathOp<"partial_product", 
+                                    [Pure, SameTypeOperands,
```
maerhart: Nit: weird indentation

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+    corresponds to the rows of a partial product array, which by default is
+    equal to the width of the inputs. 
+    
+    Verilog Example 4-bit multiplication:
+    ```verilog
+        partial_product[0][3:0] = {4{a[0]}} & b
+        ...
+        partial_product[3][3:0] = {4{a[3]}} & b
+        ab[3:0] = partial_product[0] + ... + partial_product[3] // = a*b
+    ```
```
maerhart: Super-nit:
```suggestion
    ```verilog
    partial_product[0][3:0] = {4{a[0]}} & b
    ...
    partial_product[3][3:0] = {4{a[3]}} & b
    ab[3:0] = partial_product[0] + ... + partial_product[3] // = a*b
    ```
```

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+    Example using `datapath` dialect:
+    ```mlir
+    %0:4 = datapath.partial_product %a, %b : 4 x i4
+    ```
+  }];
+  let arguments = (ins HWIntegerType:$multiplicand, 
+                       HWIntegerType:$multiplier);
+  let results   = (outs Variadic<HWIntegerType>:$results);
+
+  // let hasCustomAssemblyFormat = true;
```
maerhart:
```suggestion
```

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+  let assemblyFormat =
+    "$multiplicand `,` $multiplier attr-dict `:` functional-type(operands, results)";
```
maerhart: 80 cols?

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -0,0 +1,97 @@
...
+  // The compressor must reduce the number of operands by at least 1 otherwise
+  // it fails to perform any reduction.
+  if (getNumOperands() < 3)
+    return emitOpError("Requires 3 or more arguments - otherwise use add");
+
+  if (getNumResults() >= getNumOperands())
+    return emitOpError("Must reduce the number of operands by at least 1");
+
+  if (getNumResults() < 2)
+    return emitOpError("Must produce at least 2 results");
```
maerhart: I think the consensus is that these error strings shouldn't start with an uppercase letter (because it's not the start of a sentence and not even the start of the whole error message)

include/circt/Dialect/Datapath/DatapathOps.td
```diff
@@ -23,5 +23,85 @@ include "circt/Dialect/HW/HWTypes.td"
...
+  let description = [{
+    A compressor tree sums multiple bitvectors (often partial products in
+    multipliers or adders). Instead of adding all bitvectors sequentially, a 
+    compressor tree reduces the number of operands in parallel stages. The 
+    result is stored in a redundant (carry-save) representation, deferring the 
+    compressor tree implementation to a later stage.
+
+    Example:
+    ```mlir
+    %0:2 = datapath.compress %a, %b, %c : 3 x i16 -> (i16, i16)
```
maerhart: Since all operands and result types are always the same, why is this not just
```suggestion
    %0:2 = datapath.compress %a, %b, %c : i16
```
It would remove alll the redundant information and the entire custom parser and printer could go away. It would also be more consistent with how things are done in other CIRCT dialects (e.g. most `comb` ops).

test/Dialect/Datapath/errors.mlir
maerhart, line 1: In addition to there regression tests for the errors, we also have at least one test per operation (often in a file called `basic.mlir`) where circt-opt is only invoked with the round-trip option.

lib/Dialect/Datapath/DatapathFolds.cpp
```diff
@@ -0,0 +1,260 @@
...
+      return failure();
+
+    // Create a new CompressOp with all collected operands
+    rewriter.replaceOpWithNewOp<datapath::CompressOp>(
+        compOp, newCompressOperands, compOp.getNumResults());
+    return success();
+  }
+};
+
+struct FoldAddIntoCompress : public OpRewritePattern<comb::AddOp> {
```
maerhart, line 96: Nit: it often helps readability a lot to add a typical example of an application of a canonicalization to the docsting of the pattern struct.

include/circt/Dialect/Datapath/DatapathOps.td
cowardsa: My ideal format would be 
`%0:2 = datapath.compress %a, %b, %c : 3 x i16`

Reasoning - when these compressors become large - it is valuable to be able to quickly read off how many rows the compressor is summing - without this annotation you need to count the number of arguments

Problems: have had great difficulties understanding how assemblyFormat works... So if anyone is able to help achieve the above that would be super helpful?

maerhart: I'd argue that you can easily use your editor/IDE (e.g., vim) to give you this information by counting the number of `%` minus 1 in that line instead of annotating these things in the IR.

If you really want this format, you can significantly reduce what the custom directive is doing. It should just print `num-operands x` and in the parser it should parse a number and `x` and just throw it away as it is redundant anyway. The operands and type should be part of the `assemblyFormat` not the directive and with the right traits (the ones you already added should be enough) it will automatically infer the other operands and all result types.

cowardsa: @maerhart - do you have an example of how to print type($inputs[0[) as the comb example just uses the result - which in this case we have variadic inputs and outputs?

maerhart: Ah sorry! This is trickier than I thought. One way I have seen people work around this is to define one extra operand or result outside of the variadic and use that as the anchor for the type but that leads to other annoyances, so I'd probably rather avoid that and pay the cost of the custom parser/printer.

maerhart: I think there's also no way to get the number of results from the `%0:2 =` prefix (which I guess is the reason you kept the `(i16, i16)`). So I guess there are two options:
1. Custom parser/printer for something like `%0:2 = datapath.compress %a, %b, %c : 2 x i16`
2. Use the ODS `functional-type`

cowardsa: Ok how about the following that just involves one custom-directive and prints the type only once - I accept there is replication in the 2 but given the additional code required to get %0:2 this is perhaps more maintainable?

```mlir
%0:2 = datapath.compress %a, %b, %c : i16 [3 -> 2]
```

lib/Dialect/Datapath/DatapathFolds.cpp
cowardsa: Have added comments of basic examples and where we may have mutliple folds e.g. in Compress constant fold - have mimicked the CombFolds commenting style

cowardsa:
Believe I've addressed all the comments above now - please let me know if any further concerns @maerhart or @uenoku?

lib/Dialect/Datapath/DatapathFolds.cpp
```diff
@@ -0,0 +1,261 @@
...
+      }
+
+      // Regular operand - just add it to our list
+      shouldFold |= !newCompressOperands.empty();
+      newCompressOperands.push_back(operand);
+    }
+
+    // Only fold if we have constructed a larger compressor than what was
+    // already there
+    if (!(shouldFold))
```
uenoku:
```suggestion
    if (!shouldFold)
```

test/Dialect/Datapath/basic.mlir
```diff
@@ -0,0 +1,15 @@
...
+  %0:2 = datapath.compress %a, %b, %c : i4 [3 -> 2]
+  hw.output %0#0, %0#1 : i4, i4
+}
+
+// CHECK-LABEL: @partial_product
+hw.module @partial_product(in %a : i3, in %b : i3, out pp0 : i3, out pp1 : i3, out pp2 : i3) {
+  // CHECK-NEXT: datapath.partial_product %a, %b : (i3, i3) -> (i3, i3, i3)
+  %0:3 = datapath.partial_product %a, %b : (i3, i3) -> (i3, i3, i3)
+  hw.output %0#0, %0#1, %0#2 : i3, i3, i3
+}
```
uenoku: nit: insert new line
```suggestion
}

```

test/Dialect/Datapath/basic.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: circt-opt %s | circt-opt | FileCheck %s
```
uenoku: nit: You ca ncheck with -verify-roundtrip
```suggestion
// RUN: circt-opt %s -verify-roundtrip| FileCheck %s
```

lib/Dialect/Datapath/DatapathFolds.cpp
```diff
@@ -0,0 +1,261 @@
...
+      if (size - 1 == op.getNumResults()) {
+        rewriter.replaceOp(op, inputs.drop_back());
+        return success();
+      }
+
+      // Default create a compressor with fewer arguments
+      auto newCompressOp = rewriter.create<CompressOp>(
+          op.getLoc(), inputs.drop_back(), op.getNumResults());
+
+      rewriter.replaceOp(op, newCompressOp.getResults());
```
uenoku: Does this work?
```suggestion
  rewriter.replaceOpWithNewOp<CompressOp>(
          op, inputs.drop_back(), op.getNumResults());
```

lib/Dialect/Datapath/DatapathFolds.cpp
```diff
@@ -0,0 +1,261 @@
...
+        op.getLoc(), op.getOperands(), maxNonZeroBits);
+
+    auto zero = rewriter.create<hw::ConstantOp>(op.getLoc(),
+                                                APInt::getZero(inputWidth));
+
+    // Collect newPP results and pad with zeros if needed
+    SmallVector<Value> newResults(newPP.getResults().begin(),
+                                  newPP.getResults().end());
+    while (newResults.size() < op.getNumResults())
+      newResults.push_back(zero);
```
uenoku:
```suggestion
      newResults.append(op.getNumResults() - newResults.size(), zero);
```

lib/Dialect/Datapath/DatapathFolds.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+  results
+      .add<FoldCompressIntoCompress, FoldAddIntoCompress, ConstantFoldCompress>(
+          context);
+}
+
+//===----------------------------------------------------------------------===//
+// Partial Product Operation
+//===----------------------------------------------------------------------===//
+struct ConstantFoldPartialProduct : public OpRewritePattern<PartialProductOp> {
```
uenoku: nit: This canonicalization pattern looks like width narrowing. MLIR has folder API for constant folding so can we  rename this pattern to avoid confusion?

uenoku (approved):
LGTM other than several nits :+1:

lib/Dialect/Datapath/DatapathFolds.cpp
cowardsa: Thanks for this - was looking for a neater way to do this!

cowardsa:
Have now addressed the nits and still awaiting commit access approval so will need someone else to merge for me if satisfied please?

PR #8648, opened 2025-07-03
[RTG] Add EmbedValidationValuesPass

maerhart:


lib/Dialect/RTG/Transforms/EmbedValidationValuesPass.cpp
```diff
@@ -0,0 +1,128 @@
...
+  void runOnOperation() override;
+  LogicalResult parseFile(DenseMap<StringAttr, rtg::ValidateOp> &opById,
+                          DenseMap<StringAttr, TypedAttr> &valueMap);
+};
+} // namespace
+
+LogicalResult EmbedValidationValuesPass::parseFile(
+    DenseMap<StringAttr, rtg::ValidateOp> &opById,
+    DenseMap<StringAttr, TypedAttr> &valueMap) {
+  std::ifstream file(filename);
```
uenoku: Is there specific reason to use ifstream instead of llvm::MemoryBuffer? I think you can simplify line parsing with line_iterator like https://github.com/llvm/llvm-project/blob/4cd9c894d1ef27c26a825be5dac2824b7c30659b/llvm/utils/split-file/split-file.cpp#L75

lib/Dialect/RTGTest/IR/RTGTestTypes.cpp
```diff
@@ -33,12 +33,15 @@ bool IntegerRegisterType::isValidContentType(Type type) const {
...
-Attribute IntegerRegisterType::parseContentValue(llvm::StringRef valueString,
+TypedAttr IntegerRegisterType::parseContentValue(llvm::StringRef valueString,
                                                  Type contentType) const {
   auto immType = dyn_cast<rtg::ImmediateType>(contentType);
   if (!immType)
     return {};
 
+  if (valueString.starts_with("0x"))
+    valueString = valueString.drop_front(2);
+
```
uenoku: Why is this necessary? Isn't getAsIntegr automatically dropping 0x?

maerhart: I thought that as well, but it didn't work. Just looked at the implementation in StringRef and apparently it does not check these kind of prefixes when specifying the radix, but when invoking it with radix=0 is checks for those prefixes and does the right thing, so I'll change it to that for more generality. Thanks for pointing out!

uenoku: Oh I see. I also thought radix=16 would automatically drop 0x. Good to know!

uenoku (approved):
LGTM other than few nits.

fabianschuiki (approved):
LGTM!

PR #8652, opened 2025-07-04
[AIG][CAPI] Add C API for LongestPathAnalysis

uenoku:
This commit introduces C API bindings for AIG LongestPathAnalysis and LongestPathCollection, enabling longest path analysis of AIG circuits from C and other languages.

The API uses JSON serialization for path data exchange, providing a stable interface while the underlying data structures evolve. Paths are automatically sorted by delay in descending order for efficient critical path analysis.

### DIFF
--- a/lib/CAPI/Dialect/AIG.cpp
+++ b/lib/CAPI/Dialect/AIG.cpp
@@ -9,10 +9,129 @@
 #include "circt-c/Dialect/AIG.h"
 #include "circt/Dialect/AIG/AIGDialect.h"
 #include "circt/Dialect/AIG/AIGPasses.h"
+#include "circt/Dialect/AIG/Analysis/LongestPathAnalysis.h"
+#include "mlir-c/BuiltinAttributes.h"
+#include "mlir-c/IR.h"
+#include "mlir-c/Support.h"
 #include "mlir/CAPI/IR.h"
 #include "mlir/CAPI/Registration.h"
 #include "mlir/CAPI/Support.h"
+#include "mlir/Pass/AnalysisManager.h"
+#include "llvm/Support/JSON.h"
+
+using namespace circt;
+using namespace circt::aig;
 
 MLIR_DEFINE_CAPI_DIALECT_REGISTRATION(AIG, aig, circt::aig::AIGDialect)
 
 void registerAIGPasses() { circt::aig::registerPasses(); }
+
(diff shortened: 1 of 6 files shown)


test/CAPI/aig.c
```diff
@@ -0,0 +1,177 @@
+//===- aig.c - Test for AIG C API ----------------------------------------===//
```
maerhart:
```suggestion
//===---------------------------------------------------------------------===//
```

test/CAPI/aig.c
```diff
@@ -0,0 +1,177 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+void testAIGDialectRegistration(void) {
+  MlirContext ctx = mlirContextCreate();
+  mlirDialectHandleRegisterDialect(mlirGetDialectHandle__aig__(), ctx);
+
+  MlirDialect aig =
+      mlirContextGetOrLoadDialect(ctx, mlirStringRefCreateFromCString("aig"));
+  assert(!mlirDialectIsNull(aig));
```
maerhart: Could we use printf directly instead of relying on assertions? Assertions can be turned off with the right compiler flags and the test would pass trivially. Printf in the success case is more robust.
```suggestion
  if (!mlirDialectIsNull(aig));
    printf("success message");
```

lib/CAPI/Dialect/AIG.cpp
```diff
@@ -9,10 +9,129 @@
...
+}
+
+static inline AIGLongestPathCollection wrap(LongestPathCollection *collection) {
+  return AIGLongestPathCollection{collection};
+}
+
+static inline LongestPathCollection *
+unwrap(AIGLongestPathCollection collection) {
+  return static_cast<LongestPathCollection *>(collection.ptr);
+}
```
maerhart: Why isn't it possible to use the helper macro? https://github.com/llvm/llvm-project/blob/c7d3b818963aede5fc80a3246fb9af3f21f8eadd/mlir/include/mlir/CAPI/Wrap.h#L25

maerhart (approved):
LGTM

lib/CAPI/Dialect/AIG.cpp
uenoku: Didn't know that macro, thank you for pointing out!

PR #8654, opened 2025-07-07
[MooreToCore] Lower exponentiation to math.ipowi (PowUOpConversion)

liamslj13:
This pull request addresses part of issue #8476, and is related to PR #8574. Similar to the signed integer conversion, I have removed the SCF dialect for-loop logic and replaced it directly with `math.ipowi` from MLIR's Math dialect. 

Initially, it was suggested to zero extend the inputs (would this require `mlir.llvm` dialect?), and to ensure this conversion is secure I can reevaluate my changes to reflect this suggestion. However, I submit this PR now due to tests currently passing in this state.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -1325,29 +1325,8 @@ struct PowUOpConversion : public OpConversionPattern<PowUOp> {
                   ConversionPatternRewriter &rewriter) const override {
     Type resultType = typeConverter->convertType(op.getResult().getType());
 
-    Location loc = op.getLoc();
-    auto intType = cast<IntType>(op.getRhs().getType());
-
-    // transform a ** b into scf.for 0 to b step 1 { init *= a }, init = 1
-    Type integerType = rewriter.getIntegerType(intType.getWidth());
-    Value lowerBound = rewriter.create<hw::ConstantOp>(loc, integerType, 0);
-    Value upperBound =
-        rewriter.create<ConversionOp>(loc, integerType, op.getRhs());
-    Value step = rewriter.create<hw::ConstantOp>(loc, integerType, 1);
-
-    Value initVal = rewriter.create<hw::ConstantOp>(loc, resultType, 1);
-    Value lhsVal = rewriter.create<ConversionOp>(loc, resultType, op.getLhs());
-
-    auto forOp = rewriter.create<scf::ForOp>(
-        loc, lowerBound, upperBound, step, ValueRange(initVal),
-        [&](OpBuilder &builder, Location loc, Value i, ValueRange iterArgs) {
-          Value loopVar = iterArgs.front();
-          Value mul = rewriter.create<comb::MulOp>(loc, lhsVal, loopVar);
-          rewriter.create<scf::YieldOp>(loc, ValueRange(mul));
-        });
-
-    rewriter.replaceOp(op, forOp.getResult(0));
-
+    rewriter.replaceOpWithNewOp<mlir::math::IPowIOp>(
+        op, resultType, adaptor.getLhs(), adaptor.getRhs());
     return success();
   }
 };
--- a/test/Conversion/MooreToCore/basic.mlir
+++ b/test/Conversion/MooreToCore/basic.mlir
@@ -1111,11 +1111,9 @@ func.func @Conversions(%arg0: !moore.i16, %arg1: !moore.l16) {
 
 // CHECK-LABEL: func.func @PowUOp
 func.func @PowUOp(%arg0: !moore.l32, %arg1: !moore.l32) {
-  // CHECK: %{{.*}} = scf.for %{{.*}} = %{{.*}} to %arg1 step %{{.*}} iter_args([[VAR:%.+]] = %{{.*}}) -> (i32)  : i32 {
-  // CHECK: [[MUL:%.+]] = comb.mul %arg0, [[VAR]] : i32
-  // CHECK: scf.yield [[MUL]] : i32
-  %0 = moore.powu %arg0, %arg1 : l32
-  return
+  // CHECK: %[[RES:.*]] = math.ipowi %arg0, %arg1 : i32
+    %0 = moore.powu %arg0, %arg1 : l32
+    return
 }
 
 // CHECK-LABEL: func.func @PowSOp


test/Conversion/MooreToCore/basic.mlir
```diff
@@ -1111,11 +1111,9 @@ func.func @Conversions(%arg0: !moore.i16, %arg1: !moore.l16) {
...
 // CHECK-LABEL: func.func @PowUOp
 func.func @PowUOp(%arg0: !moore.l32, %arg1: !moore.l32) {
-  // CHECK: %{{.*}} = scf.for %{{.*}} = %{{.*}} to %arg1 step %{{.*}} iter_args([[VAR:%.+]] = %{{.*}}) -> (i32)  : i32 {
-  // CHECK: [[MUL:%.+]] = comb.mul %arg0, [[VAR]] : i32
-  // CHECK: scf.yield [[MUL]] : i32
-  %0 = moore.powu %arg0, %arg1 : l32
-  return
+  // CHECK: %[[RES:.*]] = math.ipowi %arg0, %arg1 : i32
+    %0 = moore.powu %arg0, %arg1 : l32
+    return
```
fabianschuiki: Nit: indentation
```suggestion
  %0 = moore.powu %arg0, %arg1 : l32
  return
```

liamslj13:
Thank you for the thorough explanation! This makes things quite a bit clearer, and I was able to get some work done on it today. I had some issues with types but after looking through some docs and the previous implementation of PowSOpConversion I was able to scrap together some code which I will push shortly.

Just a question or two I have about some of the code that I wasn't fully able to understand! In lines 
```cpp
auto lhs = rewriter.create<comb::ConcatOp>(loc, zero, adaptor.getLhs());
auto rhs = rewriter.create<comb::ConcatOp>(loc, zero, adaptor.getRhs());
```
does `comb::ConcatOp` concatenate to the left or right side of the unsigned number (this might not actually matter)? Also would using a different dialect such as `mlir::llvm` not be viable here with its built in `mlir::llvm::zext` function? This might be overkill or add some unneeded bulk, but I will try anything that could make this work nicely 👍

Just maybe another small question as well; are both input integers guaranteed to be of the same type width, or would you recommend I look for length of the bigger one and convert the other side to the same width?

Anyways, if these changes look good to you then I can review the test in `basic.mlir` and try to learn more about the IR syntax as a whole. The current tests will not pass so I will work on writing some that do. Sorry about the indentation issue as well, should be fixed now.

fabianschuiki:
> Just a question or two I have about some of the code that I wasn't fully able to understand! In lines [...] does `comb::ConcatOp` concatenate to the left or right side of the unsigned number (this might not actually matter)?

The `comb.concat` operands go from most significant bit to least significant bit. So the first operands will land at the highest bits of the result (left-most if you write the number down), and the last operands will land at the lowest bits of the result (right-most if you write the number down). This tries to reflect how Verilog concatenation works (`{a, b}`), but is a bit unfortunate and has caused a lot of pain and bugs in various parts of CIRCT :see_no_evil:.

> Also would using a different dialect such as `mlir::llvm` not be viable here with its built in `mlir::llvm::zext` function?

From an MLIR perspective, and when you want to do something like simulate the design, that would definitely work. The `MooreToCore` conversion tries to only use operations in the HW, Comb, Seq, LLHD, Sim, Verif, and LTL dialects, though. So producing an LLVM dialect operation would make it harder for other parts of CIRCT to then process the IR further.

> Just maybe another small question as well; are both input integers guaranteed to be of the same type width, or would you recommend I look for length of the bigger one and convert the other side to the same width?

Good point, and yes, they are! The `PowUOp` is based on `BinaryOpBase`, which has the `SameOperandsAndResultType` trait: https://github.com/llvm/circt/blob/56625bf8112d4b3d53569dba1a017ff83f57a67d/include/circt/Dialect/Moore/MooreOps.td#L722-L723

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1325,29 +1325,25 @@ struct PowUOpConversion : public OpConversionPattern<PowUOp> {
...
+    auto extendedIntType = IntegerType::get(op.getContext(), originalWidth + 1);
 
-    auto forOp = rewriter.create<scf::ForOp>(
-        loc, lowerBound, upperBound, step, ValueRange(initVal),
-        [&](OpBuilder &builder, Location loc, Value i, ValueRange iterArgs) {
-          Value loopVar = iterArgs.front();
-          Value mul = rewriter.create<comb::MulOp>(loc, lhsVal, loopVar);
-          rewriter.create<scf::YieldOp>(loc, ValueRange(mul));
-        });
+    Value zeroVal = rewriter.create<hw::ConstantOp>(loc, extendedIntType, 0);
```
fabianschuiki: I think you only need a single bit zero value here. In the concat you want to add a single 0 bit at the top of the LHS and RHS. No need to compute `originalWidth + 1` here, instead, you can just create a 1-bit constant with value 0. `ConstantOp` has a convenient shorthand (I think :grimacing:):
```suggestion
    Value zeroVal = rewriter.create<hw::ConstantOp>(loc, APInt(1, 0));
```
Then you also no longer need `originalWidth` and `extendedIntType`, because all of the types are automatically inferred by the `create<...>` function calls below.

liamslj13:
> The comb.concat operands go from most significant bit to least significant bit. So the first operands will land at the highest bits of the result (left-most if you write the number down), and the last operands will land at the lowest bits of the result (right-most if you write the number down). This tries to reflect how Verilog concatenation works ({a, b}), but is a bit unfortunate and has caused a lot of pain and bugs in various parts of CIRCT 🙈.

Alright, thank you for clearing this up! I don't exactly understand SystemVerilog/Verilog or other verification methods very well yet so please forgive me lol, the last part may have gone over my head. 

I will keep a note on the dialect usage as well so I know what is permissible to use unless specified otherwise for next time.

I have added the suggestions you have given me; thank you as always! I also rewrote the test in `basic.mlir`. The FileCheck variable(?) names or other conventions may be off, but I tried to mimic what other tests (bigger than PowSOp) were doing. Everything should run okay.

fabianschuiki (approved):
Very cool, this looks great! Thanks a lot for all your work :rocket:. If it's okay with you, I'll merge this if CI is happy.

liamslj13:
> Very cool, this looks great! Thanks a lot for all your work 🚀. If it's okay with you, I'll merge this if CI is happy.

All good with me 👍 Thank you for your patience with me and the explanations on the codebase. Glad it was all able to get done smoothly!

PR #8655, opened 2025-07-07
[seq] Simplify clock enabled when constant enabled.

jpienaar:
If always enabled, it seems equivalent/always better to just use compreg op.

I wasn't sure if just returning input is the expected behavior when enable always false.


seldridge (approved):
Makes sense to me. Please include a test of this here: https://github.com/llvm/circt/blob/main/test/Dialect/Seq/canonicalization.mlir

PR #8656, opened 2025-07-07
[ImportVerilog] Add support for SVA declarations

towoe:
Support 'sequence' and 'property' declarations.
Slang already handles the references and variable ports. Make visiting of 'AssertionInstance' legal.
Those are part of 'Expressions' of the 'Simple' case. Adapt the visitor to handle the different return types.

### DIFF
--- a/lib/Conversion/ImportVerilog/AssertionExpr.cpp
+++ b/lib/Conversion/ImportVerilog/AssertionExpr.cpp
@@ -78,18 +78,22 @@ struct AssertionExprVisitor {
   }
 
   Value visit(const slang::ast::SimpleAssertionExpr &expr) {
-    // A Simple Assertion expression has the following members:
-    // - Expression
-    // - Optional repetition
-
-    // The expression needs to have the type `i1` for the LTL operations.
-    // Convert first to a moore type and then add a conversion to `i1`.
+    // Handle expression
     auto value = context.convertRvalueExpression(expr.expr);
+    if (!value)
+      return {};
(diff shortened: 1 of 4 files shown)


towoe:
One benefit of supporting the symbols would be that the named properties are then inlined.
If there is a proper path back to SV, this could provide a way to convert the named properties for other tools, e.g. Verilator.

lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -78,18 +78,22 @@ struct AssertionExprVisitor {
...
     auto value = context.convertRvalueExpression(expr.expr);
+    if (!value)
+      return {};
     auto loc = context.convertLocation(expr.expr.sourceRange);
-    value = context.convertToI1(value);
+    auto valueType = value.getType();
+    // For assertion instances the value is already the expected type, convert
+    // boolean value
+    if (!(mlir::isa<ltl::SequenceType>(valueType) ||
+          mlir::isa<ltl::PropertyType>(valueType))) {
```
fabianschuiki: You can check for one of multiple types in one go:
```suggestion
    if (!mlir::isa<ltl::SequenceType, ltl::PropertyType>(valueType)) {
```

towoe:
Thanks for the feedback @fabianschuiki!
Yeah, doing more optimization in the IR could be quite useful. Hopefully that is not too much technical debt to get to later :)
If you could push it, that would be great. Thanks!

PR #8659, opened 2025-07-07
[AIG][Python] Add Python bindings for LongestPathAnalysis

uenoku:
This commit adds comprehensive Python bindings for AIG timing analysis, providing a high-level interface for critical path analysis

- LongestPathAnalysis class for performing timing analysis on AIG circuits
- LongestPathCollection with efficient caching and statistical methods
- Dataclasses (DataflowPath, Object, DebugPoint, OpenPath) for structured path representation with full type annotations
- Flamegraph generation for visualization
- Statistical analysis methods (percentiles, delay distribution)

The Python API builds on the C API, providing JSON deserialization into structured Python objects with caching for performance. The flamegraph output format enables visualization of critical timing paths through the circuit hierarchy.

Usage example:
```python
  analysis = LongestPathAnalysis(module)
  collection = analysis.get_all_paths("module_name")
  collection.print_summary()
  print(collection.longest_path.to_flamegraph())
````

### DIFF
--- a/integration_test/Bindings/Python/dialects/aig.py
+++ b/integration_test/Bindings/Python/dialects/aig.py
@@ -4,7 +4,7 @@
 import circt
 from circt.dialects import aig, hw
 from circt.ir import Context, Location, Module, InsertionPoint, IntegerType
-from circt.dialects.aig import LongestPathAnalysis, LongestPathCollection
+from circt.dialects.aig import LongestPathAnalysis, LongestPathCollection, DataflowPath
 
 with Context() as ctx, Location.unknown():
   circt.register_dialects(ctx)
@@ -56,6 +56,14 @@ with Context() as ctx, Location.unknown():
     # CHECK-NEXT: index -1 delay: 1
     print("index 1 delay:", collection[1].delay)
     print("index -1 delay:", collection[-1].delay)
+    # CHECK-NEXT: collection.get_path(5) == collection[5]: True
+    print(
+        "collection.get_path(5) == collection[5]:",
+        DataflowPath.from_json_string(
+            collection.collection.get_path(5)) == collection[5])
+    # Check that len and get_size are the same
+    # CHECK-NEXT: 96 96
+    print(len(collection), collection.collection.get_size())
 
     try:
       print(collection[96])
--- a/lib/Bindings/Python/AIGModule.cpp
+++ b/lib/Bindings/Python/AIGModule.cpp
@@ -47,15 +47,14 @@ void circt::python::populateDialectAIGSubmodule(nb::module_ &m) {
              MlirStringRef moduleNameRef =
                  mlirStringRefCreateFromCString(moduleName.c_str());
 
-             if (aigLongestPathCollectionIsNull(
-                     aigLongestPathAnalysisGetAllPaths(*self, moduleNameRef,
-                                                       elaboratePaths)))
-               throw nb::value_error(
-                   "Failed to get all paths, see previous error(s).");
-
              auto collection =
                  AIGLongestPathCollection(aigLongestPathAnalysisGetAllPaths(
                      *self, moduleNameRef, elaboratePaths));
+
+             if (aigLongestPathCollectionIsNull(collection))
+               throw nb::value_error(
+                   "Failed to get all paths, see previous error(s).");
+
              return collection;
            });
 
--- a/lib/Bindings/Python/dialects/aig.py
+++ b/lib/Bindings/Python/dialects/aig.py
@@ -334,6 +334,7 @@ class LongestPathCollection:
         Returns:
             DataflowPath at the specified position ratio
         """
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: integration_test/Bindings/Python/dialects/aig.py [hunk 1]
```diff
@@ -4,29 +4,75 @@
...
     # Test aig.and_inv operation
-    # CHECK: %{{.*}} = aig.and_inv %a, not %b : i1
+    # CHECK: %[[RESULT1:.*]] = aig.and_inv %a, not %b : i32
+    # CHECK: %{{.*}} = aig.and_inv not %a, %[[RESULT1]] : i32
     print(m)
+
+    # Test LongestPathAnalysis Python bindings
+    analysis = LongestPathAnalysis(m.operation)
+
+    collection = analysis.get_all_paths("test_aig")
```
cowardsa, line 40: Nit: a test that get_path(index) == collection[index]?

integration_test/Bindings/Python/dialects/aig.py
```diff
@@ -4,29 +4,75 @@
...
+    # CHECK-NEXT: 99.9th percentile delay: 2
+    collection.print_summary()
+
+    # Check index.
+    # CHECK-NEXT: index 1 delay: 2
+    # CHECK-NEXT: index -1 delay: 1
+    print("index 1 delay:", collection[1].delay)
+    print("index -1 delay:", collection[-1].delay)
+
+    try:
```
cowardsa, line 68: Nit: a test that size of collections == get_size?

lib/Bindings/Python/dialects/aig.py
```diff
@@ -2,5 +2,387 @@
...
+        Since paths are sorted by delay in descending order, higher ratios
+        correspond to paths with higher delays (closer to the critical path).
+        Args:
+            ratio: Position ratio between 0.0 and 1.0
+                  (e.g., 1.0 = longest delay path, 0.0 = shortest delay path,
+                   0.95 = path among the top 5% slowest paths)
+        Returns:
+            DataflowPath at the specified position ratio
+        """
+    index = int(len(self) * (1 - ratio))
```
cowardsa, line 338: Nit: assert 0<= ratio <= 1

lib/Bindings/Python/dialects/aig.py
```diff
@@ -2,5 +2,387 @@
...
+class DebugPoint:
+  """
+    Represents a debug point in the timing path history.
+    Debug points are intermediate points along a timing path that provide
+    insight into the delay accumulation and signal propagation through
+    the circuit. Each point captures the state at a specific location.
+    Attributes:
+        object: The signal/object at this debug point
+        delay: Accumulated delay up to this point (in timing units)
+        comment: Optional descriptive comment about this point
```
cowardsa, line 89: Non-Blocking: If comments are largely derived from sv name hints, then presumably many share the same name and have a unique index into the bitvector - could storing table names and only indices help reduce memory overhead? (apologies if misunderstood intent here)

cowardsa (commented):
LGTM - few additional testing suggestions and queries about debug point (note reviewer experience low - very much still learning)

lib/Bindings/Python/AIGModule.cpp
```diff
@@ -22,4 +26,53 @@ using namespace mlir::python::nanobind_adaptors;
...
+             aigLongestPathAnalysisDestroy(self);
+           })
+      .def("get_all_paths",
+           [](AIGLongestPathAnalysis *self, const std::string &moduleName,
+              bool elaboratePaths) -> AIGLongestPathCollection {
+             MlirStringRef moduleNameRef =
+                 mlirStringRefCreateFromCString(moduleName.c_str());
+
+             if (aigLongestPathCollectionIsNull(
+                     aigLongestPathAnalysisGetAllPaths(*self, moduleNameRef,
```
mikeurbach: Should we call the `aigLongestPathAnalysisGetAllPaths` just once? We could get the paths, check if they're null, then make the AIGLongestPathCollection if not.

mikeurbach (commented):
Nice! No major comments from me, the bindings look good.

lib/Bindings/Python/AIGModule.cpp
uenoku: Oh my, thank you for pointing out! I debugged why LongestPath analysis is fairly slower than C++ but this duplication explains a lot :)

lib/Bindings/Python/dialects/aig.py
uenoku: Yes, that's good point. Eventually i'll modify comment to use MLIR's StringAttr instead of python str. That would prevent unnecessary duplication.

integration_test/Bindings/Python/dialects/aig.py
uenoku: Added a test, thanks!

PR #8662, opened 2025-07-08
[LLHD] Make Deseq pass emit seq.firreg instead of seq.compreg

fabianschuiki:
The LLHD Deseq pass detects clocks and optional asynchronous resets. It currently, mistakenly, creates `seq.compreg` ops with those clocks and resets. The `seq.compreg` op turns out to have a _synchronous_ reset though, making the lowering invalid. Use the `seq.firreg` op instead, as it allows the reset type to be specified explicitly. This requries an additional `comb.mux` to be created for potential enable signals, which `seq.firreg` does not support directly.

We should really nail down the semantics of `seq.compreg` in a follow-up commit. The documentation is very vague, and Deseq is likely not the last pass mistakenly assuming the explicit reset of `seq.compreg` is assumed to be asynchronous.

### DIFF
--- a/include/circt/Dialect/Seq/SeqOps.td
+++ b/include/circt/Dialect/Seq/SeqOps.td
@@ -239,7 +239,6 @@ def FirRegOp : SeqOp<"firreg",
   );
   let results = (outs AnyType:$data);
 
-  let skipDefaultBuilders = 1;
   let hasCanonicalizeMethod = true;
   let hasFolder = true;
   let builders = [
--- a/lib/Dialect/LLHD/Transforms/Deseq.cpp
+++ b/lib/Dialect/LLHD/Transforms/Deseq.cpp
@@ -1004,13 +1004,13 @@ void Deseq::implementRegisters() {
     implementRegister(drive);
 }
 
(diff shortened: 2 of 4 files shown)


include/circt/Dialect/Seq/SeqOps.td
```diff
@@ -239,7 +239,6 @@ def FirRegOp : SeqOp<"firreg",
   );
   let results = (outs AnyType:$data);
 
-  let skipDefaultBuilders = 1;
```
maerhart, line 242: Why is this necessary? Aren't you just using the second custom builder or am I missing something?

maerhart (approved):
LGTM! Would be really nice to get that firreg+compreg unification over the finish line indeed.

include/circt/Dialect/Seq/SeqOps.td
fabianschuiki: That's what I was trying initially, but it turns out that the second custom builder _requires_ the reset and reset value to be non-null. So you'd have to call either the first or second builder depending on whether you want a reset or not. Having the default builder available feels like a good thing.

Issue #8657, opened 2025-07-07
Imported verilog flips output


jpienaar:
```verilog
module conc(
  input wire [63:0] arg0,
  output wire [63:0] result0
);
  wire [31:0] arg0_unflattened[2];
  assign arg0_unflattened[0] = arg0[31:0];
  assign arg0_unflattened[1] = arg0[63:32];

  assign result0 = {arg0_unflattened[1], arg0_unflattened[0]};
endmodule
```

Post

```
./bin/circt-verilog | ./bin/circt-opt --export-verilog
```

we get

```verilog
module conc(	// <stdin>:2:3
  input  [63:0] arg0,	// <stdin>:2:22
  output [63:0] result0	// <stdin>:2:39
);
module conc(

  assign result0 = {arg0[31:0], arg0[63:32]};	// <stdin>:3:10, :4:10, :5:10, :6:5
endmodule
```

(doing it inlined rather than assigning to unflattened first results in result0 = argo)

uenoku:
It seems like related to unpacked array handling in ImportVerilog. It works fine for packed array. 

```verilog
$ cat bar.sv
module conc(
  input wire [63:0] arg0,
  output wire [63:0] result0
);
  wire [1:0][31:0] arg0_unflattened;
  assign arg0_unflattened[0] = arg0[31:0];
  assign arg0_unflattened[1] = arg0[63:32];

  assign result0 = {arg0_unflattened[1], arg0_unflattened[0]};
endmodule

$ circp-verilog bar.sv | circt-opt -export-verilog
// Generated by CIRCT firtool-1.124.0-12-g4bb54cdf9-dirty
module conc(    // <stdin>:2:3
  input  [63:0] arg0,   // <stdin>:2:22
  output [63:0] result0 // <stdin>:2:39
);

  wire [31:0] _arg0_63to32;     // <stdin>:5:10
  wire [31:0] _arg0_31to0;      // <stdin>:4:10
  assign _arg0_31to0 = arg0[31:0];      // <stdin>:4:10
  assign _arg0_63to32 = arg0[63:32];    // <stdin>:5:10
  assign result0 = /*cast(bit[63:0])*/{{_arg0_63to32}, {_arg0_31to0}};  // <stdin>:3:10, :4:10, :5:10, :6:10, :7:5
endmodule

```

jpienaar:
Which is weird ... if I look at ImportVerilog, nothing jumps out. The two different forms (packed and unpacked probably the closest).

[...]

PR #8663, opened 2025-07-08
[importverilog] Lvalue RangelSelect and ElementSelect to use range.

jpienaar:
This makes it uniform with how the RvalueExprVisitor handles it (copied getSelectIndex verbatim, could also make utility function shared between visitors).

Let me know how best to test this, I could do a simple Verilog import/export FileCheck test.

Resolves: #8657

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -1006,12 +1006,42 @@ struct LvalueExprVisitor {
     return builder.create<moore::ConcatRefOp>(loc, operands);
   }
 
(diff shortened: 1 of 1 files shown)


uenoku:
Thanks! Could you add a test to https://github.com/llvm/circt/blob/main/test/Conversion/ImportVerilog/basic.sv?

jpienaar:
Done, could do it simpler than expected by just checking write and read indices match.

fabianschuiki:
I was just running circt-verilog on a few internal design snippets and noticed that your fix here got rid of a lot of LEC mismatches I was seeing. Thanks :grin:

PR #8664, opened 2025-07-08
[Datapath] Conversion Pass Comb to Datapath

cowardsa:
Add support for lowering variadic adders and two-input multipliers to datapath operations. Now automiatng 
```
%0 = comb.mul %a, %b : i4
%1 = comb.add %0, %c : i4
```

Resulting from comb-to-datapath and canonicalize:
```
%0:4 = datapath.partial_product %a, %b : (i4, i4) -> (i4, i4, i4, i4)
%1:2 = datapath.compress %0#0, %0#1, %0#2, %0#3, %c : i4 [5 -> 2]
%2 = comb.add %1#0, %1#1 : i4
```

This will be integrated into circt-synth during a subsequent PR.


lib/Conversion/CombToDatapath/CombToDatapath.cpp
```diff
@@ -0,0 +1,122 @@
...
+  patterns.add<CombAddOpConversion, CombMulOpConversion>(patterns.getContext());
+}
+
+void ConvertCombToDatapathPass::runOnOperation() {
+  ConversionTarget target(getContext());
+
+  target.addLegalDialect<datapath::DatapathDialect, comb::CombDialect,
+                         hw::HWDialect>();
+
+  target.addIllegalOp<comb::AddOp, comb::MulOp>();
```
maerhart: Isn't this redundant due to the below dynamic legality constraints?

lib/Conversion/CombToDatapath/CombToDatapath.cpp
```diff
@@ -0,0 +1,122 @@
...
+namespace {
+// add(a1, a2, ...) -> add(compress(a1, a2, ...))
+struct CombAddOpConversion : OpConversionPattern<AddOp> {
+  using OpConversionPattern<AddOp>::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AddOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Can only introduce compressor for three-inputs or more
+    if (op.getNumOperands() <= 2)
+      return failure();
```
maerhart: Nit: I think the preferred way to fail is to use `rewriter.notifyMatchFailure` which will also notify the listener (also in the pattern below).

lib/Conversion/CombToDatapath/CombToDatapath.cpp
```diff
@@ -0,0 +1,122 @@
...
+    rewriter.replaceOpWithNewOp<AddOp>(op, results.getResults(), true);
+    return success();
+  }
+};
+
+// mul(a,b) -> add(pp(a,b))
+// multi-input adder will be converted to a compressor by other pattern
+struct CombMulOpConversion : OpConversionPattern<MulOp> {
+  using OpConversionPattern<MulOp>::OpConversionPattern;
+  using OpAdaptor = typename OpConversionPattern<MulOp>::OpAdaptor;
```
maerhart: Nit: why add that here but not in the other pattern?
```suggestion
```

test/Conversion/CombToDatapath/comb-to-datapath.mlir
```diff
@@ -0,0 +1,22 @@
...
+  // CHECK-NEXT: comb.add bin %[[COMP1]]#0, %[[COMP1]]#1 : i4
+  %1 = comb.add %arg0, %arg1, %arg2, %arg3 : i4
+  
+  // CHECK-NEXT: %[[PP:.+]]:4 = datapath.partial_product %arg0, %arg1 : (i4, i4) -> (i4, i4, i4, i4)
+  // CHECK-NEXT: %[[COMP2:.+]]:2 = datapath.compress %[[PP]]#0, %[[PP]]#1, %[[PP]]#2, %[[PP]]#3 : i4 [4 -> 2]
+  // CHECK-NEXT: comb.add bin %[[COMP2]]#0, %[[COMP2]]#1 : i4
+  %2 = comb.mul %arg0, %arg1 : i4
+
+  // CHECK-NEXT: comb.mul %arg0, %arg1, %arg2 : i4
+  %7 = comb.mul %arg0, %arg1, %arg2 : i4
```
maerhart, line 21: The addition of i0 -> constant case is not tested

maerhart (approved):
Seems like you're missing a link target in https://github.com/llvm/circt/blob/main/lib/CAPI/Conversion/CMakeLists.txt

Otherwise LGTM!

lib/Conversion/CombToDatapath/CombToDatapath.cpp
```diff
@@ -0,0 +1,122 @@
...
+  ConversionTarget target(getContext());
+
+  target.addLegalDialect<datapath::DatapathDialect, comb::CombDialect,
+                         hw::HWDialect>();
+
+  target.addIllegalOp<comb::AddOp, comb::MulOp>();
+
+  // Permit 2-input adders (carry-propagate adders)
+  target.addDynamicallyLegalOp<comb::AddOp>(
+      [](comb::AddOp op) { return op.getNumOperands() == 2; });
```
uenoku: nit: Can we make unary add legal as well? They will be canonicalized anyway
```suggestion
      [](comb::AddOp op) { return op.getNumOperands() <= 2; });
```

lib/Conversion/CombToDatapath/CombToDatapath.cpp
uenoku: I think operations with `numOperands() <= 2` are legal so I believe patterns won't apply in the first place?

lib/Conversion/CombToDatapath/CombToDatapath.cpp
cowardsa: Thanks Martin - didn't realise that addDynamicallyLegalOp would mark the remainder as illegal - useful to know!

lib/Conversion/CombToDatapath/CombToDatapath.cpp
cowardsa: Removed!

test/Conversion/CombToDatapath/comb-to-datapath.mlir
cowardsa: Added

lib/Conversion/CombToDatapath/CombToDatapath.cpp
cowardsa: Removed this check as suggested

cowardsa:
Thanks reviewers - suggestions implemented ready to merge once checks are completed - still awaiting commit access approval...

PR #8665, opened 2025-07-08
Remove single state restriction on FSM to SV

jpienaar:
The conversion seems to work and gets to Verilog. It seems correct, if not most efficient lowered form if there is only one state. Not sure if there was a different reason for this restriction (beyond, this would be better as just HWModule, although didn't see a pattern that would do that).


jpienaar:
Sure, sorry for delay. I'm not sure where to put it, I can do a unit test of just the conversion to SV, but it feels like seeing the verilog shows no further issues. But could also stop at this point and just test output of this pass.

I could also split it in 2: do one here and on in export pass that uses the output of the first two as input.

PR #8666, opened 2025-07-08
[RTG] Add LowerValidateToLabels pass

maerhart:
The pass to be run in the first compilation run to generate sideband command labels understood by the ISA simulator (e.g. Spike).

### DIFF
--- a/lib/Dialect/RTG/Transforms/LowerValidateToLabelsPass.cpp
+++ b/lib/Dialect/RTG/Transforms/LowerValidateToLabelsPass.cpp
@@ -1,0 +1,64 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/RTG/IR/RTGOps.h"
+#include "circt/Dialect/RTG/Transforms/RTGPasses.h"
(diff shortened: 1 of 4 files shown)


lib/Dialect/RTG/Transforms/LowerValidateToLabelsPass.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    OpBuilder builder(validateOp);
+    auto intrinsicLabel = validateOp.getRef().getType().getIntrinsicLabel(
+        regOp.getReg(), validateOp.getId().value());
+    Value lbl = builder.create<rtg::LabelDeclOp>(
+        loc, StringAttr::get(&getContext(), intrinsicLabel), ValueRange());
+    builder.create<rtg::LabelOp>(loc, rtg::LabelVisibility::global, lbl);
+    validateOp.getValue().replaceAllUsesWith(validateOp.getDefaultValue());
+
+    pruner.eraseNow(validateOp);
```
uenoku, line 56: nit: Strictly speaking it's not allowed to erase operations in walk other than the current operation (`validateOp`). Though I think it's always legal with the current implementation of walk specifically.

uenoku (approved):
LGTM

lib/Dialect/RTG/Transforms/LowerValidateToLabelsPass.cpp
maerhart: This call only erases the `validateOp` and remembers now unused operations defining the operands in the `eraseNow` call below.

uenoku: Oh, thanks. I misunderstood that eraseNow walks and erases operations transitively....

PR #8667, opened 2025-07-08
Convert to CF before running LLHD inline.

jpienaar:
The LLHD inliner doesn't check if the parent op only supports single block regions when inlining a call (while the function body of the call may already be multiblock). This flips the lowering to CF to before LLHD inline to avoid a common variant of this (inlining into SCF ops).


fabianschuiki (approved):
LGTM! Thanks for fixing this :smiley:

PR #8668, opened 2025-07-08
[FIRRTL] Lint XMRs in the "design"

seldridge:
Add an option to FIRRTL's linter (and expose it up to `firtool`) that will
lint (error) if any XMRs exist in the "design".  This is intended to catch
problems where downstream consumers of CIRCT-generated Verilog (e.g.,
Physical Design teams) get unexpected unsynthesizable XMRs in the design.
This is almost _never_ what you want.

Unfortunately, this is highly error prone.  It's difficult to tell what
is/is not the "design".  This relies on the InstanceInfo analysis and
intentionally uses its notion of "design" and not "effective design".  In
other words, this will only trigger this kind of linting if there is an
explicit `MarkDUTAnnotation`.  It will not assume that the main module has
an implicit `MarkDUTAnnotation`.  This is intended to avoid false
positives.

In order to make this work, there is a lot of prequisite work batched in
this PR.  A new pass, which only exists to run the `InstanceInfo` analysis
(and thereby cache it) is created.  This allows for use of the
`getCachedParentAnalysis` function and guaranteeing that it will succeed.
(This is a pattern borrowed from IREE.)  Options are added for the
singular, existing lint check as well.

This PR is best reviewed as individual commits.  I am intending to land
this as individual commits as well.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Lint.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Lint.cpp
@@ -31,6 +31,9 @@ struct Config {
   /// If true, then assertions that are statically false (and will trivially
   /// fail simulation) will result in an error.
   bool lintStaticAsserts;
+  /// If true, then XMRs that exist in the "design" (not the "effective design")
+  /// will be flagged as an error.
+  bool lintXmrsInDesign;
 };
 
 /// Class that stores state related to linting.  This exists to avoid needing to
@@ -53,7 +56,7 @@ public:
           failed = true;
 
       if (auto xmrDerefOp = dyn_cast<XMRDerefOp>(op))
-        if (checkXmr(xmrDerefOp).failed())
+        if (config.lintXmrsInDesign && checkXmr(xmrDerefOp).failed())
           failed = true;
 
       return WalkResult::advance();
@@ -143,6 +146,7 @@ private:
 
 struct LintPass : public circt::firrtl::impl::LintBase<LintPass> {
   using LintBase::lintStaticAsserts;
+  using LintBase::lintXmrsInDesign;
 
   void runOnOperation() override {
 
@@ -155,7 +159,9 @@ struct LintPass : public circt::firrtl::impl::LintBase<LintPass> {
       return failure();
     };
     auto transform = [&](FModuleOp moduleOp) -> LogicalResult {
-      return Linter(moduleOp, instanceInfo, {lintStaticAsserts}).lint();
+      return Linter(moduleOp, instanceInfo,
+                    {lintStaticAsserts, lintXmrsInDesign})
+          .lint();
     };
 
     SmallVector<FModuleOp> modules(circuitOp.getOps<FModuleOp>());
@@ -168,8 +174,10 @@ struct LintPass : public circt::firrtl::impl::LintBase<LintPass> {
 };
 } // namespace
 
-std::unique_ptr<Pass> firrtl::createLintingPass(bool lintStaticAsserts) {
+std::unique_ptr<Pass> firrtl::createLintingPass(bool lintStaticAsserts,
+                                                bool lintXmrsInDesign) {
   auto pass = std::make_unique<LintPass>();
   pass->lintStaticAsserts = lintStaticAsserts;
+  pass->lintXmrsInDesign = lintXmrsInDesign;
   return pass;
 }
--- a/lib/Dialect/FIRRTL/Transforms/Lint.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Lint.cpp
@@ -26,24 +26,15 @@ using namespace circt;
 using namespace firrtl;
 
 namespace {
-/// Linter configuration.
-struct Config {
-  /// If true, then assertions that are statically false (and will trivially
-  /// fail simulation) will result in an error.
-  bool lintStaticAsserts;
-  /// If true, then XMRs that exist in the "design" (not the "effective design")
-  /// will be flagged as an error.
-  bool lintXmrsInDesign;
-};
-
 /// Class that stores state related to linting.  This exists to avoid needing to
 /// clear members of `LintPass` and instead just rely on `Linter` objects being
 /// deleted.
 class Linter {
 
 public:
-  Linter(FModuleOp fModule, InstanceInfo &instanceInfo, const Config &config)
-      : fModule(fModule), instanceInfo(instanceInfo), config(config){};
+  Linter(FModuleOp fModule, InstanceInfo &instanceInfo,
+         const LintOptions &options)
+      : fModule(fModule), instanceInfo(instanceInfo), options(options){};
 
   /// Lint the specified module.
   LogicalResult lint() {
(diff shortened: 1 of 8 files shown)


uenoku (commented):
If it was necessary to add a new pass to cache InstanceInfo can't we just make linter pass circuit op pass? For this specific use case it's safe since Lint pass doesn't mutate operations and circuit passes are executed before and after Lint pass but it means you cannot put Lint pass to arbitrary

seldridge:
@uenoku: That's the alternative way to proceed here.

I was toying with this approach as it seemed beneficial to keep the `FModuleOp` parallelism exposed to MLIR infrastructure as opposed to manually managing it. However, there appear to be no actual users of `getCachedParentAnalysis` other than that one user in IREE.

lib/Dialect/FIRRTL/Transforms/Lint.cpp
```diff
@@ -83,18 +102,67 @@ class Linter {
...
+  LogicalResult checkXmr(XMRDerefOp op) {
+    // XMRs under layers are okay.
+    if (op->getParentOfType<LayerBlockOp>() ||
+        op->getParentOfType<sv::IfDefOp>())
+      return success();
+
+    // The XMR is not under a layer.  This module must never be instantiated in
+    // the design.  Intentionally do NOT use "effective" design as this could
+    // lead to false positives.
+    if (!instanceInfo.allInstancesInDesign(fModule))
```
rwy7: I think this should be "not any". It would be nice if we could move this check out of the tight loop.

Here is an example of a missed XMR:
```mlir

firrtl.circuit "TB" {
  hw.hierpath private @xmrPath [@Foo::@sym]

  firrtl.module @Foo() {
    %a = firrtl.wire sym @sym : !firrtl.uint<1>
    %0 = firrtl.xmr.deref @xmrPath : !firrtl.uint<1>
  }

  firrtl.module @TB() {
    firrtl.instance dut @DUT()
    firrtl.instance foo @Foo()
  }

  firrtl.module @DUT() attributes {
    annotations = [
      {class = "sifive.enterprise.firrtl.MarkDUTAnnotation"}
    ]
  } {
    firrtl.instance foo @Foo()
  }
}
```

seldridge: Good catch! This is fixed in the forced update to: https://github.com/llvm/circt/pull/8668/commits/b5fd923a6755190d5f9b86dcdd583492244b7303

seldridge: Also, nice test. I included a distilled version of this in: 7e6892a49

lib/Dialect/FIRRTL/Transforms/Lint.cpp
```diff
@@ -83,18 +94,60 @@ class Linter {
...
 struct LintPass : public circt::firrtl::impl::LintBase<LintPass> {
+  using LintBase::LintBase;
+
   void runOnOperation() override {
-    if (failed(Linter(getOperation()).lint()))
+
+    CircuitOp circuitOp = getOperation();
+    auto instanceInfo = getAnalysis<InstanceInfo>();
+
+    if (failed(mlir::failableParallelForEach(
```
uenoku: Maybe we want to run Lint on all modules even when there is a failure?

lib/Dialect/FIRRTL/Transforms/Lint.cpp
```diff
@@ -83,18 +94,60 @@ class Linter {
...
+    // the design.  Intentionally do NOT use "effective" design as this could
+    // lead to false positives.
+    if (!instanceInfo.anyInstanceInDesign(fModule))
+      return success();
+
+    // If the op has a single user, that user is a connect, and the connect is
+    // to an instance which is marked `lowerToBind`, then this is a pattern for
+    // inlining the XMR into the bound instance site.  This pattern is used by
+    // Grand Central, but not elsewhere.
+    if (op.getResult().hasOneUse()) {
```
uenoku: Is single user restriction necessary? I think it's ok if all users are used by bound instances.

lib/Firtool/Firtool.cpp
```diff
@@ -819,7 +831,7 @@ circt::firtool::FirtoolOptions::FirtoolOptions()
       stripDebugInfo(false), fixupEICGWrapper(false), addCompanionAssume(false),
       disableCSEinClasses(false), selectDefaultInstanceChoice(false),
       symbolicValueLowering(verif::SymbolicValueLowering::ExtModule),
-      disableWireElimination(false) {
+      disableWireElimination(false), lintStaticAsserts(true) {
```
uenoku: missing litXmrsInDesign.

lib/Firtool/Firtool.cpp
```diff
@@ -780,6 +781,17 @@ struct FirtoolCmdOptions {
...
+  //===----------------------------------------------------------------------===
+  // Lint options
+  //===----------------------------------------------------------------------===
+
+  llvm::cl::opt<bool> lintStaticAsserts{
+      "lint-static-asserts", llvm::cl::desc("Lint static assertions"),
+      llvm::cl::init(true)};
+  llvm::cl::opt<bool> lintXmrsInDesign{
+      "lint-xmrs-in-design", llvm::cl::desc("Lint XMRs in the design"),
+      llvm::cl::init(true)};
```
uenoku, line 794: Do we need options to disable them? 🤔 I feel it's beneficial to run always.

uenoku (approved):
LGTM

lib/Firtool/Firtool.cpp
seldridge: We get that for free with `-lint-xmrs-in-design=0` / `-lint-xmrs-in-design=false`. This is slightly not great as users may not realize that they can disable these like that. WDYT?

uenoku: I think having options are good for me as long as there are use cases.

lib/Dialect/FIRRTL/Transforms/Lint.cpp
seldridge: Hey, great idea. This especially matters after I made it so that all lint errors are reported in: https://github.com/llvm/circt/commit/d9e5efd4d83d40276cf60fd71a73d4d2d28eca20

This is fixed by switching to `transformReduce` in: https://github.com/llvm/circt/pull/8668/commits/1629b842d019c66b2ba5bdb6a9426c088944a9e2

(For this and future "fixed in" comments, I had to do a rebase and force push due to the interaction of the commits here.)

lib/Dialect/FIRRTL/Transforms/Lint.cpp
seldridge: It shouldn't be. Additionally, this made me realize that _no user_ is also fine as an unused expression will also not be emitted. I switched to an `all_of` check in the updated https://github.com/llvm/circt/pull/8668/commits/d8aa3b69582e705482dd94f1b4f8a69c060a1dd2. This also updated all the tests to make sure there was a user of the XMR and adds a test of no-user in "design" being okay.

lib/Firtool/Firtool.cpp
seldridge: Whoops. Fixed in the forcibly updated: https://github.com/llvm/circt/pull/8668/commits/921a4d328a01cb855c6d50573580e8d4465adc3a

PR #8669, opened 2025-07-08
[FIRRTL] Delete BlackBoxResourceFileNameAnno

seldridge:
This is two logical commits:

1. Remove `BlackBoxResourceFileNameAnno` since this has not been load beagin
   after it was removed from Chisel [[1]].  This annotation was used to control
   the name of a file which was produced that listed all blackboxes in the
   design.  Instead, use the default name always.

2. Completely remove the generation of the vestigial
   `firrtl_black_box_resource_files.f`.  This file is intended to be unused by
   downstream flows.

[1]: https://github.com/chipsalliance/chisel/commit/70f5c7066b1e65cbb9d580703449acb5912a2b05

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
@@ -269,11 +269,6 @@ void BlackBoxReaderPass::runOnOperation() {
     SmallVector<Attribute> symbols;
     for (emit::FileOp file : fileListFiles)
       symbols.push_back(FlatSymbolRefAttr::get(file.getSymNameAttr()));
-
-    builder.create<emit::FileListOp>(
-        loc, builder.getStringAttr("firrtl_black_box_resource_files.f"),
-        builder.getArrayAttr(symbols),
-        builder.getStringAttr(ns.newName("blackbox_filelist")));
   }
 
   // If nothing has changed we can preserve the analysis.
--- a/test/Dialect/FIRRTL/SFCTests/GrandCentralInterfaces/Wire.fir
+++ b/test/Dialect/FIRRTL/SFCTests/GrandCentralInterfaces/Wire.fir
@@ -469,9 +469,3 @@ circuit Top :
     ; EXTRACT:        FILE ".{{[/\]}}BlackBox_DUT.v"
     ; EXTRACT:        FILE "Wire{{[/\]}}firrtl{{[/\]}}gct{{[/\]}}BlackBox_GCT.v"
     ; EXTRACT:        FILE ".{{[/\]}}BlackBox_DUTAndGCT.v"
-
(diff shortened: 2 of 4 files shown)


fabianschuiki (approved):
Awesome! CIRCT 1 : Bit-Rot 0.

tymcauley:
@seldridge after removing `firrtl_black_box_resource_files.f`, what is the intended way to get the list of black-box files produced by `firtool`?

seldridge:
@tymcauley: Do you need to know the names of any of the generated blackboxes given that these will always show up in one of the output build directories?

I.e., `BlackBoxPathAnno` is a way of telling CIRCT to copy some file into another file relative to the output directory. `BlackBoxInlineAnno` is a way of having verbatim text written to a file relative to the output directory. Given this, the build system should be able to do a file glob of the know build areas to get all blackboxes generated in this way.

Note: I'm spitballing this here as I dropped this without much concern for existing users. 😬 Are you using this currently and how are you using it?

tymcauley:
Ya, I just need to know the names of those generated blackbox files. Right now, they're showing up in my output directory along with all of the other generated SystemVerilog files. I'm using a build system similar to what Chipyard does: https://github.com/ucb-bar/chipyard/blob/973f8732d4ffeb8cc231edf109e84eef114e3d6e/common.mk#L188

Is there a way to place all of the blackbox files defined in Chisel into a subdirectory of the `firtool` output area?

seldridge:
> Is there a way to place all of the blackbox files defined in Chisel into a subdirectory of the firtool output area?

`firrtl.transforms.BlackBoxTargetDirAnno` sets this. (You could alternatively do something in Chisel to setup the blackbox directory for all the annotations, though that may be really annoying.) Using this, you should be able to duplicate the behavior with a file glob or `ls`.

Note: I am entirely fine with reverting this if we need to! However, if dropping this shakes out a simpler solution with fewer ways of doing things that would be good. 😉

tymcauley:
Okay, that annotation looks like it should do the trick, thanks! For the path provided in that annotation, can it be a relative path, and I'm assuming it would be relative to wherever `firtool` is executed from?

seldridge:
Yes, it can be relative. It should be relative to the `-o` path, not necessarily the path where `firtool` is executed from.

seldridge:
This breaks upstream Chisel as it relies on this file to find blackboxes through logic here: https://github.com/chipsalliance/chisel/blob/177579fe6159217b607d898ee5180445627a4427/src/main/scala/chisel3/simulator/package.scala#L236

I will plan to update svsim so that this will continue to work. svsim really shouldn't be relying on these undocumented filelists.

tymcauley:
I also ran into #8672 while trying to experiment with `firrtl.transforms.BlackBoxTargetDirAnno`.

PR #8670, opened 2025-07-08
Add SitestBlackBoxLibrariesAnnotation; process in CreateSiFiveMetadataPass

tmckay-sifive:
This allows extmodules to be annotated with additional library names to be included in the blackbox metadata. Downstream tooling can consume this to determine additional libraries that should be compiled alongside extmodules.

- Add SitestBlackBoxLibrariesAnnotation; process in CreateSiFiveMetadataPass
- Update docs
- Update and add tests

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
@@ -188,8 +188,13 @@ struct ObjectModelIR {
     // Create list of blackbox resources from the provided list
     SmallVector<Value> libValues;
     for (StringRef libName : libraries) {
-      auto libNameAttr = builderOM.create<StringConstantOp>(
-          builderOM.getStringAttr(libName));
+      Value libNameAttr;
+      if (libName == defName) {
+        libNameAttr = modEntry;
+      } else {
+        libNameAttr = builderOM.create<StringConstantOp>(
+            builderOM.getStringAttr(libName));
+      }
       libValues.push_back(libNameAttr);
     }
     auto blackBoxResourcesList = builderOM.create<ListCreateOp>(
--- a/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
@@ -165,7 +165,7 @@ struct ObjectModelIR {
     auto unknownLoc = mlir::UnknownLoc::get(context);
     auto builderOM = mlir::ImplicitLocOpBuilder::atBlockEnd(
         unknownLoc, circtOp.getBodyBlock());
-    Type classFieldTypes[] = {StringType::get(context), BoolType::get(context)};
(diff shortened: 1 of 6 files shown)


lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -819,30 +838,51 @@ CreateSiFiveMetadataPass::emitSitestBlackboxMetadata(ObjectModelIR &omir) {
...
+        for (auto lib : libsAttr) {
+          if (auto libStr = dyn_cast<StringAttr>(lib)) {
+            libs.push_back(libStr.getValue());
+            hasLibs = true;
+          }
+        }
+      }
+    }
+
+    if (!libs.empty()) {
```
mikeurbach: nit: I think our [preferred style](https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code) would be more like

```
if (libs.empty())
  continue;
```

And to generally pull early returns / continues as soon as possible. I don't think this could be pulled any earlier, but that is the general gist.

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -184,13 +184,32 @@ struct ObjectModelIR {
...
     auto modEntry = builderOM.create<StringConstantOp>(module.getDefnameAttr());
     auto inDutAttr = builderOM.create<BoolConstantOp>(inDut);
+
+    // Create list of blackbox resources from the provided list
+    SmallVector<Value> libValues;
+    for (StringRef libName : libraries) {
+      Value libNameAttr;
+      if (libName == defName) {
+        libNameAttr = modEntry;
+      } else {
```
mikeurbach, line 196: nit: I think the `else` case is simple enough that parentheses could be elided by our [rule of thumb](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements).

mikeurbach (approved):
The implementation generally looks good to me. Just a couple style nits. @prithayan since you're familiar with this, do you mind taking a look?

lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp
```diff
@@ -184,13 +184,32 @@ struct ObjectModelIR {
         module.getLoc(), blackBoxMetadataClass.getBodyBlock());
     auto modEntry = builderOM.create<StringConstantOp>(module.getDefnameAttr());
     auto inDutAttr = builderOM.create<BoolConstantOp>(inDut);
+
+    // Create list of blackbox resources from the provided list
```
tmckay-sifive:
```suggestion
```

stray comment

prithayan (approved):
The metadata updates look good.

prithayan:
I think `clang-format -i     lib/Dialect/FIRRTL/Transforms/CreateSiFiveMetadata.cpp` needs to be run.

PR #8674, opened 2025-07-09
Insert probes at least post def.

jpienaar:
Handle failure case where the probe was being inserted before def.

### DIFF
--- a/test/Dialect/LLHD/Transforms/mem2reg.mlir
+++ b/test/Dialect/LLHD/Transforms/mem2reg.mlir
@@ -992,3 +992,26 @@ hw.module @CombCreateDynamicInject(in %u: i42, in %v: i10, in %q: i1) {
 func.func private @use_i42(%arg0: i42)
 func.func private @use_inout_i42(%arg0: !hw.inout<i42>)
 func.func private @use_array_i42(%arg0: !hw.array<4xi42>)
+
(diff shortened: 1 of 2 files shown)


test/Dialect/LLHD/Transforms/mem2reg.mlir
```diff
@@ -992,3 +992,26 @@ hw.module @CombCreateDynamicInject(in %u: i42, in %v: i10, in %q: i1) {
...
+    %false_0 = hw.constant false
+    %e = llhd.sig %false_0 : i1
+    %3 = comb.concat %c0_i31, %1 : i31, i1
+    %4 = llhd.prb %e : !hw.inout<i1>
+    %5 = comb.icmp ult %3, %c7_i32 : i32
+    %6 = comb.and %5, %4 : i1
+    llhd.yield %6 : i1
+  }
+  hw.output
+}
```
maerhart: Nit: this can probably be trimmed down a bit
I guess something like this is enough to trigger the issue?
```
hw.module @ProbePostDef() {
  %2 = llhd.combinational -> i1 {
    %false = hw.constant false
    %e = llhd.sig %false : i1
    %4 = llhd.prb %e : !hw.inout<i1>
    llhd.yield %4 : i1
  }
  hw.output
}
```

maerhart (approved):
Thanks!

PR #8675, opened 2025-07-09
[RTG] Add Randomization Pipeline

maerhart:


include/circt/Dialect/RTG/Transforms/RTGPassPipelines.h
```diff
@@ -0,0 +1,42 @@
...
+//===----------------------------------------------------------------------===//
+// Building and Registering.
+//===----------------------------------------------------------------------===//
+
+/// Adds the randomization pipeline to the `OpPassManager`.
+void buildRandomizationPipeline(mlir::OpPassManager &pm,
+                                const RandomizationPipelineOptions &options);
+
+/// Registers all pipelines for the `rtg` dialect.
+void registerRTGPipelines();
```
fabianschuiki: Nit: maybe drop "RTG" here because it's already in the namespace name. Then you have symmetrical `rtg::registerPasses` and `rtg::registerPipelines`.

fabianschuiki (approved):
Really cool! Our first pass pipeline in CIRCT?

fabianschuiki:
@uenoku That would be very nice. Also maybe for arcilator and circt-verilog, which have similar pipelines. Kind of a nice idea to factor the pipeline out of the tool itself :thinking:

PR #8678, opened 2025-07-09
[FIRRTL] Fix inject-dut-hier deleting new top

seldridge:
Fix an issue with the `moveDut=true` behavior of the `InjectDUTHierarchy`
pass that would occur if the design-under-test (DUT) was the main module
in a circuit.  If this happens, this would create a new wrapper module
above the DUT, mark it private, and keep the main module of the circuit
pointing at the wrapper.  This would then cause the new DUT to be deleted
as it was a private module.

This is not FIRRTL ABI compatible as it will create a new public module
and cause things like additional ref files to be created.

CC: @azidar


uenoku (approved):
This makes sense to me.

PR #8680, opened 2025-07-09
[ImportVerilog] Incorporate block names to variable/instance names

### LINKED ISSUE #8679: [ImportVerilog] Incorporate block names into signal/instance names
Consider this Verilog snippet:
```systemverilog
module Top;
  int x;
  begin : foo
    int y;
    for (genvar i = 0; i < 4; ++i) begin : bar
      int z;
    end
  end
endmodule
```
In most EDA tools this would generate the following signals:
```
Top/x
Top/foo.y
Top/foo.bar[0].z
Top/foo.bar[1].z
Top/foo.bar[2].z
Top/foo.bar[3].z
```
We should do the same thing. ImportVerilog currently ignores block names, thus creating the following signals:
```
Top/x
Top/y
Top/z
Top/z_0
Top/z_1
Top/z_2
```
When picking names for signals and instances, try to incorporate the enclosing block names.

fabianschuiki:
Use the names of surrounding generate blocks as a prefix for variable and instance names. This makes the names match up more closely with how existing EDA tools would label variables and instances embededded in generate blocks.

Consider the following Verilog input:

```systemverilog
module Top;
  int x;
  begin : foo
    int y;
    for (genvar i = 2; i < 6; ++i) begin : bar
      int z;
    end
  end
endmodule
```

The ImportVerilog conversion pass will now produce the following variables:

```mlir
%x = moore.variable : <i32>            // before: %x
%foo.y = moore.variable : <i32>        // before: %y
%foo.bar_2.z = moore.variable : <i32>  // before: %z
%foo.bar_3.z = moore.variable : <i32>  // before: %z_0
%foo.bar_4.z = moore.variable : <i32>  // before: %z_1
%foo.bar_5.z = moore.variable : <i32>  // before: %z_2
```

This closely matches names like `foo.bar[2].z` or `foo_bar_2_z` that would be generated by common EDA tools. Ideally we would be able to use a name like `foo.bar[2].z` directly, but ExportVerilog currently does not support escaped identifiers, causing the name to be sanitized to something like `foo_bar5B25D_z`.

By incorporating for-generate loop indices and block names into the variable name, logical equivalence checking should become easier and the signal names in Arcilator waveforms will match user expectations more closely.

Fixes #8679.

### DIFF
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -228,8 +228,14 @@ static moore::NetKind convertNetKind(slang::ast::NetType::NetKind kind) {
 
 namespace {
 struct ModuleVisitor : public BaseVisitor {
-  using BaseVisitor::BaseVisitor;
   using BaseVisitor::visit;
+
+  // A prefix of block names such as `foo.bar.` to put in front of variable and
+  // instance names.
+  StringRef blockNamePrefix;
+
+  ModuleVisitor(Context &context, Location loc, StringRef blockNamePrefix = "")
+      : BaseVisitor(context, loc), blockNamePrefix(blockNamePrefix) {}
 
   // Skip ports which are already handled by the module itself.
   LogicalResult visit(const slang::ast::PortSymbol &) { return success(); }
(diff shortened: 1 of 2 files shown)


uenoku (approved):
LGTM

PR #8681, opened 2025-07-10
[circt-synth] [Synthesis] Add synthesis pipeline and refactor the lib structure

uenoku:
This commit introduces a standardized synthesis pipeline and restructures the codebase:

* Creates a new SynthesisPipeline class to define the default synthesis pipeline. This pipeline serves both circt-synth and is exposed through the C API for Python bindings.
* Added a dedicated Synthesis directory under `lib/` to house synthesis-related code. This architectural change is aimed to promote synthesis capabilities to a first-class component within CIRCT rather than limiting it to the circt-synth tool.


include/circt/Synthesis/SynthesisPipeline.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Pass/PassOptions.h"
+#include <string>
+#include <vector>
+
+namespace mlir {
+class PassManager;
+class OpPassManager;
+} // namespace mlir
```
maerhart: Why forward declare this when including "PassManager.h" that contains the declarations? I get that it would be more convenient if there is a large number of segments, but when invoking the pipeline using the CLI flag it feels a bit strange to add a flag for the pipeline but then a sub-flag to truncate the pipeline again. This flag would then stay in the circt-synth tool if there is a need for the user to stop at various points.

include/circt/Synthesis/SynthesisPipeline.h
```diff
@@ -0,0 +1,74 @@
...
+//===----------------------------------------------------------------------===//
+namespace circt {
+namespace synthesis {
+
+/// Options for the synthesis pipeline.
+struct SynthesisPipelineOptions
+    : public mlir::PassPipelineOptions<SynthesisPipelineOptions> {
+
+  /// Enumeration for pipeline stopping points.
+  enum UntilPoint { UntilAIGLowering = 0, UntilEnd = 1 };
```
maerhart: Why not define one pipeline per segment and potentially another one that runs all the segment pipelines?

include/circt/Synthesis/SynthesisPipeline.h
```diff
@@ -0,0 +1,74 @@
...
+
+/// Populate the synthesis pipeline with the given options.
+void buildSynthesisPipeline(mlir::OpPassManager &pm,
+                            const SynthesisPipelineOptions &options);
+
+/// Register the synthesis pipeline.
+void registerSynthesisPipeline();
+} // namespace synthesis
+
+} // namespace circt
```
maerhart:
```suggestion
} // namespace circt

```

maerhart (commented):
Really nice!
What kind of code do you expect to end up in the 'Synthesis' directory and what would be other tools next to circt-synth that could take advantage of it? I'm wondering whether it's worth adding that already or if the pipeline definition should just be in the 'Transforms ' directory for now. But either way sounds fine to me.

include/circt/Synthesis/SynthesisPipeline.h
uenoku: Good points. I separated the pipeline 👍

uenoku:
> What kind of code do you expect to end up in the 'Synthesis' directory and what would be other tools next to circt-synth

I have a plan to add Mapper framework and eventually want to implement STA stuffs there. I considered putting under Transform too but I thought Synthesis pipeline might be too specific sets of pipelines for Transform. I may revisit later but please let me continue with the current structure. Thank you for the suggestion!!

Issue #4483, opened 2022-12-23
[Comb] Flow-sensitive? mux condition canonicalizer


uenoku:
Input:
```
circuit Foo:
  module Foo:
     input clock: Clock
     input a: UInt<1>
     input b: UInt<4>
     output out: UInt<4>
     reg release_state: UInt<4>, clock 
     release_state <= b
     node _T_2238 = eq(release_state, UInt<4>("h5"))
     when _T_2238 : 
       release_state <= UInt<4>("h0") 
     node _T_2240 = eq(release_state, UInt<4>("h2")) 
     when _T_2240 : 
       release_state <= UInt<4>("h7") 
     when _T_2238 :
       release_state <= UInt<4>("h2") 

     out <= release_state
```

Current output:
```
  always @(posedge clock) begin
    automatic logic _T_2238;
    _T_2238 = release_state == 4'h5; 
    if (_T_2238) 
      release_state <= 4'h2; 
    else if (release_state == 4'h2)
      release_state <= 4'h7;
    else if (_T_2238)  
      release_state <= 4'h0;
    else 
      release_state <= b;
  end // always @(posedge)
```

The second `if(_T_2238)` should be removed because that's clearly dead code. The canonicalizer can be implemented in comb mux canonilizer by analyzing mux chain's conditions.

```mlir
hw.module @Foo(%clock: i1, %a: i1, %b: i4) -> (out: i4) {
  %c5_i4 = hw.constant 5 : i4
  %c0_i4 = hw.constant 0 : i4
  %c2_i4 = hw.constant 2 : i4
  %c7_i4 = hw.constant 7 : i4
  %release_state = seq.firreg %4 clock %clock {firrtl.random_init_start = 0 : ui64} : i4
  %0 = comb.icmp bin eq %release_state, %c5_i4 {sv.namehint = "_T_2238"} : i4
  %1 = comb.mux bin %0, %c0_i4, %b : i4
  %2 = comb.icmp bin eq %release_state, %c2_i4 {sv.namehint = "_T_2240"} : i4
  %3 = comb.mux bin %2, %c7_i4, %1 : i4
  %4 = comb.mux bin %0, %c2_i4, %3 : i4
  hw.output %release_state : i4
}
```

Should be optimized into 
```
```mlir
hw.module @Foo(%clock: i1, %a: i1, %b: i4) -> (out: i4) {
  %c5_i4 = hw.constant 5 : i4
  %c0_i4 = hw.constant 0 : i4
[...]

PR #8685, opened 2025-07-11
[Comb] Replace mux cond uses in true/false operand with constant

fabianschuiki:
Implement the following canonicalizations which essentially assume the value of a mux condition to be a constant in the op that supplies the true or false value of the mux:

```
mux(cond, op(cond), x) -> mux(cond, op(1), x)
mux(cond, x, op(cond)) -> mux(cond, x, op(0))
```

This only works if the true or false value only have a single use.


test/Dialect/Comb/canonicalization.mlir
```diff
@@ -1668,3 +1668,25 @@ hw.module @mul_0(in %arg0: i0, in %arg1: i0, out out: i0) {
   %0 = comb.mul %arg0, %arg1 : i0
   hw.output %0 : i0
 }
+
+// CHECK-LABEL: @assumeMuxCondInOperand
+func.func @assumeMuxCondInOperand(%arg0: i1, %arg1: i1) -> i2 {
```
uenoku, line 1673: Could you add a test case that muxValue has multiple uses?

uenoku (approved):
LGTM!

FYI https://github.com/llvm/circt/issues/4483, thank you for adding it!

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -2240,6 +2240,29 @@ static bool foldMuxOfUniformArrays(MuxOp op, PatternRewriter &rewriter) {
...
 }
 
+/// If the mux condition is an operand to the op defining its true or false
+/// value, replace the condition with 1 or 0.
+static bool assumeMuxCondInOperand(Value muxCond, Value muxValue,
+                                   bool constCond, PatternRewriter &rewriter) {
+  if (!muxValue.hasOneUse())
+    return false;
+  auto *op = muxValue.getDefiningOp();
+  if (!op)
```
uenoku: I think we need more restriction here. I think it's illegal to replace operands if op is observable in a different way (XMR to seq.firreg, hw.wire). 

I think it would be sufficient to restrict to combinatorial operations.

fabianschuiki: Oh yeah excellent point :+1: Thanks!

test/Dialect/Comb/canonicalization.mlir
fabianschuiki: Should be just below :smiley:

uenoku: Yeah I missed that. Thanks.

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -2240,6 +2240,29 @@ static bool foldMuxOfUniformArrays(MuxOp op, PatternRewriter &rewriter) {
...
 }
 
+/// If the mux condition is an operand to the op defining its true or false
+/// value, replace the condition with 1 or 0.
+static bool assumeMuxCondInOperand(Value muxCond, Value muxValue,
+                                   bool constCond, PatternRewriter &rewriter) {
+  if (!muxValue.hasOneUse())
+    return false;
+  auto *op = muxValue.getDefiningOp();
+  if (!op || !isa<CombDialect>(op->getDialect()))
```
teqdruid, line 2250: Shouldn't this be checking that an operation is pure? (I'm not sure what traits assure that it doesn't have side effects.) I think this could apply more generally.

fabianschuiki: I think this also technically works for side-effecting operations, since it just changes an operand, but doesn't move or replicate the op itself.

@uenoku pointed out that we'd want to limit this to Comb for now, since things like XMRs on hw.wire would be able to observe the optimization.

uenoku: My suggestion was too conservative and maybe pure is legal. I thought HW wire was pure but actually it's not marked as pure.

teqdruid: > I think this also technically works for side-effecting operations, since it just changes an operand, but doesn't move or replicate the op itself.

Good point... Though I think it depends on your definition of "side-effect". Would the XMR case count? `hw.wire` isn't marked as having side effects.

fabianschuiki: Yeah I think the XMR would count as a side-effect. Or maybe if the XMR would cause a use on the `hw.wire`'s SSA value somehow, that would even be better. But currently we implement those with inner symbols, which makes this additional channel of observability of a value harder to see and reason about.

teqdruid: OK. I understand the problem now. It's actually broader than this -- anything which has an inner symbol could potentially be observed, kinda like it has an implicit consumer.

fabianschuiki: Yes exactly! I would _love_ for us to have a different system where some of this XMR business is funneled through SSA values of some special reference type or so, such that regular SSA reasoning applies. But that's pretty tricky to do :cry:

PR #8687, opened 2025-07-11
[HW] MaterializeConstant: check for null block

teqdruid:
Builders don't always specify the block. If it doesn't, this function segfaults.

### DIFF
--- a/lib/Dialect/HW/HWDialect.cpp
+++ b/lib/Dialect/HW/HWDialect.cpp
@@ -102,7 +102,10 @@ Operation *HWDialect::materializeConstant(OpBuilder &builder, Attribute value,
   }
 
   // Parameter expressions materialize into hw.param.value.
-  auto parentOp = builder.getBlock()->getParentOp();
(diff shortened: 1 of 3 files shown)


teqdruid:
This was the root cause of a bug reported by a PyCDE user. In this case, it was triggered by the `lower-seq-hlmem` but it's been a problem for me in the past. I was able to fix it for this specific case with 2d75698, but that doesn't address the root cause.

Problem with this bug is that I don't know how to craft a test for it! How would I create a test for it?

uenoku (approved):
LGTM, agree with copilot suggestion. For test probably you can create unittest under HW something like this https://github.com/llvm/circt/blob/2d75698c6396769afa084b80d133c4ad0532ed6d/unittests/Dialect/RTG/MaterializerTest.cpp.

teqdruid:
> For test probably you can create unittest under HW something like this

I totally forgot we had C++ unit tests!

teqdruid:
Thanks @uenoku !

PR #8688, opened 2025-07-11
[MooreToCore] Lower empty string_constant to expected bit width.

mvpant:
Fixes type bit width mismatch when lowering empty string constants.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -605,13 +605,17 @@ struct StringConstantOpConv : public OpConversionPattern<StringConstantOp> {
   matchAndRewrite(moore::StringConstantOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     const auto str = op.getValue();
-    const unsigned byteWidth = str.size() * 8;
+    unsigned byteWidth = str.size() * 8;
     const auto resultType =
         typeConverter->convertType(op.getResult().getType());
     if (const auto intType = mlir::dyn_cast<IntegerType>(resultType)) {
       if (intType.getWidth() < byteWidth) {
         return rewriter.notifyMatchFailure(op,
                                            "invalid string constant type size");
+      }
+      // Zero-fill empty string constant to expected type size in bits.
+      if (byteWidth == 0) {
+        byteWidth = intType.getWidth();
       }
     } else {
       return rewriter.notifyMatchFailure(op, "invalid string constant type");
--- a/test/Conversion/MooreToCore/basic.mlir
+++ b/test/Conversion/MooreToCore/basic.mlir
@@ -1049,6 +1049,8 @@ moore.module @StringConstant() {
     %str = moore.string_constant "Test" : i32
     // CHECK: hw.constant 0 : i0
     %str_empty = moore.string_constant "" : i0
+    // CHECK: hw.constant 0 : i8
+    %str_empty_extended = moore.string_constant "" : i8
     moore.return
   }
 }


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -605,14 +605,18 @@ struct StringConstantOpConv : public OpConversionPattern<StringConstantOp> {
...
         typeConverter->convertType(op.getResult().getType());
     if (const auto intType = mlir::dyn_cast<IntegerType>(resultType)) {
       if (intType.getWidth() < byteWidth) {
         return rewriter.notifyMatchFailure(op,
                                            "invalid string constant type size");
       }
+      // Zero-fill empty string constant to expected type size in bits.
+      if (byteWidth == 0) {
+        byteWidth = intType.getWidth();
+      }
```
maerhart: Shouldn't that apply in all cases instead of just `byteWidth == 0`?
@fabianschuiki I'm assuming `string_constant` is supposed to allow implicit zext since it is not restricted in its verifier?

mvpant: I assume that it can be applied in all cases as long as 1) result type bit width is multiple of 8  (bit width % 8 == 0), 2) result type bit width >= str.size() * 8 (can fit the string value). Otherwise, to be conservative, we should probably emit a warning.

fabianschuiki: Thanks for working on this @mvpant!

It looks like the Slang frontend will always generate `len(str)*8` as type for `"..."` literals, except for the empty `""` literal, which is counted as 1 byte. Feels like a workaround for SV's lack of 0-width types.

```systemverilog
module Foo(
  output bit [12:0] x,
  output bit [12:0] y,
  output bit [12:0] z
);
  assign x = "";
  assign y = "a";
  assign z = "ab";
endmodule
```
produces the following after `circt-verilog --ir-moore`:
```mlir
moore.module @Foo(out x : !moore.i13, out y : !moore.i13, out z : !moore.i13) {
  %0 = moore.string_constant "" : i8
  %1 = moore.zext %0 : i8 -> i13
  %2 = moore.string_constant "a" : i8
  %3 = moore.zext %2 : i8 -> i13
  %4 = moore.string_constant "ab" : i16
  %5 = moore.trunc %4 : i16 -> i13
  moore.output %1, %3, %5 : !moore.i13, !moore.i13, !moore.i13
}
```
Looks like the frontend already inserts zero-extension and truncation as needed, with the `""` case being the weird one that doesn't follow `len(str)*8`.

I see two options:
1. add a verifier to `moore.string_constant` that forces the result type to be exactly `len(str)*8` wide, and then special-case zero-length strings in ImportVerilog directly; or
2. zero-extend the characters in this conversion, similar to how `moore.constant` integers don't require you to make use of all bits in the result.

I think just zero-extending is sensible :+1:, keeping things in line with integer constants. We should probably not even check whether the string is too small or too big in this conversion, and instead just fill the characters into an APInt that is _exactly_ as wide as expected by the result. Any errors about the string being wider than fits the result could go into the `moore.string_constant` verifier, similar to how `moore.constant` complains about values that don't fit into the result type.

Maybe all that's neede here is to not use a `byteWidth` for the APInts, but use `intType.getWidth()` directly. The loop then pushes the characters into the APInt 8 bits at a time, regardless of whether there are fewer or more characters than fits. That makes the conversion behave very predictably, and we can add any errors to the verifier if we want :smiley:

What do you guys think?

maerhart: That's a great analysis! So that means `""` is equivalent to `"\0"` in Verilog?
I think both approaches are reasonable with a slight preference for the more restrictive verifier and inserting an additional zext for empty strings in ImportVerilog to keep things as explicit and easy to understand as possible.

mvpant: I find the second option more fitting (possibly due to my lack of project knowledge), as the problem arose on a Moore ---> Core path. Setting up an APInt as large as the expected type, initializing it with a default value of 0, feeding it all the characters, and then verifying `hw::ConstantOp` soundness in the verifier sounds good to me.

mvpant: It seems we can't blindly use the expected type's bitwidth as an APInt parameter, since there is an assert that checks the shift amount,
```
/include/llvm/ADT/APInt.h:786: APInt &llvm::APInt::operator<<=(unsigned int): Assertion `ShiftAmt <= BitWidth && "Invalid shift amount"' failed.
```
which cannot exceed its bitwidth. So for now, I'm taking the maximum of the expected type's bitwidth and the string length * 8, and making that an attribute type, so that the verifier can later check for any mismatch between the return type's bitwidth and the attribute value's bitwidth.

Implicitly truncating characters by fitting only those that fit into the expected result type is also an option, but I'm not sure it's a good one, as errors could easily go unnoticed. An explicit `trunc` at the Moore level should be used for such cases instead.

fabianschuiki: Hmmm, I'd just go with the exact result type the op guarantees. Instead, I'd make the loop stop as soon as the shift amount goes beyond the width of the type, where the additional characters shifted up would just be lost anyway.

Generally, the conversion should just translate from one dialect to another. But it's up to the op verifier to enforce any guarantees that must be upheld for the op to make sense. I wouldn't introduce any error messages in the conversion itself. The conversion should just work in a predictable, consistent way. The errors get presented to the user, and conversion errors are often pretty cryptic and hard to fix for the user. Instead, if we really need any guarantees, we should probably add them as constraints to the op such that they pop up very early on when the op gets created, e.g. during parsing of the user's input.

In this case I'd just remove all error messages from the conversion, and make the conversion populate the APInt of the precise result bit width as best as possible.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -604,23 +604,26 @@ struct StringConstantOpConv : public OpConversionPattern<StringConstantOp> {
...
-    if (const auto intType = mlir::dyn_cast<IntegerType>(resultType)) {
-      if (intType.getWidth() < byteWidth) {
-        return rewriter.notifyMatchFailure(op,
-                                           "invalid string constant type size");
-      }
-    } else {
+    const auto intType = mlir::dyn_cast<IntegerType>(resultType);
+    if (!intType) {
       return rewriter.notifyMatchFailure(op, "invalid string constant type");
     }
```
fabianschuiki: Nit: I think this message is not needed (probably a holdover from earlier Moore dialect days), since the `StringConstantOp` verifier verifies that its result type is a `moore::IntType`. Instead of `dyn_cast`, a `cast` works :smiley: 

```suggestion
    const auto intType = mlir::cast<IntegerType>(resultType);
```

fabianschuiki (approved):
Really cool, thanks a lot! LGTM. Thanks for adding all the test cases :raised_hands:.

lib/Conversion/MooreToCore/MooreToCore.cpp
mvpant: Oh, sure, missed that. Thanks (Let's assume `typeConverter` doesn't do anything weird with integer input type)

fabianschuiki:
Thanks @mvpant, really cool stuff! Landing this 🚀

PR #8689, opened 2025-07-11
[VerifToSMT] Fix lowering of initial integer values for BMC

fzi-hielscher:
Fixes a problem during VerifToSMT lowering when creating the `smt::BVConstantOp`s for initial integer values of registers in the BMC operation. The current implementation calls `getSExtValue()` on the `APInt` provided by the initial value attribute. This causes a crash in either of the following situations:
- The initial value type is lager than 64 bits
- The initial value type is smaller than 64 bits and the MSB of the value is set. By sign-extending to 64 bits, the resulting integer value exceeds the width of the created BitVector.
 
This is fixed by passing the `APInt` directly to the builder of `smt::BVConstantOp` and adding a check beforehand ensuring that the types of the initial attribute and the initialized register match.

### DIFF
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -312,7 +312,7 @@ struct VerifBoundedModelCheckingOpConversion
           const auto &cstInt = initIntAttr.getValue();
           assert(cstInt.getBitWidth() ==
                      cast<smt::BitVectorType>(newTy).getWidth() &&
-                 "With mismatch between initial value and target type");
+                 "Width mismatch between initial value and target type");
           inputDecls.push_back(rewriter.create<smt::BVConstantOp>(loc, cstInt));
           continue;
         }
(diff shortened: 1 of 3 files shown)


lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -309,9 +309,11 @@ struct VerifBoundedModelCheckingOpConversion
         auto initVal =
             initialValues[curIndex - oldCircuitInputTy.size() + numRegs];
         if (auto initIntAttr = dyn_cast<IntegerAttr>(initVal)) {
-          inputDecls.push_back(rewriter.create<smt::BVConstantOp>(
-              loc, initIntAttr.getValue().getSExtValue(),
-              cast<smt::BitVectorType>(newTy).getWidth()));
+          const auto &cstInt = initIntAttr.getValue();
+          assert(cstInt.getBitWidth() ==
+                     cast<smt::BitVectorType>(newTy).getWidth() &&
+                 "With mismatch between initial value and target type");
```
maerhart:
```suggestion
                 "Width mismatch between initial value and target type");
```

TaoBi22 (approved):
Thanks!

PR #8697, opened 2025-07-12
[AIG] Add canonicalization to simplify inversion

uenoku:
Add AIG double inversion canonicalization pattern.

This test verifies the new canonicalization pattern that flattens nested and_inv operations with single inverted inputs. The pattern handles cases where an and_inv operation takes as input another and_inv operation with a single inverted operand, effectively eliminating redundant inversions.


fabianschuiki (approved):
LGTM!

PR #8698, opened 2025-07-12
[circt-bmc] Support `seq.firreg` with sync reset

liuyic00:
This adds support for `seq.firreg` with synchronous reset in `ExternalizeRegistersPass`.

Asynchronous resets are not supported because currently there isn't a way to update a register on the negedge.
When the async reset signal becomes true before the negedge and turns false before the posedge, register should be reset and store the reset value. This needs to specify and update the register on the negedge in `VerifToSMT`.
How to implement this need further discussion, maybe pass more information to `VerifToSMT`, or first simplify the clock and circuit into a transition-system-like form.


lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -150,6 +150,69 @@ void ExternalizeRegistersPass::runOnOperation() {
           ++numRegs;
           return;
         }
+        if (auto regOp = dyn_cast<seq::FirRegOp>(op)) {
```
TaoBi22, line 129: This whole statement is a lot of code duplication that it's probably better to avoid. I believe all the differences between the two cases are just having to call different getters to fetch similar information about the respective registers right (e.g. getInitial vs getPreset)? It would probably be neater to have some generic register lowering function that takes the input, initial value, reset etc. as arguments - then, you just need to have a case for each kind of register where you extract these values and call the method.

test/Tools/circt-bmc/externalize-registers-errors.mlir
```diff
@@ -54,3 +54,11 @@ hw.module @reg_with_instance_initial(in %clk: !seq.clock, in %in: i32, out out:
...
   hw.output %1 : i32
 }
+
+// -----
+hw.module @firreg_with_async_reset(in %clk: !seq.clock, in %rst: i1, in %in: i32, out out: i32) {
+  %c0_i32 = hw.constant 0 : i32
+  // expected-error @below {{seq.firreg with async reset not yet supported}}
+  %1 = seq.firreg %in clock %clk reset async %rst, %c0_i32 : i32
+  hw.output %1 : i32
+}
```
TaoBi22: nit: this file is missing a newline at the end

test/Tools/circt-bmc/externalize-registers.mlir
```diff
@@ -98,3 +98,48 @@ hw.module @reg_with_reset(in %clk: !seq.clock, in %rst: i1, in %in: i32, out out
...
+  %0 = comb.add %in0, %in1 : i32
+  %single_reg = seq.firreg %0 clock %clk preset 0 : i32
+  hw.output %single_reg : i32
+}
+
+// CHECK:  hw.module @combreg_and_firreg(in [[CLK:%.+]] : !seq.clock, in [[IN0:%.+]] : i32, in [[IN1:%.+]] : i32, in [[OLD_REG0:%.+]] : i32, in [[OLD_REG1:%.+]] : i32, out {{.+}} : i32, out {{.+}} : i32, out {{.+}} : i32) attributes {initial_values = [unit, unit], num_regs = 2 : i32} {
+// CHECK:    [[ADD:%.+]] = comb.add [[IN0]], [[IN1]]
+// CHECK:    hw.output [[OLD_REG1]], [[ADD]], [[OLD_REG0]]
+// CHECK:  }
+hw.module @combreg_and_firreg(in %clk: !seq.clock, in %in0: i32, in %in1: i32, out out: i32) {
```
TaoBi22:
```suggestion
hw.module @compreg_and_firreg(in %clk: !seq.clock, in %in0: i32, in %in1: i32, out out: i32) {
```
(same change necessary in CHECK comment too)

TaoBi22 (commented):
Thanks! This mostly looks good, but it would be nice to not introduce so much duplicated code - see the comment below for some more detail on that

liuyic00:
Thanks for the advice!
I extracted a function `externalizeReg` to handle the common part.
I also renamed the new register value from `xxx_input` to `xxx_next`, because `input` could be confused with module input, and `xxx_next` aligns better with `xxx_state`.

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -210,3 +207,52 @@ void ExternalizeRegistersPass::runOnOperation() {
...
+  addedInputNames[module.getSymNameAttr()].push_back(newInputName);
+  addedOutputs[module.getSymNameAttr()].push_back(next.getType());
+  addedOutputNames[module.getSymNameAttr()].push_back(newOutputName);
+
+  // Replace the register with newInput and newOutput
+  auto newInput = module.appendInput(newInputName, regType).second;
+  if (reset) {
+    if (isAsync) {
+      // Async reset
+      op->emitError("registors with async reset not yet supported");
```
TaoBi22:
```suggestion
      op->emitError("registers with an async reset are not yet supported");
```

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
```diff
@@ -210,3 +207,52 @@ void ExternalizeRegistersPass::runOnOperation() {
...
+      return failure();
+    }
+    // Sync reset
+    result.replaceAllUsesWith(newInput);
```
TaoBi22: nit: we have this in both branches of the if, could it be moved outside?

test/Tools/circt-bmc/externalize-registers-errors.mlir
```diff
@@ -54,3 +54,11 @@ hw.module @reg_with_instance_initial(in %clk: !seq.clock, in %in: i32, out out:
   %1 = seq.compreg %in, %clk initial %init : i32
   hw.output %1 : i32
 }
+
+// -----
+hw.module @firreg_with_async_reset(in %clk: !seq.clock, in %rst: i1, in %in: i32, out out: i32) {
+  %c0_i32 = hw.constant 0 : i32
+  // expected-error @below {{registors with async reset not yet supported}}
```
TaoBi22:
```suggestion
  // expected-error @below {{registers with an async reset are not yet supported}}
```

TaoBi22 (approved):
This all looks sensible to me, thanks for the changes!!

lib/Tools/circt-bmc/ExternalizeRegisters.cpp
liuyic00: I somehow added the if-else back for sync/async reset (which was removed by clang-tidy 😢). Hope this looks better.
The use of reg should be replaced with another `mux` when processing async reset (if we could), so I didn't put it outside the branch.

TaoBi22: Sorry, I might not have been clear! I just mean that we have an `if (reset)` but then we call `result.replaceAllUsesWith(newInput);` regardless of the branch, so it would be nice to move this above the if.

I don't think the if that clang-tidy dropped is a big deal, I think clang-tidy's just dropping it since we return if we hit the error anyway,

liuyic00: The `result.replaceAllUsesWith(newInput);` has been moved above the `if`.
You made it very clear the first time, I was just trying a different solution. Thanks for being so patient!

TaoBi22:
Great, thank you for the changes @liuyic00! Are you happy for me to hit merge?

liuyic00:
Yes please! Thank you so much!!

PR #8700, opened 2025-07-13
[Bazel] Add Bazel support to build CIRCT

SihaoLiu:
Hi CIRCT team,

I think bazel is a nice tool in terms of building large project especially something like CIRCT.

This PR strictly follows the LLVM's bazel support style, which means all bazel build support files is under `utils/bazel` like https://github.com/llvm/llvm-project/tree/main/utils/bazel

A obvious nice about this bazel support is that it does not requires `LLVM` as git submodule and bazel handles every dependency (kind of like cargo style).


fabianschuiki:
Is there any way to auto-generate these Bazel files from the CMake setup and the submodules? The main thing that comes to my mind is that this essentially duplicates the entire CMake setup and the submodule hash. Who would maintain this in order to prevent the CMake build and submodule from falling out of sync with the Bazel build? How does LLVM handle this?

uenoku:
Yeah on LLVM contributors are not required to update bazel, and usually when something was broken bazel users (mainly Google folks) are fixing up the bazel config. I appreciate proposal but at this point there is few bazel user of CIRCT so probably we cannot support bazel in upstream.

SihaoLiu:
> Yeah on LLVM contributors are not required to update bazel, and usually when something was broken bazel users (mainly Google folks) are fixing up the bazel config. I appreciate proposal but at this point there is few bazel user of CIRCT so probably we cannot support bazel in upstream.

Yeah, totally understand that when I realized that there is a team from Google to maintain bazel build of LLVM work. I will try to maintain a bazel support in my own repo https://github.com/SihaoLIu/circt

PR #8701, opened 2025-07-14
[Verif][LEC] Make LECOp result optional to avoid unsafe conversion

fzi-hielscher:
#8497 introduced a lowering pattern for the `LogicEquivalenceCheckingOp` which changes behavior depending on whether the result of the operation is used. [This is generally not safe](https://discourse.llvm.org/t/rfc-a-new-one-shot-dialect-conversion-driver/79083) and can cause lowerings to erroneously consider the result to be used/unused. The pattern must only depend on the operation itself, not on its def-use context.

By making the result of the operation optional, we should be able to retain the current behavior. In the pattern `op.use_empty()` is simply replaced by `op.getNumResults() == 0`. However, we now have to decide at the point of creating the `LogicEquivalenceCheckingOp` whether the result should be made accessible to the IR.

### DIFF
--- a/include/circt/Dialect/Verif/VerifOps.td
+++ b/include/circt/Dialect/Verif/VerifOps.td
@@ -243,7 +243,7 @@ def LogicEquivalenceCheckingOp : VerifOp<"lec", [
 
     The operation can return an boolean result that is `true` iff equivalence
     of the two circuits has been proven. The result can be omitted for use-cases
-    which do not allow further processing (e.g., SMTLib exporting).
+    which do not allow further processing (e.g., SMT-LIB exporting).
   }];
 
   let results = (outs Optional<I1>:$areEquivalent);


include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -236,20 +236,30 @@ def LogicEquivalenceCheckingOp : VerifOp<"lec", [
...
     constructed and compared for identity, etc.
-    
+
     The number and types of the inputs and outputs of the two circuits (and thus
     also the block arguments and yielded values of both regions) have to match.
     Otherwise, the two should be considered trivially non-equivalent.
+
+    The operation can return an boolean result that is `true` iff equivalence
+    of the two circuits has been proven. The result can be omitted for use-cases
+    which do not allow further processing (e.g., SMTLib exporting).
```
TaoBi22: super-nit
```suggestion
    which do not allow further processing (e.g., SMT-LIB exporting).
```

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -236,20 +236,30 @@ def LogicEquivalenceCheckingOp : VerifOp<"lec", [
...
-  let results = (outs I1:$areEquivalent);
+  let results = (outs Optional<I1>:$areEquivalent);
   let regions = (region SizedRegion<1>:$firstCircuit,
                         SizedRegion<1>:$secondCircuit);
 
   let assemblyFormat = [{
-    attr-dict `first` $firstCircuit `second` $secondCircuit
+    attr-dict (`:` type($areEquivalent)^)?
+    `first` $firstCircuit
+    `second` $secondCircuit
```
TaoBi22, line 256: Nit, but having the type before the regions looks a bit weird to me - I would expect the syntax to be like that of e.g., `seq.initial` and `smt.check`?

TaoBi22 (approved):
This looks sensible to me, thanks for catching that what we have now is unsafe!!

include/circt/Dialect/Verif/VerifOps.td
fzi-hielscher: I faintly remember having odd parsing problems with trailing optional type specifiers. It is probably an unjustified concern here, but I deliberately put it in a position where it is clearly delimited by the `first` keyword.

PR #8702, opened 2025-07-14
Fix filecheck directive typos and fix now-active test lines

dtzSiFive:


dtzSiFive:
See https://github.com/llvm/circt/pull/7058 for example invoking filecheck lint utility from upstream on our tests.

PR #8703, opened 2025-07-14
[hw] Convert HW Passes to use ODS constructors

seldridge:
Change all HW passes to use ODS consturctors as opposed to hand-rolling
them.  The hand-rolled constructors aren't necessary and add a bunch of
boilerplate.

Note: a number of these passes do not have users and hence may break out
of tree users.  The out of tree users are likely @teqdruid and @mortbopet.


fabianschuiki (approved):
LGTM! Thanks for cleaning this up :raised_hands: :100:

PR #8704, opened 2025-07-14
[hw] Add container op

seldridge:
Add a new `hw.container` op.  This is intended to be a generic container
of hardware.  Critically, this exists to enable the checking of ops that
are users of inner symbol tables.  This is shown with the one error test
case that shows validation of a `hw.hierpath` op with an invalid inner
symbol use.  (This will not be checked if a `builtin.module` is used.)

Note: this is intended to be temporarily backwards-compatible with existing operations. This is done by allowing `hw.module` and friends to have a parent that is either a `builtin.module` or a `hw.container`. As long as there is buy-in on this approach, I will then go about updating all tests, migrating things over, and requiring that `hw.module` and friends have a `hw.container` parent.

### DIFF
--- a/include/circt/Dialect/HW/HWStructure.td
+++ b/include/circt/Dialect/HW/HWStructure.td
@@ -26,6 +26,22 @@ include "mlir/IR/SymbolInterfaces.td"
 include "mlir/Interfaces/ControlFlowInterfaces.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
 
+def ContainerOp : HWOp<"container",
+      [IsolatedFromAbove, SymbolTable, SingleBlock, NoTerminator,
+       NoRegionArguments, InnerRefNamespace]> {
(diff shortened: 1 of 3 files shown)


uenoku:
For temporary migration, are you going to change ExportVerilog etc to allow both container op and builtin.module? In that case we will need to either make passes op-agnostic or tweak `canScheduleOn` manually a lot of places.

seldridge:
> For temporary migration, are you going to change ExportVerilog etc to allow both container op and builtin.module?

I had thought that something like this could work, but didn't know the particulars of how to do it.

That said, Will did almost a full migration a while ago (https://github.com/llvm/circt/compare/main...dtzSiFive:circt:feature/hw-design-op) and that touched a lot of files, but wasn't conceptually that big of a diff. So, maybe we just get consensus and then do a hard switch?

PR #8709, opened 2025-07-16
[AIG] Add slice indexing support to LongestPathCollection in AIG Pyth…

uenoku:
…on bindings

This commit implements slice indexing functionality for the LongestPathCollection class, allowing users to access multiple timing paths using Python's standard slice notation. The implementation handles all slice types including negative indices, step values, and empty slices through Python's built-in slice.indices() method.


fabianschuiki (approved):
LGTM!

PR #8710, opened 2025-07-16
[RTG] Redefine RandomNumberInRangeOp upper bound to be inclusive

maerhart:
When we want to be able to get all possible numbers that fit in a 64 bit register, we would need a 65 bit number as the upper bound. However, an iindex typed attribute uses an APInt with 64 bits and would thus make this complicated.


fabianschuiki (approved):
LGTM

PR #8711, opened 2025-07-16
[RTG] Add custom tuple type to support empty tuples

maerhart:
The builtin tuple type does not allow empty tuples, but they can avoid special casing in the frontend and can also be used as indicators (something like a none type could otherwise be used for but which we don't have in RTG).


test/Dialect/RTG/IR/basic.mlir
```diff
@@ -188,11 +188,13 @@ rtg.test @tuples() {
...
   // CHECK-NEXT: [[TUPLE:%.+]] = rtg.tuple_create [[IDX0]], [[TRUE]] : index, i1
-  // CHECK-NEXT: rtg.tuple_extract [[TUPLE]] at 1 : tuple<index, i1>
+  // CHECK-NEXT: rtg.tuple_extract [[TUPLE]] at 1 : !rtg.tuple<index, i1>
+  // CHECK-NEXT: rtg.tuple_create
   %idx0 = index.constant 0
   %true = index.bool.constant true
   %0 = rtg.tuple_create %idx0, %true : index, i1
-  %1 = rtg.tuple_extract %0 at 1 : tuple<index, i1>
+  %1 = rtg.tuple_extract %0 at 1 : !rtg.tuple<index, i1>
+  rtg.tuple_create
```
fabianschuiki, line 198: It might be worth adding an empty tuple as an argument to the `rtg.test`, just to make sure that the empty tuple type round-trips. Unless I've overlooked a use of it somewhere 🤔

fabianschuiki (approved):
LGTM!

PR #8712, opened 2025-07-16
[RTG] Enable conditional value forwarding for ValidateOp

maerhart:


include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -801,16 +803,41 @@ def ValidateOp : RTGOp<"validate", [
     a simple integer). Therefore, it is strongly recommended to not use the
     result value of this operation in situations that expect structural
     equivalence checks such as adding it to sets or bags.
+
+    The 'defaultUsedValues' are forwarded to the 'values' results without any
+    modification whenever the 'defaultValue' is used as replacement for 'value'.
+    Otherwise, the 'elseValues' are forwarded. This can be used to conditionally
+    execute code based on whether the default value was used or a proper value
+    was used as replacement. Note that this not the most light-weight
```
fabianschuiki:
```suggestion
    was used as replacement. Note that this is not the most light-weight
```

include/circt/Dialect/RTG/IR/RTGOps.td
```diff
@@ -801,16 +803,41 @@ def ValidateOp : RTGOp<"validate", [
...
+                       Variadic<AnyType>:$elseValues);
+  let results = (outs AnyType:$value, Variadic<AnyType>:$values);
 
   let assemblyFormat = [{
-    $ref `,` $defaultValue (`,` $id^)? `:`
-    type($ref) `->` type($value) attr-dict
+    $ref `,` $defaultValue (`,` $id^)?
+    (` ``(` $defaultUsedValues^ `else` $elseValues `:`
+    qualified(type($defaultUsedValues)) `)`)? `:`
+    qualified(type($ref)) `->` qualified(type($defaultValue)) attr-dict
```
fabianschuiki, line 840: Do the types need to be qualified?

maerhart: No, but they are qualified for all other operations, so I added it for consistency.

PR #8713, opened 2025-07-16
[Verif] Add RefinementCheckingOp

fzi-hielscher:
This PR adds the `RefinementCheckingOp` to the `verif` dialect. The motivation behind this operation is to be able to automatically check whether a 'target' circuit is a refinement of a 'source' circuit. This should be a small step towards performing translation validation comparable to [Alive2](https://users.cs.utah.edu/~regehr/alive2-pldi21.pdf). The operation is structurally identical to the `LogicEquivalenceCheckingOp`, so I factored out most of the ODS into a common `CircuitRelationCheckOp`. If there is no non-determinism present in the circuits, the operation is also functionally identical to `LogicEquivalenceCheckingOp`.

Finding a good definition of the `refines` relation is hard, so please just consider this as a humble first attempt. In essence, there are (at least) three factors to consider:
- Does it permit all optimizations we want to be able to do?
- Does it reject all "optimizations" we want to be illegal?
- Is it viable to perform model checking with the tools available to us?

Our current definition of logical equivalence is too restrictive for translation validation. In the presence of non-determinism it is not guaranteed that `A == A`, see #7025 . Stating that a circuit cannot be translated to itself is problematic.
In contrast, the `RefinementCheckingOp` as described here compares the sets of possible outputs for a given input. This is 
potentially too permissive, as it allows a different non-deterministic value to be picked for each possible input assignment.

Note that the [current specification](https://circt.llvm.org/docs/Dialects/Comb/RationaleComb/#undefined-values) of the undefined value in `comb` puts the non-determinism in the domain of functions, getting us dangerously close to second-order logic. While I think it is a good model for the likely behavior of real hardware, it is probably impractical for SMT modeling.

### DIFF
--- a/include/circt/Dialect/Verif/VerifOps.td
+++ b/include/circt/Dialect/Verif/VerifOps.td
@@ -224,14 +224,31 @@ def BoundedModelCheckingOp : VerifOp<"bmc", [
   let hasRegionVerifier = true;
 }
 
-def LogicEquivalenceCheckingOp : VerifOp<"lec", [
-  IsolatedFromAbove,
-  SingleBlockImplicitTerminator<"verif::YieldOp">,
-]> {
+// Attempt to prove a binary relation between two circuits
+class CircuitRelationCheckOp<string mnemonic, list<Trait> traits = []>
+  : VerifOp<mnemonic, !listconcat(traits,
+    [IsolatedFromAbove, SingleBlockImplicitTerminator<"verif::YieldOp">])>
+{
+  let results = (outs Optional<I1>:$isProven);
+  let regions = (region SizedRegion<1>:$firstCircuit,
+                        SizedRegion<1>:$secondCircuit);
+
+  let assemblyFormat = [{
+    attr-dict (`:` type($isProven)^)?
+    `first` $firstCircuit
+    `second` $secondCircuit
+  }];
+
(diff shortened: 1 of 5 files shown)


uenoku:
Thank you for working on this! I think this is great extension. I have actually encountered this problem in translation validation of synthesis lowering. I end up adding verif.assume to prevent out-of-bounds access: https://github.com/llvm/circt/blob/5ee54acb0538127525527f334caa5cec7f3d286c/integration_test/circt-synth/array-lec.mlir#L13-L17

include/circt/Dialect/Verif/VerifOps.td
```diff
@@ -241,25 +258,37 @@ def LogicEquivalenceCheckingOp : VerifOp<"lec", [
...
+    IR. Given two (purely combinational) circuits A and B with the same
+    signature, B refines A iff for all inputs the set of possible output
+    values of B is a subset of the possible output values of A given the
+    same input.
+
+    For strictly deterministic circuits the 'refines' relation is identical to
+    logical equivalence. Informally speaking, refining allows maintaining or
+    reducing the non-determinism of a circuit.
+
+    If the signatures of the circuits does not match, the second circuit is
```
TaoBi22:
```suggestion
    If the signatures of the circuits do not match, the second circuit is
```

PR #8716, opened 2025-07-16
[Seq] Add a pass to convert an array seq.firreg to seq.firmem

prithayan:
This PR introduces a new transformation pass `RegOfVecToMem` that converts register arrays following memory access patterns into `seq.firmem` operations.

This is required for the `circt-verilog` tool, to identify memories, such that other `circt` transformations/analysis can be run in the `seq` dialect on the `mlir` parsed from verilog.
 
The pass analyzes `seq.firreg` operations with array types and identifies memory access patterns consisting of:
- A register array representing memory storage
- `hw.array_get` operations for reads  
- `hw.array_inject` operations for writes
- `comb.mux` operations controlling read/write enables
- Optional output registers for read data

When a valid pattern is detected, the pass replaces the register array with a `seq.firmem` operation and corresponding read/write ports.

TODO: 
1. This pass cannot extract the read enable pattern yet.
2. Parallelize the implementation.

## Example Transformation

**Input:**
````mlir
hw.module @memory_example(in %clk: i1, in %addr: i13, in %data: i8, in %we: i1, out result: i8) {
  %clock = seq.to_clock %clk
  
  // Memory implemented as register array
  %mem_next = comb.mux %we, %mem_updated, %mem : !hw.array<8192xi8>
  %mem_updated = hw.array_inject %mem[%addr], %data : !hw.array<8192xi8>, i13
  %mem = seq.firreg %mem_next clock %clock : !hw.array<8192xi8>
  
  // Read operation
  %read_data = hw.array_get %mem[%addr] : !hw.array<8192xi8>, i13
  
  hw.output %read_data : i8
}
````

**Output:**
````mlir
hw.module @memory_example(in %clk: i1, in %addr: i13, in %data: i8, in %we: i1, out result: i8) {
  %clock = seq.to_clock %clk
  
  // Memory implemented as seq.firmem
  %mem = seq.firmem 0, 1, undefined, undefined : <8192 x 8, mask 1>
  %read_data = seq.firmem.read_port %mem[%addr], clock %clock enable %true
  seq.firmem.write_port %mem[%addr] = %data, clock %clock enable %we
  
  hw.output %read_data : i8
}
````

### DIFF
--- a/include/circt/Dialect/Seq/SeqPasses.td
+++ b/include/circt/Dialect/Seq/SeqPasses.td
@@ -121,7 +121,6 @@ def RegOfVecToMem : Pass<"seq-reg-of-vec-to-mem", "hw::HWModuleOp"> {
     3. Updates are controlled by enable signals through mux operations
     4. Read and write operations use the same clock
   }];
-  let constructor = "circt::seq::createRegOfVecToMemPass()";
 }
 
 #endif // CIRCT_DIALECT_SEQ_SEQPASSES
--- a/include/circt/Dialect/Seq/SeqPasses.td
+++ b/include/circt/Dialect/Seq/SeqPasses.td
@@ -111,4 +111,17 @@ def LowerSeqShiftReg : Pass<"lower-seq-shiftreg", "hw::HWModuleOp"> {
   let dependentDialects = ["circt::hw::HWDialect"];
 }
 
+def RegOfVecToMem : Pass<"seq-reg-of-vec-to-mem", "hw::HWModuleOp"> {
+  let summary = "Convert register arrays to FIRRTL memories";
+  let description = [{
+    This pass identifies register arrays that follow memory access patterns
+    and converts them to seq.firmem operations. It looks for patterns where:
+    1. A register array is updated via array_inject operations
+    2. The array is read via array_get operations  
+    3. Updates are controlled by enable signals through mux operations
+    4. Read and write operations use the same clock
+  }];
+  let constructor = "circt::seq::createRegOfVecToMemPass()";
+}
+
 #endif // CIRCT_DIALECT_SEQ_SEQPASSES
--- a/lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
+++ b/lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
@@ -232,7 +232,8 @@ void RegOfVecToMemPass::runOnOperation() {
   // Erase all marked operations
   for (auto *op : opsToErase) {
     LLVM_DEBUG(llvm::dbgs()
-               << "Erasing operation: " << *op << " number of uses:" << "\n");
+               << "Erasing operation: " << *op << " number of uses:"
+               << "\n");
     op->dropAllUses();
     op->erase();
   }
--- a/lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
+++ b/lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
@@ -1,0 +1,244 @@
+//===- RegOfVecToMem.cpp - Convert Register Arrays to Memories -----------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This transformation pass converts register arrays that follow memory access
+// patterns to seq.firmem operations.
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/Seq/SeqOps.h"
+#include "circt/Dialect/Seq/SeqPasses.h"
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: test/Dialect/Seq/reg-of-vec-to-mem.mlir [hunk 1]
```diff
@@ -0,0 +1,51 @@
+// RUN: circt-opt %s --seq-reg-of-vec-to-mem | FileCheck %s
+
+// CHECK-LABEL: hw.module private @sf_tagmem_external_mem_a
+hw.module private @sf_tagmem_external_mem_a(in %CLK : i1, in %D : i46, in %ADR : i13, in %WE : i1, in %ME : i1, out Q : i46) {
```
fabianschuiki: Maybe something like `@complex_mem` to keep things short?

lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
```diff
@@ -0,0 +1,245 @@
...
+            LLVM_DEBUG(llvm::dbgs()
+                       << "  Found output register: " << outputReg << "\n");
+            pattern.outputReg = outputReg;
+            break;
+          }
+        }
+      }
+      break;
+    }
+  }
```
fabianschuiki: This loop might have to go through every user to ensure that there's only a single `array_get` using the register. Otherwise, replacing the register would potentially break because there are other users.

It's probably worthwhile to add a test that has a `seq.firreg` that looks like it qualifies for `seq.firmem` replacement, but has its value returned as an output or some other use besides an `hw.array_get`, which should then block the replacement.

lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  LLVM_DEBUG(llvm::dbgs() << "  Created write port\n");
+
+  // Replace read access
+  if (pattern.outputReg)
+    // If there's an output register, replace its input
+    pattern.outputReg.getNext().replaceAllUsesWith(readData);
+  else
+    // Replace direct read access
+    pattern.readAccess.getResult().replaceAllUsesWith(readData);
```
fabianschuiki, line 218: Does the pass need to distinguish the output register from a regular output read? As in, would it be sufficient to replace all uses of the `array_get` that reads from the memory with the `firmem.read` op?

lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
```diff
@@ -0,0 +1,245 @@
...
+  pattern.clock = reg.getClk();
+
+  // Find the mux that drives this register
+  auto nextValue = reg.getNext();
+  auto mux = nextValue.getDefiningOp<comb::MuxOp>();
+  if (!mux)
+    return false;
+
+  LLVM_DEBUG(llvm::dbgs() << "  Found driving mux: " << mux << "\n");
+  pattern.writeMux = mux;
```
fabianschuiki, line 120: Since the mux gets deleted later, it's probably necessary to check that the mux result isn't used anywhere outside this specific memory pattern. Might be worth adding a test for this.

fabianschuiki (commented):
Really cool!

include/circt/Dialect/Seq/SeqPasses.td
```diff
@@ -111,4 +111,17 @@ def LowerSeqShiftReg : Pass<"lower-seq-shiftreg", "hw::HWModuleOp"> {
...
+  let summary = "Convert register arrays to FIRRTL memories";
+  let description = [{
+    This pass identifies register arrays that follow memory access patterns
+    and converts them to seq.firmem operations. It looks for patterns where:
+    1. A register array is updated via array_inject operations
+    2. The array is read via array_get operations  
+    3. Updates are controlled by enable signals through mux operations
+    4. Read and write operations use the same clock
+  }];
+  let constructor = "circt::seq::createRegOfVecToMemPass()";
```
uenoku: nit: please use ODS generated one

lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
prithayan: You are right. Added the check and a test.

lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
prithayan: I think it makes sense to make the check more constrained. The register should only have three users. Added test.

lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
prithayan: I think the `array_get` users shouldn't matter. The memory read is enabled by default, it should be possible to replace all the users. Do you have any concerns here ?

test/Dialect/Seq/reg-of-vec-to-mem.mlir
prithayan: Right, copy paste was a bad idea here.

lib/Dialect/Seq/Transforms/RegOfVecToMem.cpp
fabianschuiki: Sounds great :+1:

fabianschuiki (approved):
Very nice! LGTM :+1:

PR #8717, opened 2025-07-17
[AIG] Use llvm::stable_sort to sort paths

uenoku:
since llvm::sort is non-determnisic it's not suitable for using it for user-facing API. Also fix inconsistent bit position emission.


maerhart (approved):
LGTM

PR #8718, opened 2025-07-17
[Python] Speed up type_to_pytype and attribute_to_var

maerhart:
Exceptions are slow. I've seen speedups of ~60x for this new type conversion approach.

### DIFF
--- a/lib/Bindings/Python/support.py
+++ b/lib/Bindings/Python/support.py
@@ -89,102 +89,54 @@ def type_to_pytype(t) -> ir.Type:
     return t
 
   from .dialects import esi, hw, seq, rtg, rtgtest
-  try:
+  if ir.IntegerType.isinstance(t):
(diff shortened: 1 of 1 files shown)


uenoku (approved):
LGTM, this way is encouraged by MLIR doc as wel  :100: https://github.com/llvm/llvm-project/blob/e73d1a5341fafb1eadb77b787eb6e65630b4db3a/mlir/docs/Bindings/Python.md?plain=1#L502

teqdruid (approved):
This must be new! Or at least I didn't know about it back when I wrote this. Thanks!

mikeurbach:
Nice, looks good to me.

> This must be new! Or at least I didn't know about it back when I wrote this. Thanks!

Relatively new yes. It certainly took quite a while for something like this to exist, and try/except was the best we could have done for far too long.

PR #8719, opened 2025-07-17
[PyRTG] Support Python config parameters

maerhart:
Allows parameterization of information that is static in the IR such as the xlen.
The plan is to add a mechanism to codegen each test for each config providing such an xlen parameter, e.g., the following should produce two tests, one for each config (possibly by introducing a common parent class and using that as the test decorator argument). But that will be added in a future PR.

```py
@config
class RV64(Config):
  xlen = PythonParam(64)

@config
class RV32(Config):
  xlen = PythonParam(32)

@test(RV64)
def test(config):
  # do something with 'config.xlen'
```


fabianschuiki (approved):
LGTM!

PR #8720, opened 2025-07-17
[circt-verilog] Enable SROA again

maerhart:
Enable SROA again since the stackoverflow bug upstream is resolved now.


fabianschuiki (approved):
Awesome! Thanks for fixing this upstream :partying_face: :100:

PR #8721, opened 2025-07-17
[circt-lec] Adding support for the datapath dialect

cowardsa:
Simple PR to incorporate the datapath dialect into circt-lec. Have added some associated tests that prove the correctness of the datapath operators according to the SMT semantics provided by DatapathToSMT.


uenoku:
Very cool!

fabianschuiki (approved):
Very nice!

PR #8724, opened 2025-07-17
[FIRRTL] Enums: Add user-defined constructor encodings

youngar:
This adds the ability to specify the constants used to encode enumeration variants in the tag. For example,
``` firrtl
wire a : {| A = 10, B = 5 |}
```

The values can be unspecified, in which case it defaults to one higher than the previous value, or zero if it is the first value. For example, these two enumerations are the same:
``` firrtl
{| A = 0, B = 1 |} == {| A, B |}
```

In FIRRTL, Two enumeration types are considered equal if they have the same variants with the same values and data, without regard to variant order.  Internally in FIRRTL dialect IR, enumerations types must have the variants sorted from low to high. For example the following two enumeration types are equivalent:
``` firrtl
{| B = 1, A = 0 |} == {| A, B |}
```

The mlir syntax has been updated to ellide the the tag values and data types if they are implicit.  For example, the first type would be roundtripped to the second type:
``` mlir
!firrtl.enum<a = 0 : uint<0>> => !firrtl.enum<a>
```

Enumerations are no longer lowered to SystemVerilog enumerations, but to regular integer types.  This is for a variety of reasons, but mainly because enumerations in SV are nominal and it was not a good fit for our structural enumerations.

In LowerToHW, the `firrtl.istag` operation lowers to a comparison to a constant value created with  a localparam, which can show up in the output depending on how the module is optimized. The  FIRRTL code below,

``` firrtl
FIRRTL version 4.0.0
circuit Enums:
  public module Enums:
    input in : {|A = 8 : UInt<8>, B = 16 |}
    output out : UInt<8>
    match in:
      A(data):
        connect out, data
      B:
        connect out, UInt<8>(16)
```
produces:
``` verilog
module Enums(
  input  struct packed {logic [4:0] tag; union packed {logic [7:0] A;/*B: Zero Width;*/ } body; } in,
  output [7:0]                                                                                    out
);

  localparam [4:0] A = 8;
  assign out = in.tag == A ? in.body.A : 8'h10;
endmodule
```

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLTypes.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLTypes.cpp
@@ -308,8 +308,6 @@ static OptionalParseResult customTypeParser(AsmParser &parser, StringRef name,
     SmallVector<StringAttr> names;
     SmallVector<APInt> values;
     SmallVector<FIRRTLBaseType> types;
-
-    std::optional<APInt> previous;
     auto parseEnumElement = [&]() -> ParseResult {
       // Parse the variant tag.
       std::string nameStr;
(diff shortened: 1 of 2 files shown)


youngar:
`TagExtractOp`'s type inference is broken by this change, but I will fix it in a later commit.

youngar:
Pushed in some additional enum type verification that I was going to add in a future PR, for completeness.

PR #8728, opened 2025-07-17
[FIRRTL] FIRParser: parse tagExtract operations

youngar:
Depends on #8724


lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -3956,6 +3965,28 @@ ParseResult FIRStmtParser::parseRefRead(Value &result) {
...
+      parseToken(FIRToken::r_paren, "expected ')' in 'tagExtract'"))
+    return failure();
+
+  locationProcessor.setLoc(startTok.getLoc());
+
+  // Check argument is an enum-type value.
+  if (!type_isa<FEnumType>(input.getType()))
+    return emitError(startTok.getLoc(),
+                     "expected enum-type expression in 'tagExtract', got ")
+           << input.getType();
```
youngar: Its not clear to me if we should be checking this here or relying on the op verifiers. I think I will remove this check for now.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -2122,6 +2123,14 @@ ParseResult FIRStmtParser::parseExpImpl(Value &result, const Twine &message,
     if (parseRefRead(result))
       return failure();
     break;
+  case FIRToken::lp_tagExtract:
+    if (requireFeature(missingSpecFIRVersion, "tagExtract"))
+      return failure();
+    if (isLeadingStmt)
+      return emitError("unexpected tagExtract() as start of statement");
```
youngar: I don't think I need this but its a little unclear.  Without it you get a flow checking error.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -2122,6 +2123,14 @@ ParseResult FIRStmtParser::parseExpImpl(Value &result, const Twine &message,
     if (parseRefRead(result))
       return failure();
     break;
+  case FIRToken::lp_tagExtract:
+    if (requireFeature(missingSpecFIRVersion, "tagExtract"))
+      return failure();
```
youngar: It would be nice to parse this as a prim op and save some typing, but then you can't check for the spec version.

youngar:
@seldridge I added a minimum version to all of the primops. I had to use the `FIRVersion` constructor since macros don't work with the initializer list syntax `{0, 0, 0}`.  I gave the integer prim ops the same version number as the rest of the integer operations.

Issue #7106, opened 2024-05-30
[FIRRTL] firrtl-lower-signatures fails for enum types


programmerjake:
using firtool-1.75.0 from github releases

I expect firtool to be able to lower enum types [as specified in the ABI](https://github.com/chipsalliance/firrtl-spec/blob/e53da0ca1b55d002e3c2c640b8a170b95b291970/abi.md?plain=1#L329-L334), however it just plain fails.

using the following FIRRTL:
```firrtl
FIRRTL version 3.2.0
circuit check_enum_literals:
    type Ty0 = {|None, Some: UInt<8>|}
    module check_enum_literals:
        input i: UInt<8>
        output o: Ty0
        connect o, {|None, Some: UInt<8>|}(Some, i)
```
I tried running:
```bash
firtool check_enum_literals.fir -o - --mlir-print-ir-after-all
```

which output:
```
<snip>
// -----// IR Dump After LowerIntrinsics (firrtl-lower-intrinsics) //----- //
firrtl.module @check_enum_literals(in %i: !firrtl.uint<8>, out %o: !firrtl.alias<Ty0, enum<None: uint<0>, Some: uint<8>>>) attributes {convention = #firrtl<convention scalarized>} {
  %0 = firrtl.enumcreate Some(%i) : (!firrtl.uint<8>) -> !firrtl.enum<None: uint<0>, Some: uint<8>>
  firrtl.connect %o, %0 : !firrtl.alias<Ty0, enum<None: uint<0>, Some: uint<8>>>, !firrtl.enum<None: uint<0>, Some: uint<8>>
}

// -----// IR Dump After LowerSignatures Failed (firrtl-lower-signatures) //----- //
firrtl.circuit "check_enum_literals" {
  firrtl.module @check_enum_literals(in %i: !firrtl.uint<8>, out %o: !firrtl.alias<Ty0, enum<None: uint<0>, Some: uint<8>>>) attributes {convention = #firrtl<convention scalarized>} {
    %0 = firrtl.enumcreate Some(%i) : (!firrtl.uint<8>) -> !firrtl.enum<None: uint<0>, Some: uint<8>>
    firrtl.connect %o, %0 : !firrtl.alias<Ty0, enum<None: uint<0>, Some: uint<8>>>, !firrtl.enum<None: uint<0>, Some: uint<8>>
  }
}
```

uenoku:
[...]

PR #8731, opened 2025-07-17
[FIRRTL] LowerSigs: Add enum support

youngar:
This adds support for enumerations to LowerSigs so that the pass no longer fails when it encounters an enum, and just passes them through unchanged.


dtzSiFive:
(Fixes https://github.com/llvm/circt/issues/7106 ? )

PR #8732, opened 2025-07-17
[ExportVerilog] localparam should always print bitwidths

youngar:
We are currently printing 1-bit localparams without a size, which causes verilator to complain that it does not have a known bitwidth when used in some expressions.  This changes ExportVerilog to always print the bitwidth of localparams, which is something we already do for parameters.

### DIFF
--- a/lib/Conversion/ExportVerilog/ExportVerilog.cpp
+++ b/lib/Conversion/ExportVerilog/ExportVerilog.cpp
@@ -5907,6 +5907,11 @@ LogicalResult StmtEmitter::emitDeclaration(Operation *op) {
   auto type = value.getType();
   auto word = getVerilogDeclWord(op, emitter);
   auto isZeroBit = isZeroBitType(type);
+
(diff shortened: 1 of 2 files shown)


lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -5907,6 +5907,13 @@ LogicalResult StmtEmitter::emitDeclaration(Operation *op) {
   auto type = value.getType();
   auto word = getVerilogDeclWord(op, emitter);
   auto isZeroBit = isZeroBitType(type);
+
+  // LocalParams always need the bitwidth, otherwise they are considered to have
+  // an unknown size.
+  bool singleBitDefaultType = true;
+  if (isa<LocalParamOp>(op))
+    singleBitDefaultType = false;
+
```
uenoku:
```suggestion
  bool singleBitDefaultType = !isa<LocalParamOp>(op);
```

PR #8734, opened 2025-07-18
[RTG] Add a pass to print a list of tests

maerhart:
This pass prints the names of the tests in the IR. Since tests can be duplicated and elaborated differently during compilation, it prints both the newly uniqued name and the original name as specified by the user in the frontend.

In the future this could be extended to print additional information about tests post-elaboration.

### DIFF
--- a/lib/Dialect/RTG/Transforms/PrintTestNamesPass.cpp
+++ b/lib/Dialect/RTG/Transforms/PrintTestNamesPass.cpp
@@ -1,0 +1,49 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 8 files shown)


lib/Dialect/RTG/Transforms/PrintTestNamesPass.cpp
```diff
@@ -0,0 +1,52 @@
...
+  if (!output) {
+    mlir::emitError(UnknownLoc::get(&getContext()), errorMessage);
+    return signalPassFailure();
+  }
+
+  llvm::raw_ostream &os = output->os();
+
+  for (auto testOp : getOperation().getOps<rtg::TestOp>())
+    os << testOp.getSymName() << "," << testOp.getTemplateName() << "\n";
+
```
uenoku, line 46: Probably markAllAnalysesPreserved can be called?

fabianschuiki (approved):
LGTM

PR #8735, opened 2025-07-18
[RTG] Add immediate concat and slice operations

maerhart:
This allows passing immediates around instead of two integers (one with the value, one with the bitwidth) and then split it, e.g., for loading that immediate into a register since there is no instruction that can load a 64bit or even 32bit immediate in one go.

### DIFF
--- a/frontends/PyRTG/src/pyrtg/immediates.py
+++ b/frontends/PyRTG/src/pyrtg/immediates.py
@@ -29,6 +29,39 @@ class Immediate(Value):
     # Note that the upper limit is exclusive
     return Immediate(width, Integer.random(0, 2**width - 1))
 
+  def concat(self, *others: Immediate) -> Immediate:
+    """
+    Concatenates this immediate with the provided immediates. The operands are
+    concatenated in order, with this immediate becoming the most significant
+    bits of the result.
+    """
+
+    return rtg.ConcatImmediateOp([self, *others])
+
+  def __getitem__(self, slice_range) -> Immediate:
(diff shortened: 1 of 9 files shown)


frontends/PyRTG/src/pyrtg/immediates.py
```diff
@@ -29,6 +29,39 @@ def random(width: int) -> Immediate:
...
+            f"Invalid slice range [{start}:{stop}] for width {self._width}")
+      return rtg.SliceImmediateOp(ImmediateType(stop - start), self, start)
+
+    if isinstance(slice_range, int):
+      if slice_range < 0 or slice_range >= self._width:
+        raise ValueError(
+            f"Index {slice_range} out of range for width {self._width}")
+      return rtg.SliceImmediateOp(ImmediateType(1), self, slice_range)
+
+    raise TypeError("Slice must be an integer or slice object")
```
fabianschuiki: Does the MSB have index 0 here? Looks like the start offset is forwarded as offset to `SliceImmediateOp`; does that op start counting from 0 at the MSB, or behave more like `comb.extract` where it starts counting from 0 at the LSB?

fabianschuiki: Just noticed the op def further down. That looks like index 0 would be the LSB, wouldn't it?

maerhart: Yeah that should say LSB. Thanks!

PR #8736, opened 2025-07-18
[Datapath] Create Datapath to Comb Pass

cowardsa:
A pass lowering datapath operators to a subset of the comb dialect, typically gate-level comb operators. The pass is for usage in the circt-synth pipeline, where we create datapath operators, perform datapath optimisations and then lower back to comb for further processing in the AIG generation pipeline. 

The compress operator can be lowered to a Wallace tree and also to a variadic add. The lower-compress-to-add option is used for comparing compressor tree synthesis against other tools (e.g. Yosys). The partial product operator can be lowered to alternative multiplier array implementations depending on the bitwidth. Currently the pass supports lowering to a simple AND array and a Radix-4 Booth multiplier.

A subsequent PR will incorporate the datapath dialect into circt-synth tool flow.

### DIFF
--- a/lib/Conversion/DatapathToComb/DatapathToComb.cpp
+++ b/lib/Conversion/DatapathToComb/DatapathToComb.cpp
@@ -92,7 +92,7 @@ struct DatapathPartialProductOpConversion
 
   DatapathPartialProductOpConversion(MLIRContext *context, bool forceBooth)
       : OpConversionPattern<PartialProductOp>(context),
-        forceBooth(forceBooth) {};
+        forceBooth(forceBooth){};
 
   const bool forceBooth;
 


cowardsa:
Initially created this in review because think we need to discuss where some of the shared infrastructure should live.

@uenoku - you'll notice that I've currently duplicated your wallace tree and full adder code for example - since this is common infrastructure - where do you think this should best reside?

uenoku:
Super cool! 

>  since this is common infrastructure - where do you think this should best reside?

Could you add function declarations to https://github.com/llvm/circt/blob/main/include/circt/Dialect/Comb/CombOps.h and define in CombOps.cpp?

cowardsa:
@uenoku - have now moved these over to comb ops as suggested - think this is now ready for a full review

test/Conversion/DatapathToComb/datapath-to-comb.mlir
```diff
@@ -0,0 +1,47 @@
+// RUN: circt-opt %s --convert-datapath-to-comb | FileCheck %s
```
uenoku, line 1: Please test lower-compress-to-add as well.

include/circt/Dialect/Comb/CombOps.h
```diff
@@ -85,6 +86,17 @@ Value createDynamicInject(OpBuilder &builder, Location loc, Value value,
 Value createInject(OpBuilder &builder, Location loc, Value value,
                    unsigned offset, Value replacement);
 
+/// Construct a full adder for three 1-bit inputs.
+std::pair<Value, Value> fullAdder(ConversionPatternRewriter &rewriter,
+                                  Location loc, Value a, Value b, Value c);
+
+/// Perform Wallace tree reduction on partial products.
+/// See https://en.wikipedia.org/wiki/Wallace_tree
+SmallVector<Value>
```
uenoku: Could you add documents for `targetAddends` and `partialProducts`?

include/circt/Dialect/Comb/CombOps.h
```diff
@@ -85,6 +86,17 @@ Value createDynamicInject(OpBuilder &builder, Location loc, Value value,
...
+wallaceReduction(Value falseValue, size_t width, size_t targetAddends,
+                 ConversionPatternRewriter &rewriter, Location loc,
+                 SmallVector<SmallVector<Value>> &partialProducts);
```
uenoku: nit: Could you refactor the signature a bit since it's exposed as a library? Specifically could you remove `falseValue` and use OpBuilder instead of ConversionRewriter, reorder to look like more a builder function. 
```suggestion
wallaceReduction(OpBuilder &builder, Location loc, size_t width, 
                               size_t targetAddends,
                               SmallVector<SmallVector<Value>>& partialProducts);
```

include/circt/Dialect/Comb/CombOps.h
```diff
@@ -85,6 +86,17 @@ Value createDynamicInject(OpBuilder &builder, Location loc, Value value,
 Value createInject(OpBuilder &builder, Location loc, Value value,
                    unsigned offset, Value replacement);
 
+/// Construct a full adder for three 1-bit inputs.
+std::pair<Value, Value> fullAdder(ConversionPatternRewriter &rewriter,
```
uenoku: ConversionPatternRewriter is sub-class of OpBuilder so OpBuilder would be appropriate (I defined this function in this so sorry).
```suggestion
std::pair<Value, Value> fullAdder(OpBuilder &builder,
```

lib/Conversion/DatapathToComb/DatapathToComb.cpp
```diff
@@ -0,0 +1,272 @@
...
+    // Booth encoding halves array height by grouping three bits at a time:
+    // partialProducts[i] = a * (-2*b[2*i+1] + b[2*i] + b[2*i-1]) << 2*i
+    // encNeg \approx (-2*b[2*i+1] + b[2*i] + b[2*i-1]) <= 0
+    // encOne = (-2*b[2*i+1] + b[2*i] + b[2*i-1]) == +/- 1
+    // encTwo = (-2*b[2*i+1] + b[2*i] + b[2*i-1]) == +/- 2
+    Value encNegPrev;
+    for (unsigned i = 0; i < width; i += 2) {
+      // Get Booth bits: b[i+1], b[i], b[i-1] (b[-1] = 0)
+      Value bim1 = (i == 0) ? zeroFalse : bBits[i - 1];
+      Value bi = (i < width) ? bBits[i] : zeroFalse;
```
uenoku, line 173: nit:
```suggestion
      Value bi = bBits[i];
```

lib/Conversion/DatapathToComb/DatapathToComb.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+struct DatapathPartialProductOpConversion
+    : OpConversionPattern<PartialProductOp> {
+  using OpConversionPattern<PartialProductOp>::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(PartialProductOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    auto inputs = op.getOperands();
+    Value a = inputs[0];
```
uenoku: nit:
```suggestion
    Value a = op.getLhs();
    Value b = op.getRhs();
```

lib/Dialect/Comb/CombOps.cpp
```diff
@@ -216,6 +216,86 @@ Value comb::createInject(OpBuilder &builder, Location loc, Value value,
...
+        // Add remaining rows as is
+        newAddends.append(addends.begin() + i, addends.end());
+      }
+    }
+    std::swap(newAddends, addends);
+  }
+
+  assert(addends.size() <= targetAddends);
+  SmallVector<Value> carrySave;
+  for (auto addend : addends) {
```
uenoku: nit: 
```suggestion
  for (auto& addend : addends) {
```

test/Conversion/DatapathToComb/datapath-to-comb.mlir
```diff
@@ -0,0 +1,47 @@
...
+  // CHECK-NEXT: comb.shl %[[PP0]], %c0_i3 : i3
+  // CHECK-NEXT: %[[B1R:.+]] = comb.replicate %[[B1]] : (i1) -> i3
+  // CHECK-NEXT: %[[PP1:.+]]  = comb.and %[[B1R]], %a : i3
+  // CHECK-NEXT: %c1_i3 = hw.constant 1 : i3
+  // CHECK-NEXT: comb.shl %[[PP1]], %c1_i3 : i3
+  // CHECK-NEXT: %[[B2R:.+]] = comb.replicate %[[B2]] : (i1) -> i3
+  // CHECK-NEXT: %[[PP2:.+]] = comb.and %[[B2R]], %a : i3
+  // CHECK-NEXT: %c2_i3 = hw.constant 2 : i3
+  // CHECK-NEXT: comb.shl %[[PP2]], %c2_i3 : i3
+  %0:3 = datapath.partial_product %a, %b : (i3, i3) -> (i3, i3, i3)
```
uenoku, line 57: Could you add a test for booth as well (at least it compiles)?

include/circt/Dialect/Comb/CombOps.h
cowardsa: Thanks for this pointer - wasn't sure on the appropriate signature for these functions!

cowardsa:
@uenoku - have moved everything over to CombOps to provide shared infrastructure - added tests to circt-synth integration tests - since this is largely about lowering? Although open to suggestions to move? - have added an option to forceBooth array lowering so that we can check equivalence in a resonable runtime. Let me know if you've got any other concerns or queries!

include/circt/Conversion/DatapathToComb.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef CIRCT_CONVERSION_DATAPATHTOCOMB_H
+#define CIRCT_CONVERSION_DATAPATHTOCOMB_H
+
+#include "circt/Support/LLVM.h"
+
+namespace circt {
+
+void populateDatapathToCombConversionPatterns(TypeConverter &converter,
+                                              RewritePatternSet &patterns,
+                                              bool lowerCompressToAdd,
```
uenoku: Do you have use case or plan for this function? If not keeping it in DatapatToComb might be better.

uenoku (approved):
Thank you for addressing comments, one question for pattern population helper but LGTM!i

> - added tests to circt-synth integration tests - since this is largely about lowering

I think that makes sense. Eventually we might want to clean up that directory as currently there are tons of translation validation are added altogether but we can do that later.

cowardsa:
Will merge tomorrow unless anyone has any final comments :)

PR #8737, opened 2025-07-18
[CombToSMT] Force conversion from bool to bv<1> after icmp

maerhart:


TaoBi22 (approved):
LGTM for now. Thanks for fixing this up so quickly!

PR #8739, opened 2025-07-18
Inline input-only modules

rwy7:
Add a pass which inlines input-only modules. This functionality was previously performed by extract-test-code, but when extract-test-code is disabled, we still need to perform this transformation. This transformation is notable because it ignores don't touch annotations.

The code in this pass is largely a copy-paste of parts of extract-test-code.


uenoku:
IMO this pass may not be required if `fixupEmptyModule` is added in verilog codegen. The transformation was added just for preventing empty module issues in the synthesis.

PR #8743, opened 2025-07-19
[RTG] Fix ValidateOp elaboration

maerhart:
Also treat them as identity values to be properly threaded through block arguments. Currently the pass errors out when passing a result of a validate op through a block argument.
This also means that two instances referring to the same value are not considered equal. This is problematic when adding it to a set or bag. However, this applies to the 'value' result as well, so we need to revisit this again anyway.


uenoku (approved):
LGTM

PR #8747, opened 2025-07-20
[Support] Add NPN class for Boolean function canonicalization

uenoku:
This commit introduces NPNClass and BinaryTruthTable classes to the CIRCT Support library for Boolean function equivalence checking and canonicalization.

NPNClass computes Negation-Permutation-Negation canonical forms for Boolean functions, enabling efficient detection of functionally equivalent circuits under input/output transformations. BinaryTruthTable provides compact representation using APInt with support for multi-input, multi-output functions.

This infrastructure is essential for technology mapping and Boolean function optimization in synthesis flows.

### DIFF
--- a/include/circt/Support/NPNClass.h
+++ b/include/circt/Support/NPNClass.h
@@ -1,0 +1,162 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This header file defines NPN (Negation-Permutation-Negation) canonical forms
+// and binary truth tables for boolean function representation and equivalence
+// checking in combinational logic optimization.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_SUPPORT_NPNCLASS_H
+#define CIRCT_SUPPORT_NPNCLASS_H
+
+#include "circt/Support/LLVM.h"
+#include "llvm/ADT/APInt.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace circt {
+
+/// Represents a boolean function as a truth table.
+///
+/// A truth table stores the output values for all possible input combinations
+/// of a boolean function. For a function with n inputs and m outputs, the
+/// truth table contains 2^n entries, each with m output bits.
+///
+/// Example: For a 2-input AND gate:
+/// - Input 00 -> Output 0
+/// - Input 01 -> Output 0
+/// - Input 10 -> Output 0
+/// - Input 11 -> Output 1
+/// This would be stored as the bit pattern 0001 in the truth table.
+struct BinaryTruthTable {
+  unsigned numInputs;  ///< Number of inputs for this boolean function
+  unsigned numOutputs; ///< Number of outputs for this boolean function
+  llvm::APInt table;   ///< Truth table data as a packed bit vector
+
+  /// Default constructor creates an empty truth table.
+  BinaryTruthTable() : numInputs(0), numOutputs(0), table(1, 0) {}
+
+  /// Constructor for a truth table with given dimensions and evaluation data.
+  BinaryTruthTable(unsigned numInputs, unsigned numOutputs,
+                   const llvm::APInt &table)
+      : numInputs(numInputs), numOutputs(numOutputs), table(table) {
+    assert(table.getBitWidth() == (1u << numInputs) * numOutputs &&
+           "Truth table size mismatch");
+  }
+
+  /// Constructor for a truth table with given dimensions, initialized to zero.
+  BinaryTruthTable(unsigned numInputs, unsigned numOutputs)
+      : numInputs(numInputs), numOutputs(numOutputs),
+        table((1u << numInputs) * numOutputs, 0) {}
+
+  /// Get the output value for a given input combination.
+  llvm::APInt getOutput(const llvm::APInt &input) const;
+
+  /// Set the output value for a given input combination.
+  void setOutput(const llvm::APInt &input, const llvm::APInt &output);
+
+  /// Apply input permutation to create a new truth table.
+  /// This reorders the input variables according to the given permutation.
+  BinaryTruthTable applyPermutation(ArrayRef<unsigned> permutation) const;
+
+  /// Apply input negation to create a new truth table.
+  /// This negates selected input variables based on the mask.
+  BinaryTruthTable applyInputNegation(unsigned mask) const;
+
+  /// Apply output negation to create a new truth table.
+  /// This negates selected output variables based on the mask.
+  BinaryTruthTable applyOutputNegation(unsigned negation) const;
+
+  /// Check if this truth table is lexicographically smaller than another.
+  /// Used for canonical ordering of truth tables.
+  bool isLexicographicallySmaller(const BinaryTruthTable &other) const;
+
+  /// Equality comparison for truth tables.
+  bool operator==(const BinaryTruthTable &other) const;
+
+  /// Debug dump method for truth tables.
+  void dump(llvm::raw_ostream &os = llvm::errs()) const;
+};
+
+/// Represents the canonical form of a boolean function under NPN equivalence.
+///
+/// NPN (Negation-Permutation-Negation) equivalence considers two boolean
+/// functions equivalent if one can be obtained from the other by:
+/// 1. Negating some inputs (pre-negation)
+/// 2. Permuting the inputs
+/// 3. Negating some outputs (post-negation)
+///
+/// This canonical form is used to efficiently match cuts against library
+/// patterns, as functions in the same NPN class can be implemented by the
+/// same circuit with appropriate input/output inversions.
+struct NPNClass {
+  BinaryTruthTable truthTable;                  ///< Canonical truth table
+  llvm::SmallVector<unsigned> inputPermutation; ///< Input permutation applied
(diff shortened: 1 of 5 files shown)


include/circt/Support/NPNClass.h
```diff
@@ -0,0 +1,162 @@
...
+///
+/// This canonical form is used to efficiently match cuts against library
+/// patterns, as functions in the same NPN class can be implemented by the
+/// same circuit with appropriate input/output inversions.
+struct NPNClass {
+  BinaryTruthTable truthTable;                  ///< Canonical truth table
+  llvm::SmallVector<unsigned> inputPermutation; ///< Input permutation applied
+  unsigned inputNegation = 0;                   ///< Input negation mask
+  unsigned outputNegation = 0;                  ///< Output negation mask
+
```
cowardsa, line 108: Nit: an example might be helpful here - ab+c is NPN equivalent to !a + b(!c)

include/circt/Support/NPNClass.h
```diff
@@ -0,0 +1,162 @@
...
+  ///
+  /// FIXME: Currently we are using exact canonicalization which doesn't scale
+  /// well. For larger truth tables, semi-canonical forms should be used
+  /// instead.
+  static NPNClass computeNPNCanonicalForm(const BinaryTruthTable &tt);
+
+  /// Get input mapping from this NPN class to another equivalent NPN class.
+  ///
+  /// When two NPN classes are equivalent, they may have different input
+  /// permutations. This function computes a mapping that allows transforming
```
cowardsa: Q: Is the mapping just another permutation or something different? If it's a permutation perhaps it could just be described as a permutation?

lib/Support/NPNClass.cpp
```diff
@@ -0,0 +1,277 @@
...
+  assert(permutation.size() == numInputs && "Permutation size mismatch");
+  BinaryTruthTable result(numInputs, numOutputs);
+
+  for (unsigned i = 0; i < (1u << numInputs); ++i) {
+    llvm::APInt input(numInputs, i);
+    llvm::APInt permutedInput(numInputs, 0);
+
+    // Apply permutation
+    for (unsigned j = 0; j < numInputs; ++j)
+      permutedInput.setBitVal(j, input[permutation[j]]);
```
cowardsa: Would it improve efficiency if only permuted bits were updated?

lib/Support/NPNClass.cpp
```diff
@@ -0,0 +1,277 @@
...
+BinaryTruthTable BinaryTruthTable::applyInputNegation(unsigned mask) const {
+  BinaryTruthTable result(numInputs, numOutputs);
+
+  for (unsigned i = 0; i < (1u << numInputs); ++i) {
+    llvm::APInt input(numInputs, i);
+    llvm::APInt negatedInput(numInputs, 0);
+
+    // Apply negation
+    for (unsigned j = 0; j < numInputs; ++j)
+      negatedInput.setBitVal(j, (mask & (1u << j)) ? !input[j] : input[j]);
```
cowardsa: Nit: could this be bitwise xor, which I think APInt supports? mask ^ input

lib/Support/NPNClass.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  for (unsigned i = 0; i < (1u << numInputs); ++i) {
+    llvm::APInt input(numInputs, i);
+    llvm::APInt output = getOutput(input);
+    llvm::APInt negatedOutput(numOutputs, 0);
+
+    // Apply negation
+    for (unsigned j = 0; j < numOutputs; ++j)
+      negatedOutput.setBitVal(j,
+                              (negation & (1u << j)) ? !output[j] : output[j]);
```
cowardsa: `negatedOutput = negation ^ output`

lib/Support/NPNClass.cpp
```diff
@@ -0,0 +1,277 @@
...
+//===----------------------------------------------------------------------===//
+
+namespace {
+// Helper functions for permutation manipulation - kept as implementation
+// details
+
+/// Create an identity permutation of the given size.
+/// Result[i] = i for all i in [0, size).
+llvm::SmallVector<unsigned> identityPermutation(unsigned size) {
+  llvm::SmallVector<unsigned> identity(size);
```
cowardsa, line 155: Can this be simpler?
`SmallVector<unsigned> identity(size, 1);`

lib/Support/NPNClass.cpp
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  return mapping;
+}
+
+NPNClass NPNClass::computeNPNCanonicalForm(const BinaryTruthTable &tt) {
+  NPNClass canonical(tt);
+  // Initialize permutation with identity
+  canonical.inputPermutation = identityPermutation(tt.numInputs);
+  assert(tt.numInputs <= 8 && "Too many inputs for input negation mask");
```
cowardsa: Nit: should this assert message say too many inputs for canonicalization?

lib/Support/NPNClass.cpp
```diff
@@ -0,0 +1,277 @@
...
+  os << "NPNClass:\n";
+  os << "  Input permutation: [";
+  for (unsigned i = 0; i < inputPermutation.size(); ++i) {
+    if (i > 0)
+      os << ", ";
+    os << inputPermutation[i];
+  }
+  os << "]\n";
+  os << "  Input negation: 0b";
+  for (int i = inputPermutation.size() - 1; i >= 0; --i) {
```
cowardsa: Should this be truthTable.numInputs? Although these are presumably equal

unittests/Support/NPNClassTest.cpp
```diff
@@ -0,0 +1,414 @@
+//===- NPNClassTest.cpp - NPNClass unit tests ---------------------------===//
```
cowardsa: Nit: File header text not required?

unittests/Support/NPNClassTest.cpp
```diff
@@ -0,0 +1,414 @@
...
+  APInt output0(1, 0);
+  APInt output1(1, 1);
+
+  // Set AND truth table
+  original.setOutput(APInt(2, 0), output0); // f(0,0) = 0
+  original.setOutput(APInt(2, 1), output0); // f(0,1) = 0
+  original.setOutput(APInt(2, 2), output0); // f(1,0) = 0
+  original.setOutput(APInt(2, 3), output1); // f(1,1) = 1
+
+  // Apply input negation mask 1 (negate the the second)
```
cowardsa: Nit: (negate the the second)

unittests/Support/NPNClassTest.cpp
```diff
@@ -0,0 +1,414 @@
...
+  // clang-format off
+  // g(0,0) gets the output from f(0,1) = 0
+  // g(1,0) gets the output from f(1,1) = 1
+  // g(0,1) gets the output from f(0,0) = 1
+  // g(1,1) gets the output from f(1,0) = 0
+  // clang-format on
+  EXPECT_EQ(result00, output0); // g(0,0) = 0
+  EXPECT_EQ(result01, output0); // g(0,1) = 0
+  EXPECT_EQ(result10, output1); // g(1,0) = 1
+  EXPECT_EQ(result11, output0); // g(1,1) = 0
```
cowardsa: Nit: These comments don't align with the results shown below?

unittests/Support/NPNClassTest.cpp
```diff
@@ -0,0 +1,414 @@
...
+  EXPECT_EQ(negated2.getOutput(APInt(2, 2)), APInt(2, 0)); // 10 -> 00
+  EXPECT_EQ(negated2.getOutput(APInt(2, 3)), APInt(2, 1)); // 11 -> 10
+
+  // Apply output negation mask 3 (negate both output bits)
+  BinaryTruthTable negated3 = original.applyOutputNegation(3);
+
+  EXPECT_EQ(negated3.getOutput(APInt(2, 0)), APInt(2, 3)); // 00 -> 11
+  EXPECT_EQ(negated3.getOutput(APInt(2, 1)), APInt(2, 2)); // 01 -> 10
+  EXPECT_EQ(negated3.getOutput(APInt(2, 2)), APInt(2, 1)); // 10 -> 01
+  EXPECT_EQ(negated3.getOutput(APInt(2, 3)), APInt(2, 0)); // 11 -> 00
```
cowardsa: Nit: The comments here don't align with the rhs of expected_eq is that intentional? E.g. APInt(2,1) != 10

cowardsa (approved):
LGTM - few minor comments but nothing substantial - would it be worth adding some tests to make sure we can't touch certain elements e.g. trigger the asserts that setOutput(largeNumber) doesn't proceed?

include/circt/Support/NPNClass.h
uenoku: Yep, this is just a permutation.

unittests/Support/NPNClassTest.cpp
uenoku: These seems like outdated, thank you for pointing it out!

lib/Support/NPNClass.cpp
uenoku: Great point, thanks!

lib/Support/NPNClass.cpp
uenoku: Maybe I'm missing something but identity[i] = i so it's a different from identity(size, 1).

uenoku:
> trigger the asserts that setOutput(largeNumber) doesn't proceed?

Thank you for the suggestion! Assertion is conditionally enabled via compiler option so it's a bit difficult to test assertion firing in the unit tests (in CI assertions are enabled but we don't enable assertions for the binary release). For this specific conditions, users of NPN/TruthTable are expected to provide APInt with correct bit width so I think CI should have good coverage for this specific asssertions.

cowardsa:
LGTM - thanks for making the changes - the updated comments also help where you've written !0, !1 - makes it a little easier to follow!

PR #8749, opened 2025-07-20
[VerifToSMT] Lower `verif.refines` to SMT

fzi-hielscher:
Implement a conversion pattern for the `verif.refines` operation to the SMT dialect.

Following the current definition of the `RefinmentCheckingOp`, the second (target) circuit is a refinement of the first (source) circuit iff for any given input assignment the set of possible outputs of the target circuit is a subset of the source circuit's possible outputs. The conversion assumes all `smt.declare_fun` operations in the circuit body to be sources of non-determinism, as they are created by the lowering of the division (by zero) and (out of bounds) array get operations. The expression we are checking for is:
_For all_ inputs: _For all_ non-deterministic choices in the target circuit: _Exists_ a non-deterministic choice in the source circuit leading to identical outputs?

Since we are trying to find a counter example, we produce the inverted expression:
_Exists_ an input and _exists_ a non-deterministic choice in the target circuit so that _for all_ non-deterministic choices in the source circuit the outputs are _not_ equal.

As we are checking for the existence of a satisfying interpretation, the outer existential quantifiers don't need to be modeled explicitly. Thus, the free variables in the target circuit stay free variables in the miter. If there are _no_ free variables in the source circuit the refinement check becomes an equivalence check. If there _are_ free variables in  the source circuit, they are bound by the universal quantifier.

This should generally match the way Alive2 handles non-determinism.

### DIFF
--- a/lib/Conversion/VerifToSMT/VerifToSMT.cpp
+++ b/lib/Conversion/VerifToSMT/VerifToSMT.cpp
@@ -231,25 +231,26 @@ struct RefinementCheckingOpConversion
   matchAndRewrite(verif::RefinementCheckingOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
 
-    // Find non-deterministic values (free variables) in the LHS circuit.
-    SmallVector<Value> lhsNonDetValues;
-    bool onlyPrimitiveNd = true;
+    // Find non-deterministic values (free variables) in the source circuit.
+    // For now, only support quantification over 'primitive' types.
+    SmallVector<Value> srcNonDetValues;
+    bool canBind = true;
     for (auto ndOp : op.getFirstCircuit().getOps<smt::DeclareFunOp>()) {
       if (!isa<smt::IntType, smt::BoolType, smt::BitVectorType>(
               ndOp.getType())) {
         ndOp.emitError("Uninterpreted function of non-primitive type cannot be "
                        "converted.");
-        onlyPrimitiveNd = false;
+        canBind = false;
       }
-      lhsNonDetValues.push_back(ndOp.getResult());
+      srcNonDetValues.push_back(ndOp.getResult());
     }
-    if (!onlyPrimitiveNd)
+    if (!canBind)
       return failure();
 
-    if (lhsNonDetValues.empty()) {
-      // If there is no non-determinism in the LHS circuit, the refinement check
-      // becomes an equivalence check, which does not need quantified
-      // expressions.
+    if (srcNonDetValues.empty()) {
+      // If there is no non-determinism in the source circuit, the
+      // refinement check becomes an equivalence check, which does not
+      // need quantified expressions.
       auto eqOp = rewriter.create<verif::LogicEquivalenceCheckingOp>(
           op.getLoc(), op.getNumResults() != 0);
       rewriter.moveBlockBefore(&op.getFirstCircuit().front(),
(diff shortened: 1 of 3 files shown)


lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -217,6 +217,127 @@ struct LogicEquivalenceCheckingOpConversion
...
+    auto *firstOutputs = adaptor.getFirstCircuit().front().getTerminator();
+    auto *secondOutputs = adaptor.getSecondCircuit().front().getTerminator();
+
+    if (firstOutputs->getNumOperands() == 0) {
+      // Trivially equivalent
+      Value trueVal =
+          rewriter.create<arith::ConstantOp>(loc, rewriter.getBoolAttr(true));
+      rewriter.replaceOp(op, trueVal);
+      return success();
+    }
```
maerhart: There doesn't seem to be a test for this case?
Isn't the result optional and we have to check first before attempting to replace it with constant 'true'?

lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -217,6 +217,127 @@ struct LogicEquivalenceCheckingOpConversion
...
+
+  LogicalResult
+  matchAndRewrite(verif::RefinementCheckingOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    // Find non-deterministic values (free variables) in the source circuit.
+    // For now, only support quantification over 'primitive' types.
+    SmallVector<Value> srcNonDetValues;
+    bool canBind = true;
+    for (auto ndOp : op.getFirstCircuit().getOps<smt::DeclareFunOp>()) {
```
maerhart, line 238: Doesn't this assume that the body regions were already converted? Usually, we don't have SMT ops already in the body, but something like a division. How do you intend to support this?
Maybe it'd be better to not use the dialect conversion framework to perform this lowering but do it in a separate pass after the dialect conversion and mark this op legal here (after converting the arguments)?

lib/Conversion/VerifToSMT/VerifToSMT.cpp
```diff
@@ -217,6 +217,127 @@ struct LogicEquivalenceCheckingOpConversion
...
+      if (!isa<smt::IntType, smt::BoolType, smt::BitVectorType>(
+              ndOp.getType())) {
+        ndOp.emitError("Uninterpreted function of non-primitive type cannot be "
+                       "converted.");
```
maerhart, line 242: It's not really good practice to emit an error in a rewrite pattern.

lib/Conversion/VerifToSMT/VerifToSMT.cpp
fzi-hielscher: Yes, you're right. I missed that when making the result of the LEC op optional in #8701. I'll write a fix+test later today.

lib/Conversion/VerifToSMT/VerifToSMT.cpp
fzi-hielscher: At the moment all pipelines that use VerifToSMT (circt-lec and circt-bmc) run it after HWToSMT and CombToSMT. But, yeah, it is not great to assume this to be the case in general. `verif.refines` is both structurally and functionally very close to `verif.lec`, so I wanted to also keep their implementation close.

If we really want to decouple the lowerings, we'd have to somehow model the non-deterministic choices explicitly in the body of the cirucuit, e.g.:

```MLIR
verif.refines first {
  [...]
  %nondet = verif.nondeterministic : i8
  %div = comb.divu %a, %b : i8
  %res = comb.mux %isBZero, %nondet, %div : i8
 [...]
```

This would have to be added when creating the `verif.refines` op and we'd have to explicitly check for patterns that are known to create nondeterminism.  Matching on `smt.declare_fun` seems like the more universal (and simpler) solution. 

Moving the lowering to a separate pass would be the other option. But marking a `verif` op as (dynamically) legal in VerifToSMT  and then having another pass that _actually_ converts the op to SMT also seems quirky to me. It would likely also cause some duplicated logic between the `verif.lec` and `verif.refines` lowerings, unless we move the `verif.lec` lowering too.

I do not love any of the options. The way I've implemented it at the moment feels like the simplest solution that 'just works' for our current use-case, while admittedly being a little hacky. Do you have a preference here?

lib/Conversion/VerifToSMT/VerifToSMT.cpp
fzi-hielscher: True. But it's just because in theory the op could be picked up by another pattern and then succeed or fail again, isn't it? That is unlikely to happen here, so emitting a diagnostic seemed more useful to me than just failing to legalize without any hint why.

lib/Conversion/VerifToSMT/VerifToSMT.cpp
maerhart: I agree with your points and I'm totally fine with merging this as is.

However, I don't think running the HWToSMT, CombToSMT, VerifToSMT passes separately in sequence is the best approach. It's just that way because it was the easiest to get running (and I was lazy xD). It'd be better to add all those patterns in a single pass run (AFAIK MLIR upstream does that for lowering to LLVM using a dialect interface to query all lowering patterns from all dialects in the IR). That way we wouldn't end up with all the unrealised conversion cast ops inbetween and it'd be less error prone because we can define a stricter lowering target. But if we do it that way, your pattern wouldn't work anymore.

When going in the direction of having a separate lowering pass for `verif.refines`, I'd also do that for the LEC and BMC ops (either to the same pass, or each in their separate but sharing as much as possible via some utility header). That pass could then use the SMT lowering dialect interface to populate the patterns for all operations in the `verif.refines` body, define a ConversionTarget that marks all non-SMT operations within the body illegal, but the refines op and everything outside of it legal, then do a full conversion (which gives us more guarantees and a more direct error than the current partial conversions, e.g., see https://github.com/llvm/circt/pull/8737 which lead to an error from Z3 for something that could have been caught in that pass). Once the body is converted, the `verif.refines`, LEC, BMC ops would be converted and we're done. I think that'd be my preferred solution that we could slowly work towards.
I wouldn't be concerned about marking LECOp, BMCOp, RefinementOp legal, dynamically legal, or ignoring them in VerifToSMT. If something isn't the perfect fit for a lowering pattern, we don't have to shoehorn it into one for completeness sake. And I don't really think there are benefits in having them as patterns (or am I missing something?)

I wouldn't go in the direction of the `verif.nondeterministic` you outlined for essentially the reasons you already mentioned.

lib/Conversion/VerifToSMT/VerifToSMT.cpp
maerhart: Agree with your reasoning and that a proper message is useful here. It was just another hint for me that the lowering might be better off outside a pattern :)

maerhart (approved):

lib/Conversion/VerifToSMT/VerifToSMT.cpp
fzi-hielscher: Awesome. Thanks a lot for your input. I'll have a look at a way to restructure the SMT lowering pipeline to make it more like LLVM.

Issue #8610, opened 2025-06-27
[FIRRTL] [Bug Report] Comb Loop Detect not aware of bitcast + extract


sequencer:
See this MLIR Module of FIRRTL, the output `output_0` should connected from `input_0` with `%3 = firrtl.bitcast %io` and `firrtl.bits %3 7 to 0`:
```mlir
module {
  firrtl.circuit "TEST" {
    firrtl.module @TEST(in %input_0: !firrtl.uint<8>, out %output_0: !firrtl.uint<8>) attributes {convention = #firrtl<convention scalarized>, layers = [@verification]} {
      %io = firrtl.wire : !firrtl.bundle<input_0 flip: uint<8>, output_0: uint<8>>
      %probe = firrtl.wire : !firrtl.bundle<>
      %0 = firrtl.subfield %io[input_0] : !firrtl.bundle<input_0 flip: uint<8>, output_0: uint<8>>
      firrtl.connect %0, %input_0 : !firrtl.uint<8>
      %1 = firrtl.subfield %io[output_0] : !firrtl.bundle<input_0 flip: uint<8>, output_0: uint<8>>
      firrtl.connect %output_0, %1 : !firrtl.uint<8>
      %2 = firrtl.subfield %io[output_0] : !firrtl.bundle<input_0 flip: uint<8>, output_0: uint<8>>
      %3 = firrtl.bitcast %io : (!firrtl.bundle<input_0 flip: uint<8>, output_0: uint<8>>) -> !firrtl.uint<16>
      %4 = firrtl.bits %3 7 to 0 : (!firrtl.uint<16>) -> !firrtl.uint<8>
      %_GEN_0 = firrtl.node interesting_name %4 : !firrtl.uint<8>
      firrtl.connect %2, %_GEN_0 : !firrtl.uint<8>
    }
    firrtl.layer @verification bind {
    }
  }
}
```
However, `combinational cycle in a FIRRTL module, sample path: TEST.{_GEN_0 <- ... <- io_output_0 <- _GEN_0}`

The `firrtl.bitcast` is only designed for lowering to `hw.bitcast`, see [here](https://github.com/llvm/circt/blob/6d08cb473337670277f12dd5857fa6e9a7e616f6/lib/Dialect/FIRRTL/Transforms/LowerTypes.cpp#L1350), so `CheckCombLoops` will detect this as a loop.
The simplest way of fixing this postpone the `CheckCombLoops` to HW, and don't execute it in FIRRTL Pass.

[...]

PR #4648, opened 2023-02-10
[FIRRTL] Limit BitCast to passive output, fix non-passive input.

dtzSiFive:
Casting bits to non-passive types is problematic re:flow, so limit the result to passive types.

Tweak BitCast's verifier to no longer reject non-passive inputs.

PR #8750, opened 2025-07-21
Add check for casting bundle with elements in different directions

unlsycn:
Add a check for the case described in #8610


dtzSiFive:
cc https://github.com/llvm/circt/pull/4648

Maybe we should just require the input to be `PassiveType` in ODS?

Disallowing it works for me, although not without uses.

If we need this we can create wire with passive version of the type and connect to it.

PR #8756, opened 2025-07-21
[Comb] Fix more recursive mux folders

### LINKED ISSUE #8754: [Firtool] Conversion of reduced Rocket HW IR to Verilog doesn't terminate or segfaults
Stumbled across another weird issue through circt-reduce - [this file](https://github.com/user-attachments/files/21347496/breaking.txt), for me, will fail to terminate, or segfault if I run it with `--debug`, with the following command: `firtool <file> --format=mlir --verilog`.

I found that if I pipe the debug-mode output into a file it won't segfault, but it still doesn't terminate and the debug output seems to be several output streams mangled together, e.g.:

```
eai3% e1%io_in_c_bits_sizet :   in sv.if : }ExportVerilog.cpp:0:0} -> , e sv.if %//===-------------------------------------------===//
ai      rs1arg1413, %r) -> _i_4% auto_icache_master_out_a_bits_address (

success%r : % {c0_i30Processing operation : 'sosv.assertts_, , ifalse_(a, 8%io_enq_bits_size : %i    #16queue_arw_deq_q.io_deq_bits_qos tpattern applied successfullyfalse%
false, hw.constant%'(s_ e2rio_in_6_ready%4,tisin =  : comb.muxifalse3sv.if,  %0x0000784e942a1ed3_
       { 
c0_i320x62fcb8354020ei%9t: arg1414,  o1s i 32 ) -> %queue_arw_deq_q.io_deq_bits_echo_tl_state_size 9// *** IR Dump After Pattern Application ***
sv.assert{       , ) {
rmfalse"_i, i1immediate_ c, e%io_in_c_bits_source4bin, {ifalse, __pthread_once_slow0
%      false%sv.assert  te,_
a%4, labellir : t , in
1 % "sv.assert,false _m     s1192, , %io_in_7_ready i3i_2%in       {queue_arw_deq_q.io_deq_bits_echo_tl_state_source./nptl/pthread_once.c:118:7 

 hw.module, %%false__immediate}siarg1416: "n, 8, 0auto_in_d_ready %auto_icache_master_out_d_validsv.assert
, %#17false 

    immediate c0_i2,ab e
2, iatiin" %io_enq_bits_source, :  i :        %,0x000062fcb1a6cfde}
 private,   @silabel    r, %1s.s3%io_in_c_bits_address
%2i, %sv.assertqueue_arw_deq_q.io_deq_bits_wen  
label   %immediatefalseTLMonitor_6st tsv.if_iarg1417, ae,  :     51,, ininfalse  = immediatestd::__future_base::_Deferred_state<std::thread::_Invoker<std::tuple<std::function<void ()>>>, void>::_M_complete_async()      * Pattern " , label%e(s" 619, io_out_validurii}32
   %io_enq_bits_sink, : %hw.instancelabelExportVerilog.cpp:0:0sv.if : 'a falserin_a%
"%: tt4,     in%%auto_icache_master_out_d_bits_opcode ifalse  
 scomb.extract, "t dsfalse} 
arg1418io_, sv.if 8 : immediatei 2,""#18%s -> (%a_%clockas{    , 
%1__i %io_in_c_bits_corrupt% 3label, ,  qa falsee )' {
sfalse6 : te
}

arg1419, ta2 : false: ininimmediateus0x000062fcb1a6c667r{Trying to match "s, 0!"seqar      t
,     } -> io_out_bits_writels, i  " a  es et
"
e%."
sv.assert_%sv.ifsuccess: _si1{, 
i%auto_icache_master_out_d_bits_sizes : %io_enq_bits_deniedlabelustd::_Function_handler<void (), std::shared_future<void> llvm::ThreadPoolInterface::asyncImpl<void>(std::function<void ()>, llvm::ThreadPoolTaskGroup*)::'lambda'()>::_M_invoke(std::_Any_data const&)r_      "rc0_i30tclock,      _arg1400  : , iie32in       3si :  e t_sv.assert" result , _, "}%a%at least one pattern matched%1nr, sv.assert%io_in_d_ready
e4i"_ExportVerilog.cpp:0:0_
a 0%_ini
falsesfalse
arg1401, _ti  : %  , rin1aa_#19s%
```

so I wasn't sure how to debug this one (though I don't know if the garbage output is due to some parallelism I can disable?).

TaoBi22:
This catches two more cases where muxes that have themselves as an input can cause a recursive folder in comb canonicalization, both triggered by the added integration test. Closes #8754

### DIFF
--- a/lib/Dialect/Comb/CombFolds.cpp
+++ b/lib/Dialect/Comb/CombFolds.cpp
@@ -2079,9 +2079,8 @@ static bool foldCommonMuxValue(MuxOp op, bool isTrueOperand,
     }
 
     auto isARecursiveMux = [](Value v) {
-      if (v.getDefiningOp())
-        if (auto muxOp = dyn_cast<MuxOp>(v.getDefiningOp()))
-          return muxOp.getTrueValue() == v || muxOp.getFalseValue() == v;
+      if (auto muxOp = v.getDefiningOp<MuxOp>())
+        return muxOp.getTrueValue() == v || muxOp.getFalseValue() == v;
       return false;
     };
 


lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -2078,6 +2078,17 @@ static bool foldCommonMuxValue(MuxOp op, bool isTrueOperand,
       return false;
     }
 
+    auto isARecursiveMux = [](Value v) {
+      if (v.getDefiningOp())
+        if (auto muxOp = dyn_cast<MuxOp>(v.getDefiningOp()))
+          return muxOp.getTrueValue() == v || muxOp.getFalseValue() == v;
```
fabianschuiki: Nit: You can combine these into a single call:

```suggestion
      if (auto muxOp = v.getDefiningOp<MuxOp>())
        return muxOp.getTrueValue() == v || muxOp.getFalseValue() == v;
```

fabianschuiki (approved):
LGTM, thanks for fixing this 🙌

fzi-hielscher (commented):
I'm wondering if it really makes sense to play whack-a-mole with this kind of issues, or if we shouldn't just bail out of `canonicalize` immediately when the op is trivially cyclic. Does anyone actually benefit from optimizations over cyclic comb expressions?

How about adding a more general helper function?
```
// Returns true if the op is in `comb` and if any of its operands
// is identical to any of its results.
static bool isTriviallyCyclicCombOp(Operation* op) {
  if (!op || !isa<comb::CombDialect>(op->getDialect()))
    return false;
  for (auto res: op->getResults()) {
    if (llvm::any_of(op->getOperands(), [res](auto operand){
      return operand == res;
    }))
      return true;
  }
  return false;
}
```
(Not tested!)
I'd expect this to come in handy at various places.

lib/Dialect/Comb/CombFolds.cpp
TaoBi22: Oh cool, thanks!

TaoBi22:
> I'm wondering if it really makes sense to play whack-a-mole with this kind of issues, or if we shouldn't just bail out of `canonicalize` immediately when the op is trivially cyclic. Does anyone actually benefit from optimizations over cyclic comb expressions?
> 
> How about adding a more general helper function?
> 
> ```c++
> // Returns true if the op is in `comb` and if any of its operands
> // is identical to any of its results.
> static bool isTriviallyCyclicCombOp(Operation* op) {
>   if (!op || !isa<comb::CombDialect>(op->getDialect()))
>     return false;
>   for (auto res: op->getResults()) {
>     if (llvm::any_of(op->getOperands(), [res](auto operand){
>       return operand == res;
>     }))
>       return true;
>   }
>   return false;
> }
> ```
> 
> (Not tested!) I'd expect this to come in handy at various places.

I agree that I think in general bailing out of canonicalization if the op is trivially cyclic probably makes sense and that having such a helper would be useful, but worth noting that these cases don't always occur when the root operation is trivially cyclic - in this case it's another operation that gets caught up in the fold. I guess maybe the answer is to call this on any op involved in a comb fold, but that's probably a topic for a later discussion (I can't make the next one, but maybe an ODM?). In the meantime I'm going to merge this just so we have the bug covered

PR #8758, opened 2025-07-21
[Comb] Add comb.reverse operation with Verilog Export

mafeguimaraes:
This PR introduces a new operation to the Comb dialect: `comb.reverse`, which performs bitwise reversal (mirroring) of an integer value.
It also adds Verilog export support for this operation using SystemVerilog’s streaming operator `{<<{}}`.

Bit reversal is a common operation in hardware design, especially in signal processing and communication protocols.
Previously, reversing bits required generating many explicit `assign` statements.
This new operation makes the IR cleaner and enables direct export to compact Verilog syntax.

### Example

**Original Verilog:**
```systemverilog
assign out[0] = in[7];
assign out[1] = in[6];
assign out[2] = in[5];
assign out[3] = in[4];
assign out[4] = in[3];
assign out[5] = in[2];
assign out[6] = in[1];
assign out[7] = in[0];
```

**MLIR**
`%rev = comb.reverse %in : (i8) -> i8`

**Verilog (exported)**
`assign rev = {<<{in}};`

### Implementation Details
The implementation is divided into three main parts:

1.  **Operation Definition (`CombOps.td`, `CombOps.cpp`):**
    * The `Comb_ReverseOp` is defined in `CombOps.td` with `Pure` and `SameOperandsAndResultType` traits.
    * The C++ implementation in `CombOps.cpp` includes:
        * **Constant Folding:** Handles bit-reversal for constant operands at compile time.
        * **Canonicalization Pattern:**  Simplifies `comb.reverse(comb.reverse(x))` to` x`.
        * Implements `verify()` for operand/result checks and `inferResultType()` for result inference.

2.  **Verilog Lowering (`ExportVerilog.cpp`):**
    * A dedicated emitter case for `comb::ReverseOp` was added.
    * Lowers the operation to SystemVerilog’s streaming operator `{<<{in}}`, generating compact and readable Verilog design.

3.  **Testing (`reverse.mlir, comb-reverse.mlir`):**
    * A new test file was created to verify:
        * Constant folding and canonicalization via `-canonicalize`.
        * Verilog emission via `-export-verilog` and `FileCheck`.


include/circt/Dialect/Comb/Combinational.td
```diff
@@ -331,3 +331,30 @@ def TruthTableOp : CombOp<"truth_table", [Pure]> {
...
+    ```
+    If %in is 4'b1101, then %out is 4'b1011.
+  }];
+
+  let arguments = (ins HWIntegerType:$input);
+  let results = (outs HWIntegerType:$result);
+
+  let hasFolder = 1;
+  let hasCanonicalizer = 1;
+  let hasVerifier = 1;
```
fabianschuiki: Since the verifier just returns `success()`, you can simplify the op by removing this `hasVerifier` line and removing the `verify` function implementation :slightly_smiling_face:

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -2598,6 +2598,14 @@ SubExprInfo ExprEmitter::emitSubExpr(Value exp,
...
                                      bool isSelfDeterminedUnsignedValue,
                                      bool isAssignmentLikeContext) {
+
+  if (auto reverseOp = dyn_cast_or_null<comb::ReverseOp>(exp.getDefiningOp())) {
+    ps << "{<<{";
+    emitSubExpr(reverseOp.getInput(), LowestPrecedence);
+    ps << "}}";
+    emittedExprs.insert(reverseOp);
+    return {Symbol, IsUnsigned};
+  }
```
fabianschuiki: It would be fantastic if you could move this emission into a separate `ExprEmitter::visitComb` function. We already have one such function per Comb operation. That will then properly insert sign casts, parenthesis, and other things around the streaming operator.

Since you are adding a Comb op, you'll also have to add the op to the Comb op visitor. The visitor is what calls the various `visitComb` functions. You'll need to do this in two places:
- https://github.com/llvm/circt/blob/main/include/circt/Dialect/Comb/CombVisitors.h#L41
- https://github.com/llvm/circt/blob/main/include/circt/Dialect/Comb/CombVisitors.h#L102-L106

lib/Dialect/Comb/CombOps.cpp
```diff
@@ -419,6 +419,55 @@ LogicalResult ConcatOp::inferReturnTypes(
...
+} 
+
+void comb::ReverseOp::getCanonicalizationPatterns(RewritePatternSet &results,
+                                                  MLIRContext *context) {
+  results.add<ReverseOfReverse>(context);
+}
+
+mlir::LogicalResult comb::ReverseOp::verify() {
+  return mlir::success();
+}
```
fabianschuiki: Not needed :slightly_smiling_face:

lib/Dialect/Comb/CombOps.cpp
```diff
@@ -419,6 +419,55 @@ LogicalResult ConcatOp::inferReturnTypes(
...
+
+void comb::ReverseOp::inferResultRanges(
+    llvm::ArrayRef<mlir::ConstantIntRanges> operandRanges,
+    llvm::function_ref<void(mlir::Value, const mlir::ConstantIntRanges &)>
+        setResultRangeFn) {
+  setResultRangeFn(getResult(), operandRanges.front());
+}
```
fabianschuiki: I'm not sure if this implementation of `inferResultRanges` is correct. As far as I understand it, that interface is supposed to compute upper and lower bounds for the integer value of the result. Since the `ReverseOp` reverses the bits of a number and therefore changes its value, the upper and lower bounds of the result are not the same as that of the input. The bit reversal probably has the annoying effect of splitting a contiguous range of integer values into potentially multiple ones.

Do any of your passes rely on having the `InferIntRangeInterface` implemented on this op? If not, I would delete this implementation and remove the `DeclareOpInterfaceMethods<InferIntRangeInterface, ["inferResultRanges"]>` trait from the op itself.

test/Conversion/ExportVerilog/comb-reverse.mlir
```diff
@@ -0,0 +1,17 @@
...
+// CHECK-NEXT:    output [7:0] out
+// CHECK-NEXT:  );
+// CHECK-EMPTY:
+// CHECK-NEXT:    assign out = {<<{in}};
+// CHECK-NEXT:  endmodule
+
+hw.module @test_reverse(in %in: i8, out out: i8) {
+  %reversed = comb.reverse %in : (i8) -> i8
+  hw.output %reversed : i8
+}
```
fabianschuiki: Nice! Could you move this into https://github.com/llvm/circt/blob/main/test/Conversion/ExportVerilog/verilog-basic.mlir? That's where the other Comb operations are tested :slightly_smiling_face:

test/Dialect/Comb/reverse.mlir
```diff
@@ -0,0 +1,32 @@
...
+  %rev_const = comb.reverse %c13 : (i8) -> i8
+
+  // Check outputs: after canonicalization,
+  // out_double_rev should be wired directly to %in,
+  // and out_const_rev should be wired to the folded constant (which is -80 in i8)
+  hw.output %rev2, %rev_const : i8, i8
+}
+
+// CHECK: %[[CST:.*]] = hw.constant -80 : i8
+// CHECK: hw.output %[[IN]], %[[CST]] : i8, i8
```
fabianschuiki: Could you move this into https://github.com/llvm/circt/blob/main/test/Dialect/Comb/canonicalization.mlir? That's where the Comb dialect canonicalization tests are located :slightly_smiling_face:

include/circt/Dialect/Comb/Combinational.td
```diff
@@ -331,3 +331,30 @@ def TruthTableOp : CombOp<"truth_table", [Pure]> {
...
+
+  let assemblyFormat = "$input attr-dict `:` functional-type($input, $result)";
```
uenoku: I think functional-type is not necessary as it has SameOperandsAndResultType

fabianschuiki: Good point, you should be able to just use
```
`:` type($input)
```
here.

mafeguimaraes:
Thank you both, @fabianschuiki and @uenoku, for the helpful review. I've implemented all the suggested changes. Please let me know if there is anything else. 🙂

fabianschuiki:
Hey @mafeguimaraes, thanks a lot! It looks like there are a few clang-format issues that CI isn't happy about. These are just code formatting guidelines. Check out "clang format patches display" in https://github.com/llvm/circt/actions/runs/16469136872/job/46573783262?pr=8758 for details.

The other CI failure seems to be related to an accidental LLVM downgrade in your branch. Could you undo the LLVM submodule change in your branch such that it lines up again with what's on "main"? That should clear up the CI build.

mafeguimaraes:
I've run clang-format to fix the style issues and reset the LLVM submodule to match main. Thank you, @fabianschuiki!

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -2598,6 +2599,7 @@ SubExprInfo ExprEmitter::emitSubExpr(Value exp,
                                      SubExprSignRequirement signRequirement,
                                      bool isSelfDeterminedUnsignedValue,
                                      bool isAssignmentLikeContext) {
+
```
darthscsi: nit: stray whitespace change

fabianschuiki:
Hey @mafeguimaraes, it looks like the LLVM submodule is still pointing to a different commit than what's on main, causing CI to break. Try something like the following to change the submodule to point to the same commit as is on main:
```
git checkout main -- llvm
```

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -3304,6 +3306,17 @@ SubExprInfo ExprEmitter::visitComb(MuxOp op) {
   });
 }
 
+SubExprInfo ExprEmitter::visitComb(ReverseOp op) {
+  if (hasSVAttributes(op))
+    emitError(op, "SV attributes emission is unimplemented for the op");
+
+  ps << "{<<{";
+  emitSubExpr(op.getInput(), LowestPrecedence);
```
darthscsi, line 3313: It's been a while since I dealt with this, but is LowestPrecedence needed?  The expression is insides brackets.  It might be worth checking the spec to avoid unnecessary parens being emitted.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -3304,6 +3306,17 @@ SubExprInfo ExprEmitter::visitComb(MuxOp op) {
...
+  ps << "}}";
+
+  return {Symbol, IsUnsigned};
```
darthscsi, line 3316: Should this be concatenation precedence?

darthscsi, line 3316: I'm having a hard time verifying unsigned results from the spec.  It seems true, but I can't find it.

darthscsi (approved):
Thanks!  a couple minor nits, but looking good.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
mafeguimaraes: You are right, since the expression is inside the {<<{...}} streaming operator, it's in a new context and will never need to be parenthesized.

In this emitter, passing LowestPrecedence to the recursive emitSubExpr call is the idiomatic way to achieve exactly that and prevent any unnecessary parens from being added.

lib/Conversion/ExportVerilog/ExportVerilog.cpp
mafeguimaraes: The streaming operator {<<{...}} should indeed have the same precedence as a regular concatenation.

mafeguimaraes: Yes, I had the same doubt. From what I found, the SystemVerilog streaming operator ({<<{}}) preserves the bit pattern without applying any sign extension or interpretation, which effectively makes the result unsigned.

mafeguimaraes:
That's great to hear! Thank you so much for the feedback and for guiding me through the process. I'm happy I could contribute. 😃

fabianschuiki:
Looks like there's a minor breakage in an LLHD pass associated with the CombVisitor. I think you just need to create a `visitComb` stub for the `ReverseOp` here: https://github.com/llvm/circt/blob/main/lib/Dialect/LLHD/Transforms/DesequentializationPass.cpp#L155

Interesting that this didn't show up for you in your local build 🤔

mafeguimaraes:
> Looks like there's a minor breakage in an LLHD pass associated with the CombVisitor. I think you just need to create a `visitComb` stub for the `ReverseOp` here: https://github.com/llvm/circt/blob/main/lib/Dialect/LLHD/Transforms/DesequentializationPass.cpp#L155
> 
> Interesting that this didn't show up for you in your local build 🤔

Good catch! My local build configuration didn't include the LLHD dialect, so I missed this dependency. I've added the necessary visitComb stub for ReverseOp to the CombInterpreter. Thanks for pointing it out!

fabianschuiki:
Looks like clang-format is unhappy about some trailing white space. Try running clang-format locally, or manually fix the whitespace :smiley:

fabianschuiki:
@mafeguimaraes You can run the tests locally with `ninja check-circt` or `make check-circt`, depending on what build system you are using. That should allow you to debug the CI failures more quickly, since CI basically runs clang-format and `ninja check-circt` as well :slightly_smiling_face:

mafeguimaraes:
Great, all checks are passing now. Thanks again for all the help!🙂

fabianschuiki:
Very nice, thanks a lot for adding this to CIRCT! :partying_face:

mafeguimaraes:
That's fantastic news! Thank you again, @fabianschuiki, for all your help and patience throughout the review process. 😁

fabianschuiki:
Happy to help :slightly_smiling_face: Looking forward to other fancy PRs and optimizations :grin:

sequencer:
Maybe this op should be expose to firrtl to replace the [Reverse API](https://www.chisel-lang.org/api/latest/chisel3/util/Reverse$.html)

PR #8763, opened 2025-07-22
[VerifToSMT] Fix lowering of no output, no result LEC op

fzi-hielscher:
#8701 made the result of the `verif.lec` operation optional. In case of a trivial `LogicEquivalenceCheckingOp` without any outputs, VerifToSMT would always attempt to replace the result without checking that it exists.

Thanks to @maerhart for spotting this.


maerhart (approved):
LGTM

PR #8765, opened 2025-07-23
[HWToSMT] Add ArrayInject lowering to HWToSMT

uenoku:
This commit implements the lowering of hw.array_inject operations to SMT expressions using smt.array.store operations. The lowering creates a new array with the specified element injected at the given index, with proper bounds checking to handle out-of-bounds accesses gracefully.

The implementation includes bounds checking that compares the index against the maximum valid index (numElements - 1) and uses smt.ite to return either the updated array (if in bounds) or the original array (if out of bounds).


maerhart (approved):
LGTM

PR #8769, opened 2025-07-23
[FIRRTL] make mux type inference support enumeration types

youngar:


include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
```diff
@@ -472,15 +472,15 @@ def BaseTypeAliasImpl : FIRRTLImplType<"BaseTypeAlias", [DeclareTypeInterfaceMet
...
+      FIRRTLBaseType getConstType(bool isConst) const;
       FIRRTLBaseType getAllConstDroppedType();
 
       /// Return the recursive properties of the type.
       RecursiveTypeProperties getRecursiveTypeProperties() const;
 
       // If a given `newInnerType` is identical to innerType, return `*this`
       // because we can reuse the type alias. Otherwise return `newInnerType`.
-      FIRRTLBaseType getModifiedType(FIRRTLBaseType newInnerType);
+      FIRRTLBaseType getModifiedType(FIRRTLBaseType newInnerType) const;
```
dtzSiFive, line 483: Maybe const-qualify these separately?

What's this for, just curious?

https://mlir.llvm.org/docs/Rationale/UsageOfConst/

Anyway separate commit works for me (as already done), thanks!

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -5404,6 +5404,33 @@ static FIRRTLBaseType inferMuxReturnType(FIRRTLBaseType high,
...
+  // of the other.
+  auto highEnum = type_dyn_cast<FEnumType>(high);
+  auto lowEnum = type_dyn_cast<FEnumType>(low);
+  if (lowEnum && highEnum) {
+    if (lowEnum.getNumElements() != highEnum.getNumElements())
+      return emitInferRetTypeError<FIRRTLBaseType>(
+          loc, "incompatible mux operand types, true value type: ", high,
+          ", false value type: ", low);
+    SmallVector<FEnumType::EnumElement> elements;
+    for (auto [high, low] : llvm::zip(highEnum, lowEnum)) {
```
dtzSiFive:
```suggestion
    for (auto [high, low] : llvm::zip_equal(highEnum, lowEnum)) {
```
?

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
youngar: Yeah, I needed these when zipping together the enum types, which gives constant references to the elements, which I then need to be able to convert to the const version of the type.  I might have been able to just add a non-const iterator to the enum type, but this seemed a bit better to me.

PR #8772, opened 2025-07-23
[FIRRTL] FRT: support creating 0-valued enums

youngar:
This adds the ability to the FullResetTransform to create enumeration values which are 0.  When the enumeration has a valid variant that is encoded as a 0, we create this value using the FEnumCreateOp.  If there is no valid variant with the value 0, we create a constant 0 and bitcast it to the enumeration type.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/InferResets.cpp
@@ -141,6 +141,21 @@ static Value createZeroValue(ImplicitLocOpBuilder &builder, FIRRTLBaseType type,
             return builder.create<ConstantOp>(
                 type, APInt::getZero(type.getWidth().value_or(1)));
           })
+          .Case<FEnumType>([&](auto type) -> Value {
+            // There might not be a variant that corresponds to 0, in which case
+            // we have to create a 0 value and bitcast it to the enum.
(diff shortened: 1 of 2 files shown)


rwy7 (approved):

lib/Dialect/FIRRTL/Transforms/InferResets.cpp
```diff
@@ -141,6 +141,20 @@ static Value createZeroValue(ImplicitLocOpBuilder &builder, FIRRTLBaseType type,
...
+            // There might not be a variant that corresponds to 0, in which case
+            // we have to create a 0 value and bitcast it to the enum.
+            if (type.getNumElements() != 0 &&
+                type.getElement(0).value.getValue().isZero()) {
+              const auto &element = type.getElement(0);
+              auto value = createZeroValue(builder, element.type, cache);
+              return builder.create<FEnumCreateOp>(type, element.name, value);
+            }
+            auto value = builder.create<ConstantOp>(
+                UIntType::get(builder.getContext(), type.getBitWidth(), true),
```
dtzSiFive: hyper-nit 
```suggestion
                UIntType::get(builder.getContext(), type.getBitWidth(), /*isConst=*/true),
```
?

Mostly because I couldn't remember what this flag did.

dtzSiFive (approved):

PR #8773, opened 2025-07-23
[circt-verilog] Add register-to-memory pass to pipeline

fabianschuiki:
Add the new RegOfVecToMem pass to the circt-verilog pipeline. This will detect memories described as `always` blocks and map them from the current `seq.firreg` representation to the correpsonding `seq.firmem`. This allows later parts of the pipeline to reason about memories more easily and transform them if needed.


uenoku (approved):
LGTM

PR #8774, opened 2025-07-24
[HWToLLVM] Add lowering support for 'hw.array_inject' op

mvpant:
This adds lowering support for the `hw.array_inject` operation into `llvm` dialect.

### DIFF
--- a/lib/Conversion/HWToLLVM/HWToLLVM.cpp
+++ b/lib/Conversion/HWToLLVM/HWToLLVM.cpp
@@ -79,26 +79,6 @@ static Value zextByOne(Location loc, ConversionPatternRewriter &rewriter,
   return rewriter.create<LLVM::ZExtOp>(loc, zextTy, value);
 }
 
-static Value reallocArrayOnStack(Location loc,
-                                 ConversionPatternRewriter &rewriter,
-                                 Value arrInput) {
-  Value arrPtr;
-  if (auto load = arrInput.getDefiningOp<LLVM::LoadOp>()) {
-    // In this case the array was loaded from an existing address, so we can
-    // just grab that address instead of reallocating the array on the stack.
-    arrPtr = load.getAddr();
-  } else {
-    auto oneC = rewriter.create<LLVM::ConstantOp>(
-        loc, rewriter.getI32Type(), rewriter.getI32IntegerAttr(1));
-    arrPtr = rewriter.create<LLVM::AllocaOp>(
-        loc, LLVM::LLVMPointerType::get(rewriter.getContext()),
-        arrInput.getType(), oneC,
-        /*alignment=*/4);
-    rewriter.create<LLVM::StoreOp>(loc, arrInput, arrPtr);
-  }
-  return arrPtr;
-}
-
 //===----------------------------------------------------------------------===//
 // Extraction operation conversions
 //===----------------------------------------------------------------------===//
@@ -167,8 +147,13 @@ struct ArrayInjectOpConversion
   LogicalResult
   matchAndRewrite(hw::ArrayInjectOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    Value arrPtr =
-        reallocArrayOnStack(op->getLoc(), rewriter, adaptor.getInput());
+    auto oneC = rewriter.create<LLVM::ConstantOp>(
+        op->getLoc(), rewriter.getI32Type(), rewriter.getI32IntegerAttr(1));
+    auto arrPtr = rewriter.create<LLVM::AllocaOp>(
+        op->getLoc(), LLVM::LLVMPointerType::get(rewriter.getContext()),
+        adaptor.getInput().getType(), oneC,
+        /*alignment=*/4);
+    rewriter.create<LLVM::StoreOp>(op->getLoc(), adaptor.getInput(), arrPtr);
 
     auto arrTy = typeConverter->convertType(op.getInput().getType());
     auto zextIndex = zextByOne(op->getLoc(), rewriter, op.getIndex());
@@ -193,8 +178,22 @@ struct ArrayGetOpConversion : public ConvertOpToLLVMPattern<hw::ArrayGetOp> {
   LogicalResult
   matchAndRewrite(hw::ArrayGetOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    Value arrPtr =
-        reallocArrayOnStack(op->getLoc(), rewriter, adaptor.getInput());
+
+    Value arrPtr;
+    if (auto load = adaptor.getInput().getDefiningOp<LLVM::LoadOp>()) {
+      // In this case the array was loaded from an existing address, so we can
+      // just grab that address instead of reallocating the array on the stack.
+      arrPtr = load.getAddr();
+    } else {
+      auto oneC = rewriter.create<LLVM::ConstantOp>(
+          op->getLoc(), IntegerType::get(rewriter.getContext(), 32),
+          rewriter.getI32IntegerAttr(1));
+      arrPtr = rewriter.create<LLVM::AllocaOp>(
+          op->getLoc(), LLVM::LLVMPointerType::get(rewriter.getContext()),
+          adaptor.getInput().getType(), oneC,
+          /*alignment=*/4);
+      rewriter.create<LLVM::StoreOp>(op->getLoc(), adaptor.getInput(), arrPtr);
+    }
 
     auto arrTy = typeConverter->convertType(op.getInput().getType());
     auto elemTy = typeConverter->convertType(op.getResult().getType());
--- a/lib/Conversion/HWToLLVM/HWToLLVM.cpp
+++ b/lib/Conversion/HWToLLVM/HWToLLVM.cpp
@@ -147,34 +147,50 @@ struct ArrayInjectOpConversion
   LogicalResult
   matchAndRewrite(hw::ArrayInjectOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    auto oneC = rewriter.create<LLVM::ConstantOp>(
-        op->getLoc(), rewriter.getI32Type(), rewriter.getI32IntegerAttr(1));
-    auto arrPtr = rewriter.create<LLVM::AllocaOp>(
-        op->getLoc(), LLVM::LLVMPointerType::get(rewriter.getContext()),
-        adaptor.getInput().getType(), oneC,
-        /*alignment=*/4);
-    rewriter.create<LLVM::StoreOp>(op->getLoc(), adaptor.getInput(), arrPtr);
+    auto inputType = cast<hw::ArrayType>(op.getInput().getType());
+    auto oldArrTy = typeConverter->convertType(inputType);
+    auto newArrTy = oldArrTy;
+    const size_t arrElems = inputType.getNumElements();
 
-    auto arrTy = typeConverter->convertType(op.getInput().getType());
-    auto zextIndex = zextByOne(op->getLoc(), rewriter, op.getIndex());
-    const size_t numElems =
-        cast<hw::ArrayType>(op.getInput().getType()).getNumElements();
-
-    if (numElems != 0 && (numElems == 1 || !llvm::isPowerOf2_64(numElems))) {
-      // Clamp index to prevent OOB access.
-      auto indexMax = rewriter.create<LLVM::ConstantOp>(
-          op->getLoc(), zextIndex.getType(),
-          rewriter.getI32IntegerAttr(numElems - 1));
-      zextIndex =
-          rewriter.create<LLVM::UMinOp>(op->getLoc(), zextIndex, indexMax);
+    if (arrElems == 0) {
+      rewriter.replaceOp(op, adaptor.getInput());
+      return success();
     }
 
+    auto oneC = rewriter.create<LLVM::ConstantOp>(
+        op->getLoc(), rewriter.getI32Type(), rewriter.getI32IntegerAttr(1));
+    auto zextIndex = zextByOne(op->getLoc(), rewriter, op.getIndex());
+
+    Value arrPtr;
+    if (arrElems == 1 || !llvm::isPowerOf2_64(arrElems)) {
+      // Clamp index to prevent OOB access. We add an extra element to the
+      // array so that OOB access modifies this element, leaving the original
+      // array intact.
+      auto maxIndex = rewriter.create<LLVM::ConstantOp>(
+          op->getLoc(), zextIndex.getType(),
+          rewriter.getI32IntegerAttr(arrElems));
+      zextIndex =
+          rewriter.create<LLVM::UMinOp>(op->getLoc(), zextIndex, maxIndex);
+
+      newArrTy = typeConverter->convertType(
+          hw::ArrayType::get(inputType.getElementType(), arrElems + 1));
+      arrPtr = rewriter.create<LLVM::AllocaOp>(
+          op->getLoc(), LLVM::LLVMPointerType::get(rewriter.getContext()),
+          newArrTy, oneC, /*alignment=*/4);
+    } else {
+      arrPtr = rewriter.create<LLVM::AllocaOp>(
+          op->getLoc(), LLVM::LLVMPointerType::get(rewriter.getContext()),
+          newArrTy, oneC, /*alignment=*/4);
+    }
+
+    rewriter.create<LLVM::StoreOp>(op->getLoc(), adaptor.getInput(), arrPtr);
+
     auto gep = rewriter.create<LLVM::GEPOp>(
-        op->getLoc(), LLVM::LLVMPointerType::get(rewriter.getContext()), arrTy,
-        arrPtr, ArrayRef<LLVM::GEPArg>{0, zextIndex});
+        op->getLoc(), LLVM::LLVMPointerType::get(rewriter.getContext()),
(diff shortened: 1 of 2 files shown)


maerhart (approved):
LGTM thanks!

fzi-hielscher:
Is there a reason to not use `llvm.insertvalue`? As we are converting `!hw.array` to `!llvm.array` this would seem like the direct equivalent to `hw.array_inject` to me.

Also, unless I'm getting my pointers mixed up, the test case in `convertConstArray` looks like it is performing a store to the global constant `GLOB1`.

maerhart:
Great catch. We have to make a copy of the array if there are other users that access the array after the inject op.

`insertvalue` does not work because its index has to be a constant (if I'm reading the reference manual correctly), which we don't always have with the inject.

mvpant:
> Is there a reason to not use `llvm.insertvalue`? As we are converting `!hw.array` to `!llvm.array` this would seem like the direct equivalent to `hw.array_inject` to me.

No specific reason, just to maintain consistency with the lowering of `hw.array_get`. Personally, i just think `extractvalue` and `insertvalue` are for struct types. Using `insertvalue` here would mean changing `hw.array_get` to `extractvalue` to keep things the same.

> Also, unless I'm getting my pointers mixed up, the test case in convertConstArray looks like it is performing a store to the global constant GLOB1

```
module {
  llvm.mlir.global internal constant @_aggregate_const_global(dense<[1, 0]> : tensor<2xi32>) {addr_space = 0 : i32} : !llvm.array<2 x i32>
  llvm.mlir.global internal constant @_aggregate_const_global_0(dense<[[3, 2], [1, 0]]> : tensor<2x2xi32>) {addr_space = 0 : i32} : !llvm.array<2 x array<2 x i32>>
  llvm.mlir.global internal @_aggregate_const_global_1() {addr_space = 0 : i32} : !llvm.array<2 x struct<(i1, i32)>> {
    %0 = llvm.mlir.undef : !llvm.array<2 x struct<(i1, i32)>>
    %1 = llvm.mlir.undef : !llvm.struct<(i1, i32)>
    %2 = llvm.mlir.constant(false) : i1
    %3 = llvm.insertvalue %2, %1[0] : !llvm.struct<(i1, i32)>
    %4 = llvm.mlir.constant(2 : i32) : i32
    %5 = llvm.insertvalue %4, %3[1] : !llvm.struct<(i1, i32)>
    %6 = llvm.insertvalue %5, %0[0] : !llvm.array<2 x struct<(i1, i32)>>
    %7 = llvm.mlir.undef : !llvm.struct<(i1, i32)>
    %8 = llvm.mlir.constant(true) : i1
    %9 = llvm.insertvalue %8, %7[0] : !llvm.struct<(i1, i32)>
    %10 = llvm.mlir.constant(0 : i32) : i32
    %11 = llvm.insertvalue %10, %9[1] : !llvm.struct<(i1, i32)>
    %12 = llvm.insertvalue %11, %6[1] : !llvm.array<2 x struct<(i1, i32)>>
    llvm.return %12 : !llvm.array<2 x struct<(i1, i32)>>
  }
  func.func @convertConstArray(%arg0: i1, %arg1: i32) {
    %0 = llvm.mlir.addressof @_aggregate_const_global : !llvm.ptr
    %1 = llvm.load %0 : !llvm.ptr -> !llvm.array<2 x i32>
    %2 = llvm.zext %arg0 : i1 to i2
    %3 = llvm.getelementptr %0[0, %2] : (!llvm.ptr, i2) -> !llvm.ptr, !llvm.array<2 x i32>
    %4 = llvm.load %3 : !llvm.ptr -> i32
    %5 = llvm.mlir.addressof @_aggregate_const_global_0 : !llvm.ptr
    %6 = llvm.load %5 : !llvm.ptr -> !llvm.array<2 x array<2 x i32>>
    %7 = llvm.mlir.addressof @_aggregate_const_global_0 : !llvm.ptr
    %8 = llvm.load %7 : !llvm.ptr -> !llvm.array<2 x array<2 x i32>>
    %9 = llvm.mlir.addressof @_aggregate_const_global_1 : !llvm.ptr
    %10 = llvm.load %9 : !llvm.ptr -> !llvm.array<2 x struct<(i1, i32)>>
    %11 = llvm.zext %arg0 : i1 to i2
    %12 = llvm.getelementptr %0[0, %11] : (!llvm.ptr, i2) -> !llvm.ptr, !llvm.array<2 x i32>
    llvm.store %arg1, %12 : i32, !llvm.ptr
    %13 = llvm.load %0 : !llvm.ptr -> !llvm.array<2 x i32>
    return
  }
}
```
and then into llvm ir:
```
; ModuleID = 'LLVMDialectModule'
source_filename = "LLVMDialectModule"

@_aggregate_const_global = internal constant [2 x i32] [i32 1, i32 0]
@_aggregate_const_global_0 = internal constant [2 x [2 x i32]] [[2 x i32] [i32 3, i32 2], [2 x i32] [i32 1, i32 0]]
@_aggregate_const_global_1 = internal global [2 x { i1, i32 }] [{ i1, i32 } { i1 false, i32 2 }, { i1, i32 } { i1 true, i32 0 }]

declare void @exit(i32)

define void @convertConstArray(i1 %0, i32 %1) {
  %3 = zext i1 %0 to i2
  %4 = getelementptr [2 x i32], ptr @_aggregate_const_global, i32 0, i2 %3
  store i32 %1, ptr %4, align 4
  ret void
}

!llvm.module.flags = !{!0}

!0 = !{i32 2, !"Debug Info Version", i32 3}
```
Looks alright?

fzi-hielscher:
Both `hw.array_inject` and `llvm.insertvalue` operate on values, not on references. So you have to guarantee that they have no effect on other users of their operands. By storing into `@_aggregate_const_global` you are modifying the array that `%1 = hw.array_get %0[%arg0] : !hw.array<2xi32>, i1` accesses, which should be a constant.
Storing into an array marked as constant will definitely cause trouble. Also, if you change your example to make it return the value of `%4` and remove the `const` attribute from `@_aggregate_const_global` you'll get after optimization:

```LLVM
; ModuleID = '<stdin>'
source_filename = "LLVMDialectModule"

@_aggregate_const_global = internal unnamed_addr global [2 x i32] [i32 1, i32 0]

; Function Attrs: mustprogress nofree norecurse nosync nounwind willreturn memory(readwrite, argmem: none, inaccessiblemem: none)
define i32 @convertConstArray(i1 %0, i32 %1) local_unnamed_addr #0 {
  %3 = zext i1 %0 to i64
  %4 = getelementptr [2 x i32], ptr @_aggregate_const_global, i64 0, i64 %3
  %5 = load i32, ptr %4, align 4
  store i32 %1, ptr %4, align 4
  ret i32 %5
}

attributes #0 = { mustprogress nofree norecurse nosync nounwind willreturn memory(readwrite, argmem: none, inaccessiblemem: none) }

!llvm.module.flags = !{!0}

!0 = !{i32 2, !"Debug Info Version", i32 3}
```

So, if you call `@convertConstArray` twice, on the second call the original `hw.aggregate_constant` will be broken. 

> insertvalue does not work because its index has to be a constant (if I'm reading the reference manual correctly), which we don't always have with the inject.

Yeah, right. That is unfortunate. So there is no other option than to create a copy. It would be nice if we had a way to collapse chains of `hw.array_inject` so they could operate on the same copy instead of creating a series of allocas. But that wont be possible with a conversion pattern. So, either we have to create a separate lowering for this op or hope that LLVM's `opt` will take care of it in the end.

fabianschuiki:
@fzi-hielscher Maybe we could do a quick check if the current inject op is the only user, and only create a copy when there are other users which we might disturb by modifying values?

fzi-hielscher:
Annoyingly, we won't be able to do this during dialect conversion, as the patterns are not allowed to look at the op's users.

mvpant:
Alright, the pattern now always makes a copy of input array.

> Yeah, right. That is unfortunate. So there is no other option than to create a copy. It would be nice if we had a way to collapse chains of hw.array_inject so they could operate on the same copy instead of creating a series of allocas. But that wont be possible with a conversion pattern. So, either we have to create a separate lowering for this op or hope that LLVM's opt will take care of it in the end.

One option would be to modify `hw.array_inject` operation to accept multiple indices and multiple values, so a chain can be collapsed into one operation. Other option might be an analysis pass, which would somehow say which `hw.array_inject` operations should make a copy and which should modify its input directly. That could be done both inside `HWToLLVMLoweringPass::runOnOperation` and `LowerArcToLLVMPass::runOnOperation`, then maybe passed into `populateHWToLLVMConversionPatterns`.

> Annoyingly, we won't be able to do this during dialect conversion, as the patterns are not allowed to look at the op's users.

```
    for (auto *user : op.getInput().getUsers()) {
      if (user != op.getOperation())
        user->dump();
    }
```
prints:
```
%0 = unrealized_conversion_cast %arg1 : !hw.array<2xi32> to !llvm.array<2 x i32>
%22 = hw.array_concat %arg1, %arg1 : !hw.array<2xi32>, !hw.array<2xi32>
%22 = hw.array_concat %arg1, %arg1 : !hw.array<2xi32>, !hw.array<2xi32>
%12 = hw.array_slice %arg1[%arg0] : (!hw.array<2xi32>) -> !hw.array<1xi32>
%6 = hw.array_get %arg1[%arg0] : !hw.array<2xi32>, i1
```
for this
```
func.func @convertArray(%arg0 : i1, %arg1: !hw.array<2xi32>, %arg2: i32, %arg3: i32, %arg4: i32, %arg5: i32) {
  %0 = hw.array_get %arg1[%arg0] : !hw.array<2xi32>, i1
  %1 = hw.array_slice %arg1[%arg0] : (!hw.array<2xi32>) -> !hw.array<1xi32>
  %2 = hw.array_concat %arg1, %arg1 : !hw.array<2xi32>, !hw.array<2xi32>
  %3 = hw.array_create %arg2, %arg3, %arg4, %arg5 : i32
  %4 = hw.array_inject %arg1[%arg0], %arg2 : !hw.array<2xi32>, i1
  return
}
```

fzi-hielscher:
Thanks @mvpant, looks good. But after realizing that the index is dynamic, I'm afraid I have to ask for yet another fix. 😅
The size of `hw.array` is not guaranteed to be a power of two, so the index value can go out out bounds in that case. While `hw.array_inject` does not specify any behavior for this, we need to make sure that the store does not overwrite adjacent stack content. Two possible ways of doing this:
- Check the index against the size and branch over the store if it is OOB.
- Just pad the allocation to the next power of two.

Assuming that we are mostly dealing with small-ish arrays, I think I'd rather waste some stack memory than to add control flow. But that would make it even more desirable to avoid adding unnecessary allocas.

(BTW, we really should fix that for Array get too, #6956 😬 )

> One option would be to modify hw.array_inject operation to accept multiple indices and multiple values, so a chain can be collapsed into one operation. 

It would turn `hw.array_inject` into a bit of a monster, but thinking about it, it might actually be the most pragmatic solution. Considering that there already is no direct equivalent in either SystemVerilog or LLVM IR, we might as well go all in and try to collapse array ops as much as possible in the HW dialect. That would probably also benefit the SV lowering as we would not have to create as many `always_comb` processes. WDYT @fabianschuiki ?

> Other option might be an analysis pass, which would somehow say which hw.array_inject operations should make a copy and which should modify its input directly. That could be done both inside HWToLLVMLoweringPass::runOnOperation and LowerArcToLLVMPass::runOnOperation, then maybe passed into populateHWToLLVMConversionPatterns.

During dialect conversion the IR is in a kind of a limbo state and it is notoriously unclear what you are allowed to do in the patterns:
https://discourse.llvm.org/t/rfc-a-new-one-shot-dialect-conversion-driver/79083

E.g., if there was a pattern in the conversion that lowered _to_ a HWArrayInjectOp, the new operation would not be committed to the IR immediately, which could then cause the `ArrayInjectOpConversion` to incorrectly see no other users of its operand. 
So, I'm not even sure if it is safe to use the result of an analysis in a conversion pattern. If we do not change the operation itself, I'd rather do a pre-conversion pass that 'manually' converts chunks of array ops to LLVM and then let the dialect conversion pick up the remaining ops.

fabianschuiki:
Great idea to carefully check the bounds. I think we've run into this before where Arcilator (and also Verilator in some situations) would access OOB memory and corrupt the stack or other signal values. Maybe a `mod` on the index would be faster than a branch misprediction?

Regarding chained injects: I wouldn't worry too much about the performance for now. @darthscsi points out that it's often worth keeping operations simple, especially if an increase in op complexity doesn't lead to a corresponding decrease in IR complexity. In practice, multiple `array_inject`s would probably be canonicalized to an `array_create`, `array_slice`, and `array_concat`, such that only a few cases of `array_inject` would survive. There's also a pass in the circt-verilog pipeline now that tries to map some of the `array_inject`s on `seq.firreg` to `seq.firmem`, where we can directly deal with memory behind a pointer 🙂. I'd just go for correctness in this conversion pass for now.

We could also think about adding a pass before LLVM lowering later that converts large `!hw.array`s to something based on memrefs and an allocation, where we can be more careful about when copies are needed. That could also be an opportunity to map small `!hw.array`s to `i64` and friends, such that we can do most ops as bit manipulation.

fzi-hielscher:
> Maybe a mod on the index would be faster than a branch misprediction?

Overwriting other elements would clash with Verilog behavior, IIRC. But I guess just allocating a single extra element and then using `llvm.intr.umin %index, %arraySize` to clamp the index should do the trick with minimal overhead.

fabianschuiki:
That's a nice idea! 🥳

mvpant:
> Overwriting other elements would clash with Verilog behavior, IIRC. But I guess just allocating a single extra element and then using `llvm.intr.umin %index, %arraySize` to clamp the index should do the trick with minimal overhead.

Clamped the index, but not sure why we need an extra element? Can't we just do `llvm.intr.umin %index, %arraySize - 1` to avoid bitcast?

BTW, while looking at `IndexBitWidthConstraint`, i noticed that currently
 ```
%0 = hw.array_inject %arg0[%arg5], %arg4 : !hw.array<0xi32>, i64
```
 is considered a valid operation. Is it right?

**Comment by fzi-hielscher** [comment]
> Clamped the index, but not sure why we need an extra element? Can't we just do llvm.intr.umin %index, %arraySize - 1 to avoid bitcast?

By doing so, you will change an element that is within the bounds of the array. Going by the formal semantics of `hw.array_inject`, [as they are currently implemented](https://github.com/llvm/circt/blob/main/lib/Conversion/HWToSMT/HWToSMT.cpp#L184), an OOB access _must_ return an array that is identical to the input argument.* By redirecting the store to a location that technically does not belong to the array, we can guarantee this without adding control flow. 
(Whether adding a branch op would actually have a measurable impact on performance is questionable. But not doing so should at least keep the lowering simpler.)
I don't think you need to bitcast. It is perfectly legal to access an array outside its static type bounds as long as it is backed-up by an allocation, see: https://llvm.org/docs/LangRef.html#array-type
So you should only need to change the `elem_type` of the AllocaOp. All accesses to the pointer can still use the "correct" type.

\* Not sure we should keep it that way, though.

> BTW, while looking at IndexBitWidthConstraint, i noticed that currently 
`%0 = hw.array_inject %arg0[%arg5], %arg4 : !hw.array<0xi32>, i64`
is considered a valid operation. Is it right?

Very well spotted! That is an interesting quirk and I'm not sure it is intentional. But, then again, there is not much to be gained from debating the "right" index type for a zero-width array. I think the best you can do in this lowering is to check for zero-width early and then just replace the result of the operation with the input array.

**Comment by mvpant** [comment]
> By doing so, you will change an element that is within the bounds of the array. Going by the formal semantics of `hw.array_inject`, [as they are currently implemented](https://github.com/llvm/circt/blob/main/lib/Conversion/HWToSMT/HWToSMT.cpp#L184), an OOB access _must_ return an array that is identical to the input argument.*

I see, thanks for the clarification. Added an extra element.

File: lib/Conversion/HWToLLVM/HWToLLVM.cpp [hunk 1]
```diff
@@ -135,6 +135,67 @@ struct StructExtractOpConversion
...
+///   load(alloca)
+struct ArrayInjectOpConversion
+    : public ConvertOpToLLVMPattern<hw::ArrayInjectOp> {
+  using ConvertOpToLLVMPattern<hw::ArrayInjectOp>::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(hw::ArrayInjectOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto inputType = cast<hw::ArrayType>(op.getInput().getType());
+    auto oldArrTy = typeConverter->convertType(inputType);
```
fzi-hielscher: Nit: `adaptor.getInput().getType()` would avoid the extra conversion.

lib/Conversion/HWToLLVM/HWToLLVM.cpp
```diff
@@ -135,6 +135,67 @@ struct StructExtractOpConversion
...
+      rewriter.replaceOp(op, adaptor.getInput());
+      return success();
+    }
+
+    auto oneC = rewriter.create<LLVM::ConstantOp>(
+        op->getLoc(), rewriter.getI32Type(), rewriter.getI32IntegerAttr(1));
+    auto zextIndex = zextByOne(op->getLoc(), rewriter, op.getIndex());
+
+    Value arrPtr;
+    if (arrElems == 1 || !llvm::isPowerOf2_64(arrElems)) {
```
fzi-hielscher, line 165: Great job catching the `arrElems == 1` case. 👍

fabianschuiki:
:partying_face: Thanks a lot for catching all these different corner cases and issues with OOB access behavior!

mvpant:
> Should I merge on your behalf?

Sure, thanks!

PR #8775, opened 2025-07-24
[SV][ETC] Quick fix for extraction involving $signed/$unsigned.

### LINKED ISSUE #8771: [SV] ETC crash with signed printf operand
Input:
```firrtl
FIRRTL version 5.1.0
circuit SignedPrintETC :

  public module SignedPrintETC :
    input clk : Clock
    input test : SInt<64>

    printf(clk, UInt<1>(0h1), "%0d", test)
```

When using `--extract-test-code` this produces invalid IR:

```
signed-print-etc.fir:4:10: error: operand #0 does not dominate this use
  public module SignedPrintETC :
         ^
signed-print-etc.fir:4:10: note: see current operation: "hw.instance"(%2, %1, %0) <{argNames = ["port_0", "port_1", "port_2"], doNotPrint, inner_sym = #hw<innerSym@__ETC_SignedPrintETC_assert>, instanceName = "SignedPrintETC_assert", moduleName = @SignedPrintETC_assert, parameters = [], resultNames = []}> : (i64, i1, i1) -> ()
signed-print-etc.fir:8:5: note: operand defined here (op in a child region)
    printf(clk, UInt<1>(0h1), "%0d", test)
    ^
```

dtzSiFive:
Just check for these specifically in the classification logic.

Fixes #8771.

### DIFF
--- a/lib/Dialect/SV/Transforms/SVExtractTestCode.cpp
+++ b/lib/Dialect/SV/Transforms/SVExtractTestCode.cpp
@@ -631,6 +631,11 @@ bool isInDesign(hw::HWSymbolCache &symCache, Operation *op,
   if (isa<TimeOp, sv::FuncCallProceduralOp>(op))
     return false;
 
+  if (auto system = dyn_cast<SystemFunctionOp>(op);
+      system &&
+      (system.getFnName() == "signed" || system.getFnName() == "unsigned"))
+    return false;
+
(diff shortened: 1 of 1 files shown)


dtzSiFive:
This isn't how I'd like to fix this but leaving this here so we can workaround the user issue immediately if needed while sorting this out.

PR #8778, opened 2025-07-25
[ESI] Always find_package Python3 Development

mortbopet:
AFAICT, pybind11 v. 3.0.0 (release 2 weeks ago) changed a bit how modules are being defined. A result of that is that it's not sufficient to just look for `Python3_FOUND`, since pybind module definition will fail inside the pybind CMake files if the `Development` component hasn't been enabled.

That can occur when e.g. building the ESI runtime alongside the rest of CIRCT, if python bindings are _not_ enabled. In that case [here](https://github.com/llvm/circt/blob/main/CMakeLists.txt) in the top-level CMakeLists, we **don't** look for the Python3 `Development` component, which then triggers the issue. I'm a bit surprised that this isn't done inside the pybind11 cmake files (i.e. ensuring that `Development` is present), but oh well... the problem is fixed by always `find_package`'ing for both `Interpreter` and `Development` in the ESI runtime CMakeLists.txt.

Also sneaks in a .gitignore for the `ESI/runtime` library, to better support in-tree builds of just the runtime.


teqdruid (approved):
I seem to recall a particular reason I did this, but as long as it works...

PR #8779, opened 2025-07-25
[ESI] Disallow calls on an unconnected `FuncService::Function`

mortbopet:
because currently, that'll result in segfaults.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
+++ b/lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
@@ -284,6 +284,7 @@ public:
     std::mutex callMutex;
     WriteChannelPort *arg;
     ReadChannelPort *result;
+    bool connected = false;
   };
 
 private:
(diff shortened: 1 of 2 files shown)


lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
```diff
@@ -284,6 +284,7 @@ class FuncService : public Service {
     std::mutex callMutex;
     WriteChannelPort *arg;
     ReadChannelPort *result;
+    bool connected = false;
```
teqdruid, line 287: Instead of storing a third connected status, you could query the underlying read/write ports to see if they're both connected. In general, I prefer to avoid redundant state.

teqdruid (approved):

lib/Dialect/ESI/runtime/cpp/include/esi/Services.h
mortbopet: Totally agree, but when digging into the sources, i was a bit unsure about what the general approach to `BundlePort`s' overall connectedness-status is. I suspect this isn't the only case where a `BundlePort` needs to connect to some underlying set of ports, which probably warrants something more generic...

PR #8782, opened 2025-07-25
[ESI] Better single-argument handling for FunctionPort

mortbopet:
Previously, if one has a function with only a single `BitsType` argument, there was no way that this code could provide a correct argument to be serialized, given that only `kwargs` would be accepted - which is a dict, and not serializeable by `BitsType`.

Modified argument handling abit s.t.
1. users have to provide either kwargs or positional
2. if positional arguments are provided, and only a single is present, it is unpacked, ensuring that the validation and serialization code doesn't recieve a tuple, but instead the underlying value.

### DIFF
--- a/lib/Dialect/ESI/runtime/python/esiaccel/types.py
+++ b/lib/Dialect/ESI/runtime/python/esiaccel/types.py
@@ -448,14 +448,27 @@ class FunctionPort(BundlePort):
     self.cpp_port.connect()
     self.connected = True
 
-  def call(self, **kwargs: Any) -> Future:
+  def call(self, *args: Any, **kwargs: Any) -> Future:
     """Call the function with the given argument and returns a future of the
     result."""
-    valid, reason = self.arg_type.is_valid(kwargs)
+
+    # Accept either positional or keyword arguments, but not both.
+    if len(args) > 0 and len(kwargs) > 0:
+      raise ValueError("cannot use both positional and keyword arguments")
+
+    # Handle arguments: for single positional arg, unwrap it from tuple
+    if len(args) == 1:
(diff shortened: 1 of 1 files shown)


lib/Dialect/ESI/runtime/python/esiaccel/types.py
```diff
@@ -448,14 +448,27 @@ def connect(self):
...
-    valid, reason = self.arg_type.is_valid(kwargs)
+
+    # Accept either positional or keyword arguments, but not both.
+    if len(args) > 0 and len(kwargs) > 0:
+      raise ValueError("cannot use both positional and keyword arguments")
+
+    # Handle arguments: for single positional arg, unwrap it from tuple
+    if len(args) == 1:
+      selected = args[0]
+    elif len(args) > 1:
```
teqdruid, line 462: This sort of special casing doesn't sit right with me. As written, multiple positional arguments would probably be serialized to a list. The case of one positional argument would not be serialized to a list. We don't want users to get in the habit of treating multiple positional arguments as an implicit list.

I was also thinking about providing the capability to call this without keywords -- just positional -- and map them into the arguments struct automatically using the position and order of arguments in the struct.

It may be better to look at the type to determine the correct behavior. But I'd also be OK with just making multiple positional args an error case and looking at the type later on.

mortbopet: Out of all of the options, i'd probably be in favor of banning the multiple positional arguments, then. This style of function is anyways one of the worst cases of footguns in python, since you virtually get zero help from any linter. Single argument or keyword arguments seems like a fair restriction for the type constructors.

PR #8785, opened 2025-07-26
Update to new builder format.

jpienaar:
Updated post https://github.com/llvm/llvm-project/pull/147168. Basically did

set(CMAKE_CXX_CLANG_TIDY local/clang-tidy -checks=-*,llvm-use-new-mlir-op-builder -fix)

and then fixed O(10) cases where temporary OpBuilders were used as create requires it passed in by reference now.


uenoku (approved):
LGTM, thank you for taking care of this migration!

seldridge:
Very nice. Thank you for doing this!

PR #8787, opened 2025-07-26
Create .git-blame-ignore-revs

jpienaar:
Enables flagging skipping over more mechanical refactoring.

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view


uenoku (approved):
Thanks!

PR #8788, opened 2025-07-27
[HW][HWAggregateToComb] Add support for hw.array_inject operation in HWAggregateToComb pass

uenoku:
This patch implements lowering of hw.array_inject operations to combinational logic in the HWAggregateToComb transformation pass.

The implementation creates a 2D array where each row represents the result of injecting the new value at a specific index. A multiplexer then selects the appropriate row based on the injection index.

### DIFF
--- a/integration_test/circt-synth/array-lec.mlir
+++ b/integration_test/circt-synth/array-lec.mlir
@@ -5,7 +5,7 @@
 
 // RUN: circt-lec %t.mlir %s -c1=array -c2=array --shared-libs=%libz3 | FileCheck %s --check-prefix=COMB_ARRAY
 // COMB_ARRAY: c1 == c2
-hw.module @array(in %arg0: i2, in %arg1: i2, in %arg2: i2, in %arg3: i2, in %sel1: i2, in %sel2: i2, out out1: i2, out out2: i2) {
+hw.module @array(in %arg0: i2, in %arg1: i2, in %arg2: i2, in %arg3: i2, in %sel1: i2, in %sel2: i2, in %sel3: i2, in %val: i2, out out1: i2, out out2: i2) {
   %0 = hw.array_create %arg0, %arg1, %arg2, %arg3 : i2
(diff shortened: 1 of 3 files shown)


fzi-hielscher:
Sorry, but after being so pedantic in #8774 I can't let the exact same thing go uncommented here. 😅 

Based on the HWToSMT semantics an OOB array inject should  deterministically return its input array unchanged.
So the following modules should be equivalent:

```MLIR
hw.module @inject_oob(out o : i4) {
  %cstneg1 = hw.constant -1 : i4
  %oobIdx = hw.constant 3 : i2
  %ibIdx = hw.constant 2 : i2
  %0 = hw.aggregate_constant [3 : i4, 2 : i4, 1 : i4] : !hw.array<3xi4>
  %1 = hw.array_inject %0[%oobIdx], %cstneg1 : !hw.array<3xi4>, i2
  %2 = hw.array_get %1[%ibIdx] : !hw.array<3xi4>, i2
  hw.output %2 : i4
}

hw.module @inject_oob_unchanged(out o : i4) {
  %cstneg1 = hw.constant -1 : i4
  %ibIdx = hw.constant 2 : i2
  %0 = hw.aggregate_constant [3 : i4, 2 : i4, 1 : i4] : !hw.array<3xi4>
  %2 = hw.array_get %0[%ibIdx] : !hw.array<3xi4>, i2
  hw.output %2 : i4
}

```

But running them through `circt-opt -hw-aggregate-to-comb --canonicalize` produces:

```MLIR
hw.module @inject_oob(out o : i4) {
  %c-1_i4 = hw.constant -1 : i4
  hw.output %c-1_i4 : i4
}
hw.module @inject_oob_unchanged(out o : i4) {
  %c3_i4 = hw.constant 3 : i4
  hw.output %c3_i4 : i4
}
```

I think we can take this as a hint that we should relax the SMT model of `hw.array_inject`.

uenoku:
That's great point, thank you for creating this example! 

> I think we can take this as a hint that we should relax the SMT model of hw.array_inject.

I agree :+1:.

PR #8789, opened 2025-07-28
[ESI] Yield gil before blocking operations in MessageDataFuture

mortbopet:
In case an ESI runtime backend uses some form of multi-threading, and a user has registered a Python callback with an ESI port, deadlock may occur when calling `MessageDataFuture` functions, due to calling a blocking operation while holding the GIL.

Instead, release the GIL while calling the blocking operations.


teqdruid (approved):
IIRC I had some problems with callbacks deadlocking some time ago... This may have been the cause.

PR #8790, opened 2025-07-28
[RTG][InlineSequences] Add option to remove sequence declarations after inlining

maerhart:
In many cases we don't need the sequence declarations anymore after they are inlined. Removing a potentially very long list of sequence declarations can speed up all passes after this significantly.

### DIFF
--- a/lib/Dialect/RTG/Transforms/InlineSequencesPass.cpp
+++ b/lib/Dialect/RTG/Transforms/InlineSequencesPass.cpp
@@ -241,4 +241,9 @@ void InlineSequencesPass::runOnOperation() {
 
   numSequencesInlined = inliner.numSequencesInlined;
   numSequencesInterleaved = inliner.numSequencesInterleaved;
+
+  if (removeSequenceDecls) {
+    for (auto op : llvm::make_early_inc_range(moduleOp.getOps<SequenceOp>()))
+      op.erase();
+  }
 }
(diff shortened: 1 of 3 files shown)


uenoku (approved):
LGTM, I personally prefer making rtg.sequence symbol private, and run SymbolDCE after   InlineSequencePass to ensure not to remove sequence symbols captured in somewhere else but I guess that won't happen practically so it's up to you :+1:

lib/Dialect/RTG/Transforms/InlineSequencesPass.cpp
```diff
@@ -241,4 +241,9 @@ void InlineSequencesPass::runOnOperation() {
 
   numSequencesInlined = inliner.numSequencesInlined;
   numSequencesInterleaved = inliner.numSequencesInterleaved;
+
+  if (removeSequenceDecls) {
+    for (auto op : llvm::make_early_inc_range(moduleOp.getOps<SequenceOp>()))
+      op.erase();
+  }
```
fabianschuiki, line 248: Does the pass guarantee that all sequences are fully inlined at this point?

maerhart: That's the idea. If it doesn't that'd be a bug.

maerhart:
Closing because I took the SymbolDCE route @uenoku suggested.

PR #8791, opened 2025-07-28
[RTG] Don't hardcode the root op of passes where not necessary

maerhart:


lib/Dialect/RTG/Transforms/LinearScanRegisterAllocationPass.cpp
```diff
@@ -62,14 +62,15 @@ static void expireOldInterval(SmallVector<RegisterLiveRange *> &active,
...
-  LLVM_DEBUG(llvm::dbgs() << "=== Processing test @" << testOp.getSymName()
+  LLVM_DEBUG(llvm::dbgs() << "=== Processing "
+                          << OpWithFlags(getOperation(),
+                                         OpPrintingFlags().skipRegions())
                           << "\n\n");
 
   DenseMap<Operation *, unsigned> opIndices;
   unsigned maxIdx;
-  for (auto [i, op] : llvm::enumerate(*testOp.getBody())) {
+  for (auto [i, op] :
```
uenoku, line 78: nit: Could you assert the op has singe region and block?

PR #8793, opened 2025-07-28
[FIRRTL] Type check in Cat operatnds in FIR parser

uenoku:
Parser crashes when cat is constructed with invalid type operands.


uenoku:
> Wondering about locations of operands but that's not touched here but since has error path would be good to ensure error is clear if we have mega cat. Not blocking, just curiosity.

That's good point. I added a note to operand location in addition to the start of the expression.

PR #8794, opened 2025-07-28
[HWToSMT] Return an unbound value for OOB hw.array_inject

uenoku:
This is a follow up to 186dcc8b0e9cd5b47f82f389770089dd040c0bdf.

h/t @fzi-hielscher for pointing out the issue


fzi-hielscher (approved):
Thanks, @uenoku, LGTM.

The `verif.refines` op should hopefully be able to work with this to have a logical equivalence-like check that also works on OOB cases. But Z3 sometimes refuses to operate on quantified array sorts. So I'll have to tweak the LEC tool first to make sure it properly handles `unknown` results.

maerhart:
It'd probably be a good thing to mention this new OOB behavior in the ODS description field of the op.

PR #8795, opened 2025-07-29
[RTG] Add InsertTestToFileMappingPass

maerhart:
Add a pass to group tests to output files. Currently, this matches what the emission pass does but can be extended to group tests according to certain requirements demanded from the execution environment. Also add a simple pass that inlines the tests without any gluecode (matching what the emission pass currently does). The emission pass will be simplified in a future PR to only iterate over the file operations and print what's inside


fabianschuiki (approved):
LGTM! Pretty cool :slightly_smiling_face:

PR #8800, opened 2025-07-30
[FIRRTL] Add layer colour verifiers to force and release ops

rwy7:
This PR adds layer verifiers to the probe force and release ops. During this work I notice that lower-layers does not remove colours from all ops (caught by the new verifiers), so I have fixed that here as well.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLOps.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOps.cpp
@@ -473,17 +473,15 @@ static bool isLayerSetCompatibleWith(const LayerSet &src, const LayerSet &dst,
   return missing.empty();
 }
 
-template <typename F>
-static LogicalResult checkLayerCompatibility(const LayerSet &src,
-                                             const LayerSet &dst, F &&error) {
+static LogicalResult checkLayerCompatibility(
+    Operation *op, const LayerSet &src, const LayerSet &dst,
+    const Twine &errorMsg,
+    const Twine &noteMsg = Twine("missing layer requirements")) {
   SmallVector<SymbolRefAttr> missing;
   if (isLayerSetCompatibleWith(src, dst, missing))
     return success();
-
-  auto diag = error();
-  auto &note = diag.attachNote();
-  note << "missing layer requirements: ";
-  interleaveComma(missing, note);
+  interleaveComma(missing, op->emitOpError(errorMsg).attachNote()
+                               << noteMsg << ": ");
   return failure();
 }
 
@@ -1672,16 +1670,8 @@ LogicalResult FExtModuleOp::verify() {
         referenced.insert(layer);
   }
 
-  SmallVector<SymbolRefAttr> missing;
-  if (!isLayerSetCompatibleWith(referenced, known, missing)) {
-    auto diag = emitOpError("references unknown layers");
-    auto &note = diag.attachNote();
-    note << "unknown layers: ";
-    interleaveComma(missing, note);
-    return failure();
-  }
-
-  return success();
+  return checkLayerCompatibility(getOperation(), referenced, known,
+                                 "references unknown layers", "unknown layers");
 }
 
 LogicalResult FIntModuleOp::verify() {
@@ -3909,15 +3899,10 @@ LogicalResult RefDefineOp::verify() {
   auto ambientLayers = getAmbientLayersAt(getOperation());
   auto dstLayers = getLayersFor(getDest());
   SmallVector<SymbolRefAttr> missingLayers;
-  if (!isLayerSetCompatibleWith(ambientLayers, dstLayers, missingLayers)) {
-    auto diag = emitOpError("has more layer requirements than destination");
-    auto &note = diag.attachNote();
-    note << "additional layers required: ";
-    interleaveComma(missingLayers, note);
-    return failure();
-  }
 
-  return success();
+  return checkLayerCompatibility(getOperation(), ambientLayers, dstLayers,
+                                 "has more layer requirements than destination",
+                                 "additional layers required");
 }
 
 LogicalResult PropAssignOp::verify() {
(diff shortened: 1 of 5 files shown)


test/Dialect/FIRRTL/layers-errors.mlir
```diff
@@ -169,3 +169,106 @@ firrtl.circuit "Top" {
...
+    firrtl.layerblock @A {
+      %wire = firrtl.wire sym @wire : !firrtl.uint<1>
+    }
+    %0 = firrtl.xmr.ref @xmr : !firrtl.probe<uint<1>>
+  }
+}
+
+// -----
+
+// Driving an outer property from inside a layerblock is not allowed.
```
dtzSiFive: I think this test is duplicated?

test/Dialect/FIRRTL/lower-layers.mlir
seldridge, line 1: An explicit check of the instance case is missing.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6346,6 +6346,67 @@ LogicalResult RWProbeOp::verifyInnerRefs(hw::InnerRefNamespace &ns) {
...
+    auto diag =
+        emitOpError("ambient layers are insufficient to release reference");
+    auto &note = diag.attachNote();
+    note << "missing layer requirements: ";
+    interleaveComma(missingLayers, note);
+    return failure();
+  }
+  return success();
+}
+
```
seldridge: Is there a way to reuse this verification code as opposed to duplicating it? The only difference appears to be in the use of `"release"` or `"force"` in the message.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6346,6 +6346,67 @@ LogicalResult RWProbeOp::verifyInnerRefs(hw::InnerRefNamespace &ns) {
   return checkFinalType(symOp.getTargetResult().getType(), symOp.getLoc());
 }
 
+LogicalResult RefForceOp::verify() {
+  auto ambientLayers = getAmbientLayersAt(getOperation());
+  auto destLayers = getLayersFor(getDest());
+  SmallVector<SymbolRefAttr> missingLayers;
+  if (!isLayerSetCompatibleWith(destLayers, ambientLayers, missingLayers)) {
+    auto diag =
+        emitOpError("ambient layers are insufficient to force reference");
```
seldridge: Consider:

```
has insufficient ambient layers to force its reference
```

test/Dialect/FIRRTL/layers-errors.mlir
```diff
@@ -169,3 +169,106 @@ firrtl.circuit "Top" {
...
+firrtl.circuit "Top" {
+  firrtl.layer @A bind {}
+  hw.hierpath @xmr [@Top::@wire]
+  firrtl.module @Top(out %o : !firrtl.probe<uint<1>>) {
+    firrtl.layerblock @A {
+      %wire = firrtl.wire sym @wire : !firrtl.uint<1>
+    }
+    %0 = firrtl.xmr.ref @xmr : !firrtl.probe<uint<1>>
+  }
+}
```
seldridge: If this is checking for something that _should_ be legal, then it should be put in a test that isn't a `-error` test.

test/Dialect/FIRRTL/layers-errors.mlir
```diff
@@ -169,3 +169,106 @@ firrtl.circuit "Top" {
...
+// Release-initial-ing a colored probe is not allowed (in an uncolored context).
+firrtl.circuit "Top" {
+  firrtl.layer @A bind {}
+  firrtl.module @Top(in %e : !firrtl.uint<1>) {
+    %p = firrtl.wire : !firrtl.rwprobe<uint<1>, @A>
+    firrtl.layerblock @A {
+      %w, %wp = firrtl.wire forceable : !firrtl.uint<1>, !firrtl.rwprobe<uint<1>>
+      %q = firrtl.ref.cast %wp : (!firrtl.rwprobe<uint<1>>) -> !firrtl.rwprobe<uint<1>, @A>
+      firrtl.ref.define %p, %q : !firrtl.rwprobe<uint<1>, @A>
+    }
```
seldridge, line 237: This section is duplicated for each test. I think the PR could share this in a circuit which had the four illegal force/release statements in the module's body. I think the verifier will report all four errors then.

seldridge (approved):
Nits only.

Nice that this is getting tightened up.

One meta point: the fact that so many of the error messages refer to "ambient" layers could be viewed as problematic if these error messages are exposed directly to Chisel users / people writing or generating FIRRTL. This is because "ambient" is never explained anywhere.

I'd move in either one of two directions with this: (1) extend the FIRRTL spec to add this concept or (2) make all the error messages closer to the language in the FIRRTL spec. (2) is much more verbose. E.g., if I was banned from using "ambient", I'd have to write these error messages like:

```
'firrtl.ref.force' op is in a region with fewer layers than its destination
'firrtl.ref.force' op must exist in a region with layer colorings 'A, B, C' to write to its destination, but only exists in a region with layer colorings `A, B`
```

While I would expect that "ambient" is semi-intuitive, this is really a compiler-internal thing at this point.

If unclear, this commentary is all for follow-up work.

test/Dialect/FIRRTL/layers-errors.mlir
rwy7: The verifier bails on the first error.

test/Dialect/FIRRTL/layers-errors.mlir
rwy7: Sorry--this wasn't meant to be checked in. Removed!

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: Took your suggestion here.

test/Dialect/FIRRTL/lower-layers.mlir
rwy7: Sorry, not sure what this means! :bow:

test/Dialect/FIRRTL/layers-errors.mlir
rwy7: Removed, thanks!

test/Dialect/FIRRTL/lower-layers.mlir
seldridge: This commit adds/enshrines two checks: (1) all layer colors are removed from ops and (2) all layer colors are removed from instances.

This is lightly checked by updating an existing test that erroneously included the `A` layer. I didn't see a new check that layer colors are removed from instances. I didn't expand the test to check if this was already being checked somewhere else.

test/Dialect/FIRRTL/layers-errors.mlir
seldridge: Got it. No problem with this as-is then.

seldridge (approved):

test/Dialect/FIRRTL/lower-layers.mlir
rwy7: Oh OK, yes this is already being checked!
https://github.com/llvm/circt/blob/main/test/Dialect/FIRRTL/lower-layers.mlir#L26-L30

PR #8801, opened 2025-07-30
[ESI] Always build zlib statically into ESI runtime

mortbopet:


teqdruid (approved):
This  _may_ introduce linking issues if the client code also uses zlib, but it could also help to resolve them. But anything to make common case linking easier.

PR #8803, opened 2025-07-30
[ESI][Runtime] Fix wheel build

teqdruid:


teqdruid:
Proof build: https://github.com/llvm/circt/actions/runs/16633017477

PR #8808, opened 2025-07-31
Add MacroErrorOp

rwy7:
Add an operation which intentionally emits as a reference to an undefined SV macro. We can use this op to crash a verilog compilation process if the design is misconfigured.


include/circt/Dialect/SV/SVStatements.td
```diff
@@ -854,6 +854,23 @@ def ForOp : SVOp<"for", [SingleBlock, NoTerminator, ProceduralOp,
...
+    reference to an undefined macro definition.
+
+    We reserve the macro prefix _ERROR for emitting static assertions.
+  }];
+  let arguments = (ins OptionalAttr<StrAttr>:$message);
+  let assemblyFormat = "($message^)? attr-dict";
+  let extraClassDeclaration = [{
+    /// Build the bogus macro identifier for this error op. Do not define this
+    /// macro identifier. The identifier will be prefixed by `_ERROR`.
+    SmallString<32> getMacroIdentifier();
```
seldridge: `SmallString<32>` (or any `SmallString`) is usually not a good return type. How about `std::string`? Or: taking a parameter of `&SmallVectorImpl<char>`?

lib/Dialect/SV/SVOps.cpp
```diff
@@ -172,6 +173,37 @@ MacroRefExprSEOp::getReferencedMacro(const hw::HWSymbolCache *cache) {
   return ::getReferencedMacro(cache, *this, getMacroNameAttr());
 }
 
+//===----------------------------------------------------------------------===//
+// MacroErrorOp
+//===----------------------------------------------------------------------===//
+
+SmallString<32> MacroErrorOp::getMacroIdentifier() {
+  SmallString<32> id;
+  id.append("_ERROR");
```
seldridge: Can save a line with initialization, I think:


```suggestion
  SmallString<32> id("_ERROR");
```

lib/Dialect/SV/SVOps.cpp
```diff
@@ -172,6 +173,37 @@ MacroRefExprSEOp::getReferencedMacro(const hw::HWSymbolCache *cache) {
...
+
+  auto msg = getMessage();
+  if (!msg)
+    return id;
+
+  bool split = true;
+  for (auto c : *msg) {
+    if (isalnum(c)) {
```
seldridge: The PR can use `llvm::isAlnum` to avoid the dependency on `ctype.h`.

lib/Dialect/SV/SVOps.cpp
```diff
@@ -172,6 +173,37 @@ MacroRefExprSEOp::getReferencedMacro(const hw::HWSymbolCache *cache) {
...
+      if (split)
+        id.push_back('_');
+      id.push_back(c);
+      split = false;
+    } else {
+      split = true;
+    }
+  }
```
seldridge, line 192: This would be simpler with replacing `llvm::isSpace` with `_` and dropping anything that isn't `llvm::isAlnum`.

I grok the "collapse strings of non-alnum to `_` except the leading non-alnum". However, this may be slightly too clever.

Another direction is to make this verified as being `isSpace` or `isAlnum` in a verifier for the op for additional predictability.

seldridge (approved):
LGTM

Consider simplifying the algorithm for the message mangling. Additionally, returning `SmallString` is a bit non-standard.

test/Conversion/ExportVerilog/sv-dialect.mlir
```diff
@@ -1978,6 +1978,15 @@ sv.bind #hw.innerNameRef<@InlineBind::@foo2>
 // CHECK-NEXT:    .out (wire_0)
 // CHECK-NEXT:  );
 
+sv.macro.error
```
uenoku: Could you include a test to check sv.macro.error being emitted in hw.module?

rwy7: Done, thanks!

lib/Dialect/SV/SVOps.cpp
rwy7: Simplified to replace non-alnum with underscore.

seldridge: LGTM.

include/circt/Dialect/SV/SVStatements.td
rwy7: Changed to string.

PR #8809, opened 2025-08-01
[utils] Add Arcilator and Verilator comparison utility

fabianschuiki:
Add the `compare-arcs-vtor.py` utility script which takes a core dialect design, compiles it with Arcilator and Verilator to obtain two simulators, and then runs the simulators in lockstep on random inputs to verify that they produce the same output.

This can be useful in combination with circt-reduce in order to isolate behavior mismatches between Arcilator and Verilator.

### DIFF
--- a/utils/compare-arcs-vtor.py
+++ b/utils/compare-arcs-vtor.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+import argparse
+import json
+import shlex
+import shutil
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+"""
+A utility that takes a piece of hardware described in the core dialects and
+runs it through arcilator and verilator in lockstep with randomized inputs,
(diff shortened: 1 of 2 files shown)


utils/compare-arcs-vtor.py
```diff
@@ -0,0 +1,167 @@
...
+    run(args.input, args.dir)
+  else:
+    with tempfile.TemporaryDirectory() as dir:
+      run(args.input, Path(dir))
+
+
+def run(input: Path, dir: Path):
+  # Figure out where the arcilator binary is.
+  arcilator = Path(
+      subprocess.check_output(["which", "arcilator"],
```
uenoku: nit: shutil.which? 

Also for consistency it would be good to check verilator/firtool/llc

fabianschuiki: Very cool, I didn't know about `shutil.which`. Going to use that :+1:

PR #8811, opened 2025-08-01
Bump LLVM to b63a9b7a3cdc1e41545df595215077e50bfd04af.

mikeurbach:
The one upstream change we had to account for was
https://github.com/llvm/llvm-project/commit/b6a98b9, where we started creating splits that don't have null terminated buffers, which are assumed by our tool and the SourceMgr. Make a copy to ensure null-termination, which is apparently what happened anyway before this change.


uenoku (approved):
LGTM, glad we don't have to update any test.

PR #8818, opened 2025-08-04
[Synth] Add cut rewrite framework (priority cuts enumeration)

uenoku:
This is the first commit introducing a cut-based rewriting framework for combinational logic optimization. Cut-based rewriting is a fundamental technique in logic synthesis which are used for not only technology mapping/LUT mappping but also logic optimization by identifying and replacing subgraphs (cuts) in combinational circuits with functionally equivalent but more optimal implementations.

This initial implementation focuses specifically on priority cuts enumeration as the foundation for future cut-based optimization passes. The framework includes a Cut data structure for representing logical cuts with truth tables, a CutEnumerator implementing the priority cuts algorithm with configurable limits, cut enumeration with size constraints and deduplication, truth table computation and NPN canonical form matching, support for AIG operations.

The implementation is based on ABC's priority cuts paper https://people.eecs.berkeley.edu/~alanmi/publications/2007/iccad07_map.pdf 

The priority cuts algorithm maintains a bounded set of promising cuts per node to avoid exponential explosion while preserving optimization quality. This foundation enables future development of technology mapping, LUT mapping, and other cut-based combinational optimization passes.


test/Dialect/Synth/priority-cuts.mlir
```diff
@@ -0,0 +1,97 @@
...
+//           @t128 = binary 10000000 = 3-input function
+//
+// - d<depth>: Logic depth from primary inputs to this cut's output
+//   Example: d0 = primary input (no logic), d1 = one logic level, etc.
+//
+// Example: {a0, a1}@t8d1 means:
+// - Cut has inputs a0 and a1
+// - Implements AND function (truth table 8 = binary 1000)
+// - Has logic depth 1 (one level from primary inputs)
+//===----------------------------------------------------------------------===//
```
fabianschuiki: This is very helpful. Might be useful to comment on which of `a0` or `a1` forms the LSB and MSB of the index into the truth table.

test/Dialect/Synth/priority-cuts.mlir
```diff
@@ -0,0 +1,97 @@
...
+
+
+// CHECK-LABEL: Enumerating cuts for module: trivial
+hw.module @trivial(in %a : i1, out result : i1) {
+    // CHECK: a 1 cuts: {a}@t2d0
+    // CHECK-NEXT: out0 2 cuts: {out0}@t2d0 {a}@t2d1
```
fabianschuiki, line 34: Very cool that this already gives you the minimal truth table 😍

include/circt/Dialect/Synth/Transforms/SynthPasses.h
```diff
@@ -0,0 +1,40 @@
...
+
+#define GEN_PASS_DECL
+#include "circt/Dialect/Synth/Transforms/SynthPasses.h.inc"
+/// Generate the code for registering passes.
+#define GEN_PASS_REGISTRATION
+#include "circt/Dialect/Synth/Transforms/SynthPasses.h.inc"
+
+} // namespace synth
+
+} // namespace circt
```
fabianschuiki:
```suggestion
} // namespace synth
} // namespace circt
```

include/circt/Dialect/Synth/Transforms/SynthPasses.h
```diff
@@ -0,0 +1,40 @@
...
+enum OptimizationStrategy {
+  OptimizationStrategyArea,  ///< Optimize for minimal area
+  OptimizationStrategyTiming ///< Optimize for minimal critical path delay
+};
+
+#define GEN_PASS_DECL
+#include "circt/Dialect/Synth/Transforms/SynthPasses.h.inc"
+/// Generate the code for registering passes.
+#define GEN_PASS_REGISTRATION
+#include "circt/Dialect/Synth/Transforms/SynthPasses.h.inc"
```
fabianschuiki: Could this work with a single include?
```suggestion
#define GEN_PASS_DECL
#define GEN_PASS_REGISTRATION
#include "circt/Dialect/Synth/Transforms/SynthPasses.h.inc"
```

include/circt/Dialect/Synth/Transforms/SynthPasses.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef CIRCT_DIALECT_SYNTH_TRANSFORMS_SYNTHPASSES_H
+#define CIRCT_DIALECT_SYNTH_TRANSFORMS_SYNTHPASSES_H
+
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Pass/Pass.h"
+#include <limits>
+
+namespace circt {
+namespace synth {
+/// Optimization strategy.
```
fabianschuiki:
```suggestion
namespace synth {

/// Optimization strategy.
```

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -0,0 +1,267 @@
...
+#include "llvm/Support/raw_ostream.h"
+#include <memory>
+#include <optional>
+
+namespace circt {
+namespace synth {
+
+// Import NPN classes from the support library
+using circt::BinaryTruthTable;
+using circt::NPNClass;
```
fabianschuiki: Shouldn't these automatically be visible because the `synth` namespace is nested in `circt`?

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -0,0 +1,267 @@
...
+
+  /// Check if this cut represents a trivial cut.
+  /// A trivial cut has no internal operations and exactly one input.
+  bool isTrivialCut() const;
+
+  /// Get the root operation of this cut.
+  unsigned getDepth() const { return depth; }
+
+  /// Default constructor creates an empty cut.
+  Cut() = default;
```
fabianschuiki: Since there are no other constructors, this is probably not needed.

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -0,0 +1,267 @@
...
+/// Represents a cut that has been successfully matched to a rewriting pattern.
+///
+/// This class encapsulates the result of matching a cut against a rewriting
+/// pattern during optimization. It stores the matched pattern, the
+/// cut that was matched, and timing information needed for optimization.
+/// TODO: Currently only added as a placeholder.
+class MatchedPattern {
+public:
+  /// Default constructor creates an invalid matched pattern.
+  MatchedPattern() = default;
```
fabianschuiki: I think you can just leave this out and get the same behavior for the placeholder.
```suggestion
```

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -0,0 +1,267 @@
...
+public:
+  /// Finalize the cut set by removing duplicates and selecting the best
+  /// pattern.
+  ///
+  /// This method:
+  /// 1. Removes duplicate cuts based on inputs and root operation
+  /// 2. Limits the number of cuts to prevent exponential growth
+  /// 3. Matches each cut against available patterns
+  /// 4. Selects the best pattern based on the optimization strategy
+  void finalize(const CutRewriterOptions &options);
```
fabianschuiki, line 155: How does this function have access to the available patterns? Is this going to be a separate argument to the function later, or do the cuts or cut sets have some reference to a common object that holds match patterns?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+//
+// This file implements a DAG-based boolean matching cut rewriting algorithm for
+// applications like technology/LUT mapping and combinational logic
+// optimization. The algorithm uses priority cuts and NPN
+// (Negation-Permutation-Negation) canonical forms to efficiently match cuts
+// against rewriting patterns.
+//
+// References:
+//  "Combinational and Sequential Mapping with Priority Cuts", Alan Mishchenko,
+//  Sungmin Cho, Satrajit Chatterjee and Robert Brayton, ICCAD 2007
```
fabianschuiki, line 17: Thanks for citing the academic work. That'll make following along easier 🙌

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+    }
+    // Evaluate the and inverter
+    eval[andOp.getResult()] = andOp.evaluate(inputs);
+    return;
+  }
+
+  assert(false &&
+         "Unsupported operation for simulation. isSupportedLogicOp should "
+         "be used to check if the operation can be simulated.");
+}
```
fabianschuiki: Would it make sense to either make all of these `assert`, or all `llvm::report_fatal_error`? At least the fatal error and the final assert feel like they should both be the same, because they effectively report a similar kind of input argument error.

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+//  Return true if the value is always a cut input.
```
fabianschuiki:
```suggestion
// Return true if the value is always a cut input.
```

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+  return !isSupportedLogicOp(op);
+}
+
+LogicalResult
+circt::synth::topologicallySortLogicNetwork(mlir::Operation *topOp) {
+
+  // Sort the operations topologically
+  if (topOp
+          ->walk([&](Region *region) {
```
fabianschuiki: Maybe break this into two separate lines to avoid the wild if breaking?
```
auto result = topOp->walk(...);
if (result.wasInterrupted())
  return mlir::emitError(...);
```

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+            // Graph region.
+            for (auto &block : *region) {
+              if (!mlir::sortTopologically(
+                      &block, [&](Value value, Operation *op) -> bool {
+                        // Topologically sort simulatable ops and purely
+                        // dataflow ops. Other operations can be scheduled.
+                        return !(isSupportedLogicOp(op) ||
+                                 isa<comb::ExtractOp, comb::ReplicateOp,
+                                     comb::ConcatOp>(op));
+                      }))
```
fabianschuiki: Very nice use of the MLIR topo utils! 💯

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// Cut
+//===----------------------------------------------------------------------===//
+
+bool Cut::isTrivialCut() const {
+  // A cut is a primary input if it has no operations and only one input
+  return operations.empty() && inputs.size() == 1;
+}
```
fabianschuiki: This feels like it could live in the header, which already contains a few simple operations defined inline. Or maybe move all of them here?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+  auto canonicalForm = NPNClass::computeNPNCanonicalForm(truthTable);
+
+  npnClass.emplace(std::move(canonicalForm));
+  return *npnClass;
+}
+
+void Cut::getPermutatedInputs(const NPNClass &patternNPN,
+                              SmallVectorImpl<Value> &permuatedInputs) const {
+  auto npnClass = getNPNClass();
+  SmallVector<Value> permutedInputs;
```
fabianschuiki:
```suggestion
```

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+  // Simulate the IR.
+  uint32_t tableSize = 1 << numInputs;
+  DenseMap<Value, APInt> eval;
+  for (uint32_t i = 0; i < numInputs; ++i) {
+    APInt value(tableSize, 0);
+    for (uint32_t j = 0; j < tableSize; ++j) {
+      // Make sure the order of the bits is correct.
+      value.setBitVal(j, (j >> i) & 1);
+    }
```
fabianschuiki: If I understand this correctly, this generates the `10101010` pattern for input 0, `11001100` for input 1, `11110000` for input 2, etc. You might be able to use `getHighBitsSet` and `getSplat` to avoid having to do this in `O(N*(2^N))`: https://github.com/llvm/circt/blob/main/lib/Dialect/LLHD/Transforms/DeseqUtils.h#L115-L121

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+    // Set the input value for the truth table
+    eval[inputs[i]] = std::move(value);
+  }
+
+  // Simulate the operations in the cut
+  for (auto *op : operations)
+    simulateLogicOp(op, eval);
```
fabianschuiki, line 236: Very cool idea to generate these input truth table vectors and then flow them through the ops, essentially vectorizing the circuit evaluation using bitwise logic ops. 🥳

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+      if (isInput) {
+        if (!other.operations.contains(defOp)) // op is in the other cut.
+          continue;
+      }
+      if (isOtherInput) {
+        if (!operations.contains(defOp)) // op is in this cut.
+          continue;
+      }
+
+      populateOperations(defOp);
```
fabianschuiki: Could compact this slightly:
```suggestion
      // Otherwise, check if the operation is in the other cut.
      auto *defOp = value.getDefiningOp();
      assert(defOp && "Value must have a defining operation since block"
                      "arguments are treated as inputs");
      if (isInput)
        if (!other.operations.contains(defOp)) // op is in the other cut.
          continue;
      if (isOtherInput)
        if (!operations.contains(defOp)) // op is in this cut.
          continue;
      populateOperations(defOp);
```

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+// different root operation.
+Cut Cut::reRoot(Operation *root) const {
+  Cut newCut;
+  newCut.inputs = inputs;
+  newCut.operations = operations;
+  // Add the new root operation to the cut
+  newCut.operations.insert(root);
+  newCut.depth = depth + 1; // Increment depth for new root
+  return newCut;
+}
```
fabianschuiki: Does this require that the old root op is an operand of the new root op? The `depth + 1` suggests that the added op must strictly extend the depth, which suggests that it must depend on the old root somehow. Should this possibly be some kind of assert?
```c++
assert(llvm::is_contained(root.getOperands(), operations.back()->getResult(0)));
```
Not sure how relevant this is.

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+    // Support only binary operations with a single bit
+    // TODO: This can be extended to support variadic multiplication.
+    if (op->getNumOperands() > 2)
+      return op->emitError() << "Cut enumeration supports at most 2 operands, "
+                             << "found: " << op->getNumOperands();
+    if (!op->getResult(0).getType().isInteger(1))
+      return op->emitError()
+             << "Supported logic operations must have a single bit "
+                "result type but found: "
+             << op->getResult(0).getType();
```
fabianschuiki: It looks like these checks are performed in `visitLogicOp` again. Might be worth removing these ones here in favor of `visitLogicOp`?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -0,0 +1,564 @@
...
+  auto *resultCutSet = createNewCutSet(result);
+
+  // Add the singleton cut first
+  resultCutSet->addCut(primaryInputCut);
+
+  // Schedule cut set finalization when exiting this scope
+  auto prune = llvm::make_scope_exit([&]() {
+    // Finalize cut set: remove duplicates, limit size, and match patterns
+    resultCutSet->finalize(options);
+  });
```
fabianschuiki, line 501: Clever!

include/circt/Dialect/Synth/Transforms/CutRewriter.h
uenoku: Yes, good question and sorry that it was not clear from comments that I changed the signature to split the PR. 
In the follow up it will be changed to take a lambda which will be passed from `CutRewriter` class that holds all the available patterns.  
 
https://github.com/uenoku/circt/blob/4e6d3e0b80745b221b3da3a9893a1303386ad198/lib/Synthesis/CutRewriter.cpp#L421
https://github.com/uenoku/circt/blob/4e6d3e0b80745b221b3da3a9893a1303386ad198/lib/Synthesis/CutRewriter.cpp#L590-L592

lib/Dialect/Synth/Transforms/CutRewriter.cpp
uenoku: Oh, no. Thank you for catching this.

lib/Dialect/Synth/Transforms/CutRewriter.cpp
uenoku: Yeah I moved everything to cpp file for the consistency.

lib/Dialect/Synth/Transforms/CutRewriter.cpp
uenoku: I didn't know the function, very clever!

lib/Dialect/Synth/Transforms/CutRewriter.cpp
uenoku: Yeah I added an assertion, thank you for suggestion.

cowardsa:
Sorry I couldn't get to this quickly enough - but looks really interesting - did have two questions though:

1. The ABC paper largely talks about FPGA mapping where the target is K-LUTs - do you know how effective the technique is for standard cell libraries?
2. Do you have a benchmark suite (potentially internal) to evaluate "progress" on in terms of runtime vs logic depth vs number of cells as would be helpful to understand where gains are coming from?

uenoku:
Thank you for taking a look! Sorry I was a bit rush to merge :) 

1. https://dl.acm.org/doi/10.1145/2593069.2593158 has comparison for MIG/AIG-based mapping with commercial tools for standard cell mapping. MIG has better results in the paper but AIG(ABC) is on par with commercial tool. ABC uses priority-cut + NPN + supergate based mapping. Though for MIG experiments setup has short description ("traditional optimization mapping") and implementation is not disclosed as far as I see but I believe they are doing the same (considered that mockturtle's mapper works in the same way https://mockturtle.readthedocs.io/en/latest/algorithms/mapper.html). 
2. No at this moment but I'm planning to add https://github.com/lsils/benchmarks to https://github.com/circt/circt-tests test-suites and track (combinatorial) mapping results with open source PDK. Also for e2e testing I'm hoping we can use large chisel open source projects as benchmark (rocket, boom, XiangShan, t1 etc). I eventually want to track results on internal designs but it's difficult to disclose the design so first I'd like to focus on setting up benchmark on open source projects.

Issue #8693, opened 2025-07-11
llhd mem2reg failure


jpienaar:
For input (derived from SV import minized and then MLIR minized)

```mlir
module {
  hw.module @a() {
    %false = hw.constant false
    %b = llhd.sig %false : i1
    llhd.combinational {
      cf.br ^bb2
    ^bb1:  // no predecessors
      %0 = llhd.prb %b : !hw.inout<i1>
      %1 = llhd.constant_time <0ns, 0d, 1e>
      %g = llhd.sig %false : i1
      llhd.drv %g, %0 after %1 : !hw.inout<i1>
      cf.br ^bb2
    ^bb2:  // 2 preds: ^bb0, ^bb1
      llhd.yield
    }
    hw.output
  }
}
```

results post llhd-mem2reg in

```
error: operand #0 does not dominate this use
note: see current operation: "llhd.drv"(%6, %7, %9, %8) : (!hw.inout<i1>, i1, !llhd.time, i1) -> ()
/tmp/t2.mlir:10:12: note: operand defined here (op in the same region)
      %g = llhd.sig %false : i1
           ^
```

print IR after:

```mlir
"builtin.module"() ({
  "hw.module"() <{module_type = !hw.modty<>, parameters = [], sym_name = "a"}> ({
    %0 = "hw.constant"() <{value = false}> : () -> i1
    %1 = "llhd.sig"(%0) <{name = "b"}> : (i1) -> !hw.inout<i1>
    %2 = "llhd.prb"(%1) : (!hw.inout<i1>) -> i1
    "llhd.combinational"() ({
      %3 = "hw.constant"() <{value = false}> : () -> i1
      %4 = "hw.constant"() <{value = false}> : () -> i1
      "cf.br"(%3, %4)[^bb2] : (i1, i1) -> ()
    ^bb1:  // no predecessors
      %5 = "hw.constant"() <{value = true}> : () -> i1
      %6 = "llhd.sig"(%0) <{name = "g"}> : (i1) -> !hw.inout<i1>
      "cf.br"(%2, %5)[^bb2] : (i1, i1) -> ()
    ^bb2(%7: i1, %8: i1):  // 2 preds: ^bb0, ^bb1
      %9 = "llhd.constant_time"() <{value = #llhd.time<0ns, 0d, 1e>}> : () -> !llhd.time
      "llhd.drv"(%6, %7, %9, %8) : (!hw.inout<i1>, i1, !llhd.time, i1) -> ()
      // ^
[...]

PR #8821, opened 2025-08-05
[LLHD] Propogate signal definitions only to dominated blocks in Mem2Reg

mvpant:
Currently, after Mem2Reg inserts drives and probes, it does not check whether `llhd.sig` (or the slot) still dominates all of its users. For cases where the operand is `hw.constant`, it should be safe to move the `llhd.sig` into region's entry block to ensure dominance, if it doesn't already. What do you think? 

Error, without the patch:
```
<stdin>:29:16: error: operand #0 does not dominate this use
    %ready_T = llhd.sig %c2_i4 : i4
               ^
<stdin>:29:16: note: see current operation: "llhd.drv"(%18, %9, %11, %10) : (!hw.inout<i4>, i4, !llhd.time, i1) -> ()
<stdin>:29:16: note: operand defined here (op in the same region)
```

### DIFF
--- a/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
+++ b/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
@@ -630,8 +630,10 @@ struct Promoter {
   void constructLattice();
   void propagateBackward();
   void propagateBackward(LatticeNode *node);
-  void propagateForward(bool optimisticMerges);
-  void propagateForward(LatticeNode *node, bool optimisticMerges);
+  void propagateForward();
+  void propagateForward(bool optimisticMerges, DominanceInfo &dominance);
+  void propagateForward(LatticeNode *node, bool optimisticMerges,
+                        DominanceInfo &dominance);
   void markDirty(LatticeNode *node);
 
   void insertProbeBlocks();
@@ -708,14 +710,16 @@ LogicalResult Promoter::promote() {
   LLVM_DEBUG({
     llvm::dbgs() << "After probe insertion:\n";
     lattice.dump();
+    llvm::dbgs() << *region.getParentOp() << "\n";
   });
 
   // Propagate the reaching definitions forward across the lattice.
-  propagateForward(true);
-  propagateForward(false);
+  DominanceInfo dominance(region.getParentOp());
+  propagateForward();
   LLVM_DEBUG({
     llvm::dbgs() << "After forward propagation:\n";
     lattice.dump();
+    llvm::dbgs() << *region.getParentOp() << "\n";
   });
 
   // Resolve definitions.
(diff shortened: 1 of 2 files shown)


fabianschuiki:
Do you have an example input that triggers this issue for you? Mem2Reg creating probes and drives with domination issues is pretty bad. That's definitely a bug. Relocating slots if possible can be a very useful optimization, but Mem2Reg also shouldn't produce invalid IR if we can't relocate a slot.

mvpant:
> Do you have an example input that triggers this issue for you?

Yes, a testcase from PR produces an error:

```
sed -n '1008,1050p' test/Dialect/LLHD/Transforms/mem2reg.mlir | circt-opt --mlir-disable-threading --debug-only="llhd-mem2reg" --llhd-mem2reg --mlir-print-ir-after-failure
```

Or even shorter version:
```
hw.module @Test() {
  %true = hw.constant true
  %false = hw.constant false
  %3 = llhd.constant_time <0ns, 0d, 1e>
  %clock_0 = llhd.sig name "clock" %false : i1
  llhd.process {
    cf.br ^bb1
  ^bb1:
    %5 = llhd.prb %clock_0 : !hw.inout<i1>
    llhd.wait (%5 : i1), ^bb3
  ^bb3:
    %c2_i4 = hw.constant 0 : i4
    %ready_T = llhd.sig %c2_i4 : i4
    llhd.drv %ready_T, %c2_i4 after %3 : !hw.inout<i4>
    %160 = llhd.prb %ready_T : !hw.inout<i4>
    cf.br ^bb1
  }
  hw.output
}
```

#8693 -- looks like a similar issue

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
```diff
@@ -1145,10 +1158,18 @@ void Promoter::propagateForward(LatticeNode *node, bool optimisticMerges) {
...
+        continue;
+      // ...
+      for (auto &defEntry : predecessor->valueBefore->reachingDefs) {
+        auto def = defEntry.getFirst().getPointer();
+        auto *defBlock = def.getDefiningOp()->getBlock();
+        if (!dominance.dominates(defBlock, entry->block))
+          continue;
+        reachingDefs.insert(defEntry);
+      }
+    }
```
mvpant: How about this? Following the suggestion from issue #8693.

fabianschuiki: Nice, I like that approach :raised_hands: Love the small test cases related to dominance, too! :smiley:

fabianschuiki (approved):
LGTM! Thank you very much for fixing this :partying_face:

mvpant:
> LGTM! Thank you very much for fixing this 🥳 (Happy to merge for you if needed)

Sure, thanks!

PR #8824, opened 2025-08-06
[ImportVerilog] Add time literals

fabianschuiki:
Add support for time literal expressions such as `42us` or `1.234ns` to the ImportVerilog conversion.

Time values in SystemVerilog are pretty messy and easy to misuse because they are essentially just unit-less integers or doubles. To make the CIRCT's Verilog frontend a bit more robust, introduce a new `!moore.time` type that represents a time value. The `time` and `realtime` types in SV both map to this IR type.

All time values are converted and rounded to femtoseconds and then stored as an `i64`. A new `moore.constant_time` can be used to materialize a constant time value. This means that time values in the IR are represented with an unambiguous unit.

In contrast, SystemVerilog eagerly converts time values to the underlying integer or double by scaling the value by the `timeunit` specified by the user. This is problematic when time values are passed between modules that have different `timeunit` specifications. Since time values are packed and thus equivalent to an integer, a time value will flow across this module boundary _without_ having its units rescaled. This can lead to a value like `10ns` read as a `100ps` in another module due to a time unit mismatch.

Since the IR does not apply the time unit immediately when a time value is created, future cast operations from and to integers and doubles will have to apply this time unit scaling factor instead. As a result, time values become harder to abuse, and if the user never casts a time to an integer or double, all arithmetic on time values can be done directly on the `time` type without ever interacting with the local `timeunit` specification. We can later add a faithful-but-unsafe mode where we eagerly map time values to integers and doubles.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -492,6 +492,13 @@ def ConstantOp : MooreOp<"constant", [Pure, ConstantLike]> {
   ];
 }
 
+def ConstantTimeOp : MooreOp<"constant_time", [Pure]> {
+  let summary = "A constant time value in femtoseconds";
+  let arguments = (ins UI64Attr:$value);
+  let results = (outs TimeType:$result);
+  let assemblyFormat = "$value `fs` attr-dict";
+}
+
 def StringConstantOp : MooreOp<"string_constant", [Pure]> {
   let summary = "Produce a constant string value";
   let description = [{
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -8,6 +8,7 @@
 
 #include "ImportVerilogInternals.h"
 #include "slang/ast/Compilation.h"
+#include "llvm/ADT/ScopeExit.h"
 
 using namespace circt;
 using namespace ImportVerilog;
@@ -650,6 +651,13 @@ struct ModuleVisitor : public BaseVisitor {
 LogicalResult Context::convertCompilation() {
   const auto &root = compilation.getRoot();
 
+  // Keep track of the local time scale. `getTimeScale` automatically looks
+  // through parent scopes to find the time scale effective locally.
+  auto prevTimeScale = timeScale;
+  timeScale = root.getTimeScale().value_or(slang::TimeScale());
+  auto timeScaleGuard =
+      llvm::make_scope_exit([&] { timeScale = prevTimeScale; });
+
   // First only to visit the whole AST to collect the hierarchical names without
   // any operation creating.
   for (auto *inst : root.topInstances)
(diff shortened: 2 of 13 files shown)


lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -696,6 +703,12 @@ Context::convertModuleHeader(const slang::ast::InstanceBodySymbol *module) {
   using slang::ast::PortSymbol;
   using slang::ast::TypeParameterSymbol;
 
+  // Keep track of the local time scale.
+  auto prevTimeScale = timeScale;
+  timeScale = module->getTimeScale().value_or(slang::TimeScale());
```
uenoku, line 710: ~Is it correct to use default constructor? Shouldn't outer timescale override the inner if not exist?~

Could you mention that `getTimeScale` automatically looks at outer scope to find time scale?

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -3028,3 +3028,20 @@ function void rangeSelectRValue(
...
+  time a1 = 2.345ns;
+  // CHECK: moore.constant_time 350000 fs
+  time a2 = 345ps;
+  // CHECK: moore.constant_time 45000000 fs
+  realtime b0 = 45ns;
+  // CHECK: moore.constant_time 5680000 fs
+  realtime b1 = 5.678ns;
+  // CHECK: moore.constant_time 680000 fs
+  realtime b2 = 678ps;
+endmodule
```
uenoku, line 3060: Please add tests for default behavior and nested scopes.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -600,6 +600,19 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
 
+struct ConstantTimeOpConv : public OpConversionPattern<ConstantTimeOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ConstantTimeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOpWithNewOp<llhd::ConstantTimeOp>(
+        op, llhd::TimeAttr::get(op->getContext(), op.getValue(),
+                                StringRef("fs"), 0, 0));
```
uenoku, line 611: Not directly related but I'm wondering if we can unify to fs in LLHD as well.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -492,6 +492,13 @@ def ConstantOp : MooreOp<"constant", [Pure, ConstantLike]> {
   ];
 }
 
+def ConstantTimeOp : MooreOp<"constant_time", [Pure]> {
```
uenoku, line 495: ConstantLike?

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: Yes that would be great!

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: That requires a folder for the op, and an attribute to represent time constants. Will add that later :+1:

PR #8830, opened 2025-08-07
[FIRRT] Domains: Types Approach

seldridge:
This is an experimental approach to add domains to FIRRTL.  This approach
sees what it looks like to put domain information into the MLIR type
system.

This gets very funky, very quickly because this is doing many things which
are very against the ethos of MLIR:

  1. Some types have symbol storage.
  2. Some types have inner symbol storage.

(2) is specifically working around the fact that MLIR provides no built-in
way to associate a value to a type.  (There are good reasons for this!)

This will be contrasted with a future approach that does an association of
domain information on a per-operation basis.

This is broken up into the following logical commits:

  - [FIRRTL] Add `firrtl.domain`
  - [FIRRTL] Add firrtl.domain type
  - [FIRRTL] Add domain information to types
  - [FIRRTL] Canonicalize domains on scalar types

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLStructure.td
@@ -597,4 +597,24 @@ def SimulationOp : FIRRTLOp<"simulation", [
   }];
 }
 
+def DomainOp : FIRRTLOp<"domain", [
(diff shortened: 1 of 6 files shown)


seldridge:
Closed in favor or #8843.

PR #8834, opened 2025-08-08
[Support] Use git commit hash as version in non-release builds

fabianschuiki:
Instead of just saying "unknown git version", use the current git commit hash as CIRCT version in non-release builds. This makes the output of `--version` of our tools more useful, because it points to an actual commit.

With this change, the [sv-tests-results] website can provide an actual git commit tooltip for the circt-verilog tool being used.

[sv-tests-results]: https://chipsalliance.github.io/sv-tests-results/


cmake/modules/GenVersionFile.cmake
```diff
@@ -1,36 +1,64 @@
...
-#   IN_FILE             - An absolute path of Version.cpp.in
-#   OUT_FILE            - An absolute path of Version.cpp
-#   RELEASE_PATTERN     - A pattern to search release tags
-#   DRY_RUN             - If true, make the version unknown.
-#   SOURCE_ROOT         - Path to root directory of source
+#
+# - `IN_FILE`:          Absolute path of `Version.cpp.in`
+# - `OUT_FILE`:         Absolute path of `Version.cpp`
+# - `RELEASE_PATTERN`:  A pattern like `firtool*` to search release tags
+# - `RELEASE_ENABLED`:  Use release tag if true, or git hash if false
```
uenoku: nit: When I first looked at `RELEASE_ENABLED` I thought this is about build option (Release/Debug/RelWithDebugInfo). Could you rename the variables to `RELEASE_TAG_PATTERN` and `RELEASE_TAG_ENABLED`?

fabianschuiki: Done :+1:

PR #8837, opened 2025-08-08
[Reducer] Allow reductions to match multiple times per op

fabianschuiki:
Instead of only allowing reductions to match or not match an entire op, allow reductions to declare a list of zero, one, or multiple matches. Each match has a separate `uint64_t` integer to identify it. This can be an operand index for an operand forwarding reduction, or an argument number for an argument removing reduction, or anything else.

This commit makes the reducer gather all of the potentially multiple matches per op into a list. It then sorts this list by benefit and, as before, uses the `rangeBase` and `rangeLength` to perform a binary search through that list of matches. To apply reductions, the reducer collects all matches on the same op and calls a new `rewriteMatches` function with that list of matches. The reduction patterns can then decide how to perform the rewrite with the given selection of matches.

In a subsequent PR, I will update a few of the HW reduction patterns to use this new capability.

### DIFF
--- a/tools/circt-reduce/circt-reduce.cpp
+++ b/tools/circt-reduce/circt-reduce.cpp
@@ -165,6 +165,12 @@ static LogicalResult writeOutput(ModuleOp module) {
   return success();
 }
 
+struct Match {
+  Operation *op;
+  uint64_t id;
+  uint64_t benefit;
+};
+
 /// Execute the main chunk of work of the tool. This function reads the input
 /// module and iteratively applies the reduction strategies until no options
 /// make it smaller.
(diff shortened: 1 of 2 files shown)


tools/circt-reduce/circt-reduce.cpp
```diff
@@ -262,31 +268,91 @@ static LogicalResult execute(MLIRContext &context) {
...
+
+      // Drop all matches before `rangeBase`.
+      if (rangeBase < matches.size())
+        matches.erase(matches.begin(), matches.begin() + rangeBase);
+      else
+        matches.clear();
+
+      // Pick the next `rangeLength` of matches and apply the pattern. This may
+      // not work for all matches, which is why we do this in a loop until we
+      // ahve actually applied `rangeLength` matches.
```
dtzSiFive:
```suggestion
      // have actually applied `rangeLength` matches.
```

PR #8838, opened 2025-08-08
[HW] Make constantifier reduction treat results separately

fabianschuiki:
Make the reduction pattern that replaces operation results with constants more fine-grained. Since the reducer now supports multiple matches per op, this reduction can generate a separate match for each of the op results. This allows the reducer to replace _some_ of an op's results with a constant, but not necessarily all of them.

### DIFF
--- a/lib/Dialect/HW/HWReductions.cpp
+++ b/lib/Dialect/HW/HWReductions.cpp
@@ -113,19 +113,20 @@ struct HWOperandForwarder : public Reduction {
 /// A sample reduction pattern that replaces integer operations with a constant
 /// zero of their type.
 struct HWConstantifier : public Reduction {
-  uint64_t match(Operation *op) override {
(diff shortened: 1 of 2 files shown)


lib/Dialect/HW/HWReductions.cpp
```diff
@@ -113,19 +113,20 @@ struct HWOperandForwarder : public Reduction {
...
-  uint64_t match(Operation *op) override {
-    if (op->getNumResults() == 0 || op->getNumOperands() == 0)
-      return 0;
-    return llvm::all_of(op->getResults(), [](Value result) {
-      return isa<IntegerType>(result.getType());
-    });
+  void matches(Operation *op,
+               llvm::function_ref<void(uint64_t, uint64_t)> addMatch) override {
+    for (auto result : op->getResults())
+      if (!result.use_empty())
```
dtzSiFive, line 119: Filtering on results that are unused seems reasonable to me.

PR #8839, opened 2025-08-08
[HW] Use multiple reduction matches in port pruner

fabianschuiki:
Extend the HW reduction pattern that removes input and output ports. Since the reducer now supports multiple matches per op for a pattern, we can generate separate matches for every input and output port. The reducer then picks some of these and hands them back to the `rewriteMatches` function, where we can remove all the selected inputs and outputs of a module.

Also make the reductions adjust instance ops such that all modules can have their ports pruned, instead of just the top-level. Support extern modules as well, which are common due to the module externalizer reduction.

This allows the reducer to get rid of ports very aggressively.

### DIFF
--- a/lib/Dialect/HW/HWReductions.cpp
+++ b/lib/Dialect/HW/HWReductions.cpp
@@ -135,81 +135,156 @@ struct HWConstantifier : public Reduction {
   std::string getName() const override { return "hw-constantifier"; }
 };
 
-/// Remove the first or last output of the top-level module depending on the
-/// 'Front' template parameter.
-template <bool Front>
-struct ModuleOutputPruner : public OpReduction<HWModuleOp> {
+/// Remove unused module input ports.
+struct ModuleInputPruner : public Reduction {
   void beforeReduction(mlir::ModuleOp op) override {
-    useEmpty.clear();
-
-    SymbolTableCollection table;
-    SymbolUserMap users(table, op);
-    for (auto module : op.getOps<HWModuleOp>())
-      if (users.useEmpty(module))
-        useEmpty.insert(module);
+    symbolTables = std::make_unique<SymbolTableCollection>();
+    symbolUsers = std::make_unique<SymbolUserMap>(*symbolTables, op);
   }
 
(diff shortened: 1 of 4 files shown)


lib/Dialect/HW/HWReductions.cpp
```diff
@@ -135,81 +135,156 @@ struct HWConstantifier : public Reduction {
...
-        useEmpty.insert(module);
+    symbolTables = std::make_unique<SymbolTableCollection>();
+    symbolUsers = std::make_unique<SymbolUserMap>(*symbolTables, op);
   }
 
-  uint64_t match(HWModuleOp op) override {
-    return op.getNumOutputPorts() != 0 && useEmpty.contains(op);
+  void matches(Operation *op,
+               llvm::function_ref<void(uint64_t, uint64_t)> addMatch) override {
+    auto module = dyn_cast<HWModuleLike>(op);
```
maerhart: Nit: 'module' is a keyword starting with C++20, so maybe avoid it as a variable name

lib/Dialect/HW/HWReductions.cpp
```diff
@@ -135,81 +135,156 @@ struct HWConstantifier : public Reduction {
...
+
+    // Remove the ports from the module.
+    SmallVector<unsigned> indexList;
+    BitVector indexSet(module.getNumInputPorts());
+    for (auto idx : matches) {
+      indexList.push_back(idx);
+      indexSet.set(idx);
+    }
+    llvm::sort(indexList);
+    module.erasePorts(indexList, {});
```
maerhart: Hmm, I thought we deprecated all accessors using separate input/output lists, but can't find anything else that could be used instead. 🤔

fabianschuiki: I haven't seen anything, either! This also isn't marked as deprecated explicitly 🤔

lib/Dialect/HW/HWReductions.cpp
fabianschuiki: Good point, let me rename this 👍

PR #8840, opened 2025-08-10
[ImportVerilog] Fix file locations for auto-discovered files

fabianschuiki:
The ImportVerilog conversion currently uses a hacky `bufferFilePaths` table to map from `slang::BufferID`s to the corresponding file path. This only works for files that we explicitly add to the source manager. Files discovered automatically by Slang, e.g. through the `-y` flag, have no entry in that table. As a result, any errors in such files will report their file path as `-`, which is not very helpful.

This commit fixes this by using whatever Slang's own `SourceManager::getFileName` returns for a given location. This removes the need for that `bufferFilePaths` table. As a result, errors in files discovered via `-y` will now properly show the file name.

While we're at it, also attach the include file stack as a note to any reported diagnostic. This makes it easier to understand how the compiler arrived at an error location.


fabianschuiki:
Great idea! Added a test :+1:

Issue #8831, opened 2025-08-08
[ImportVerilog] Crash when assignment LHS is a multi-dimensional array concat


fabianschuiki:
The following crashes:
```systemverilog
// circt-verilog
module Foo(input logic [1023:0] x);
  logic [7:0][63:0] a, b;
  always_comb {a, b} = x;
endmodule
```
Stack trace:
```
 #8 SmallVectorBase<unsigned int>::set_size(...) SmallVector.h:90:5
 #9 SmallVectorTemplateBase<Type, true>::push_back(...) SmallVector.h:565:11
#10 moore::ConcatRefOp::inferReturnTypes(...) MooreOps.cpp:672:11
#11 moore::ConcatRefOp::build(...) Moore.cpp.inc:8081:7
```

AtticusKuhn:
I wrote a PR about this:
https://github.com/llvm/circt/pull/8841

PR #8841, opened 2025-08-11
[ImportVerilog] Fix Bug in Packed Array Lvalue

AtticusKuhn:
Fixes:
https://github.com/llvm/circt/issues/8831

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -670,119 +670,72 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
   LogicalResult
   matchAndRewrite(ExtractOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    // TODO: return X if the domain is four-valued for out-of-bounds accesses
-    // once we support four-valued lowering
+    Location loc = op.getLoc();
     Type resultType = typeConverter->convertType(op.getResult().getType());
     Type inputType = adaptor.getInput().getType();
     int32_t low = adaptor.getLowBit();
 
+    // Case 1: The input is an integer type.
     if (isa<IntegerType>(inputType)) {
-      int32_t inputWidth = inputType.getIntOrFloatBitWidth();
-      int32_t resultWidth = resultType.getIntOrFloatBitWidth();
-      int32_t high = low + resultWidth;
+      int64_t resultWidth = hw::getBitWidth(resultType);
+      if (resultWidth < 0)
+        return rewriter.notifyMatchFailure(
+            op, "cannot determine bit width of result type");
 
-      SmallVector<Value> toConcat;
-      if (low < 0)
-        toConcat.push_back(hw::ConstantOp::create(
-            rewriter, op.getLoc(), APInt(std::min(-low, resultWidth), 0)));
+      // We are extracting a slice of bits from an integer. This is a simple
+      // part-select. The slice will have an integer type.
+      Value extractedInt = rewriter.create<comb::ExtractOp>(
+          loc, adaptor.getInput(), low, resultWidth);
 
-      if (low < inputWidth && high > 0) {
-        int32_t lowIdx = std::max(low, 0);
-        Value middle = rewriter.createOrFold<comb::ExtractOp>(
-            op.getLoc(),
-            rewriter.getIntegerType(
-                std::min(resultWidth, std::min(high, inputWidth) - lowIdx)),
-            adaptor.getInput(), lowIdx);
-        toConcat.push_back(middle);
+      // If the target result type is already this integer type, we are done.
+      if (extractedInt.getType() == resultType) {
+        rewriter.replaceOp(op, extractedInt);
+        return success();
       }
 
-      int32_t diff = high - inputWidth;
-      if (diff > 0) {
-        Value val =
-            hw::ConstantOp::create(rewriter, op.getLoc(), APInt(diff, 0));
-        toConcat.push_back(val);
-      }
-
-      Value concat =
-          rewriter.createOrFold<comb::ConcatOp>(op.getLoc(), toConcat);
-      rewriter.replaceOp(op, concat);
+      // If the target result type is something else (e.g., an array), we must
+      // bitcast the extracted integer slice to that type. This handles the
+      // `logic [7:0] a = int_val;` case.
+      rewriter.replaceOpWithNewOp<hw::BitcastOp>(op, resultType,
+                                                 extractedInt);
       return success();
     }
 
+    // Case 2: The input is an array type.
     if (auto arrTy = dyn_cast<hw::ArrayType>(inputType)) {
       int32_t width = llvm::Log2_64_Ceil(arrTy.getNumElements());
       int32_t inputWidth = arrTy.getNumElements();
 
+      // Subcase 2.1: The result is also an array (Array Slice).
       if (auto resArrTy = dyn_cast<hw::ArrayType>(resultType)) {
-        int32_t elementWidth = hw::getBitWidth(arrTy.getElementType());
-        if (elementWidth < 0)
-          return failure();
-
-        int32_t high = low + resArrTy.getNumElements();
         int32_t resWidth = resArrTy.getNumElements();
+        Value lowIdxVal = rewriter.create<hw::ConstantOp>(
+            loc, rewriter.getIntegerType(width), low);
+        Value middle = rewriter.createOrFold<hw::ArraySliceOp>(
+            loc,
+            hw::ArrayType::get(arrTy.getElementType(),
+                               std::min(resWidth, inputWidth - low)),
+            adaptor.getInput(), lowIdxVal);
 
-        SmallVector<Value> toConcat;
-        if (low < 0) {
-          Value val = hw::ConstantOp::create(
-              rewriter, op.getLoc(),
-              APInt(std::min((-low) * elementWidth, resWidth * elementWidth),
-                    0));
-          Value res = rewriter.createOrFold<hw::BitcastOp>(
-              op.getLoc(), hw::ArrayType::get(arrTy.getElementType(), -low),
-              val);
-          toConcat.push_back(res);
-        }
-
-        if (low < inputWidth && high > 0) {
-          int32_t lowIdx = std::max(0, low);
-          Value lowIdxVal = hw::ConstantOp::create(
-              rewriter, op.getLoc(), rewriter.getIntegerType(width), lowIdx);
-          Value middle = rewriter.createOrFold<hw::ArraySliceOp>(
-              op.getLoc(),
-              hw::ArrayType::get(
-                  arrTy.getElementType(),
-                  std::min(resWidth, std::min(inputWidth, high) - lowIdx)),
-              adaptor.getInput(), lowIdxVal);
-          toConcat.push_back(middle);
-        }
-
-        int32_t diff = high - inputWidth;
-        if (diff > 0) {
-          Value constZero = hw::ConstantOp::create(
-              rewriter, op.getLoc(), APInt(diff * elementWidth, 0));
-          Value val = hw::BitcastOp::create(
-              rewriter, op.getLoc(),
-              hw::ArrayType::get(arrTy.getElementType(), diff), constZero);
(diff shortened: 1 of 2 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -670,119 +670,70 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
   LogicalResult
   matchAndRewrite(ExtractOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    // TODO: return X if the domain is four-valued for out-of-bounds accesses
-    // once we support four-valued lowering
```
TaoBi22: This TODO has been removed, is it resolved now?

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -670,119 +670,70 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
...
+    Location loc = op.getLoc();
     Type resultType = typeConverter->convertType(op.getResult().getType());
     Type inputType = adaptor.getInput().getType();
     int32_t low = adaptor.getLowBit();
 
+    // The input is an integer type.
```
TaoBi22: This comment seems superfluous

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -670,119 +670,70 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
...
-      if (diff > 0) {
-        Value val =
-            hw::ConstantOp::create(rewriter, op.getLoc(), APInt(diff, 0));
-        toConcat.push_back(val);
+      int64_t resultWidth = hw::getBitWidth(resultType);
+      if (resultWidth < 0)
+        return rewriter.notifyMatchFailure(
+            op, "cannot determine bit width of result type");
+
+      //  The slice will have an integer type.
```
TaoBi22: Not sure this comment is necessary

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -670,119 +670,70 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
...
-
-      int32_t diff = high - inputWidth;
-      if (diff > 0) {
-        Value val =
-            hw::ConstantOp::create(rewriter, op.getLoc(), APInt(diff, 0));
-        toConcat.push_back(val);
+      int64_t resultWidth = hw::getBitWidth(resultType);
+      if (resultWidth < 0)
+        return rewriter.notifyMatchFailure(
+            op, "cannot determine bit width of result type");
```
TaoBi22: If we add a new error message it tends to be good to include a test that triggers it

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -670,119 +670,70 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
...
-          rewriter.createOrFold<comb::ConcatOp>(op.getLoc(), toConcat);
-      rewriter.replaceOp(op, concat);
+      // If the target result type is something else (e.g., an array), we must
+      // bitcast the extracted integer slice to that type. This handles the
+      // `logic [7:0] a = int_val;` case.
+      rewriter.replaceOpWithNewOp<hw::BitcastOp>(op, resultType, extractedInt);
       return success();
     }
 
+    // The input is an array type.
```
TaoBi22: Same as above

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -670,119 +670,70 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
...
-                                         rewriter.getIntegerType(width),
-                                         adaptor.getLowBit());
+      Value idx = rewriter.create<hw::ConstantOp>(
+          loc, rewriter.getIntegerType(width), adaptor.getLowBit());
       rewriter.replaceOpWithNewOp<hw::ArrayGetOp>(op, adaptor.getInput(), idx);
       return success();
     }
 
-    return failure();
+    return rewriter.notifyMatchFailure(op, "unhandled extract pattern");
```
TaoBi22: As above, new test for the new error message please!

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -663,11 +663,30 @@ LogicalResult ConcatRefOp::inferReturnTypes(
...
+  for (Value operand : operands) {
+    UnpackedType nestedType = cast<RefType>(operand.getType()).getNestedType();
+    PackedType packedType = dyn_cast<PackedType>(nestedType);
+
+    if (!packedType) {
+      if (loc)
+        llvm::outs() << "operand of moore.concat_ref must be a reference to a "
+                        "packed type, but got "
+                     << nestedType;
+      return failure();
```
TaoBi22: Is there a reason these are being streamed to llvm::outs() rather than using emitError? Also, new test for error message please

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -663,11 +663,30 @@ LogicalResult ConcatRefOp::inferReturnTypes(
...
       domain = Domain::FourValued;
-    width += type.getWidth();
+
+    // getBitSize() for PackedType returns an optional, so we must check it.
+    std::optional<int> bitSize = packedType.getBitSize();
+    if (!bitSize) {
+      if (loc)
+        llvm::outs() << "operand type " << packedType
+                     << " in moore.concat_ref has no known bit size";
+      return failure();
```
TaoBi22: As above

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -663,11 +663,30 @@ LogicalResult ConcatRefOp::inferReturnTypes(
     mlir::RegionRange regions, SmallVectorImpl<Type> &results) {
   Domain domain = Domain::TwoValued;
   unsigned width = 0;
-  for (auto operand : operands) {
-    auto type = cast<IntType>(cast<RefType>(operand.getType()).getNestedType());
-    if (type.getDomain() == Domain::FourValued)
+  for (Value operand : operands) {
+    UnpackedType nestedType = cast<RefType>(operand.getType()).getNestedType();
```
TaoBi22, line 667: is this definitely a safe cast?

lib/Dialect/Moore/MooreOps.cpp
AtticusKuhn: `emitError()` says
`Call to non-static member function without an object argumentclang(member_call_without_object)`

lib/Dialect/Moore/MooreOps.cpp
AtticusKuhn: I think so. Maybe?

- RefType: A wrapper is used to wrap any SystemVerilog type. It's aimed to work for 'moore.variable', 'moore.blocking_assign', and 'moore.read', which are related to memory, like alloca/write/read.

- PackedType: Packed types are the core types of SystemVerilog.

lib/Dialect/Moore/MooreOps.cpp
TaoBi22: Oh, I missed that we were in inferReturnTypes - I'd imagine we want this kind of thing to be caught in a verifier, not in inferReturnTypes

test/circt-verilog/packed_lvalue.sv
```diff
@@ -0,0 +1,8 @@
+// RUN: circt-verilog %s | FileCheck %s
+// REQUIRES: slang
+
+// CHECK-LABEL: hw.module @Foo(
+module Foo(input logic [1023:0] x);
+  logic [7:0][63:0] a, b;
+  always_comb {a, b} = x;
+endmodule
```
fabianschuiki: Thanks for adding a test :raised_hands:! Could you move this test into `test/Conversions/ImportVerilog/basic.sv` and add a few `// CHECK:` lines to ensure that the conversion works as expected? If you run circt-verilog in parse-only mode (`circt-verilog --parse-only`), the canonicalizers won't remove the unused variables, so you'll be able to see what ops got created. (The `basic.sv` file is already setup to do that.)

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1901,4 +1903,4 @@ void MooreToCorePass::runOnOperation() {
 
   if (failed(applyFullConversion(module, target, std::move(patterns))))
     signalPassFailure();
-}
+}
```
fabianschuiki: Nit: I think you a accidentally removed the final newline here.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -678,7 +679,8 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
 
     if (isa<IntegerType>(inputType)) {
       int32_t inputWidth = inputType.getIntOrFloatBitWidth();
-      int32_t resultWidth = resultType.getIntOrFloatBitWidth();
+      int32_t resultWidth =
+          hw::getBitWidth(resultType); // resultType.getIntOrFloatBitWidth();
```
fabianschuiki: Hmmm, I don't think this would help fix the issue. If the input type is an integer, the result should definitely also be an integer, in which case `getIntOrFloatBitWidth()` should work just fine. If this line caused a crash, chances are that we are missing an op verifier on `ExtractOp` that ensures that the result type matchse the input type, just with fewer bits.

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -663,11 +663,23 @@ LogicalResult ConcatRefOp::inferReturnTypes(
...
+
+    // getBitSize() for PackedType returns an optional, so we must check it.
+    std::optional<int> bitSize = packedType.getBitSize();
+    if (!bitSize) {
+      return failure();
+    }
+    width += *bitSize;
   }
   results.push_back(RefType::get(IntType::get(context, width, domain)));
   return success();
```
fabianschuiki, line 685: The original `inferReturnTypes` implementation was pretty primitive: it just assumes that all operands are `IntType` (for both `ConcatOp` and `ConcatRefOp`). To support arrays, we'll need to extend this a bit and handle `IntType` and `PackedArray` types separately. To do this, all operands must have the same type, except for a varying width. I would iterate over the operands, track what the common type is across all of them (maybe `IntType` and `PackedArray` to get started), and sum up the number of bits/elements of each operand. Then create a new instance of that common type with the summed-up number of bits/elements and add that to the `results` array. If you encounter an operand that does not match the previous operand's type (except for different number of bits/elements), return `failure()`.

AtticusKuhn: I think that my changes resolve this. I added four new test cases, and all of them pass `ninja check-circt`. Are you saying that my does not resolve this issue, and that this above algorithm would be better?

lib/Conversion/MooreToCore/MooreToCore.cpp
AtticusKuhn: I have an example where `resultType.getIntOrFloatBitWidth();` crashes but `hw::getBitWidth(resultType);`, so I think this change is necessary.

fabianschuiki: Oh interesting! Can you paste this example here? I'm curious what goes wrong there.

lib/Dialect/Moore/MooreOps.cpp
fabianschuiki: Hmmm I think you're right. I was thinking about unpacked array concatenation, but it looks like the standard treats those completely separately. I like the approach of just dealing with packed types here.

fabianschuiki (approved):
Thanks for adding the additional tests! LGTM 🥳

lib/Conversion/MooreToCore/MooreToCore.cpp
AtticusKuhn:
```systemverilog
// circt-verilog
module Foo(input logic [1023:0] x);
  logic [7:0][63:0] a, b;
  always_comb {a, b[0]} = x;
endmodule
```

PR #8842, opened 2025-08-11
[circt-synth] Incorporate datapath operations into circt-synth pipeline

cowardsa:
To improve synthesis area and delay results, we:

1. Lower arithmetic comb operators to datapath operators,
2. Perform datapath canonicalizations (greedily fusing datapath operators)
3. Lower datapath operators to the gate-level subset of comb, that also includes two-input adders.

Datapath optimisations are enabled by default with a flag to disable datapath optimisations when required.


uenoku (approved):
Awesome, thanks!

PR #3976, opened 2022-09-22
Revert "[FIRRTL] Verify FModuleLike's have unique port names."

dtzSiFive:
Reverts llvm/circt#3891, until have better fix for #3968 .

Conflicts in this way are not quite invalid FIRRTL per spec ($ vs _) and it seems to be expected to "work" in some designs.  FWIW SFC has tests for handling this sort of lowering conflict specifically.  Neither compiler does what the spec says re:"$" however which would avoid the conflict here.

These duplicate names will be resolved/legalized before verilog emission.

They should not be trusted until that point (similarly wires and regs and other namable things), and the way the conflicts are resolved should not be relied upon (at best implementation-defined behavior).

This sort of conflict is likely a problem if occurs on public/top module or extmodule-- that is, where used in contexts where the lowered names need to be reliably referenced from the outside.

Anyway until have better story for all this let's not error out.

PR #8843, opened 2025-08-12
[FIRRTL] Domains: Operation Approach

seldridge:
This is an alternative approach to implementing FIRRTL domains from #8830.
This represents domains as additional operands to operations instead of
capturing which domain something is colored by in its type.

This approach relies on an `InferDomains` domain inference pass to infer
and check the correctness of all domains in the system.

At present, the actual kind of a domain, e.g., if a domain is a clock
domain, a reset domain, or something else, is not linked to the underlying
domain values.

This PR is broken up into individual logical commits making the following
changes:
  - Add `firrtl.domain`
  - Add domain type
  - Add domains to operations
  - Add `unsafe_domain_cast` operation
  - Add InferDomains pass


test/Dialect/FIRRTL/round-trip.mlir
```diff
@@ -196,4 +196,24 @@ firrtl.module @Fprintf(
...
+
+// CHECK-LABEL: firrtl.module @Domains
+// CHECK-SAME:    in %A: !firrtl.domain
+// CHECK-SAME:    in %B: !firrtl.domain
+// CHECK-SAME:    in %a: !firrtl.uint<1> domains [%A]
+// CHECK-SAME:    out %b: !firrtl.uint<1> domains [%B]
+firrtl.module @Domains(
+  in %A: !firrtl.domain,
+  in %B: !firrtl.domain,
+  in %a: !firrtl.uint<1> domains [%A],
```
uenoku, line 216: Does this work with anonymous block arguments?

seldridge:
> DomainInfoAttr is a 2d integer array which has port indices, correct? 

Yes, 2D integer array with port indices. Mechanically this is being enforced in a verifier for modules and instances (with slight tweaks between them). For modules, this must be either an empty `ArrayAttr` or an `ArrayAttr` of `ArrayAttr`s where the outer dimension is the number of ports. This is the same requirements for port annotations.

For instances, this is the same as modules, except it cannot be an empty `ArrayAttr`. This is also the same as port annotations on instances, though I don't think port annotations on instances are used.

> In that case how do we keep that invariant through the pipeline?

Two ways (first is a non-way):

1. We don't. After the forthcoming `EraseDomains` pass, all domain information is removed and the domain info on modules and instances is empty.
2. We do enforce it and we will probably need either new operations or tracking a bit that we are after domain inference and we need stricter checks.

> Is it legal to define a port before Domain?

Currently, no. This is only being checked in the FIRRTL parser and MLIR parser, though, not with a verifier. There's no actual problem with, though, just we've never allowed FIRRTL text to ever refer to an identifier that comes later. The underlying storage would allow this, though.

test/Dialect/FIRRTL/round-trip.mlir
seldridge: No, it does not. This probably needs to use an alternative code path when there is a block arg that doesn't have a name.

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
```diff
@@ -607,4 +607,11 @@ def FStringImpl : FIRRTLImplType<"FString", [], "::circt::firrtl::FIRRTLType" >
   let typeName = "firrtl.fstring";
 }
 
+def DomainTypeImpl : FIRRTLImplType<"Domain", [], "::circt::firrtl::FIRRTLType"> {
+  let summary = "A domain type";
+  let parameters = (ins);
```
uenoku, line 612: Does this eventually take a symbol defined by DomainOp?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -824,6 +828,7 @@ static void insertPorts(FModuleLike op,
       newDirections.push_back(existingDirections[oldIdx]);
       newNames.push_back(existingNames[oldIdx]);
       newTypes.push_back(existingTypes[oldIdx]);
+      newDomains.push_back(op.getDomainInfoAttrForPort(oldIdx));
```
uenoku: `insertPorts` implicitly assumes domain port indices are not changed. Could you insert an assertion that domain ports indexes are not mutated, or properly update domain port indexes?

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -294,7 +297,8 @@ def FMemModuleOp : FIRRTLModuleLike<"memmodule"> {
            DefaultValuedAttr<RUWBehaviorAttr,
              "firrtl::RUWBehavior::Undefined">:$ruw,
            AnnotationArrayAttr:$annotations,
-           DefaultValuedAttr<LayerArrayAttr, "{}">:$layers
+           DefaultValuedAttr<LayerArrayAttr, "{}">:$layers,
+           DefaultValuedAttr<ArrayRefAttr, "{}">:$domainInfo
```
uenoku, line 301: Allowing intrinsics to specify domains for operands/results make sense but certainly that will require special handling for each operation in infer domain (or maybe we can prepare op interface to specify domain inference). For now could you reject intrinsics with domain info  in LowerIntrinsic or somewhere?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -855,11 +862,23 @@ static void insertPorts(FModuleLike op,
...
+  // The lack of *any* domains is also represented by an empty `domainInfo`
+  // attribute.
+  if (llvm::all_of(newDomains, [](Attribute attr) {
+        return !attr || cast<ArrayAttr>(attr).empty();
+      }))
+    newDomains.clear();
+
+  for (auto domain : newDomains) {
+    llvm::errs() << domain << "\n";
+  }
```
uenoku:
```suggestion

```

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -1200,6 +1231,21 @@ printModulePorts(OpAsmPrinter &p, Block *block, ArrayRef<bool> portDirections,
...
+    // Print domain information.
+    if (!domainInfo.empty()) {
+      auto domains = cast<ArrayAttr>(domainInfo[i]);
+      if (!domains.empty()) {
+        p << " domains [";
+        llvm::interleaveComma(domains, p, [&](Attribute attr) {
+          if (block)
+            p << "%";
+          p << cast<StringAttr>(portNames[cast<IntegerAttr>(attr).getUInt()])
+                   .getValue();
```
uenoku: nit: Port names may not correspond to SSA names so can we just print block arguments when there is a block?

include/circt/Dialect/FIRRTL/FIRRTLTypesImpl.td
seldridge: Conceptually, yes, each domain type is associated with a `DomainOp`. Mechanically, no, as types don't work well with symbols.

What I was planning to do here is to allow for the `domainInfo` `ArrayAttr` to contain either `SymbolRefAttr` (to refer to a domain for domain typed ports) or an `ArrayAttr` (to track associations of non-domain type ports). I need to hook this up, though.

seldridge: https://github.com/llvm/circt/pull/8843/commits/092f9e8f80ddd6873f88e6d1530c5acbefc47a6b updates the `domainInfo` storage as discussed above. Each port of domain type now has to indicate what domain kind it is.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: f8ae35e81 adds logic for properly updating ports when inserted or deleted.

One problem exists with this approach related to needing to both insert domain ports and reference them. There is currently no way to do this as the inserted `PortInfo` must refer to existing ports. Such an insertion can be accomplished with two insertions: (1) insert new domain ports and (2) insert ports associated with them.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: Thanks for catching this. Fixed in: 95924cb91

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: Good call. This is fixed in: https://github.com/llvm/circt/pull/8843/commits/8758bdef6aedb20899af148a137124dbaadc0015 That includes a test of this, so something like the following will now round trip:

``` mlir
firrtl.circuit "Bar" {
  firrtl.domain @ClockDomain {
  }
  firrtl.module @Bar(
    in %arg0: !firrtl.domain of @ClockDomain, 
    in %a: !firrtl.uint<1> domains [%arg0]
  ) attributes {portNames = ["", "a"]} {
  }
}
```

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
seldridge: I think this is going to be the same as with external modules. They can define domains, but they need to be completely specified. Same thing for public modules, no inferred domains. WDYT?

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -877,6 +922,82 @@ static void insertPorts(FModuleLike op,
...
+    }
+    // This contains indexes to domain ports.  Update them.
+    for (auto domain : domains) {
+      auto oldIdx = cast<IntegerAttr>(domain).getUInt();
+      auto newIdx = oldIdx - numDeleted[oldIdx];
+      if (oldIdx == newIdx)
+        newPortDomains.push_back(attr);
+      else
+        newPortDomains.push_back(IntegerAttr::get(
+            IntegerType::get(context, 32, IntegerType::Unsigned), newIdx));
```
seldridge: There is a third case that is not captured here: a domain port could be deleted. If this happens, then the domain information should be dropped from those ports that reference it.

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
```diff
@@ -596,4 +602,24 @@ def SimulationOp : FIRRTLOp<"simulation", [
...
+def DomainOp : FIRRTLOp<"domain", [
+  DeclareOpInterfaceMethods<Symbol>,
+  HasParent<"firrtl::CircuitOp">,
+  NoTerminator,
+  SingleBlock
+]> {
+  let summary = "Define a domain type";
+  let description = [{
+    A `firrtl.domain` operation defines a type of domain that exists in this
+    circuit.  E.g., this can be used to delcare a `ClockDomain` type when
```
dtzSiFive:
```suggestion
    circuit.  E.g., this can be used to declare a `ClockDomain` type when
```

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1228,7 +1228,8 @@ FIRRTLModuleLowering::lowerModule(FModuleOp oldModule, Block *topLevelModule,
       "annotations",   "convention",      "layers",
       "portNames",     "sym_name",        "portDirections",
       "portTypes",     "portAnnotations", "portSymbols",
-      "portLocations", "parameters",      SymbolTable::getVisibilityAttrName()};
```
dtzSiFive, line 1231: Adjust smallvector size to 13 please.

lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
```diff
@@ -120,6 +120,39 @@ LogicalResult circt::firrtl::verifyModuleLikeOpInterface(FModuleLike module) {
...
+    if (llvm::any_of(domains.getValue(), [](Attribute attr) {
+          auto arrayAttr = dyn_cast_or_null<ArrayAttr>(attr);
+          if (!arrayAttr)
+            return false;
+          return llvm::any_of(arrayAttr, [](Attribute attr) {
+            return !isa<IntegerAttr>(attr);
+          });
+        }))
+      return module.emitOpError()
+             << "domain information for non-domain port" << portNames[index]
```
dtzSiFive:
```suggestion
             << "domain information for non-domain port " << portNames[index]
```

lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
```diff
@@ -120,6 +120,39 @@ LogicalResult circt::firrtl::verifyModuleLikeOpInterface(FModuleLike module) {
...
+  //   2. An ArrayAttr, one entry-per-port, of:
+  //      a. SymbolRefAttrs if the port is a domain type
+  //      b. IntegerAttrs if the port is a non-domain type
+  auto domains = module.getDomainInfoAttr();
+  if (!domains)
+    return module.emitOpError("requires valid port domains");
+  if (!domains.empty() && domains.size() != numPorts)
+    return module.emitOpError("requires ")
+           << numPorts << " port domains, but has " << domains.size();
+  for (auto [index, domain] : llvm::enumerate(module.getDomainInfo())) {
```
dtzSiFive:
```suggestion
  for (auto [index, domain] : llvm::enumerate(domains.getValue())) {
```

lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
```diff
@@ -120,6 +120,39 @@ LogicalResult circt::firrtl::verifyModuleLikeOpInterface(FModuleLike module) {
...
+    auto type = cast<TypeAttr>(portTypes[index]).getValue();
+    if (isa<DomainType>(type)) {
+      if (!isa<FlatSymbolRefAttr>(domain))
+        return module.emitOpError()
+               << "domain information for domain port " << portNames[index]
+               << " must be a 'FlatSymbolRefAttr'";
+      continue;
+    }
+    if (llvm::any_of(domains.getValue(), [](Attribute attr) {
```
dtzSiFive: Should this only walk `domain`? Probably moved from outside the loop? As-is walks all arrays of all elements of the domains array for each of the outer loop.

Something like (wasn't sure whether should skip or error if not an array):
``` patch
From 49a7ea4176a19c616f8bbb8de4fedea5850f74c3 Mon Sep 17 00:00:00 2001
From: Will Dietz <will.dietz@sifive.com>
Date: Wed, 27 Aug 2025 10:06:42 -0500
Subject: [PATCH] walk specific array

---
 lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp | 13 +++++--------
 1 file changed, 5 insertions(+), 8 deletions(-)

diff --git a/lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp b/lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
index af1199f3a6..19ce8d768f 100644
--- a/lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
@@ -140,14 +140,11 @@ LogicalResult circt::firrtl::verifyModuleLikeOpInterface(FModuleLike module) {
                << " must be a 'FlatSymbolRefAttr'";
       continue;
     }
-    if (llvm::any_of(domains.getValue(), [](Attribute attr) {
-          auto arrayAttr = dyn_cast_or_null<ArrayAttr>(attr);
-          if (!arrayAttr)
-            return false;
-          return llvm::any_of(arrayAttr, [](Attribute attr) {
-            return !isa<IntegerAttr>(attr);
-          });
-        }))
+    auto arrayAttr = dyn_cast_or_null<ArrayAttr>(domain);
+    if (!arrayAttr)
+      continue;
+    if (llvm::any_of(arrayAttr,
+                     [](Attribute attr) { return !isa<IntegerAttr>(attr); }))
       return module.emitOpError()
              << "domain information for non-domain port" << portNames[index]
              << " must be an 'ArrayAttr<IntegerAttr>'";
--
2.44.1
```

dtzSiFive: Consider adding error tests for these.

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
```diff
@@ -79,12 +79,15 @@ struct SpecializeOptionPass
...
             ImplicitLocOpBuilder builder(inst.getLoc(), inst);
+            // TODO: Fixup this to lower domains.
+            SmallVector<Attribute> domains;
             auto newInst = InstanceOp::create(
                 builder, inst->getResultTypes(), target, inst.getNameAttr(),
                 inst.getNameKindAttr(), inst.getPortDirectionsAttr(),
-                inst.getPortNamesAttr(), inst.getAnnotationsAttr(),
-                inst.getPortAnnotationsAttr(), builder.getArrayAttr({}),
-                UnitAttr{}, UnitAttr{}, inst.getInnerSymAttr());
+                inst.getPortNamesAttr(), builder.getArrayAttr(domains),
```
dtzSiFive: Consider `getArrayAttr({})` (or doing this once and re-using) like other empty-array arguments.
With the TODO, maybe it's done this way to build on later so 🤷 .

test/Dialect/FIRRTL/round-trip.mlir
dtzSiFive: FWIW FIRRTL allows duplicate port names, previously had verifier that needed to be rolled back: https://github.com/llvm/circt/pull/3976 .

Consider:
``` firrtl
FIRRTL version 5.1.0
circuit Domains:
  domain ClockDomain:

  public module Domains:
    input a_b: Domain of ClockDomain
    input a : { b: String }
    output b: UInt<5> domains [a_b]
    invalidate b
```

With current PR this input errors:
```
oops.fir:5:10: error: 'firrtl.module' op domain information for domain port "a_b" must be a 'FlatSymbolRefAttr'
  public module Domains:
         ^
oops.fir:5:10: note: see current operation:
"firrtl.module"() <{annotations = [], convention = #firrtl<convention scalarized>, domainInfo = [[], [], [0 : ui32]], layers = [], portAnnotations = [], portDirections = array<i1: false, false, true>, portLocations = [loc("oops.fir":6:11), loc("oops.fir":7:11), loc("oops.fir":8:12)], portNames = ["a_b", "a_b", "b"], portSymbols = [], portTypes = [!firrtl.domain, !firrtl.string, !firrtl.uint<5>], sym_name = "Domains"}> ({
^bb0(%arg0: !firrtl.domain, %arg1: !firrtl.string, %arg2: !firrtl.uint<5>):
  %0 = "firrtl.invalidvalue"() : () -> !firrtl.uint<5>
  "firrtl.matchingconnect"(%arg2, %0) : (!firrtl.uint<5>, !firrtl.uint<5>) -> ()
}) : () -> ()
```

dtzSiFive (commented):
Here we go!

Left some feedback 👍 .

Does this need a connect operation? I suggest adding one that behaves similarly to define and propassign (static single connect).

Some interesting inputs to consider:

This should error (symbol used as domain is not domain):
```firrtl
FIRRTL version 5.1.0
circuit Domains:
  public module ClockDomain:

  public module Domains:
    input A: Domain of ClockDomain
    input b: UInt<5> domains [A]
```

LowerOpenAggs lowers the open bundle to put domain in its own port causing an error.  Putting `b` in same domain twice is needed, FWIW.
Generally, where can "Domain" types be used?  From what I'm seeing in the syntax, probably Domain is not meant to be valid in a port type if it's not top-level?

```
FIRRTL version 5.1.0
circuit Domains:
  domain ClockDomain:

  public module Domains:
    input A: Domain of ClockDomain
    input x: { y: Domain }
    output b: UInt<5> domains [A, A]
    invalidate b
```

Instantiation involving domains (this example doesn't make sense yet but just wanted to see an instance):
```
FIRRTL version 5.1.0
circuit Domains:
  domain ClockDomain:

  module Child:
    output B: Domain of ClockDomain
    input b: UInt<5> domains [B]

  public module Domains:
    input A: Domain of ClockDomain
    input a: UInt<5> domains [A]

    inst c of Child
    connect c.b, a
```

include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
```diff
@@ -37,6 +37,22 @@ def PortAnnotationsAttr : ArrayAttrBase<
...
+//===----------------------------------------------------------------------===//
+
+def DomainInfoAttr : ArrayAttrBase<
+    And<[
+      // Guarantee this is an ArrayAttr first
+      CPred<"llvm::isa<::mlir::ArrayAttr>($_self)">,
+      // Guarantee all elements are a FlatSymbolRefAttr or ArrayAttr.
+      CPred<"::llvm::all_of(cast<mlir::ArrayAttr>($_self), "
+            "[&](::mlir::Attribute attr) { return isa_and_nonnull<"
+            "::mlir::ArrayAttr>(attr);})">]>,
```
dtzSiFive:
``` patch
diff --git a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
index 23f436d5a5..7293d1d30b 100644
--- a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
@@ -48,7 +48,7 @@ def DomainInfoAttr : ArrayAttrBase<
       // Guarantee all elements are a FlatSymbolRefAttr or ArrayAttr.
       CPred<"::llvm::all_of(cast<mlir::ArrayAttr>($_self), "
             "[&](::mlir::Attribute attr) { return isa_and_nonnull<"
-            "::mlir::ArrayAttr>(attr);})">]>,
+            "::mlir::ArrayAttr, ::mlir::FlatSymbolRefAttr>(attr);})">]>,
     "domain attributes must be an empty array or array-of-arrays"> {
   let constBuilderCall = "$_builder.getArrayAttr($0)";
 }
```
(modulo formatting; this fixes the instance example posted earlier)

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: Fixed in: 2c4ac9fb4

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
seldridge: Fixed in: 21ea28be6

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
seldridge: Fixed in: dc4b28825

seldridge:
> Does this need a connect operation? I suggest adding one that behaves similarly to define and propassign (static single connect).

It will. Though that can come in a subsequent PR.

> Some interesting inputs to consider:
> 
> This should error (symbol used as domain is not domain):
> ```firrtl
> FIRRTL version 5.1.0
> circuit Domains:
>   public module ClockDomain:
> 
>   public module Domains:
>     input A: Domain of ClockDomain
>     input b: UInt<5> domains [A]
> ```

Fixed in b171092ad.

> Generally, where can "Domain" types be used? From what I'm seeing in the syntax, probably Domain is not meant to be valid in a port type if it's not top-level?

I don't see a reason to not allow it on any module (or a corresponding instantiation). Domains need to be valid in any port, wire, or bundle. Support for bundles is not rolled out yet, but will be added.

> Instantiation involving domains (this example doesn't make sense yet but just wanted to see an instance)
> 
> FIRRTL version 5.1.0
> circuit Domains:
>   domain ClockDomain:
> 
>   module Child:
>     output B: Domain of ClockDomain
>     input b: UInt<5> domains [B]
> 
>   public module Domains:
>     input A: Domain of ClockDomain
>     input a: UInt<5> domains [A]
> 
>     inst c of Child
>     connect c.b, a

Your example is correct and how it should work. This is not super interesting without the connection operator for domains. However, this will force the question on if you even have to connect it or if the connection can be inferred.

dtzSiFive:
> Your example is correct and how it should work. This is not super interesting without the connection operator for domains.

Awesome! I meant this was maybe weird but is just the first way I came up with to instantiate a child with domain ports, and wasn't meant to be anything valid or not.  That wasn't the point.

I forgot to mention this had a verifier error at the time of review.

> However, this will force the question on if you even have to connect it or if the connection can be inferred.

Whoa, fascinating!

lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
seldridge: Good catch. Just `domains` should work here. Fixed in: bc7e104d6

lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
seldridge: Fixed in: 1d91cf685

include/circt/Dialect/FIRRTL/FIRRTLStructure.td
uenoku: Yes that makes sense .

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -37,7 +37,9 @@
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Casting.h"
 #include "llvm/Support/FormatVariadic.h"
+#include <mlir/IR/BuiltinAttributes.h>
```
uenoku:
```suggestion
#include "mlir/IR/BuiltinAttributes.h"
```

lib/Dialect/FIRRTL/FIRRTLOpInterfaces.cpp
seldridge: Thanks for catching this. Tests should have been there and they would have identified the problems here. Fixed in 08e72932b.

seldridge:
> FWIW FIRRTL allows duplicate port names, previously had verifier that needed to be rolled back: https://github.com/llvm/circt/pull/3976 .

I don't think this has been true for a while as the FIRRTL spec states how to handle the collisions (and provides an example of this exact case): https://github.com/chipsalliance/firrtl-spec/blob/main/spec.md#the-scalarized-convention The problem is that the type lowering passes don't properly implement the spec. The second `a_b` should be `a_b_0` and CIRCT is relying on the Verilog emitter to do this uniquification when it should happen eagerly.

include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
seldridge: 👍 Good catch. Yes, this was incorrectly overly restrictive. Fixed in: c1de2b9ade.

lib/Dialect/FIRRTL/Transforms/SpecializeOption.cpp
seldridge: Good catch. The _real problem_ here is that `InstanceChoiceOp` doesn't support domains and it should. I'll add that.

What would be nice is if we had a common `OpInterface` the shared some of the commonality between these two.

seldridge: This is fixed in https://github.com/llvm/circt/pull/8843/commits/8465e1ed86a73f695b676748a97ce8cd322d2379.

dtzSiFive:
I meant FIRRTL dialect.  I think we're agreeing, FIRRTL language itself is
clear here.

I agree with your assessment and that's why I added the verifier 😁👍.
Might have been your idea, I'm not sure 😅.  I think you made the
clarification to the spec, thanks for that!

There was an argument that it's inappropriate to ask passes to worry about
uniquing names (or providing them at all) that seems worth a consideration
but generally this is surprising behavior at the very least.

Anyway until that's resolved there are concerns about round tripping if
using their names, depending how they're grabbed (their ssa names should
roundtrip, so if this works the same way we're good, I haven't checked).

seldridge:
Looking at that example again, the error seems to be due to something unrelated to the FIRRTL Dialect name collision. This is actually just dropping the domain information of the domain port. This will trip the verifier, but it looks unrelated to naming.

seldridge:
> Looking at that example again, the error seems to be due to something unrelated to the FIRRTL Dialect name collision. This is actually just dropping the domain information of the domain port. This will trip the verifier, but it looks unrelated to naming.

Fixed in: 84f502f8c A port insertion utility was dropping the `FlatSymbolRefAttr` from a Domain type port when it should have passed it through. That commit also fixes issues identified with port insertion needing to _not_ generate `<<NULL ATTRIBUTE>>` for ports. There's a messy discontinuity between the storage for `PortInfo` which tracks "no domains" as a null pointer and the underlying operation storage for domains which needs to be an empty `ArrayAttr` for this.

test/Dialect/FIRRTL/parse-basic.fir
```diff
@@ -2191,3 +2191,34 @@ circuit NullaryCat:
     input a: UInt<1>
     ; CHECK: firrtl.cat : () -> !firrtl.uint<0>
     node b = cat()
+
+;// -----
+FIRRTL version 5.1.0
```
uenoku, line 2196: Just a question but what FIRRTL version do you target for domain?

uenoku (approved):
The IR design looks solid and good to me, great work! The change looks already large so I'm fine with merging the current state and incrementally fixing up remaining things. 

In the follow could you add tests for Transforms to make sure these passes don't drop domain info? Also LowerTypes and LowerSignature could insert ports so the currently IR will be broken when there is a domain port in the middle of the ports. Could you open tracking issues for that?

test/Dialect/FIRRTL/parse-basic.fir
seldridge: I'd target this for 6.0. However, this is currently `nextFIRVersion` which is 5.1.0. I'll deal with updating all this once the FIRRTL spec version solidifies.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: Whoops. Removed in the rebase. My editor helpfully added this...

seldridge:
> In the follow could you add tests for Transforms to make sure these passes don't drop domain info? Also LowerTypes and LowerSignature could insert ports so the currently IR will be broken when there is a domain port in the middle of the ports. Could you open tracking issues for that?

Good idea. Right now, this is only suitable for `firtool -parse-only`, `circt-opt`, and `circt-translate -export-firrtl`. Tracking issue to fix the bugs with insertion/deletion (and thereby get this working with the rest of the pipeline) here: https://github.com/llvm/circt/issues/8906

PR #8846, opened 2025-08-12
[SV] Remove extract-test-code

rwy7:
Remove the extract test code feature from firtool, which has been deprecated in favour of using bound layers.


tools/circt-synth/circt-synth.cpp
```diff
@@ -183,11 +183,6 @@ static void partiallyLegalizeCombToAIG(SmallVectorImpl<std::string> &ops) {
 
 // Add a default synthesis pipeline and analysis.
 static void populateCIRCTSynthPipeline(PassManager &pm) {
-  // ExtractTestCode is used to move verification code from design to
-  // remove registers/logic used only for verification.
-  pm.addPass(sv::createSVExtractTestCodePass(
```
rwy7: @uenoku please take a look and let me know if you are still relying on ETC :)

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
seldridge, line 1: Can the PR also delete the annotations in `AnnotationDetails.h` and their handling in the `LowerAnnotations` pass?

tools/circt-synth/circt-synth.cpp
uenoku: I'm fine as long as they are removed from designs so it's all OK! Thank you for checking!

mikeurbach (approved):
Good to see this finally go 🫡 

LGTM modulo the other open comments.

docs/Dialects/FIRRTL/FIRRTLAnnotations.md
seldridge: I think we found that there were some usages of these annotations and we need to delay removing these for now.

PR #8847, opened 2025-08-13
[HW] Accept most reductions even if module size increases

fabianschuiki:
The HW reductions that replace values with constants and remove module ports are beneficial even when they increase the size of the MLIR text. Removing ports and replacing values with zero removes dependencies in the IR, making it easier to apply other reductions.

Make these reductions return true from `acceptSizeIncrease` to have the reducer accept the result even if the module size increases.


fabianschuiki:
Yeah good point, the port reduction should always reduce the module size. I marked them here as well since they are always beneficial, so accepting them should not be contingent on module size.

lib/Dialect/HW/HWReductions.cpp
```diff
@@ -115,6 +116,8 @@ struct HWOperandForwarder : public Reduction {
 struct HWConstantifier : public Reduction {
   void matches(Operation *op,
                llvm::function_ref<void(uint64_t, uint64_t)> addMatch) override {
+    if (op->hasTrait<OpTrait::ConstantLike>())
```
dtzSiFive, line 119: --> don't try to constant-ify something already a constant? ✅

dtzSiFive (approved):

lib/Dialect/HW/HWReductions.cpp
fabianschuiki: Exactly 😁

PR #8848, opened 2025-08-13
[ImportVerilog] Add support for %f/%F format specifier

liamslj13:
This PR aims to add a FormatRealOp, similar to FormatIntOp, to the Moore dialect. I have tested it very briefly and will continue working on it tomorrow ~~but it should work okay in its current state (?)~~ (it did not). Opening as draft just to be safe for now.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -1617,17 +1617,11 @@ def FormatRealOp : MooreOp<"fmt.real", [Pure]> {
   }];
   let arguments = (ins
   RealType:$value,
-  RealFormatAttr:$format,
-  I32Attr:$width,
-  IntAlignAttr:$alignment,
-  IntPaddingAttr:$padding
+  RealFormatAttr:$format
   );
   let results = (outs FormatStringType:$result);
   let assemblyFormat = [{
       $format $value `,`
-      `width` $width `,`
-      `align` $alignment `,`
-      `pad` $padding
       attr-dict `:` type($value)
   }];
 }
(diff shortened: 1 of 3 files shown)


liamslj13:
At the moment formatting specifiers `%e`/`%E` and `%g`/`%G` from `IEEE 1800-2017 § 21.2.1.2 "Format specifications"`'s real numbers section are coupled in with `%f`/`%F`. Errors such as `unsupported format specifier %` should not be thrown for any real number specifiers, but behavior is not fully fleshed out yet.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1599,6 +1599,39 @@ def FormatIntOp : MooreOp<"fmt.int", [Pure]> {
...
+  Format a real value as a string according to the specified format.
+
+  See IEEE 1800-2017 § 21.2.1.2 "Format specifications".
+  }];
+  let arguments = (ins
+  RealType:$value,
+  RealFormatAttr:$format,
+  I32Attr:$width,
+  IntAlignAttr:$alignment,
+  IntPaddingAttr:$padding
```
fabianschuiki: Since we aren't really populating width, alignment, and padding yet, could we just delete them for now? That would make the code simpler and easier to extend. And in a second iteration later on, we could add supprot for all the formatting options for `%f` and friends, and add the necessary fields. I suspect we'll end up with a `width` again, but it might be optional, alongside a number of significant digits. Might be worth starting out easy first.

liamslj13: I think this is a good idea too. I've added a comment just to remind me or someone else taking a look for later.

liamslj13:
Thank you! I pushed the fix you suggested, just let me know if anything else comes up 👍

fabianschuiki:
Nice, thanks a lot!

PR #8850, opened 2025-08-13
[ci] Bump static build stack to 16MiB

seldridge:
It looks like 8MiB is not enough.  Apply exponential backoff and see if
this helps.

Possible fix for #8595.

Building the static binary here: https://github.com/llvm/circt/actions/runs/16944828793


dtzSiFive (approved):
Let's give this a go, sure!

seldridge:
It appears that we do not need this. It didn't fix the issue identified. Instead, we should figure out how to build the static binary with mimalloc.

PR #8852, opened 2025-08-14
[Feat][circt-opt] Add pass `--convert-core-to-fsm`

AtticusKuhn:
This PR adds a pass to `circt-opt` called `--convert-core-to-fsm`. This pass converts code written in RTL (`comb` + `seq`) and converts it to the `fsm` dialect.

It recognises and extracts FSM structure from an RTL netlist. It has similar functionality to `fsm_extract` in yosys:
https://yosyshq.readthedocs.io/projects/yosys/en/0.46/cmd/fsm.html

### DIFF
--- a/include/circt/Conversion/CoreToFSM.h
+++ b/include/circt/Conversion/CoreToFSM.h
@@ -1,0 +1,28 @@
+//===- CoreToFSM.h - Core to FSM conversions --------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CIRCT_CONVERSION_CORETOFSM_H
+#define CIRCT_CONVERSION_CORETOFSM_H
+
+#include <memory>
+
+namespace mlir {
+class Pass;
+class ModuleOp;
+} // namespace mlir
+
+namespace circt {
+
+std::unique_ptr<mlir::Pass> createConvertCoreToFSMPass();
+
+#define GEN_PASS_DECL_CONVERTCORETOFSM
+#include "circt/Conversion/Passes.h.inc"
+
+} // namespace circt
+
+#endif // CIRCT_CONVERSION_CORETOFSM_H
--- a/lib/Conversion/CoreToFSM/CoreToFSM.cpp
+++ b/lib/Conversion/CoreToFSM/CoreToFSM.cpp
@@ -1,0 +1,933 @@
+//===- CoreToFSM.cpp - Convert Core to FSM Dialect ------------------------===//
+#include "circt/Conversion/CoreToFSM.h"
+#include "circt/Dialect/Comb/CombDialect.h"
+#include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/FSM/FSMOps.h"
+#include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/Seq/SeqOps.h"
+#include "circt/Dialect/Seq/SeqTypes.h"
+#include "circt/Support/BackedgeBuilder.h"
+#include "circt/Support/LLVM.h"
+#include "mlir/IR/Block.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/IRMapping.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/Value.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/LogicalResult.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include "circt/Conversion/CoreToFSM.h"
+#include "mlir/Analysis/TopologicalSortUtils.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Passes.h"
+#include <boost/pending/disjoint_sets.hpp>
+#include <memory>
+#include <string>
+#include <vector>
+namespace circt {
+#define GEN_PASS_DEF_CONVERTCORETOFSM
+#include "circt/Conversion/Passes.h.inc"
+} // namespace circt
+
+using namespace mlir;
+using namespace circt;
+using namespace hw;
+using namespace fsm;
+
+namespace {
+
+// Forward declaration for our recursive helper function
+static void generateConcatenatedValues(
+    const std::vector<llvm::DenseSet<size_t>> &allOperandValues,
+    const std::vector<unsigned> &shifts,
+    llvm::DenseSet<size_t> &finalPossibleValues, size_t operandIdx,
+    size_t currentValue);
+
+static void getPossibleValues(llvm::DenseSet<size_t> &possibleValues, Value v) {
+  if (circt::hw::ConstantOp c =
+          dyn_cast_or_null<hw::ConstantOp>(v.getDefiningOp())) {
+    possibleValues.insert(c.getValueAttr().getValue().getZExtValue());
+    return;
+  }
+  if (circt::comb::MuxOp m = dyn_cast_or_null<comb::MuxOp>(v.getDefiningOp())) {
+    getPossibleValues(possibleValues, m.getTrueValue());
+    getPossibleValues(possibleValues, m.getFalseValue());
+    return;
+  }
+
+  if (circt::comb::ConcatOp concatOp =
+          dyn_cast_or_null<comb::ConcatOp>(v.getDefiningOp())) {
+    std::vector<llvm::DenseSet<size_t>> allOperandValues;
+    std::vector<unsigned> operandWidths;
+
+    for (Value operand : concatOp.getOperands()) {
+      llvm::DenseSet<size_t> operandPossibleValues;
+      getPossibleValues(operandPossibleValues, operand);
+
+      // It's crucial to handle the case where a sub-computation is too complex.
+      // If we can't determine specific values for an operand, we must
+      // pessimistically assume it can be any value its bitwidth allows.
+      IntegerType opType = dyn_cast<IntegerType>(operand.getType());
+      unsigned width = opType.getWidth();
+      if (operandPossibleValues.empty()) {
+        uint64_t numStates = 1ULL << width;
+        // Add a threshold to prevent combinatorial explosion on large unknown
+        // inputs.
+        if (numStates > 256) { // Heuristic threshold
+          // If the search space is too large, we abandon the analysis for this
+          // path. The outer function will fall back to its own full-range
+          // default.
+          return;
+        }
+        for (uint64_t i = 0; i < numStates; ++i)
+          operandPossibleValues.insert(i);
+      }
+
+      allOperandValues.push_back(operandPossibleValues);
+      operandWidths.push_back(width);
+    }
+
+    // The shift for operand `i` is the sum of the widths of operands `i+1` to
+    // `n-1`.
+    std::vector<unsigned> shifts(concatOp.getNumOperands(), 0);
+    for (int i = concatOp.getNumOperands() - 2; i >= 0; --i) {
+      shifts[i] = shifts[i + 1] + operandWidths[i + 1];
+    }
+
+    generateConcatenatedValues(allOperandValues, shifts, possibleValues, 0, 0);
+    return;
+  }
+
+  // --- Fallback Case ---
+  // If the operation is not recognized, assume all possible values for its
+  // bitwidth.
+  Operation *o = v.getDefiningOp();
+
+  IntegerType addrType = dyn_cast<IntegerType>(v.getType());
+  if (!addrType)
+    return; // Not an integer type we can analyze
+
+  unsigned bitWidth = addrType.getWidth();
+  // Again, use a threshold to avoid trying to enumerate 2^64 values.
+  if (bitWidth > 16) {
+    return;
+  }
+
+  uint64_t numRegStates = 1ULL << bitWidth;
+  for (size_t i = 0; i < numRegStates; i++) {
+    possibleValues.insert(i);
+  }
+  return;
+};
+
+/// @brief Recursively builds all possible concatenated integer values.
+/// @param allOperandValues Vector of sets, where each set contains the possible
+/// values for an operand.
+/// @param shifts The pre-calculated left-shift amount for each operand index.
+/// @param finalPossibleValues The output set to populate with final combined
+/// values.
+/// @param operandIdx The current operand index we are processing.
+/// @param currentValue The bitwise-OR'd value accumulated so far from previous
+/// operands.
+static void generateConcatenatedValues(
+    const std::vector<llvm::DenseSet<size_t>> &allOperandValues,
+    const std::vector<unsigned> &shifts,
+    llvm::DenseSet<size_t> &finalPossibleValues, size_t operandIdx,
+    size_t currentValue) {
+
+  // Base case: If we've processed all operands, the currentValue is complete.
+  if (operandIdx >= allOperandValues.size()) {
+    finalPossibleValues.insert(currentValue);
+    return;
+  }
+
+  // Recursive step: For each possible value of the current operand,
+  // combine it with the accumulated value and recurse for the next operand.
+  const auto &currentOperandPossibleValues = allOperandValues[operandIdx];
+  unsigned shift = shifts[operandIdx];
+
+  for (size_t val : currentOperandPossibleValues) {
+    // Combine the current operand's value by shifting it and ORing it.
+    size_t nextValue = currentValue | (val << shift);
+    generateConcatenatedValues(allOperandValues, shifts, finalPossibleValues,
+                               operandIdx + 1, nextValue);
+  }
+}
+
+static llvm::DenseMap<mlir::Value, int> intToRegMap(std::vector<seq::CompRegOp> v, int i){
+    llvm::DenseMap<mlir::Value, int> m;
+    // int i = 0;
+    // int width = 0;
+    for(size_t ci = 0; ci < v.size(); ci++){
+        seq::CompRegOp reg = v[ci];
+        int bits = reg.getType().getIntOrFloatBitWidth();
+        int v = i & ((1 << bits) - 1);
+        m[reg] = v;
+        i = i >> bits;
+        // i += m[reg] * 1ULL << width;
+        // width += (bits);
+    }
+    return m;
+    // return i;
+}
+static int regMapToInt(std::vector<seq::CompRegOp> v, llvm::DenseMap<mlir::Value, int> m){
+    int i = 0;
+    int width = 0;
+    for(size_t ci = 0; ci < v.size(); ci++){
+        seq::CompRegOp reg = v[ci];
+        i += m[reg] * 1ULL << width;
+        width += (reg.getType().getIntOrFloatBitWidth());
+    }
+    return i;
+}
+/// @brief Computes the Cartesian product of a list of sets.
+/// This function takes a vector of sets, where each set contains the possible
+/// values for a particular element (e.g., a register). It returns a set of
+/// vectors, where each vector represents one complete and unique combination of
+/// values, drawing one value from each of the input sets.
+/// @param valueSets A vector of DenseSets, each representing the possible
+/// values for one component of a state vector.
+/// @return A std::set of std::vectors, representing all possible complete
+/// state vectors.
+static std::set<std::vector<size_t>>
+calculateCartesianProduct(const std::vector<llvm::DenseSet<size_t>> &valueSets) {
+  std::set<std::vector<size_t>> product;
+  if (valueSets.empty()) {
+    // The Cartesian product of zero sets is a set containing one element:
+    // the empty tuple (represented here by an empty vector).
+    product.insert({});
+    return product;
+  }
+
+  // Initialize the product with the elements of the first set, each in its
+  // own vector.
+  for (size_t value : valueSets.front()) {
+    product.insert({value});
+  }
+
+  // Iteratively build the product. For each subsequent set, create a new
+  // temporary product by appending each of its elements to every combination
+  // already generated.
+  for (size_t i = 1; i < valueSets.size(); ++i) {
+    const auto &currentSet = valueSets[i];
+    if (currentSet.empty()) {
+      // The Cartesian product with an empty set results in an empty set.
+      return {};
+    }
+
+    std::set<std::vector<size_t>> newProduct;
+    for (const auto &existingVector : product) {
+      for (size_t newValue : currentSet) {
+        std::vector<size_t> newVector = existingVector;
+        newVector.push_back(newValue);
+        newProduct.insert(std::move(newVector));
+      }
+    }
+    product = std::move(newProduct);
+  }
+
+  return product;
+}
+
+static FrozenRewritePatternSet loadPatterns(MLIRContext &context){
+
+    RewritePatternSet patterns(&context);
+    // Collect canonicalization patterns from the dialects you are using.
+    // This is what the canonicalizer pass does internally.
+    for (auto *dialect : context.getLoadedDialects())
+      dialect->getCanonicalizationPatterns(patterns);
+    comb::ICmpOp::getCanonicalizationPatterns(patterns, &context);
+    comb::AndOp::getCanonicalizationPatterns(patterns, &context);
+    comb::XorOp::getCanonicalizationPatterns(patterns, &context);
+    comb::MuxOp::getCanonicalizationPatterns(patterns, &context);
+    comb::ConcatOp::getCanonicalizationPatterns(patterns,
+                                                &context);
+    comb::ExtractOp::getCanonicalizationPatterns(patterns,
+                                                 &context);
+    comb::AddOp::getCanonicalizationPatterns(patterns, &context);
+    comb::OrOp::getCanonicalizationPatterns(patterns, &context);
+    comb::MulOp::getCanonicalizationPatterns(patterns, &context);
+    hw::ConstantOp::getCanonicalizationPatterns(patterns,
+                                                &context);
+    fsm::TransitionOp::getCanonicalizationPatterns(patterns,
+                                                   &context);
+    fsm::StateOp::getCanonicalizationPatterns(patterns, &context);
+    fsm::MachineOp::getCanonicalizationPatterns(patterns,
+                                                &context);
+
+    FrozenRewritePatternSet frozenPatterns(std::move(patterns));
+    return frozenPatterns;
+}
+
+static void
+getReachableStates(llvm::DenseSet<size_t> &vistableStates,
+                   circt::hw::HWModuleOp moduleOp, size_t currentStateIndex,
+                   std::vector<seq::CompRegOp> registers,
+                   OpBuilder opBuilder, bool isInitialState) {
+
+  IRMapping mapping;
+  HWModuleOp clonedBody = llvm::dyn_cast<circt::hw::HWModuleOp>(
+      opBuilder.clone(*moduleOp, mapping));
+
+  llvm::DenseMap<mlir::Value, int> stateMap = intToRegMap(registers,currentStateIndex);
+  Operation *terminator = clonedBody.getBody().front().getTerminator();
+  circt::hw::OutputOp output = dyn_cast<circt::hw::OutputOp>(terminator);
+  int i = 0;
+  std::vector<mlir::Value> values;
+  std::vector<mlir::Type> types;
+  llvm::DenseMap<int, mlir::Value> regMap;
+
+  for (auto [originalRegValue, constStateValue] : stateMap) {
+
+    mlir::Value clonedRegValue = mapping.lookup(originalRegValue);
+    Operation *clonedRegOp = clonedRegValue.getDefiningOp();
+    circt::seq::CompRegOp reg = dyn_cast<circt::seq::CompRegOp>(clonedRegOp);
+    mlir::Type constantType = reg.getType();
+    IntegerAttr constantAttr =
+        opBuilder.getIntegerAttr(constantType, constStateValue);
+    opBuilder.setInsertionPoint(clonedRegOp);
+    circt::hw::ConstantOp otherStateConstant =
+        opBuilder.create<hw::ConstantOp>(reg.getLoc(), constantAttr);
+    values.push_back(reg.getInput());
+    types.push_back(reg.getType());
+    clonedRegValue.replaceAllUsesWith(otherStateConstant.getResult());
+    regMap[i] = originalRegValue;
+    reg.erase();
+    i++;
+  }
+  opBuilder.setInsertionPointToEnd(clonedBody.front().getBlock());
+  circt::hw::OutputOp newOutput =
+      opBuilder.create<circt::hw::OutputOp>(output.getLoc(), values);
+  output.erase();
+  // Collect canonicalization patterns from the dialects you are using.
+  // This is what the canonicalizer pass does internally.
+  FrozenRewritePatternSet frozenPatterns = loadPatterns(*moduleOp.getContext());
+
+  SmallVector<Operation *> opsToProcess;
+  clonedBody.walk([&](Operation *op) { opsToProcess.push_back(op); });
+
+  bool changed = false;
+  mlir::GreedyRewriteConfig config;
+  LogicalResult converged = mlir::applyOpPatternsGreedily(
+      opsToProcess, frozenPatterns, config, &changed);
+
+    std::vector<llvm::DenseSet<size_t>> pv;
+    for (size_t j = 0; j < newOutput.getNumOperands(); j++) {
+      llvm::DenseSet<size_t> possibleValues;
+
+      Value v = newOutput.getOperand(j);
+      getPossibleValues(possibleValues, v);
+      pv.push_back(possibleValues);
+    }
+    std::set<std::vector<size_t>> flipped =  calculateCartesianProduct(pv);
+    for(std::vector<size_t> v : flipped) {
+        llvm::DenseMap<mlir::Value, int> m;
+        for(int k = 0; k < v.size(); k++){
+            seq::CompRegOp r = registers[k];
+            m[r] = v[k];
+        }
+
+        int i = regMapToInt(registers, m);
+        vistableStates.insert(i);
+    }
+
+  clonedBody.erase();
+};
+
+
+
+
+
+// A converter class to handle the logic of converting a single hw.module.
+class HWModuleOpConverter {
+public:
+  HWModuleOpConverter(OpBuilder &builder, HWModuleOp moduleOp)
+      : moduleOp(moduleOp), opBuilder(builder) {}
+  LogicalResult run() {
+    llvm::SmallVector<circt::seq::CompRegOp> stateRegs;
+    llvm::SmallVector<circt::seq::CompRegOp> variableRegs;
+    moduleOp.walk([&](circt::seq::CompRegOp reg) {
+      if (reg.getName()->contains("state")) {
+        stateRegs.push_back(reg);
+      } else {
+        variableRegs.push_back(reg);
+      }
+    });
+    if (stateRegs.empty()) {
+      llvm::outs()
+          << "Cannot find state register in this FSM. You might need to "
+             "manually specify which registers are state registers.\n";
+      return mlir::success();
+    }
+    llvm::DenseMap<mlir::Value, size_t> regToIndexMap;
+    int regIndex = 0;
+    std::vector<seq::CompRegOp> registers;
+    for(seq::CompRegOp c : stateRegs){
+        regToIndexMap[c] = regIndex;
+        regIndex++;
+        registers.push_back(c);
+    }
+
+    llvm::DenseMap<size_t, circt::fsm::StateOp> stateToStateOp;
+    llvm::DenseMap<circt::fsm::StateOp, size_t> stateOpToState;
+    // gather async reset arguments to delete them from function type
+    llvm::DenseSet<size_t> asyncResetArguments;
+    auto regsInGroup = stateRegs;
+    mlir::Location loc = moduleOp.getLoc();
+    SmallVector<mlir::Type> inputTypes = moduleOp.getInputTypes();
+
+    // Create a new FSM machine with the current state.
+    auto resultTypes = moduleOp.getOutputTypes();
+    FunctionType machineType =
+        FunctionType::get(opBuilder.getContext(), inputTypes, resultTypes);
+    StringRef machineName = moduleOp.getName();
+
+    // int initialStateIndex = 0;
+    // std::vector<llvm::DenseMap<mlir::Value, int>> states =
+
+    //     enumerateStates(regsInGroup);
+    llvm::DenseMap<mlir::Value, int> initialStateMap;
+    for(seq::CompRegOp reg : moduleOp.getOps<seq::CompRegOp>()){
+        mlir::Value resetValue = reg.getResetValue();
+        circt::hw::ConstantOp definingConstant =
+            resetValue.getDefiningOp<circt::hw::ConstantOp>();
+        if (!definingConstant) {
+          reg->emitError(
+              "Cannot find defining constant for reset value of register: ");
+          return failure();
+        }
+        int resetValueInt = definingConstant.getValueAttr().getValue().getZExtValue();
+        initialStateMap[reg] = resetValueInt;
+    }
+    int initialStateIndex = regMapToInt(registers, initialStateMap);
+
+    std::string initialStateName = "state_" + std::to_string(initialStateIndex);
+
+    // Preserve argument and result names, which are stored as attributes.
+    SmallVector<NamedAttribute> machineAttrs;
+    if (auto argNames = moduleOp->getAttrOfType<ArrayAttr>("argNames"))
+      machineAttrs.emplace_back(opBuilder.getStringAttr("argNames"), argNames);
+    if (auto resNames = moduleOp->getAttrOfType<ArrayAttr>("resultNames"))
+      machineAttrs.emplace_back(opBuilder.getStringAttr("resNames"), resNames);
+
+    // The builder for fsm.MachineOp will create the body region and block
+    // arguments.
+    opBuilder.setInsertionPoint(moduleOp);
+    fsm::MachineOp machine = opBuilder.create<fsm::MachineOp>(
+        loc, machineName, initialStateName, machineType, machineAttrs);
+
+    OpBuilder::InsertionGuard guard(opBuilder);
+    opBuilder.setInsertionPointToStart(&machine.getBody().front());
+    llvm::DenseMap<circt::seq::CompRegOp, circt::fsm::VariableOp> variableMap;
+    for (circt::seq::CompRegOp varReg : variableRegs) {
+      ::mlir::TypedValue<::mlir::Type> initialValue = varReg.getResetValue();
+      circt::hw::ConstantOp definingConstant =
+          initialValue.getDefiningOp<circt::hw::ConstantOp>();
+      circt::fsm::VariableOp variableOp =
+          opBuilder.create<circt::fsm::VariableOp>(
+              varReg->getLoc(), varReg.getInput().getType(),
+              definingConstant.getValueAttr(), varReg.getName().value_or("a"));
+      variableMap[varReg] = variableOp;
+    }
+
+    // A valid machine needs at least its initial state defined.
+    FrozenRewritePatternSet frozenPatterns = loadPatterns(*moduleOp.getContext());
+
+    SetVector<int> reachableStates;
+    SmallVector<int> worklist;
+
+    worklist.push_back(initialStateIndex);
+    reachableStates.insert(initialStateIndex);
+    unsigned i = 0;
+    while (i < worklist.size()) {
+
+      int currentStateIndex = worklist[i++];
+
+     llvm::DenseMap<mlir::Value, int> stateMap = intToRegMap(registers, currentStateIndex);
+      
+
+      opBuilder.setInsertionPointToEnd(&machine.getBody().front());
+
+      circt::fsm::StateOp stateOp;
+
+      if (!stateToStateOp.contains(currentStateIndex)) {
+        stateOp = opBuilder.create<fsm::StateOp>(
+            loc, "state_" + std::to_string(currentStateIndex));
+        stateToStateOp.insert({currentStateIndex, stateOp});
+        stateOpToState.insert({stateOp, currentStateIndex});
+      } else {
+        stateOp = stateToStateOp.lookup(currentStateIndex);
+      }
+      mlir::Region &outputRegion = stateOp.getOutput();
+      mlir::Block *outputBlock = &outputRegion.front();
+      opBuilder.setInsertionPointToStart(outputBlock);
+      IRMapping mapping;
+      opBuilder.cloneRegionBefore(moduleOp.getModuleBody(), outputRegion,
+                                  outputBlock->getIterator(), mapping);
+      outputBlock->erase();
+
+      auto *terminator = outputRegion.front().getTerminator();
+      auto hwOutputOp = dyn_cast<hw::OutputOp>(terminator);
+      assert(hwOutputOp && "Expected terminator to be hw.OutputOp");
+
+      // Position the builder to insert the new terminator right before the
+      // old one.
+      OpBuilder::InsertionGuard stateGuard(opBuilder);
+      opBuilder.setInsertionPoint(hwOutputOp);
+
+      // Create the new fsm.OutputOp with the same operands.
+
+      opBuilder.create<fsm::OutputOp>(hwOutputOp.getLoc(),
+                                      hwOutputOp.getOperands());
+
+      // Erase the old terminator.
+      hwOutputOp.erase();
+
+      // Iterate through the state configuration to replace registers
+      // with constants.
+      for (auto &[originalRegValue, variableOp] : variableMap) {
+        Value clonedRegValue = mapping.lookup(originalRegValue);
+        Operation *clonedRegOp = clonedRegValue.getDefiningOp();
+        circt::seq::CompRegOp reg =
+            dyn_cast<circt::seq::CompRegOp>(clonedRegOp);
+        const auto res = variableOp.getResult();
+        clonedRegValue.replaceAllUsesWith(res);
+        reg.erase();
+      }
+      for (auto const &[originalRegValue, constStateValue] : stateMap) {
+        //  Find the cloned register's result value using the mapping.
+        Value clonedRegValue = mapping.lookup(originalRegValue);
+        assert(clonedRegValue && "Original register value not found in "
+                                 "mapping; this is an internal error.");
+        Operation *clonedRegOp = clonedRegValue.getDefiningOp();
+        
+        assert(clonedRegOp && "Cloned value must have a defining op.");
+        opBuilder.setInsertionPoint(clonedRegOp);
+        circt::seq::CompRegOp r = dyn_cast<circt::seq::CompRegOp>(clonedRegOp);
+        assert(r && "Must be a register.");
+        auto input = r.getInput();
+        TypedValue<IntegerType> registerReset = r.getReset();
+        if (registerReset) { // Ensure the register has a reset.
+          if (BlockArgument blockArg =
+                  mlir::dyn_cast<mlir::BlockArgument>(registerReset)) {
+            asyncResetArguments.insert(blockArg.getArgNumber());
+            // blockArg.dump();
+            ConstantOp falseConst = opBuilder.create<hw::ConstantOp>(
+                blockArg.getLoc(), clonedRegValue.getType(), 0);
+            blockArg.replaceAllUsesWith(falseConst.getResult());
+            // Check if this argument belongs to the top-level module block.
+          }
+          // also check for rst_ni case, because OpenTitan uses ni.
+          if (circt::comb::XorOp xorOp =
+                  registerReset.getDefiningOp<circt::comb::XorOp>()) {
+            if (xorOp.isBinaryNot()) {
+              mlir::Value rhs = xorOp.getOperand(0);
+              if (BlockArgument blockArg =
+                      mlir::dyn_cast<mlir::BlockArgument>(rhs)) {
+                asyncResetArguments.insert(blockArg.getArgNumber());
+                ConstantOp trueConst = opBuilder.create<hw::ConstantOp>(
+                    blockArg.getLoc(), blockArg.getType(), 1);
+                blockArg.replaceAllUsesWith(trueConst.getResult());
+              }
+            }
+          }
+        }
+        // Create the hw.constant operation for the specific state value.
+        auto constantOp = opBuilder.create<hw::ConstantOp>(
+            clonedRegValue.getLoc(), clonedRegValue.getType(), constStateValue);
+
+        // Replace all uses of the cloned register's result with the new
+        // constant.
+        clonedRegValue.replaceAllUsesWith(constantOp.getResult());
+
+        // Erase the now-dead cloned register operation.
+
+        clonedRegOp->erase();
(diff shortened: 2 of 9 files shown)


include/circt/Conversion/CoreToFSM.h
```diff
@@ -0,0 +1,28 @@
...
+namespace circt {
+
+std::unique_ptr<mlir::Pass> createConvertCoreToFSMPass();
+
+#define GEN_PASS_DECL_CONVERTCORETOFSM
+#include "circt/Conversion/Passes.h.inc"
+
+} // namespace circt
+
+#endif // CIRCT_CONVERSION_CORETOFSM_H
```
TaoBi22: Nit: missing EOF newline

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/LogicalResult.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include "circt/Conversion/CoreToFSM.h"
+#include "mlir/Analysis/TopologicalSortUtils.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Passes.h"
+#include <boost/pending/disjoint_sets.hpp>
```
TaoBi22: I don't think we have boost as a dependency in CIRCT. Can you use some of the set structures from the LLVM codebase instead?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+#include <memory>
+#include <string>
+#include <vector>
```
TaoBi22: We try to use llvm::SmallVectors instead of std::vectors, which would allow you to drop this include

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/LogicalResult.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include "circt/Conversion/CoreToFSM.h"
+#include "mlir/Analysis/TopologicalSortUtils.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Passes.h"
+#include <boost/pending/disjoint_sets.hpp>
+#include <memory>
```
TaoBi22: What's this used for?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+
+/// @brief Recursively builds all possible concatenated integer values.
+/// @param allOperandValues Vector of sets, where each set contains the possible
+/// values for an operand.
+/// @param shifts The pre-calculated left-shift amount for each operand index.
+/// @param finalPossibleValues The output set to populate with final combined
+/// values.
+/// @param operandIdx The current operand index we are processing.
+/// @param currentValue The bitwise-OR'd value accumulated so far from previous
+/// operands.
```
TaoBi22: I'm not sure what this syntax is, but I don't think we tend to use it in CIRCT. Just a /// function description should be fine here.

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+}
+/// @brief Computes the Cartesian product of a list of sets.
+/// This function takes a vector of sets, where each set contains the possible
+/// values for a particular element (e.g., a register). It returns a set of
+/// vectors, where each vector represents one complete and unique combination of
+/// values, drawing one value from each of the input sets.
+/// @param valueSets A vector of DenseSets, each representing the possible
+/// values for one component of a state vector.
+/// @return A std::set of std::vectors, representing all possible complete
+/// state vectors.
```
TaoBi22: As above

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -2950,9 +2950,67 @@ static void combineEqualityICmpWithXorOfConstant(ICmpOp cmpOp, XorOp xorOp,
...
+    circt::comb::ICmpOp eq2 = rewriter.create<circt::comb::ICmpOp>(
+        loc, circt::comb::ICmpPredicate::eq, y, b);
+    circt::comb::MuxOp newMux = rewriter.create<circt::comb::MuxOp>(
+        loc, mux.getCond(), eq1.getResult(), eq2.getResult());
+    Operation *o = newMux;
+    op.replaceAllUsesWith(o);
+    op.erase();
+    return llvm::success();
+  }
   // icmp 1, x -> icmp x, 1
```
TaoBi22, line 3101: As discussed offline, these might make more sense as patterns defined in CoreToFSM (and if we stick with adding them to CombFolds we'd probably want this to be a separate PR)

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+      if (reg.getName()->contains("state")) {
+        stateRegs.push_back(reg);
+      } else {
+        variableRegs.push_back(reg);
+      }
+    });
+    if (stateRegs.empty()) {
+      llvm::outs()
+          << "Cannot find state register in this FSM. You might need to "
+             "manually specify which registers are state registers.\n";
```
TaoBi22: Can this be an emitError instead of being piped to outs()?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+      return mlir::success();
```
TaoBi22: This looks like it should return failure?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+        clonedRegValue.replaceAllUsesWith(constantOp.getResult());
+
+        // Erase the now-dead cloned register operation.
+
+        clonedRegOp->erase();
+      }
+      mlir::GreedyRewriteConfig config;
+      // You could add more patterns here if needed.
+      // This function will apply folding and DCE, which is exactly what
+      // you need.
```
TaoBi22: Not sure these comments are necessary

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+        mlir::Block &newGuardBlock = guardRegion.front();
+        Operation *terminator = newGuardBlock.getTerminator();
+        hw::OutputOp hwOutputOp = dyn_cast<hw::OutputOp>(terminator);
+        assert(hwOutputOp && "Expected terminator to be hw.OutputOp");
+
+        // Position the builder to insert the new terminator right before
+        // the old one.
+
+        llvm::DenseMap<mlir::Value, int> toStateMap =
+            intToRegMap(registers, j); // states[j];
```
TaoBi22:
```suggestion
            intToRegMap(registers, j);
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,926 @@
...
+          circt::seq::CompRegOp r =
+              dyn_cast<circt::seq::CompRegOp>(clonedRegOp);
+
+          mlir::Value registerInput = r.getInput();
+          TypedValue<IntegerType> registerReset = r.getReset();
+          if (registerReset) { // Ensure the register has a reset.
+            if (BlockArgument blockArg =
+                    mlir::dyn_cast<mlir::BlockArgument>(registerReset)) {
+              // asyncResetArguments.insert(blockArg.getArgNumber());
+              // blockArg.dump();
```
TaoBi22:
```suggestion
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,954 @@
...
+      } else {
+        variableRegs.push_back(reg);
+      }
+    });
+    if (stateRegs.empty()) {
+      emitError(moduleOp.getLoc())
+          << "Cannot find state register in this FSM. You might need to "
+             "manually specify which registers are state registers.\n";
+      return mlir::failure();
+    }
```
AndreyVV-100: This analysis does not look reliable. Is it possible to replace it with an analogue of `fsm_detect` from yosys?

AtticusKuhn: Yes, that's true, but I think that to have modularity, it would be fine to split up the total pipeline into `fsm_detect` and `fsm_extract` (like Yosys does). So, theoretically there could be a pass `circt-opt --annotate-fsm-state-registers` that works in combination with `circt-opt --convert-core-to-fsm`. In the meantime, it's not too difficult for the user to manually annotate the FSM state registers in an RTL design.

TaoBi22: Maybe for now we could just require it to always have the name of the state register explicitly given by the user? That way we don't need to worry about users having it opaquely pick up a register they don't expect, and we can always later change it to a flag that will get it to look for an attribute

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,953 @@
+//===- CoreToFSM.cpp - Convert Core to FSM Dialect ------------------------===//
```
uenoku: nit: please add a header (`Part of the LLVM Project,..`)

include/circt/Conversion/CoreToFSM.h
```diff
@@ -0,0 +1,28 @@
+//===- CoreToFSM.h - Core to FSM conversions --------------------*- C++ -*-===//
```
luisacicolini:
```suggestion
//===- CoreToFSM.h - Core to FSM conversion pass --------------------*- C++ -*-===//
```
NIT to be consistent with the CIRCT conventions

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
+//===- CoreToFSM.cpp - Convert Core Dialects (Seq + Comb) to FSM Dialect---===//
```
luisacicolini:
```suggestion
//===- CoreToFSM.cpp - Convert Core Dialects (HW + Seq + Comb) to FSM Dialect---===//
```
other nit

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/LogicalResult.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include "circt/Conversion/CoreToFSM.h"
```
TaoBi22: Repeated include (I'm surprised clang-format didn't catch this - I suspect it's due to having a space in the middle of the include list, which it would also be nice to drop)

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+// Forward declaration for our recursive helper function
+static void generateConcatenatedValues(
+    const llvm::SmallVector<llvm::DenseSet<size_t>> &allOperandValues,
+    const llvm::SmallVector<unsigned> &shifts,
+    llvm::DenseSet<size_t> &finalPossibleValues, size_t operandIdx,
+    size_t currentValue);
+
+static void getPossibleValues(llvm::DenseSet<size_t> &possibleValues, Value v) {
+  if (circt::hw::ConstantOp c =
```
TaoBi22: nit:
```suggestion
  if (ConstantOp c =
```

TaoBi22: actually can probably just be an auto?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+          dyn_cast_or_null<hw::ConstantOp>(v.getDefiningOp())) {
+    possibleValues.insert(c.getValueAttr().getValue().getZExtValue());
+    return;
+  }
+  if (circt::comb::MuxOp m = dyn_cast_or_null<comb::MuxOp>(v.getDefiningOp())) {
```
TaoBi22: As above

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+namespace {
+
+// Forward declaration for our recursive helper function
+static void generateConcatenatedValues(
+    const llvm::SmallVector<llvm::DenseSet<size_t>> &allOperandValues,
+    const llvm::SmallVector<unsigned> &shifts,
+    llvm::DenseSet<size_t> &finalPossibleValues, size_t operandIdx,
+    size_t currentValue);
+
+static void getPossibleValues(llvm::DenseSet<size_t> &possibleValues, Value v) {
```
TaoBi22: Nit: the name of this function makes it sound like a getter, but it's a void. Maybe addPossibleValues or something would make more sense?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+      unsigned width = opType.getWidth();
+      if (operandPossibleValues.empty()) {
+        uint64_t numStates = 1ULL << width;
+        // Add a threshold to prevent combinatorial explosion on large unknown
+        // inputs.
+        if (numStates > 256) { // Heuristic threshold
+          // If the search space is too large, we abandon the analysis for this
+          // path. The outer function will fall back to its own full-range
+          // default.
+          return;
```
TaoBi22, line 102: Would it make sense to emit a warning/error here?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+};
+
+static bool isConstant(mlir::Value value) {
+  Operation *getDefiningOp = value.getDefiningOp();
+  if (!getDefiningOp) {
+    return false;
+  }
+  if (circt::hw::ConstantOp constantOp =
+          mlir::dyn_cast<circt::hw::ConstantOp>(getDefiningOp)) {
+    return true;
```
TaoBi22: this can be replaced with `isa<ConstantOp>` to avoid an unnecessary cast

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+  }
+  if (circt::comb::MuxOp muxOp =
+          mlir::dyn_cast<circt::comb::MuxOp>(getDefiningOp)) {
+    return isConstant(muxOp.getTrueValue()) &&
+           isConstant(muxOp.getFalseValue());
+  }
```
TaoBi22: If it returns true in this case, I guess isConstant might not be the most informative name for the function, since this isn't a constant term?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+    mlir::Value b = op.getRhs();
+    Location loc = op.getLoc();
+    circt::comb::ICmpOp eq1 = rewriter.create<circt::comb::ICmpOp>(
+        loc, circt::comb::ICmpPredicate::eq, x, b);
+    circt::comb::ICmpOp eq2 = rewriter.create<circt::comb::ICmpOp>(
+        loc, circt::comb::ICmpPredicate::eq, y, b);
+    circt::comb::MuxOp newMux = rewriter.create<circt::comb::MuxOp>(
+        loc, mux.getCond(), eq1.getResult(), eq2.getResult());
+    Operation *o = newMux;
+    op.replaceAllUsesWith(o);
```
TaoBi22: Looks like we can probably sink newMux in directly? (Maybe with a cast)

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+    mlir::Value b = op.getLhs();
+    Location loc = op.getLoc();
+    circt::comb::ICmpOp eq1 = rewriter.create<circt::comb::ICmpOp>(
+        loc, circt::comb::ICmpPredicate::eq, x, b);
+    circt::comb::ICmpOp eq2 = rewriter.create<circt::comb::ICmpOp>(
+        loc, circt::comb::ICmpPredicate::eq, y, b);
+    circt::comb::MuxOp newMux = rewriter.create<circt::comb::MuxOp>(
+        loc, mux.getCond(), eq1.getResult(), eq2.getResult());
+    Operation *o = newMux;
+    op.replaceAllUsesWith(o);
```
TaoBi22: As above

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+    op.erase();
+    return llvm::success();
+  }
+  return llvm::failure();
+}
+/// Recursively builds all possible concatenated integer values.
+static void generateConcatenatedValues(
```
TaoBi22, line 229: It looks like this wouldn't be too hard to do this iteratively instead of recursively - it's probably worth doing iteratively if possible to avoid any risk of stack overflow

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+                               operandIdx + 1, nextValue);
+  }
+}
+
+static llvm::DenseMap<mlir::Value, int>
+intToRegMap(llvm::SmallVector<seq::CompRegOp> v, int i) {
+  llvm::DenseMap<mlir::Value, int> m;
+  for (size_t ci = 0; ci < v.size(); ci++) {
+    seq::CompRegOp reg = v[ci];
+    int bits = reg.getType().getIntOrFloatBitWidth();
```
TaoBi22, line 266: Do we check anywhere that the reg's type is integer (i.e. it has a bitwidth)? We definitely don't need to support designs where that's not the case, but we should error out with a message

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+// A converter class to handle the logic of converting a single hw.module.
+class HWModuleOpConverter {
+public:
+  HWModuleOpConverter(OpBuilder &builder, HWModuleOp moduleOp)
+      : moduleOp(moduleOp), opBuilder(builder) {}
+  LogicalResult run() {
+    llvm::SmallVector<circt::seq::CompRegOp> stateRegs;
+    llvm::SmallVector<circt::seq::CompRegOp> variableRegs;
+    moduleOp.walk([&](circt::seq::CompRegOp reg) {
+      if (reg.getName()->contains("state")) {
```
TaoBi22: Re: the conversation above it would be nice to make this an input to the pass if possible rather than this hardcoded name check, or an attribute on the state register perhaps

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+    opBuilder.setInsertionPointToStart(&machine.getBody().front());
+    llvm::DenseMap<circt::seq::CompRegOp, circt::fsm::VariableOp> variableMap;
+    for (circt::seq::CompRegOp varReg : variableRegs) {
+      ::mlir::TypedValue<::mlir::Type> initialValue = varReg.getResetValue();
+      circt::hw::ConstantOp definingConstant =
+          initialValue.getDefiningOp<circt::hw::ConstantOp>();
+      circt::fsm::VariableOp variableOp =
+          opBuilder.create<circt::fsm::VariableOp>(
+              varReg->getLoc(), varReg.getInput().getType(),
+              definingConstant.getValueAttr(), varReg.getName().value_or("a"));
```
TaoBi22: Might be worth replacing "a" with something more informative

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+    FrozenRewritePatternSet frozenPatterns =
+        loadPatterns(*moduleOp.getContext());
+
+    SetVector<int> reachableStates;
+    SmallVector<int> worklist;
+
+    worklist.push_back(initialStateIndex);
+    reachableStates.insert(initialStateIndex);
+    unsigned i = 0;
+    while (i < worklist.size()) {
```
TaoBi22: nit: can this just be 
```suggestion
    for (auto currentStateIndex: worklist) {
```
?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+      mlir::Block *outputBlock = &outputRegion.front();
+      opBuilder.setInsertionPointToStart(outputBlock);
+      IRMapping mapping;
+      opBuilder.cloneRegionBefore(moduleOp.getModuleBody(), outputRegion,
+                                  outputBlock->getIterator(), mapping);
+      outputBlock->erase();
+
+      auto *terminator = outputRegion.front().getTerminator();
+      auto hwOutputOp = dyn_cast<hw::OutputOp>(terminator);
+      assert(hwOutputOp && "Expected terminator to be hw.OutputOp");
```
TaoBi22:
```suggestion
      assert(hwOutputOp && "Expected terminator to be hw.output op");
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+            dyn_cast<circt::seq::CompRegOp>(clonedRegOp);
+        const auto res = variableOp.getResult();
+        clonedRegValue.replaceAllUsesWith(res);
+        reg.erase();
+      }
+      for (auto const &[originalRegValue, constStateValue] : stateMap) {
+        //  Find the cloned register's result value using the mapping.
+        Value clonedRegValue = mapping.lookup(originalRegValue);
+        assert(clonedRegValue && "Original register value not found in "
+                                 "mapping; this is an internal error.");
```
TaoBi22: Probably fine to leave off the 'this is an internal error', we usually use assertions for internal errors and explicit errors for user error/invalid or unsupported input

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+        assert(clonedRegOp && "Cloned value must have a defining op.");
+        opBuilder.setInsertionPoint(clonedRegOp);
+        circt::seq::CompRegOp r = dyn_cast<circt::seq::CompRegOp>(clonedRegOp);
+        assert(r && "Must be a register.");
+        TypedValue<IntegerType> registerReset = r.getReset();
+        if (registerReset) { // Ensure the register has a reset.
+          if (BlockArgument blockArg =
+                  mlir::dyn_cast<mlir::BlockArgument>(registerReset)) {
+            asyncResetArguments.insert(blockArg.getArgNumber());
+            // blockArg.dump();
```
TaoBi22:
```suggestion
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+        // Replace all uses of the cloned register's result with the new
+        // constant.
+        clonedRegValue.replaceAllUsesWith(constantOp.getResult());
+
+        // Erase the now-dead cloned register operation.
+
+        clonedRegOp->erase();
+      }
+      mlir::GreedyRewriteConfig config;
```
TaoBi22: super-nit: can we sink this to its use?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+      bool changed = false;
+      LogicalResult converged = mlir::applyOpPatternsGreedily(
+          opsToProcess, patterns, config, &changed);
+      opBuilder.setInsertionPoint(stateOp);
+      if (!sortTopologically(&outputRegion.front())) {
+        moduleOp.emitError("could not resolve cycles in module");
+        return failure();
+      }
+      mlir::Region &transitionRegion = stateOp.getTransitions();
+      llvm::DenseSet<size_t> vistableStates;
```
TaoBi22:
```suggestion
      llvm::DenseSet<size_t> visitableStates;
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+        opBuilder.setInsertionPointToStart(&guardBlock);
+        IRMapping mapping;
+        opBuilder.cloneRegionBefore(moduleOp.getModuleBody(), guardRegion,
+                                    guardBlock.getIterator(), mapping);
+        guardBlock.erase();
+        mlir::Block &newGuardBlock = guardRegion.front();
+        Operation *terminator = newGuardBlock.getTerminator();
+        hw::OutputOp hwOutputOp = dyn_cast<hw::OutputOp>(terminator);
+        assert(hwOutputOp && "Expected terminator to be hw.OutputOp");
```
TaoBi22:
```suggestion
        assert(hwOutputOp && "Expected terminator to be hw.output op");
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+            seq::CompRegOp reg = dyn_cast<seq::CompRegOp>(clonedRegOp);
+            opBuilder.setInsertionPointToStart(&newActionBlock);
+            fsm::UpdateOp updateOp = opBuilder.create<fsm::UpdateOp>(
+                reg.getLoc(), variableOp, reg.getInput());
+            const mlir::Value res = variableOp.getResult();
+            clonedRegValue.replaceAllUsesWith(res);
+            reg.erase();
+          }
+          Operation *terminator = actionRegion.back().getTerminator();
+          hw::OutputOp hwOutputOp = dyn_cast<hw::OutputOp>(terminator);
```
TaoBi22: This cast looks unnecessary

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+                constStateValue);
+
+            // Replace all uses of the cloned register's result with the new
+            // constant.
+            clonedRegValue.replaceAllUsesWith(constantOp.getResult());
+
+            clonedRegOp->erase();
+          }
+
+          // delete the arguments from the output block
```
TaoBi22: This comment looks like it's describing something other than what follows it

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+        mlir::GreedyRewriteConfig config2;
+        config2.setScope(&stateOp.getTransitions());
+        mlir::applyOpPatternsGreedily(transitionOps, frozenPatterns, config2,
+                                      &changed);
+
+        if (failed(converged)) {
+          stateOp.emitError("Failed to canonicalize the generated state op");
+          return failure();
+        }
```
TaoBi22: It looks like we can probably catch this after the first call directly?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+        for (TransitionOp transition :
+             stateOp.getTransitions().getOps<TransitionOp>()) {
+          StateOp nextState = transition.getNextStateOp();
+          int nextStateIndex = stateOpToState.lookup(nextState);
+          hw::ConstantOp guardConst =
+              transition.getGuardReturn()
+                  .getOperand()
+                  .getDefiningOp<circt::hw::ConstantOp>();
```
TaoBi22: is this guaranteed not to return null?

AtticusKuhn:
Thanks for the feedback @TaoBi22 . I'm somewhat busy currently, but I'll be sure to work through and address your feedback in the coming days.

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+
+    for (Value operand : concatOp.getOperands()) {
+      llvm::DenseSet<size_t> operandPossibleValues;
+      addPossibleValues(operandPossibleValues, operand);
+
+      // It's crucial to handle the case where a sub-computation is too complex.
+      // If we can't determine specific values for an operand, we must
+      // pessimistically assume it can be any value its bitwidth allows.
+      auto opType = dyn_cast<IntegerType>(operand.getType());
+      unsigned width = opType.getWidth();
```
TaoBi22, line 89: I think we'll need an error here if the cast to an IntegerType fails

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+/// Checks if a value is a constant or a tree of muxes with constant leaves.
+static bool isConstantLike(mlir::Value value) {
+  Operation *definingOp = value.getDefiningOp();
+  if (!definingOp)
+    return false;
+  if (isa<hw::ConstantOp>(definingOp))
+    return true;
+  if (auto muxOp = dyn_cast<MuxOp>(definingOp))
+    return isConstantLike(muxOp.getTrueValue()) &&
+           isConstantLike(muxOp.getFalseValue());
```
TaoBi22: It looks to me like this might throw a stack overflow given something in the core input like `%2 = comb.mux %0, %1, %2`

TaoBi22: (which isn't necessarily a useful input, but it is feasible and would be nice to catch)

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+  return false;
+}
+
+LogicalResult pushIcmp(ICmpOp op, PatternRewriter &rewriter) {
```
TaoBi22, line 190: A comment to explain what this method is doing would be great

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+      llvm::dyn_cast<HWModuleOp>(opBuilder.clone(*moduleOp, mapping));
+
+  llvm::DenseMap<mlir::Value, int> stateMap =
+      intToRegMap(registers, currentStateIndex);
+  Operation *terminator = clonedBody.getBody().front().getTerminator();
+  auto output = dyn_cast<hw::OutputOp>(terminator);
+  int i = 0;
+  llvm::SmallVector<mlir::Value> values;
+  llvm::SmallVector<mlir::Type> types;
+  llvm::DenseMap<int, mlir::Value> regMap;
```
TaoBi22: Nit: here (and in other places), there are a lot of `mlir::` prefixes, but there's a `using namespace mlir` directive at the start of the file

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+
+  for (auto [originalRegValue, constStateValue] : stateMap) {
+
+    mlir::Value clonedRegValue = mapping.lookup(originalRegValue);
+    Operation *clonedRegOp = clonedRegValue.getDefiningOp();
+    auto reg = dyn_cast<seq::CompRegOp>(clonedRegOp);
+    mlir::Type constantType = reg.getType();
```
TaoBi22: Is this guaranteed to be a successful cast? If not, we should check and throw an error if it fails; if it's guaranteed to be safe, we can just use `cast`

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+      : moduleOp(moduleOp), opBuilder(builder), stateRegNames(stateRegNames) {}
+  LogicalResult run() {
+    llvm::SmallVector<seq::CompRegOp> stateRegs;
+    llvm::SmallVector<seq::CompRegOp> variableRegs;
+    bool hasNonIntegerReg = false;
+    moduleOp.walk([&](seq::CompRegOp reg) {
+      // Check that the register type is an integer.
+      if (!isa<IntegerType>(reg.getType())) {
+        reg.emitError("FSM extraction only supports integer-typed registers");
+        hasNonIntegerReg = true;
```
TaoBi22: Nit, but you should be able to avoid the need for this intermediate variable by returning a [WalkResult](https://mlir.llvm.org/doxygen/classmlir_1_1WalkResult.html) and checking if the walk was interrupted

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+        stateRegs.push_back(reg);
+      } else {
+        variableRegs.push_back(reg);
+      }
+    });
+    if (hasNonIntegerReg)
+      return mlir::failure();
+    if (stateRegs.empty()) {
+      emitError(moduleOp.getLoc())
+          << "Cannot find state register in this FSM. Use the --state-regs "
```
TaoBi22: Super-nit just because of the syntax of specifying options for passes in circt-opt and the fact it could be being called in a pass pipeline defined in C++:
```suggestion
          << "Cannot find state register in this FSM. Use the state-regs "
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+    llvm::SmallVector<seq::CompRegOp> registers;
+    for (seq::CompRegOp c : stateRegs) {
+      regToIndexMap[c] = regIndex;
+      regIndex++;
+      registers.push_back(c);
+    }
+
+    llvm::DenseMap<size_t, StateOp> stateToStateOp;
+    llvm::DenseMap<StateOp, size_t> stateOpToState;
+    // gather async reset arguments to delete them from function type
```
TaoBi22: I'm not sure I understand what this comment means. Are async resets just thrown away? (If so it's probably worth returning an error rather than silently ignoring them, since we're then changing the circuit's semantics)

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+                          "variable register");
+        return failure();
+      }
+      auto variableOp = opBuilder.create<VariableOp>(
+          varReg->getLoc(), varReg.getInput().getType(),
+          definingConstant.getValueAttr(), varReg.getName().value_or("var"));
+      variableMap[varReg] = variableOp;
+    }
+
+    // A valid machine needs at least its initial state defined.
```
TaoBi22: Nit: This comment doesn't seem relevant to what directly follows it?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+          if (nextStateIsReachable &&
+              !reachableStates.contains(nextStateIndex)) {
+            worklist.push_back(nextStateIndex);
+            reachableStates.insert(nextStateIndex);
+          }
+        }
+      }
+    }
+
+    SmallVector<StateOp> statesToErase;
```
TaoBi22, line 885: A comment to explain what's happening here would be helpful

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,934 @@
...
+
+    // First, collect the states that need to be erased.
+    for (StateOp stateOp : machine.getOps<StateOp>()) {
+      if (!stateOp.getOutputOp()) {
+        statesToErase.push_back(stateOp);
+      }
+    }
+
+    // Now, erase them in a separate loop.
```
TaoBi22: How come this needs to happen in a separate loop?

test/Conversion/CoreToFSM/basic.mlir
```diff
@@ -0,0 +1,33 @@
...
+    %add = comb.concat %c0_i1, %inp : i1, i1
+    %next_state = comb.add %state, %add : i2
+    %is_2 = comb.icmp eq %state, %c2_i2 : i2
+    hw.output %is_2 : i1
+}
+
+// CHECK: fsm.state @state_0 output {
+// CHECK:   %[[FALSE:.*]] = hw.constant false
+// CHECK:   fsm.output %[[FALSE]] : i1
+// CHECK: } transitions {
```
TaoBi22, line 26: Can we check the transitions are as expected?

test/Conversion/CoreToFSM/counter.mlir
```diff
@@ -0,0 +1,33 @@
...
+    %c2_i2 = hw.constant 2 : i2
+    %true = hw.constant true
+    %state = seq.compreg name "state" %next_state, %clk reset %rst, %c0_i2 : i2
+    %add = comb.concat %c0_i1, %inp: i1, i1
+    %next_state = comb.add %state, %add : i2
+    %is_2 = comb.icmp eq %state, %c2_i2 : i2
+    hw.output %is_2 : i1
+}
+
+// Verify that states are created
```
TaoBi22: As above, I think we want to check more than just creation of states - we for sure want to know that transitions are created correctly too

test/Conversion/CoreToFSM/counter.mlir
```diff
@@ -0,0 +1,33 @@
+// RUN: circt-opt -convert-core-to-fsm %s | FileCheck %s
+
+// Test RTL-to-FSM conversion with a counter that increments based on input
```
TaoBi22: Could we have a test that checks variables/actions too?

test/Conversion/CoreToFSM/state-regs-option.mlir
```diff
@@ -0,0 +1,35 @@
...
+    %true = hw.constant true
+    // Register named "my_fsm_reg" - does not contain "state" but is specified via --state-regs
+    %my_fsm_reg = seq.compreg name "my_fsm_reg" %next_val, %clk reset %rst, %c0_i2 : i2
+    %add = comb.concat %c0_i1, %inp : i1, i1
+    %next_val = comb.add %my_fsm_reg, %add : i2
+    %is_2 = comb.icmp eq %my_fsm_reg, %c2_i2 : i2
+    hw.output %is_2 : i1
+}
+
+// Verify that states are created correctly
```
TaoBi22: As above

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,988 @@
...
+      regToIndexMap[c] = regIndex;
+      regIndex++;
+      registers.push_back(c);
+    }
+
+    llvm::DenseMap<size_t, StateOp> stateToStateOp;
+    llvm::DenseMap<StateOp, size_t> stateOpToState;
+    // Collect async reset arguments to exclude from the FSM's function type.
+    // FSMs in CIRCT don't have an async reset concept, so these signals are
+    // not passed through. The reset behavior is captured in the initial state.
```
TaoBi22: Just noticed playing around with this pass - this probably isn't something we need to warn users about since CompReg resets aren't explicitly synchronous (as I recently discovered, we don't technically enforce a reset behaviour over them) so we should be fine to implicitly assume they're synchronous. Thanks for having thought of it and having made the test though!

AtticusKuhn: This code is to erase the reset arguments from the inputs to the FSM.
If we can assume that reset arguments to the CompReg are synchronous, then in the future, we could keep the reset arguments in the list of block arguments, and simply add a special transition that says "If the reset argument is 1, transition to the initial state". Right now, we just ignore/erase the `reset` inputs.

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,988 @@
...
+}
+
+static void addPossibleValues(llvm::DenseSet<size_t> &possibleValues, Value v) {
+  llvm::DenseSet<Value> visited;
+  addPossibleValuesImpl(possibleValues, v, visited);
+}
+
+/// Checks if a value is a constant or a tree of muxes with constant leaves.
+/// Uses an iterative approach with a visited set to handle cycles.
+static bool isConstantLike(Value value) {
```
TaoBi22: Super-nit, but constantlike is an existing op trait in MLIR, so this could be a slightly ambiguous name. Maybe something like `isConstantOrConstantTree` would be clearer?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,988 @@
...
+    Value x = mux.getTrueValue();
+    Value y = mux.getFalseValue();
+    Value b = op.getLhs();
+    Location loc = op.getLoc();
+    auto eq1 = rewriter.create<ICmpOp>(loc, ICmpPredicate::eq, x, b);
+    auto eq2 = rewriter.create<ICmpOp>(loc, ICmpPredicate::eq, y, b);
+    auto newMux = rewriter.create<MuxOp>(loc, mux.getCond(), eq1.getResult(),
+                                         eq2.getResult());
+    op.replaceAllUsesWith(newMux.getOperation());
+    op.erase();
```
TaoBi22: Another nit but this could be cleaner with `rewriter.replaceOpWithNewOp`

lib/Conversion/CoreToFSM/CoreToFSM.cpp
TaoBi22: Is this resolved?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
TaoBi22: I think ignoring them sounds reasonable for the time being (though that would be a nice feature to have later down the line) - I guess this comment should just be updated to reflect that all compreg resets are treated like that

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,988 @@
...
+    OpBuilder builder(module);
+
+    SmallVector<HWModuleOp> modules;
+    for (auto hwModule : module.getOps<HWModuleOp>()) {
+      modules.push_back(hwModule);
+    }
+
+    for (auto hwModule : modules) {
+      builder.setInsertionPoint(hwModule);
+      HWModuleOpConverter converter(builder, hwModule, stateRegs);
```
TaoBi22, line 997: Is there any code that I've missed that replaces module instances with FSM instances? If not then I think this pass would emit invalid IR given a module that's instantiated in the same file - we can always do the actual instance conversion in a later PR but for now we should at least have an error that throws (with a test) if any of the modules are instantiated

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,988 @@
...
+        stateOp.getTransitions().walk(
+            [&](Operation *op) { transitionOps.push_back(op); });
+
+        GreedyRewriteConfig config2;
+        config2.setScope(&stateOp.getTransitions());
+        applyOpPatternsGreedily(transitionOps, frozenPatterns, config2,
+                                &changed);
+
+        
+
```
TaoBi22: this is a mega-nit but spurious newlines

TaoBi22 (approved):
Thanks @AtticusKuhn, this is looking great! Just a few (hopefully final) comments before merge - almost all nits though

lib/Conversion/CoreToFSM/CoreToFSM.cpp
AtticusKuhn: Yes, we check that the register's type is an integer on line 438

```cpp

      if (!isa<IntegerType>(reg.getType())) {
        reg.emitError("FSM extraction only supports integer-typed registers");
        return WalkResult::interrupt();
      }
```

And it's tested in `error.mlir`:

```mlir

// Test: Non-integer register type should produce an error.
// The pass only supports registers with integer types.
hw.module @non_integer_register(in %clk : !seq.clock, in %rst : i1, out output : i1) {
    %c0_i1 = hw.constant 0 : i1
    %zero_array = hw.aggregate_constant [0 : i8, 0 : i8] : !hw.array<2xi8>
    // expected-error @+1 {{FSM extraction only supports integer-typed registers}}
    %state = seq.compreg name "state" %state, %clk reset %rst, %zero_array : !hw.array<2xi8>
    hw.output %c0_i1 : i1
}
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
AtticusKuhn: That's a good catch. I hadn't considered that until now. 


Here's a temporary fix:

```cpp
    // Check if any modules being converted are instantiated in the same file.
    // The pass does not currently support converting instances, so we error
    // out.
    for (auto hwModule : module.getOps<HWModuleOp>()) {
      for (auto instance : hwModule.getOps<hw::InstanceOp>()) {
        StringRef referencedModule = instance.getReferencedModuleName();
        if (moduleNames.contains(referencedModule)) {
          instance.emitError()
              << "module '" << referencedModule
              << "' is instantiated but will be converted to FSM; "
              << "instance conversion is not yet supported";
          signalPassFailure();
          return;
        }
      }
    }
 ```

    And test:

    ```mlir
    hw.module @top_module(in %clk : !seq.clock, in %rst : i1, in %inp : i1, out output : i1) {
    // expected-error @+1 {{module 'instantiated_fsm' is instantiated but will be converted to FSM; instance conversion is not yet supported}}
    %out = hw.instance "child" @instantiated_fsm(clk: %clk: !seq.clock, rst: %rst: i1, inp: %inp: i1) -> (output: i1)
    hw.output %out : i1
}
```
    
    I'll throw an error for now, but in the future, we could also replace `hw.instance` with `fsm.instance`.

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -969,8 +969,27 @@ struct CoreToFSMPass : public circt::impl::ConvertCoreToFSMBase<CoreToFSMPass> {
...
+      moduleNames.insert(hwModule.getName());
+    }
+
+    // Check if any modules being converted are instantiated in the same file.
+    // The pass does not currently support converting instances, so we error
+    // out.
+    for (auto hwModule : module.getOps<HWModuleOp>()) {
+      for (auto instance : hwModule.getOps<hw::InstanceOp>()) {
+        StringRef referencedModule = instance.getReferencedModuleName();
+        if (moduleNames.contains(referencedModule)) {
```
TaoBi22: I think we're fine without this check, the verifier for the instance operation will fail if it can't find a module definition

lib/Conversion/CoreToFSM/CoreToFSM.cpp
TaoBi22: This looks perfect (modulo a note I added below)! Thanks!

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,1002 @@
...
+      addPossibleValuesImpl(operandPossibleValues, operand, visited);
+
+      // It's crucial to handle the case where a sub-computation is too complex.
+      // If we can't determine specific values for an operand, we must
+      // pessimistically assume it can be any value its bitwidth allows.
+      auto opType = dyn_cast<IntegerType>(operand.getType());
+      if (!opType) {
+        concatOp.emitError(
+            "FSM extraction only supports integer-typed operands "
+            "in concat operations");
```
TaoBi22: I think we're missing a test for this error

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,1002 @@
...
+          [](BlockArgument arg) { return true; });
+      FrozenRewritePatternSet patterns(opBuilder.getContext());
+      config.setScope(&outputRegion);
+
+      bool changed = false;
+      LogicalResult converged =
+          applyOpPatternsGreedily(opsToProcess, patterns, config, &changed);
+      opBuilder.setInsertionPoint(stateOp);
+      if (!sortTopologically(&outputRegion.front())) {
+        moduleOp.emitError("could not resolve cycles in module");
```
TaoBi22: Could we add a test for this too?

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,1002 @@
...
+          SmallVector<Operation *> opsToProcess;
+          actionRegion.walk([&](Operation *op) { opsToProcess.push_back(op); });
+          config.setScope(&actionRegion);
+
+          bool changed = false;
+          LogicalResult converged =
+              applyOpPatternsGreedily(opsToProcess, patterns, config, &changed);
+
+          if (!sortTopologically(&actionRegion.front())) {
+            transitionOp.emitError(
```
TaoBi22: As above

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,1002 @@
...
+                "could not resolve cycles in action block of" +
+                std::to_string(currentStateIndex) + " to " + std::to_string(j));
+            return failure();
+          }
+        }
+
+        if (!sortTopologically(&newGuardBlock)) {
+          transitionOp.emitError("could not resolve cycles in guard block of" +
```
TaoBi22: As above

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,1002 @@
...
+        stateOp.getOutput().walk(
+            [&](Operation *op) { outputOps.push_back(op); });
+
+        bool changed = false;
+        GreedyRewriteConfig config;
+        config.setScope(&stateOp.getOutput());
+        LogicalResult converged = applyOpPatternsGreedily(
+            outputOps, frozenPatterns, config, &changed);
+        if (failed(converged)) {
+          stateOp.emitError("Failed to canonicalize the generated state op");
```
TaoBi22: As above

TaoBi22 (commented):
Sorry, 1 final thing! Realised we're missing tests for a couple of errors (possibly ones that were previously assertions), it would be great to have those covered if possible

AtticusKuhn:
>  Realised we're missing tests for a couple of errors (possibly ones that were previously assertions), it would be great to have those covered if possible

Actually, I'm not sure what .mlir inputs could trigger these errors. I think they're unreachable errors. For example, I'm not sure in what case the pattern rewriter would fail to converge.

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -644,10 +648,10 @@ class HWModuleOpConverter {
...
           applyOpPatternsGreedily(opsToProcess, patterns, config, &changed);
       opBuilder.setInsertionPoint(stateOp);
-      if (!sortTopologically(&outputRegion.front())) {
-        moduleOp.emitError("could not resolve cycles in module");
-        return failure();
-      }
+      // MLIR's SSA form guarantees no cycles in the use-def chain, so
+      // sortTopologically should always succeed for valid IR.
+      bool sorted = sortTopologically(&outputRegion.front());
+      assert(sorted && "SSA form guarantees no cycles in output region");
```
TaoBi22: This isn't true when the IR comes from an hw.module, since they use graph regions that allow cycles (since non-combinational cycles are fine and necessary in hardware). Guessing that could be the case here? (i.e. is this assertion triggered if you have some IR with a register that feeds back into itself?)

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -966,27 +966,8 @@ struct CoreToFSMPass : public circt::impl::ConvertCoreToFSMBase<CoreToFSMPass> {
...
-        StringRef referencedModule = instance.getReferencedModuleName();
-        if (moduleNames.contains(referencedModule)) {
-          instance.emitError()
-              << "module '" << referencedModule
-              << "' is instantiated but will be converted to FSM; "
-              << "instance conversion is not yet supported";
-          signalPassFailure();
-          return;
-        }
-      }
```
TaoBi22: Did you mean to drop all of this? I may have been unclear above, sorry if so - we still need to check whether there are instances, the check we can drop is just `if (moduleNames.contains(referencedModule))`

test/Conversion/CoreToFSM/errors.mlir
```diff
@@ -1,5 +1,37 @@
...
+//
+// 4. "could not resolve cycles in guard block of X to Y"
+//    - Same as above - SSA form prevents cycles.
+//
+// 5. "Failed to canonicalize the generated state op"
+//    - applyOpPatternsGreedily() fails only in edge cases like exceeding
+//      iteration limits, which cannot be reliably triggered in tests.
+//
+//===----------------------------------------------------------------------===//
+
```
TaoBi22: I think we're probably fine without this, it can be inferred from the .cpp file

lib/Conversion/CoreToFSM/CoreToFSM.cpp
```diff
@@ -0,0 +1,983 @@
...
+            [&](Operation *op) { outputOps.push_back(op); });
+
+        bool changed = false;
+        GreedyRewriteConfig config;
+        config.setScope(&stateOp.getOutput());
+        LogicalResult converged = applyOpPatternsGreedily(
+            outputOps, frozenPatterns, config, &changed);
+        // applyOpPatternsGreedily fails only in edge cases like exceeding
+        // iteration limits, which cannot realistically occur in this context.
+        assert(succeeded(converged) && "canonicalization should not fail");
```
TaoBi22: It's probably more helpful to have an error message here explaining what happened for debugging:

```suggestion
        assert(succeeded(converged) && "canonicalization failed to converge");
```

lib/Conversion/CoreToFSM/CoreToFSM.cpp
AtticusKuhn: If a register feeds back into itself, we get the following:
Input:
```mlir
hw.module @non_integer_register(in %clk : !seq.clock, in %rst : i1, out output : i1) {
    %c0_i1 = hw.constant 0 : i1
    %state = seq.compreg name "state" %state, %clk reset %c0_i1, %c0_i1 : i1
    hw.output %c0_i1 : i1
}
```

Output:
```
ak2518@autobot:/local/scratch/ak2518$ ./circt/build/bin/circt-opt --convert-core-to-fsm ./circt/thing.mlir
module {
  fsm.machine @non_integer_register(%arg0: i1) -> i1 attributes {initialState = "state_0"} {
    fsm.state @state_0 output {
      %false = hw.constant false
      fsm.output %false : i1
    } transitions {
      fsm.transition @state_0 guard {
        %true = hw.constant true
        fsm.return %true
      }
    }
  }
}
```

Essentially, we assume if the output of a register feeds back into itself, that the FSM stays constantly in the initial state.

TaoBi22: Thanks for checking this! I missed that the registers will be replaced with constants - however, if we have a cycle that doesn't contain a register it looks like we still hit that assertion. For example, adding a couple of cyclic muxes to that example:

```
hw.module @cylic_muxes(in %clk : !seq.clock, in %rst : i1, out output : i1) {
    %c0_i1 = hw.constant 0 : i1
    %state = seq.compreg name "state" %state, %clk reset %c0_i1, %c0_i1 : i1
    %mux1 = comb.mux %rst, %mux2, %state : i1
    %mux2 = comb.mux %state, %mux1, %rst : i1
    hw.output %c0_i1 : i1
}
```

we get a crash as we hit the assertion. Obviously this is a weird case that we don't need to support, but it's something we should gracefully error out on since it's feasible input.

AtticusKuhn: Sorry I took so long to respond. Thanks for catching that. I changed that `assert` to a `emitError()`, which fixes it and I added that example as a test-case to the test-bench.

PR #8853, opened 2025-08-14
[FIRRTL] Firtool: Add option to emit bind files for private modules

rwy7:
The spec does not guarantee the existence of layer bindfiles for private modules, but we have a need to use the verilog of internal/private modules of a circuit and it is helpful if their bindfiles always exist. This PR adds an option to firtool that forces lower layers to emit bindfiles for private modules, even when not necessary.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -199,7 +199,7 @@ struct BindFileInfo {
 
 class LowerLayersPass
     : public circt::firrtl::impl::LowerLayersBase<LowerLayersPass> {
-  using LowerLayersBase::LowerLayersBase;
+  using Base::Base;
 
   hw::OutputFileAttr getOutputFile(SymbolRefAttr layerName) {
     auto layer = symbolToLayer.lookup(layerName);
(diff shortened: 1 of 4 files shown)


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -199,6 +199,8 @@ struct BindFileInfo {
 
 class LowerLayersPass
     : public circt::firrtl::impl::LowerLayersBase<LowerLayersPass> {
+  using LowerLayersBase::LowerLayersBase;
```
seldridge: Nit: I've seen the pattern in upstream of the slightly terser:


```suggestion
  using Base::Base;
```

seldridge (approved):
LGTM, so long as this is not something we keep around forever.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
rwy7: TIL :+1: 

```c++
template <typename DerivedT>
class LowerLayersBase : public ::mlir::OperationPass<firrtl::CircuitOp> {
public:
  using Base = LowerLayersBase;
 ...
```

PR #8857, opened 2025-08-15
[LLHD] Handle eq and ult loop conditions in UnrollLoops pass

fabianschuiki:
The UnrollLoops pass currently only recognizes loops that use an `slt` comparison between the induction variable and an upper bound. Add support for `ult`. Also add support for the degenerate `for (int i = 0; i == 0; ++i)` case, which the canonicalizer will often produce as a strength reduction of `i < 1`.

### DIFF
--- a/test/Dialect/LLHD/Transforms/unroll-loops.mlir
+++ b/test/Dialect/LLHD/Transforms/unroll-loops.mlir
@@ -1,5 +1,7 @@
 // RUN: circt-opt --llhd-unroll-loops %s | FileCheck %s
 
+func.func private @marker()
+
 // CHECK-LABEL: @SimpleLoop
 hw.module @SimpleLoop(out x : i42) {
   %c0_i42 = hw.constant 0 : i42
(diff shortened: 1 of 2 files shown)


test/Dialect/LLHD/Transforms/unroll-loops.mlir
```diff
@@ -350,3 +352,50 @@ hw.module @DontCrashOnSingleBlocks() {
...
+
+// CHECK-LABEL: @DegenerateSingleTripLoopWithEq
+hw.module @DegenerateSingleTripLoopWithEq() {
+  %c0_i32 = hw.constant 0 : i32
+  %c1_i32 = hw.constant 1 : i32
+  llhd.combinational {
+    // CHECK: cf.br
+    // CHECK: func.call @marker()
+    // CHECK: cf.br
+    // CHECK: llhd.yield
```
maerhart: wouldn't that filecheck also succeed if the IR stays untouched? Maybe we need some additional CHECK-NOT?

maerhart (approved):
LGTM

test/Dialect/LLHD/Transforms/unroll-loops.mlir
fabianschuiki: 🤦 You're totally right! Let me adjust this a bit such that we can at least check the exact sequence of lines.

PR #8859, opened 2025-08-15
[LLHD] Properly handle local signals during mem2reg

fabianschuiki:
Extend the LLHD Mem2Reg pass to deal with local signal declarations properly. Since these signals have an initial value, a `llhd.sig` op should propagate that value as an initial definition to subsequent ops. This allows a probe on a local signal that has never been driven to resolve to that initial value.

Additionally, since local signals often become entirely unread after propagation to SSA values, make the Mem2Reg pass perform a cleanup of local signals that are only projected into and driven, but never probed.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
+++ b/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
@@ -1797,7 +1797,7 @@ void Promoter::removeUnusedLocalSignals() {
 /// never probed. Since the signal is local and cannot be observed in any other
 /// way, we can safely remove it along with any projection ops and drives.
 void Promoter::removeUnusedLocalSignal(SignalNode *signal) {
-  // Check if the signal is only ever projected into an driven, but never
+  // Check if the signal is only ever projected into and driven, but never
   // probed.
   SmallSetVector<Operation *, 8> users;
   SmallVector<Operation *> worklist;
--- a/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
+++ b/lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
@@ -31,6 +31,7 @@ using namespace circt;
 using namespace llhd;
 using llvm::PointerIntPair;
 using llvm::SmallDenseSet;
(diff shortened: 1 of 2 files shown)


lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
```diff
@@ -1721,6 +1782,43 @@ void Promoter::replaceValueWith(Value oldValue, Value newValue) {
...
+/// Remove all unused local signals.
+void Promoter::removeUnusedLocalSignals() {
+  for (auto *node : lattice.nodes)
+    if (auto *signal = dyn_cast<SignalNode>(node))
+      removeUnusedLocalSignal(signal);
+}
+
+/// Remove a local signal if it is only used by projection ops and drives, but
+/// never probed. Since the signal is local and cannot be observed in any other
+/// way, we can safely remove it along with any projection ops and drives.
```
maerhart, line 1798: Technically, it could be observed in a simulation trace. In that case, we might want to replace the remaining signal-drive pair with a `hw.wire`. But that's something for a future PR.

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
```diff
@@ -1721,6 +1782,43 @@ void Promoter::replaceValueWith(Value oldValue, Value newValue) {
...
+void Promoter::removeUnusedLocalSignal(SignalNode *signal) {
+  // Check if the signal is only ever projected into an driven, but never
```
maerhart:
```suggestion
  // Check if the signal is only ever projected into and driven, but never
```

lib/Dialect/LLHD/Transforms/Mem2Reg.cpp
fabianschuiki: Yeah that's a good point! I think we should generally remove the "name" attribute from `llhd.sig`, and use `dbg.variable` instead to track named values. And if we ever need to support forcing of signals, we can introduce dedicate "force point" ops.

PR #8862, opened 2025-08-15
Put memories in the same directory as their parent

rwy7:
By the time lower-to-hw runs, the directories have already been assigned. Colocate the memory module with its parent. Remove logic which would set the output directory to the testbench--this is not always right. This allows memories to be emitted into layer specific directories.

### DIFF
--- a/test/firtool/cmem-under-layerblock.fir
+++ b/test/firtool/cmem-under-layerblock.fir
@@ -1,0 +1,19 @@
+; RUN: firtool %s | FileCheck %s
+
+FIRRTL version 5.0.0
+circuit Top:
+  layer Verification, bind, "verification":
+
+  public module Top:
+    input  c : Clock
+    input  i : UInt<1>
+    input  j : UInt<1>
+    input  d : UInt<1>
+    layerblock Verification:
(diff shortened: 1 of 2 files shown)


dtzSiFive (approved):

test/firtool/cmem-under-layerblock.fir
```diff
@@ -0,0 +1,19 @@
...
+    input  j : UInt<1>
+    input  d : UInt<1>
+    layerblock Verification:
+      cmem memory : UInt<1>[2]
+      infer mport rd = memory[i], c
+      infer mport wr = memory[j], c
+      connect wr, d
+      assert(c, rd, UInt<1>(1), "hello")
+
+; CHECK: ----- 8< ----- FILE "verification/memory_2x1.sv" ----- 8< -----
```
dtzSiFive: CI points out need these separators to match windows ones as well.

seldridge: Thanks! I'm commandeering this PR with: https://github.com/llvm/circt/pull/8862/commits/ee8355424d85e453f9b8aab552a7da3396629e94

PR #8868, opened 2025-08-18
[Synth] Add CutRewriter framework and TechMapper pass

uenoku:
This commit implements a cut-based rewriting framework for combinational logic optimization 
and adds the TechMapper pass that uses NPN-equivalence matching with area and delay
metrics to map AIG circuits to technology libraries. The implementation is based on: 
* "Combinational and Sequential Mapping with Priority Cuts", Alan Mishchenko, Sungmin Cho, Satrajit Chatterjee and Robert Brayton, ICCAD 2007
* "Fast Boolean matching based on NPN classification",  Zheng Huang Lingli Wang Yakov Nasikovskiy and Alan Mishchenko, FPT 2013

The Technology Mapping Pass shows how to use this framework in practice by 
mapping generic logic operations to specific hardware gates from a technology 
library. This pass also serves as a test for the framework since 
testing the cut rewriting algorithms directly would be very difficult without 
a real application to exercise them.
The TechMapper pass processes `HWModuleOp` instances with `hw.techlib.info`
attributes as library patterns for now. This must be replaced with more well-designed representation.

The framework is designed to be easily extended with new optimization patterns 
and strategies while maintaining good performance for cut enumeration and 
pattern matching.

### DIFF
--- a/include/circt/Dialect/Synth/Transforms/CutRewriter.h
+++ b/include/circt/Dialect/Synth/Transforms/CutRewriter.h
@@ -31,7 +31,7 @@ namespace circt {
 namespace synth {
 // Type for representing delays in the circuit. It's user's responsibility to
 // use consistent units, i.e., all delays should be in the same unit (usually
-// femto, but not limited to it).
+// femtoseconds, but not limited to it).
 using DelayType = int64_t;
 
 /// Maximum number of inputs supported for truth table generation.
@@ -140,7 +140,8 @@ class MatchedPattern {
 private:
   const CutRewritePattern *pattern = nullptr; ///< The matched library pattern
   Cut *cut = nullptr;                         ///< The cut that was matched
-  SmallVector<DelayType, 2> arrivalTimes; ///< Arrival time through this pattern
+  SmallVector<DelayType, 2>
+      arrivalTimes; ///< Arrival times of outputs from this pattern
 
 public:
   /// Default constructor creates an invalid matched pattern.
@@ -187,7 +188,7 @@ private:
 
 public:
   /// Get the best matched pattern for this cut set.
-  std::optional<MatchedPattern> getMatchedPattern() const;
+  std::optional<MatchedPattern> getBestMatchedPattern() const;
 
   /// Check if this cut set has a valid matched pattern.
   bool isMatched() const;
@@ -340,7 +341,10 @@ struct CutRewritePattern {
   ///
   /// If this method returns true, the pattern matcher will use truth table
   /// comparison for efficient pre-filtering. Only cuts with matching truth
-  /// tables will be passed to the match() method.
+  /// tables will be passed to the match() method. If it returns false, the
+  /// pattern will be checked against all cuts regardless of their truth tables.
+  /// This is useful for patterns that match regardless of their truth tables,
+  /// such as LUT-based patterns.
   virtual bool
   useTruthTableMatcher(SmallVectorImpl<NPNClass> &matchingNPNClasses) const;
 
@@ -465,11 +469,13 @@ public:
 
 private:
   /// Enumerate cuts for all nodes in the given module.
+  /// Note: This preserves module boundaries and does not perform
+  /// rewriting across the hierarchy.
   LogicalResult enumerateCuts(Operation *topOp);
 
   /// Find patterns that match a cut's truth table.
   ArrayRef<std::pair<NPNClass, const CutRewritePattern *>>
-  getMatchingPatternFromTruthTable(const Cut &cut) const;
+  getMatchingPatternsFromTruthTable(const Cut &cut) const;
 
   /// Match a cut against available patterns and compute arrival time.
   std::optional<MatchedPattern> patternMatchCut(Cut &cut);
--- a/include/circt/Dialect/Synth/Transforms/CutRewriter.h
+++ b/include/circt/Dialect/Synth/Transforms/CutRewriter.h
@@ -29,18 +29,33 @@
 
 namespace circt {
 namespace synth {
+// Type for representing delays in the circuit. It's user's responsibility to
+// use consistent units, i.e., all delays should be in the same unit (usually
+// femto, but not limited to it).
+using DelayType = int64_t;
+
+/// Maximum number of inputs supported for truth table generation.
+/// This limit prevents excessive memory usage as truth table size grows
+/// exponentially with the number of inputs (2^n entries).
+static constexpr unsigned maxTruthTableInputs = 16;
 
 // This is a helper function to sort operations topologically in a logic
 // network. This is necessary for cut rewriting to ensure that operations are
 // processed in the correct order, respecting dependencies.
 LogicalResult topologicallySortLogicNetwork(mlir::Operation *op);
 
+// Get the truth table for a specific operation within a block.
+// Block must be a SSACFG or topologically sorted.
+FailureOr<BinaryTruthTable> getTruthTable(ValueRange values, Block *block);
+
 //===----------------------------------------------------------------------===//
 // Cut Data Structures
 //===----------------------------------------------------------------------===//
 
 // Forward declarations
+class CutRewritePatternSet;
 class CutRewriter;
+struct CutRewritePattern;
 struct CutRewriterOptions;
 
 /// Represents a cut in the combinational logic network.
@@ -121,9 +136,36 @@ public:
 /// This class encapsulates the result of matching a cut against a rewriting
 /// pattern during optimization. It stores the matched pattern, the
 /// cut that was matched, and timing information needed for optimization.
-/// TODO: Currently only added as a placeholder.
 class MatchedPattern {
+private:
+  const CutRewritePattern *pattern = nullptr; ///< The matched library pattern
+  Cut *cut = nullptr;                         ///< The cut that was matched
+  SmallVector<DelayType, 2> arrivalTimes; ///< Arrival time through this pattern
+
 public:
+  /// Default constructor creates an invalid matched pattern.
+  MatchedPattern() = default;
+
+  /// Constructor for a valid matched pattern.
+  MatchedPattern(const CutRewritePattern *pattern, Cut *cut,
+                 SmallVector<DelayType, 2> arrivalTimes)
+      : pattern(pattern), cut(cut), arrivalTimes(std::move(arrivalTimes)) {}
+
+  /// Get the arrival time of signals through this pattern.
+  DelayType getArrivalTime(unsigned outputIndex) const;
+  ArrayRef<DelayType> getArrivalTimes() const;
+
+  /// Get the library pattern that was matched.
+  const CutRewritePattern *getPattern() const;
+
+  /// Get the cut that was matched to the pattern.
+  Cut *getCut() const;
+
+  /// Get the area cost of using this pattern.
+  double getArea() const;
+
+  /// Get the delay between specific input and output pins.
+  DelayType getDelay(unsigned inputIndex, unsigned outputIndex) const;
 };
 
 /// Manages a collection of cuts for a single logic node using priority cuts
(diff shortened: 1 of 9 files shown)


include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -29,18 +29,33 @@
 
 namespace circt {
 namespace synth {
+// Type for representing delays in the circuit. It's user's responsibility to
+// use consistent units, i.e., all delays should be in the same unit (usually
+// femto, but not limited to it).
```
cowardsa: Super nit: femto seconds?

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -121,9 +136,36 @@ class Cut {
 /// This class encapsulates the result of matching a cut against a rewriting
 /// pattern during optimization. It stores the matched pattern, the
 /// cut that was matched, and timing information needed for optimization.
-/// TODO: Currently only added as a placeholder.
 class MatchedPattern {
+private:
+  const CutRewritePattern *pattern = nullptr; ///< The matched library pattern
+  Cut *cut = nullptr;                         ///< The cut that was matched
+  SmallVector<DelayType, 2> arrivalTimes; ///< Arrival time through this pattern
```
cowardsa: Nit: arrival times of inputs to matched pattern (I found through confusing)

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -144,6 +186,16 @@ class CutSet {
   bool isFrozen = false; ///< Whether cut set is finalized
 
 public:
+  /// Get the best matched pattern for this cut set.
+  std::optional<MatchedPattern> getMatchedPattern() const;
```
cowardsa: Nit: getBestMatchedPattern(), could also have a getMatchedPatterns() function - as presumably there will be some area/delay tradeoff to balance?

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -252,6 +312,180 @@ class CutEnumerator {
...
+  /// This method orchestrates the entire rewriting process:
+  /// 1. Enumerate cuts for all nodes in the combinational logic
+  /// 2. Match cuts against available patterns
+  /// 3. Select optimal patterns based on strategy
+  /// 4. Rewrite the circuit with selected patterns
+  LogicalResult run(Operation *topOp);
+
+private:
+  /// Enumerate cuts for all nodes in the given module.
+  LogicalResult enumerateCuts(Operation *topOp);
```
cowardsa: Q: will the current implementation will preserve module boundaries? Namely, there will be no rewriting across the hierarchy?

integration_test/circt-synth/mapping-lec.mlir
```diff
@@ -0,0 +1,54 @@
...
+// RUN: circt-opt %s --convert-aig-to-comb -o %t2.mlir
+// RUN: circt-lec %t1.mlir %t2.mlir -c1=mul -c2=mul --shared-libs=%libz3 | FileCheck %s --check-prefix=COMB_MUL_TECHMAP
+
+// COMB_MUL_TECHMAP: c1 == c2
+
+// Set delay for binary and inv op to 5 so that others will be prioritized
+hw.module @and_inv(in %a : i1, in %b : i1, out result : i1) attributes {hw.techlib.info = {area = 1.0 : f64, delay = [[5], [5]]}} {
+    %0 = aig.and_inv %a, %b : i1
+    hw.output %0 : i1
+}
```
cowardsa, line 15: This is really cool - like the idea of specifying the technology library within the IR itself! Maybe its added elsewhere but perhaps a safety mechanism would be useful to check that a pattern's techlib.info is only being set once?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -424,9 +532,69 @@ void CutSet::finalize(const CutRewriterOptions &options) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// CutRewritePattern
+//===----------------------------------------------------------------------===//
+
+bool CutRewritePattern::useTruthTableMatcher(
+    SmallVectorImpl<NPNClass> &matchingNPNClasses) const {
+  return false;
+}
```
cowardsa, line 571: Why does this always return false - thought the default matching procedure used truth tables?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -424,9 +532,69 @@ void CutSet::finalize(const CutRewriterOptions &options) {
...
+// CutRewritePatternSet
+//===----------------------------------------------------------------------===//
+
+CutRewritePatternSet::CutRewritePatternSet(
+    llvm::SmallVector<std::unique_ptr<CutRewritePattern>, 4> patterns)
+    : patterns(std::move(patterns)) {
+  // Initialize the NPN to pattern map
+  for (auto &pattern : this->patterns) {
+    SmallVector<NPNClass, 2> npnClasses;
+    if (pattern->useTruthTableMatcher(npnClasses)) {
```
cowardsa: Nit: dead code?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -578,3 +746,225 @@ const CutSet *CutEnumerator::getCutSet(Value value) {
...
+    assert(cutSet && "Input must have a valid cut set");
+
+    auto matchedPattern = cutSet->getMatchedPattern();
+    // If there is no matching pattern, it means it's not possilbe to use the
+    // input in the cut rewriting. So abort early.
+    if (!matchedPattern)
+      return {};
+
+    // This must be a block argument must have been a cut input.
+    auto resultNumber = cast<mlir::OpResult>(input);
```
cowardsa, line 844: Nit: does this need a check that the case is safe?

include/circt/Dialect/Synth/Transforms/CutRewriter.h
```diff
@@ -252,6 +312,180 @@ class CutEnumerator {
...
+
+  /// Find patterns that match a cut's truth table.
+  ArrayRef<std::pair<NPNClass, const CutRewritePattern *>>
+  getMatchingPatternFromTruthTable(const Cut &cut) const;
```
cowardsa: Nit: getMatchingPatternsFromTruthTable

lib/Dialect/Synth/Transforms/TechMapper.cpp
```diff
@@ -0,0 +1,255 @@
...
+      auto npnClass = getNPNClassFromModule(hwModule);
+      if (failed(npnClass)) {
+        signalPassFailure();
+        return;
+      }
+
+      // Create a CutRewritePattern for the library module
+      std::unique_ptr<CutRewritePattern> pattern =
+          std::make_unique<TechLibraryPattern>(hwModule, area, std::move(delay),
+                                               std::move(*npnClass));
```
cowardsa, line 225: Q: do we have any checks on unique names and unique pattern definitions? Does it matter if we have two patterns implementing the same circuit with different costs?

test/Dialect/Synth/tech-mapper-error.mlir
```diff
@@ -0,0 +1,77 @@
...
+    hw.output %0 : i1
+}
+
+// -----
+
+hw.module @do_nothing(in %a : i1, out result : i1) attributes {hw.techlib.info = {area = 1.0 : f64, delay = [[1]]}} {
+    hw.output %a : i1
+}
+
+// No tech library found so pattern matching is failed.
```
cowardsa: Nit: comment doesn't seem to match detected error?

test/Dialect/Synth/tech-mapper-error.mlir
```diff
@@ -0,0 +1,77 @@
...
+// expected-error@+1 {{All input ports must be single bit}}
+hw.module @multibit(in %a : i2, in %b: i1, out result : i1) attributes {hw.techlib.info = {area = 1.0 : f64, delay = [[1], [2]]}} {
+  hw.output %b: i1
+}
+
+// -----
+
+
+// expected-error@+1 {{All input ports must be single bit}}
+hw.module @multibit(in %a : i1, in %b: i2, out result : i2) attributes {hw.techlib.info = {area = 1.0 : f64, delay = [[1], [2]]}} {
```
cowardsa: Q: are multi-bit outputs supported?

test/Dialect/Synth/tech-mapper.mlir
```diff
@@ -0,0 +1,160 @@
...
+}
+
+hw.module @area_flow(in %a : i1, in %b : i1, in %c: i1, out result : i1) attributes {hw.techlib.info = {area = 1.5 : f64, delay = [[10], [10], [10], [10], [10]]}} {
+    %0 = aig.and_inv not %a, not %b : i1
+    %1 = aig.and_inv not %c, %0 : i1
+    hw.output %1 : i1
+}
+
+// This is a test that needs area-flow to get an optimal result.
+// It produces sub-optimal mappings since currently area-flow is not implemented.
```
cowardsa: Q: What is area-flow?

include/circt/Dialect/Synth/Transforms/CutRewriter.h
uenoku: Yes, it preserves module boundaries. Rewriting across module boundaries would be very complicated (and also it's hard to parallelize across modules) so currently I don't have a plan. Users are expected to perform inlining or flattening if context-sensitivity is important. I think we can still use global timing analysis results as arrival times of inputs (input ports or instance results).

include/circt/Dialect/Synth/Transforms/CutRewriter.h
uenoku: That's good point. Will change to `getBestMatchedPattern`. I think keeping few matched patterns are great and will do in the follow-up

lib/Dialect/Synth/Transforms/CutRewriter.cpp
uenoku: `isAlwaysCutInput` returns `true` for a block argument, so it's safe to cast here. `cast` internally has an assertion so I think an additional check is not necessary. Thank you for suggestion!

lib/Dialect/Synth/Transforms/CutRewriter.cpp
uenoku: This is a virtual function so a pattern can override the function. So if a pattern implements a LUT, we can return false since LUT can match regardless of truth table. The computation of truth table is a bit expensive so this tries to avoid truth table computation when we only target LUT operations.

lib/Dialect/Synth/Transforms/TechMapper.cpp
uenoku: Yes, it's possible that there are multiple patterns that are equivalent circuit with different costs, e.g. pattern A/B has smallest delay from input 0/1 respectively.  CutRewriter will try all available patterns and picks a best pattern.

test/Dialect/Synth/tech-mapper-error.mlir
uenoku: Good point. It's not supported now. I'll add a test. It's possible to support but it will be simply treated as multiple single bits outputs.

test/Dialect/Synth/tech-mapper.mlir
uenoku: I added a comment and reference to "Heuristics for Area Minimization in LUT-Based FPGA Technology Mapping" which proposed area-flow.

uenoku:
>  my only major concern is around useTruthTableMatcher

Good point, this is is not used in the current PR so I removed a handling useTrtuTableMatcher in `CutRewritePatternSet` constructor. This is used when a pattern is for LUT where it doesn't need truth table while matching. In the follow-up I will add (generic) LUT mapper so I'd like to test there.

integration_test/circt-synth/mapping-lec.mlir
uenoku: Yeah eventually we would like to introduce a cell operation but using attributes on hw.module would be good enough until then. The hw.techlib.info is an attribute so MLIR verifier should ensure the uniqueness.

PR #8872, opened 2025-08-19
[Arc] Consider array/struct field alignment when computing state size

fabianschuiki:
The `StateType` in the Arc dialect currently just defers to `hw::getBitWidth` to compute the number of bits that need to be allocated in the simulator state. This is problematic, since types like structs and arrays have field and element alignment requirements, making LLVM require more bytes for these types than the summed-up bit width would suggest. As a result, the AllocateState pass would allocate 2 bytes for `struct<tag: i5, sign_ext: i1, offset: i3, size: i2>`, even though LLVM would consider the individual fields to be aligned to the nearest byte, making the struct cover 4 bytes in memory.

To fix this issue, make `StateType` more carefully compute the actual number of bits required in simulator storage by requiring all elements of a array or struct to be at least byte-aligned, and to more carefully track the alignmenmt of individual fields. This allows us to predict how many bytes more accurately.

This fixes a few nasty simulation mismatches between Arcilator and Verilator due to the memory corruption that would result from not allocating sufficient storage for arrays and structs.

### DIFF
--- a/lib/Dialect/Arc/ArcTypes.cpp
+++ b/lib/Dialect/Arc/ArcTypes.cpp
@@ -21,18 +21,66 @@ using namespace mlir;
 #define GET_TYPEDEF_CLASSES
 #include "circt/Dialect/Arc/ArcTypes.cpp.inc"
 
-unsigned StateType::getBitWidth() {
-  if (llvm::isa<seq::ClockType>(getType()))
+/// Compute the bit width a type will have when allocated as part of the
+/// simulator's storage. This includes any padding and alignment that may be
+/// necessary once the type has been mapped to LLVM. The idea is for this
+/// function to be conservative, such that we provide sufficient storage bytes
+/// for any type.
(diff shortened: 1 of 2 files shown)


lib/Dialect/Arc/ArcTypes.cpp
```diff
@@ -21,18 +21,61 @@ using namespace mlir;
 #define GET_TYPEDEF_CLASSES
 #include "circt/Dialect/Arc/ArcTypes.cpp.inc"
 
-unsigned StateType::getBitWidth() {
-  if (llvm::isa<seq::ClockType>(getType()))
+static std::optional<uint64_t> computeBitWidth(Type type) {
```
maerhart: maybe call it `computeLLVMBitWidth` or something to make it super clear that this includes alignment/padding

fabianschuiki: Good idea!

PR #8874, opened 2025-08-20
Remove some unneeded defs and unused variable warnings.

jpienaar:
NFC.

### DIFF
--- a/lib/Conversion/ImportVerilog/AssertionExpr.cpp
+++ b/lib/Conversion/ImportVerilog/AssertionExpr.cpp
@@ -116,6 +116,7 @@ struct AssertionExprVisitor {
         return {};
 
       Type valueType = sequenceValue.getType();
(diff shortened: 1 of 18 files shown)


lib/Conversion/ImportVerilog/AssertionExpr.cpp
```diff
@@ -116,6 +116,7 @@ struct AssertionExprVisitor {
         return {};
 
       Type valueType = sequenceValue.getType();
+      (void)valueType;
```
seldridge: `[[maybe_unused]]` would provide better intent and to align with the LLVM developer guidelines around sole-usage in an assert: https://llvm.org/docs/CodingStandards.html#prefer-c-style-casts Ditto throughout.

fabianschuiki (approved):
Thanks for doing this cleanup. LGTM!

PR #8879, opened 2025-08-21
[ESI] Add serde capabilities to ESI C++ runtime types

mortbopet:
Adds serialization/deserialization capabilities to the C++ side of the ESI runtime. Naturally, this is more limited than the python side, most notably around integers, since we don't (yet) have a way to represent infinite precision integers. However, for most usecases, serde capabilities up-to-and-including [s|u)int64 should be a common, and generically useful thing.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
+++ b/lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
@@ -106,9 +106,7 @@ public:
   std::ptrdiff_t getBitWidth() const override { return inner->getBitWidth(); };
 
   void ensureValid(const std::any &obj) const override;
-
   MessageData serialize(const std::any &obj) const override;
-
   std::pair<std::any, std::span<const uint8_t>>
   deserialize(std::span<const uint8_t> data) const override;
 
@@ -124,9 +122,7 @@ public:
   std::ptrdiff_t getBitWidth() const override { return 1; };
 
   void ensureValid(const std::any &obj) const override;
-
   MessageData serialize(const std::any &obj) const override;
-
   std::pair<std::any, std::span<const uint8_t>>
   deserialize(std::span<const uint8_t> data) const override;
 };
@@ -160,9 +156,7 @@ public:
   using BitVectorType::BitVectorType;
 
   void ensureValid(const std::any &obj) const override;
-
   MessageData serialize(const std::any &obj) const override;
-
   std::pair<std::any, std::span<const uint8_t>>
   deserialize(std::span<const uint8_t> data) const override;
 };
@@ -180,9 +174,7 @@ public:
   using IntegerType::IntegerType;
 
   void ensureValid(const std::any &obj) const override;
-
   MessageData serialize(const std::any &obj) const override;
-
   std::pair<std::any, std::span<const uint8_t>>
   deserialize(std::span<const uint8_t> data) const override;
 };
@@ -193,9 +185,7 @@ public:
   using IntegerType::IntegerType;
 
   void ensureValid(const std::any &obj) const override;
-
   MessageData serialize(const std::any &obj) const override;
-
   std::pair<std::any, std::span<const uint8_t>>
   deserialize(std::span<const uint8_t> data) const override;
 };
@@ -221,9 +211,7 @@ public:
   }
 
   void ensureValid(const std::any &obj) const override;
-
   MessageData serialize(const std::any &obj) const override;
-
   std::pair<std::any, std::span<const uint8_t>>
   deserialize(std::span<const uint8_t> data) const override;
 
@@ -255,9 +243,7 @@ public:
   }
 
   void ensureValid(const std::any &obj) const override;
-
   MessageData serialize(const std::any &obj) const override;
-
   std::pair<std::any, std::span<const uint8_t>>
   deserialize(std::span<const uint8_t> data) const override;
 
--- a/lib/Dialect/ESI/runtime/cpp/lib/Types.cpp
+++ b/lib/Dialect/ESI/runtime/cpp/lib/Types.cpp
@@ -219,6 +219,12 @@ void StructType::ensureValid(const std::any &obj) const {
     }
 
     for (const auto &[fieldName, fieldType] : fields) {
+      if (fieldType->getBitWidth() % 8 != 0)
+        throw std::runtime_error(std::format(
+            "C++ ser/de of struct types only supports "
+            "structs with byte-aligned fields, but field '{}' has {} bits",
+            fieldName, fieldType->getBitWidth()));
+
       auto it = structData.find(fieldName);
       if (it == structData.end())
         throw std::runtime_error(std::format("missing field '{}'", fieldName));
@@ -246,12 +252,21 @@ MessageData StructType::serialize(const std::any &obj) const {
   if (totalSize > 0)
     result.reserve((totalSize + 7) / 8);
 
-  // Serialize fields in reverse order (to match Python implementation)
-  for (auto it = fields.rbegin(); it != fields.rend(); ++it) {
-    const auto &[fieldName, fieldType] = *it;
+  auto serializeField = [&](const std::pair<std::string, const Type *> &field) {
+    auto &fieldName = field.first;
+    auto &fieldType = field.second;
     auto fieldData = fieldType->serialize(structData.at(fieldName));
     const auto &fieldBytes = fieldData.getData();
     result.insert(result.end(), fieldBytes.begin(), fieldBytes.end());
+  };
+
+  // Serialize fields in reverse order.
+  if (isReverse()) {
+    for (auto it = fields.rbegin(); it != fields.rend(); ++it)
+      serializeField(*it);
+  } else {
+    for (const auto &field : fields)
+      serializeField(field);
   }
 
   return MessageData(std::move(result));
@@ -262,12 +277,22 @@ StructType::deserialize(std::span<const std::uint8_t> data) const {
   std::map<std::string, std::any> result;
   std::span<const std::uint8_t> remaining = data;
 
-  // Deserialize fields in reverse order (to match Python implementation)
-  for (auto it = fields.rbegin(); it != fields.rend(); ++it) {
-    const auto &[fieldName, fieldType] = *it;
-    auto [fieldValue, newRemaining] = fieldType->deserialize(remaining);
-    result[fieldName] = fieldValue;
-    remaining = newRemaining;
+  auto deserializeField =
+      [&](const std::pair<std::string, const Type *> &field) {
+        auto &fieldName = field.first;
+        auto &fieldType = field.second;
+        auto [fieldValue, newRemaining] = fieldType->deserialize(remaining);
+        result[fieldName] = fieldValue;
+        remaining = newRemaining;
+      };
+
+  if (isReverse()) {
+    // Deserialize fields in reverse order.
+    for (auto it = fields.rbegin(); it != fields.rend(); ++it)
+      deserializeField(*it);
+  } else {
+    for (const auto &field : fields)
+      deserializeField(field);
   }
 
   return {std::any(result), remaining};
--- a/unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
+++ b/unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
@@ -23,11 +23,11 @@ TEST(ESITypesTest, VoidTypeSerialization) {
 
   // Test valid void value (empty std::any)
   std::any voidValue;
-  EXPECT_NO_THROW(voidType.isValid(voidValue));
+  EXPECT_NO_THROW(voidType.ensureValid(voidValue));
 
   // Test valid void value (nullptr)
   std::any nullptrValue = std::any(nullptr);
-  EXPECT_NO_THROW(voidType.isValid(nullptrValue));
+  EXPECT_NO_THROW(voidType.ensureValid(nullptrValue));
 
   // Test invalid void value
   std::any invalidValue = std::any(42);
@@ -35,13 +35,17 @@ TEST(ESITypesTest, VoidTypeSerialization) {
 
   // Test serialization
   MessageData serialized = voidType.serialize(voidValue);
-  EXPECT_EQ(serialized.getSize(), 1UL);
-  EXPECT_EQ(serialized.getData()[0], 0);
+  EXPECT_EQ(serialized.getSize(), 1UL)
+      << "VoidType serialization should produce exactly 1 byte";
+  EXPECT_EQ(serialized.getData()[0], 0)
+      << "VoidType serialization should produce a zero byte";
 
   // Test deserialization
   auto [deserialized, remaining] = voidType.deserialize(serialized);
-  EXPECT_FALSE(deserialized.has_value());
-  EXPECT_EQ(remaining.size(), 0UL);
+  EXPECT_FALSE(deserialized.has_value())
+      << "VoidType deserialization should not have a value";
+  EXPECT_EQ(remaining.size(), 0UL)
+      << "VoidType deserialization should consume all data";
 }
 
 // Test BitsType serialization and deserialization
@@ -51,7 +55,7 @@ TEST(ESITypesTest, BitsTypeSerialization) {
   // Test valid bits value
   std::vector<uint8_t> bitsValue = {0xAB};
   std::any validBits = std::any(bitsValue);
-  EXPECT_NO_THROW(bitsType.isValid(validBits));
+  EXPECT_NO_THROW(bitsType.ensureValid(validBits));
 
   // Test invalid size
   std::vector<uint8_t> wrongSize = {0xAB, 0xCD};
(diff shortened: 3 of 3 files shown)


lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
```diff
@@ -83,6 +121,41 @@ class VoidType : public Type {
...
+      return {true, ""};
+    } catch (const std::bad_any_cast &) {
+      return {false,
+              "void type must be represented by empty std::any or nullptr"};
+    }
+  }
+
+  MessageData serialize(const std::any &obj) const override {
+    auto [valid, reason] = isValid(obj);
+    if (!valid) {
```
teqdruid: nit: plz comply with LLVM style.

lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
```diff
@@ -125,12 +234,137 @@ class IntegerType : public BitVectorType {
...
+      throw std::runtime_error("Invalid object for sint type: " + reason);
+    }
+    auto value = std::any_cast<int64_t>(obj);
+    size_t byteSize = (getWidth() + 7) / 8;
+    std::vector<uint8_t> data(byteSize);
+
+    // Little-endian serialization with sign extension
```
teqdruid: Maybe I'm daft, but I don't see the sign-extension logic.

lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
```diff
@@ -125,12 +234,137 @@ class IntegerType : public BitVectorType {
...
+        return {false, "value " + std::to_string(value) + " out of range for " +
+                           std::to_string(getWidth()) + "-bit signed integer"};
+      }
+      return {true, ""};
+    } catch (const std::bad_any_cast &) {
+      return {false, "must be int64_t"};
+    }
+  }
+
+  MessageData serialize(const std::any &obj) const override {
```
teqdruid: Can't the (de-)serialization logic be just pointer casting + sign extension logic? That seems much simpler to me.

teqdruid: MessageData has a pointer+size constructor, so this could be as simple as checking/setting a sign bit then passing a casted pointer and the byte size into MessageData.

lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
```diff
@@ -125,12 +234,137 @@ class IntegerType : public BitVectorType {
...
+    auto [valid, reason] = isValid(obj);
+    if (!valid) {
+      throw std::runtime_error("Invalid object for uint type: " + reason);
+    }
+    auto value = std::any_cast<uint64_t>(obj);
+    size_t byteSize = (getWidth() + 7) / 8;
+    std::vector<uint8_t> data(byteSize);
+
+    // Little-endian serialization
```
teqdruid: Same comment as above re pointer tricks.

lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
```diff
@@ -153,6 +387,69 @@ class StructType : public Type {
...
+  }
+
+  MessageData serialize(const std::any &obj) const override {
+    auto [valid, reason] = isValid(obj);
+    if (!valid) {
+      throw std::runtime_error("Invalid object for struct type: " + reason);
+    }
+
+    auto structData = std::any_cast<std::map<std::string, std::any>>(obj);
+    std::vector<uint8_t> result;
```
teqdruid: For performance, might it be a good idea to reserve space for the entire struct?

teqdruid (commented):
A few comments:

- Since this serialization will be slow regardless of any optimizations (due to the input and output `std::any`s and contents thereof), I'm not sure any performance improvement is worth it. There is, however, a bunch of data movement in this implementation. It may be a good idea during serialization to pre-allocate space for the entire message then track the offset/iterator into it. Similarly, the deserialization code could return 'slices' of the `MessageData` blob instead of copying data.

- Please follow the LLVM style guide. I see a lot of `{}` surrounding single-line blocks.

- This is too much code for unnecessarily being in a header. It's not performance sensitive, so inlining doesn't make sense. It's also not templated code. Please move this into a new `Types.cpp` file.

mortbopet:
@teqdruid thanks for the comments. Re. commens regarding MessageData `ptr+size` constructors. Since we're anyways opting into C++20 for the ESI runtime, i've changed the API to instead pass around `std::span<const uint8_t>>`. That's a much better abstraction for popping arbitrary amounts of bytes here and there, while ensuring zero copies during a deserialization scenario (which is the case when we'd be passing back a new `MessageData` from every deserializer).  
I also modified the `isValid` logic - it felt a bit backwards/too copied from python. Instead, `isValid` will now throw instead, and users should catch this if they want to handle the invalid case. This also DRY's up the serialization functions.

lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
```diff
@@ -33,6 +39,26 @@ class Type {
...
+  /// Deserialize from a span of bytes to an object. Returns the deserialized
+  /// object as a std::any and a span to the remaining bytes.
+  virtual std::pair<std::any, std::span<const uint8_t>>
+  deserialize(std::span<const uint8_t> data) const {
+    throw std::runtime_error("Deserialization not implemented for type " + id);
+  }
+
+  /// Check if a std::any object is valid for this type. Throws
+  /// std::runtime_error if the object is not valid.
+  virtual void isValid(const std::any &obj) const {
```
teqdruid: One of the reasons the Python API is the way it is is because it doesn't make sense to have a method called `isValid` throw an exception rather than return bool. If we're going to go this way, rename it to `checkValid` or `ensureValid`. We can then have an `isValid` with a try-catch.

Though I prefer it the other way. I don't like relying on exceptions to do something perfectly valid. "Exceptions should be for exceptional circumstances" -Some book I once read.

teqdruid:
Running a build to ensure that our publishing pipeline has support for `std::span`: https://github.com/llvm/circt/actions/runs/17165704652

teqdruid:
> Running a build to ensure that our publishing pipeline has support for `std::span`: https://github.com/llvm/circt/actions/runs/17165704652

Looks like it failed. If you want to use `std::format` (and likely `std::span`), you'll have to figure out how to upgrade the std C++ lib in our Docker containers. I'd like to use these new features as well as they significantly clean up the code.

lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
mortbopet: > Though I prefer it the other way. I don't like relying on exceptions to do something perfectly valid. "Exceptions should be for exceptional circumstances" -Some book I once read.

I used to be of the same opinion, but have lately come to, errr, embrace exceptions more. I think they're a perfectly valid, and extremely useful, way to get another form of control flow without going overboard in C++ boilerplate.
Renamed to `ensureValid`, and changes `isValid` to be with a try-catch statement.

mortbopet:
> Looks like it failed. If you want to use std::format (and likely std::span), you'll have to figure out how to upgrade the std C++ lib in our Docker containers. I'd like to use these new features as well as they significantly clean up the code.

Looks like the only issue here was the [esiruntime-wheel](https://github.com/orgs/circt/packages/container/package/images%2Fesiruntime-wheel) being out of date (since it already is [based on an image which _should_ have c++20 support](https://github.com/circt/images/blob/trunk/esiruntime-wheel/Dockerfile#L1)). I kicked it off, [and re-ran the publish pipeline, and things work!
](https://github.com/llvm/circt/actions/runs/17404946760)

lib/Dialect/ESI/runtime/cpp/include/esi/Types.h
```diff
@@ -73,6 +105,13 @@ class ChannelType : public Type {
   const Type *getInner() const { return inner; }
   std::ptrdiff_t getBitWidth() const override { return inner->getBitWidth(); };
 
+  void ensureValid(const std::any &obj) const override;
+
```
teqdruid: Nit: the whitespace here looks odd to me. These methods are all related so let's get rid of the blank lines between them in all of the classes without the doc comments.

lib/Dialect/ESI/runtime/cpp/lib/Types.cpp
```diff
@@ -0,0 +1,337 @@
...
+    for (const auto &[fieldName, fieldType] : fields) {
+      auto it = structData.find(fieldName);
+      if (it == structData.end())
+        throw std::runtime_error(std::format("missing field '{}'", fieldName));
+
+      try {
+        fieldType->ensureValid(it->second);
+      } catch (const std::runtime_error &e) {
+        throw std::runtime_error(
+            std::format("invalid field '{}': {}", fieldName, e.what()));
```
teqdruid, line 236: This creates the same problem as my objection to the other PR. It's not ideal, but in the C++ case I think it's better than a stack trace. Also, see my upcoming suggestion in the Python error message PR.

lib/Dialect/ESI/runtime/cpp/lib/Types.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+  auto structData = std::any_cast<std::map<std::string, std::any>>(obj);
+
+  // Pre-allocate space for performance
+  std::ptrdiff_t totalSize = getBitWidth();
+  std::vector<uint8_t> result;
+  if (totalSize > 0)
+    result.reserve((totalSize + 7) / 8);
+
+  // Serialize fields in reverse order (to match Python implementation)
```
teqdruid: Do the same thing here as you did in the Python serializer.

lib/Dialect/ESI/runtime/cpp/lib/Types.cpp
```diff
@@ -0,0 +1,337 @@
...
+  auto arrayData = std::any_cast<std::vector<std::any>>(obj);
+
+  // Pre-allocate space for performance
+  std::ptrdiff_t totalSize = getBitWidth();
+  std::vector<uint8_t> result;
+  if (totalSize > 0)
+    result.reserve((totalSize + 7) / 8);
+
+  // Serialize elements in reverse order (to match Python implementation)
+  for (auto it = arrayData.rbegin(); it != arrayData.rend(); ++it) {
```
teqdruid, line 335: Is this correct? We might check if it is...

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+namespace {
+
+// Test VoidType serialization and deserialization
+TEST(ESITypesTest, VoidTypeSerialization) {
+  VoidType voidType("void");
+
+  // Test valid void value (empty std::any)
+  std::any voidValue;
+  EXPECT_NO_THROW(voidType.isValid(voidValue));
```
teqdruid: `isValid` doesn't ever throw so `EXPECT_NO_THROW` will always evaluate true. This issue is repeated a bunch of times in this test.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+  // Test deserialization
+  auto [deserialized, remaining] = sintType.deserialize(serialized);
+  auto deserializedSInt = std::any_cast<int64_t>(deserialized);
+  EXPECT_EQ(deserializedSInt, 0x1234);
+  EXPECT_EQ(remaining.size(), 0UL);
+
+  // Test negative value serialization/deserialization
+  MessageData negSerialized = sintType.serialize(validNegSInt);
+  auto [negDeserialized, negRemaining] = sintType.deserialize(negSerialized);
```
teqdruid, line 158: Please add a check of the serialized bytes like are in the unsigned cases above.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+  EXPECT_EQ(resultMaxPos4, 7);
+
+  // Test 12-bit signed integers (non-byte-aligned case)
+  SIntType sint12("sint12", 12);
+
+  // Test -1 in 12-bit (should be 0xFFF in lower 12 bits)
+  int64_t minus1w12bit = -1;
+  MessageData serialized12bit = sint12.serialize(std::any(minus1w12bit));
+  EXPECT_EQ(serialized12bit.getSize(), 2UL); // 12 bits = 2 bytes
+
```
teqdruid, line 255: Please add the 0xFFF check here.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+    std::span<const uint8_t> spanMax(serializedMax.getData().data(),
+                                     serializedMax.getSize());
+    auto [deserializedMax, remainingMax] = sintType.deserialize(spanMax);
+    auto resultMax = std::any_cast<int64_t>(deserializedMax);
+    EXPECT_EQ(resultMax, maxVal)
+        << "Failed for width " << width << " max value";
+
+    // Test maximum negative value
+    MessageData serializedMin = sintType.serialize(std::any(minVal));
+    std::span<const uint8_t> spanMin(serializedMin.getData().data(),
```
teqdruid: Doesn't MessageData implicitly convert to a span? Why is this necessary? It's not elsewhere.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+                                     serializedMin.getSize());
+    auto [deserializedMin, remainingMin] = sintType.deserialize(spanMin);
+    auto resultMin = std::any_cast<int64_t>(deserializedMin);
+    EXPECT_EQ(resultMin, minVal)
+        << "Failed for width " << width << " min value";
+
+    // Test -1 (all bits set case)
+    MessageData serializedMinusOne =
+        sintType.serialize(std::any(static_cast<int64_t>(-1)));
+    std::span<const uint8_t> spanMinusOne(serializedMinusOne.getData().data(),
```
teqdruid: Same comment here.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+                                          serializedMinusOne.getSize());
+    auto [deserializedMinusOne, remainingMinusOne] =
+        sintType.deserialize(spanMinusOne);
+    auto resultMinusOne = std::any_cast<int64_t>(deserializedMinusOne);
+    EXPECT_EQ(resultMinusOne, -1)
+        << "Failed for width " << width << " value -1";
```
teqdruid, line 301: You can print an error message like this? Get copilot to generate error messages for all of the checks!

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+      {"field2", std::any(static_cast<int64_t>(-10))}};
+  std::any validStruct = std::any(structValue);
+  EXPECT_NO_THROW(structType.isValid(validStruct));
+
+  // Test missing field
+  std::map<std::string, std::any> incompleteStruct = {
+      {"field1", std::any(static_cast<uint64_t>(42))}};
+  std::any invalidStruct = std::any(incompleteStruct);
+  EXPECT_THROW(structType.ensureValid(invalidStruct), std::runtime_error);
+
```
teqdruid, line 328: Add a test for a field which isn't in the type.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+  // Test wrong size array
+  std::vector<std::any> wrongSizeArray = {std::any(static_cast<uint64_t>(10)),
+                                          std::any(static_cast<uint64_t>(20))};
+  std::any invalidArray = std::any(wrongSizeArray);
+  EXPECT_THROW(arrayType.ensureValid(invalidArray), std::runtime_error);
+
+  // Test serialization
+  MessageData serialized = arrayType.serialize(validArray);
+  EXPECT_EQ(serialized.getSize(), 3UL); // 1 byte per element
+
```
teqdruid, line 409: Add a check of the serialized bytes, please.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+  StructType structType("testStruct", fields);
+  EXPECT_EQ(structType.getBitWidth(), 24); // 8 + 16
+
+  // Test array bit width
+  ArrayType arrayType("uint8Array", uintType8.get(), 5);
+  EXPECT_EQ(arrayType.getBitWidth(), 40); // 8 * 5
+}
+
+// Test round-trip serialization/deserialization
+TEST(ESITypesTest, RoundTripSerialization) {
```
teqdruid: This check seems redundant with the other tests.

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+  // Test missing field
+  std::map<std::string, std::any> incompleteStruct = {
+      {"field1", std::any(static_cast<uint64_t>(42))}};
+  std::any invalidStruct = std::any(incompleteStruct);
+  EXPECT_THROW(structType.ensureValid(invalidStruct), std::runtime_error);
+
+  // Test serialization
+  MessageData serialized = structType.serialize(validStruct);
+  EXPECT_EQ(serialized.getSize(), 2UL); // 1 byte for each field
+
```
teqdruid, line 340: Please add a serialized bytes check.

teqdruid (approved):
In struct and array type (de-)serialization: I don't think that this supports fields / elements of `bitwidth % 8 != 0`. That case is not common enough to bother supporting (as it is quite complex) but it's gotta be checked as it does occur.

It's not checked elsewhere since it's easier to support as a packed struct which is statically generated. This limitation is specific to the dynamic encoding here.

Nice tests. Congratulations, your dynamic serialization is the best tested part of the ESI runtime. (Not that that's saying much.)

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
mortbopet: Woops, forgot to change to `ensureValid`!

unittests/Dialect/ESI/runtime/ESIRuntimeTest.cpp
mortbopet: AI on it 🫡

Issue #8873, opened 2025-08-20
[Comb] Crash in `assumeMuxCondInOperand`.


uenoku:
```
builtin.module(sv-extract-test-code,hierarchical-runner{top-name=SiFive_CoreIPSubsystem pipeline=synth-aig-lowering-pipeline,synth-aig-optimization-pipeline})
2025-08-19 17:09:40 - __main__ - INFO - Parsing and running synthesis pipeline...
 #0 0x00007ff936bfde78 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (circt-synth/mlirdev/.env/lib/python3.13/site-packages/circt/_mlir_libs/libCIRCTBindingsPythonCAPI.so+0x103fe78)
 #1 0x00007ff936bfb024 SignalHandler(int, siginfo_t*, void*) Signals.cpp:0:0
 #2 0x00007ff948426cf0 __restore_rt (/lib64/libpthread.so.0+0x12cf0)
 #3 0x00007ff936fadbf7 assumeMuxCondInOperand(mlir::Value, mlir::Value, bool, mlir::PatternRewriter&) CombFolds.cpp:0:0
 #4 0x00007ff936fb5f2f (anonymous namespace)::MuxRewriter::matchAndRewrite(circt::comb::MuxOp, mlir::PatternRewriter&) const CombFolds.cpp:0:0
```

Maybe `isa<CombDialect>(op->getDialect())` crashes when `op` is an unknown dialect?

dtzSiFive:
Good call!

Yeah thats right, if that's not this issue it sounds wrong anyway.

https://mlir.llvm.org/doxygen/classmlir_1_1Operation.html#a288119cc7050878064090a5c21ed3e54

Just to confirm.

PR #8880, opened 2025-08-22
Use isa_and_nonnull for op->getDialect()

uenoku:
op->getDialect() returns nullptr for unregistered dialect so generally it's preferable to use isa_and_nonnull. 

Fix https://github.com/llvm/circt/issues/8873.


fabianschuiki:
Thansk for fixing this :pray:

PR #8881, opened 2025-08-22
[circt-synth] Parallel Prefix Lowering Performance Bug Fix

cowardsa:
Small fix for the construction of the parallel prefix tree - previously the code was overwriting values that got used in the same outer loop iteration creating unnecessarily deep logic circuits. Interestingly they were still functional adder circuits so were not detected by existing tests - only detected via performance benchmarking.


uenoku (approved):
LGTM!

PR #8883, opened 2025-08-22
[Moore] Add more precise conversion ops

fabianschuiki:
Extend the Moore dialect with the following new conversion operations:

- `moore.packed_to_sbv` to convert any packed type to its simple bit vector equivalent (i.e., an integer);
- `moore.sbv_to_packed` to do the reverse;
- `moore.logic_to_int` to convert from a four-valued integer to a two-valued integer;
- `moore.int_to_logic` to convert from a two-valued integer to a four-valued integer; and
- `moore.to_builtin_bool` to convert from a single bit `moore::IntType` to a single bit builtin `mlir::IntegerType`.

Adjust the `ImportVerilog` pass to generate these new conversion ops instead of the old whole-sale `moore.conversion`. Having smaller, more precise conversion ops will make lowering to the core dialects easier in the future, especially once we fully support four-valued types and time values.

In the future, we may also want to create conversion ops from and to some of the unpacked types such as strings, unpacked arrays, queues, etc. These will require special care in the lowering, likely expanding to runtime function calls.

The long-term plan is to get rid of `moore.conversion` altogether and replace it with simpler, dedicated conversion ops.

Most of this commit is just mechanically threading-through the new cast operations. They all lower to the same bitcast (or a passthrough) in HW as the `moore.conversion` op. This should result in no effective change at the core dialect level.

I ran this on [circt-tests] and the Snitch core in [arc-tests], with no change in output.

[circt-tests]: github.com/circt/circt-tests
[arc-tests]: github.com/circt/arc-tests

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -41,6 +41,13 @@ class TypeDomainsMatch<list<string> values> :
     cast<moore::UnpackedType>($_self.getType()).getDomain()
   }], "domain">;
 
+class TypeIsSimpleBitVectorOf<string sbvType, string packedType> :
+  TypesMatchWith<
+    sbvType # " type is simple bit vector equivalent of " # packedType,
+    sbvType, packedType, [{
+      cast<PackedType>($_self).getSimpleBitVector()
+    }]>;
+
 //===----------------------------------------------------------------------===//
 // Structure
 //===----------------------------------------------------------------------===//
@@ -564,6 +571,70 @@ def ConversionOp : MooreOp<"conversion", [Pure]> {
     $input attr-dict `:` type($input) `->` type($result)
   }];
   let hasFolder = 1;
+}
+
+def PackedToSBVOp : MooreOp<"packed_to_sbv", [
+  Pure,
+  TypeIsSimpleBitVectorOf<"input", "result">,
+]> {
+  let summary = "Convert a packed type to its simple bit vector equivalent";
+  let arguments = (ins PackedTypeWithKnownSize:$input);
+  let results = (outs IntType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input)
+  }];
+}
+
+def SBVToPackedOp : MooreOp<"sbv_to_packed", [
+  Pure,
+  TypeIsSimpleBitVectorOf<"result", "input">,
+]> {
+  let summary = "Convert a simple bit vector to an equivalent packed type";
+  let arguments = (ins IntType:$input);
+  let results = (outs PackedTypeWithKnownSize:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($result)
+  }];
+}
+
+def LogicToIntOp : MooreOp<"logic_to_int", [
+  Pure,
+  TypesMatchWith<"result matches input with two-valued domain",
+    "input", "result", "cast<IntType>($_self).getTwoValued()">
+]> {
+  let summary = "Convert a four-valued to a two-valued integer";
+  let arguments = (ins FourValuedIntType:$input);
+  let results = (outs TwoValuedIntType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input)
+  }];
+  let hasFolder = 1;
+}
+
+def IntToLogicOp : MooreOp<"int_to_logic", [
+  Pure,
+  TypesMatchWith<"result matches input with four-valued domain",
+    "input", "result", "cast<IntType>($_self).getFourValued()">
+]> {
+  let summary = "Convert a two-valued to a four-valued integer";
+  let arguments = (ins TwoValuedIntType:$input);
+  let results = (outs FourValuedIntType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input)
+  }];
+  let hasFolder = 1;
+}
+
+def ToBuiltinBoolOp : MooreOp<"to_builtin_bool", [Pure]> {
+  let summary = "Convert a `!moore.l1` or `!moore.i1` to a builtin `i1`";
+  let description = [{
+    Maps `X` and `Z` to 0, and passes through 1 and 0 unchanged.
+  }];
+  let arguments = (ins AnySingleBitType:$input);
+  let results = (outs I1:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input)
+  }];
 }
 
 def TruncOp : MooreOp<"trunc", [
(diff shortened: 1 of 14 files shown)


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -566,6 +573,70 @@ def ConversionOp : MooreOp<"conversion", [Pure]> {
...
+  }];
+  let hasFolder = 1;
+}
+
+def IntToLogicOp : MooreOp<"int_to_logic", [
+  Pure,
+  TypesMatchWith<"result matches input with four-valued domain",
+    "input", "result", "cast<IntType>($_self).getFourValued()">
+]> {
+  let summary = "Convert a four-valued to a two-valued integer";
```
maerhart: Should be the other way around

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -1050,6 +1050,39 @@ OpFoldResult ConversionOp::fold(FoldAdaptor adaptor) {
...
+
+  return {};
+}
+
+//===----------------------------------------------------------------------===//
+// IntToLogicOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult IntToLogicOp::fold(FoldAdaptor adaptor) {
+  // int_to_logic(logic_to_int(x)) -> x
```
maerhart: Wouldn't that map all X and Z to 0 rather than being the identity function?

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -1050,6 +1050,39 @@ OpFoldResult ConversionOp::fold(FoldAdaptor adaptor) {
...
+  if (auto reverseOp = getInput().getDefiningOp<LogicToIntOp>())
+    return reverseOp.getInput();
+
+  // Simply pass through constants.
+  if (auto intInput = dyn_cast_or_null<FVIntegerAttr>(adaptor.getInput()))
+    return intInput;
+
+  return {};
+}
+
```
maerhart, line 1084: There don't seem to be tests for the folders.

lib/Dialect/Moore/MooreTypes.cpp
```diff
@@ -115,6 +116,7 @@ std::optional<unsigned> PackedType::getBitSize() const {
   return TypeSwitch<PackedType, std::optional<unsigned>>(*this)
       .Case<VoidType>([](auto) { return 0; })
       .Case<IntType>([](auto type) { return type.getWidth(); })
+      .Case<TimeType>([](auto type) { return 64; })
```
maerhart, line 119: Is this tested somewhere?

maerhart (approved):
Having dedicated conversion ops for the various kinds of conversions makes sense to me 👍

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -566,6 +573,70 @@ def ConversionOp : MooreOp<"conversion", [Pure]> {
...
+]> {
+  let summary = "Convert a packed type to its simple bit vector equivalent";
+  let arguments = (ins PackedTypeWithKnownSize:$input);
+  let results = (outs IntType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input) `->` type($result)
+  }];
+}
+
+def SBVToPackedOp : MooreOp<"sbv_to_packed", [
```
jpienaar, line 588: You could add InferTypeOpInterface  to these as you can deduce the correct output type given input - not blocking here.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1377,39 +1367,52 @@ Value Context::convertToSimpleBitVector(Value value) {
...
-        value = moore::ZExtOp::create(builder, loc, dstWidthType, value);
+        value = moore::ZExtOp::create(builder, loc, resizedType, value);
+    }
+
+    // Convert the domain if needed.
+    if (dstInt.getDomain() != srcInt.getDomain()) {
+      if (dstInt.getDomain() == moore::Domain::TwoValued)
+        value = moore::LogicToIntOp::create(builder, loc, value);
+      else if (dstInt.getDomain() == moore::Domain::FourValued)
+        value = moore::IntToLogicOp::create(builder, loc, value);
```
jpienaar, line 1404: These are the only two cases here right? E.g., post this one requires the domain's to match.

lib/Conversion/ImportVerilog/Statements.cpp
```diff
@@ -288,8 +287,9 @@ struct StmtVisitor {
 
         // Take note if the expression is a constant.
         auto maybeConst = value;
-        if (auto defOp = maybeConst.getDefiningOp<moore::ConversionOp>())
-          maybeConst = defOp.getInput();
+        while (isa_and_nonnull<moore::ConversionOp, moore::LogicToIntOp,
```
jpienaar: Micro nit sort in alphabetical order.

jpienaar (approved):
Looks good modulo the folder which seems to change semantics.

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: Fantastic catch, thanks! 🥳

lib/Dialect/Moore/MooreOps.cpp
fabianschuiki: Good point. I'll remove the `logic_to_int(int_to_logic(x)) -> x` fold, since that would map X/Z to 0 as you describe. I'll leave a comment in the code to make sure we don't accidentally add this in the future.

fabianschuiki: Actually the other way around, as you said: `int_to_logic(logic_to_int(x)) -> x` can't be done. Got myself confused by the parenthesis and the nesting 😅

lib/Dialect/Moore/MooreOps.cpp
fabianschuiki: I forgot about those. Adding them now. 👍

lib/Dialect/Moore/MooreTypes.cpp
fabianschuiki: Added an explicit use of `packed_to_sbv` and `sbv_to_packed` with a time type, which calls this to ensure the types match.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Exactly! 👍

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: This is already done implicitly by `TypesMatchWith` through `TypeIsSimpleBitVectorOf`. Let me drop the redundant types from the assembly format 👍

PR #8884, opened 2025-08-23
[Moore] Add WaitDelayOp, support delay control in ImportVerilog

fabianschuiki:
Add a `moore.wait_delay` operation to the Moore dialect. This op simply suspends execution for an amount of time determined by its operand. This lowers directly to `llhd.wait delay %0, ^bb0`. Make use of this new op in ImportVerilog by adding support for basic `#10ns` delay control statements.


maerhart (approved):
LGTM

PR #8887, opened 2025-08-25
Bump slang to dd16a7947e0586d0541477f1b4b60eda7c986e35

jpienaar:
We encountered some failures that were addressed since 9.0 tag, given no API changes seemed like good to bump to new version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -564,8 +564,7 @@ if(CIRCT_SLANG_FRONTEND_ENABLED)
     FetchContent_Declare(
       slang
       GIT_REPOSITORY https://github.com/MikePopoloski/slang.git
-      GIT_TAG dd16a7947e0586d0541477f1b4b60eda7c986e35 # v9.0
-      GIT_SHALLOW ON)
+      GIT_TAG dd16a7947e0586d0541477f1b4b60eda7c986e35)
     set(FETCHCONTENT_TRY_FIND_PACKAGE_MODE "NEVER")
 
     # Force Slang to be built as a static library to avoid messing around with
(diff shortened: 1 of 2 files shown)


fabianschuiki:
Sounds good to me! Thanks for establishing a precedence to pull in up-to-date Slang versions similar to how we bump LLVM :100:

jpienaar:
@fabianschuiki with this change though I get the following diff which I'm not sure if acceptable/bug introduced 
image

From

```verilog
module GenerateConstructs;
  genvar i;
  parameter p = 2;
  
  generate
    // CHECK: [[TMP:%.+]] = moore.constant 0
    // CHECK: dbg.variable "i", [[TMP]]
    // CHECK: [[TMP:%.+]] = moore.constant 0
    // CHECK: %g1 = moore.variable [[TMP]]
    // CHECK: [[TMP:%.+]] = moore.constant 1
    // CHECK: dbg.variable "i", [[TMP]]
    // CHECK: [[TMP:%.+]] = moore.constant 1
    // CHECK: moore.variable name "g1" [[TMP]]
    for (i = 0; i < 2; i = i + 1) begin
      integer g1 = i;
    end
```

I can just update the test but unsure if expected.

fabianschuiki:
That diff looks okay. It seems like Slang now calls unnamed generate blocks `genblk`, which feels like a reasonable thing to do.

jpienaar:
Seems the cmake fetch fails if one uses a hash that's not near tip (https://gitlab.kitware.com/cmake/cmake/-/issues/23379), given its opt in, would making it non-shallow be fine?

fabianschuiki:
I'm definitely fine with doing a non-shallow fetch. Maybe keep the shallow option as a comment for whenever we switch to a specific version tag?

fabianschuiki (approved):
LGTM once CI is happy 👍 Thanks for keeping us up-to-date with Slang bug fixes!

PR #8890, opened 2025-08-25
[AIGLongestPathAnalysis] Implement incremental longest path analysis for IR transformations

uenoku:
This commit introduces IncrementalLongestPathAnalysis to support lazy delay computation during IR mutations in synthesis passes, with the DatapathToComb conversion pass as the first use case.

The implementation adds the IncrementalLongestPathAnalysis class that extends the existing analysis with incremental capabilities through the PatternRewriter::Listener interface to track mutations and maintain validity. New getOrComputeDelay() and getOrComputePaths() methods enable on-demand computation, while an OperationAnalyzer class handles dynamic analysis of Comb and HW dialect operations through temporary module synthesis.

The analysis requires a top-down rewriting approach to maintain correctness, as bottom-up mutations can invalidate paths that depend on erased values. The incremental framework validates mutations to prevent breaking existing timing paths and provides early error detection for inappropriate modifications to critical path values.

### DIFF
--- a/lib/Conversion/DatapathToComb/DatapathToComb.cpp
+++ b/lib/Conversion/DatapathToComb/DatapathToComb.cpp
@@ -316,4 +316,16 @@ void ConvertDatapathToCombPass::runOnOperation() {
   if (failed(applyPatternsGreedilyWithTimingInfo(
           getOperation(), std::move(patterns), analysis)))
     return signalPassFailure();
+
+  // Verify that all Datapath operations have been successfully converted.
+  // Walk the operation and check for any remaining Datapath dialect operations.
+  auto result = getOperation()->walk([&](Operation *op) {
+    if (llvm::isa_and_nonnull<datapath::DatapathDialect>(op->getDialect())) {
+      op->emitError("Datapath operation not converted: ") << *op;
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  if (result.wasInterrupted())
+    return signalPassFailure();
 }
--- a/lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
+++ b/lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
@@ -1339,6 +1339,10 @@ OperationAnalyzer::getOrComputeLocalVisitor(Operation *op) {
 
   // Connect module inputs to cloned operation operands
   // Handle type mismatches with bitcast operations
+  // Type mismatches can occur when the original operation uses non-integer
+  // types (e.g., structs, arrays) that get converted to integer types for the
+  // wrapper module ports. Since we use the same bit width via
+  // hw::getBitWidth(), bitcast is safe for bit-compatible types.
   for (auto arg : hwModule.getBodyBlock()->getArguments()) {
     Value input = arg;
     auto idx = arg.getArgNumber();


lib/Conversion/DatapathToComb/DatapathToComb.cpp
```diff
@@ -73,17 +79,45 @@ struct DatapathCompressOpConversion : OpConversionPattern<CompressOp> {
...
       addends.push_back(
           extractBits(rewriter, input)); // Extract bits from each input
+
+      // NOTE: Following change will be splitted into a separate PR.
+      if (analysis) {
+        auto delay = analysis->getOrComputePaths(input, 0);
+        if (failed(delay))
+          return op.emitError("Failed to get delay for input");
+        // TODO: Use the delay information to sort the inputs.
+      }
```
cowardsa: Here you only check for bitpos 0 delay information - within the compressor tree synthesis (and in many other lowering passes) we will want bit-level delay information for every bit - do you expect if we have bitpos 0 then we will also have all other bits of the input?

lib/Conversion/DatapathToComb/DatapathToComb.cpp
```diff
@@ -254,6 +288,22 @@ struct ConvertDatapathToCombPass
 };
 } // namespace
 
+static LogicalResult
+applyConversionWithTimingInfo(Operation *op, const ConversionTarget &target,
+                              RewritePatternSet &&patterns,
+                              aig::IncrementalLongestPathAnalysis *analysis) {
```
cowardsa: Presumably we expect this to be computationally expensive? Would it be safer to make timing aware synthesis optional?

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -73,7 +74,7 @@
 using namespace circt;
 using namespace aig;
 
-static size_t getBitWidth(Value value) {
+static int64_t getBitWidth(Value value) {
```
cowardsa, line 80: Nit: presumably hw::getBitWidth checks for positive bitwidth?

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -1456,7 +1506,13 @@ LongestPathAnalysis::~LongestPathAnalysis() { delete impl; }
 LongestPathAnalysis::LongestPathAnalysis(
     Operation *moduleOp, mlir::AnalysisManager &am,
     const LongestPathAnalysisOption &option)
-    : impl(new Impl(moduleOp, am, option)), ctx(moduleOp->getContext()) {}
+    : impl(new Impl(moduleOp, am, option)), ctx(moduleOp->getContext()) {
+  llvm::errs() << "LongestPathAnalysis created\n";
```
cowardsa: Nit: why llvm::errs() not llvm::dbgs()?

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -1506,6 +1562,82 @@ ArrayRef<hw::HWModuleOp> LongestPathAnalysis::getTopModules() const {
...
+  }
+}
+
+void IncrementalLongestPathAnalysis::notifyOperationErased(Operation *op) {
+  for (auto value : op->getResults()) {
+    if (isValueValidToErase(value)) {
+      mlir::emitError(value.getLoc())
+          << "value " << value
+          << " replaced but contained in longest path analysis. It's required "
+             "to perform the IR mutation from buttom-up";
```
cowardsa: Nit: "bottom-up"

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -1506,6 +1562,82 @@ ArrayRef<hw::HWModuleOp> LongestPathAnalysis::getTopModules() const {
...
+}
+
+void IncrementalLongestPathAnalysis::notifyOperationReplaced(
+    Operation *op, ValueRange replacement) {
+  for (auto value : op->getResults()) {
+    if (isValueValidToErase(value)) {
+      mlir::emitError(value.getLoc())
+          << "value " << value
+          << " replaced but contained in longest path analysis. It's required "
+             "to perform the IR mutation from buttom-up";
```
cowardsa: Nit: "bottom-up"

cowardsa (commented):
Thanks for this connection Hideto - looks really valuable for improving synthesis results! I also like the interface but wanted to clarify usage a little:

1. Is it the responsibility of the pass author to call notifyOperation... whenever you make changes to the IR? Or is this automatically hooked up with existing MLIR infra?
2. Does knowing precisely where changes have been made (due to IR transformations) mean we can only look at local changes to the delay paths? 

P.s. if you hook up the comb delay model interface then others (myself included) can add additional or more precise models as needed - similar to the KnownBits Analysis

lib/Conversion/DatapathToComb/DatapathToComb.cpp
uenoku: Ah this is mistake. I thought I iterated over elements of addends (so everything is single bit). Every bitpos is supported so both `getDelay(input, i)` and `getDelay(addends[j][i], 0)` should work.

uenoku:
> Is it the responsibility of the pass author to call notifyOperation... whenever you make changes to the IR? Or is this automatically hooked up with existing MLIR infra?

Yes, it's user's responsibility to notify the mutation but MLIR's DialectConversion/PartialConversion/PatternRewriter automatically send notification to the listener as long as the listener is registered (`config.listener = analysis;` is the corresponding code). So it's shouldn't be a much problem if MLIR infra was used.

> Does knowing precisely where changes have been made (due to IR transformations) mean we can only look at local changes to the delay paths?

Yes, that's current limitation. Currently IncrementalLongestPathAnalysis will only accept local scope (hw::HWModuleOp).
User could use normal LongestPathAnalysis for mlir::ModuleOp but it doesn't support incremental updates now. 

For inter-module timing analysis I'm inclined to prefer `synth.required_times/arrival_times` operation (or attribute?) to attach timing information across modules (similar SDC) so that local timing analysis can use information of global timing analysis, and periodically sync the timing info globally.  
```mlir
hw.module @Foo(in %x: i3) {
    // bit 0/1/2 arrive at time 1/5/10.
    %x_with_timing = synth.arrival_times %x [10, 5, 1] : i3
    %out = hw.instance @Bar(%x_with_timing): (i3) -> i2
    %out_with_timing = synth.arrival_times %out_timing [100, 20] : i2
    
}
```

lib/Conversion/DatapathToComb/DatapathToComb.cpp
uenoku: Yes, I'll add an option.

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: `hw::getBitWidth` return int64_t and could return -1 if the bitwidth is unknown, so it's safer to return int64_t here as well

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: This was for debugging, I'll remove them :) Thank you for pointing out!

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: It was actually top-down :sweat_smile: Fixed it, thank you for pointing it out.

uenoku:
@cowardsa LongestPathAnalysis.cpp requires more clean up and testing, so not ready for review but the analysis now returns timing model (= logic depth in AIG) for most of comb/hw operations (so it should be usable for datapath lowering). 
Certainly we can introduce an op interface to override some of the analysis but I'd like to do that in the follow-up.  
I made changes on DatapathToComb which requires analysis to work. Specifically I replaced DialectConversion with GreedyRewirter due to the lack of top-down rewrite in DialectConversion. I hope that's fine change. 

This is a change of AIG depth for your benchmark set for the current PR FYI.  
```
add_three 12 -> 12
blend 49 -> 49
dot_product 43 -> 43
-----------
fmaa 72 -> 67
fmaa_shae 45 -> 43
fma 64  -> 61
```

cowardsa:
This is fantastic! Very pleasing to see this play out in the results as well - you need the top-down traversal so that we have lowered everything which the current operator depends on to ensure that we can compute the timing information. Is that correct? I actually like this a lot as it also means that partial_product operators will be lowered before we reach the compressor trees that consume their outputs. Nice!

I'm hoping to have a timing-driven compressor tree PR ready today or early next week - but will perhaps separate out the datapathToComb changes to avoid conflicts :)

uenoku:
>  ensure that we can compute the timing information. Is that correct? 

Exactly! It's necessary to topologically sort the IR beforehand because HWModule has a graph region but after that top-down lowering should ensure that operand is lowered before the operation.

lib/Dialect/AIG/Analysis/PrintLongestPathAnalysis.cpp
```diff
@@ -273,7 +273,12 @@ void PrintLongestPathAnalysisPass::printPathHistory(const OpenPath &timingPath,
...
 
 void PrintLongestPathAnalysisPass::runOnOperation() {
-  auto &analysis = getAnalysis<aig::LongestPathAnalysisWithTrace>();
+  LongestPathAnalysis *analysis;
+  if (showTopKPercent != 0) {
+    analysis = &getAnalysis<aig::LongestPathAnalysisWithTrace>();
+  } else {
+    analysis = &getAnalysis<aig::LongestPathAnalysis>();
+  }
   igraph::InstancePathCache pathCache(
```
uenoku, line 282: This change is unrelated to the PR itself, will commit separately.

include/circt/Dialect/Synth/Transforms/SynthesisPipeline.h
```diff
@@ -31,6 +31,10 @@ struct AIGLoweringPipelineOptions
       *this, "disable-datapath",
       llvm::cl::desc("Disable datapath optimization passes"),
       llvm::cl::init(false)};
+  PassOptions::Option<bool> timingAware{
+      *this, "timing-aware",
+      llvm::cl::desc("Lower operators in a timing-aware fashion"),
+      llvm::cl::init(false)};
```
cowardsa, line 37: Would actually advocate for this being true by default as we've enabled Datapath optimisations by default.

lib/Conversion/DatapathToComb/DatapathToComb.cpp
```diff
@@ -76,29 +82,49 @@ struct DatapathCompressOpConversion : OpConversionPattern<CompressOp> {
...
+          delays.push_back(std::make_pair(*delay, addend[j]));
+        }
+        std::stable_sort(delays.begin(), delays.end(),
+                         [](const std::pair<int64_t, Value> &a,
+                            const std::pair<int64_t, Value> &b) {
+                           return a.first < b.first;
+                         });
+        for (size_t i = 0; i < addends.size(); ++i)
+          addends[i][j] = delays[i].second;
+      }
```
cowardsa, line 107: FYI - will pass the "delays" directly to the constructor in an upcoming PR - but current placeholder looks good

lib/Conversion/DatapathToComb/DatapathToComb.cpp
```diff
@@ -254,25 +280,40 @@ struct ConvertDatapathToCombPass
...
+static LogicalResult applyPatternsGreedilyWithTimingInfo(
+    Operation *op, RewritePatternSet &&patterns,
+    aig::IncrementalLongestPathAnalysis *analysis) {
+  // TODO: Topologically sort the operations in the module to ensure that all
+  // dependencies are processed before their users.
+  mlir::GreedyRewriteConfig config;
+  // Set the listener to update timing information
+  // HACK: Setting max iterations to 2 to ensure that the patterns are one-shot,
+  // making sure target operations are datapath operations are replaced.
+  config.setMaxIterations(2).setListener(analysis).setUseTopDownTraversal(true);
```
cowardsa, line 292: Is it possible to add a check after applying the rewriter to ensure that there are no "Illegal" operators from the Datapath dialect - to ensure we satisfy the conditions imposed by the conversion pass?

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -409,19 +421,102 @@ class Context {
...
+// haven't been converted to AIG yet. It creates isolated modules for each
+// unique operation type/signature combination, runs the conversion pipeline
+// (HW -> Comb -> AIG), and analyzes the resulting AIG representation.
+//
+// This is used as a fallback when the main analysis encounters operations
+// that don't have direct AIG equivalents. The analyzer:
+// 1. Creates a wrapper HW module for the operation
+// 2. Runs the standard conversion pipeline to lower it to AIG
+// 3. Analyzes the resulting AIG to compute timing paths
+// 4. Caches results based on operation name and function signature
```
cowardsa, line 455: This is great! It really nicely captures any improvements to the lowering code itself e.g. adding a new faster adder architecture! This assumes uniform arrival times for all inputs is that correct?

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -409,19 +421,102 @@ class Context {
...
+
+  // Standard conversion pipeline: HW -> Comb -> AIG -> cleanup passes
+  // This pipeline is applied to wrapper modules to prepare them for analysis
+  constexpr static StringRef pipelineStr =
+      "hw.module(hw-aggregate-to-comb,convert-comb-to-aig,cse,canonicalize)";
+  std::unique_ptr<mlir::PassManager> passManager;
+
+  // Temporary module used to hold wrapper modules during analysis
+  // Each operation gets its own wrapper module created inside this parent
+  mlir::OwningOpRef<mlir::ModuleOp> moduleOp;
```
cowardsa, line 508: Non-blocking! Presumably the scope is a single operation, therefore no input limits are known e.g. if I perform:
`{4'd0, a} + {4'd0, b}`

Does this get viewed as an 8-bit adder for the purpose of this local lowering? I think this is a perfectly reasonable over-approximation for guiding initial lowering decisions. If we envision a future iterative timing closure loop then we may need a more accurate delay model.

cowardsa (commented):
Incremental review - will continue but hit a weird GitHub interface bug...

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -1129,6 +1263,168 @@ const LocalVisitor *Context::getAndWaitLocalVisitor(StringAttr name) const {
...
+// OperationAnalyzer
+// ===----------------------------------------------------------------------===//
+
+FailureOr<LocalVisitor *>
+OperationAnalyzer::getOrComputeLocalVisitor(Operation *op) {
+  // Check cache first.
+  auto opName = op->getName();
+  auto functionType = getFunctionTypeForOp(op);
+  auto key = std::make_pair(opName, functionType);
+  auto it = cache.find(key);
```
cowardsa, line 1285: Nice!

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
```diff
@@ -1129,6 +1263,168 @@ const LocalVisitor *Context::getAndWaitLocalVisitor(StringAttr name) const {
...
+    Value input = arg;
+    auto idx = arg.getArgNumber();
+
+    // Insert bitcast if input port type differs from operand type
+    if (input.getType() != cloned->getOperand(idx).getType())
+      input = builder.create<hw::BitcastOp>(
+          op->getLoc(), cloned->getOperand(idx).getType(), input);
+
+    cloned->setOperand(idx, input);
+  }
```
cowardsa: Q: Why do we get type mismatches? Is this always safe to cast to different bitwidths?

cowardsa (approved):
This is really great! I'll be interested to test the performance but the interface looks ideal and will track changes to the lowering code. The tests are also very nice - look forward to seeing how we might incorporate this across the synthesis pipeline :)

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: Yes, input arrival times are assumed to be uniform.

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: > Does this get viewed as an 8-bit adder for the purpose of this local lowering?

Yes, constants are not visible so it will be viewed as 8-bit adder. Maybe we can include the known bits as a part of analysis, e.g. `cache[{opName, funcType, inputKnownBits]}` that makes things more accurate.

lib/Conversion/DatapathToComb/DatapathToComb.cpp
uenoku: Added in `runOnOperation` :+1:

lib/Dialect/AIG/Analysis/LongestPathAnalysis.cpp
uenoku: LongestPathAnalysis doesn't look dataflow through array/struct, so it's necessary to use integers for ports. Integers will be bit-cast to original type and they always have the same bitwidth.

PR #8891, opened 2025-08-25
[FIRRTL] Lower-Layers: capture operands of subaccess op

rwy7:
When we don't need to (or, aren't able to) hoist a subaccess op out of a layer block, because the input operand is passive, we still need to capture the operands.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -608,26 +608,61 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
       // transparently referencing a spilled op, spill the node, too.  The node
       // provides an anchor for an inner symbol (which subfield, subindex, and
       // subaccess do not).
-      if (isa<SubfieldOp, SubindexOp>(op)) {
-        auto input = op->getOperand(0);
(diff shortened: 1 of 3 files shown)


test/Dialect/FIRRTL/lower-layers.mlir
```diff
@@ -219,6 +219,54 @@ firrtl.circuit "Test" {
     }
   }
 
+  // Test that subfield and subindex ops are captured as normal when the input is passive.
+  //
```
dtzSiFive, line 223: Nice tests! 💯

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -608,26 +608,61 @@ LogicalResult LowerLayersPass::runOnModuleBody(FModuleOp moduleOp,
...
+        }
+
+        // Otherwise, move the subaccess op out of the layerblock, if possible.
+        if (!isAncestorOfValueOwner(layerBlock, index)) {
           subOp->moveBefore(layerBlock);
           spilledSubOps.insert(op);
           return WalkResult::advance();
         }
+
+        // When the input is not passive, but the index is defined inside this
```
dtzSiFive, line 664: Already error test here, woohoo!

fabianschuiki (approved):
LGTM!

PR #8893, opened 2025-08-25
[Sim] Create clocked and procedural sim.exit/pause ops

fabianschuiki:
Combine the existing `sim.finish` and `sim.fatal` ops into a single `sim.clocked_exit` op. Create a `sim.exit` op for the procedural case. This op now accepts a `success` operand to indicate whether the simulation should terminate with a success or failure exit code, and a `verbose` operand to indicate whether the exit should behave like a `$finish(0)` or a `$finish(1)` in SystemVerilog.

Also add a `sim.pause` and `sim.clocked_pause` operation which only suspends simulation and gives control back to the user in an interactive simulation. This corresponds to `$stop` in SystemVerilog.

These new ops will allow circt-verilog to lower `$finish`, `$fatal`, `$stop`, and `$exit` in the input.

Adjust the Sim-to-SV conversion pass to support the new operations. Also try to reuse existing ifdef and always/if ops, effectively combining multiple consecutive Sim ops into a single ifdef guard or always process.


include/circt/Dialect/Sim/SimOps.td
```diff
@@ -379,4 +361,96 @@ def PrintFormattedProcOp : SimOp<"proc.print"> {
...
 
+//===----------------------------------------------------------------------===//
+// Simulation Control
+//===----------------------------------------------------------------------===//
+
+def ClockedExitOp : SimOp<"clocked_exit"> {
+  let summary = "Terminate the simulation";
+  let description = [{
+    Implements the semantics of `sim.exit` if the given condition is true during
+    the rising edge of the clock operand.
```
fzi-hielscher: Ultra nit: As a clock edge represents an event and not a time interval I prefer "on the rising edge" instead of "during".

include/circt/Dialect/Sim/SimOps.td
```diff
@@ -379,4 +361,96 @@ def PrintFormattedProcOp : SimOp<"proc.print"> {
...
+    custom<KeywordBool>($success, "\"success\"", "\"failure\"") `,`
+    custom<KeywordBool>($verbose, "\"verbose\"" , "\"quiet\"") attr-dict
+  }];
+}
+
+def ClockedPauseOp : SimOp<"clocked_pause"> {
+  let summary = "Pause the simulation";
+  let description = [{
+    Implements the semantics of `sim.pause` if the given condition is true
+    during the rising edge of the clock operand.
```
fzi-hielscher: As above.

include/circt/Dialect/Sim/SimOps.td
```diff
@@ -379,4 +361,96 @@ def PrintFormattedProcOp : SimOp<"proc.print"> {
...
+    BoolAttr:$verbose
+  );
+  let assemblyFormat = [{
+    $clock `,` $condition `,`
+    custom<KeywordBool>($verbose, "\"verbose\"" , "\"quiet\"") attr-dict
+  }];
+}
+
+def ExitOp : SimOp<"exit"> {
+  let summary = "Terminate the simulation";
```
fzi-hielscher: Can't we just call it `TerminateOp` to avoid the confusion with `$exit`?

lib/Conversion/SimToSV/SimToSV.cpp
```diff
@@ -36,6 +36,20 @@ namespace circt {
...
+/// from affecting synthesis runs.
+static bool needsIfdefGuard(Operation *op) {
+  return isa<ClockedExitOp, ClockedPauseOp, ExitOp, PauseOp>(op);
+}
+
+/// Check whether an op should be placed inside an always process triggered on a
+/// clock, and an if statement checking for a condition.
+static std::pair<Value, Value> needsClockAndConditionWrapper(Operation *op) {
+  if (isa<ClockedExitOp, ClockedPauseOp>(op))
+    return {op->getOperand(0), op->getOperand(1)};
```
fzi-hielscher: Nit: Maybe use a `TypeSwitch` to avoid hardcoding operand numbers?

lib/Conversion/SimToSV/SimToSV.cpp
```diff
@@ -315,6 +326,96 @@ void LowerDPIFunc::addFragments(hw::HWModuleOp module,
...
+                              guardOp.hasElse())
+                            return guardOp.getElseBlock();
+                          return nullptr;
+                        })
+                    .Default([](auto) { return nullptr; });
+
+      // If there was no pre-existing guard, create one.
+      if (!block) {
+        OpBuilder builder(op);
+        if (op->getParentOp()->hasTrait<sv::ProceduralRegion>())
```
fzi-hielscher, line 357: Is this sufficient at this point? What if the parent op has not been lowered to SV yet, but will become a procedural region? I've been struggling in the past with the lack of a cross-dialect notion of procedural regions.

lib/Conversion/SimToSV/SimToSV.cpp
```diff
@@ -315,6 +326,96 @@ void LowerDPIFunc::addFragments(hw::HWModuleOp module,
...
+                      builder, loc, "SYNTHESIS", [] {}, [] {})
+                      .getElseBlock();
+        usedSynthesisMacro = true;
+      }
+
+      // Move the op into the guard block.
+      op->moveBefore(block, block->end());
+    }
+
+    // Check if the op require an clock and condition wrapper.
```
fzi-hielscher:
```suggestion
    // Check if the op requires a clock and condition wrapper.
```

test/Conversion/SimToSV/simulation-control.mlir
```diff
@@ -0,0 +1,52 @@
...
+    sim.exit failure, quiet
+    // CHECK-NEXT: sv.stop 1
+    sim.pause verbose
+    // CHECK-NEXT: sv.stop 0
+    sim.pause quiet
+
+    // CHECK-NEXT: }
+  }
+  // CHECK-NEXT: }
+}
```
fzi-hielscher, line 52: A test that makes sure that blocks that cannot be reused (due to different clocks/conditions) do not get reused would be nice.

include/circt/Dialect/Sim/SimOps.td
fabianschuiki: Good idea :+1:

include/circt/Dialect/Sim/SimOps.td
fabianschuiki: That's a good idea. I've renamed this to `TerminateOp`.

lib/Conversion/SimToSV/SimToSV.cpp
fabianschuiki: I think it is, at least for now. We only ever nest into ifdefs and always processes, which implement the trait. And the verification of SV ops simply checks if the parent has one of these traits.

Maybe we should just switch over to checking if the parent is a graph region or not :thinking:, which is universal across dialects, instead of relying on an SV trait.

test/Conversion/SimToSV/simulation-control.mlir
fabianschuiki: Added :+1:

fabianschuiki:
@fzi-hielscher Yeah I also don't like that we have all these clocked and unclocked flavors of ops. :disappointed:. How about we start by just factoring the clock and condition operand out into a surrounding op in the Sim dialect? Basically a trivial version of `sim.triggered` that just just takes two operands, and then let's see how far we get with that.

PR #8894, opened 2025-08-26
[ExportVerilog] Fix crash for variadic ops with a single operand

7FM:
Fixes https://github.com/llvm/circt/issues/7698 and hopefully doesn't break anything else :upside_down_face:


lib/Conversion/ExportVerilog/PrepareForEmission.cpp
```diff
@@ -1196,7 +1196,7 @@ static LogicalResult legalizeHWModule(Block &block,
     // statements.
     // TODO: This is checking the Commutative property, which doesn't seem
     // right in general.  MLIR doesn't have a "fully associative" property.
-    if (op.getNumOperands() > 2 && op.getNumResults() == 1 &&
```
uenoku, line 1199: Could you update the comment above (`...with more than two operands...`)?

7FM: Done :)

PR #8895, opened 2025-08-27
[MooreToCore] Convert unreachable, stop, finish, finish_message ops

fabianschuiki:
Add conversion patterns for the following ops:

- `moore.unreachable`
- `moore.builtin.stop`
- `moore.builtin.finish`
- `moore.builtin.finish_message`

Also create a `ConversionPatternSet` utility that simplifies creating small conversion patterns. This basically allows us to define patterns simply through a conversion function instead of having to define an entire pattern struct. The `RewritePatternSet` already supports this, but only for functions with the `matchAndRewrite` signature of rewrite patterns:

    LogicalResult (Op, PatternRewriter&)

This new utility allows us to do the same for functions with the usual `matchAndRewrite` signature of conversion patterns, too:

    LogicalResult (Op, Op::Adaptor, ConversionPatternRewriter&)


include/circt/Support/ConversionPatternSet.h
```diff
@@ -0,0 +1,69 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
uenoku, line 7: nit: Please add a header description

include/circt/Support/ConversionPatternSet.h
```diff
@@ -0,0 +1,69 @@
...
+
+#ifndef CIRCT_SUPPORT_UNUSEDOPPRUNER_H
+#define CIRCT_SUPPORT_UNUSEDOPPRUNER_H
```
maerhart: should be CIRCT_SUPPORT_CONVERSIONPATTERNSET_H

uenoku (commented):
I think this is reasonable but ConversionPatternSet is not actually used in the PR yet, is that correct?

include/circt/Support/ConversionPatternSet.h
```diff
@@ -0,0 +1,69 @@
...
+        return implFn(op, adaptor, rewriter);
+      }
+    };
+
+    auto pattern =
+        std::make_unique<FnPattern>(typeConverter, patterns.getContext());
+    pattern->implFn = implFn;
+    patterns.add(std::move(pattern));
+    return *this;
+  }
```
maerhart: This part doesn't seem to be used, thus also not tested. Maybe it'd be a good idea to add this upstream? Seems like it's also useful outside of CIRCT.

maerhart (approved):
LGTM

include/circt/Support/ConversionPatternSet.h
fabianschuiki: Yikes, good catch

include/circt/Support/ConversionPatternSet.h
fabianschuiki: Ah I forgot to migrate one of the patterns that actually needs the `adaptor`. Added that. Good idea about upstreaming this!

maerhart: I think you might actually have to migrate over all the patterns, because the upstream implementation uses OpRewritePattern as base. I'm not sure if it's still the case with all the dialect conversion framework changes happening, but it used to be (and possibly still is) incompatible with conversion patterns.

fabianschuiki: Seems to work :thinking:

maerhart: See slide 19 of https://llvm.org/devmtg/2024-10/slides/techtalk/Springer-Pattern-Based-IR-Rewriting-in-MLIR.pdf

fabianschuiki: Hmm yeah, the `RewritePatternSet` uses `OpRewritePattern`. But it's still used to collect the `OpConversionPattern`s in a dialect conversion :thinking:. This new pattern set just factors out creating a subclass of `OpConversionPattern` into a helper function, but the rest is still the same as any other dialect conversion.

maerhart: Yes, that's part of what's meant with "API design suggests that Conversion/RewritePattern are compatible" in the slides. What I was suggesting is to change all `convert` functions in the pass to have the adaptor argument even if it is not used (or change the `add` function in `ConversionPatternSet` that currently delegates to `RewritePatternSet` to also create a `OpConversionPattern` instead)

fabianschuiki: Ah I see! Sorry, slow brain today. Let me change those over to use the adaptor :+1:.

PR #8900, opened 2025-08-29
[Moore] Add proper conversions between time and integer values

fabianschuiki:
Time values in Verilog are basically just plain 64 bit integers. This is an infinite source of frustration when trying to send time values between units with different timescales. The Moore dialect has a separate `!moore.time` type for this reason, allowing time values to be captured more precisely and converted to the equivalent `!llhd.time`.

To allow users to write `#5`, basically delaying execution for 5 times the local time scale value, add a `moore.time_to_int` and complementary `moore.int_to_time` operation. These allow time values to be cast from and to their equivalent integer number of femtoseconds.

Make type conversions in ImportVerilog more conservative by refusing any conversion between packed and simple bit vector types if the packed type has any nested time value. For trivial casts between an integer and a time value, use the new `moore.time_to_int` and `moore.int_to_time` operations together with a `moore.mul` or `moore.divu` of the integer value by the local timescale value.

Add unsigned and signed division operators to `FVInt`.

Add a bunch of constant folders to the Moore dialect and switch over to `createOrFold` in most conversion functions to produce already-folded IR where possible.

This now allows circt-verilog to process statements like `#5` or `time t = 42`.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -512,11 +512,12 @@ def ConstantOp : MooreOp<"constant", [Pure, ConstantLike]> {
   ];
 }
 
-def ConstantTimeOp : MooreOp<"constant_time", [Pure]> {
+def ConstantTimeOp : MooreOp<"constant_time", [Pure, ConstantLike]> {
   let summary = "A constant time value in femtoseconds";
   let arguments = (ins UI64Attr:$value);
   let results = (outs TimeType:$result);
   let assemblyFormat = "$value `fs` attr-dict";
+  let hasFolder = 1;
 }
 
 def StringConstantOp : MooreOp<"string_constant", [Pure]> {
@@ -552,7 +553,7 @@ def RealLiteralOp : MooreOp<"real_constant", [Pure]> {
 }
 
 //===----------------------------------------------------------------------===//
-// Casting and Resizing
+// Casting
 //===----------------------------------------------------------------------===//
 
 def ConversionOp : MooreOp<"conversion", [Pure]> {
@@ -649,6 +650,26 @@ def ToBuiltinBoolOp : MooreOp<"to_builtin_bool", [Pure]> {
     $input attr-dict `:` type($input)
   }];
 }
+
+def TimeToLogicOp : MooreOp<"time_to_logic", [Pure]> {
+  let summary = "Convert a time type to an integer number of femtoseconds";
+  let arguments = (ins TimeType:$input);
+  let results = (outs FourValuedI64:$result);
+  let assemblyFormat = "$input attr-dict";
+  let hasFolder = 1;
+}
+
+def LogicToTimeOp : MooreOp<"logic_to_time", [Pure]> {
+  let summary = "Convert an integer number of femtoseconds to a time type";
+  let arguments = (ins FourValuedI64:$input);
+  let results = (outs TimeType:$result);
+  let assemblyFormat = "$input attr-dict";
+  let hasFolder = 1;
+}
+
+//===----------------------------------------------------------------------===//
+// Resizing
+//===----------------------------------------------------------------------===//
 
 def TruncOp : MooreOp<"trunc", [
   Pure,
@@ -839,7 +860,6 @@ def SubOp : BinaryOpBase<"sub"> {
 
     See IEEE 1800-2017 § 11.4.3 "Arithmetic operators".
   }];
-
   let hasFolder = 1;
 }
 
@@ -851,6 +871,7 @@ def MulOp : BinaryOpBase<"mul", [Commutative]> {
 
     See IEEE 1800-2017 § 11.4.3 "Arithmetic operators".
   }];
+  let hasFolder = 1;
 }
 
 class DivOpBase<string mnemonic> : BinaryOpBase<mnemonic> {
@@ -863,6 +884,7 @@ class DivOpBase<string mnemonic> : BinaryOpBase<mnemonic> {
 
     See IEEE 1800-2017 § 11.4.3 "Arithmetic operators".
   }];
+  let hasFolder = 1;
 }
 
 def DivUOp : DivOpBase<"divu">;
--- a/include/circt/Support/FVInt.h
+++ b/include/circt/Support/FVInt.h
@@ -44,6 +44,10 @@ public:
       : FVInt(APInt(numBits, value, isSigned)) {}
 
   /// Construct an `FVInt` from an `APInt`. The result has no X or Z bits.
+  FVInt(APInt &&value)
+      : value(value), unknown(APInt::getZero(value.getBitWidth())) {}
+
+  /// Construct an `FVInt` from an `APInt`. The result has no X or Z bits.
   FVInt(const APInt &value)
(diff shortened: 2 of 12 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -60,6 +60,28 @@ static Value getSelectIndex(Context &context, Location loc, Value index,
...
+
+  static_assert(int(slang::TimeScaleMagnitude::One) == 1);
+  static_assert(int(slang::TimeScaleMagnitude::Ten) == 10);
+  static_assert(int(slang::TimeScaleMagnitude::Hundred) == 100);
+
+  auto exp = static_cast<unsigned>(context.timeScale.base.unit);
+  assert(exp <= 5);
+  exp = 5 - exp;
+  auto scale = static_cast<uint64_t>(context.timeScale.base.magnitude);
+  while (exp-- > 0)
```
uenoku, line 80: TIL there is no integer version of std::pow :smile: Does slang warn if there is time literal that is out-of-range (and causes overflow here)?

lib/Dialect/Moore/MooreTypes.cpp
```diff
@@ -156,6 +174,19 @@ IntType PackedType::getSimpleBitVector() const {
   return {};
 }
 
+bool PackedType::containsTimeType() const {
```
uenoku, line 177: Maybe you can walk the type attribute (https://mlir.llvm.org/doxygen/classmlir_1_1AttrTypeWalker.html) to find time type which might be more flexible for newly added types.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -650,6 +651,26 @@ def ToBuiltinBoolOp : MooreOp<"to_builtin_bool", [Pure]> {
   }];
 }
 
+def TimeToIntOp : MooreOp<"time_to_int", [Pure]> {
```
uenoku: nit: maye `time_to_logic` is better name? I thought it must produce a two state value because of `_int` suffix.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1380,33 +1466,35 @@ Value Context::materializeConversion(Type type, Value value, bool isSigned,
...
+    if (!value)
+      return {};
 
     // Create truncation or sign/zero extension ops depending on the source and
     // destination width.
     auto resizedType = moore::IntType::get(
         value.getContext(), dstInt.getWidth(), srcPacked.getDomain());
     if (dstInt.getWidth() < srcInt.getWidth()) {
-      value = moore::TruncOp::create(builder, loc, resizedType, value);
+      value = builder.createOrFold<moore::TruncOp>(loc, resizedType, value);
```
uenoku, line 1478: Unfortunate that TruncOp::createOrFold doesn't exist yet

include/circt/Support/FVInt.h
```diff
@@ -543,6 +547,32 @@ class FVInt {
     return v;
   }
 
+  /// Compute an unsigned division.
```
uenoku: Do you have references for these semantics? Should div-by-zero return X value?

uenoku (approved):
LGTM for conversion operation stuff except for one nit about naming of time_to_int. I think canonicalization and FVInt changes are orthogonal to conversion changes and I have a few question regarding behavior of div-by-zero.

include/circt/Dialect/Moore/MooreOps.td
jpienaar: Why is it four valued here? (e.g., I'd have assumed time would be two state valued)

lib/Conversion/ImportVerilog/Expressions.cpp
jpienaar: Yeah this does change the uniformity a bit ... I'll see about adding.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -650,6 +651,26 @@ def ToBuiltinBoolOp : MooreOp<"to_builtin_bool", [Pure]> {
...
+  let summary = "Convert a time type to an integer number of femtoseconds";
+  let arguments = (ins TimeType:$input);
+  let results = (outs FourValuedI64:$result);
+  let assemblyFormat = "$input attr-dict";
+  let hasFolder = 1;
+}
+
+def IntToTimeOp : MooreOp<"int_to_time", [Pure]> {
```
jpienaar: Can either of these either fail or are all inputs valid?

jpienaar (approved):

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: That would be very neat 💯

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: Excellent point, let me rename these to contain `logic`. I hate that Verilog makes time values four-state, but then also tells you that any X or Z bits are treated as 0 for the purpose of delay control or delayed assignments 🤦.

From IEEE 1800-2017 § 6.11 Integer data types:

Screenshot 2025-08-29 at 09 29 26

lib/Dialect/Moore/MooreTypes.cpp
fabianschuiki: Oh that's really nice. I wasn't aware of these walk/replace utilities. This probably needs a bit of help in the case of structs and unions, because we have this custom `StructLikeMember` array in there.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Yeah 😅. I think this here is always safe, because scale is 1, 10, or 100, and exp is between 0 and 5. So this produces `1000^5` in the worst case, which fits into 52 bits.

lib/Dialect/Moore/MooreTypes.cpp
fabianschuiki: Hmmmm I can't seem to get this to work. Somehow the walker never descends into the struct members. I'll need to read up on this a bit more. Having walkers work properly on types sounds like a very useful thing. But I might be missing some context on how to ensure this works for custom types. 😢

include/circt/Support/FVInt.h
fabianschuiki: I've added a reference to the SV spec. Good point about the div-by-zero producing X. I forgot about that. Added a test for it 👍

include/circt/Dialect/Moore/MooreOps.td
fabianschuiki: All inputs are valid. Sadly, time values are just unit-less integers in Verilog, so any conversion issues or unit mismatches are basically mandated by the standard 🤦‍♂️

lib/Dialect/Moore/MooreTypes.cpp
uenoku: Hmm, I see. I think the current implementation does look good then.

include/circt/Support/FVInt.h
uenoku: Cool, thanks!

PR #8901, opened 2025-08-29
[ESI] Extend dll dirs on esiaccel module load

mortbopet:
To ensure that things (especially other pybind packages) which build against ESI libraries have the `ESICppRuntime.dll` freely available upon load time, by adding the `esiaccel` directory to the `dll_directory` list.

### DIFF
--- a/lib/Dialect/ESI/runtime/python/esiaccel/__init__.py
+++ b/lib/Dialect/ESI/runtime/python/esiaccel/__init__.py
@@ -2,6 +2,8 @@
 #  See https://llvm.org/LICENSE.txt for license information.
 #  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+import sys
+import os
 from .accelerator import AcceleratorConnection
 
 from .esiCppAccel import (AppID, Type, BundleType, ChannelType, ArrayType,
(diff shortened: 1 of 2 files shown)


teqdruid (commented):
I'm confused... `from .esiCppAccel ...` above implicitly pulls in ESIRuntime.dll so it's already loaded into the Python process. Is this for executables run by python scripts which happen to import `esiaccel`? (I assume that the DLL directory is inherited by child processes.)

I'm also uncomfortable with yet another place to update when I have to move the location of the DLL in the wheel... Can you add a function to utils which returns the location of the so/dll? `get_dynamic_lib_dir()` or something like that?

mortbopet:
The specific case that i'm dealing with is a separate pybind package which is built against ESICppRuntime.dll. I agree with you that intuitively, importing just esiaccel would be sufficient because (as you mention) that itself has to link against ESICppRuntime.dll. However, I'm assuming that some Windows concoction of dll lookup order and whatnot is preventing that pybind package from accepting the in-memory ESICppRuntime.dll to be accepted as the loaded module. And, that's why this appears to be working.

> I'm also uncomfortable with yet another place to update when I have to move the location of the DLL in the wheel... Can you add a function to utils which returns the location of the so/dll? get_dynamic_lib_dir() or something like that?

Realistically, will this ever happen? Distributing windows .exe's/.dll's with some form of hierarchy seem to me like asking for trouble with the dynamic linker. So, assuming that, and assuming that the `esiCppAccel....pyd` pybind package will always be directly nested in the `esiaccel` package, that would imply that the dll's are placed alongside it, always.

teqdruid:
> Realistically, will this ever happen? Distributing windows .exe's/.dll's with some form of hierarchy seem to me like asking for trouble with the dynamic linker. So, assuming that, and assuming that the `esiCppAccel....pyd` pybind package will always be directly nested in the `esiaccel` package, that would imply that the dll's are placed alongside it, always.

You'd think so but historically I've changed it a few times. In fact, I just did a few weeks ago. Please create a function in utils to return this path and use it here.

After that, I think you're good to go.

PR #8902, opened 2025-08-29
[Comb] Expand Known-Bits Analysis to Additional Operators

cowardsa:
Add support for extract, and logical shift operators to KnownBits analysis - helping to improve QoR for synth results. A future PR may aim to improve performance by invoking a caching mechanism that uses the MLIR analysis runners.

### DIFF
--- a/unittests/Dialect/CMakeLists.txt
+++ b/unittests/Dialect/CMakeLists.txt
@@ -1,4 +1,5 @@
 add_subdirectory(AIG)
+add_subdirectory(Comb)
 add_subdirectory(Moore)
 add_subdirectory(FIRRTL)
 add_subdirectory(ESI)
--- a/unittests/Dialect/Comb/CMakeLists.txt
+++ b/unittests/Dialect/Comb/CMakeLists.txt
@@ -1,0 +1,11 @@
+add_circt_unittest(CIRCTCombTests
+  KnownBitsTest.cpp
+)
+
+target_link_libraries(CIRCTCombTests
+  PRIVATE
+  CIRCTHW
+  CIRCTComb
+  MLIRIR
+  MLIRParser
+)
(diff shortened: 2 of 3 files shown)


lib/Dialect/Comb/CombAnalysis.cpp
```diff
@@ -78,6 +99,22 @@ static KnownBits computeKnownBits(Value v, unsigned depth) {
...
+    return res;
+  }
+
+  // `shr(x, y)` is the known bits of `x` >> known bits of `y`.
+  if (auto shrOp = dyn_cast<ShrUOp>(op)) {
+    auto lhs = computeKnownBits(shrOp.getOperand(0), depth + 1);
+    auto rhs = computeKnownBits(shrOp.getOperand(1), depth + 1);
+    auto res = KnownBits::lshr(lhs, rhs);
+    return res;
+  }
```
uenoku, line 116: Cool, didn't know that this smart shift API for `KnownBits` exists.

uenoku (approved):
LGTM. We need unit tests for this but I don't block the PR as there hasn't been a test for other operations as well.

cowardsa:
@uenoku - sorry meant to add a question in the description. Don't think we have any existing KnownBits tests - any thoughts on where such tests should go? Would it require a test pass to be written?

uenoku:
I think we can create a unite test folder for Comb (here is for AIG and probably we can put IR and query knowBits analysis to them https://github.com/llvm/circt/blob/3f3b55927cb0b0fc2b5aebf6caa6491187f1a6a2/unittests/Dialect/AIG/LongestPathTest.cpp#L29) but test pass also makes sense to me

cowardsa:
@uenoku and @fabianschuiki - added KnownBits tests - let me know what you think? Wanted to add them now as worth checking things are behaving as expected!

PR #8903, opened 2025-08-29
[Datpath] Improve Booth Encoding based on the sign-extension trick

cowardsa:
Reduce the number of partial product bits generated using a classical sign-extension trick where repeated bits can be optimised away. Currently only implemented when the inputs to a partial product operator are zero-extended. Current folding mechanisms are unable to recognise the zero-padding in a single step (although canonicalization will incrementally strip away). The KnownBits analysis is able to recognise this at some efficiency penalty - although this could be improved using bounded depth traversal. 

A future PR will exploit this simplified partial product array via timing driven compression and improve the algorithm further to detect sign-extended inputs.

### DIFF
--- a/integration_test/circt-synth/datapath-lowering-lec.mlir
+++ b/integration_test/circt-synth/datapath-lowering-lec.mlir
@@ -18,6 +18,31 @@ hw.module @partial_product_4(in %a : i4, in %b : i4, out sum : i4) {
   hw.output %1 : i4
 }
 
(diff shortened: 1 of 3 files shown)


uenoku (approved):
Looks really awesome! Could you add normal FileCheck tests and LEC for partial products with zero/sign extension?

cowardsa:
Now with tests added :)

uenoku (approved):
Thanks, looks awesome!

PR #8909, opened 2025-08-30
[Synth] Refactor and improve cut rewriter infrastructure

uenoku:
This commit refactors the cut rewriter and implements correct priority sort metric for depth. Previously `depth` of a cut represents a logic depth *within* a cut but this is not correct. What the original paper (ICCAD 2007) proposed was to use arrival time of cut outputs instead.

For that this commit changes to perform cut matching for each cut, and use worst arrival time as a priority of cuts.

LUT6 depth for 16-bit multiplier improved from 17 to 7, which is on per with Yosys/ABC.

### DIFF
--- a/lib/Dialect/Synth/Transforms/CutRewriter.cpp
+++ b/lib/Dialect/Synth/Transforms/CutRewriter.cpp
@@ -22,6 +22,7 @@
 
 #include "circt/Dialect/AIG/AIGOps.h"
 #include "circt/Dialect/Comb/CombOps.h"
+#include "circt/Dialect/HW/HWOps.h"
 #include "circt/Support/LLVM.h"
 #include "circt/Support/NPNClass.h"
 #include "circt/Support/UnusedOpPruner.h"
@@ -44,9 +45,11 @@
 #include "llvm/ADT/iterator.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
+#include "llvm/Support/LogicalResult.h"
 #include <functional>
 #include <memory>
 #include <optional>
+#include <string>
 
 #define DEBUG_TYPE "synth-cut-rewriter"
 
@@ -224,8 +227,6 @@ bool Cut::isTrivialCut() const {
   return operations.empty() && inputs.size() == 1;
 }
 
-unsigned Cut::getDepth() const { return depth; }
-
 mlir::Operation *Cut::getRoot() const {
   return operations.empty()
              ? nullptr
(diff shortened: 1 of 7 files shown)


lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -546,49 +527,101 @@ static void removeDuplicateAndNonMinimalCuts(SmallVectorImpl<Cut> &cuts) {
...
-  for (auto &cut : cuts) {
-    // Match the cut against the pattern set
-    auto matchResult = matchCut(cut);
-    if (!matchResult)
-      continue;
+  // Step 3: Sort cuts by priority to select the best ones
+  // Priority is determined by timing optimization strategy:
+  // - Trivial cuts (direct connections) have highest priority
+  // - Among matched cuts, prefer better timing then lower complexity
+  // - Matched cuts are preferred over unmatched cuts
```
cowardsa: Nit: perhaps a link to the paper could be provided here?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -546,49 +527,101 @@ static void removeDuplicateAndNonMinimalCuts(SmallVectorImpl<Cut> &cuts) {
...
+    const auto &currentMatch = cut.getMatchedPattern();
+    if (!currentMatch)
+      continue; // Skip cuts without matched patterns
+
+    // Compare this cut against the current best cut
+    if (!bestCut ||
+        compareDelayAndArea(options.strategy, currentMatch->getArea(),
+                            currentMatch->getArrivalTimes(),
+                            bestCut->getMatchedPattern()->getArea(),
+                            bestCut->getMatchedPattern()->getArrivalTimes())) {
```
cowardsa: Is the sorting criteria used above identical to the compareDelayAndArea?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
```diff
@@ -872,26 +978,28 @@ std::optional<MatchedPattern> CutRewriter::patternMatchCut(Cut &cut) {
...
+    if (!bestCut)
       return {};
 
+    const auto &matchedPattern = *bestCut->getMatchedPattern();
+
     // This must be a block argument must have been a cut input.
     auto resultNumber = cast<mlir::OpResult>(input);
     inputArrivalTimes.push_back(
-        matchedPattern->getArrivalTime(resultNumber.getResultNumber()));
+        matchedPattern.getArrivalTime(resultNumber.getResultNumber()));
```
cowardsa: Nit: mixture of result and input is a bit confusing here?

cowardsa (approved):
LGTM - only comment on slightly improving documentation for heuristics chosen - perhaps this could also be reflected in dialect or synthesis documentation to make it clear what features are implemented and where the ideas come from?

lib/Dialect/Synth/Transforms/CutRewriter.cpp
uenoku: Great point! I modified the sorting function to use `compareDelayAndArea`, and change the second loop just to return the first matched one.

uenoku:
> perhaps this could also be reflected in dialect or synthesis documentation to make it clear what features are implemented and where the ideas come from

Great point, I'll write down a rational document for Synth dialect and cut rewriter in a follow up.

PR #8911, opened 2025-09-02
[ESI] Don't assume __dict__ attribute for all objects

mortbopet:
... when checking if a given object is valid for `StructType` serialization. E.g., `bytes` objects do not have a `__dict__` attribute.

### DIFF
--- a/lib/Dialect/ESI/runtime/python/esiaccel/types.py
+++ b/lib/Dialect/ESI/runtime/python/esiaccel/types.py
@@ -245,6 +245,8 @@ class StructType(ESIType):
   def is_valid(self, obj) -> Tuple[bool, Optional[str]]:
     fields_count = 0
     if not isinstance(obj, dict):
+      if not hasattr(obj, "__dict__"):
+        return (False, "must be a dict or have __dict__ attribute")
       obj = obj.__dict__
 
     for (fname, ftype) in self.fields:


lib/Dialect/ESI/runtime/python/esiaccel/types.py
```diff
@@ -260,6 +262,8 @@ def is_valid(self, obj) -> Tuple[bool, Optional[str]]:
 
   def serialize(self, obj) -> bytearray:
     ret = bytearray()
+    if not isinstance(obj, dict):
```
teqdruid, line 265: Is `is_valid` called before this? If not, it should be. Problem is that it should either be called at the top of the serialize call or be called here and not descend into the inner objects. Or is it better to require the user code to check and put it somewhere in the documentation?

teqdruid (approved):
Not strictly related to this PR: we've (I've) played a bit fast and loose with testing, especially the Python API. Now that we've got the ability to create Python types from Python (right?), we could create some pytests to test the type / value system without invoking PyCDE to create a system (manifest). This is even more important for Python APIs since they don't get compiled. I'm less worried about the C++ side as a result of this (and its functionality is significantly simpler) but we definitely need some mechanism to test that better also.

mortbopet:
> Not strictly related to this PR: we've (I've) played a bit fast and loose with testing, especially the Python API. Now that we've got the ability to create Python types from Python (right?), we could create some pytests to test the type / value system without invoking PyCDE to create a system (manifest). This is even more important for Python APIs since they don't get compiled. I'm less worried about the C++ side as a result of this (and its functionality is significantly simpler) but we definitely need some mechanism to test that better also.

Completely agree. But, before that, we should probably have some dummy backend, such that we can test the ESI runtime proper. I.e., what it'd **really** like to be able to do is to just open the ESIRuntime subdirectory, and not care at all about building the rest of CIRCT to test it.

lib/Dialect/ESI/runtime/python/esiaccel/types.py
mortbopet: The main place where `is_valid` is called is when [writing to a port](https://github.com/llvm/circt/blob/78344510fddd232b4f5e22bb1650ea6a0ea8b0f4/lib/Dialect/ESI/runtime/python/esiaccel/types.py#L356). Since it (as you mention) is somewhat critical that a given type `is_valid` when trying to serialize and object, that `is_valid` call should probably be moved to the serialization function.

teqdruid:
> > Not strictly related to this PR: we've (I've) played a bit fast and loose with testing, especially the Python API. Now that we've got the ability to create Python types from Python (right?), we could create some pytests to test the type / value system without invoking PyCDE to create a system (manifest). This is even more important for Python APIs since they don't get compiled. I'm less worried about the C++ side as a result of this (and its functionality is significantly simpler) but we definitely need some mechanism to test that better also.
> 
> Completely agree. But, before that, we should probably have some dummy backend, such that we can test the ESI runtime proper. I.e., what it'd **really** like to be able to do is to just open the ESIRuntime subdirectory, and not care at all about building the rest of CIRCT to test it.

We do have a dummy backend (`trace`) what we don't have is an easy way build an ESI system. The ESI runtime tests could pull down a PyCDE from a package on pypi...

mortbopet:
> We do have a dummy backend (trace) what we don't have is an easy way build an ESI system. The ESI runtime tests could pull down a PyCDE from a package on pypi...

In my mind, a proper "dummy" backend also has APIs to dynamically construct the underlying accelerator (99% of the time, it's probably just loopback, but i can imagine an API of that dummy backend which takes an input type, output type, and a lambda, which defines the transformation in between). Then we take PyCDE/complexity out of the equation, and keep everything internral to the ESI runtime project.

Issue #8786, opened 2025-07-26
[Synthesis][AIG] Add Synthesis dialect and change AIG dialect to a sub-dialect of synthesis


uenoku:
Currently AIG dialect is used as core representation for circt-synth flow but for synthesis there are few other operations I'd like to introduce but not limited to AIG. From that perspective I'm thinking it's better to introduce a dialect for synthesis and move AIG to sub-dialect of Synthesis. This is my current sketch for what looks like. 

### Meta operations for synthesis 
* `synth.choice` operation -- defer lowering decisions to the late stage/actual tech mapping
```mlir
%0 = synth.choice %slow_but_small, %balance, %fast_but_large : i42
```
 inspired by https://people.eecs.berkeley.edu/~alanmi/publications/2005/tech05_map.pdf but not limited to AIG node

* `synth.bundle` operation -- This is basically a compiler hint to map multiple output cells. Without this kind of operations it's necessary to recognize pairs of root operations in memory and DAG rewriting becomes much annoying to manage. 
```mlir
%carry = comb.and %a, %b: i1
%sum = comb.xor %a, %b: i1
%carry', %sum' = synth.bundle %carry, %sum : i1, i1
```
* `synth.store` operation -- rollback to original operation
```mlir
%original = synth.store (%a, %b) {
    %1 = comb.mul %a, %b : i32
   synth.store.return %0: i32
}
%tech_mapped = hw.instance @some_gate (..) {arrivatl_time=1500}
%0 = synth.choice %tech_mapped, %original : i32

==> Restore pass

%original = comb.mul %a, %b : i32
%tech_mapped = hw.instance @some_gate (...) {arrival_time=1500}
%0 = synth.choice %tech_mapped, %original : i32

==> Rerun synthesis pipeline with post mapped timing analysis
%tech_mapped = hw.instance @some_gate (...) {arrival_time=1500}
%tech_mapped_2 = hw.instance @some_gate (...) {arrival_time=1400}
%0 = synth.choice %tech_mapped, %tech_mapped_2 : i32
```

### Logic Representation 
Currently both AIG and MIG are widely used and being able to experiment both representations would be beneficial.

* `synth.aig.and_inv` operation (renamed from `aig.and_inv`)
* `synth.mig.majority_inv` operation.  MIG has emerged as a new efficient representation for logic synthesis https://si2.epfl.ch/demichel/publications/archive/2016/07293649.pdf. 

[...]

PR #8912, opened 2025-09-02
[Synth] Add MajorityInverterOp (synth.mig.maj_inv) to Synth dialect

uenoku:
Towards https://github.com/llvm/circt/issues/8786, this commit introduces a new MajorityInverterOp operation to represent majority-inverter gates in Majority-Inverter Graphs (MIG) for synthesis (https://ieeexplore.ieee.org/document/7293649). Majority-Inverter Graphs have become a prominent representation for logic synthesis in recent years, offering advantages in certain optimization scenarios alongside And-Inverter Graphs (AIG).

The implementation provides a synth.mig.maj_inv operation that computes the majority function of its inputs with optional inversion per operand. The operation requires an odd number of operands to avoid ties in majority computation and uses a custom assembly format with 'not' keyword prefix for inverted inputs. 

The majority function returns 1 when more than half of the inputs are 1. For three inputs, it's equivalent to: (a & b) | (a & c) | (b & c).

This addition aligns with the goal of making the circt-synth pipeline handle both AIG and MIG representations seamlessly, providing flexibility in choosing the most appropriate intermediate representation (or mix them). Also MIG can represent and/or/carry in a single depth (and-inverter takes 2 depth for or/carry) so would be slightly accurate metric as logic depth. 

Example usage:
```
  %r1 = synth.mig.maj_inv %a, %b, %c : i1
  %r2 = synth.mig.maj_inv not %a, %b, not %c : i1
```

### DIFF
--- a/include/circt/Dialect/AIG/AIGOps.td
+++ b/include/circt/Dialect/AIG/AIGOps.td
@@ -56,9 +56,9 @@ def AndInverterOp : AIGOp<"and_inv", [SameOperandsAndResultType, Pure]> {
   let arguments = (ins Variadic<AnyType>:$inputs, DenseBoolArrayAttr:$inverted);
   let results = (outs AnyType:$result);
 
-  // NOTE: Custom assembly format is needed to pretty print the `inverted`
-  // attribute.
-  let hasCustomAssemblyFormat = 1;
+  let assemblyFormat = [{
+    custom<VariadicInvertibleOperands>($inputs, type($result), $inverted, attr-dict)
+  }];
 
   let builders = [
     OpBuilder<(ins "Value":$input, CArg<"bool", "false">:$invert), [{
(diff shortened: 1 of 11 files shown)


include/circt/Dialect/AIG/AIGOps.td
```diff
@@ -56,9 +56,9 @@ def AndInverterOp : AIGOp<"and_inv", [SameOperandsAndResultType, Pure]> {
   let arguments = (ins Variadic<AnyType>:$inputs, DenseBoolArrayAttr:$inverted);
   let results = (outs AnyType:$result);
 
-  // NOTE: Custom assembly format is needed to pretty print the `inverted`
-  // attribute.
-  let hasCustomAssemblyFormat = 1;
+  let assemblyFormat = [{
+    custom<VariadicInvertibleOperands>($inputs, type($result), $inverted, attr-dict)
+  }];
```
fabianschuiki, line 61: Very cool! Love the use of `custom` here 🙌

fabianschuiki (approved):
Very cool!

PR #8916, opened 2025-09-02
[ESI] Unpack dictionary arguments for CallbackPort callbacks

mortbopet:
... Allowing for more pythonic syntax when connecting to callback ports.


teqdruid (commented):
Please create a test for this in the PyCDE integration tests.

PR #8917, opened 2025-09-02
[LLHD] Simplify single-block llhd.process terminated by llhd.halt

mvpant:
This simplifies an `llhd.process` that has only a single basic block with an `llhd.halt` terminator.
1) Lifts constant-like yield operands of `llhd.halt` operation and replaces uses of `llhd.process` results with them directly.
2) Deduplicates the yield operands of `llhd.halt`.

### DIFF
--- a/lib/Dialect/LLHD/IR/LLHDOps.cpp
+++ b/lib/Dialect/LLHD/IR/LLHDOps.cpp
@@ -491,14 +491,14 @@ LogicalResult ProcessOp::canonicalize(ProcessOp op, PatternRewriter &rewriter) {
     return failure();
 
   auto &block = op.getBody().front();
-  if (op.getNumResults() == 0) {
-    rewriter.eraseOp(op);
-    return success();
-  }
-
   auto haltOp = dyn_cast<HaltOp>(block.getTerminator());
   if (!haltOp)
     return failure();
+
+  if (op.getNumResults() == 0 && block.getOperations().size() == 1) {
+    rewriter.eraseOp(op);
+    return success();
+  }
 
   // Only constants and halt terminator are expected in a single block.
   if (!llvm::all_of(block.without_terminator(), [](auto &bodyOp) {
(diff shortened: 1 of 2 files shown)


mvpant:
> Regarding the initial process that just ends in an `llhd.halt`:
> 
> ```mlir
>     %2:2 = llhd.process -> i1, i1 {
>       llhd.halt %false, %reset : i1, i1
>     }
>     llhd.drv %reg_0, %2#0 after %0 if %2#1 : !hw.inout<i1>
> ```
> 
> This should actually be pretty straightforward to handle. I think we're just missing a folder or canonicalizer that replaces `llhd.process` results with the corresponding `llhd.halt` constant operand, if there's only a single `llhd.halt` in the last block.
> 
> It's important to only do this for constant operands, since the process runs only once at very beginning. This means that any non-constant value would basically be "snapshot" at that specific point in time and produced as a result from the process. This result won't change throughout the simulation, even if the original non-constant value does. Doing this only for constants works around this problem.

This is a follow up to the [discussion](https://github.com/llvm/circt/pull/8870#issuecomment-3237758156)

lib/Dialect/LLHD/IR/LLHDOps.cpp
```diff
@@ -554,6 +615,55 @@ LogicalResult WaitOp::verify() {
...
+  }
+
+  rewriter.eraseOp(processOp);
+  return success();
+}
+
+LogicalResult HaltOp::canonicalize(HaltOp op, PatternRewriter &rewriter) {
+  return deduplicateYieldOperands(rewriter, op);
+}
+
```
fabianschuiki: This looks like it's doing the same thing as the canonicalizer for `ProcessOp` above 🤔. Since deduplicating yield operands requires us to change the process' results, which requires a new process op, it probably makes sense to do all of this as part of the `ProcessOp` canonicalizer above. WDYT?

lib/Dialect/LLHD/IR/LLHDOps.cpp
```diff
@@ -487,11 +487,72 @@ DrvOp::ensureOnlySafeAccesses(const MemorySlot &slot,
...
-    if (block.getOperations().size() == 1 && isa<HaltOp>(block.getTerminator()))
-      rewriter.eraseOp(op);
+  if (!op.getBody().hasOneBlock())
+    return failure();
+
+  auto &block = op.getBody().front();
+  if (op.getNumResults() == 0) {
+    rewriter.eraseOp(op);
+    return success();
+  }
```
fabianschuiki: I think still needs `block.getOperations().size() == 1`, since the body may contain side-effecting ops that we cannot just delete.

fabianschuiki (approved):
Very cool! Thanks a lot for adding this 👍 🙌

fabianschuiki:
Want me to merge this, @mvpant?

mvpant:
> Want me to merge this, @mvpant?

Yes, thanks!

PR #8920, opened 2025-09-03
[Datapath] Implement timing-driven compression algorithm

cowardsa:
Building on recent updates to incorporate longest path analysis for higher-level dialects (e.g. comb), this PR incorporates the arrival time information into the design of the compressor tree which improves the logic depth of a datpath.compress lowering to AIG. The CompressorTree class was moved into datapath.ops to avoid additional clutter in the comb ops files. 

Future PRs will further build upon these algorithms to produce even more efficient compressors.


lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,323 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+      auto col = columns[i];
+
+      // Sort the column by arrival time - fastest at the end
+      std::stable_sort(
+          col.begin(), col.end(),
+          [](const auto &a, const auto &b) { return a.delay > b.delay; });
+      // Only compress to reach the target stage height - Dadda's Algorithm
+      while (col.size() + newColumns[i].size() > targetStageHeight) {
+        if (col.size() < 2) {
+          llvm::dbgs() << "CompressorTree: Not enough bits in column " << i
```
uenoku:
```suggestion
          llvm::errs() << "CompressorTree: Not enough bits in column " << i
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,323 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+      auto it = std::find_if(columns[j].begin(), columns[j].end(),
+                             [val = originalAddends[i][j]](const auto &bit) {
+                               return bit.val == val;
+                             });
+      if (it != columns[j].end())
+        it->delay = inputDelays[i][j];
+    }
+  }
+  usingTiming = true;
+}
```
uenoku: Can we use callback for timing info? I think it makes here a bit simpler(also we don't need to copy originalAddends in that case).

```suggestion
LogicalResult CompressorTree::withInputDelays(llvm::function_ref<FailureOr<int64_t>(Value)> getDelay)) {
  for (auto& column: columns) {
      for(auto& [value, result] : column) {
          auto delay = getDelay(value);
          if(failed(delay)) return failure();
          result = delay;
      }
  }
  usingTiming = true;
}
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,323 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+// Compressor Tree Logic.
+//===----------------------------------------------------------------------===//
+
+// Construct a full adder for three 1-bit inputs.
+std::pair<CompressorBit, CompressorBit>
+CompressorTree::fullAdderWithDelay(OpBuilder &builder, CompressorBit a,
+                                   CompressorBit b, CompressorBit c) {
+
+  auto aXorB = builder.createOrFold<comb::XorOp>(loc, a.val, b.val, true);
+  Value sum_val = builder.createOrFold<comb::XorOp>(loc, aXorB, c.val, true);
```
uenoku: nit: please follow camelCase. 
```suggestion
  Value sumVal = builder.createOrFold<comb::XorOp>(loc, aXorB, c.val, true);
```

include/circt/Dialect/Datapath/DatapathOps.h
```diff
@@ -17,4 +17,86 @@
...
+private:
+  // Original addends representation as bitvectors (kept for reference)
+  SmallVector<SmallVector<Value>> originalAddends;
+
+  // Column-wise bit storage - columns[i] contains all bits at bit position i
+  SmallVector<SmallVector<CompressorBit>> columns;
+
+  // Whether to use a timing driven compression algorithm
+  // If true, use a timing driven compression algorithm.
+  // If false, use a simple compression algorithm (e.g., Wallace).
```
uenoku: Is it possible to use timing driven algorithm with uniform arrival time even when timing info is not provided?

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,323 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+        continue;
+      // Add non-zero bit to the column
+      columns[i].push_back(bit);
+    }
+  }
+}
+
+// Update the input delays based on longest path analysis
+void CompressorTree::withInputDelays(
+    const SmallVector<SmallVector<int64_t>> inputDelays) {
```
uenoku: nit:
```suggestion
    const SmallVector<SmallVector<int64_t>> &inputDelays) {
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,323 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+  usingTiming = true;
+}
+
+size_t CompressorTree::getMaxHeight() const {
+  size_t maxSize = 0;
+  for (const auto &column : columns) {
+    maxSize = std::max(maxSize, column.size());
+  }
+  return maxSize;
+}
```
uenoku: nit: please remove single brace
```suggestion
  for (const auto &column : columns)
    maxSize = std::max(maxSize, column.size());
  return maxSize;
}
```

include/circt/Dialect/Datapath/DatapathOps.h
```diff
@@ -17,4 +17,86 @@
...
+  bool usingTiming;
+
+  // Bitwidth of compressor tree
+  size_t width;
```
uenoku: Can you pass width explicitly and make it const?
```suggestion
  const size_t width;
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,323 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+      numFullAdders(0), loc(loc) {
+  assert(addends.size() > 2);
+  // Number of bits in a row == bitwidth of input addends
+  // Compressors will be formed of uniform bitwidth addends
+  width = addends[0].size();
+  SmallVector<SmallVector<CompressorBit>> initColumns(width);
+  columns = initColumns;
+  // Convert addends rows to columns
+  // Known bits analysis constructs a minimal array - skipping zeros
+  for (auto row : addends) {
```
uenoku: Can we initialize width and columns with constructors and add an assertion for width?
```suggestion
// Map input rows to column representation
CompressorTree::CompressorTree(size_t width, const SmallVector<SmallVector<Value>>& addends,
                               Location loc)
    : originalAddends(addends), usingTiming(false), numStages(0),
      numFullAdders(0), loc(loc), width(width), columns(width) {
  assert(addends.size() > 2);
  // Convert addends rows to columns
  // Known bits analysis constructs a minimal array - skipping zeros
  for (const auto& row : addends) {
    assert(row.size() == width);
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,323 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+  auto carry_delay = sum_delay + 1;
+
+  CompressorBit sum = {sum_val, sum_delay};
+  CompressorBit carry = {carry_val, carry_delay};
+  std::pair<CompressorBit, CompressorBit> fa{sum, carry};
+  ++numFullAdders;
+  return fa;
+}
+
+// Construct a full adder for three 1-bit inputs.
```
uenoku:
```suggestion
// Construct a half adder for two 1-bit inputs.
```

lib/Dialect/Datapath/DatapathOps.cpp
cowardsa: Nice - I constructed this daft approach in an attempt to keep the separation - but this is so much easier thanks for the suggestion!

cowardsa:
Now with changes recommended including removal of wallace reduction code and using a function call back for delay calculations

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+    maxSize = std::max(maxSize, column.size());
+
+  return maxSize;
+}
+
+// Use Dadda's ALAP alogrithm to determine the target height of the next stage
+// https://en.wikipedia.org/wiki/Dadda_multiplier
+size_t CompressorTree::getNextStageTargetHeight() const {
+  auto maxHeight = getMaxHeight();
+  size_t m_prev = 2;
```
uenoku:
```suggestion
  size_t mPrev = 2;
```

lib/Conversion/DatapathToComb/DatapathToComb.cpp
```diff
@@ -74,41 +74,27 @@ struct DatapathCompressOpConversion : mlir::OpRewritePattern<CompressOp> {
...
-    // Wallace tree reduction
-    // TODO: Implement a more efficient compression algorithm to compete with
-    // yosys's `alumacc` lowering - a coarse grained timing model would help to
-    // sort the inputs according to arrival time.
+    // Compressor tree reduction
+    auto width = inputs[0].getType().getIntOrFloatBitWidth();
     auto targetAddends = op.getNumResults();
+    datapath::CompressorTree comp(width, addends, loc);
+
+    SmallVector<SmallVector<int64_t>> addendsDelays(addends.size());
```
uenoku:
```suggestion
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+      // Only compress to reach the target stage height - Dadda's Algorithm
+      while (col.size() + newColumns[i].size() > targetStageHeight) {
+        if (col.size() < 2) {
+          llvm::errs() << "CompressorTree: Not enough bits in column " << i
+                       << " to compress further.\n New Columns size: "
+                       << newColumns[i].size()
+                       << ", Current Column size: " << col.size() << "\n";
+        }
+        assert(col.size() >= 2 &&
+               "Expected at least two bits in compressor column");
```
uenoku:
```suggestion
                       << ", Current Column size: " << col.size() << "\n";
         llvm::report_fatal_error("Expected at least two bits in compressor column");

        }
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+            bit2 = *it;
+            col.erase(it);
+          } else {
+            // If no bit satisfies the delay constraint pick the fastest one
+            bit2 = col.pop_back_val();
+          }
+          auto [sum, carry] = fullAdderWithDelay(builder, bit0, bit1, bit2);
+
+          newColumns[i].push_back(sum);
+          if (i + 1 < newColumns.size())
```
uenoku, line 275: Is newColumns.size() always equal to width?

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+            newColumns[i + 1].push_back(carry);
+        }
+      }
+
+      // Pass through remaining bits
+      for (auto bit : col)
+        newColumns[i].push_back(bit);
```
uenoku:
```suggestion
        newColumns[i].append(col);
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+    }
+
+    // Compute another stage of reduction
+    columns = newColumns;
```
uenoku:
```suggestion
    columns = std::move(newColumns);
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+
+// Perform recursive compression using timing information until reduced to the
+// target height - this currently uses Dadda's algorithm and timing driven
+// signal selection
+// TODO: Dadda's algorithm is redundant here since it assumes uniform arrival so
+// need to implement a more timing driven approach
+SmallVector<Value> CompressorTree::compressUsingTiming(OpBuilder &builder,
+                                                       size_t targetHeight) {
+  while (getMaxHeight() > targetHeight) {
+    dump();
```
uenoku:
```suggestion
    LLVM_DEBUG(dump());
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+
+      // Pass through remaining bits
+      for (auto bit : col)
+        newColumns[i].push_back(bit);
+    }
+
+    // Compute another stage of reduction
+    columns = newColumns;
+  }
+  dump();
```
uenoku:
```suggestion
  LLVM_DEBUG(dump(););
```

lib/Dialect/Datapath/DatapathOps.cpp
```diff
@@ -59,6 +66,269 @@ static void printCompressFormat(OpAsmPrinter &printer, Operation *op,
...
+  return columnsToAddends(builder, targetHeight);
+}
+
+void CompressorTree::dump() const {
+  LLVM_DEBUG({
```
uenoku: nit: Could you make `dump` dumps information regardless of DEBUG, and wrap dump callers with `LLVM_DEBUG`?
```suggestion
```

lib/Dialect/Datapath/DatapathOps.cpp
cowardsa: I believe so - we initialize it:
SmallVector<SmallVector<CompressorBit>> newColumns(width);

And then we never push_back to newColumns itself only newColumns[i]

PR #8922, opened 2025-09-04
[FIRRTL] Use the defname for deciding bindfile names of extmodules

rwy7:
Lowerlayers is using the name of an extmodule to determine the name of the extmodule's bindfiles. Instead, we should be using the defname.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
@@ -257,6 +257,10 @@ class LowerLayersPass
   void preprocessModule(CircuitNamespace &ns, InstanceGraphNode *node,
                         FModuleOp module);
 
+  /// Get the verilog name of an extmodule, which could be recorded in its
+  /// defname property.
+  StringRef getExtModuleName(FExtModuleOp extModule);
+
   /// Record the supposed bindfiles for any known layers of the ext module.
   void preprocessExtModule(CircuitNamespace &ns, InstanceGraphNode *node,
                            FExtModuleOp extModule);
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -257,6 +257,10 @@ class LowerLayersPass
   void preprocessModule(CircuitNamespace &ns, InstanceGraphNode *node,
                         FModuleOp module);
 
+  /// Get the verilog name of an extmodule, which could be recorded in its
+  /// defname property.
+  StringRef getExtModuleName(FExtModuleOp extModule);
```
seldridge, line 262: This feels like a missing method on `FExtModuleOp` itself. Any idea if this shows up elsewhere in the code base which would definitely warrant raising this to the class?

test/firtool/lower-layers.fir
```diff
@@ -316,3 +316,22 @@ circuit Foo:
...
+
+  extmodule Foo knownlayer A:
+    defname = Bar
+
+  public module Top:
+    inst foo of Foo
+
+; CHECK:      FILE "layers-Top-A.sv"
+; CHECK:      `include "layers-Bar-A.sv"
+; CHECK-NOT:  `include "layers-Foo-A.sv"
```
seldridge: Nit: This could technically pass if the `` `include `` for Foo showed up before the `` `include `` for Bar. Consider:


```suggestion
; CHECK-NOT:  `include
; CHECK:      `include "layers-Bar-A.sv"
; CHECK-NOT:  `include
```

rwy7: I dropped the check-not, honestly it wasn't really functional, it was only "explanatory", if that makes sense.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
rwy7: I think it is something we should consider, this pattern shows up in a few places, but I don't want to address this in this PR. I will open an issue.

rwy7: https://github.com/llvm/circt/issues/8923

PR #8927, opened 2025-09-05
Reapply "[ci] Add Slang to UBTI workflows"

seldridge:
This reverts commit ac490f1128580ad754411aef557d7bbba74fa819.

Reapply changes that include Slang in the UBTI builds.  This originally
broke the Windows build because it does not, by default, support UTF-8.
This is now fixed in: 8ca2e97e8.


fabianschuiki:
Looks like `fmt` is complaining about a missing CLI flag:
```
static_assert failed: 'Unicode support requires compiling with /utf-8'
```
Interesting that this is coming from the fmt user code, not msvc itself. Maybe turning on UTF-8 support in msvc builds changes some of the types or stdlib stuff that fmt relies on?

fzi-hielscher:
#8926 only added the UTF flag to the standalone build, not the unified build the CI is using. I tried adding it to both in a test run and it does compile:
https://github.com/llvm/circt/actions/runs/18022778620/job/51283784148
It should hopefully pass after fixing the path separators in the lit test.

It took me way too long to figure this out. The `/EHs-c- /GR-` flags _are_ present in this PR's logs, but apparently they are not coming form our CMake file. And attempts to recreate the problem locally had all sorts of outcomes, just not the one seen here. It's a bit of a mess...

PR #8928, opened 2025-09-05
[ci] Release circt-{opt,synth,test} in firrtl-bin

seldridge:
Add additional, useful tools as part of the `firrtl-bin` release package.
This will avoid users having to get one of the "full" installations (and
also deal with the fact that these installations are not truly static).

This has a cost of bloating the `firrtl-bin` tarball.  However, that seems
acceptable given that users may not have easy alternatives for how to get
these tools.


uenoku (approved):
LGTM

PR #8929, opened 2025-09-05
[FIRRTL] Add LowerDomains pass

seldridge:
Add a pass that will lower FIRRTL domains to properties.  This has the
effect of removing all domain information from the circuit.

### DIFF
--- a/include/circt/Dialect/FIRRTL/Passes.td
+++ b/include/circt/Dialect/FIRRTL/Passes.td
@@ -905,4 +905,12 @@ def CheckLayers : Pass<"firrtl-check-layers", "firrtl::CircuitOp"> {
   }];
 }
 
+def LowerDomains : Pass<"firrtl-lower-domains", "firrtl::CircuitOp"> {
+  let summary = "lower domain information to properties";
+  let description = [{
+    Lower all domain information into FIRRTL properties.  This has the effect of
+    erasing all domain information.
+  }];
+}
+
 #endif // CIRCT_DIALECT_FIRRTL_PASSES_TD
--- a/lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
@@ -1,0 +1,254 @@
+//===- LowerDomains.cpp - Erase all FIRRTL Domain Information -------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This lower all FIRRTL domain information into properties.  This is part of
+// the compilation of FIRRTL domains where they are inferred and checked (See:
+// the InferDomains pass) and then lowered (this pass).  This pass has the
+// additional effect of removing all domain information from the circuit.
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/FIRRTL/FIRRTLInstanceGraph.h"
+#include "circt/Dialect/FIRRTL/FIRRTLOpInterfaces.h"
+#include "circt/Dialect/FIRRTL/FIRRTLOps.h"
+#include "circt/Support/Debug.h"
+#include "circt/Support/InstanceGraphInterface.h"
+#include "llvm/ADT/PostOrderIterator.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "firrtl-lower-domains"
+
+namespace circt {
+namespace firrtl {
+#define GEN_PASS_DEF_LOWERDOMAINS
+#include "circt/Dialect/FIRRTL/Passes.h.inc"
+} // namespace firrtl
+} // namespace circt
+
+using namespace circt;
+using namespace firrtl;
+
+namespace {
+class Constants {
+
+public:
+  Constants(MLIRContext *context) : context(context) {}
+
+  ArrayRef<PortInfo> getDomainPorts();
+
+private:
+  MLIRContext *context;
+  SmallVector<PortInfo, 1> domainPorts;
+};
+
+ArrayRef<PortInfo> Constants::getDomainPorts() {
+  if (domainPorts.empty())
+    domainPorts.append(
+        {{/*name=*/
+          StringAttr::get(context, "associations"),
+          /*type=*/
+          ListType::get(context, cast<PropertyType>(PathType::get(context))),
+          /*dir=*/Direction::In}});
+  return domainPorts;
+}
+
+class LowerModule {
+
+public:
+  LowerModule(FModuleLike &op, const DenseMap<Attribute, ClassOp> &classes)
+      : op(op), eraseVector(BitVector(op.getNumPorts())), classes(classes) {}
+
+  // Lower the associated module.  Remove domain ports and remove all domain
+  // information.
+  [[nodiscard]] LogicalResult lowerModule();
+
+  // Lower all instances of the associated module.  This relies on state built
+  // up during `lowerModule`.
+  void lowerInstances(InstanceGraph &);
+
+private:
+  FModuleLike &op;
+  BitVector eraseVector;
+  const DenseMap<Attribute, ClassOp> &classes;
+};
+
+struct Associations {
+  ObjectOp object;
+  SmallVector<DistinctAttr> associations = SmallVector<DistinctAttr>();
+};
+
+LogicalResult LowerModule::lowerModule() {
+  auto *body =
+      llvm::TypeSwitch<FModuleLike, Block *>(op)
+          .Case<FModuleOp>([](FModuleOp op) { return op.getBodyBlock(); })
+          .Default([](auto op) { return nullptr; });
+  if (!body) {
+    op.emitOpError() << "is unsupported in LowerDomains";
+    return failure();
+  }
+  auto builder = OpBuilder::atBlockBegin(body);
+  DenseMap<int, Associations> objects;
+  SmallVector<Attribute> newPortAnnotations;
+  for (auto [index, port] : llvm::enumerate(op.getPorts())) {
+    // Mark domain type ports for removal.
+    auto domain = dyn_cast<FlatSymbolRefAttr>(port.domains);
+    if (domain) {
+      objects[index] = {ObjectOp::create(
+          builder, port.loc, classes.lookup(domain.getAttr()), port.name)};
+      eraseVector.set(index);
+      newPortAnnotations.push_back(port.annotations.getArrayAttr());
+      continue;
+    }
+
+    // If this port has domain associations, then add port annotation trackers.
+    // These will be hooked up to the Object's associations later.
+    ArrayAttr domainAttr = cast<ArrayAttr>(port.domains);
+    if (!domainAttr || domainAttr.empty())
+      continue;
+
+    SmallVector<Annotation> newAnnotations;
+    for (auto attr : domainAttr) {
+      auto index = cast<IntegerAttr>(attr).getUInt();
+      auto id = DistinctAttr::create(UnitAttr::get(op.getContext()));
+      newAnnotations.push_back(Annotation(DictionaryAttr::getWithSorted(
+          op.getContext(),
+          {{"class", builder.getStringAttr("circt.tracker")}, {"id", id}})));
+      objects[index].associations.push_back(id);
+    }
+    if (!newAnnotations.empty())
+      port.annotations.addAnnotations(newAnnotations);
+    newPortAnnotations.push_back(port.annotations.getArrayAttr());
+  }
+  op.setPortAnnotationsAttr(builder.getArrayAttr(newPortAnnotations));
+
+  // Hook up all assocaitions to the created objects.
+  for (size_t i = 0, e = op.getNumPorts(); i != e; ++i) {
+    auto itr = objects.find(i);
+    if (itr == objects.end())
+      continue;
+    auto assoc = itr->second;
+    builder.setInsertionPointAfter(assoc.object);
+    auto sub = ObjectSubfieldOp::create(builder, builder.getUnknownLoc(),
+                                        assoc.object, 0);
+    SmallVector<Value> paths;
+    for (auto id : assoc.associations) {
+      paths.push_back(PathOp::create(
+          builder, builder.getUnknownLoc(),
+          TargetKindAttr::get(op.getContext(), TargetKind::MemberReference),
+          id));
+    }
+    auto list = ListCreateOp::create(
+        builder, builder.getUnknownLoc(),
+        ListType::get(op.getContext(),
+                      cast<PropertyType>(PathType::get(op.getContext()))),
+        paths);
+    PropAssignOp::create(builder, builder.getUnknownLoc(), sub, list);
+  }
+
+  // Erase domain port users, domain ports, and clear the DomainInfo attr.
+  if (auto moduleOp = dyn_cast<FModuleOp>(*op)) {
+    for (auto idx = eraseVector.find_first(); idx >= 0;
+         idx = eraseVector.find_next(idx)) {
+      auto arg = moduleOp.getArgument(idx);
+      for (auto *user : llvm::make_early_inc_range(arg.getUsers())) {
+        if (auto castOp = dyn_cast<UnsafeDomainCastOp>(user)) {
+          castOp.getResult().replaceAllUsesWith(castOp.getInput());
+          castOp.erase();
+          continue;
+        }
+        user->emitOpError() << "cannot be lowered by LowerDomains";
+        return failure();
+      }
+    }
+  }
+  op.erasePorts(eraseVector);
+  op.setDomainInfoAttr(ArrayAttr::get(op.getContext(), {}));
+
+  return success();
+}
+
+void LowerModule::lowerInstances(InstanceGraph &instanceGraph) {
+  auto *node = instanceGraph.lookup(cast<igraph::ModuleOpInterface>(*op));
+  for (auto *use : llvm::make_early_inc_range(*node)) {
+    auto instanceOp = cast<InstanceOp>(*use->getInstance());
+    ImplicitLocOpBuilder builder(instanceOp.getLoc(), instanceOp);
+    auto newInstanceOp = instanceOp.erasePorts(builder, eraseVector);
+    instanceGraph.replaceInstance(instanceOp, newInstanceOp);
+    instanceOp.erase();
+  }
+}
+
+class LowerCircuit {
+
+public:
+  LowerCircuit(CircuitOp circuit, InstanceGraph &instanceGraph)
+      : circuit(circuit), constants(circuit.getContext()),
+        instanceGraph(instanceGraph) {}
+
+  [[nodiscard]] LogicalResult lowerDomain(DomainOp);
+  [[nodiscard]] LogicalResult lowerCircuit();
+
+private:
+  CircuitOp circuit;
+  Constants constants;
+  InstanceGraph &instanceGraph;
+  DenseMap<Attribute, ClassOp> classes;
+};
+
+LogicalResult LowerCircuit::lowerDomain(DomainOp op) {
+  ImplicitLocOpBuilder builder(op.getLoc(), op);
+  auto name = op.getNameAttr();
+  classes.insert(
+      {name, ClassOp::create(builder, name, constants.getDomainPorts())});
+
+  op.erase();
+  return success();
+}
+
+LogicalResult LowerCircuit::lowerCircuit() {
+  LLVM_DEBUG(llvm::dbgs() << "Processing domains:\n");
+  for (auto domain : llvm::make_early_inc_range(circuit.getOps<DomainOp>())) {
+    LLVM_DEBUG(llvm::dbgs() << "  - " << domain.getName() << "\n");
+    if (failed(lowerDomain(domain)))
+      return failure();
+  }
+
+  LLVM_DEBUG(llvm::dbgs() << "Processing modules:\n");
+  DenseSet<InstanceGraphNode *> visited;
+  for (auto *root : instanceGraph) {
+    for (auto *node : llvm::post_order_ext(root, visited)) {
+      auto moduleOp = dyn_cast<FModuleLike>(node->getModule<Operation *>());
+      if (!moduleOp)
+        continue;
+      LLVM_DEBUG(llvm::dbgs() << "  - " << moduleOp.getName() << "\n");
+      LowerModule lowerModule(moduleOp, classes);
+      if (failed(lowerModule.lowerModule()))
+        return failure();
+      lowerModule.lowerInstances(instanceGraph);
+    }
+  }
+
+  return success();
+}
+} // namespace
+
+class LowerDomainsPass : public impl::LowerDomainsBase<LowerDomainsPass> {
+  void runOnOperation() override;
+};
+
+void LowerDomainsPass::runOnOperation() {
+  LLVM_DEBUG(debugPassHeader(this) << "\n";);
+
+  LowerCircuit lowerCircuit(getOperation(), getAnalysis<InstanceGraph>());
+  if (failed(lowerCircuit.lowerCircuit()))
+    return signalPassFailure();
+
+  LLVM_DEBUG(debugFooter() << "\n";);
+}
--- a/include/circt/Dialect/FIRRTL/FIRRTLOpInterfaces.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLOpInterfaces.td
@@ -238,6 +238,9 @@ def FModuleLike : OpInterface<"FModuleLike", [Symbol, PortList, InstanceGraphMod
       return domains[portIndex];
     }]>,
 
+    InterfaceMethod<"Set the domain info of all ports", "void",
+    "setDomainInfoAttr", (ins "::mlir::ArrayAttr":$domainInfo)>,
+
     //===------------------------------------------------------------------===//
     // All Port Information
     //===------------------------------------------------------------------===//
--- a/lib/Dialect/FIRRTL/Transforms/CMakeLists.txt
+++ b/lib/Dialect/FIRRTL/Transforms/CMakeLists.txt
@@ -31,6 +31,7 @@ add_circt_dialect_library(CIRCTFIRRTLTransforms
   LowerCHIRRTL.cpp
   LowerClasses.cpp
   LowerDPI.cpp
+  LowerDomains.cpp
   LowerIntmodules.cpp
   LowerIntrinsics.cpp
   LowerLayers.cpp


test/Dialect/FIRRTL/lower-domains.mlir
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// -----
+
+firrtl.circuit "Foo" {
+  firrtl.domain @ClockDomain {}
+  // The external module should have all domain ports erased and domain
+  // associations removed.
+  //
+  // CHECK-LABEL: firrtl.extmodule @Bar(
```
dtzSiFive, line 206: I would expect extmodule's to have their signature mutated in the same way as module definitions?

dtzSiFive:
Consider this example (mutated second test circuit to not have an extmodule), which given current IR I think is close to simplest example with instantiation:
```mlir
firrtl.circuit "Foo" {
  firrtl.domain @ClockDomain {}
  firrtl.module @Bar(
    in %A: !firrtl.domain of @ClockDomain,
    in %a: !firrtl.uint<1> domains [%A]
  ) { }
  firrtl.module @Foo(
    in %A: !firrtl.domain of @ClockDomain,
    in %a: !firrtl.uint<1> domains [%A]
  ) {
    %bar_A, %bar_a = firrtl.instance bar @Bar(
      in A: !firrtl.domain of @ClockDomain,
      in a: !firrtl.uint<1> domains[A]
    )
    firrtl.matchingconnect %bar_a, %a : !firrtl.uint<1>
  }
}
```

Looks like this doesn't work in the presence of instances? (no test exists demonstrating otherwise and trivial example does not work)

Error:
```
/home/will/src/sifive/circt/test/Dialect/FIRRTL/lower-domains.mlir:11:22: error: 'firrtl.instance' op has a wrong number of results; expected 3 but got 1
    %bar_A, %bar_a = firrtl.instance bar @Bar(
                     ^
/home/will/src/sifive/circt/test/Dialect/FIRRTL/lower-domains.mlir:11:22: note: see current operation: %5 = "firrtl.instance"() <{annotations = [], domainInfo = [[]], layers = [], moduleName = @Bar, name = "bar", nameKind = #firrtl<name_kind droppable_name>, portAnnotations = [[]], portDirections = array<i1: false>, portNames = ["a"]}> : () -> !firrtl.uint<1>
/home/will/src/sifive/circt/test/Dialect/FIRRTL/lower-domains.mlir:3:3: note: original module declared here
  firrtl.module @Bar(
  ^
```

test/Dialect/FIRRTL/lower-domains.mlir
seldridge: I had been thinking that it didn't make sense to include this information as the bodies could not be updated to represent it. However, that shouldn't matter. In a hypothetical linking flow at the FIRRTL level, we'd like to be able to substitute modules for external modules and this wouldn't work if the signatures were not manipulated in the same way.

Good thought.

seldridge:
Something is broken. The `lowerInstances` member function is supposed to be replacing all instances with the result of the module lowering. Thanks for the test case.

_Edit: Typical mistake of walking instances in modules instead of instances of modules._

seldridge:
Instances are fixed in https://github.com/llvm/circt/pull/8929/commits/5c42aeacbe6a7d6cc47ffe69e1d61f253b125b0c. I still need to look at the changes to external modules. (External modeuls are fixed in https://github.com/llvm/circt/pull/8929/commits/9f28b0999f5fa3629e4e39ae7d2e816111a4e56a.)

test/Dialect/FIRRTL/lower-domains.mlir
seldridge: Changed to lower like modules in: https://github.com/llvm/circt/pull/8929/commits/9f28b0999f5fa3629e4e39ae7d2e816111a4e56a

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
+//===- LowerDomains.cpp - Erase all FIRRTL Domain Information -------------===//
```
dtzSiFive: Nit: update this now that the pass is lowering domains to properties?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+// the InferDomains pass) and then lowered (this pass).  This pass has the
+// additional effect of removing all domain information from the circuit.
+//
+//===----------------------------------------------------------------------===//
+
+#include "circt/Dialect/FIRRTL/FIRRTLInstanceGraph.h"
+#include "circt/Dialect/FIRRTL/FIRRTLOpInterfaces.h"
+#include "circt/Dialect/FIRRTL/FIRRTLOps.h"
+#include "circt/Support/Debug.h"
+#include "circt/Support/InstanceGraphInterface.h"
```
dtzSiFive: This should be already included with `FIRRTLInstanceGraph`, I was curious, and probably will always do so.  A bit of IWYU is nice though, no change requested.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+#include "llvm/ADT/MapVector.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "firrtl-lower-domains"
```
dtzSiFive: It'd be neat if we could magically set this to the pass name, as that's usually what we want!

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  return success();
+}
+} // namespace
+
+class LowerDomainsPass : public impl::LowerDomainsBase<LowerDomainsPass> {
+  void runOnOperation() override;
+};
+
+void LowerDomainsPass::runOnOperation() {
+  LLVM_DEBUG(debugPassHeader(this) << "\n";);
```
dtzSiFive: Might be nice to have something like:

```
ScopedDebugPassLogger x(this);
```

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+namespace {
+/// Track information about a domain.
+struct DomainInfo {
+  /// The instantiated class inside a module which contains domain information.
+  ObjectOp op;
+
+  /// The index of the created input port for this domain.  This port does not
+  /// include association information and only contains the information that the
+  /// user must provide.
+  size_t inputPort;
```
dtzSiFive: These are `size_t`, in the code it looks like `unsigned` is used for port indices, and `indexToDomain` uses `int` as the key but is accessed with unsigned key in practice.  I think `unsigned` is used in a few more places, but I'm still going through.

Please pick one type (of some unsigned variety) for port indices and use it.  The `int` made me go looking for using negative as a sentinel 😄 .

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+#include "circt/Dialect/FIRRTL/Passes.h.inc"
+} // namespace firrtl
+} // namespace circt
+
+using namespace circt;
+using namespace firrtl;
+
+namespace {
+/// Track information about a domain.
+struct DomainInfo {
```
dtzSiFive, line 83: A short description (perhaps earlier?) of what "Lowering domains to properties" means would go a long way.

Reading along, "of the created input port for this domain" (why is it created?), "This port does not include association information" (what's that? I am just hearing about that now).

--> 
What ARE these ports?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  /// The instantiated class inside a module which contains domain information.
```
dtzSiFive: Stray thought: will this be a domain once they're classlike (I'm not sure)? Anyway consider making pass to align comments as these land.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  size_t inputPort;
+
+  /// The index of the created output port for this domain, which includes
+  /// association information.
+  size_t outputPort;
+
+  /// The associations to other ports for this domain.  This is in terms of
+  /// distinct attributes that have already been established in the annotations
+  /// for the associated ports.
+  SmallVector<DistinctAttr> associations = SmallVector<DistinctAttr>();
```
dtzSiFive: nit:
```suggestion
  SmallVector<DistinctAttr> associations{};
```

Mostly the full type on both sides is overkill IMO.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+};
+
+class LowerModule {
+
+public:
+  LowerModule(FModuleLike &op,
+              const DenseMap<Attribute, std::pair<ClassOp, ClassOp>> &classes)
```
dtzSiFive: Instead of `std::pair<ClassOp, ClassOp>`, consider a struct to document its meaning, although in at least some places the destructuring `auto [classIn, classOut] =` helps with that presently.

If don't like the struct, maybe a `using` to name this might help?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+      : op(op), eraseVector(BitVector(op.getNumPorts())),
```
dtzSiFive:
```suggestion
      : op(op), eraseVector(op.getNumPorts()),
```

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+        domainToClasses(classes) {}
+
+  // Lower the associated module.  Remove domain ports and remove all domain
+  // information.
```
dtzSiFive: Maybe clarify lowering replaces domain ports and domain information with property ports?

I think I'm suggesting this in different ways, so having this somewhere might be plenty 😆 .

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  // A mapping of old result to new result
+  SmallVector<std::pair<unsigned, unsigned>> resultMap;
+
+  // Mapping of domain name to the lowered input and output class
+  const DenseMap<Attribute, std::pair<ClassOp, ClassOp>> &domainToClasses;
+};
+
+LogicalResult LowerModule::lowerModule() {
+  // Only lower modules or external modules.
+  if (!isa<FModuleOp, FExtModuleOp>(op))
```
dtzSiFive: Can this isa check be combined with grabbing the body? I like it logically separate but generally good to avoid repeated classof checks.  Not worth anything dramatic, but give that a thought please 👍 .

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+    return success();
+
+  // Much of the lowering is conditioned on whether or not his module has a
```
dtzSiFive: or "HER" module, jeez.

jk, should be "this module"

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  // body.  If it has a body, then we need to instantiate an object for each
+  // domain port and hook up all the domain ports to annotations added to each
+  // associated port.
+  Block *body = nullptr;
+  std::optional<OpBuilder> builder;
+  if (auto moduleOp = dyn_cast<FModuleOp>(*op)) {
+    body = moduleOp.getBodyBlock();
+    builder = OpBuilder::atBlockBegin(body);
```
dtzSiFive: Is beginning important? If it's at the end (which to my sensibilities right now seems where new things might naturally go unless there are other considerations-- do these need to be live for use throughout the current body?), you can drop the builder entirely (IIRC MLIR builders are super cheap to construct with virtually no state) which also collapses having two things that must be valid iff the other is.  Instead, just `auto builder = OpBuilder::atBlockBegin(body)` in the two places, and losing the insertion point between is no issue since end is end.

Just a thought, only if that seems better to you 👍 .

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  auto *context = op.getContext();
+
+  // Information about a domain.  This is built up during the first iteration
+  // over the ports.  This needs to preserve insertion order.
+  llvm::MapVector<int, DomainInfo> indexToDomain;
```
dtzSiFive:
```suggestion
  llvm::MapVector<unsigned, DomainInfo> indexToDomain;
```
?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  // insertions.
+  SmallVector<Attribute> portAnnotations;
+
+  // Iterate over the ports, staging domain ports for removal and recording the
+  // associations of non-domain ports.  After this, domain ports will be deleted
+  // and then class ports will be inserted.  This loop therefore needs to track
+  // three indices:
+  //   1. i tracks the original port index.
+  //   2. iDel tracks the port index after deletion.
+  //   3. iIns tracks the port index after insertion.
```
dtzSiFive, line 282: 🏅

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  auto ports = op.getPorts();
+  for (unsigned i = 0, iDel = 0, iIns = 0, e = op.getNumPorts(); i != e; ++i) {
+    auto port = cast<PortInfo>(ports[i]);
+    // Mark domain type ports for removal.  Add information to `domainInfo`.
+    auto domain = dyn_cast<FlatSymbolRefAttr>(port.domains);
+    if (domain) {
+      eraseVector.set(i);
+
+      // Instantiate a domain object with association information.
+      auto port = cast<PortInfo>(op.getPorts()[i]);
```
dtzSiFive: This shadows `port` above and more significantly calls `getPorts()` again instead of using `ports`, or just deleting this line and using `port` from L125? Presumably no mutation is done until after the loop, so no reason to grab the whole PortInfo thing again.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+      auto name = cast<FlatSymbolRefAttr>(port.domains);
+      auto [classIn, classOut] = domainToClasses.lookup(name.getAttr());
```
dtzSiFive:
```suggestion
      auto [classIn, classOut] = domainToClasses.at(name.getAttr());
```
?

Mostly, you very much don't want to have a "default value" constructed for you should it not exist.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+      // Add input and output property ports that encode the property inputs
+      // (which the user must provide for the domain) and the outputs that
+      // encode this information and the associations.
+      newPorts.append(
+          {{iDel,
+            PortInfo(port.name, classIn.getInstanceType(), Direction::In)},
+           {iDel, PortInfo(StringAttr::get(context, Twine(port.name) + "_out"),
+                           classOut.getInstanceType(), Direction::Out)}});
+      portAnnotations.append(
+          {ArrayAttr::get(context, {}), ArrayAttr::get(context, {})});
```
dtzSiFive: Consider creating this once and reusing it (as far outside loops as seems reasonable 😉 ).

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+        // Erase users of the domain in the module body.
+        auto arg = body->getArgument(i);
+        for (auto *user : llvm::make_early_inc_range(arg.getUsers())) {
+          if (auto castOp = dyn_cast<UnsafeDomainCastOp>(user)) {
+            castOp.getResult().replaceAllUsesWith(castOp.getInput());
+            castOp.erase();
+            continue;
+          }
+          user->emitOpError()
```
dtzSiFive: Consider `return user->emitOpError() `...

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+      // Don't increment the iDel since we deleted one port.
+      // Increment the iIns by 2 since we added two ports.
+      iIns += 2;
+      continue;
+    }
+
+    // Record the mapping of the old port to the new port.  This can be used
+    // later to update instances.  This port will not be deleted, so
+    // post-increment both indices.
+    resultMap.push_back({iDel++, iIns++});
```
dtzSiFive:
```suggestion
    resultMap.emplace_back(iDel++, iIns++);
```

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    // If this port has domain associations, then we need to add port annotation
+    // trackers.  These will be hooked up to the Object's associations later.
+    // However, if there is no domain information, then annotations do not need
+    // to be modified.  Early continue first, adding trackers otherwise.
```
dtzSiFive: ✔️

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  for (unsigned i = 0, iDel = 0, iIns = 0, e = op.getNumPorts(); i != e; ++i) {
+    auto port = cast<PortInfo>(ports[i]);
+    // Mark domain type ports for removal.  Add information to `domainInfo`.
+    auto domain = dyn_cast<FlatSymbolRefAttr>(port.domains);
+    if (domain) {
+      eraseVector.set(i);
+
+      // Instantiate a domain object with association information.
+      auto port = cast<PortInfo>(op.getPorts()[i]);
+      auto name = cast<FlatSymbolRefAttr>(port.domains);
```
dtzSiFive:
```suggestion
```
And use `domain` from above instead of `name`.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+    // However, if there is no domain information, then annotations do not need
+    // to be modified.  Early continue first, adding trackers otherwise.
+    ArrayAttr domainAttr = cast<ArrayAttr>(port.domains);
+    if (!domainAttr || domainAttr.empty()) {
+      portAnnotations.push_back(port.annotations.getArrayAttr());
+      continue;
+    }
+
+    SmallVector<Annotation> newAnnotations;
+    for (auto attr : domainAttr) {
```
dtzSiFive: `domainAttr.getAsRange<IntegerAttr>()`

Not huge win, shrug.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+      auto domainIndex = cast<IntegerAttr>(attr).getUInt();
+      auto id = DistinctAttr::create(UnitAttr::get(context));
```
dtzSiFive: If you're caching attributes, `UnitAttr::get(context)` maybe too.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+      newAnnotations.push_back(Annotation(DictionaryAttr::getWithSorted(
```
dtzSiFive: 🙌 `getWithSorted`

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  // Insert new property ports and hook these up to the object that was
+  // instantiated earlier.
+  op.insertPorts(newPorts);
+  if (body) {
+    for (auto const &[_, info] : indexToDomain) {
+      auto [object, inputPort, outputPort, associations] = info;
+      builder->setInsertionPointAfter(object);
+      // Assign input domain info.
+      auto subDomainInfoIn = ObjectSubfieldOp::create(
+          *builder, builder->getUnknownLoc(), object, 0);
```
dtzSiFive: Just wondering, why unknown loc?  Is there not a suitable location to use? It likely won't reach users (?) but if setting one we might as well set to one that tells the right story.  Or, no?

Same for the code that follows.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+      PropAssignOp::create(*builder, builder->getUnknownLoc(), subDomainInfoIn,
+                           body->getArgument(inputPort));
+      auto subAssociations = ObjectSubfieldOp::create(
+          *builder, builder->getUnknownLoc(), object, 2);
+      // Assign associations.
+      SmallVector<Value> paths;
+      for (auto id : associations) {
+        paths.push_back(PathOp::create(
+            *builder, builder->getUnknownLoc(),
+            TargetKindAttr::get(context, TargetKind::MemberReference), id));
```
dtzSiFive: Is this the right TargetKind for what we need? Offhand I believe so, but just asking the question for the expected "of course" 😉 .

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+    llvm::dbgs() << "    portMap:\n";
+    for (auto [oldIndex, newIndex] : resultMap)
+      llvm::dbgs() << "      - " << oldIndex << ": " << newIndex << "\n";
+  });
+
+  return success();
+}
+
+LogicalResult LowerModule::lowerInstances(InstanceGraph &instanceGraph) {
+  auto *node = instanceGraph.lookup(cast<igraph::ModuleOpInterface>(*op));
```
dtzSiFive, line 425: If we like `at` for a checking lookup, then here as well.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  for (auto *use : llvm::make_early_inc_range(node->uses())) {
+    auto instanceOp = cast<InstanceOp>(*use->getInstance());
```
dtzSiFive: InstanceChoiceOp ?

If that's tricky or doesn't make sense for whatever reason, consider just failing (vs crashing).

IIRC instance choice and things like properties/probes are sketchy/unsupported presently, but discussions were made about how probes could basically do the ref ABI thing locally to "continue" the reference string.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+    LLVM_DEBUG(llvm::dbgs()
+               << "      - " << instanceOp.getInstanceName() << "\n");
+
+    for (auto bit : eraseVector.set_bits()) {
+      auto result = instanceOp.getResult(bit);
+      for (auto *user : llvm::make_early_inc_range(result.getUsers())) {
+        if (auto castOp = dyn_cast<UnsafeDomainCastOp>(user)) {
+          castOp.getResult().replaceAllUsesWith(castOp.getInput());
```
dtzSiFive: Consider pulling this pattern out to a helper? At least in its current form. If in the future the lowering of other domain-using operations changes instance-side vs module-side nevermind.

```c++
if (failed(
```

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+          castOp.erase();
+          continue;
+        }
+        user->emitOpError()
```
dtzSiFive: `return user->emitOpError() ` ... ?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+            << "has an unimplemented lowering in the LowerDomains pass.";
+        return failure();
+      }
+    }
+
+    ImplicitLocOpBuilder builder(instanceOp.getLoc(), instanceOp);
+    auto erased = instanceOp.erasePorts(builder, eraseVector);
```
dtzSiFive, line 441: Once of these days we'll make a helper that doesn't require an intermediate extra IntsanceOp .
I believe I put a TODO or note in LOA on the subject, but similarly the reuse with the fiddly arrays is /very/ valuable to not roll in each pass haha, so for now this is the right way!

dtzSiFive: The operation itself isn't a huge deal (if a little unusual), but creating all these attribute arrays that we basically know are immediately (and likely forever) dead is a little unfortunate.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+  ImplicitLocOpBuilder builder(op.getLoc(), op);
+  auto *context = op.getContext();
+  auto name = op.getNameAttr();
+  // TODO: Update this once DomainOps have properties.
+  auto classIn = ClassOp::create(builder, name, {});
+  auto classInType = classIn.getInstanceType();
+  auto pathListType =
+      ListType::get(context, cast<PropertyType>(PathType::get(context)));
+  auto classOut =
+      ClassOp::create(builder, StringAttr::get(context, Twine(name) + "_out"),
```
dtzSiFive, line 502: If attribute caching, consider caching the field name strings as well.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+} // namespace
+
+class LowerDomainsPass : public impl::LowerDomainsBase<LowerDomainsPass> {
+  void runOnOperation() override;
+};
+
+void LowerDomainsPass::runOnOperation() {
+  LLVM_DEBUG(debugPassHeader(this) << "\n";);
+
+  LowerCircuit lowerCircuit(getOperation(), getAnalysis<InstanceGraph>());
```
dtzSiFive: Consider marking instance graph preserved if it indeed is (I see `replaceInstance`).  It may not be since we're adding new nodes and instances.

test/Dialect/FIRRTL/lower-domains.mlir
```diff
@@ -0,0 +1,86 @@
...
+
+// Deeply check the lowering of domains and how associations are lowered.  Later
+// tests will ignore these deep checks.
+firrtl.circuit "Foo" {
+  // Two classes should be created:
+  //   1. A first that just represents the domain.  This has information that a
+  //      user needs to provide.
+  //   2. A second that returns this user-provided information and includes the
+  //      domain information.
+  // CHECK-LABEL: firrtl.class @ClockDomain
```
dtzSiFive, line 11: I have questions about domains and classes and names -- you probably can guess, the usual kind.

Modulo making domains classlike, insofar as classes are nominally typed it may be important to give consideration here.  Probably the associations class is the main question (the other will continue as a domain (?) or at least can re-use the class name, assuming running with same namespace for them), while it's squarely compiler-generated it needs to be the same across SC units.

Something to think about.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,376 @@
...
+          context,
+          {{"class", StringAttr::get(context, "circt.tracker")}, {"id", id}})));
+      indexToDomain[domainIndex].associations.push_back(id);
+    }
+    if (!newAnnotations.empty())
+      port.annotations.addAnnotations(newAnnotations);
+    portAnnotations.push_back(port.annotations.getArrayAttr());
+  }
+
+  // Erease domain ports and clear domain association information.
```
dtzSiFive:
```suggestion
  // Erase domain ports and clear domain association information.
```

test/Dialect/FIRRTL/lower-domains.mlir
```diff
@@ -0,0 +1,86 @@
...
+  firrtl.domain @ClockDomain {}
+  // CHECK-LABEL: firrtl.extmodule @Bar(
+  // CHECK-SAME:    in A: !firrtl.class<@ClockDomain()>
+  // CHECK-SAME:    out A_out: !firrtl.class<@ClockDomain_out(
+  // CHECK-SAME:    in a: !firrtl.uint<1>
+  firrtl.extmodule @Bar(
+    in A: !firrtl.domain of @ClockDomain,
+    in a: !firrtl.uint<1> domains [A]
+  )
+  firrtl.module @Foo(
```
dtzSiFive, line 214: Please include more tests, this is a good start.

More than one domain kind, more than one association.

dtzSiFive (commented):
Looking good! Left a bunch of feedback.

GH started bugging on me so will add a few more comments shortly.

include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -905,4 +905,12 @@ def CheckLayers : Pass<"firrtl-check-layers", "firrtl::CircuitOp"> {
   }];
 }
 
+def LowerDomains : Pass<"firrtl-lower-domains", "firrtl::CircuitOp"> {
```
dtzSiFive, line 912: Consider adding statistics.

include/circt/Dialect/FIRRTL/Passes.td
```diff
@@ -905,4 +905,12 @@ def CheckLayers : Pass<"firrtl-check-layers", "firrtl::CircuitOp"> {
...
+  let summary = "lower domain information to properties";
+  let description = [{
+    Lower all domain information into FIRRTL properties.  This has the effect of
+    erasing all domain information.
```
dtzSiFive, line 916: Does this pass have any prerequisites?  Should these be stated, and can/should they be checked for?
Having neither seems dangerous, but I'm not sure sure what that might entail.

This maybe needs InferDomains to have run (or be equivalent to having run already, maybe, if input is already fully set)? Where in the pipeline is this expected to live -- hopefully part of the overall LowerToHW pipeline (which includes LowerClasses)?

I know there's advice going around (super fair characterization follows!) that passes should yolo their way and crash on folks not swoll enough to use them "properly" but that's basically the same argument against defensive assertions that also is contrary to LLVM ethos and building quality products; if one piece of advice is blindly taken I'd much rather have a compiler (and especially be on a compiler team that works towards) that works every time over ... I'll spare my critique of one blindly following the other way 👍.  Of course there are trade-offs and consideration of best engineering best practices at play here... not everything can or should check everything, so on.

Use your judgement, as I'm sure you would regardless 😁 .

Making passes that don't have silent assumptions they crash on really helps a compiler project overall, and is part of being a high-quality pass in general.  Pipeline debugging (bugpoint-style), reductions (circt-reduce style), pass rearranging, so on.

Anyway 😄 .  I'm not sure this needed so many words, please do what you think is best here 👍 .

I'm honestly partly sincerely interested in what this needs for it to work! So if nothing else please spare words to answer that for me if you would.

test/Dialect/FIRRTL/lower-domains.mlir
dtzSiFive: Can there be output domain ports? If so, them as well.

I assume this requires running long after both LOA and LowerTypes (probably examples of things not worth going out of the way to check? one day that'll be pipeline-guaranteed !), otherwise the usual sort of "does flip work" test.

Having a few more ports to check the insertion/deletion logic.

Multiple instances?

Anyway I'm sure you got it, just offering a few more suggestions.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: I pruned this down to what was only needed in: a19e7d83d

I do like the IWYU notion, though. One thing I noticed is that things like `circt/Support/LLVM.h` inclusion in some headers mean that there is very little you have to include. (I was surprised that I could prune this down as much as I could in a19e7d83d.)

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: b269c5d9b

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Ugh... this has annoyed me for a while. There's no nice `char *` readily available. However, I did find a way to do it that looks fine. See what you think about: 4ab6d8541

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Added in d410fa83a. Used a bit in 30cd87cc2. Will switch to that here, too.

seldridge: This had some issues... I reverted, fixed in a6742c5e3 and re-rolled out in e45f124df and a6c5d82cb.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Good catch. The `int` stemmed from an earlier version which used `-1` as a sentinel for when this was not set. However, the current version doesn't have this. I'll switch this to `unsigned` everywhere because this is the type that is used for `insertPorts`. I'd probably prefer `size_t` as that is the type returned by `getNumPorts()`, however, that doesn't align nicely with `insertPorts`.

seldridge: Fixed in: 39455de4c

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: This makes it a little terse for my liking. However, I think you're onto something here. I switch to an `else if` after the failed `FModuleOp` dynamic cast in: 0bc4a351b

There's some formulation involving a `TypeSwitch` that I think would be the most readable. I tried a couple of formulations using this and couldn't get something that I liked.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Cleaned up with earlier comments in 9a3fbbb27, also 879d221e9.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: 2af51927d

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Switched to a struct in: 63c613e40

Yeah, I agree, there's no places using `.first` and `.second`, so it's not as it could have been.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: 👍 Fixed in: 2b6082c62

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Some of the new language in 9a3fbbb27 refers to this as the "domain-lowered class". The class-like vs. not class-like is really just about allowing for domains to have input properties, output properties, and operations in their bodies. Currently, they have no ports and have bodies (but you can't put anything in them).

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Added better commentary in: 9a3fbbb27

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: https://github.com/llvm/circt/commit/39455de4c14372f3f93dbb11b5b379cca7f17247

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: There was also careful selection of variable names so that the loop would all fit on one line. An earlier version of this was doing an `llvm::enumerate(op.getPorts)` with these indices declared before the loop, but only used in the loop. I like the current formulation a lot better.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Nice catch, just a mistake. Fixed in: cf119aa92

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: This used to be inverted with `if (has associations) { update assocations }`. However, I found it cleaner inverted. However, this necessitated this verbose comment as I figured this might be confusing. I do wish we had a good pattern for "this comment applies to this region of code". Maybe such a thing is commonplace elsewhere?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: I will loop back to this a bit later. I originally had a whole `Constants` class that I was using for this kind of thing which I eventually deleted. I'll revive this in light of this comment and some later ones about attribute caching.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: It's completely free here and no reason not to. Leaving performance on the table if it doesn't do this. 😅

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: No `at` member function in `InstanceGraph`. We should add that, though!

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Yeah, this situation is unfortunate. The other confusing bit was that `erasePorts` RAUWs ports, but `cloneAndInsert` doesn't. I initially wrote this with a bug where I was trying to RAUW from the original to the inserted (which seemed expected) while I had to go from deleted to inserted.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: afbff057a

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: 6b6bcece9

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Saves a line. SGTM. Changed in: 0f424d166

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Thanks. Fixed in: bf7eed166

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Switched in: 5aa80fad7

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: 👍 Changed in: d7e38d4fd

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: I didn't know about this member function. This is exactly what I want here. Thanks! Fixed in: f2ba298c5

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Nice idea to not keep a builder around and instead create it when needed. I didn't like the fact that the `Block *body` and `std::optional<OpBuilder>` had to be kept in sync. This avoids all of it. This may help switch to something like a `TypeSwitch` earlier when branching based on the module type.

Changed to create a builder on demand in 0656383a0. I will loop back to the placement question in a later commit. Putting it at the end has some benefits in not disturbing any existing IR, e.g., numbering. This is admittedly superficial, but "appending" this seems more logical.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: 5a4731c06

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Switched to `atBlockEnd` in: 4c16c47aa

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Over-engineered constant pool added in: 30836741f

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Over-engineered constant pool added in 30836741f.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Over-engineered constant pool added in: 30836741f

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: It actually looks like `.lookup` is closer to `.at` already if building with asserts on. However, there is no dereferencing of the `node` that occurs (like there is with the `DenseMap` dereferencing its iterator). I think there is more of a discussion here about the architecture of the `InstanceGraph`. Should `.lookup` be returning a default-constructed (sentinel) `InstanceGraphNode` as opposed to asserting / returning null? This would allow separation into `lookup` / `at`.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Surprisingly, this is actually unreachable because `InstanceChoiceOp` does not implement the `InstanceOpInterface` so the instance graph never indexes it. I think this is hinting at known, serious deficiencies with `InstanceChoiceOp` that need to be rectified. I'll figure out some way to patch this up so that this does signal pass failure.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Marking everything preserved in: 93219bc65

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Added an unreachable `dyn_cast` and check in: e64bb6e22

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Good question. I checked and according to internal documentation this should be a `TargetKind::Reference`. Switched in: 8a38ce3c2

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Moved to actual location info in: d382a4d01 Almost all of the location info should be associated with the domain, which can be grabbed from the object if properly set on the object. I started treating the association info as coming from the original port (as this is where it is in the FIRRTL text). This requires using a new struct to track both the `DistinctAttr` and the `Location`.

include/circt/Dialect/FIRRTL/Passes.td
seldridge: Added one statistic in: dd1d08a0e

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Nice idea. I added a member function to `lowerModule` to handle this in: 930ec0d34

test/Dialect/FIRRTL/lower-domains.mlir
seldridge: Yeah, good point. The re-use of the existing domain name for the "input" class works. The output class will need to have an ABI-determined name. While I think it's probably fine to keep this as `<name>_out` for now, this should likely be something better like `<name>$domainOut`. Or: something that uses `$` to prevent collision and is impossible to collide with. Same thing for the port name.

I won't treat this as blocking.

test/Dialect/FIRRTL/lower-domains.mlir
seldridge: - More than one kind and multiple associations: 2a2751901
- Output domain ports (logic to handle this) and a test: a3d66f7a1
- Multiple ports: 813c31747

I don't think that multiple instances adds much just based on how the logic works. I'll punt on that one.

include/circt/Dialect/FIRRTL/Passes.td
seldridge: Known prerequisites: this needs to run after infer domains (which is WIP right now) and before lower classes. If run before infer domains, it won't crash or anything like that, but it will produce incomplete results that differ from what would happen if run after infer domains. E.g., if association information is missing, then it will claim that those ports have no associations.

The current limitations of the representation of domains (that ports are either domain ports or have associations) means that this assumes a post-LOA representation. However, there is no way to actually represent this pre-LOA right now. The pass doesn't require LowerTypes currently, though that is again due to the lack of the ability to have associations on fields. Once per-field associations are available, the pass will need an update.

Clearly, as the representational restrictions are lifted, the pass will need to gain capabilities.

Yes, the pass is expected to be in the `LowerToHW` pipeline. This adds _a lot_ of distinct attr annotations to ports which will have a detrimental effect to the ability of the module to be optimized. For that reason it needs to be as late as possible.

seldridge: 1d3f95aa8 adds top-level commentary summarizing the above comment.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,517 @@
...
+// domain port is lowered to an output of type (2).
+//
+// Intuitively, (1) is the information that a user must specify about a domain
+// and (2) is the associations for that domain.
+//
+// This pass needs to run after InferDomains and before LowerClasses.  This pass
+// assumes that all domain information is available.  It is not written in such
+// a way that partial domain information can be lowered incrementally, e.g.,
+// interleaving InferDomains and LowerDomains with passes that incrementally add
+// comain information will not work.  This is because LowerDomains is closer to
```
dtzSiFive:
```suggestion
// domain information will not work.  This is because LowerDomains is closer to
```

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,517 @@
...
+// properties.  This is part of the compilation of FIRRTL domains where they are
+// inferred and checked (See: the InferDomains pass) and then lowered (this
+// pass).  After this pass runs, all domain information has been removed from
+// its original representation.
+//
+// Each domain is lowered into two classes: (1) a class that has the exact same
+// input/output properties as its corresponding domain and (2) a class that is
+// used to track the associations of the domain.  Every input domain port is
+// lowered to an input of type (1) and an output of type (2).  Every output
+// domain port is lowered to an output of type (2).
```
dtzSiFive, line 19: Until we have our domain define connectlike (#9067) maybe this is moot, but....

What drives the added object of type (1) that are now input ports?

Conceptually, I mean, if nothing else.  In some cases this can be from the parent's input domain port.  When local to a module (whatever "allocate a domain" looks like), it's from that point I'd imagine.

This brings us to the example I was bringing up the other day on this subject re:proposed lowering and asymmetric handling of input/output domain information.

Basically, I'd expect the lowered domain plumbing to mirror that of the domain connections originally:
If a domain input port is (in the FIRRTL) driven by a sibling's output port, where does the domain object of type (1) come from?

By lowering (2) to associations.. ah, there it is.
Okay so they're both (at least) objects of type (1), (2) additionally includes a list of association paths.

So presumably the answer to this question is that we'd grab the domain object out of the output of type (2) from the sibling instance and drive that to the input port (now expecting something of type (1)).

Okay so -- just how I like thinking about these things:

1) All domain values and connectivity become type (1) and propassign instead of domain define.
2) Output ports have association information bundled into the required-for-correctness-of-IR threading of type (1) objects (lowered domain values) which implicitly also ties the two together for easy processing agnostic of knowledge of, say, domain port names.

Anyway this breakdown isn't how the code does things, and that's all good, but hopefully you follow.  We transform domains to classes, and replace values with objects of that class, and connectivity is basically the same throughout.  For output ports we instead drive into the associations object (and grab it back out as needed).

Does this make sense, anything wrong/missing here?
Thanks!

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,517 @@
...
+// Intuitively, (1) is the information that a user must specify about a domain
+// and (2) is the associations for that domain.
+//
+// This pass needs to run after InferDomains and before LowerClasses.  This pass
+// assumes that all domain information is available.  It is not written in such
+// a way that partial domain information can be lowered incrementally, e.g.,
+// interleaving InferDomains and LowerDomains with passes that incrementally add
+// comain information will not work.  This is because LowerDomains is closer to
+// a conversion than a pass.  It is expected that this is part of the FIRRTL to
+// HW pass pipelined.
```
dtzSiFive:
```suggestion
// HW pass pipeline.
```

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,517 @@
...
+  /// A vector of minimal association info that will be hooked up to the
+  /// associations of this ObjectOp.
+  SmallVector<AssociationInfo> associations{};
+};
+
+/// Store of the two classes created from a domain, an input class (which is
+/// one-to-one with the domain) and an output class (which tracks the input
+/// class and any associations).
+struct Classes {
+
```
dtzSiFive:
```suggestion
```

Consistent with other structs in this file (and offhand, I think most of CIRCT).

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,517 @@
...
+  /// The index of the output port that the ObjectOp will be connected to.  This
+  /// port communicates back to the user information about the associations.
+  unsigned outputPort;
+
+  /// A vector of minimal association info that will be hooked up to the
+  /// associations of this ObjectOp.
+  SmallVector<AssociationInfo> associations{};
+};
+
+/// Store of the two classes created from a domain, an input class (which is
```
dtzSiFive:
```suggestion
/// Struct of the two classes created from a domain, an input class (which is
```
?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Yeah, this is correct. The lowered class plumbing will mirror that of the original domain plumbing. Things may get optimized, though. Also, aliasing is possible (but not problematic).

### Aliasing Example

Consider the following two circuits. The first has two output aliased domains of `A`:

``` firrtl
domain ClockDomain:
  input name_in: String
  output name_out: String

  propassign name_out, name_in

module Foo:
  input A: Domain of ClockDomain
  output B: Domain of ClockDomain
  output C: Domain of ClockDomain
  output a: UInt<1> domains [A]
  output b: UInt<1> domains [B]
  output c: UInt<1> domains [C]

  invalidate a
  invalidate b
  invalidate c
  
  domain_define B = A
  domain_define C = A
```

``` firrtl
domain ClockDomain:
  input name_in: String
  output name_out: String

  propassign name_out, name_in

module Foo:
  input A: Domain of ClockDomain
  output a: UInt<1> domains [A]
  output b: UInt<1> domains [A]
  output c: UInt<1> domains [A]

  invalidate a
  invalidate b
  invalidate c
```

When this eventually hits the OM APIs in the final MLIR, the shape of the module `Foo` will look different for the first circuit and for the second. In the first, the user will have to provide a single clock domain object, setting the `name_in` parameter. They will then get three domains out, each associated with one port. However, the user will need to recognize that all of these are aliases as they have the same name.

There is no aliasing in the latter.

If this is an internal module, it is reasonable that the first module is optimized into the second. If this is a public module, then (I think) it cannot be optimized and it's up to the user to deal with the aliasing.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Changed in: d683f147f

I was trying to avoid saying "Struct" to avoid the tautological comment. ("Of course a struct is a struct!") Anyway, it's too cute.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: d683f147f

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: d683f147f

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: d683f147f

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+/// Thread safe, lazy pool of constant attributes
+class Constants {
+
+public:
+  Constants(MLIRContext *context) : context(context) {}
+
+  // Return an empty ArrayAttr.
+  ArrayAttr getEmptyArrayAttr() {
+    if (!emptyArrayAttr) {
```
dtzSiFive: Awesome! Thanks for putting this together!  I wonder if we could reuse this / similar pattern to extract out re:attribute constants on [pass initialization](https://mlir.llvm.org/docs/PassManagement/#initialization) or otherwise get a bit of reuse out of the similar structures across some of our passes? Just a thought / future work 👍 .

Unfortunately, this has a [race](https://en.wikipedia.org/wiki/Race_condition#Example_definitions_of_data_races_in_particular_concurrency_models) ([bit more C++-specific here](https://en.cppreference.com/w/cpp/language/multithread.html)), suffers from classic singleton initialization problem (two threads can take the "true" branch before either writes; usually that means double-allocation or conflicting writes but here it's two writes of same; see https://en.wikipedia.org/wiki/Double-checked_locking for some related flavor).

Maybe just initialize this on pass construction (or `lowerCircuit`) unconditionally?

They're not so expensive that creating them is worth avoiding doing once (especially because soon we expect literally every circuit to have domains, yeah? so is avoiding one-time construction for no-domain case more important than slowing down access when domain-rich?), but they add up and become costly when done either in tight loops or from multiple threads at same time, and especially both.

OTOH lazy initialization like what's here works if we drop the mutex and embrace this pass not having internal parallelism 👍 .

On the subject, we leverage word-granular data-races and redundant writes of same value being "fine" in the way we mutate some FIRRTL types internally re:caching things like passive type.  This is wrong according to the language specification but is compiled to something that works on most CPU's and rightfully things like thread-sanitizer will complain.  That's a little trickier, but this comment is long enough already 😄 .

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+  }
+
+  StringAttr getAssociationsOut() {
+    constructFieldNameAttrs();
+    assert(associationsOut);
+    return associationsOut;
+  }
+
+private:
+  /// Mutex indicating that, when held, allows for mutation.
```
dtzSiFive: mutex needs to be on read as well as write, but I think I hammered that enough in my earlier comment.

I am not categorically against relying on word-size writes being atomic, but language specification says this is a race and would prefer we resorted to that only in places where it really helps save us (and ideally only relied on it on platforms we know that's true for?).

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+  StringAttr domainInfoOut;
+  StringAttr associationsIn;
+  StringAttr associationsOut;
+};
+
+class LowerModule {
+
+public:
+  LowerModule(FModuleLike &op, const DenseMap<Attribute, Classes> &classes,
+              Constants &constants)
```
dtzSiFive: IF this is eagerly constructed, please make this const.  Lazy is good too, just noting if go other way.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+  // ports.
+  LogicalResult lowerModule();
+
+  // Lower all instances of the associated module.  This relies on state built
+  // up during `lowerModule` and must be run _afterwards_.
+  LogicalResult lowerInstances(InstanceGraph &);
+
+private:
+  /// Erase all users of domain type ports.
+  LogicalResult eraseDomainUsers(Value value) {
```
dtzSiFive, line 212: 🎉

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+        continue;
+      }
+      return user->emitOpError()
+             << "has an unimplemented lowering in the LowerDomains pass.";
+    }
+    return success();
+  }
+
+  // The module this class is lowering
+  FModuleLike &op;
```
dtzSiFive: I think this should be by-value, like most MLIR things (operations/values/attributes for sure).

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+  // three indices:
+  //   1. i tracks the original port index.
+  //   2. iDel tracks the port index after deletion.
+  //   3. iIns tracks the port index after insertion.
+  auto ports = op.getPorts();
+  for (unsigned i = 0, iDel = 0, iIns = 0, e = op.getNumPorts(); i != e; ++i) {
+    auto port = cast<PortInfo>(ports[i]);
+    // Mark domain type ports for removal.  Add information to `domainInfo`.
+    auto domain = dyn_cast<FlatSymbolRefAttr>(port.domains);
+    if (domain) {
```
dtzSiFive:
```suggestion
    if (auto domain = dyn_cast<FlatSymbolRefAttr>(port.domains)) {
```
If you want

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+    // post-increment both indices.
+    resultMap.emplace_back(iDel++, iIns++);
+
+    // If this port has domain associations, then we need to add port annotation
+    // trackers.  These will be hooked up to the Object's associations later.
+    // However, if there is no domain information, then annotations do not need
+    // to be modified.  Early continue first, adding trackers otherwise.  Only
+    // create one tracker for all associations.
+    ArrayAttr domainAttr = cast<ArrayAttr>(port.domains);
+    if (!domainAttr || domainAttr.empty()) {
```
dtzSiFive: Can this be null and have the cast succeed? I don't /think/ so but I've seem something weird in this area that I won't say this is for sure wrong.  Do you know?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+      portAnnotations.push_back(port.annotations.getArrayAttr());
+      continue;
+    }
+
+    SmallVector<Annotation> newAnnotations;
+    DistinctAttr id;
+    for (auto indexAttr : domainAttr.getAsRange<IntegerAttr>()) {
+      if (!id) {
+        id = DistinctAttr::create(UnitAttr::get(context));
```
dtzSiFive, line 350: Do we usually reuse distinctattr's?  I don't think so, but not sure that that is important.  Do you know/remember @youngar ?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+  // Erase domain ports and clear domain association information.
+  op.erasePorts(eraseVector);
+  op.setDomainInfoAttr(constants.getEmptyArrayAttr());
+
+  // Insert new property ports and hook these up to the object that was
+  // instantiated earlier.
+  op.insertPorts(newPorts);
+
+  if (body) {
+    for (auto const &[_, info] : indexToDomain) {
```
dtzSiFive, line 371: https://wg21.link/p2169 -- c++26!

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+        auto subDomainInfoIn =
+            ObjectSubfieldOp::create(builder, object.getLoc(), object, 0);
+        PropAssignOp::create(builder, object.getLoc(), subDomainInfoIn,
+                             body->getArgument(*inputPort));
+      }
+      auto subAssociations =
+          ObjectSubfieldOp::create(builder, object.getLoc(), object, 2);
+      // Assign associations.
+      SmallVector<Value> paths;
+      for (auto [id, loc] : associations) {
```
dtzSiFive, line 389: I love the destructuring!

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+void LowerDomainsPass::runOnOperation() {
+  CIRCT_DEBUG_SCOPED_PASS_LOGGER(this);
+
+  LowerCircuit lowerCircuit(getOperation(), getAnalysis<InstanceGraph>(),
+                            numDomains);
+  if (failed(lowerCircuit.lowerCircuit()))
+    return signalPassFailure();
+
+  markAllAnalysesPreserved();
```
dtzSiFive, line 561: I mentioned in other comment but I don't know it was resolved and can't find it now:

I don't see where we add the new FModuleLike's and their instances to the instance graph?
On a related note, quick bounce through and I don't see a way to do this.

But since I'm not sure this landed before: my concrete concern is advertising the InstanceGraph as preserved (==correct for the IR after the pass is done) but due to lacking added instances and modules is now incomplete (=incorrect, in a sense).

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
dtzSiFive: Thank you! This makes sense.  I was wondering how much inspect-what-is-there will be done in the consuming API's.  I'm a little surprised we're on board with having the set of domain association objects being dependent on optimizations, but that almost takes us to point of preserving a module just to run a property out that says the domain has no associations.

OTOH since we're not splitting these apart at the MLIR level we could gather associations at the top of public modules (like OMIR would do for children nodes, IIRC?).
Nevermind that, mostly, for now 😄 .

That does raise question of how the API's might reason about domains they are provided associations for that they don't know the origin of (similarly but slightly less obvious than needing to provide information for domains not originally part of the module's interface).

I'll move this conversion to a chat 👍 .

dtzSiFive (approved):
LGTM!

Awesome, excellent, work!! I love how very well structured your code always is ❤️ makes it a joy to read.  Even tests are aligned (a pattern I think we've started "enforcing"/requesting generally).  Comments are fantastic and really help, appreciate it.  Thank you!

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: I didn't think this was actually `c++26` as if the compiler isn't interpreting this as being 26, it would be no different from assigning this to a variable called `_`? Do you know if there is any problem here?

Okay, I think this is okay. `clang-17` is only throwing warnings if I have two `_` variables in scope at the same time. If I just have one, then that should be okay. WDYT?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: I wasn't able to make the cast success if this was null. I think it would have to be a `cast_or_null`.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: As I was writing this, it definitely felt like something that should be reusable.

> OTOH lazy initialization like what's here works if we drop the mutex and embrace this pass not having internal parallelism 👍 .

Note: this pass has no internal parallelism, but I have an eye towards making it parallel. The one trick here is that there's a missing utility missing for organizing a thread pool based on dependencies in the instance graph. I wanted to build this as a follow on to @fabianschuiki's work here: https://github.com/llvm/circt/commit/8b46321ea2f9db0f50808b654cff3003dad50ae8). I just haven't gotten around to it as I wanted to do the parallelism using this unwritten utility. In a very quick follow-on, let me see if I can do this the way I want so that this lazy initialization code isn't over-engineered for a pass that isn't parallel.

The double set should be fixed in: e61396770 Can you check that this looks right?

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Should be fixed in: e61396770 Check me, though.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Opting for continued lazy in: e61396770

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Yeah, the reference here doesn't make sense. Switched in: 2a14f64cd Good catch.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Yep, this is cleaner. Fixed in: d8ba5206c

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Ah, I missed a few spots for this. This _should_ be fixed with: 475f994b8

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
dtzSiFive: Oh no I'm so sorry I was just having fun!  It's valid as an identifier AFAIK in all versions of c++ (especially any we might possibly care about).  I should have explained, apologies.

I've been going through new and proposed C++ features lately and it'll be fun to have its use in the way you're doing here enshrined -- specifically as it allows redeclaring where today that's an error like any other identifier, and tidies a few other things.

There is no issue here AFAIK.
Your assessment is correct! Thanks!

seldridge: SGTM. Yeah, this is one of the features that I'm looking forward to using in 26. 🥺

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Switched to `llvm::call_once` in 3b6cb9a83. Thanks for the pointer to this!

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Resolving this in favor of the earlier discussion.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
dtzSiFive: Yeah, it shouldn't work I don't think! Okay so the `!domainAttr` portion is unnecessary (and perhaps misleading)?

seldridge: Ah, sorry, yes. This check is entirely dead now. This is vestigial from an earlier version.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Talked with Andrew about this offline and, while we don't seem to do this currently, there should be no problem with it.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: d4ead2dc9

PR #8932, opened 2025-09-08
[Synth] Support MIG lowering in CombToAIG conversion pass

uenoku:
This patch extends the synthesis pipeline to support lowering combinational operations to Majority-Inverter Graphs (MIG) in addition to And-Inverter Graphs (AIG). The implementation adds a CombToAIGTarget enum with AIG/MIG options, implements MIG-specific conversion patterns for OR operations using majority functions, updates adder carry logic to use majority functions in MIG mode, and renames the pipeline from AIGLowering to CombLowering to reflect the generalized capability. The MIG representation uses 3-input majority functions (MAJ(a,b,c) = a&b | a&c | b&c) which are more efficient than AIG for arithmetic operations. Integration tests verify functional equivalence using circt-lec.

In the follow-up I'm planning to rename CombToAIG to CombToSynth so reflect the pass can lower IR to both AIG and MIG.


lib/Conversion/CombToAIG/CombToAIG.cpp
```diff
@@ -95,6 +112,27 @@ static Value createShiftLogic(ConversionPatternRewriter &rewriter, Location loc,
...
 
+// Return a majority operation if MIG is enabled, otherwise return a majority
+// function implemented with Comb operations. In that case `carry` has slightly
+// smaller depth than the other inputs.
+static Value createMajorityFunction(OpBuilder &rewriter, Location loc, Value a,
+                                    Value b, Value carry,
+                                    bool useMajorityInverterOp) {
+  if (useMajorityInverterOp) {
+    SmallVector<Value, 3> inputs = {a, b, carry};
+    SmallVector<bool, 3> inverts = {false, false, false};
```
fabianschuiki: Maybe `std::array` would work here, too?

PR #8935, opened 2025-09-10
[ConvertToArcs] Allow ops with regions

fabianschuiki:
Tweak the ConvertToArcs pass to allow operations with regions. The pass would just give up on these ops until now. However, in most cases we can handle these ops just fine, as long as we take care to properly gather all values used in nested operations. Also mark all side-effecting ops as arc breakers, to prevent them from being gobbled up into arc defines, where the verifier would be unhappy about them.


lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
```diff
@@ -132,32 +135,43 @@ LogicalResult Converter::runOnModule(HWModuleOp module) {
...
+  SmallSetVector<Value, 2> seenOperands;
+  auto addToWorklist = [&](Operation *op) {
+    seenOperands.clear();
+    for (auto operand : op->getOperands())
+      seenOperands.insert(operand);
+    for (auto &region : op->getRegions())
+      mlir::visitUsedValuesDefinedAbove(region, region, [&](auto *operand) {
+        seenOperands.insert(operand->get());
+      });
+    worklist.emplace_back(op, seenOperands.takeVector());
```
uenoku: I think takeVector moves internal vector so I think it's illegal to reuse `seenOperands` later even when there is `clear`.

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
```diff
@@ -132,32 +135,43 @@ LogicalResult Converter::runOnModule(HWModuleOp module) {
...
+  SmallVector<std::tuple<Operation *, SmallVector<Value, 2>>> worklist;
+  SmallSetVector<Value, 2> seenOperands;
+  auto addToWorklist = [&](Operation *op) {
+    seenOperands.clear();
+    for (auto operand : op->getOperands())
+      seenOperands.insert(operand);
+    for (auto &region : op->getRegions())
+      mlir::visitUsedValuesDefinedAbove(region, region, [&](auto *operand) {
+        seenOperands.insert(operand->get());
+      });
```
uenoku: Maybe this works? https://github.com/llvm/llvm-project/blob/5539daf8120cfe20c0c36d115319e14077e5fa79/mlir/lib/Transforms/Utils/RegionUtils.cpp#L77
```suggestion
mlir::getUsedValuesDefinedAbove(op->getRegions(), seenOperands);
```

fabianschuiki: I tried `getUsedValuesDefinedAbove` first, but that uses a `SetVector` instead of a `SmallSetVector`. I was thinking that with most operations probably having 0-2 operands it might make sense to avoid allocations :thinking:

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
fabianschuiki: Ah good point, I forgot about that :+1:

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
fabianschuiki: Actually now that I think about it, it's probably better to just use `getUsedValuesDefinedAbove` and create `SmallVector`s from it. Makes the whole thing less messy to read.

PR #8937, opened 2025-09-10
[circt-verilog] Add commandfile input option.

jpienaar:
This follow behavior of same flag in slang. Goal is to be able to use filelists as inputs rather than specify individual files.

### DIFF
--- a/tools/circt-verilog/circt-verilog.cpp
+++ b/tools/circt-verilog/circt-verilog.cpp
@@ -267,6 +267,13 @@ struct CLOptions {
           "One or more library files, which are separate compilation units "
           "where modules are not automatically instantiated."),
       cl::value_desc("filename"), cl::Prefix, cl::cat(cat)};
+
+  cl::list<std::string> commandFiles{
+      "C",
+      cl::desc(
+          "One or more library files, which are independent compilation units "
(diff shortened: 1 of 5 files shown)


tools/circt-verilog/circt-verilog.cpp
```diff
@@ -267,6 +267,13 @@ struct CLOptions {
           "One or more library files, which are separate compilation units "
           "where modules are not automatically instantiated."),
       cl::value_desc("filename"), cl::Prefix, cl::cat(cat)};
+
+  cl::list<std::string> commandFiles{
+      "C",
+      cl::desc(
+          "One or more library files, which are independent compilation units "
+          "where modules are automatically instantiated."),
```
fabianschuiki: This might be outdated / copied from the library files option above. Doesn't this open a file wiht additional command line arguments? Or a Verilog file list?

fabianschuiki (approved):
LGTM!

tools/circt-verilog/circt-verilog.cpp
jpienaar: With additional arguments, but a restricted set of options (if I see correctly: macros, libraries, include dirs). I'll update to say command files. Then at least its consistent and discoverable yes.

PR #8938, opened 2025-09-10
[Reduce] Use stable sort, prune unused symbols, tweak FIRRTL reductions

fabianschuiki:
Use a stable sort in the reducer to sort matches. The binary search over the matchese expects them to be in the order that they were matched in the IR when matches have the same benefit.

Add a generic reduction to prune unused symbols.

Make the FIRRTL annotation remover check if annotations are already empty, and not match on those ops. This significantly improves reduction speed.

Fix an issue in the FIRRTL module externalizer, where `FExtModule`s require an additional list of known layers that needs to be created. Also, don't externalize modules that participate in NLAs, since removing those would invalidate the NLA.

### DIFF
--- a/tools/circt-reduce/circt-reduce.cpp
+++ b/tools/circt-reduce/circt-reduce.cpp
@@ -284,7 +284,7 @@ static LogicalResult execute(MLIRContext &context) {
 
       // Sort the matches by benefit. This will cause us to try the most
       // beneficial matches first.
-      llvm::sort(matches, [](auto &a, auto &b) {
+      llvm::stable_sort(matches, [](auto &a, auto &b) {
         if (a.benefit > b.benefit)
           return true;
         if (a.benefit < b.benefit)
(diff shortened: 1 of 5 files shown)


uenoku (approved):

tools/circt-reduce/circt-reduce.cpp
```diff
@@ -366,8 +366,9 @@ static LogicalResult execute(MLIRContext &context) {
       // Reduce the chunk size to achieve the minimum number of chunks requested
       // by the user.
       if (minChunks > 0)
-        rangeLength = std::min<size_t>(rangeLength,
-                                       std::max<size_t>(opIdx / minChunks, 1));
+        rangeLength = std::min<size_t>(
```
uenoku: Could you use std::clamp and llvm::divideCeil?

fabianschuiki: Nice, let me tweak that! :+1:

PR #8944, opened 2025-09-11
[SeqToSV] Use XMR to set up async reset and preset of firreg under ifdef

rwy7:
With inline layers, firregs may now appear under ifdef blocks. This PR updates the generation of the reset IR to reference a firreg by XMR if the register is buried under an ifdef.

### DIFF
--- a/lib/Conversion/SeqToSV/FirRegLowering.cpp
+++ b/lib/Conversion/SeqToSV/FirRegLowering.cpp
@@ -357,6 +357,8 @@ void FirRegLowering::createRandomInitialization(ImplicitLocOpBuilder &builder) {
 
 void FirRegLowering::createPresetInitialization(ImplicitLocOpBuilder &builder) {
   for (auto &svReg : presetInitRegs) {
+    OpBuilder::InsertionGuard guard(builder);
+
     auto loc = svReg.reg.getLoc();
     auto elemTy = svReg.reg.getType().getElementType();
     auto cst = getOrCreateConstant(loc, svReg.preset.getValue());
@@ -367,7 +369,13 @@ void FirRegLowering::createPresetInitialization(ImplicitLocOpBuilder &builder) {
     else
       rhs = hw::BitcastOp::create(builder, loc, elemTy, cst);
 
-    sv::BPAssignOp::create(builder, loc, svReg.reg, rhs);
+    buildRegConditions(builder, svReg.reg);
+    Value target = svReg.reg;
+    if (svReg.path)
+      target = buildXMRTo(builder, svReg.path, svReg.reg.getLoc(),
+                          svReg.reg.getType());
+
+    sv::BPAssignOp::create(builder, loc, target, rhs);
   }
 }
 
--- a/test/Dialect/Seq/firreg.mlir
+++ b/test/Dialect/Seq/firreg.mlir
@@ -1043,3 +1043,29 @@ hw.module @AsyncResetRegUnderIfdef(in %clock : !seq.clock, in %reset : i1, in %v
   // CHECK: }
   hw.output
 }
+
+// Test for registers with "preset".
+
+// CHECK:  hw.hierpath @[[reg_path:.+]] [@PresetRegUnderIfdef::@reg]
+hw.module @PresetRegUnderIfdef(in %clock : !seq.clock, in %value : i1) {
+  %c = hw.constant 0 : i1
+
+  // CHECK: sv.ifdef @MyMacro {
+  // CHECK:   %reg = sv.reg sym @reg : !hw.inout<i1>
+  // CHECK:   %0 = sv.read_inout %reg : !hw.inout<i1>
+  // CHECK:   sv.always posedge %clock {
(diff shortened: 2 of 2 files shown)


rwy7:
Updated to also fix the same issue for preset registers.

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -360,7 +367,15 @@ void FirRegLowering::createPresetInitialization(ImplicitLocOpBuilder &builder) {
...
-    sv::BPAssignOp::create(builder, loc, svReg.reg, rhs);
+    auto *block = builder.getBlock();
+    buildRegConditions(builder, svReg.reg);
+    Value target = svReg.reg;
+    if (svReg.path)
+      target = buildXMRTo(builder, svReg.path, svReg.reg.getLoc(),
+                          svReg.reg.getType());
+
+    sv::BPAssignOp::create(builder, loc, target, rhs);
+    builder.setInsertionPointToEnd(block);
```
seldridge: Consider using `OpBuilder::InsertionGuard` here as opposed to recording the block, then restoring it.

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -373,11 +388,19 @@ void FirRegLowering::createAsyncResetInitialization(
...
+      builder.setInsertionPointToEnd(thenBlock);
+      buildRegConditions(builder, reg.reg);
+      Value target = reg.reg;
+      if (reg.path)
+        target =
+            buildXMRTo(builder, reg.path, reg.reg.getLoc(), reg.reg.getType());
+      sv::BPAssignOp::create(builder, reg.reg.getLoc(), target,
+                             reg.asyncResetValue);
+    }
+    builder.setInsertionPointAfter(ifOp);
```
seldridge: Same as above. Consider `InsertionGuard`. Suggestion only, non-blocking.

lib/Conversion/SeqToSV/FirRegLowering.cpp
```diff
@@ -373,11 +388,19 @@ void FirRegLowering::createAsyncResetInitialization(
...
     //  end
-    sv::IfOp::create(builder, reset.first, [&] {
-      for (auto &reg : reset.second)
-        sv::BPAssignOp::create(builder, reg.reg.getLoc(), reg.reg,
-                               reg.asyncResetValue);
-    });
+    auto ifOp = sv::IfOp::create(builder, reset.first);
+    auto *thenBlock = ifOp.getThenBlock();
+    for (auto &reg : reset.second) {
+      builder.setInsertionPointToEnd(thenBlock);
```
seldridge: Can this line be hoisted out of the `for` loop given that this will, I think, always be inserting at the end?

test/Dialect/Seq/firreg.mlir
```diff
@@ -1011,3 +1011,61 @@ hw.module @RegUnderIfdefDupName(in %clock : !seq.clock, in %reset : i1, in %valu
...
+  }
+
+  // CHECK: sv.initial {
+  // CHECK:   sv.ifdef.procedural  @MyMacro {
+  // CHECK:     %0 = sv.xmr.ref @PresetRegUnderIfdef_reg : !hw.inout<i1>
+  // CHECK:     sv.bpassign %0, %false : i1
+  // CHECK:   }
+  // CHECK: }
+  hw.output
+}
```
seldridge, line 1071: Very readable test structure!

test/Dialect/Seq/firreg.mlir
```diff
@@ -1011,3 +1011,61 @@ hw.module @RegUnderIfdefDupName(in %clock : !seq.clock, in %reset : i1, in %valu
...
+  // CHECK: sv.ifdef  @MyMacro {
+  // CHECK: %reg = sv.reg sym @reg : !hw.inout<i1> 
+  // CHECK: %0 = sv.read_inout %reg : !hw.inout<i1>
+  // CHECK: sv.always posedge %clock, posedge %reset {
+  // CHECK:   sv.if %reset {
+  // CHECK:     sv.passign %reg, %false_0 : i1
+  // CHECK:   } else {
+  // CHECK:     sv.passign %reg, %value : i1
+  // CHECK:   }
+  // CHECK: }
```
seldridge: Nit: one more level of indentation for the `sv.ifdef` body may further improve readability.

test/Dialect/Seq/firreg.mlir
```diff
@@ -1011,3 +1011,61 @@ hw.module @RegUnderIfdefDupName(in %clock : !seq.clock, in %reset : i1, in %valu
   }
   hw.output
 }
+
+// Test for registers with async resets.
+
+// CHECK:  hw.hierpath @AsyncResetRegUnderIfdef_reg [@AsyncResetRegUnderIfdef::@reg]
```
seldridge: Nit: using captures/variables for the symbols would help if this ever changes in the future.

PR #8949, opened 2025-09-12
[Arc] Add ExecuteOp

fabianschuiki:
Add a new `arc.execute` op which is similar to `scf.execute_region`. It allows an SSACFG region to be embedded in a graph region like an `hw.module`. In a later PR, a conversion from `llhd.combinational` to this new `arc.execute` will be added. We can't directly use `scf.execute_region` for this since that op eagerly inlines into the parent region even if the parent is a graph region. This does not preserve semantics in the case of HW modules.

### DIFF
--- a/include/circt/Dialect/Arc/ArcOps.td
+++ b/include/circt/Dialect/Arc/ArcOps.td
@@ -113,7 +113,7 @@ def DefineOp : ArcOp<"define", [
 
 def OutputOp : ArcOp<"output", [
   Terminator,
-  ParentOneOf<["DefineOp", "LutOp", "ClockDomainOp"]>,
+  ParentOneOf<["DefineOp", "LutOp", "ClockDomainOp", "ExecuteOp"]>,
   Pure,
   ReturnLike
 ]> {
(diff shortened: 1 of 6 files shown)


include/circt/Dialect/Arc/ArcOps.td
```diff
@@ -599,6 +599,35 @@ def StateWriteOp : ArcOp<"state_write", [
...
+//===----------------------------------------------------------------------===//
+
+def ExecuteOp : ArcOp<"execute", [
+  IsolatedFromAbove,
+  RecursiveMemoryEffects,
+]> {
+  let summary = "Execute an SSACFG region";
+  let description = [{
+    The `arc.execute` op allows an SSACFG region to be embedded in a parent
+    graph region, or another SSACFG region. Whenever execution reaches this op,
```
Moxinilian, line 613: Could you explain what execution reaching the op means in this context? Am I correct in understanding that it means "when the value of the result is needed" in an intentionally non-specific way?

fabianschuiki: That's exactly it 😁. I think the deal with `hw.module` and graph regions in Arcilator is that all side-effecting operations are always executed (unless the compiler can prove that not executing them has no observable effect), and pure operations are executed opportunistically when the results are needed. Although that last part is kind of redundant; maybe we could just say that all ops are always executed, unless the compiler can prove that skipping them has no observable effect. Or something about the side effects of operations in the module take effect in the order they appear in the module.

PR #8950, opened 2025-09-12
[ConvertToArcs] Add llhd.combinational conversion

fabianschuiki:
Introduce a dialect conversion step into the `ConvertToArcs` pass. We'll use this to map various core dialect operations to Arc-specific ones in the future. As a starting point, add a conversion from `llhd.combinational` to `arc.execute`. This, alongside lowering to LLVM to be added in later PRs, will implement some basic LLHD support in the Arc dialect.

### DIFF
--- a/lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
+++ b/lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
@@ -9,11 +9,14 @@
 #include "circt/Conversion/ConvertToArcs.h"
 #include "circt/Dialect/Arc/ArcOps.h"
 #include "circt/Dialect/HW/HWOps.h"
+#include "circt/Dialect/LLHD/IR/LLHDOps.h"
 #include "circt/Dialect/Seq/SeqOps.h"
 #include "circt/Dialect/Sim/SimOps.h"
+#include "circt/Support/ConversionPatternSet.h"
 #include "circt/Support/Namespace.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/RegionUtils.h"
 #include "llvm/Support/Debug.h"
 
@@ -24,6 +27,7 @@ using namespace arc;
 using namespace hw;
 using llvm::MapVector;
 using llvm::SmallSetVector;
+using mlir::ConversionConfig;
 
 static bool isArcBreakingOp(Operation *op) {
   if (isa<TapOp>(op))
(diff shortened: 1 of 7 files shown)


lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
```diff
@@ -507,29 +511,95 @@ LogicalResult Converter::absorbRegs(HWModuleOp module) {
...
+  // Gather the conversion patterns.
+  ConversionPatternSet patterns(&getContext(), converter);
+  patterns.add<llhd::CombinationalOp>(convert);
+  patterns.add<llhd::YieldOp>(convert);
+
+  // Setup the legal ops. (Sort alphabetically.)
+  ConversionTarget target(getContext());
+  target.addIllegalDialect<llhd::LLHDDialect>();
+  target.markUnknownOpDynamicallyLegal(
+      [](Operation *op) { return !isa<llhd::LLHDDialect>(op->getDialect()); });
```
uenoku, line 587: Is it required even when llhd::LHDDialect is registered as an illegal dialect?

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
```diff
@@ -507,29 +511,95 @@ LogicalResult Converter::absorbRegs(HWModuleOp module) {
...
+      return std::nullopt;
+    return type;
+  });
+
+  // Gather the conversion patterns.
+  ConversionPatternSet patterns(&getContext(), converter);
+  patterns.add<llhd::CombinationalOp>(convert);
+  patterns.add<llhd::YieldOp>(convert);
+
+  // Setup the legal ops. (Sort alphabetically.)
```
TaoBi22, line 583: What does the 'sort alphabetically' here mean?

fabianschuiki: Just a hint to keep the list of patterns sorted alphabetically :sweat_smile:

lib/Conversion/ConvertToArcs/ConvertToArcs.cpp
fabianschuiki: Yeah I ran into the issue that all operations/dialects not explicitly listed as legal leading to a "failed to legalize" error. I would have thought that the partial dialect conversion would just rewrite ops that were explicitly marked as illegal. But that didn't work for me :thinking:

uenoku: Hmm, interesting! Thank you for trying it out!

jpienaar: One of the reasons I often just use greedy pattern rewriter or simple walk, as I always have to remember again and mostly I want any pattern that can apply to apply. (dialect conversion still very useful for ones where type conversion needed and I haven't tried the new oneshot one).

fabianschuiki: Yeah same here. I have _no_ mental model of the dialect conversion framework. It feels completely random to me what I have to put into legalization, type conversion, and what gets converted where. But the plain old rewriting framework is annoying when a lot of type conversions are involved. I'm also having trouble finding good documentation for the dialect conversion framework. Seems to be a RTFC kind of deal :grin:

PR #8951, opened 2025-09-12
[ArcToLLVM] Add arc.execute conversion

fabianschuiki:
Add a conversion from `arc.execute` to the LLVM dialect. The conversion mimics `scf.execute_region`, but tries to convert types and block signatures along the way. (No idea why SCF does not do this.) This allows `arc.execute` ops to be lowered to LLVM by inlining the control flow graph into the parent SSACFG region, most likely an `llvm.func`. This now allows Arcilator to simulate designs that contain `llhd.combinational` ops.

### DIFF
--- a/test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
+++ b/test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
@@ -243,3 +243,38 @@ func.func @DontCrashOnI0(%arg0: i1, %arg1: !hw.array<1xi42>) -> i42 {
   %1 = hw.array_get %arg1[%0] : !hw.array<1xi42>, i0
   return %1 : i42
 }
+
+// CHECK-LABEL: llvm.func @ExecuteEmpty
+func.func @ExecuteEmpty() {
+  // CHECK-NEXT: llvm.br [[BB:\^.+]]
+  // CHECK-NEXT: [[BB]]:
+  arc.execute {
+    // CHECK-NEXT: llvm.br [[BB:\^.+]]
+    arc.output
+  }
+  // CHECK-NEXT: [[BB]]:
+  // CHECK-NEXT: llvm.return
+  return
+}
+
+// CHECK-LABEL: llvm.func @ExecuteWithOperandsAndResults
+func.func @ExecuteWithOperandsAndResults(%arg0: i42, %arg1: !hw.array<4xi19>, %arg2: !arc.storage) {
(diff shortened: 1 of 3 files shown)


test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
```diff
@@ -243,3 +243,34 @@ func.func @DontCrashOnI0(%arg0: i1, %arg1: !hw.array<1xi42>) -> i42 {
...
+    cf.br ^bb1(%2 : !arc.storage)
+  ^bb1(%3: !arc.storage):
+    // CHECK-NEXT: [[BB]]([[ARG2:%.+]]: !llvm.ptr):
+    // CHECK-NEXT: llvm.br [[BB:\^.+]]([[ARG0]], [[ARG1]], [[ARG2]] : i42, !llvm.array<4 x i19>, !llvm.ptr)
+    arc.output %0, %1, %3 : i42, !hw.array<4xi19>, !arc.storage
+  }
+  // CHECK-NEXT: [[BB]]([[ARG0:%.+]]: i42, [[ARG1:%.+]]: !llvm.array<4 x i19>, [[ARG2:%.+]]: !llvm.ptr):
+  // CHECK-NEXT: llvm.return
+  return
+}
```
maerhart, line 278: There's no test where `arc.execute` has results.

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
```diff
@@ -243,3 +243,34 @@ func.func @DontCrashOnI0(%arg0: i1, %arg1: !hw.array<1xi42>) -> i42 {
...
 }
+
+// CHECK-LABEL: llvm.func @ExecuteEmpty
+func.func @ExecuteEmpty() {
+  // CHECK-NEXT: llvm.br [[BB:\^.+]]
+  // CHECK-NEXT: [[BB]]:
+  arc.execute {
+    // CHECK-NEXT: llvm.br [[BB:\^.+]]
+    arc.output
+  }
```
maerhart, line 254: Couldn't we have a folder for this case that just removes it? In that case, this test would become invalid.

fabianschuiki: Good point! I'll add that folder in a follow-up PR and get rid of this test :slightly_smiling_face:

test/Conversion/ArcToLLVM/lower-arc-to-llvm.mlir
fabianschuiki: :+1: Good point. I added a use for the results here.

PR #8953, opened 2025-09-12
[ESI] Librarify ESI cosim classes

mortbopet:


teqdruid:
@mortbopet plz review my changes and merge.

PR #8957, opened 2025-09-13
[Reduce] Various reducer improvements

fabianschuiki:
A bunch of improvements and new reduction patterns, mainly for FIRRTL. Mostly vibe-coded with Claude Sonnet 4. Works pretty well for this kind of thing!


lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -1118,6 +1168,325 @@ struct ModuleNameSanitizer : OpReduction<firrtl::CircuitOp> {
...
+
+  /// Create a vector of port type-direction pairs for the given FIRRTL module.
+  /// This ignores port names, allowing modules with the same port types and
+  /// directions but different port names to be considered equivalent for
+  /// swapping.
+  PortSignature getModulePortSignature(FModuleLike module) {
+    PortSignature signature;
+    for (unsigned i = 0, e = module.getNumPorts(); i < e; ++i) {
+      signature.emplace_back(module.getPortType(i), module.getPortDirection(i));
+    }
```
uenoku:
```suggestion
    for (unsigned i = 0, e = module.getNumPorts(); i < e; ++i)
      signature.emplace_back(module.getPortType(i), module.getPortDirection(i));
 
```

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -1118,6 +1168,325 @@ struct ModuleNameSanitizer : OpReduction<firrtl::CircuitOp> {
...
+    });
+  }
+  void afterReduction(mlir::ModuleOp op) override { nlaRemover.remove(op); }
+
+  /// Create a vector of port type-direction pairs for the given FIRRTL module.
+  /// This ignores port names, allowing modules with the same port types and
+  /// directions but different port names to be considered equivalent for
+  /// swapping.
+  PortSignature getModulePortSignature(FModuleLike module) {
+    PortSignature signature;
```
uenoku, line 1209: nit: `signature.reserve(..)`

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -1118,6 +1168,325 @@ struct ModuleNameSanitizer : OpReduction<firrtl::CircuitOp> {
...
+      state.moduleTypeGroups[signature].push_back(module);
+    }
+
+    // For each group, find the smallest module
+    for (auto &[signature, modules] : state.moduleTypeGroups) {
+      if (modules.size() <= 1)
+        continue;
+
+      FModuleLike smallestModule = nullptr;
+      uint64_t smallestSize = UINT64_MAX;
```
uenoku: nit: please use std::numeric_limits<uint64_t>::max()

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -1118,6 +1168,325 @@ struct ModuleNameSanitizer : OpReduction<firrtl::CircuitOp> {
...
+  ::detail::SymbolCache symbols;
+  NLARemover nlaRemover;
+  ModuleSizeCache moduleSizes;
+
+  // Per-circuit state containing all module swapping information
+  DenseMap<CircuitOp, CircuitState> circuitStates;
+};
+
+/// A reduction pattern that handles MustDedup annotations by replacing all
+/// module names in a dedup group with a single module name. This helps reduce
```
uenoku, line 1308: This is super smart 👍

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -1118,6 +1168,325 @@ struct ModuleNameSanitizer : OpReduction<firrtl::CircuitOp> {
...
+  };
+
+  void beforeReduction(mlir::ModuleOp op) override {
+    symbols.clear();
+    nlaRemover.clear();
+    moduleSizes.clear();
+    circuitStates.clear();
+
+    // Collect module type groups and NLA tables for all circuits up front
+    op.walk([&](CircuitOp circuitOp) {
```
uenoku: This walks entire IR so it would be better to make this preorder and skip the walk for circuit op.

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -617,38 +622,79 @@ struct ConnectInvalidator : public Reduction {
...
+    // Convert matches to a set for fast lookup
+    llvm::SmallDenseSet<uint64_t, 4> matchesSet(matches.begin(), matches.end());
+
+    // Lambda to process annotations and filter out matched ones
+    uint64_t matchId = 0;
+    auto processAnnotations =
+        [&](ArrayRef<Attribute> annotations) -> ArrayAttr {
+      SmallVector<Attribute> newAnnotations;
+      for (auto anno : annotations) {
+        if (!matchesSet.contains(matchId)) {
```
uenoku, line 660: Maybe I'm missing something but isn't every annotation marked as matched in the current `matches`?

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -617,38 +622,79 @@ struct ConnectInvalidator : public Reduction {
...
+  void matches(Operation *op,
+               llvm::function_ref<void(uint64_t, uint64_t)> addMatch) override {
+    uint64_t matchId = 0;
+
+    // Generate matches for regular annotations
+    if (auto annos = op->getAttrOfType<ArrayAttr>("annotations")) {
+      for (auto anno : annos) {
+        (void)anno;
+        addMatch(1, matchId++);
+      }
```
seldridge: This is kind of weird. Maybe:


```suggestion
      for (auto _ : annos)
        addMatch(1, matchId++);
```

Ditto throughout.

fabianschuiki: Good idea :+1:

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
fabianschuiki: Good point! I made this a pre-order walk that skips the circuit.

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
fabianschuiki: Nope you're correct: all annotations are marked as an individual match in `matches`. The reducer then selects some of those matches and passes them to `rewrite`. Usually the reducer will first try to pass all matches to `rewrite` to see if it can remove all annotations in this case. If that doesn't work, it will start to perform a binary search over the matches and pass subsets of them to `rewrite`, to find annotations that can be deleted.

PR #8958, opened 2025-09-13
[AIG] feat : and_inv fold

### LINKED ISSUE #8905: [AIG] Add a missing folder
The following patterns are currently implemented as canonicalizaton but it would be to add as an folder. 
```
aig.and_inv(x, 0) -> 0
aig.and_inv(x, 1) -> x
```

markram1729:
Resolves #8905

### DIFF
--- a/lib/Dialect/Synth/SynthOps.cpp
+++ b/lib/Dialect/Synth/SynthOps.cpp
@@ -159,24 +159,22 @@ OpFoldResult AndInverterOp::fold(FoldAdaptor adaptor) {
   if (getNumOperands() == 1 && !isInverted(0))
     return getOperand(0);
 
-  auto inputs  = adaptor.getInputs();
-  if(inputs.size()==2 && inputs[1])
-   {
-	  auto value = cast<IntegerAttr>(inputs[1]).getValue();
-	  if(value.isZero())
-	   {
-		  if(isInverted(1)) return getOperand(0);
-		  return getIntAttr(value,getContext());
-	   }
-	  if(value.isAllOnes())
-	   {	
+  auto inputs = adaptor.getInputs();
+  if (inputs.size() == 2 && inputs[1]) {
+    auto value = cast<IntegerAttr>(inputs[1]).getValue();
+    if (value.isZero()) {
+      if (isInverted(1))
+        return getOperand(0);
+      return getIntAttr(value, getContext());
+    }
+    if (value.isAllOnes()) {
 
-		  if(isInverted(1)) 
-			  value = 0;
-			  return getIntAttr(value,getContext());
-		  return getOperand(0);
-	   }
-   }
+      if (isInverted(1))
+        value = 0;
+      return getIntAttr(value, getContext());
+      return getOperand(0);
+    }
+  }
   return {};
 }
 
--- a/lib/Dialect/Synth/SynthOps.cpp
+++ b/lib/Dialect/Synth/SynthOps.cpp
@@ -7,7 +7,6 @@
 //===----------------------------------------------------------------------===//
 
 #include "circt/Dialect/Synth/SynthOps.h"
-#include "circt/Dialect/HW/HWAttributes.h"
 #include "circt/Dialect/HW/HWOps.h"
 #include "circt/Support/CustomDirectiveImpl.h"
 #include "circt/Support/Naming.h"
@@ -164,9 +163,9 @@ OpFoldResult AndInverterOp::fold(FoldAdaptor adaptor) {
   if(inputs.size()==2 && inputs[1])
    {
 	  auto value = cast<IntegerAttr>(inputs[1]).getValue();
-	  if(value.isZero())
+	  if(value.isZero()&& !(isInverted(1)))
 		  return getIntAttr(value,getContext());
-	  if(value.isOne())
+	  if(value.isAllOnes())
 		  return getOperand(0);
    }
   return {};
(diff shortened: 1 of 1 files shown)


lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "circt/Dialect/Synth/SynthOps.h"
+#include "circt/Dialect/HW/HWAttributes.h"
```
uenoku: nit: you probably don't need this.

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -31,9 +32,24 @@ LogicalResult MajorityInverterOp::verify() {
...
 OpFoldResult AndInverterOp::fold(FoldAdaptor adaptor) {
   if (getNumOperands() == 1 && !isInverted(0))
     return getOperand(0);
+
+  auto inputs  = adaptor.getInputs();
+  if(inputs.size()==2 && inputs[1])
+   {
+	  auto value = cast<IntegerAttr>(inputs[1]).getValue();
+	  if(value.isZero())
+		  return getIntAttr(value,getContext());
```
uenoku: and_inverter could have inverted flag so it's necessary to check the flag. 
```suggestion
	  if(value.isZero() && !isInverted(1))
		  return getIntAttr(value,getContext());
```

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -31,9 +32,24 @@ LogicalResult MajorityInverterOp::verify() {
...
+	  if(value.isOne())
```
uenoku: Value might be multibit so could you check with value.isAllOnes()?

uenoku (commented):
Thank you for sending a PR! Could you insert checks for inversion flags?

markram1729:
Hi @uenoku , Just small clarification 
Should i handle these as well 
```
 aig.and_inv  x 0 -> 0
 aig.and_inv  x ~0 -> x
 aig.and_inv x 1 ->x 
 aig.and_inv  x ~1 -> 0
 ```

File: lib/Dialect/Synth/SynthOps.cpp (hunk 2 above)
Comment (markram1729) [reply lib/Dialect/Synth/SynthOps.cpp]: Done

File: lib/Dialect/Synth/SynthOps.cpp (hunk 3 above)
Comment (markram1729) [reply lib/Dialect/Synth/SynthOps.cpp]: True, added change now

**Comment by uenoku** [comment]
Yes, that would be great! Also can we add tests in test/Dialect/Synth/canonicalizer.mlir?

**Comment by markram1729** [comment]
> Yes, that would be great! Also can we add tests in test/Dialect/Synth/canonicalizer.mlir?

Sure , I will add the Necessary tests

**Comment by markram1729** [comment]
Hi @uenoku  I see , there are already test that is covering all 4 conditions  at https://github.com/llvm/circt/blob/e7baaf683039b5fd48f56a4260878890dacf118e/test/Dialect/Synth/canonicalizer.mlir#L4

File: lib/Dialect/Synth/SynthOps.cpp [hunk 4]
```diff
@@ -150,9 +150,31 @@ LogicalResult MajorityInverterOp::canonicalize(MajorityInverterOp op,
...
   if (getNumOperands() == 1 && !isInverted(0))
     return getOperand(0);
+
+  auto inputs = adaptor.getInputs();
+  if (inputs.size() == 2 && inputs[1]) {
+    auto value = cast<IntegerAttr>(inputs[1]).getValue();
+    if (value.isZero()) {
+      if (isInverted(1))
+        return getOperand(0);
+      return getIntAttr(value, getContext());
```
uenoku: Could you insert a check for operand 0 as well?

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -150,9 +150,31 @@ LogicalResult MajorityInverterOp::canonicalize(MajorityInverterOp op,
...
+    }
+    if (value.isAllOnes()) {
+
+      if (isInverted(1))
+        value = 0;
+      return getIntAttr(value, getContext());
+      return getOperand(0);
+    }
```
uenoku: Maybe inverting value first makes the control flow a bit simpler. 
```suggestion
    auto value = cast<IntegerAttr>(inputs[1]).getValue();
    if(isInverted(1))
       value = ~value;
    if (value.isZero()) 
        return IntergertAttr::get(context, value);
    if (value.isAllOnes()) {
        if (isInverted(0))
           return {}; // Failure because we cannot remove operand in folder 
        return  getOperand(0);
    }
```

markram1729: True, Thank you for the suggestion 
I will change it

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -150,9 +150,29 @@ LogicalResult MajorityInverterOp::canonicalize(MajorityInverterOp op,
 // AIG Operations
 //===----------------------------------------------------------------------===//
 
+static TypedAttr getIntAttr(const APInt &value, MLIRContext *context) {
```
uenoku: nit: can you inline this function since there is only one user?

markram1729: sure!

markram1729: Hi @uenoku ,added the inline

uenoku:
Thank you for the PR!

PR #8959, opened 2025-09-13
[Synth] Implement basic canonicalization/folder to mig.maj_inv

uenoku:
Add canonicalize method with pattern matching for single operands, identical/complementary operands, and constants.
Also implement fold method for constant evaluation along with constant materialization


fabianschuiki (approved):
Nice, LGTM!

PR #8963, opened 2025-09-15
[ESI] Bump zlib tag

mortbopet:
The latest zlib release (v1.3.1) is fairly old (Jan 22. 2024) and is missing some key changes/fixes for building w/ a static library. This seems to be somewhat system/cmake version dependent. In any case, solved by bumping to head of their development branch (still a 7 month old commit) which cleans up their CMake build system, and introduces an alias target for the static library.


teqdruid:
I'm running a test build here: https://github.com/llvm/circt/actions/runs/17750296543 . If it passes, you're good to go.

teqdruid (approved):
Please wait for this to go green: https://github.com/llvm/circt/actions/runs/17750296543

PR #8964, opened 2025-09-15
[ESI] Factor out inner execution in Simulator::run

mortbopet:
When using esi-cosim as a library, it is very inconvenient that the default behavior of the `run` command assumes that `esi-cosim` is executed as a standalone script (i.e. either tries to run a user-provided command or waits for user input). To better serve as a library, factor out the inner simulator process execution to `run_proc` s.t. other python code can call this, and get a handle to the simulator process and port.

### DIFF
--- a/lib/Dialect/ESI/runtime/cosim_dpi_server/esi-cosim.py
+++ b/lib/Dialect/ESI/runtime/cosim_dpi_server/esi-cosim.py
@@ -82,6 +82,24 @@ class SourceFiles:
     return self.dpi_sv + self.user
 
 
+class SimProcess:
+
+  def __init__(self, proc: subprocess.Popen, port: int):
+    self.proc = proc
+    self.port = port
+
+  def force_stop(self):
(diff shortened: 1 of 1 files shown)


lib/Dialect/ESI/runtime/cosim_dpi_server/esi-cosim.py
```diff
@@ -146,60 +205,7 @@ def run(self,
...
-      checkCount = 0
-      while not is_port_open(port):
-        checkCount += 1
-        if checkCount > 200:
-          raise Exception(f"Cosim RPC port ({port}) never opened")
-        if simProc.poll() is not None:
-          raise Exception("Simulation exited early")
-        time.sleep(0.05)
-
+      simProc, port = self.run_proc(gui=gui)
```
teqdruid: In the `finally:` block below, it'd be neat if the `simProc` returned an object which did all the stuff in the finally block. Perhaps something we could run in a `with` block.

At the very least, the `finally:` block should be factored out as well.

PR #8968, opened 2025-09-15
[Moore] Add builtin for $urandom

Scheremo:
First time I try to add a new Op, so let me know if I went wrong somewhere! 

This PR adds a new builtin in moore for the $urandom system call.
Per the newest version of the spec, $urandom may be called with or with a integer seed argument, and always returns an integer produced by a pseudo-random number generator.

As such I added an importer conversion and tests for both cases; the allowed expressions are 
```
moore.builtin.urandom
moore.builtin.urandom %arg : !moore.i32
```

I struggled a bit with constraining the output type of `moore.builtin.urandom`, since it is constant. I added
```
def MooreI32 :
  Type<CPred<
    "(::llvm::isa<::circt::moore::IntType>($_self) && "
    "::llvm::cast<::circt::moore::IntType>($_self).getWidth() == 32 && "
    "::llvm::cast<::circt::moore::IntType>($_self).getDomain() == "
    "::circt::moore::Domain::TwoValued)"
  >, "!moore.i32"> {
  let builderCall =
    "::circt::moore::IntType::getInt($_builder.getContext(), 32)";
}
```
in place to constrain an IntType to be exactly a moore.i32 for this purpose, but please let me know if there is a smarter way; this seems a bit cumbersome.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -1742,6 +1742,38 @@ class Builtin<string mnemonic, list<Trait> traits = []> :
     MooreOp<"builtin." # mnemonic, traits>;
 
 //===----------------------------------------------------------------------===//
+// Pseudo-Random Generators
+//===----------------------------------------------------------------------===//
+
+// Buildable alias for exactly !moore.i32
+def MooreI32 :
+  Type<CPred<
+    "(::llvm::isa<::circt::moore::IntType>($_self) && "
+    "::llvm::cast<::circt::moore::IntType>($_self).getWidth() == 32 && "
+    "::llvm::cast<::circt::moore::IntType>($_self).getDomain() == "
+    "::circt::moore::Domain::TwoValued)"
+  >, "!moore.i32"> {
+  let builderCall =
+    "::circt::moore::IntType::getInt($_builder.getContext(), 32)";
+}
+
+def UrandomBIOp : Builtin<"urandom"> {
+  let summary = "Generate a pseudo-random unsigned integer (optionally seeded)";
+  let description = [{
+    Corresponds to the `$urandom` system function. Returns a 32-bit
+    pseudo-random integer. The seed is optional; when provided, it initializes
+    the generator. If not provided, treat it as 0 in semantics/lowering.
+
+    See IEEE 1800-2023 § 18.13 "Random number system functions and methods".
+  }];
+
+  // Optional positional attribute for the seed
+  let arguments = (ins Optional<IntType>:$seed);
+  let results = (outs MooreI32:$result);
+  let assemblyFormat = "($seed^ `:` type($seed))? attr-dict";
+}
+
+//===----------------------------------------------------------------------===//
 // Simulation Control Builtins
 //===----------------------------------------------------------------------===//
 
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -1745,18 +1745,6 @@ class Builtin<string mnemonic, list<Trait> traits = []> :
 // Pseudo-Random Generators
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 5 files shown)


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1741,6 +1741,38 @@ def FormatRealOp : MooreOp<"fmt.real", [Pure]> {
...
+def MooreI32 :
+  Type<CPred<
+    "(::llvm::isa<::circt::moore::IntType>($_self) && "
+    "::llvm::cast<::circt::moore::IntType>($_self).getWidth() == 32 && "
+    "::llvm::cast<::circt::moore::IntType>($_self).getDomain() == "
+    "::circt::moore::Domain::TwoValued)"
+  >, "!moore.i32"> {
+  let builderCall =
+    "::circt::moore::IntType::getInt($_builder.getContext(), 32)";
+}
```
fabianschuiki: Unfortunately this is roughly what you have to do. :disappointed:. We have a few of those in MooreTypes.td, for example `FourValuedI64`: https://github.com/llvm/circt/blob/main/include/circt/Dialect/Moore/MooreTypes.td#L479-L484. You could create a corresponding `TwoValuedI32` there for consistency and use that for your result.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1741,6 +1741,38 @@ def FormatRealOp : MooreOp<"fmt.real", [Pure]> {
...
+
+def UrandomBIOp : Builtin<"urandom"> {
+  let summary = "Generate a pseudo-random unsigned integer (optionally seeded)";
+  let description = [{
+    Corresponds to the `$urandom` system function. Returns a 32-bit
+    pseudo-random integer. The seed is optional; when provided, it initializes
+    the generator. If not provided, treat it as 0 in semantics/lowering.
+
+    See IEEE 1800-2023 § 18.13 "Random number system functions and methods".
```
fabianschuiki, line 1755: Nice! Thanks for adding descriptions and a reference into the standard. Very valuable.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1741,6 +1741,38 @@ def FormatRealOp : MooreOp<"fmt.real", [Pure]> {
...
+  }];
+
+  // Optional positional attribute for the seed
+  let arguments = (ins Optional<IntType>:$seed);
```
fabianschuiki: I think the seed is also an `int` in SV, so you could use a `TwoValuedI32` here as well.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -984,17 +984,29 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+      break;
+
+    default:
+      break;
     }
 
+    if (failed(result))
+      return {};
+    if (*result)
+      return *result;
```
fabianschuiki, line 1009: Nice!

include/circt/Dialect/Moore/MooreOps.td
Scheremo: Done! Thanks for the pointer 👍

include/circt/Dialect/Moore/MooreOps.td
Scheremo: Changed to use new `TwoValuedI32`

include/circt/Dialect/Moore/MooreOps.td
Scheremo: Took a leaf out of the book of whoever documented the other system calls 😆

Scheremo:
Move the new Type to `MooreTypes.td` and renamed to `TwoValuedI32` and used it to construct the `seed` optional argument. Thanks for the quick and very helpful feedback! 
If there is no other comments, I am okay with merging whenever you get to it.

PR #8969, opened 2025-09-15
[FIRRTL] Add reduction that moves MustDedup onto children

fabianschuiki:
Add the `MustDedupChildren` reduction which takes must dedup annotations and replaces them with must dedup annotations on child modules. More Claude Sonnet 4 magic!

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLReductions.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLReductions.cpp
@@ -9,6 +9,7 @@
 #include "circt/Dialect/FIRRTL/FIRRTLReductions.h"
 #include "circt/Dialect/FIRRTL/AnnotationDetails.h"
 #include "circt/Dialect/FIRRTL/CHIRRTLDialect.h"
+#include "circt/Dialect/FIRRTL/FIRRTLAnnotationHelper.h"
 #include "circt/Dialect/FIRRTL/FIRRTLAnnotations.h"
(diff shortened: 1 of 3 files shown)


lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -1478,6 +1480,185 @@ struct ForceDedup : public OpReduction<CircuitOp> {
...
+    SmallVector<FModuleLike> modules;
+    for (auto moduleRef : modulesAttr.getAsRange<StringAttr>()) {
+      // Parse "~CircuitName|ModuleName" format.
+      auto refStr = moduleRef.getValue();
+      auto pipePos = refStr.find('|');
+      if (pipePos != StringRef::npos && pipePos + 1 < refStr.size()) {
+        auto modName = StringAttr::get(context, refStr.substr(pipePos + 1));
+        if (auto mod = symbolTable.lookup<FModuleLike>(modName))
+          modules.push_back(mod);
+      }
```
uenoku: Could you use TokenAnnoTarget? https://github.com/llvm/circt/blob/a2d030176fd7e119a4ca0801b6fcdb8b502fea82/lib/Dialect/FIRRTL/FIRRTLAnnotationHelper.cpp#L291

fabianschuiki: Great idea, thanks! I forgot about `TokenAnnoTarget` 😍. That also works for creating the annotation string again later!

PR #8970, opened 2025-09-16
[Moore] Add builtins for simulation time measurements

Scheremo:
This PR adds builtins for `$time`, `$stime` and `$realtime` in the `moore` Dialect.
All these builtins correspond to System Verilog system tasks and only slightly differ in their functionality.

To represent the return value of `$time`, this PR adds `TwoValueI64`, which represents a 64-Bit integer used to express the simulation time value in.

EDIT: Just spotted there is a `TimeType` in the MooreTypes as well - maybe that would be a better output target? 
An "added complication" in this matter is that `TimeType` currently does not enforce the spec's rounding / truncation / overflow semantics as far as I can see.
I'm thinking it might make sense to first introduce support for `$timeformat`, maybe by adding a `TimeFormatResource`?
Then these system calls can be resolved by checking the `TimeFormatResource` with `timetype_to_time`, `timetype_to_stime`, `timetype_to_real` helper ops, which can implement the exact rounding, truncations & overflow semantics.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -1762,7 +1762,7 @@ def UrandomBIOp : Builtin<"urandom"> {
 }
 
 //===----------------------------------------------------------------------===//
-// Time Formatting And Time Stamping
+// Simulation Time Measurement Builtins
 //===----------------------------------------------------------------------===//
 
 def TimeBIOp : Builtin<"time"> {
@@ -1776,10 +1776,9 @@ def TimeBIOp : Builtin<"time"> {
   }];
 
   // Optional positional attribute for the seed
-  let results = (outs TwoValuedI64:$result);
+  let results = (outs TimeType:$result);
   let assemblyFormat = "attr-dict";
 }
-
 
 //===----------------------------------------------------------------------===//
 // Simulation Control Builtins
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -1762,6 +1762,26 @@ def UrandomBIOp : Builtin<"urandom"> {
 }
 
 //===----------------------------------------------------------------------===//
+// Time Formatting And Time Stamping
+//===----------------------------------------------------------------------===//
+
+def TimeBIOp : Builtin<"time"> {
+  let summary = "Return the current simulation time";
+  let description = [{
+    Corresponds to the `$time` system function. Returns a int-rounded 64-bit
+    integer corresponding to the elapsed simulation time scaled by the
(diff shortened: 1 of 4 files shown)


Scheremo:
Glad I'm looking at useful things! :) 
What you suggest definitely makes sense and is very likely the right way to do it.  Probably I got ahead of myself with thinking about what happens if time scales get redefined with something like `$timeformat, $timeunit, $timeprecision` calls - is this handled by Slang?

Scheremo:
> Amazing, I've wanted to add these for a long time. Thanks for taking a stab at them 😁!
> 
> You're right about the types here: the Moore dialect has a `TimeType` which we try to use as much as possible to represent time values. This type has an unambiguous unit, working around the SV weirdness of representing time as an `i64` with user-defined scaling 🤦 🤯, per-module, such that you can send time values between modules and have the two modules interpret them as two totally different values.
> 
> At the moment, operations like `moore.constant_time` create a `TimeType` value, postponing the conversion to an integer to when the user actually tries to assign a time to an integer, or pack a time up in a struct field. When a conversion to/from integer is needed, the `materializeSBVToPackedConversion` (and the reverse function) look at the local time scale and perform scaling and conversion between integers and time values.
> 
> For your builtins, I think you can just make them return a `TimeType` directly. The conversion materialization will then take care of applying a timescale and converting to an int/real when the user requests it. You may even want to map `$time`, `$stime`, and `$realtime` to a single `TimeBIOp` returning a `TimeType`. (The time-real conversion and int-real conversion isn't yet implemented, but that's something we should definitely add 😏.)

Changed the implementation to a single `TimeBIOp` which returns a `TimeType` - I'm happy to merge, but please let me know if you spot anything else!

fabianschuiki (approved):
Really nice! LGTM! :rocket:

fabianschuiki:
Minor nit: GitHub auto-populates a pull request with the commit message if there's a single commit. And GitHub then uses that as a message for the squash-and-merge commit when landing the PR. Instead of typing a PR comment, try typing the same as a commit message such that it's tracked in git. That makes the PR wrangling easier :slightly_smiling_face:

fabianschuiki:
Btw, looks like your PRs make the sv-tests pass rate for Moore go up and up :grin: https://chipsalliance.github.io/sv-tests-results/history

Scheremo:
> Minor nit: GitHub auto-populates a pull request with the commit message if there's a single commit. And GitHub then uses that as a message for the squash-and-merge commit when landing the PR. Instead of typing a PR comment, try typing the same as a commit message such that it's tracked in git. That makes the PR wrangling easier 🙂

Please remind me again if this is not done on the next few PRs, but I will try my best to keep it in mind 😄

PR #8974, opened 2025-09-17
[Support] Add walkPostOrder and walkInversePostOrder to InstanceGraph

fabianschuiki:
Add new helper functions to the `InstanceGraph` that implement a full post-order and inverse-post-order walk across _all_ modules. This pulls a common pattern into the instance graph itself, where a user would iterate over the graph's nodes and then use an `llvm::post_order_ext` iterator to not revisit prior nodes with an externally-held set.

The instance graph implements the graph traits, which makes using `llvm::post_order` it tempting. But this is usually not what we want. The graph does not have a unique root node. Instead, the FIRRTL flavor of the graph uses the "top" module as specified by the circuit, and the HW flavor of the graph uses a dummy node that points to all _public_ modules. In most passes we want to iterate over _all_ modules in the IR with a specific order, not just a subset of the modules. These new helper functions make this common use case more obvious and may allow us to remove the graph traits from `InstanceGraph` again in the future.

Most of this PR is just mechanical identation changes. One brittle test for instance choice needs to be updated since that relied on the fact that the dedup pass would ignore uninstantiated modules.

### DIFF
--- a/include/circt/Support/InstanceGraph.h
+++ b/include/circt/Support/InstanceGraph.h
@@ -16,6 +16,7 @@
 #include "circt/Support/LLVM.h"
 #include "mlir/IR/OpDefinition.h"
 #include "llvm/ADT/GraphTraits.h"
+#include "llvm/ADT/PostOrderIterator.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/iterator.h"
 #include "llvm/Support/DOTGraphTraits.h"
@@ -266,6 +267,22 @@ public:
   iterator begin() { return nodes.begin(); }
   iterator end() { return nodes.end(); }
 
+  /// Perform a post-order walk across the modules. Guaranteed to visit all
+  /// modules. Child modules are visited before their parent modules. If `Fn`
+  /// returns a `LogicalResult`, the walk can be interrupted by returning
+  /// `failure()` from the callback, in which case the overall walk will return
+  /// `failure()`.
+  template <typename Fn>
+  decltype(auto) walkPostOrder(Fn &&fn);
+
+  /// Perform an inverse-post-order walk across the modules. Guaranteed to visit
+  /// all modules. Child modules are visited before their parent modules. If
+  /// `Fn` returns a `LogicalResult`, the walk can be interrupted by returning
+  /// `failure()` from the callback, in which case the overall walk will return
+  /// `failure()`.
+  template <typename Fn>
+  decltype(auto) walkInversePostOrder(Fn &&fn);
+
   //===-------------------------------------------------------------------------
   // Methods to keep an InstanceGraph up to date.
   //
@@ -458,6 +475,16 @@ struct llvm::GraphTraits<llvm::Inverse<circt::igraph::InstanceGraphNode *>> {
 };
 
 /// Graph traits for the instance graph.
+///
+/// **Deprecated:** This uses `getTopLevelNode()` as the root node of the graph,
+/// which does not contain all modules in the instance graph. Instead, the
+/// behaviour is dependent on concrete subclasses of the instance graph. The HW
+/// and FIRRTL dialects define this to be variants of "the top module" and "all
+/// public modules", which is usually not what you want in a pass. You are more
+/// likely to want to visit _all_ modules in the IR in post order (children
+/// before parents), not just a subset of the modules depending on whether
+/// things are transitively instantiated. Use `walkPostOrder` or
+/// `walkInversePostOrder` on `InstanceGraph` instead.
 template <>
 struct llvm::GraphTraits<circt::igraph::InstanceGraph *>
     : public llvm::GraphTraits<circt::igraph::InstanceGraphNode *> {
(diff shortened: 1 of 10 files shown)


include/circt/Support/InstanceGraph.h
seldridge, line 1: If convenient, land any of the comment / cleanup stuff directly.

Thanks for doing some gardening of this file.

include/circt/Support/InstanceGraph.h
```diff
@@ -254,6 +267,22 @@ class InstanceGraph {
   iterator begin() { return nodes.begin(); }
   iterator end() { return nodes.end(); }
 
+  /// Perform a post-order walk across the modules. Guaranteed to visit all
+  /// modules. Child modules are visited before their parent modules. If `Fn`
+  /// returns a `LogicalResult`, the walk can be interrupted by returning
+  /// `failure()` from the callback, in which case the overall walk will return
+  /// `failure()`.
+  template <typename Fn>
+  decltype(auto) walkPostOrder(Fn &&fn);
```
seldridge, line 276: The one slightly weird thing about this is the use of `LogicalResult` instead of `WalkResult`. Maybe this could try to pretend it was a real walk with `WalkResult`, but that would have to support the notion of "skipping" which may not make sense here.

I do think this is fine as is.

include/circt/Support/InstanceGraph.h
```diff
@@ -439,7 +474,17 @@ struct llvm::GraphTraits<llvm::Inverse<circt::igraph::InstanceGraphNode *>> {
...
+/// behaviour is dependent on concrete subclasses of the instance graph. The HW
+/// and FIRRTL dialects define this to be variants of "the top module" and "all
+/// public modules", which is usually not what you want in a pass. You are more
+/// likely to want to visit _all_ modules in the IR in post order (children
+/// before parents), not just a subset of the modules depending on whether
+/// things are transitively instantiated. Use `walkPostOrder` or
+/// `walkInversePostOrder` on `InstanceGraph` instead.
 template <>
 struct llvm::GraphTraits<circt::igraph::InstanceGraph *>
     : public llvm::GraphTraits<circt::igraph::InstanceGraphNode *> {
```
seldridge, line 490: Can this use `[[deprecated]]` to give a compilation warning? Is there a plan to remove this?

Does telling the compiler this is deprecated have the intended effect?

include/circt/Support/InstanceGraph.h
seldridge: Is there a way to make the return `[[nodiscard]]`. I've been putting that on almost every return of `LogicalResult` and it's been helpful.

include/circt/Support/InstanceGraph.h
```diff
@@ -481,7 +526,62 @@ struct llvm::DOTGraphTraits<circt::igraph::InstanceGraph *>
...
+        if (failed(fn(*node)))
+          return failure();
+      } else {
+        fn(*node);
+      }
+    }
+  }
+  if constexpr (fallible)
+    return success();
+}
```
seldridge, line 575: These walks look correct to me. It could be good to have somebody else given them a look, too.

lib/Analysis/FIRRTLInstanceInfo.cpp
```diff
@@ -108,65 +108,61 @@ InstanceInfo::InstanceInfo(Operation *op, mlir::AnalysisManager &am) {
...
-      AnnotationSet annotations(moduleOp);
-      auto isDut = annotations.hasAnnotation(dutAnnoClass);
-      auto isGCCompanion = annotations.hasAnnotation(companionAnnoClass);
-
-      if (isDut) {
-        attributes.underDut.markConstant(true);
-        attributes.inDesign.markConstant(true);
-        attributes.inEffectiveDesign.markConstant(true);
+  iGraph.walkInversePostOrder([&](auto &modIt) {
+    auto moduleOp = modIt.getModule();
```
seldridge, line 112: So much nicer. Also saves an indentation level. 💯

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
```diff
@@ -100,6 +100,9 @@ void AssignOutputDirsPass::runOnOperation() {
   bool changed = false;
 
   LLVM_DEBUG(llvm::dbgs() << "Updating modules:\n");
+  // TODO: Use InstanceGraph::walkInversePostOrder instead. That breaks a few
+  // tests though, probably since more modules are assigned to output
+  // directories.
```
seldridge, line 105: Please create a tracking issue.

lib/Dialect/FIRRTL/Transforms/LowerLayers.cpp
```diff
@@ -1062,10 +1062,7 @@ void LowerLayersPass::preprocessModuleLike(CircuitNamespace &ns,
 /// Create the bind file skeleton for each layer, for each module.
 void LowerLayersPass::preprocessModules(CircuitNamespace &ns,
                                         InstanceGraph &ig) {
-  DenseSet<InstanceGraphNode *> visited;
-  for (auto *root : ig)
-    for (auto *node : llvm::post_order_ext(root, visited))
-      preprocessModuleLike(ns, node);
+  ig.walkPostOrder([&](auto &node) { preprocessModuleLike(ns, &node); });
```
seldridge, line 1065: So much better!

include/circt/Support/InstanceGraph.h
fabianschuiki: Done

include/circt/Support/InstanceGraph.h
fabianschuiki: I think the `LogicalResult` itself is marked as `[[nodiscard]]`, which should do the right thing already :+1:

lib/Dialect/FIRRTL/Transforms/AssignOutputDirs.cpp
fabianschuiki: Realized that this now works, too. Got fixed with a tweak to one of the tests I had to do anyway.

include/circt/Support/InstanceGraph.h
fabianschuiki: Yeah that's definitely worth doing. I think I'll do that in a follow-up PR, since we have a few uses in the code base that need fixing. Don't want to add the `deprecated` marker and then have a bunch of warnings in the code base.

dtzSiFive:
Thanks Fabian! 🏅

PR #8975, opened 2025-09-17
[FIRRTL] Handle ClassTypes properly in Dedup

fabianschuiki:
Add special handling for `!firrtl.class` types in dedup. These used to be treated just as regular types, which meant that `!firrtl.class<@A>` and `!firrtl.class<@B>` would be treated as distinct types, even when the pass deduplicates `@A` and `@B` everywhere else. We already have an established pattern to skip name references and tracking them separately. All that's needed applying this to `ClassType`.

This fixes a subtle OFIRRTL-related issue I've encountered in the wild.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -224,9 +224,25 @@ private:
   }
 
   // NOLINTNEXTLINE(misc-no-recursion)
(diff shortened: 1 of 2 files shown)


mikeurbach (approved):
Awesome, makes sense to me, thanks for this. Surprised we didn't see it yet.

uenoku (approved):
LGTM

PR #8977, opened 2025-09-17
[ESI] Add skid buffer to Cosim_Endpoint_ToHost

mortbopet:
It has been observed that some simulators (verilator) does not handle combinational driving from DPI well (resulting in non-determinism wrt. some of the combinational outputs being dropped/replicated). Questa does not show this behavior. A mitigation is to add a skid buffer to decouple the DPI interface from the output interface.


teqdruid:
I was just gonna insert an instantiation of the existing skid buffer in ESIPrimitives.sv. I'll make that change now.

PR #8978, opened 2025-09-17
[ESI] Set '--output-split' for verilator compilation

mortbopet:
... which also enables parallel compilation and better ccache support.

E.g. an internal simple test:
- without split: 8.2s compile time
- with split: 1.3s compile time


teqdruid:
@mortbopet plz review my changes and merge if you approve.

PR #8979, opened 2025-09-17
[ImportVerilog] Add support for %t format specifier, introduce `moore.fmt.time`

Scheremo:
This PR adds support for using the `%t` format specifier in format strings by mapping them to the newly introduced `moore.fmt.time / moore::TimeFormatOp` in the ImportVerilog string formatter.

According to the system verilog spec IEEE 1800-2023 § 20.4.3 "$timeformat", the exact formatting for any time value is dictated by the last `$timeformat` system task invoked within the scope or parent scopes of the format string, or alternatively, by a mix of constant default values and 
> The smallest time precision argument of all the `timescale compiler directives in the source description.

Because of the formatting relying on context (who came up with this anyway? seems like a bad idea), and use of `$timeformat` is permitted in conditional settings, meaning it can't always be determined statically, I decided to not add a format attribute to `moore::FormatTimeOp`. Instead the semantics for implementing `moore.fmt.time` should rely on querying some global state that models the configuration set by the latest `$timeformat` system call within the scope.


lib/Conversion/ImportVerilog/FormatStrings.cpp
```diff
@@ -207,6 +210,36 @@ struct FormatStringParser {
...
+  // Format an integer with the %t specifier according to IEEE 1800-2023
+  // § 20.4.3 "$timeformat"
+  LogicalResult emitTime(const slang::ast::Expression &arg,
+                         const FormatOptions &options) {
+
+    // Only handle `TimeType` values.
+    auto value = context.convertRvalueExpression(
+        arg, moore::TimeType::get(context.getContext()));
+    if (!value)
+      return failure();
```
fabianschuiki, line 222: Nice! I think this will already handle integers properly, too. The conversion will insert casts from integers to the time type 🚀

fabianschuiki (approved):
Very cool! I love the idea of keeping this simple and just work on `TimeType`. That will allow us to handle any `$timeformat` related business later on somewhere else.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1734,6 +1734,29 @@ def FormatRealOp : MooreOp<"fmt.real", [Pure]> {
   }];
 }
 
+
```
fabianschuiki:
```suggestion
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1734,6 +1734,29 @@ def FormatRealOp : MooreOp<"fmt.real", [Pure]> {
...
+  );
+  let results = (outs FormatStringType:$result);
+  let assemblyFormat = [{
+      $value
+      (`,` `width` $width^)?
+      attr-dict
+  }];
+}
+
+
```
fabianschuiki:
```suggestion
```

test/Conversion/ImportVerilog/builtins.sv
```diff
@@ -285,4 +285,4 @@ function TimeBuiltins();
   // CHECK: [[REALTIME:%.+]] = moore.builtin.time
   // TODO: There is no int-to-real conversion yet; change this to dummyB once int-to-real works!
   dummyA($realtime());
-endfunction
+endfunction  // TimeBuiltins
```
fabianschuiki:
```suggestion
endfunction
```

PR #8981, opened 2025-09-17
[FIRRTL] Emit fopen calls to get fd's not mcd's.

dtzSiFive:
Multichannel descriptors are limited to 30, where file descriptors usually have a much higher limit.

Specify opening files "w" which will truncate and is not binary; this may be a change depending on default behavior of mcd's.

### DIFF
--- a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
+++ b/lib/Conversion/FIRRTLToHW/LowerToHW.cpp
@@ -909,7 +909,7 @@ package __circt_lib_logging;
     static int global_id [string];
     static function int get(string name);
       if (global_id.exists(name) == 32'h0) begin
-        global_id[name] = $fopen(name);
+        global_id[name] = $fopen(name, "w");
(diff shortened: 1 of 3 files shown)


uenoku (approved):
Thanks!

dtzSiFive:
I think verilator uses `"w"` for this, although maybe `"wb"` for `vpi_mcd_open`, if I'm following this right:

fopen paths:
https://github.com/verilator/verilator/blob/1432bb8889ca444bb9475f4bc897dd7cb1d0ef58/include/verilated.cpp#L1603
which calls:
https://github.com/verilator/verilator/blob/1432bb8889ca444bb9475f4bc897dd7cb1d0ef58/include/verilated_imp.h#L280

`vpi_mcd_open` implementation:
https://github.com/verilator/verilator/blob/1432bb8889ca444bb9475f4bc897dd7cb1d0ef58/include/verilated_vpi.cpp#L3570-L3574

PR #8984, opened 2025-09-18
[FIRRTL] Various small reduction pattern tweaks

fabianschuiki:
- Make the eager inliner not remove instances that are part of an NLA.
- Make the annotation remover not remove NLAs that have remaining uses after some annotations referencing it were removed.
- Make the constantifier not match on existing constants, and not replace operations that have the "inner_sym" attribute.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLReductions.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLReductions.cpp
@@ -776,14 +776,6 @@ struct ExtmoduleInstanceRemover : public OpReduction<firrtl::InstanceOp> {
 
 /// A sample reduction pattern that pushes connected values through wires.
 struct ConnectForwarder : public Reduction {
-  void beforeReduction(mlir::ModuleOp op) override { opsToErase.clear(); }
-  void afterReduction(mlir::ModuleOp op) override {
-    for (auto *op : opsToErase)
-      op->dropAllReferences();
-    for (auto *op : opsToErase)
-      op->erase();
-  }
-
   uint64_t match(Operation *op) override {
     if (!isa<firrtl::FConnectLike>(op))
       return 0;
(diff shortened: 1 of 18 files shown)


lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -956,24 +966,47 @@ struct NodeSymbolRemover : public OpReduction<firrtl::NodeOp> {
 };
 
 /// A sample reduction pattern that eagerly inlines instances.
-struct EagerInliner : public OpReduction<firrtl::InstanceOp> {
+struct EagerInliner : public OpReduction<InstanceOp> {
```
seldridge, line 977: Nice to keep this tidy.

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
```diff
@@ -1009,11 +1041,39 @@ struct EagerInliner : public OpReduction<firrtl::InstanceOp> {
...
+          if (parentTable.lookup(name)) {
+            // Collision found, return failure to stop the walk
+            return failure();
+          }
+          return success();
+        });
+
+    // If the walk failed, it means we found a collision
+    return failed(walkResult);
+  }
```
seldridge, line 1076: I kind of like this style of eagerly exit if the reduction pattern isn't trivially applicable. This is in contrast with trying to do a bunch of work to clean this up. I think this makes sense because the inner symbols will get dropped with other reductions.

seldridge (approved):
LGTM

lib/Dialect/FIRRTL/FIRRTLReductions.cpp
fabianschuiki: Totally agreed! It's also very annoying if the pattern applies too eagerly and then leads to invalid IR in a lot of cases. That just sends the reducer into busy work rabbit holes that don't really improve the reduction. :disappointed:

PR #8985, opened 2025-09-18
[Moore] Add `shortreal` type, bit <-> real conversions

Scheremo:
This PR disambiguates the `RealType` into 32-Bit real types ("shortreal") and 64-Bit real types ("real").
It adds support for materializing `moore.shortreal` from specified statements like
```
localparam shortreal sr = 4.532;
```
as well. 
Apart from this new type disctinction, this PR also adds support for the builtins `$shortrealtobits`, `$bitstoshortreal`, `$realtobits` and `$bitstoreal`.


include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -140,6 +140,11 @@ def IntType : MooreTypeDef<"Int", [], "moore::PackedType"> {
 def RealType : MooreTypeDef<"Real", [], "moore::UnpackedType"> {
   let mnemonic = "real";
   let summary = "a SystemVerilog real type";
+  // Default to 64 so old sites keep working.
+  let parameters = (ins
+    DefaultValuedParameter<"unsigned", "64">:$width
+  );
+  let hasCustomAssemblyFormat = 1;
```
fabianschuiki: I would remove the default value from this. It's usually a good idea to make sites that create these types be explicit about the exact width they want, even if it creates a little bit of churn in this PR. We probably don't create too many real types at the moment that need fixing :grin:

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -148,12 +153,23 @@ def RealType : MooreTypeDef<"Real", [], "moore::UnpackedType"> {
...
+    |-------------|--------------------|
+    | `shortreal` | `!moore.shortreal` |
+    | `real`      | `!moore.real`      |
+    | `realtime`  | `!moore.time`      |
+  }];
+  let extraClassDeclaration = [{
+    // Convenience helpers
+    static RealType getShortReal(MLIRContext *ctx) { return get(ctx, 32); }
+    static RealType getReal     (MLIRContext *ctx) { return get(ctx, 64); }
   }];
```
fabianschuiki, line 160: Since we only support two distinct widths, would it be possible to create an `enum class` to represent the two kinds of reals we support? Something like
```c++
enum class Width { F32 = 32, F64 = 64 };
```
in `extraClassDeclaration`, and then using `RealType::Width` as the parameter for this type. You could then still cast it to the bit width as `static_cast<unsigned>(width)`, but the compiler would introduce a cast hurdle to non-32-64 reals, which we don't support.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -483,13 +499,41 @@ def FourValuedI64 : MooreType<CPred<[{
...
+  let builderCall = "RealType::getShortReal($_builder.getContext())";
+}
+
+// A 64-bit real value.
+def VerilogReal : MooreType<CPred<[{
+    moore::isRealType($_self, 64)
+  }]>, "64-bit real type, aka. real", "moore::RealType"> {
+  let builderCall = "RealType::getReal($_builder.getContext())";
+}
+
```
fabianschuiki, line 524: Supreme nit: I'm wondering if we should call these `RealF32` and `RealF64` to be closer to how MLIR will call these in later dialects?

lib/Dialect/Moore/MooreTypes.cpp
```diff
@@ -111,6 +117,35 @@ void UnpackedType::print(mlir::AsmPrinter &printer) const {
...
+      return Type(); // signal failure, diagnostic already emitted
+  }
+
+  if (width == 32 || width == 64)
+    return RealType::get(p.getContext(), width);
+  p.emitError(p.getCurrentLocation())
+      << "RealType parser error: expected 64 or 32 bit width, but got "
+      << width;
+  return Type();
+}
```
fabianschuiki: I would make this parameter non-optional. We probably have a few tests that just call out `!moore.real` at the moment, but these should be fairly easy to fix up to `!moore.real<64>` :slightly_smiling_face:. It's usually a good idea to not have too much syntactic sugar in the IR.

fabianschuiki: That would also allow you to continue to use the `assemblyFormat` and not have custom C++ code for the parser/printer :slightly_smiling_face:. Which then also makes the `real` short form work when the `!moore.` is clear from context :slightly_smiling_face:.

fabianschuiki (approved):
Very cool! Just a few nits here and there :grin:

lib/Dialect/Moore/MooreTypes.cpp
fabianschuiki: Or actually coming to think of it, you could just add a `f32` and `f64` detection to `parseMooreType`, to make this real type look similar to MLIR's builtin `f32` and `f64`, which we might want to use in the future :grin:

include/circt/Dialect/Moore/MooreTypes.td
Scheremo: Fair enough! I removed the default and remodelled this to take an enum class `RealWidth`.

include/circt/Dialect/Moore/MooreTypes.td
Scheremo: Done! Implemented it in `MooreTypes.h` rather than `extraClassDeclaration` to align with how it's done in `IntType`.

include/circt/Dialect/Moore/MooreTypes.td
Scheremo: Done!

lib/Dialect/Moore/MooreTypes.cpp
Scheremo: I think I got all of those changed, but feel free to let me know if I missed any!

Scheremo:
I think I addressed all comments, and CI passes again. Ready to merge unless you got other comments 😄

PR #8987, opened 2025-09-18
[SCFToCalyx] Adapt to upstream change in `LoopLikeOpInterface`.

ingomueller-net:
This PR adapts to a change in `LoopLikeOpInterface` from llvm/llvm-project#158679, which changed the API for getting a statically known iteration count. The PR adapts to the change by using the accessor of the `ForOp` for exactly that information instead of relying on a free function that takes the properties/operands of the op.


ingomueller-net:
The PR will currently fail because it adapts to a change of MLIR that CIRCT doesn't see yet. It will have to be rebased onto or be made part of the next update of MLIR.

cgyurgyik (approved):

PR #8988, opened 2025-09-18
[Moore] Fix missing RValue conversion for struct_create

Scheremo:
This PR fixes a bug where RValue Conversion is not called when a struct is created from immediate values.


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1024,55 +1024,127 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+      assert(structType.getMembers().size() == elements->size());
+
+      return moore::StructCreateOp::create(builder, loc, structType, *elements);
     }
 
     // Handle packed arrays.
     if (auto arrayType = dyn_cast<moore::ArrayType>(type)) {
-      assert(arrayType.getSize() == elements.size());
-      return moore::ArrayCreateOp::create(builder, loc, arrayType, elements);
+      auto elements = convertElements(expr, {}, replCount);
```
fabianschuiki: Would it make sense to pass the expected element type in for arrays? They might suffer from the same issue as the struct patterns. You could have an `std::variant<Type, ArrayRef<Type>>` on `convertElements` to distinguish between a uniform type across all elements, and a specific type per element. Or maybe a separate `Type uniformType` operand? Or treat `expectedTypes` of size 1 as the uniform type for all elements?

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -3196,3 +3196,20 @@ function void TimeFormat();
...
+      logic [63:0] structField;
+   } testStruct;
+
+    // CHECK: [[TS:%.+]] = moore.struct_create [[IMM]] : !moore.l64 -> struct<{structField: l64}>
+   testStruct ts = '{structField: immediate};
+    // CHECK: [[CAST:%.+]] = moore.packed_to_sbv [[ARRAY]] : array<8 x l8>
+    // CHECK-NEXT: [[TS2:%.+]] = moore.struct_create [[CAST]] : !moore.l64 -> struct<{structField: l64}>
+   testStruct ts2 = '{structField: array};
+
+endfunction
```
fabianschuiki, line 3215: This exact pattern might also break for arrays. Might be worth a test to see if you can trigger the issue there.

fabianschuiki (approved):
LGTM! Excellent find :octopus:! Might be worth extending to the arrays as well :slightly_smiling_face:

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1024,55 +1024,127 @@ struct RvalueExprVisitor : public ExprVisitor {
...
-      std::reverse(elements.begin(), elements.end());
-      return moore::ConcatOp::create(builder, loc, intType, elements);
+      auto elements = convertElements(expr, {}, replCount);
+
+      if (failed(elements))
+        return {};
+
+      assert(intType.getWidth() == elements->size());
+      std::reverse(elements->begin(), elements->end());
+      return moore::ConcatOp::create(builder, loc, intType, *elements);
```
fabianschuiki, line 1113: Integers might also suffer from this lack of a cast. Maybe we need to convert all elements to simple bit vectors with the same domain as the `intType`? Probably worth adding a test to see if you can produce the same failure you've seen for structs also for an assignment pattern to an integer :thinking:

test/Conversion/ImportVerilog/basic.sv
Scheremo: So I *think* arrays are okay; I tried the following:

```verilog
logic [7:0][7:0] array;
logic [63:0] ts3 = array;
```

And it gets lowered correctly to:
```verilog
%arg0: !moore.array<8 x l8>
...
%3 = moore.packed_to_sbv %arg0 : array<8 x l8>
%ts3 = moore.variable %3 : <l64>
```

Scheremo: I also tried struct -> arrary, which also works.

```verilog
   typedef struct packed {
      logic [63:0] structField;
   } testStruct;
   logic [63:0] immediate;
   testStruct ts = '{structField: immediate};
   logic [7:0][7:0] ts3 = ts;
```
turns into 
```verilog
    %0 = moore.struct_create %arg1 : !moore.l64 -> struct<{structField: l64}>
    %ts = moore.variable %0 : <struct<{structField: l64}>>
    ...
    %3 = moore.read %ts : <struct<{structField: l64}>>
    %4 = moore.packed_to_sbv %3 : struct<{structField: l64}>
    %5 = moore.sbv_to_packed %4 : array<8 x l8>
    %ts3 = moore.variable %5 : <array<8 x l8>>
```

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: From what I've tried so far I haven't been able to trigger the same issue with direct-to-int assignment.
I adapted the test code described above for `int ts`, `logic [63:0] ts`, and `integer ts`. All seem to work fine on my end.

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: I can definitely adapt that. I haven't been able to produce the same bug for Array conversions though - should we still go for it?

Scheremo: Refactored to use a `std::variant<Type, ArrayRef<Type>>` and applying it to array elements as well. 
Since int-casts are anyway on the "direct and happy path" in `convertRvalueExpression` and the current code treats `IntType` assignment over optionally multiple RValues, I left those out as it seems unsafe (e.g. you could concat a 2-bit and two 1-bit expressions to a 4-bit expression, explicit casts would mess this up, probably).

Scheremo:
As far as I can tell I've addressed all comments, please feel free to raise more or point out ones I might've missed. Otherwise I am okay to merge 😄

test/Conversion/ImportVerilog/basic.sv
fabianschuiki: Oh that's very interesting! So might truly just be the structs that have this behavior. :+1: No need to fix stuff that isn't broken :grin:

PR #8989, opened 2025-09-19
[ESI2Phy] Adapt usage of rewriter to upcoming version.

ingomueller-net:
This PR changes the combination of `replaceAllUsesWith` followed by `eraseOp` with `replaceOp`, which has the same effect. The original pattern, however, will be unsupported by the upcoming version of the dialect conversion; see llvm/llvm-project#158298.


teqdruid:
Thanks!

PR #8997, opened 2025-09-20
[Synth] LowerWordsToBits: Improve scalability with bit-sensitive constprop

uenoku:
Previously, GreedyPatternRewriter was naively used to bit-blast multi-bit values but it didn't scale well due to worklist iteration and unnecessary IR creation which occasionally ended up with constant values. Hence the pass was often not scalable for large designs.

This commit fixes the issue by applying constant propagation during the bit-blast process and replaces GreedyPatternRewriter with a manual one-shot walk with recursion.


fabianschuiki (approved):
LGTM! Cool to see the per-bit constant propagation prune out some unneeded ops :smile:

Issue #8673, opened 2025-07-09
[LTL][SVA] Proposal for Explicit Clocking in the CIRCT LTL(SVA) Intermediate Representation


sequencer:
# Analysis of SystemVerilog Assertions (SVA) Clocking Semantics: A Foundation for Ambiguity

In SystemVerilog, concurrent assertions without an explicit clocking event infer the clock from the surrounding context. This "action at a distance" violates the principle of locality, a key tenet of effective compiler IR design, where an operation's semantics should be self-contained.

The inference hierarchy, in descending order of priority, is as follows:
- **Explicit Clock in the `Property`/`Sequence` Instance**: An explicit clocking event specified directly in the assertion statement or property instance takes highest precedence.
- **Explicit Clock in the `Property`/`Sequence` Declaration**: A leading clocking event within a named property or sequence declaration is used if the instance itself has no explicit clock.9
- **Inferred Clock from a Procedural Block**: If a concurrent assertion is embedded within a procedural block (e.g., `always`, `always_ff`), the clocking event of that block is inferred as the clock for the assertion. This is a common source of confusion, as the procedural activation mechanism becomes entangled - with the temporal sampling clock of the assertion.
**Inferred Clock from a default clocking Block**: A module- or interface-level default clocking block can specify a default clock for all concurrent assertions within its scope that have not otherwise determined a clock.

Clock determination for SVA is complex, hindering automated analysis. A simple sequence like ##1 b requires checking for explicit clocks in assert statements or property definitions, then traversing the IR tree for procedural block sensitivity (e.g., always @(posedge clk)), and finally scanning for default clocking definitions. This non-local dependency complicates IR transformations. For example, hoisting a common sub-sequence from blocks with different clocks could silently alter program meaning, creating "murky semantics" and analytical "annoyance." IR should facilitate, not obstruct, transformations with hidden semantic dependencies.

Opaque Semantics of Clock Flow and Propagation
---
SVA's intricate rules for clock flow, especially concerning named sequences and parenthesized expressions, lead to non-compositional semantics problematic for IR.

The LRM defines two key concepts: clock flow and clock trapping:
- **Clock Flow**: A leading clocking event generally flows from left to right across sequence operators like ## and implication operators like `|=>`. For example, in `@(posedge clk) a ##1 b`, the `posedge clk` event serves as the clock for both a and the `##1` delay leading to `b`.
[...]

PR #8999, opened 2025-09-21
[LTL][ImportVerilog] Add support for $rose, $stable, $fell

Scheremo:
This commit adds support for lowering `$rose` `$stable` and `$fell` from assertion contexts into LTL.
To this end, this PR adds a new operation for look-back semantics, `ltl.PastOp`.

* **New op**
  * `ltl.past`: takes an integer and a delay, returning the sequence observed `$delay` cycles earlier.
* **Assertion system call lowering**
  * Added `Context::convertAssertionSystemCallArity1` to map single-argument system calls into LTL constructs:
    * `$rose(x)` → `x ∧ ¬past(x,1)`
    * `$fell(x)` → `¬x ∧ past(x,1)`
    * `$stable(x)` → `(x ∧ past(x,1)) ∨ (¬x ∧ ¬past(x,1))`
  * Integrated into `convertAssertionCallExpression` to dispatch on `$rose/$fell/$stable` from the AST.
* **Conversion plumbing**
  * Added entry point `convertAssertionCallExpression` to `Context`.
  * Adjusted `Expressions.cpp` visitor to recognize these system calls and redirect them to the new assertion conversion path.
* **Tests**
  * Extended `builtins.sv` with assertions using `$rose`, `$fell`, `$stable`.
  * CHECK lines verify correct lowering into `ltl.past`, `ltl.not`, `ltl.and`, `ltl.or` sequences.

### DIFF
--- a/include/circt/Dialect/LTL/LTLOps.td
+++ b/include/circt/Dialect/LTL/LTLOps.td
@@ -106,7 +106,7 @@ def DelayOp : LTLOp<"delay", [Pure]> {
 
 def PastOp : LTLOp<"past", [Pure]> {
   let arguments = (ins
-    LTLAnySequenceType:$input,
+    I1:$input,
     I64Attr:$delay);
   let results = (outs LTLSequenceType:$result);
   let assemblyFormat = [{


sequencer:
I think this PR should be implemented after the https://github.com/llvm/circt/issues/8673 is finished. Wdyt?

include/circt/Dialect/LTL/LTLOps.td
```diff
@@ -104,6 +104,21 @@ def DelayOp : LTLOp<"delay", [Pure]> {
   }];
 }
 
+def PastOp : LTLOp<"past", [Pure]> {
+  let arguments = (ins
+    LTLAnySequenceType:$input,
```
fabianschuiki: What are the semantics of `ltl.past` when used on a sequence instead of a boolean? Does that mean the sequence must have finished matching in the previous cycle? And how would we print an `ltl.past` with a sequence input back as Verilog?

If I remember correctly, the `$past` function in Verilog only operates on booleans. In that case it might make sense to make `ltl.past` operate on just `i1` for now. It's always easier to start with an overconstrained op and relax later, than figuring out semantics later 😃

fabianschuiki (approved):
LGTM! Only a minor comment on past on sequences.

fabianschuiki:
> I think this PR should be implemented after the #8673 is finished. Wdyt?

@sequencer Good point, this also has the concerning quality of magically inferring what clock the `$past` refers to. Since #8673 has to do an overhaul of all implicitly clocked ops anyway, I think it would make sense to not block this PR behind #8673 and just add `ltl.past` to the list of ops that will need a clock there. Otherwise we create a dependency between the PRs which might not be necessary.

Do you have an idea how we could do that clock inference in ImportVerilog? Would we apply the clock inference rules of the SV standard and search for a parent construct that has a clock, and then use that when we create the LTL ops?

include/circt/Dialect/LTL/LTLOps.td
Scheremo: That's a fair point, I'll limit it to i1 for this PR!

Scheremo:
> I think this PR should be implemented after the #8673 is finished. Wdyt?

Hi, I wasn't aware of the discussion in #8673. It's clearly important & useful work, but judging from the duration from opening until now (2.5 months) I would suppose it will take some time to land.
I tend to agree with Fabian's viewpoint that the semantic issue of `past` is the same as `delay`, and adjusting both at once seems reasonable. I would argue having more existing test surface for refactoring/restructuring the dialect by merging this PR might actually be helpful. From my personal point of view having at least intermediate support for rose, fell, and stable is very useful, as lots of SV code use these constructs.
Does that seem reasonable to you?

MikePopoloski:
FWIW slang implements all of the clock flow and clock inference rules from the LRM in its analysis layer, e.g. https://github.com/MikePopoloski/slang/blob/6637781e9315193e6ca2cfd3128d91eb05acd41a/source/analysis/AnalyzedAssertion.cpp#L269 (and yes, the rules are arcane and messy).

This is not easily exposed via a nice API at the moment, mostly because I wasn't sure how downstream consumers would like it to be presented. I'm happy to accept a proposal for how that information could be extracted and made usable by the ImportVerilog pass.

fabianschuiki:
> FWIW slang implements all of the clock flow and clock inference rules from the LRM in its analysis layer

@MikePopoloski Reusing this from Slang would be absolutely fantastic :partying_face:. Not having to go through all the pain of distilling out those rules again, and instead helping maintain a single implementation sounds much more reasonable. 

ImportVerilog simply walks the AST and conjures up IR ops along the way. It would be nice if we could either run the analysis for asserts and sampled value funcs as we encounter them, and then basically ask what the implied clock is, or if we could run the analysis once on the entire AST, and then just query that analysis for the implied clocks. Would something like that make sense?

fabianschuiki:
I think I'm going to land this now to not hold up subsequent work. Hope that's okay with you, @sequencer!

MikePopoloski:
> > FWIW slang implements all of the clock flow and clock inference rules from the LRM in its analysis layer
> 
> @MikePopoloski Reusing this from Slang would be absolutely fantastic 🥳. Not having to go through all the pain of distilling out those rules again, and instead helping maintain a single implementation sounds much more reasonable.
> 
> ImportVerilog simply walks the AST and conjures up IR ops along the way. It would be nice if we could either run the analysis for asserts and sampled value funcs as we encounter them, and then basically ask what the implied clock is, or if we could run the analysis once on the entire AST, and then just query that analysis for the implied clocks. Would something like that make sense?

Yes, the AnalysisManager runs in one pass once the AST is fully constructed. The question is how to store and query the results of the assertion analysis.

fabianschuiki:
Could we populate a hash map with the analysis results for each relevant AST node? Something that makes it easy to later pick up one of the asserts or sampled value function calls, and ask the analysis for the stored information associated with that.

If the analysis is just visiting the AST and generating diagnostics, you could also leave the burden of storing the analysis results in a meaningful format to the user by accepting a callback function that the analysis calls whenever it has gathered some information. ImportVerilog could then supply its own callback that just stores the info in a format convenient for it. 

On the other hand, I could see other users of Slang be interested in this info. Having the analysis result stored in some form on the Slang side would be useful for that.

MikePopoloski:
I opened https://github.com/MikePopoloski/slang/issues/1528 to track adding the API. I'm leaning toward a callback but we'll see how it looks when I start working on it.

Scheremo:
> I opened [MikePopoloski/slang#1528](https://github.com/MikePopoloski/slang/issues/1528) to track adding the API. I'm leaning toward a callback but we'll see how it looks when I start working on it.

You're a legend, Mike! Let us know if we can help in any way.

PR #9000, opened 2025-09-21
[Arc] Add MergeTaps pass

fzi-hielscher:
Implements the MergeTaps (previously called CombineTaps) pass and integrates it into the arcilator pipeline.

The MergeTaps pass collects `arc.tap` operations that observe the same (SSA) value and merges them into a single TapOp if they are within the same HWModule body and not nested under another region. To do so, the TapOp itself had to be changed to carry a `StrArrayAttr` for the names instead of just a single string.

Applying the MergeTaps pass after flattening the module hierarchy can significantly reduce the number of redundant taps. Testing it on the `rocket-small` configuration of the arc tests repo
```arcilator rocket/build/small-master/rocket.mlir --observe-wires --observe-ports --observe-named-values --observe-registers --observe-memories --until-before=state-lowering | grep arc\.tap | wc -l```
gives 9427 now vs. 19348 before.

Part of the VCD tracer implementation #8986.

### DIFF
--- a/lib/Dialect/Arc/ModelInfo.cpp
+++ b/lib/Dialect/Arc/ModelInfo.cpp
@@ -55,8 +55,16 @@ LogicalResult circt::arc::collectStates(Value storage, unsigned offset,
     if (!isa<AllocStateOp, RootInputOp, RootOutputOp, AllocMemoryOp>(op))
       continue;
 
+    SmallVector<StringAttr> names;
     auto opName = op->getAttrOfType<StringAttr>("name");
-    if (!opName || opName.getValue().empty())
+    if (opName && !opName.getValue().empty()) {
+      names.push_back(opName);
+    } else if (auto nameAttrs = op->getAttrOfType<ArrayAttr>("names")) {
+      for (auto attr : nameAttrs)
+        names.push_back(cast<StringAttr>(attr));
+    }
+
+    if (names.empty())
       continue;
 
     auto opOffset = op->getAttrOfType<IntegerAttr>("offset");
@@ -64,9 +72,9 @@ LogicalResult circt::arc::collectStates(Value storage, unsigned offset,
       return op->emitOpError(
           "without allocated offset; run state allocation first");
 
+    StateInfo stateInfo;
     if (isa<AllocStateOp, RootInputOp, RootOutputOp>(op)) {
       auto result = op->getResult(0);
-      auto &stateInfo = states.emplace_back();
       stateInfo.type = StateInfo::Register;
       if (isa<RootInputOp>(op))
         stateInfo.type = StateInfo::Input;
@@ -76,9 +84,12 @@ LogicalResult circt::arc::collectStates(Value storage, unsigned offset,
         if (alloc.getTap())
           stateInfo.type = StateInfo::Wire;
       }
-      stateInfo.name = opName.getValue();
       stateInfo.offset = opOffset.getValue().getZExtValue() + offset;
       stateInfo.numBits = cast<StateType>(result.getType()).getBitWidth();
+      for (auto name : names) {
+        stateInfo.name = name.getValue();
+        states.push_back(stateInfo);
+      }
       continue;
     }
 
@@ -89,13 +100,15 @@ LogicalResult circt::arc::collectStates(Value storage, unsigned offset,
             "without allocated stride; run state allocation first");
       auto memType = memOp.getType();
       auto intType = memType.getWordType();
-      auto &stateInfo = states.emplace_back();
       stateInfo.type = StateInfo::Memory;
-      stateInfo.name = opName.getValue();
       stateInfo.offset = opOffset.getValue().getZExtValue() + offset;
       stateInfo.numBits = intType.getWidth();
       stateInfo.memoryStride = stride.getValue().getZExtValue();
       stateInfo.memoryDepth = memType.getNumWords();
+      for (auto name : names) {
+        stateInfo.name = name.getValue();
+        states.push_back(stateInfo);
+      }
       continue;
     }
   }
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: lib/Dialect/Arc/Transforms/LowerState.cpp [hunk 1]
```diff
@@ -794,7 +794,7 @@ LogicalResult OpLowering::lower(TapOp op) {
     auto alloc = AllocStateOp::create(module.allocBuilder, op.getLoc(),
                                       StateType::get(value.getType()),
                                       module.storageArg, true);
-    alloc->setAttr("name", op.getNameAttr());
+    alloc->setAttr("names", op.getNamesAttr());
```
fzi-hielscher, line 797: We should consider making this an inherent attribute of AllocStateOp. It is quite easy to mess up name propagation otherwise.

fabianschuiki: A while ago @maerhart and I were wondering if we shouldn't just remove names from state ops and alloc ops altogether. Instead, we could just rely on the `arc.tap` op to point to the things that should be observed, and have the tap reuse and existing state op's allocated storage if possible as an optimization. :thinking: WDYT?

fabianschuiki: In the future we'd ideally just rely on `dbg.variable` to mark things that should be preserved in waveform dumps. We could lower that to `arc.tap` ops, or we might be able to lower `dbg.variable` directly to your waveform streaming API calls.

lib/Dialect/Arc/Transforms/MergeTaps.cpp
```diff
@@ -0,0 +1,74 @@
...
+struct MergeTapsPass : public arc::impl::MergeTapsBase<MergeTapsPass> {
+  using MergeTapsBase::MergeTapsBase;
+
+  void mergeTaps(HWModuleOp hwMod);
+
+  void runOnOperation() override {
+    mlir::parallelForEach(getOperation().getContext(),
+                          getOperation().getBody()->getOps<hw::HWModuleOp>(),
+                          [this](HWModuleOp hwMod) { mergeTaps(hwMod); });
+  }
```
uenoku: Maybe it's not necessary to make this mlir::ModuleOp if there is no global information required?

lib/Dialect/Arc/ModelInfo.cpp
```diff
@@ -55,18 +55,26 @@ LogicalResult circt::arc::collectStates(Value storage, unsigned offset,
...
       continue;
 
+    SmallVector<StringAttr> names;
     auto opName = op->getAttrOfType<StringAttr>("name");
-    if (!opName || opName.getValue().empty())
+    if (opName && !opName.getValue().empty()) {
+      names.push_back(opName);
+    } else if (auto nameAttrs = op->getAttrOfType<ArrayAttr>("names")) {
+      for (auto attr : nameAttrs)
+        names.push_back(cast<StringAttr>(attr));
```
maerhart: Do we have to check that it is actually a `StringAttr` since "names" is not a verified attribute?
Also, why is it not necessary to check for empty strings in this case?

lib/Dialect/Arc/Transforms/AddTaps.cpp
```diff
@@ -95,7 +95,8 @@ struct AddTapsPass : public arc::impl::AddTapsBase<AddTapsPass> {
       return;
     if (isa<seq::ClockType>(value.getType()))
       value = builder.createOrFold<seq::FromClockOp>(loc, value);
-    arc::TapOp::create(builder, loc, value, name);
+    arc::TapOp::create(builder, loc, value,
+                       builder.getArrayAttr({builder.getStringAttr(name)}));
```
maerhart: Nit: might be useful to have a convenience builder for that case

maerhart (approved):
Really nice!

fzi-hielscher:
Thank you all for your feedback. Thinking about it, I have started to wonder if this pass is even necessary. Wouldn't it also be sufficient to ensure in LowerState that only one `AllocStateOp` per SSA value is created?

> In the future we'd ideally just rely on dbg.variable to mark things that should be preserved in waveform dumps. We could lower that to arc.tap ops, or we might be able to lower dbg.variable directly to your waveform streaming API calls.

I had also considered using the Debug Dialect, I just could not figure out how to make it work quickly. I would definitely prefer having a hierarchical representation of the signal names in the compiled model, rather than painstakingly reconstruct it at runtime as I'm currently doing it. Inserting `dbg.variable` ops instead of `arc.tap` and then somehow propagating them through LowerState seems like a good idea to me. I'd have to give it a shot to see if/how it works. 

So - should I continue upstreaming the current approach or just go back to the drawing board immediately? 🤔

fabianschuiki:
@fzi-hielscher I suspect a lot of the work you've done isn't really tied to `arc.tap` and would just as well work with the debug dialect, with minor tweaks to encode the hierarchy. But it seems very useful to get VCD output going in the current state of the Arc dialect, and then later do a focused effort of replacing `arc.tap` with `dbg.variable` and `dbg.scope`. From that point of view, `MergeTaps` clearly provides and improvement over what we currently have and would therefore be very useful to have :slightly_smiling_face: :+1:

maerhart:
> Thank you all for your feedback. Thinking about it, I have started to wonder if this pass is even necessary. Wouldn't it also be sufficient to ensure in LowerState that only one `AllocStateOp` per SSA value is created?

I like your approach here. It keeps the already complicated LowerState pass simpler and the added compile time should be minimal.

> So - should I continue upstreaming the current approach or just go back to the drawing board immediately? 🤔

I'm also in favour of merging this approach and attempt replacing the taps and name attributes with debug dialect later on.

fzi-hielscher:
Right, then let's go ahead with this. 👍 

Annoyingly, the rocket tests currently fail to fully compile on AArch64 when the --observe-* flags are set, due to this LLVM bug: https://godbolt.org/z/aW7aKWP9f
I've hotfixed it locally, but haven't yet opened an issue or PR in upstream.

PR #9001, opened 2025-09-23
[circt-test] accept LoweringOptions to control the generated Verilog

ollef:
Since SBY doesn't support all System Verilog constructs, it's useful to be able to control how lowering is done. We may e.g. pass `disallowLocalVariables` when using SBY as a runner.

Ideally, this would be more automated (i.e. automatically pick a set of lowering options that is known to work for the specific runner), but that would be a bigger change since the different runners would then require different lowering. We also don't currently have all the lowering options we would need for this.


ollef:
Thanks for the review. I don't have commit access, so it would be great if you landed my PRs for me. :)

PR #8304, opened 2025-03-06
[circt-verilog-lsp] Add hover support

uenoku:
Add hover functionality to the CIRCT Verilog LSP server. This includes hover provider capability in server initialization and hover context line configuration option.

The implementation adds hover support for Verilog symbols showing type information, definition location with source code, and links to source locations. It also implements hover for external source locations showing context-aware source code snippets and links to external files.

The hover feature provides rich information when hovering over Verilog symbols (showing type and definition) and source location annotations (showing external source contents such as MLIR and scala).

PR #8280, opened 2025-02-27
[circt-verilog-lsp] Add definition and reference providers

uenoku:
This commit adds "Go to Definition" and "Find References" functionality to the CIRCT Verilog LSP server. These features allow users to navigate between symbol definitions and usages within Verilog source files. Besides it provides cross references for source location comments.

- Added VerilogIndex class to track symbol definitions and references
- Implemented source location parsing for external file references
- Added support for LSP definition and reference requests
- Extended server capabilities to advertise the new functionality
- Added test cases to verify the new features

Following PDLL's LSP implemetation, the implementation uses an IntervalMap to efficiently map source positions
to symbols, allowing quick lookups when users request definitions or references. It also supports source location annotations in comments that can link to external MLIR files, enabling cross-language navigation.

PR #9003, opened 2025-09-23
[circt-verilog-lsp] Add support for `-C` command files for Slang, project-wide defintion lookup

Scheremo:
This PR extends the CIRCT Verilog LSP server with support for project command files (`-C` / `--command-file`) and improves robustness around source locations and diagnostics.

- **Command file support**
  - Added `commandFiles` option to `VerilogServerOptions`.
  - Extended the CLI with `-C`/`--command-file` flags to pass one or more command files.
  - Each buffer’s `Driver` now processes command files, filtering out the current main buffer to avoid duplication.

- **Source location handling**
  - Improved absolute/real path resolution when constructing LSP URIs.
  - Fallbacks added for when `slang::SourceManager` has no full path info.

- **Compilation setup**
  - Ensure all definitions in the main buffer scope are treated as top modules.

- **Indexing improvements**
  - Skip top instances not defined in the current file.
  - Added guards against invalid or empty ranges (especially macro expansions).

~~Currently the main thing I am unhappy with is that two compilations are created, where the first one is only to determine all module names - this is necessary, since we index only instantiated modules, and only top modules are instantiated. I tried looking, but didn't find a flag to treat all module definitions as top modules. Maybe someone else has a deeper understanding of the Slang driver and knows a way around this?~~

~~The other thing that is a bit hacky is the mangling of the command file; this is necessary because we do definition lookups based on raw pointers, so if the main document is also in the command file the pointers do not match. To prevent this I currently scan the command file and remove the main file (if present). I think we could prevent this if instead of matching raw pointers we matched against a file id / filename + offset. However, this might be slower than the current solution since it adds another layer of indirection. Let me know if you have opinions!~~

Refactored top module determination to only process the main buffer file; this is sufficient, no need to compile twice. 
Also refactored command file mangling to instead not import the main buffer into slang compilation if it is present in any command file, and then after compilation bind it to the llvm source manager's buffer id.

### DIFF
--- a/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
+++ b/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
@@ -21,8 +21,11 @@
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/Location.h"
 #include "mlir/IR/MLIRContext.h"
+#include "mlir/Support/FileUtilities.h"
 #include "slang/ast/ASTVisitor.h"
 #include "slang/ast/Compilation.h"
+#include "slang/ast/Scope.h"
+#include "slang/ast/symbols/CompilationUnitSymbols.h"
 #include "slang/diagnostics/DiagnosticClient.h"
 #include "slang/diagnostics/Diagnostics.h"
 #include "slang/driver/Driver.h"
@@ -38,11 +41,14 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/LSP/Logging.h"
 #include "llvm/Support/LSP/Protocol.h"
+#include "llvm/Support/LineIterator.h"
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/SMLoc.h"
 #include "llvm/Support/SourceMgr.h"
 
+#include <filesystem>
+#include <fstream>
 #include <memory>
 #include <optional>
 #include <string>
@@ -248,20 +254,86 @@ void LSPDiagnosticClient::report(const slang::ReportedDiagnostic &slangDiag) {
 // VerilogDocument
 //===----------------------------------------------------------------------===//
 
+static std::filesystem::path
+canonicalizeFileName(const std::filesystem::path &file) {
+  std::error_code ec;
+  std::filesystem::path path = std::filesystem::weakly_canonical(file, ec);
+  if (ec)
+    path = std::filesystem::absolute(file).lexically_normal();
+  return path;
+}
+
+// Filter out the main buffer file from the command file list, if it is in
+// there.
+static inline bool
+mainBufferFileInCommandFileList(const std::string &cmdfileStr,
+                                const std::string &targetAbsStr) {
+  const std::filesystem::path targetAbs =
+      canonicalizeFileName(std::filesystem::path(targetAbsStr));
+
+  std::string error;
+  auto cmdFile = mlir::openInputFile(cmdfileStr, &error);
+  if (!cmdFile) {
+    circt::lsp::Logger::error(Twine("Failed to open command file ") +
+                              cmdfileStr + ": " + error);
+    return false;
+  }
+
+  const std::filesystem::path base =
+      std::filesystem::path(cmdFile->getBufferIdentifier().str()).parent_path();
+
+  // Read line by line, ignoring empty lines and comments.
+  for (llvm::line_iterator i(*cmdFile); !i.is_at_eof(); ++i) {
+    llvm::StringRef line = i->trim();
+
+    if (line.empty())
+      continue;
+
+    static constexpr llvm::StringRef commandPrefixes[] = {"+", "-"};
+    auto isCommand = [&line](llvm::StringRef s) { return line.starts_with(s); };
+    if (llvm::any_of(commandPrefixes, isCommand))
+      continue;
+
+    auto candRel = std::filesystem::path(line.str());
+    auto candAbs = canonicalizeFileName(
+        candRel.is_absolute() ? candRel : (base / candRel));
+
+    if (candAbs == targetAbs)
+      return true;
+  }
+  return false;
+}
+
 VerilogDocument::VerilogDocument(
     VerilogServerContext &context, const llvm::lsp::URIForFile &uri,
     StringRef contents, std::vector<llvm::lsp::Diagnostic> &diagnostics)
     : globalContext(context), index(*this), uri(uri) {
-  unsigned int bufferId;
-  if (auto memBufferOwn =
-          llvm::MemoryBuffer::getMemBufferCopy(contents, uri.file())) {
+  bool skipMainBufferSlangImport = false;
 
-    bufferId = sourceMgr.AddNewSourceBuffer(std::move(memBufferOwn), SMLoc());
-  } else {
+  llvm::SmallString<256> canonPath(uri.file());
+  if (std::error_code ec = llvm::sys::fs::real_path(uri.file(), canonPath))
+    canonPath = uri.file(); // fall back, but try to keep it absolute
+
+  // --- Apply project command files (the “-C”s) to this per-buffer driver ---
+  for (const std::string &cmdFile : context.options.commandFiles) {
+    if (!driver.processCommandFiles(cmdFile, false, true)) {
+      circt::lsp::Logger::error(Twine("Failed to open command file ") +
+                                cmdFile);
+    }
+    skipMainBufferSlangImport |=
+        mainBufferFileInCommandFileList(cmdFile, canonPath.str().str());
+  }
+
+  auto memBufferOwn =
+      llvm::MemoryBuffer::getMemBufferCopy(contents, uri.file());
+  if (!memBufferOwn) {
     circt::lsp::Logger::error(
         Twine("Failed to create memory buffer for file ") + uri.file());
     return;
   }
+
+  const unsigned int mainBufferId =
+      sourceMgr.AddNewSourceBuffer(std::move(memBufferOwn), SMLoc());
 
   // Build the set of include directories for this file.
   llvm::SmallString<32> uriDirectory(uri.file());
@@ -273,17 +345,60 @@ VerilogDocument::VerilogDocument(
                  context.options.libDirs.end());
 
   // Populate source managers.
-  const llvm::MemoryBuffer *memBuffer = sourceMgr.getMemoryBuffer(bufferId);
+  const llvm::MemoryBuffer *memBuffer = sourceMgr.getMemoryBuffer(mainBufferId);
+
+  // This block compiles the top file to determine all definitions
+  // This is used in a second pass to declare all those definitions
+  // as top modules, so they are elaborated and subsequently indexed.
+  {
+    slang::driver::Driver topDriver;
+
+    auto topSlangBuffer =
+        topDriver.sourceManager.assignText(uri.file(), memBuffer->getBuffer());
+    topDriver.sourceLoader.addBuffer(topSlangBuffer);
+
+    topDriver.options.compilationFlags.emplace(
+        slang::ast::CompilationFlags::LintMode, false);
+    topDriver.options.compilationFlags.emplace(
+        slang::ast::CompilationFlags::DisableInstanceCaching, false);
+
+    if (!topDriver.processOptions()) {
+      return;
+    }
+
+    if (!topDriver.parseAllSources()) {
+      circt::lsp::Logger::error(Twine("Failed to parse Verilog file ") +
+                                uri.file());
+      return;
+    }
+
+    FailureOr<std::unique_ptr<slang::ast::Compilation>> topCompilation =
+        topDriver.createCompilation();
+    if (failed(topCompilation))
+      return;
+
+    std::vector<std::string> topModules;
+    for (const auto *defs : (*topCompilation)->getDefinitions())
+      topModules.emplace_back(defs->name);
+
+    // Make sure that all possible definitions in the main buffer are
+    // topModules!
+    driver.options.topModules = std::move(topModules);
+  }
 
   for (const auto &libDir : libDirs) {
     driver.sourceLoader.addSearchDirectories(libDir);
   }
 
-  // Assign text to slang.
-  auto slangBuffer =
-      driver.sourceManager.assignText(uri.file(), memBuffer->getBuffer());
-  driver.sourceLoader.addBuffer(slangBuffer);
-  bufferIDMap[slangBuffer.id.getId()] = bufferId;
+  // If the main buffer is **not** present in a command file, add it into
+  // slang's source manager and bind to llvm source manager.
+  if (!skipMainBufferSlangImport) {
+    auto slangBuffer =
+        driver.sourceManager.assignText(uri.file(), memBuffer->getBuffer());
+    driver.sourceLoader.addBuffer(slangBuffer);
+    bufferIDMap[slangBuffer.id.getId()] = mainBufferId;
+  }
+
   auto diagClient = std::make_shared<LSPDiagnosticClient>(*this, diagnostics);
   driver.diagEngine.addClient(diagClient);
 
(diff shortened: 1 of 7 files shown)


Scheremo:
> Hi looks super cool! Leveraging slang -C command is great! As for testing I had a PR for adding definition provider #8280 so maybe you can check tests there (test/Tools/circt-verilog-lsp-server/find-definition.test). Non blocking but regarding location look-up we might want to store a mapping from location range to symbol using llvm::IntervalMap since it's very frequently used (like hovering #8304).

Hi @uenoku! Since you're working on virtually the same thing, why don't we join forces?
I think your implementation in #8280 looks very complete & good, is there anything preventing us from merging this?
I believe I can add `-C` support on top of that pretty easily. WDYT?

uenoku:
> Since you're working on virtually the same thing, why don't we join forces?

That would be really cool!

> I think your implementation in https://github.com/llvm/circt/pull/8280 looks very complete & good
Thanks! 

> is there anything preventing us from merging this?

No, I should have merged more earlier :) Let me merge today or tomorrow.

> I believe I can add -C support on top of that pretty easily. WDYT?

Yes, I agree! I think you can add SlangProjectLookup to VerilogIndex or VerilogDocument.

Scheremo:
> > Since you're working on virtually the same thing, why don't we join forces?
> 
> That would be really cool!
> 
> > I think your implementation in #8280 looks very complete & good
> > Thanks!
> 
> > is there anything preventing us from merging this?
> 
> No, I should have merged more earlier :) Let me merge today or tomorrow.
> 
> > I believe I can add -C support on top of that pretty easily. WDYT?
> 
> Yes, I agree! I think you can add SlangProjectLookup to VerilogIndex or VerilogDocument.

Awesome! Just give me a ping once that's done and I'll work out the rest from there :smile:

uenoku:
Merged the PR, sorry for the merge conflict!

Scheremo:
> Merged the PR, sorry for the merge conflict!

No worries about that @uenoku! I adapted this version to integrate with the pointer-based lookup you implemented and added a test for cross-file resolution; I also extended the way we index by setting the top modules explicitly and only indexing the ones whose location is in the actual file that's open. Without this optimization, on a large project like the [snitch_cluster](https://github.com/pulp-platform/snitch_cluster) with a ~400 line file list the LSP takes ~30 seconds to index (on my MBP M4 Max). With this optimization I didn't manage to eyeball-measure the time it takes anymore as it's too small (<1s).

While the regression doesn't capture performance, inter-file lookup in an appreciably large project is reasonably performant. No feeling of lag with the snitch_cluster on my end.

uenoku:
Super cool! Thank you for rebasing it. 

> where the first one is only to determine all module names - this is necessary, since we index only instantiated modules, and only top modules are instantiated. I tried looking, but didn't find a flag to treat all module definitions as top modules.

Does "all module names" here mean modules defined in the main file? I think this is ok if the modules not defined in the main file is not elaborated. 

> To prevent this I currently scan the command file and remove the main file (if present). I think we could prevent this if instead of matching raw pointers we matched against a file id / filename + offset. However, this might be slower than the current solution since it adds another layer of indirection.

I think we can tweak the `getSMLoc()` to return a pointer in main buffer if slang location points to the current main file by comparing a file path. Changing the key of IntervalMap to offset looks reasonable as well :+1:

>  I also extended the way we index by setting the top modules explicitly and only indexing the ones whose location is in the actual file that's open. Without this optimization, on a large project like the [snitch_cluster](https://github.com/pulp-platform/snitch_cluster) with a ~400 line file list the LSP takes ~30 seconds to index (on my MBP M4 Max). With this optimization I didn't manage to eyeball-measure the time it takes anymore as it's too small (<1s).

That's super cool and clever! Thank you for adding the optimization. Relevantly in the long term it would be really cool if we can store indexed data to disk like clangd https://clangd.llvm.org/design/indexing

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -358,8 +477,24 @@ llvm::SMLoc VerilogDocument::getSMLoc(slang::SourceLocation loc) {
...
+      if (!llvm::sys::fs::real_path(fname, realPath)) {
+        name = realPath.str().str();
+      } else {
+        name = fname.str();
+      }
+    } else {
+      name = path.string();
+    }
+
+    auto memBuffer = llvm::MemoryBuffer::getFile(name);
```
uenoku: I think we can look up the main buffer from here instead of opening a new memory buffer
```cpp
if (getFileName(uri) == name) { 
  bufferIDMap[bufferID] = 0; // Insert the id for the next call.  
  const auto *buffer = sourceMgr.getMemoryBuffer(0); // get main buffer
  return return llvm::SMLoc::getFromPointer(buffer->getBufferStart() + loc.offset());
}

**Comment by Scheremo** [comment]
> Super cool! Thank you for rebasing it.
> 
> > where the first one is only to determine all module names - this is necessary, since we index only instantiated modules, and only top modules are instantiated. I tried looking, but didn't find a flag to treat all module definitions as top modules.
> 
> Does "all module names" here mean modules defined in the main file? I think this is ok if the modules not defined in the main file is not elaborated.
> 

That is a good point, and I adapted it. Only the main buffer file is parsed to figure out which top modules to compile the entire project for.

> > To prevent this I currently scan the command file and remove the main file (if present). I think we could prevent this if instead of matching raw pointers we matched against a file id / filename + offset. However, this might be slower than the current solution since it adds another layer of indirection.
> 
> I think we can tweak the `getSMLoc()` to return a pointer in main buffer if slang location points to the current main file by comparing a file path. Changing the key of IntervalMap to offset looks reasonable as well 👍

I think I'll leave changing the key of IntervalMap to a later day; I gave it some thought, and I think there are also issue with just offsets. We would really want a data structure as the key that normalizes both llvm source buffer IDs and slang buffer IDs over an offset. Seems non-obvious to me! Instead I fixed the current implementation by setting up all slang buffers as llvm membuffers as well; This is pessimistic since most might not be accessed very often, but the most stright forward way I could think of.

> 
> > I also extended the way we index by setting the top modules explicitly and only indexing the ones whose location is in the actual file that's open. Without this optimization, on a large project like the [snitch_cluster](https://github.com/pulp-platform/snitch_cluster) with a ~400 line file list the LSP takes ~30 seconds to index (on my MBP M4 Max). With this optimization I didn't manage to eyeball-measure the time it takes anymore as it's too small (<1s).
> 
> That's super cool and clever! Thank you for adding the optimization. Relevantly in the long term it would be really cool if we can store indexed data to disk like clangd https://clangd.llvm.org/design/indexing

Absolutely! 

I think I've addressed most or all of the points you raised, and I am happy with the state. Please let me know if you find other points we can improve, otherwise I am okay with merging :)

File: lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp [hunk 2]
```diff
@@ -248,15 +252,87 @@ void LSPDiagnosticClient::report(const slang::ReportedDiagnostic &slangDiag) {
...
+    if (cand == target)
+      return true;
+  }
+  return false;
+}
+
+static inline bool mainBufferFileInCommandFileList(llvm::StringRef cmdfile,
+                                                   llvm::StringRef target_abs) {
+  return mainBufferFileInCommandFileList(cmdfile.str(), target_abs.str());
+}
```
uenoku: Can we create a helper function to weakly canonicalize + lexically_normal? Also there are a few LLVM functionalities so it would be great if we can adapt them (`StringRef::trim(), llvm::line_iterator, StringRef::starts_with`). Also can we inline the second `mainBufferFileInCommandFileList`? 
```suggestion

static std::filesystem::path canonicalizeFileName(const std::string &file) {
  std::error_code ec;
  std::filesystem::path path = std::filesystem::weakly_canonical(file, ec);
  if (ec)
    path = std::filesystem::absolute(file).lexically_normal();

  return path;
}

// Filter out the main buffer file from the command file list, if it is in
// there.
static inline bool
mainBufferFileInCommandFileList(const std::string &cmdfileStr,
                                const std::string &targetAbsStr) {

  std::filesystem::path cmdfile(cmdfileStr);
  std::filesystem::path targetAbs(targetAbsStr);
  const std::filesystem::path base = cmdfile.parent_path();

  std::string error;
  auto cmdFile = mlir::openInputFile(cmdfileStr, &error);
  if (!cmdFile) {
    circt::lsp::Logger::error(Twine("Failed to open command file ") +
                              cmdfileStr + ": " + error);
    return false;
  }

  // Read line by line, ignoring empty lines and comments.
  for (llvm::line_iterator i(*cmdFile); !i.is_at_eof(); ++i) {
    auto line = i->trim();
    // Check if this line is a path.
    if (line.empty())
      continue;
    static constexpr StringRef commandPrefixes[] = {"+incdir+", "+define+",
                                                    "-I", "-D"};
    auto isCommand = [&line](StringRef s) { return line.starts_with(s); };
    if (llvm::any_of(commandPrefixes, isCommand))
      continue;
    // Treat everything else as a path (relative entries resolved vs. cmdfile
    // dir).
    auto candRel = std::filesystem::path(line.str());
    auto candAbs = canonicalizeFileName(
        candRel.is_absolute() ? candRel : (base / candRel));
    if (candAbs == targetAbs)
      return true;
  }

  return false;
}
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -248,15 +252,87 @@ void LSPDiagnosticClient::report(const slang::ReportedDiagnostic &slangDiag) {
...
+    }
+    skipMainBufferSlangImport |=
+        mainBufferFileInCommandFileList(cmdFile, canonPath);
+  }
+
   if (auto memBufferOwn =
           llvm::MemoryBuffer::getMemBufferCopy(contents, uri.file())) {
 
-    bufferId = sourceMgr.AddNewSourceBuffer(std::move(memBufferOwn), SMLoc());
+    mainBufferId =
```
uenoku: super nit: can we declare `mainBufferId` just above? I thought maybe there is a code path that mainBufferId is not initialized.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -273,17 +349,58 @@ VerilogDocument::VerilogDocument(
...
+  if (failed(topCompilation))
+    return;
+
+  std::vector<std::string> topModules;
+  for (const auto *defs : (*topCompilation)->getDefinitions())
+    topModules.emplace_back(defs->name);
+
+  // Make sure that all possible definitions in the main buffer are
+  // topModules!
+  driver.options.topModules = topModules;
```
uenoku:
```suggestion
  driver.options.topModules = std::move(topModules);
```

uenoku: nit: could you put these parts into a block `{}` so that we can ensure that `topDriver`'s lifetime is not used anywherelse?  
```suggestion
  {
  slang::driver::Driver topDriver;

  auto topSlangBuffer =
      topDriver.sourceManager.assignText(uri.file(), memBuffer->getBuffer());
  topDriver.sourceLoader.addBuffer(topSlangBuffer);

  topDriver.options.compilationFlags.emplace(
      slang::ast::CompilationFlags::LintMode, false);
  topDriver.options.compilationFlags.emplace(
      slang::ast::CompilationFlags::DisableInstanceCaching, false);

  if (!topDriver.processOptions()) {
    return;
  }

  if (!topDriver.parseAllSources()) {
    circt::lsp::Logger::error(Twine("Failed to parse Verilog file ") +
                              uri.file());
    return;
  }

  FailureOr<std::unique_ptr<slang::ast::Compilation>> topCompilation =
      topDriver.createCompilation();
  if (failed(topCompilation))
    return;

  std::vector<std::string> topModules;
  for (const auto *defs : (*topCompilation)->getDefinitions())
    topModules.emplace_back(defs->name);

  // Make sure that all possible definitions in the main buffer are
  // topModules!
  driver.options.topModules = topModules;
  }
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -248,15 +252,87 @@ void LSPDiagnosticClient::report(const slang::ReportedDiagnostic &slangDiag) {
...
+  std::ifstream in(cmdfile);
+  if (!in)
+    return false;
+
+  std::string line;
+  while (std::getline(in, line)) {
+    std::string s = line;
+    s.erase(0, s.find_first_not_of(" \t\r\n")); // left-trim
+    if (!s.empty() &&
+        (s.rfind("+incdir+", 0) == 0 || s.rfind("+define+", 0) == 0 ||
```
uenoku: Maybe just checking `-`/`+` is sufficient and more robust?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -308,6 +425,50 @@ VerilogDocument::VerilogDocument(
   if (failed(compilation))
     return;
 
+  // If the main buffer is present in a command file, compile it only once
+  // and import directly from the command file; then figure out which buffer id
+  // it was assigned and bind to llvm source manager.
+  llvm::SmallString<256> slangCanonPath("");
```
uenoku:
```suggestion
  llvm::SmallString<256> slangCanonPath;
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -308,6 +425,50 @@ VerilogDocument::VerilogDocument(
...
+  std::unique_ptr<llvm::MemoryBuffer> newBuffer;
+  uint32_t newBufferId;
+
+  // Iterate through all buffers in the slang compilation and set up
+  // a binding to the LLVM Source Manager.
+  auto *sourceManager = (**compilation).getSourceManager();
+  for (auto slangBuffer : sourceManager->getAllBuffers()) {
+    slangRawPath = sourceManager->getRawFileName(slangBuffer);
+    if (std::error_code ec =
+            llvm::sys::fs::real_path(slangRawPath, slangCanonPath))
```
uenoku, line 439: super nit: please inline `slangRawPath` if not used anywhere else. 
```suggestion
            llvm::sys::fs::real_path(sourceManager->getRawFileName(slangBuffer), slangCanonPath))
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Good catch!

Scheremo: Fine with me!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: No strong opinions from my end. I think it's a pretty closed specification (`-I` / `+incdir+` are the industry standards), but we might also change it.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Done!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Sure! Done.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Moved mainBufferId declaration + initialization down. Made it const for good measure.

Scheremo:
Hi Hideto, thanks for the comprehensive and constructive feedback! I think I addressed all of it, but please let me know if I missed any or there is further. Otherwise I am okay with merging 😄 

EDIT: I also agree with eventually using only slang buffers, rather than LLVM Source buffers + Slang buffers. Should make life easier and tooling faster 😃

uenoku:
Cool, thank you for updating this! Looks great! I'm going to merge :)

PR #9004, opened 2025-09-23
[Reduce] Various reduction tweaks

fabianschuiki:
Going to land this as a rebase merge to keep the tweaks in separate commits. These help reduce complex OFIRRTL inputs significantly.


fabianschuiki:
Landing this since the tweaks are generally useful and easy to revert/adjust.

PR #9007, opened 2025-09-24
[FIRRTL] Ensure all types and attributes are walkable

fabianschuiki:
Upstream MLIR requires type and attribute storages to define a `getAsKey` function, and for the return type to be something that implements the corresponding attribute/type walk. Most of our FIRRTL types and attributes predate these requirements. This PR is mostly busy work, adding `getAsKey` to the types and attributes that don't yet have it, and changing `std::pair` to `std::tuple`. (For some reason upstream type/attribute walking works on tuples, but not pairs.) Also add unit tests for attributes and types that ensure the walks actually work.

These tweaks allow circt-reduce to be a lot more clever about which ops it touches, because it can see a lot more symbol uses now. I also expect the `InnerSymbolDCE` pass to potentially benefit from this, at least if inner symbols are weirdly nested within FIRRTL attributes and types.

Most of this was done by Claude Sonnet 4, with a lot of handholding.


lib/Dialect/FIRRTL/FIRRTLTypes.cpp
```diff
@@ -1438,15 +1437,15 @@ struct circt::firrtl::detail::BundleTypeStorage
   KeyTy getAsKey() const { return KeyTy(elements, isConst); }
 
   static llvm::hash_code hashKey(const KeyTy &key) {
-    return llvm::hash_combine(
-        llvm::hash_combine_range(key.first.begin(), key.first.end()),
-        key.second);
+    return llvm::hash_combine(llvm::hash_combine_range(std::get<0>(key).begin(),
+                                                       std::get<0>(key).end()),
+                              std::get<1>(key));
```
seldridge: Consider use of a temporary for duplicate `std::get<0>(key)` calls.

seldridge: Ditto throughout / throw this at the LLM and see what it does.

unittests/Dialect/FIRRTL/AttributesTest.cpp
seldridge, line 1: Lightly reviewed. This looks good.

unittests/Dialect/FIRRTL/TypesTest.cpp
```diff
@@ -47,4 +50,399 @@ TEST(TypesTest, TypeAliasCast) {
...
+
+class TypeWalkTest : public ::testing::Test {
+protected:
+  void SetUp() override { context.loadDialect<FIRRTLDialect>(); }
+
+  MLIRContext context;
+  Builder builder{&context};
+};
+
+TEST_F(TypeWalkTest, BundleTypeWalk) {
```
seldridge, line 65: This kind of test could be golfed down to a single field. Not critical.

unittests/Dialect/FIRRTL/TypesTest.cpp
seldridge, line 1: Only lightly reviewed.

seldridge (approved):
Only lightly reviewed. Looks good.

seldridge:
Also, thank you for figuring out that this was the issue and fixing this!

lib/Dialect/FIRRTL/FIRRTLTypes.cpp
```diff
@@ -1438,15 +1437,15 @@ struct circt::firrtl::detail::BundleTypeStorage
   KeyTy getAsKey() const { return KeyTy(elements, isConst); }
 
   static llvm::hash_code hashKey(const KeyTy &key) {
-    return llvm::hash_combine(
-        llvm::hash_combine_range(key.first.begin(), key.first.end()),
-        key.second);
+    return llvm::hash_combine(llvm::hash_combine_range(std::get<0>(key).begin(),
```
uenoku: Maybe it works? 
```suggestion
    return llvm::hash_value(key);
```

fabianschuiki: Excellent point, thanks! :raised_hands:

PR #9009, opened 2025-09-24
[ESI] Route simulator compilation/exec output through callbacks

mortbopet:
To better support "library mode"-usage of these scripts, we need a way for users to get a hold of the stdout/stderr from the compilation and simulation parts of the Simulator class.

This PR changes the compile/sim logging to be performed through a callback, as opposed to directly to a file descriptor. If no callbacks are provided by the user during Simulator construction, the default (file-based) logging will be used.

### DIFF
--- a/lib/Dialect/ESI/runtime/python/esiaccel/cosim/verilator.py
+++ b/lib/Dialect/ESI/runtime/python/esiaccel/cosim/verilator.py
@@ -4,7 +4,7 @@
 
 import os
 from pathlib import Path
-from typing import List
+from typing import List, Optional, Callable
 
 from .simulator import CosimCollateralDir, Simulator, SourceFiles
 
@@ -14,10 +14,28 @@ class Verilator(Simulator):
 
   DefaultDriver = CosimCollateralDir / "driver.cpp"
 
-  def compile_commands(self) -> List[List[str]]:
+  def __init__(self,
+               sources: SourceFiles,
+               run_dir: Path,
+               debug: bool,
+               run_stdout_callback: Optional[Callable[[str], None]] = None,
+               run_stderr_callback: Optional[Callable[[str], None]] = None,
+               compile_stdout_callback: Optional[Callable[[str], None]] = None,
+               compile_stderr_callback: Optional[Callable[[str], None]] = None,
+               make_default_logs: bool = True):
+    super().__init__(sources=sources,
+                     run_dir=run_dir,
+                     debug=debug,
+                     run_stdout_callback=run_stdout_callback,
+                     run_stderr_callback=run_stderr_callback,
+                     compile_stdout_callback=compile_stdout_callback,
+                     compile_stderr_callback=compile_stderr_callback,
+                     make_default_logs=make_default_logs)
     self.verilator = "verilator"
     if "VERILATOR_PATH" in os.environ:
       self.verilator = os.environ["VERILATOR_PATH"]
+
+  def compile_commands(self) -> List[List[str]]:
     cmd: List[str] = [
         self.verilator,
         "--cc",


lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
```diff
@@ -93,11 +94,66 @@ class Simulator:
...
 
+    # Install callbacks (possibly wrapping default file log writers).
+    self._run_stdout_cb = run_stdout_callback
+    self._run_stderr_cb = run_stderr_callback
+    self._compile_stdout_cb = compile_stdout_callback
+    self._compile_stderr_cb = compile_stderr_callback
+
+    self._default_files: List[IO[str]] = []
+
+    def _ensure_default_log(cb_attr: str, log_attr: str, filename: str):
```
teqdruid: I don't like the usage of `setattr` and `getattr` here. It's something to be avoided unless it's necessary or saves a lot of code.

Suggested refactor:
`(self._run_stdout_cb, self._run_stdout_log) = ensure_default(run_stdout_callback, 'sim_stdout.log')`

and necessary changes to `_ensure_default`.

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
```diff
@@ -116,23 +172,28 @@ def compile_commands(self) -> List[List[str]]:
...
+          wait=True)
+      if isinstance(ret, int) and ret != 0:
+        print("====== Compilation failure")
+
+        # If we have the default file loggers, print the compilation logs to
+        # console. Else, assume that the user has already captured them.
+        if self.UsesStderr:
+          if hasattr(self, "_compile_stderr_log"):
+            self._compile_stderr_log.seek(0)
+            print(self._compile_stderr_log.read())
```
teqdruid, line 205: As long as we're reading from the file, is there a way to stream the file directly to stdout (instead of reading the whole file into memory and then printing it)?

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
```diff
@@ -198,6 +259,54 @@ def run_proc(self, gui: bool = False) -> SimProcess:
...
+      if pipe is None:
+        return
+      for raw in pipe:
+        if raw.endswith('\n'):
+          raw = raw[:-1]
+        if cb:
+          try:
+            cb(raw)
+          except Exception:
+            # Swallow callback exceptions to avoid killing streaming threads.
```
teqdruid: I don't think it's ever a good idea to just swallow exceptions. Maybe output it via the callback?

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
```diff
@@ -161,14 +223,13 @@ def run_proc(self, gui: bool = False) -> SimProcess:
...
-    simProc = subprocess.Popen(self.run_command(gui),
-                               stdout=simStdout,
-                               stderr=simStderr,
-                               env=simEnv,
-                               cwd=self.run_dir,
-                               preexec_fn=os.setsid)
-    simStderr.close()
-    simStdout.close()
+    rcmd = self.run_command(gui)
+    simProc = self._start_process_with_callbacks(rcmd,
```
teqdruid: Where do you join with the two threads you started and didn't join with because `wait=False`?

lib/Dialect/ESI/runtime/python/esiaccel/cosim/verilator.py
```diff
@@ -14,14 +14,10 @@ class Verilator(Simulator):
 
   DefaultDriver = CosimCollateralDir / "driver.cpp"
 
-  def __init__(self, sources: SourceFiles, run_dir: Path, debug: bool):
-    super().__init__(sources, run_dir, debug)
-
+  def compile_commands(self) -> List[List[str]]:
     self.verilator = "verilator"
```
teqdruid, line 34: Why'd you move this? I like to set up everything in constructors so that if they happen to get used in other places they can be. Set up shouldn't be a side-effect of a function not called by the constructor. If you're gonna move it, it should be a local variable.

teqdruid (changes requested):

lib/Dialect/ESI/runtime/python/esiaccel/cosim/verilator.py
mortbopet: To avoid complicating the constructor (i.e. duplicating the `Simulator` constructor in `Verilator`). But i get your point - I'll reconsider!

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
mortbopet: As in, stream it and print it during compilation? If that's the behavior we want, that could just be added to the logging callback for when the user didn't provide a logging callback.

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
mortbopet: I don't think it's a good idea to do that, when the callback has just demonstrated that it's capable of exceptional behavior. We could add a regular print, or go nuclear and `sys.exit()`... regular print for now!

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
mortbopet: Nothing joins with the threads. Those threads have two ways of terminating; either when the pipe closes (i.e. the `for` loop terminates), or when the python program terminates (`daemon=true`). The thought was that the first condition would suffice, that is, tying the lifetime of the threads to the underlying simulator process.

However, i _can_ imagine a race condition, where the underlying simulator process has terminated, and thus the callback endpoint may also have terminated/destructed, but the thread is about to call the callback, which could result in unintentional behavior. I'll modify s.t. these threads are joined on force_stop.

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
teqdruid: No, I meant this line and the one above it. I'm assuming that `read()` reads the whole file into memory, though it may be more intelligent than that.

But this is very minor.

lib/Dialect/ESI/runtime/python/esiaccel/cosim/verilator.py
teqdruid: You *could* pass all of the args by `kwargs` but that fucks with code completion so I prefer the solution you've done here.

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
teqdruid: Yeah, that works too.

teqdruid (approved):

lib/Dialect/ESI/runtime/python/esiaccel/cosim/simulator.py
teqdruid: There is another race condition wherein the simulator process has stopped but the threads have not finished processing the input. esi-cosim then exits without the callbacks having processed all of the inputs. This is actually plausible if the simulation outputs a bunch of text at the very end of the sim.

The other reason to `join()` on all threads is that we probably want to add a feature wherein the callbacks detect a failure. We want this to change the exit code but don't want the callbacks to terminate immediately by calling `sys.exit(1)`.

Issue #8994, opened 2025-09-19
[circt] LowerMemory might not update module hierarchy properly


tianrui-wei:
When running the latest firtool with both `--export-module-hierarchy` and `--repl-seq-mem` flags, the sram names transformation might only happen in one place. 

This manifests as in the smems.conf, there is only `name rob_debug_inst_mem_ext ...`, but in the model hierarchy json, the module only exists as `rob_debug_inst_mem`.

This can be verified as running the following commands in the chipyard repo at main branch after initializing the repo with https://chipyard.readthedocs.io/en/main/Chipyard-Basics/Initial-Repo-Setup.html.

```
make verilog USE_CHISEL7=1 CONFIG=MediumBoomV3Config

```

Note the curious thing is that this works for the first 17 sram transformation, just not for the last one. Let me know what I can help to diagnose this issue.

tianrui-wei:
edit: the bug turned out to occur in conjunction with `sifive.enterprise.firrtl.ModuleHierarchyAnnotation`, will work with chisel folks to further diagnose the issue. Closing for now.

tianrui-wei:
Reopening this issue after conversation with @jackkoenig. This seems to be a bug in the firtool. The issue can be reproduced by running `run.sh` in the attachment. The bug, mentioned above, manifests as the `smem` instance of `ebtb` gets transformed properly, but the `rob_debug_inst_mem` does not gets renamed properly in the module hierarchy json file.

[repro.tar.gz](https://github.com/user-attachments/files/22456722/repro.tar.gz)

PR #9012, opened 2025-09-24
fix: use NLA to preserve externalized memory from being dropped

tianrui-wei:
This commit fixes a corner-case, where a memory module itself is not effectful, but it is connected to an effectful *ext* memory module. If we simply mark all memory modules as pure, this would optimise out the ext memory module, causing inconsistencies in the sifive -repl-seq-mem flow. The previous firrtl transforms and hierarchies would be generated correctly, yet the final _ext sram would be omitted completely.

This fixes https://github.com/llvm/circt/issues/8994, where the offending chisel looks like

```scala
  val rob_debug_inst_mem   = SyncReadMem(numRobRows, Vec(coreWidth, UInt(32.W)))
  val rob_debug_inst_wmask = WireInit(VecInit(0.U(coreWidth.W).asBools))
  val rob_debug_inst_wdata = Wire(Vec(coreWidth, UInt(32.W)))
  rob_debug_inst_mem.write(rob_tail, rob_debug_inst_wdata, rob_debug_inst_wmask)
  val rob_debug_inst_rdata = rob_debug_inst_mem.read(rob_head, will_commit.reduce(_||_))
```

Before this patch, the following would have all the wires optimised out by this pass under rob_debug_inst_mem

```
 @rob_debug_inst_mem(in %R0_addr: !firrtl.uint<5>, in %R0_en: !firrtl.uint<1>, in %R0_clk: !firrtl.clock, in %W0_addr: !firrtl.uint<5>, in %W0_clk: !firrtl.clock, in %W0_data: !firrtl.uint<64>, in %W0_mask: !
 firrtl.uint<2>) {
     %c1_ui1 = firrtl.constant 1 : !firrtl.uint<1>
     %rob_debug_inst_mem_ext_R0_addr, %rob_debug_inst_mem_ext_R0_en, rob_debug_inst_mem_ext_R0_clk, %rob_debug_inst_mem_ext_R0_data, rob_debug_inst_mem_ext_W0_addr, %rob_debug_inst_mem_ext_W0_en,
 %rob_debug_inst_mem_ext_W0_clk, %rob_debug_inst_mem_ext_W0_data, %rob_debug_inst_mem_ext_W0_mask = firrtl.instance rob_debug_inst_mem_ext @rob_debug_inst_mem_ext(in R0_addr: !firrtl.uint<5>, in R0_en: !
 firrtl.uint<1>, in R0_clk: !firrtl.clock, out R0_data: !firrtl.uint<64>, in W0_addr: !firrtl.uint<5>, in W0_en: !firrtl.uint<1>, in W0_clk: !firrtl.clock, in W0_data: !firrtl.uint<64>, in W0_mask: !
 firrtl.uint<2>)
     firrtl.matchingconnect %rob_debug_inst_mem_ext_R0_addr, %R0_addr : !firrtl.uint<5>
     firrtl.matchingconnect %rob_debug_inst_mem_ext_R0_en, %R0_en : !firrtl.uint<1>
     firrtl.matchingconnect %rob_debug_inst_mem_ext_R0_clk, %R0_clk : !firrtl.clock
     firrtl.matchingconnect %rob_debug_inst_mem_ext_W0_addr, %W0_addr : !firrtl.uint<5>
     firrtl.matchingconnect %rob_debug_inst_mem_ext_W0_en, %c1_ui1 : !firrtl.uint<1>
     firrtl.matchingconnect %rob_debug_inst_mem_ext_W0_clk, %W0_clk : !firrtl.clock
     firrtl.matchingconnect %rob_debug_inst_mem_ext_W0_data, %W0_data : !firrtl.uint<64>
     firrtl.matchingconnect %rob_debug_inst_mem_ext_W0_mask, %W0_mask : !firrtl.uint<2>
   }
```

Open to suggestions on how to mitigate this issue in a potentially cleaner manner :)

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LayerSink.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LayerSink.cpp
@@ -13,6 +13,7 @@
 #include "circt/Dialect/FIRRTL/FIRRTLInstanceGraph.h"
 #include "circt/Dialect/FIRRTL/FIRRTLOps.h"
 #include "circt/Dialect/FIRRTL/Passes.h"
+#include "circt/Dialect/HW/HWOps.h"
 #include "circt/Support/Debug.h"
 #include "mlir/IR/Dominance.h"
 #include "mlir/IR/Threading.h"
@@ -33,6 +34,9 @@ namespace firrtl {
 using namespace circt;
 using namespace firrtl;
 using namespace mlir;
+
+static constexpr llvm::StringLiteral
+    kExtMemoryEffectNLAAttrName("circt.layerSink.externalMemoryEffect");
 
 //===----------------------------------------------------------------------===//
 // Helpers
@@ -87,7 +91,10 @@ static bool cloneable(Operation *op) {
 namespace {
 class EffectInfo {
 public:
-  EffectInfo(CircuitOp circuit, InstanceGraph &instanceGraph) {
+  EffectInfo(CircuitOp circuit, InstanceGraph &instanceGraph,
+             DenseSet<StringAttr> explicitEffectfulModules)
+      : explicitEffectfulModules(std::move(explicitEffectfulModules)) {
+    DenseSet<InstanceGraphNode *> visited;
     instanceGraph.walkPostOrder(
         [&](auto &node) { update(node.getModule().getOperation()); });
   }
@@ -111,6 +118,7 @@ public:
   }
 
 private:
+  DenseSet<StringAttr> explicitEffectfulModules;
   /// Record whether the module contains any effectful ops.
   void update(FModuleOp moduleOp) {
     moduleOp.getBodyBlock()->walk([&](Operation *op) {
@@ -129,6 +137,14 @@ private:
     if (auto annos = getAnnotationsIfPresent(moduleOp))
       if (!annos.empty())
         return markEffectful(moduleOp);
+
+    if (explicitEffectfulModules.contains(moduleOp.getModuleNameAttr()))
+      return markEffectful(moduleOp);
+
+    // Treat generated memory modules as effectful since they model stateful
+    // hardware even though they have no body.
+    if (isa<FMemModuleOp>(moduleOp.getOperation()))
+      return markEffectful(moduleOp);
 
     for (auto annos : moduleOp.getPortAnnotations())
       if (!cast<ArrayAttr>(annos).empty())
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LayerSink.cpp
```diff
@@ -138,9 +138,9 @@ class EffectInfo {
     // Regular modules may be pure.
     if (auto m = dyn_cast<FModuleOp>(op))
       return update(m);
-    // Memory modules are pure.
-    if (auto m = dyn_cast<FMemModuleOp>(op))
-      return;
+//    // Memory modules are pure.
+//    if (auto m = dyn_cast<FMemModuleOp>(op))
+//      return;
```
seldridge: High level: commented out code should never be committed.

seldridge (changes requested):
Thanks for jumping in with a contribution!

This doesn't seem quire right to me. That said, the change to change how `LayerSink` works helps highlight what the problem is: memories that are sunk after going through the `-repl-seq-mem` flow is causing incorrect information in the files produce by `-repl-seq-mem`.

What test should be introduced to check the correctness of this? Something that is `firtool` end-to-end is fine. I expect it's something like a memory whose only users are in a layer block. This should then be able to show that one of the metadata files is wrong. Having this test/minimal example will help guide what the solution should look like.

I think there are three alternatives:

1. Metadata needs to be encoded in such a way that it works across relocations that happen after `LayerSink`.
2. The presence of metadata (likely evidenced by inner symbols on things) should prevent `LayerSink` from moving things. There's something weird going on here where things are moving when they shouldn't.
3. The metadata formats should be dropped entirely and an alternative like the use of the Python APIs should be used instead.

lib/Dialect/FIRRTL/Transforms/LayerSink.cpp
tianrui-wei: Thanks, updated to a newer patch :)

tianrui-wei:
@seldridge Hi Schuyler, thanks for your feedback for my patch, both on discord and github. I've rewritten the patch to use non-local annotations to mark the lowered memory modules, and it is currently working for me locally (as well as passing the tests). Please let me know if you have any feedback on this :)

tianrui-wei:
force-pushed to comply with clang-format rules

PR #9015, opened 2025-09-24
[Sim] Rename FormatLitOp to FormatLiteralOp, NFC

fzi-hielscher:
Just a mechanical replacement.

I have grown up with DOS, so my brain is wired to not use more than three characters after a period. While the additional 4 bytes per literal op will no doubt hurt when storing large `.MLI` files on floppies, I am willing to make that sacrifice for the sake of clarity and consistency.


fabianschuiki (approved):
Nice, thanks! This has been triggering my OCD in the past, too 😁

PR #9016, opened 2025-09-24
[FIRRTL] Update more symbol-sensitive ops after dedup

fabianschuiki:
The Dedup pass currently handles module instances and class objects explictly in order to update the referenced module/class after deduplication. This does not consider types and attributes on any other operations, such as wires with a `!firrtl.class<...>` type. These also need to be updated.

Instead of adding more special cases, make use of the attribute/type walker and replacer infrastructure of MLIR that now works properly with FIRRTL types. When hashing a module, take note of all operations that have a symbol name nested somewhere in their attributes of types, and store them in a `symbolSensitiveOps` list as part of the module info. Once deduplication is done, visit those ops and replace all occurrences of outdated symbols with their post-dedup replacement. This can get rid of the special handling of ClassOp and ObjectOp, and also covers WireOp and attributes/types on any other op.

This came up as a dedup failure in a real-world design.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -84,6 +84,10 @@ struct ModuleInfo {
   std::array<uint8_t, 32> structuralHash;
   // Module names referred by instance op in the module.
   std::vector<StringAttr> referredModuleNames;
+  // The operations that contain references to symbols that may be changed by
+  // dedup. These need a fixup pass after dedup. This is just an optimization
+  // and does not factor into the hash or the equality between `ModuleInfo`s.
+  std::vector<Operation *> symbolSensitiveOps;
 };
 
 static bool operator==(const ModuleInfo &lhs, const ModuleInfo &rhs) {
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/Dedup.cpp
seldridge, line 1: The mutation of `hasSeenSymbol` is a bit difficult to follow. I was trying to think if there was a better way to handle this like adding a return value to the `update` methods. This is non-blocking and only for consideration.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -1482,76 +1547,10 @@ struct Deduper {
 // Fixup
 //===----------------------------------------------------------------------===//
 
-/// This fixes up ClassLikes with ClassType ports, when the classes have
-/// deduped. For each ClassType port, if the object reference being assigned is
-/// a different type, update the port type. Returns true if the ClassOp was
-/// updated and the associated ObjectOps should be updated.
```
seldridge, line 1488: Getting rid of this is great!

seldridge (approved):
This is _way_ cleaner! This is also future proofing this for any other ops which could have symbols in attributes as opposed to needing to keep doing fixups. Nice work.

fabianschuiki:
@mikeurbach pointed out that classes were supposed to be nominally typed, and we have added them to FIRRTL's dedup as a hack at some point. So instead of requiring class/object op fixups, these ops should never participate in dedup in the first place. I'm going to give that a spin in parallel.

fabianschuiki:
I think I'm going to land this, since it fixes bugs in the current implementation. We'll probably want an option on firtool to disable class deduplication entirely, such that various flows can migrate over gracefully.

PR #9021, opened 2025-09-25
[comb-to-synth] Implement Sklanskey Tree and Architecture Selection based on Attribute

cowardsa:
Introduce the Sklanskey Tree lowering option for `comb.add` and make it the default option (used in the PULP Platform library of arithmetic circuits). The Sklanskey Tree is as fast as a Kogge-Stone adder but tradesoff greater fan-out for less logic (preference will depend on target technology).

To aide testing purposes - make architecture selection optionally controlled by an attribute on the operator - in future this could be expanded to provide a way in which analysis passes could modify operator implementation selection, without interacting with the lowering pass itself. Open to suggestions if this is not a preferred method for passing "synthesis" controls?

```
comb.add %arg0, %arg1, %arg2 {synth.arch = "SKLANSKEY"} : i4
```

### DIFF
--- a/lib/Conversion/CombToSynth/CombToSynth.cpp
+++ b/lib/Conversion/CombToSynth/CombToSynth.cpp
@@ -497,11 +497,23 @@ struct CombAddOpConversion : OpConversionPattern<AddOp> {
           .Case("BRENT-KUNG", BrentKung)
           .Case("RIPPLE-CARRY", RippleCarry);
     }
-    // Determine based on width.
+    // Determine using width as a heuristic.
+    // TODO: Perform a more thorough analysis to motivate the choices or
+    // implement an adder synthesis algorithm to construct an optimal adder
+    // under the given timing constraints - see the work of Zimmermann
+
+    // For very small adders, overhead of a parallel prefix adder is likely not
+    // worth it.
     if (width < 8)
       return AdderArchitecture::RippleCarry;
+
+    // Sklanskey is a good compromise for high-performance, but has high fanout
+    // which may lead to wiring congestion for very large adders.
     if (width <= 32)
       return AdderArchitecture::Sklanskey;
+
+    // Kogge-Stone uses greater area than Sklanskey but has lower fanout thus
+    // may be preferable for larger adders.
     return AdderArchitecture::KoggeStone;
   }
 


lib/Conversion/CombToSynth/CombToSynth.cpp
```diff
@@ -480,7 +480,17 @@ struct CombAddOpConversion : OpConversionPattern<AddOp> {
       return success();
     }
```
uenoku, line 482: Could you add a enum and interpret string only once? e.g. 
```suggestion
enum AdderArch {
  RippleCarry,
  Sklansky,
  KoggeStone,
  BrentKung
};

static AdderArchitecture determineAdderArch(Operation* op, int64_t width) {
    auto strAttr = op->getAttrOfType<StringAttr>("synth.arch");
    if (strAttr) {
        return StringSwitch<AdderArch>(strAttr.getValue()).Case("SKLANSKEY", Sklansky).Case(...);
   }
   // Determine based on width.
   if (width < 8) 
     return AdderArchitecture::Ripple_Carry;
   if(width <= 32)
     return AdderArchitecture::Sklanskey;
   return AdderArchitecture::KoggeStone;
}

auto arch = determineAdderArch(op, width);
if (arch == AdderArchitecture::Ripple_Carry)
  return lowerRippleCarryAdder(...);
return lowerParallelPrefixAdder(...);
```

uenoku (approved):
LGTM, using attributes for testing is great idea! I want to make sure that these attributes are not part of public API so could you add explicitly include `test` as attribute name? like `synth.test.arch` maybe. I think it might make sense to eventually expose these architectures as a stand-alone operation `synth.add.sklanskey` if an user really wants to use specific architecture in a similar way to LLVM intrinsic.

lib/Conversion/CombToSynth/CombToSynth.cpp
```diff
@@ -480,17 +480,35 @@ struct CombAddOpConversion : OpConversionPattern<AddOp> {
...
+          .Case("BRENT-KUNG", BrentKung)
+          .Case("RIPPLE-CARRY", RippleCarry);
+    }
+    // Determine based on width.
+    if (width < 8)
+      return AdderArchitecture::RippleCarry;
+    if (width <= 32)
+      return AdderArchitecture::Sklanskey;
+    return AdderArchitecture::KoggeStone;
   }
```
uenoku: Could you write down comments for rational of heuristic here?

uenoku (approved):
LGTM! Thank you for making the change.

seldridge:
This is causing nightly debug builds to timeout when running the tests added to `comb-lowering-lec.mlir`. See: https://github.com/llvm/circt/actions/runs/18129142989/job/51613927639 The tests seem fine, but we may need to either disable them for debug builds or modify the lit timeout.

seldridge:
☝️ I bumped the timeout to fix this in: https://github.com/llvm/circt/commit/2bc2607a560b7be73e9d03003095a5a3e93b244a

cowardsa:
Thanks for catching this @seldridge - we could also split the test across multiple files to avoid the timeout issues - then can reduce the limit back down - would that be preferable?

PR #9026, opened 2025-09-26
[LTL][SVA] Add explict clocked operations and types

Clo91eaf:
- Introduced new clocked operations: ClockedAndOp, ClockedOrOp, ClockedIntersectOp, CreateClockedSequenceOp, ClockedDelayOp, ClockedConcatOp, and ClockedImplicationOp.
- Added LTLClockedSequenceType to represent clocked sequences.
- Updated LTLTypes and LTLVisitors to support new clocked operations.


Clo91eaf:
@fabianschuiki Sorry! Some things caused this PR to be late. Now I have time to implement #8673, wdyt about this PR?

include/circt/Dialect/LTL/LTLOps.td
```diff
@@ -61,10 +61,159 @@ def IntersectOp : AssocLTLOp<"intersect"> {
...
+  let assemblyFormat = [{
+    $edge $clock `,` $inputs attr-dict
+  }];
+  let hasCanonicalizeMethod = 1;
+
+  let summary = "A disjunction of explicitly clocked sequences with a specified clock domain.";
+  let description = [{
+    See `ltl.or` op. $edge and $clock specify the clock domain for the result.
+  }];
+}
```
fabianschuiki, line 163: Do and/or operations require a clock? These oparations sound like they only take two sequences and performa a boolean and/or on them. But if those sequences contain any delays, those delays would already specify a clock. Wouldn't that suffice?

include/circt/Dialect/LTL/LTLOps.td
```diff
@@ -61,10 +61,159 @@ def IntersectOp : AssocLTLOp<"intersect"> {
...
+    - `ltl.clocked_delay posedge %clk, %seq, 2, 0` creates a delay that matches
+      exactly 2 cycles after the current time on the positive edge of %clk.
+    - `ltl.clocked_delay posedge %clk, %seq, 2, 2` creates a delay that matches
+      2, 3, or 4 cycles after the current time.
+    - `ltl.clocked_delay posedge %clk, %seq, 2` creates an unbounded but finite
+      delay of 2 or more cycles.
+
+    This operation enables clean lowering of SVA expressions like `##1 b`
+  }];
+}
```
fabianschuiki, line 215: It's very cool that we get a properly clocked delay operation! Could we just name this `ltl.delay` and delete the old unclocked `ltl.delay`? I don't think we need both ops.

include/circt/Dialect/LTL/LTLOps.td
```diff
@@ -119,6 +268,24 @@ def PastOp : LTLOp<"past", [Pure]> {
...
+  }];
+  let hasFolder = 1;
+  let hasCanonicalizer = 1;
+
+  let summary = "Concatenate explicitly clocked sequences with a specified clock domain.";
+  let description = [{
+    See `ltl.concat` op. $edge and $clock specify the clock domain for the result.
+  }];
+}
+
```
fabianschuiki, line 288: Same as with the and/or operation: `ltl.concat` probably doesn't need to know what clock the input sequences are sensitive to. The concat corresponds to `a ##0 b` in Verilog, which means `a` and `b` happen at the exact same time, not a specific clocked delay apart. And since there's no delay, the op probably doesn't need a clock 🤔

include/circt/Dialect/LTL/LTLTypes.td
```diff
@@ -28,17 +28,41 @@ def LTLSequenceType : LTLTypeDef<"Sequence", "sequence"> {
...
+    This type is fundamentally clock-independent - it represents a quantified
+    statement about explicitly clocked sequences rather than being a clocked
+    entity itself. The "lifting" from clocked sequences to clock-agnostic
+    properties occurs via property operators like `ltl.implication`,
+    `ltl.always`, and `ltl.eventually`.
+
+    Properties can describe relationships between temporal patterns that may
+    exist in different clock domains, enabling cross-domain verification.
   }];
 }
```
fabianschuiki, line 67: Do we need to distinguish between these two? If we have all clocked operations (basically just delay, repeat, and some of the goto stuff I'd guess) have an explicit clock, and all other operations are clock agnostic (because they don't need a clock to express their semantics), I think all of that would just be a plain old `!ltl.property`. I don't think we need an explicit distinction betweem clocked and unclocked here 🤔. When we emit things to Verilog, we can just scan the LTL operations to find all the clocks used, and insert `@(posedge clock)` around all the LTL ops that share the same clock 🤔

include/circt/Dialect/LTL/LTLOps.td
Clo91eaf: > Do and/or operations require a clock? These oparations sound like they only take two sequences and performa a boolean and/or on them. But if those sequences contain any delays, those delays would already specify a clock. Wouldn't that suffice?

The clock in the operand is mainly used to specify the clock of the result type, because when two operands are in different clock domains, the clock of the result becomes difficult to determine

include/circt/Dialect/LTL/LTLOps.td
Clo91eaf: > It's very cool that we get a properly clocked delay operation! Could we just name this `ltl.delay` and delete the old unclocked `ltl.delay`? I don't think we need both ops.

Yes! This is a good suggestion, in fact I'm not sure if break changes should be made in this PR, if you think it's okay then it would be a good idea to replace the delay entirely

include/circt/Dialect/LTL/LTLTypes.td
Clo91eaf: > Do we need to distinguish between these two? If we have all clocked operations (basically just delay, repeat, and some of the goto stuff I'd guess) have an explicit clock, and all other operations are clock agnostic (because they don't need a clock to express their semantics), I think all of that would just be a plain old `!ltl.property`. I don't think we need an explicit distinction betweem clocked and unclocked here 🤔. When we emit things to Verilog, we can just scan the LTL operations to find all the clocks used, and insert `@(posedge clock)` around all the LTL ops that share the same clock 🤔

Yes, thank you for your suggestion. After thinking about it, I agree with your point. There are actually two core aspects here: the "clock of the operands" and "clocked delays". For operations like concat/and/or, if the operands are in different clock domains, the semantics become unclear, so these operations do not need to explicitly carry a clock—it's enough to ensure that all their input operands have the same clock.

Although the proposal in #8673 suggests enforcing clock consistency for operands at the type system level, the fundamental goal is to prevent combining operands from different clock domains. In other words, these logical/combination operations can be designed as clock-agnostic—as long as the type system guarantees that operands with different clocks cannot be mixed, the semantics remain clear.

Additionally, I will remove APIs such as clocked_and and clocked_or. Based on our discussion and the recommendations from #8673, as long as these logical operations ensure operand clocks are consistent, there's no need to explicitly distinguish "clocked" versions of and/or. This will make the API simpler and the type system clearer.

Thank you again for your insightful suggestions!

Clo91eaf:
I will make a new pr to make the `ltl.delay` fix complete soon

PR #9027, opened 2025-09-26
[Build] Set `/utf-8` flag specifically for slang targets

fzi-hielscher:
As an alternative to #8926, add the MSVC `/utf-8` flag only to CMake targets which include the slang headers.

I am not particularly confident that mixing translation units with and without this flag is not going to cause trouble. But for the moment it appears to be working in the UBTI CI build.

Closes #8927.


seldridge:
This looks great. I'm giving it a run on the UBTI static build here: https://github.com/llvm/circt/actions/runs/18042294978/job/51344181670 I'd like to use this for this release (adding `circt-verilog` as a published tool).

seldridge:
That UBTI static shouldn't be affected by this... UBTI closer to a Windows release is here: https://github.com/llvm/circt/actions/runs/18043093695/job/51347037515 I realize that this should be fine given that this is running in the pre-merge Windows job. I just want to make sure the release goes out fine.

fzi-hielscher:
I'm a bit lost regarding what workflow ends up where. But if we can get `circt-verilog` in the next release, that would be really awesome. Thanks!

seldridge (approved):
Nice, thanks for figuring this out. LGTM

This unblocks building `circt-verilog` in the release. 💯

PR #9030, opened 2025-09-26
fix: prevent canonicalize looping for extremely long time

tianrui-wei:
A hanging bug can manifest with very large inputs to minimize at canonicalize pass. This is because if the maxNumRewrites is not set, it defaults to kNoLimit, which could be an arbitrary large number. This optional command line argument makes it possible to set the max number of rewrites for large input files to minimize.

### DIFF
--- a/tools/circt-reduce/circt-reduce.cpp
+++ b/tools/circt-reduce/circt-reduce.cpp
@@ -104,6 +104,12 @@ static cl::opt<bool> verbose("v", cl::init(true),
                              cl::desc("Print reduction progress to stderr"),
                              cl::cat(mainCategory));
 
+cl::opt<int64_t> maxGreedyRewrites(
+    "max-greedy-rewrites", cl::init(-1),
+    cl::desc("Maximum number of rewrites GreedyPatternRewriteDriver may "
+             "apply (negative value keeps the default)"),
+    cl::cat(mainCategory));
+
 static cl::opt<unsigned>
     maxChunks("max-chunks", cl::init(0),
               cl::desc("Stop increasing granularity beyond this number of "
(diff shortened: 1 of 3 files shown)


tianrui-wei:
@fabianschuiki @seldridge from our discussions on discord.

tianrui-wei:
Rebased to origin/main, and fixed the CI (hopefully).

tianrui-wei:
Ran clang-format and pushed again.

tianrui-wei:
Sorry, fixed newer formatting errors... Could you rerun the CI again? :)

tools/circt-reduce/circt-reduce.cpp
```diff
@@ -104,6 +104,12 @@ static cl::opt<bool> verbose("v", cl::init(true),
                              cl::desc("Print reduction progress to stderr"),
                              cl::cat(mainCategory));
 
+cl::opt<int64_t> maxGreedyRewrites(
+    "max-greedy-rewrites", cl::init(-1),
+    cl::desc("Maximum number of rewrites GreedyPatternRewriteDriver may "
+             "apply (negative value keeps the default)"),
+    cl::cat(mainCategory));
```
seldridge: Nit: I'd just use "num" instead of "greedy" to match the `circt-opt` help text.

```suggestion
cl::opt<int64_t> maxNumRewrites(
    "max-num-rewrites", cl::init(-1),
    cl::desc("Maximum number of rewrites GreedyPatternRewriteDriver may "
             "apply (negative value keeps the default)"),
    cl::cat(mainCategory));
```

seldridge (approved):
Nice, this looks great. Thanks for iterating on a nice solution.

One nit related to the name of the option. One of can land this once that is cleaned up.

tools/circt-reduce/circt-reduce.cpp
tianrui-wei: Thanks Schuyler! I applied these changes locally and force-pushed again.

Issue #3836, opened 2022-09-07
Reconsider llvm_unreachable usage: audit, policy, alternatives


dtzSiFive:
`llvm_unreachable` in release builds instructs the compiler that it is, in fact, unreachable (UB if reached).  As a result, this can result in rather unexpected situations if it ever becomes reachable, causing compiler to omit bounds checks or always take a branch knowing it 'must' be true.

Recently this caused a user to be confused when presented with a diagnostic describing a problem obviously not relevant for the code in question, which is not a good experience to provide.

The justification for this behavior is performance, and while that is a serious concern we may wish to consider our usage to be sure this is what we want.

One approach is we require/prefer the use of the cmake option that converts these to a trap in Release (see below).  This is straightforward, hardens occurrences of this in LLVM itself (which personally I'm less concerned about), and easy to deploy.
Personally, I would prefer our code worked regardless of build option, especially those that are not the default.

If keeping these, perhaps we only use this in cases where it's very much guaranteed and ideally locally obvious that the path in question is and always will be unreachable (for example, after a covered switch statement, and only because our compiler warns if we miss one).  Basically, audit+policy(+ideally checked automatically, but don't have story there).

If not keeping these, they could be assertions instead with some code for that return path, or hopefully in many cases could be converted to emit proper diagnostics (perhaps we could have our own `llvm_unreachable` alternative that does this for us)?
(good parts of this idea should be credited to @fabianschuiki :smile:).

Discourse discussion about this for some more detail: https://discourse.llvm.org/t/llvm-unreachable-is-widely-misused/60587/23 .

LLVM CMake option for converting these to trap: https://github.com/llvm/llvm-project/commit/6316129e066e .
FWIW the IREE project replaced these in their code (see discourse for link, don't want to ping them here).

PR #9032, opened 2025-09-27
[FIRRTL] Make Domains ClassLike

seldridge:
Change domains to behave almost exactly like classes.  They now have
input and outputs (which can be used by their existing bodies).

This could alternatively be solved by using a simpler representation for
domains.  However, this would be better handled with a full cleanup of how
classes are modeled.

### DIFF
--- a/lib/Dialect/FIRRTL/FIRRTLOps.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOps.cpp
@@ -6909,6 +6909,96 @@ LogicalResult BindOp::verifyInnerRefs(hw::InnerRefNamespace &ns) {
 }
 
 //===----------------------------------------------------------------------===//
+// DomainOp
+//===----------------------------------------------------------------------===//
+
+void DomainOp::build(OpBuilder &builder, OperationState &result,
+                     StringAttr name, ArrayRef<PortInfo> ports) {
+  assert(
+      llvm::all_of(ports,
+                   [](const auto &port) { return port.annotations.empty(); }) &&
+      "domain ports may not have annotations");
+  buildClass<DomainOp>(builder, result, name, ports);
+
+  // Create a region and a block for the body.
+  auto *bodyRegion = result.regions[0].get();
+  Block *body = new Block();
+  bodyRegion->push_back(body);
+
+  // Add arguments to the body block.
+  for (auto &elt : ports)
+    body->addArgument(elt.type, elt.loc);
+}
+
+SmallVector<PortInfo> DomainOp::getPorts() {
+  return ::getPortImpl(cast<FModuleLike>((Operation *)*this));
+}
+
+void DomainOp::erasePorts(const llvm::BitVector &portIndices) {
+  ::erasePorts(cast<FModuleLike>((Operation *)*this), portIndices);
+  getBodyBlock()->eraseArguments(portIndices);
+}
+
+void DomainOp::insertPorts(ArrayRef<std::pair<unsigned, PortInfo>> ports) {
+  ::insertPorts(cast<FModuleLike>((Operation *)*this), ports);
+}
+
+Convention DomainOp::getConvention() { return Convention::Internal; }
+
+ConventionAttr DomainOp::getConventionAttr() {
+  return ConventionAttr::get(getContext(), getConvention());
+}
+
+ArrayAttr DomainOp::getParameters() { return {}; }
+
+ArrayAttr DomainOp::getPortAnnotationsAttr() {
+  return ArrayAttr::get(getContext(), {});
+}
+
+ArrayRef<Attribute> DomainOp::getPortAnnotations() { return {}; }
+
+void DomainOp::setPortAnnotationsAttr(ArrayAttr annotations) {
+  llvm_unreachable("domains do not support annotations");
+}
+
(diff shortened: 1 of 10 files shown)


lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -6908,6 +6908,96 @@ LogicalResult BindOp::verifyInnerRefs(hw::InnerRefNamespace &ns) {
...
+ArrayAttr DomainOp::getParameters() { return {}; }
+
+ArrayAttr DomainOp::getPortAnnotationsAttr() {
+  return ArrayAttr::get(getContext(), {});
+}
+
+ArrayRef<Attribute> DomainOp::getPortAnnotations() { return {}; }
+
+void DomainOp::setPortAnnotationsAttr(ArrayAttr annotations) {
+  llvm_unreachable("domains do not support annotations");
```
dtzSiFive, line 6963: Consider report fatal error instead?  I don't know why this would be unreachable.

cc https://github.com/llvm/circt/issues/3836 .

We don't want compilers being told this is unreachable, as that means anything reaching it is also "unreachable" and can to bizarre behaviors, like eliding branches/checks entirely knowing they lead to unreachable code.
OTOH that's exactly what it's good for.

I see that we do this for the ClassOp definitions, which hopefully is almost exactly a copy of :), so leaving it as-is is fine.

Since presumably it is presently unreachable.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -5745,18 +5745,31 @@ ParseResult FIRCircuitParser::parseDomain(CircuitOp circuit, unsigned indent) {
...
+
   auto builder = circuit.getBodyBuilder();
-  DomainOp::create(builder, info.getLoc(), name)
-      ->getRegion(0)
-      .push_back(new Block());
+  auto domainOp = DomainOp::create(builder, info.getLoc(), name, portList);
+  domainOp.setPublic();
+  deferredModules.emplace_back(DeferredModuleToParse{
+      domainOp, portLocs, getLexer().getCursor(), indent});
+  // domainOp->getRegion(0).push_back(new Block());
```
dtzSiFive: (I know this is draft, commented out code here FWIW)

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -307,11 +310,22 @@ LogicalResult LowerCircuit::lowerDomain(DomainOp op) {
...
   auto name = op.getNameAttr();
-  // TODO: Update this once DomainOps have properties.
-  auto classIn = ClassOp::create(builder, name, {});
+
+  // Create the new input class.  This is what the user will need to specify to
+  // evaluate OM. Clone the body of the domain into the class.
+  auto classIn = ClassOp::create(builder, name, op.getPorts());
+  builder.setInsertionPointToStart(classIn.getBodyBlock());
+  mlir::IRMapping mapper;
+  for (auto [dom, cls] : llvm::zip(op.getArguments(), classIn.getArguments()))
```
dtzSiFive: Probably we want to essentially always use `zip_equal`.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -307,11 +310,22 @@ LogicalResult LowerCircuit::lowerDomain(DomainOp op) {
   ImplicitLocOpBuilder builder(op.getLoc(), op);
   auto *context = op.getContext();
   auto name = op.getNameAttr();
-  // TODO: Update this once DomainOps have properties.
-  auto classIn = ClassOp::create(builder, name, {});
+
+  // Create the new input class.  This is what the user will need to specify to
```
dtzSiFive, line 498: Could the class just take the body as-is, instead of cloning?

.... do we even need to replace the domain with a class?

lib/Dialect/FIRRTL/Import/FIRParser.cpp
seldridge: Whoops. Fixed in: 283d2194f

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Fixed in: 283d2194f

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Ah, good point. This can just be `takeBody`. Switched in: 283d2194f

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: Yeah, this is all copied from the `ClassOp` definitions. I agree that this isn't great.

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
seldridge: Wrong commit. Fixed in: https://github.com/llvm/circt/pull/9032/commits/1d6774cd8cf928020bd1e59d078a1422c1c02fc4

Issue #8918, opened 2025-09-02
[Synth] Add folder/canonicalization to maj_inv


uenoku:
Implement local rewrite pattern based on MIG Boolean Algebra section in https://ieeexplore.ieee.org/document/7293649.
```
// Folder
maj_inv(x, !x, y) -> y
maj_inv(x, x, y) -> x
maj_inv(x, 0, 1) -> x

// Canonicalization
maj_inv(x, y, maj_inv(x, y, z)) ->(Distribute) maj(maj(x, y, x), maj(x, y, y), z) -> maj(x, y, z)
maj_inv(x, y, maj_inv(x, not y,z))  -> (Complementary Associativity)  maj_inv(x, y, maj_inv(x, x, z)) -> x
```

uenoku:
https://github.com/llvm/circt/pull/8959 implements first three patterns as canonicalization pattern. It would be also good to implement them as folder as well (but it may be restrictive due to inversion flag).

Abhiramjampani:
can i work on this @uenoku ?

markram1729:
Hi @uenoku  , 

Just want to confirm second statement in canonicalisation leads to x ?  
```suggestion
  maj_inv(x, y, maj_inv(x, not y,z))  -> (Complementary Associativity)  maj_inv(x, y, maj_inv(x, x, z)) -> maj_inv(x, y, x) -> x 
```

uenoku:
```mlir
$cat bar.mlir
 hw.module @c1(in %x: i1, in %y: i1, in %z: i1, out out: i1) {
    %0 = synth.mig.maj_inv %x, not %y, %z : i1
    %1 = synth.mig.maj_inv %x, %y, %0 : i1
    hw.output %1 : i1
  }

  hw.module @c2(in %x: i1, in %y: i1, in %z: i1, out out: i1) {
    hw.output %x : i1
  }

$circt-lec bar.mlir --c1 c1 --c2 c2

c1 == c2

```
Yes! Great catch.

markram1729:
Hey @uenoku , is this issue still relevant ?

uenoku:
Unfortunately it's not relevant at this point (and sorry that I had to drop your contribution). 
I think at some point i think I can introduce `synth.majority` which is semantically same but not focusing its homogeneous representation (=MIG, which is patented unfortunately). I'll reopen the issue once it's ready.

markram1729:
Sure @uenoku 
Thank you

PR #9033, opened 2025-09-27
[Synth ] feat : Fold Majority Graph

markram1729:
Resolves Patial #8918 
// Folder for 
maj_inv(x, !x, y) -> y
maj_inv(x, x, y) -> x
maj_inv(x, 0, 1) -> x


lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -67,12 +68,78 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
...
+              // auto value = cast<IntegerAttr>(getInputs()[i]).getValue();
+              //  return IntegerAttr::get(IntegerType::get(getContext(),
+              //  value.getBitWidth()),value);
+              return IntegerAttr::get(getType(), *c1);
+            }
+            // If constants are complementary, we can fold.
+            if (*c1 == ~*c2)
+              return getOperand(k);
+          }
+        }
```
uenoku: This block would be dead as two constants are already handled.

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -49,6 +49,7 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
   auto width = inputs[0].getBitWidth();
   APInt result(width, 0);
 
+  // is every input of same width ?
```
uenoku: yes

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -67,12 +68,78 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
...
+      inputValues.push_back(value);
+    }
+    i++;
+  }
+  if (!isOpConstant) {
+    // x 0 0
+    // x 1 1
+    if (inputValues.size() == 2) {
+      if (inputValues[0] != inputValues[1])
+        return {};
```
uenoku: I think it's pessimistic. If two constants are same you can return constants, if they are complement you can return third variable.

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -67,12 +68,78 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
...
+      for (int j = i + 1; j < 3; ++j) {
+        int k = 3 - (i + j);
+        assert(k >= 0 && k < 3);
+        // If we have two identical operands, we can fold.
+        if (getOperand(i) == getOperand(j)) {
+          // If they are inverted differently, we can fold to the third.
+          if (isInverted(i) != isInverted(j)) {
+            return getOperand(k);
+          }
+          return getOperand(i);
```
uenoku: This is a my bug but it's necessary to check if operand i is inverted. Could you fix it as well?

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -67,12 +68,78 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
...
+              //  value.getBitWidth()),value);
+              return IntegerAttr::get(getType(), *c1);
+            }
+            // If constants are complementary, we can fold.
+            if (*c1 == ~*c2)
+              return getOperand(k);
+          }
+        }
+      }
+    }
```
uenoku, line 117: you need to return {} here as evaluate expects same operands. Also `evaluate` expects uninverted values so the current code is not correct.

uenoku (commented):
Thanks!

lib/Dialect/Synth/SynthOps.cpp
markram1729: yes handled !

lib/Dialect/Synth/SynthOps.cpp
markram1729: okay!

lib/Dialect/Synth/SynthOps.cpp
markram1729: true , resolved !

lib/Dialect/Synth/SynthOps.cpp
markram1729: Hi @uenoku ,
if operand(i) is inverted , should i return {} ? 

I can't return with inverted operand(i)  , let canonicalise handle it !

lib/Dialect/Synth/SynthOps.cpp
markram1729: yes , it is handled now 

Thank you for pointing out

markram1729:
Hi @uenoku , Can you please check if this resolves the folding issue ?

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -39,9 +39,9 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
          "Number of inputs must match number of operands");
 
   if (inputs.size() == 3) {
```
uenoku, line 48: Could you keep the previous implementation? `evaluate` is meant to evaluate the op based on input values so it's necessary to look at inversion.

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -67,12 +67,60 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
...
-    inputValues.push_back(attr.getValue());
+    // x 0 0
+    // x 1 1
+    if (inputValues.size() == 2) {
+      if (inputValues[0] == inputValues[1])
+        return IntegerAttr::get(
+            IntegerType::get(getContext(), inputValues[0].getBitWidth()),
+            inputValues[0]);
+      else
+        return getOperand(index); // fix if it is a Inverted
```
uenoku: Could you check the inversion?

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -112,39 +160,6 @@ LogicalResult MajorityInverterOp::canonicalize(MajorityInverterOp op,
     return success();
   };
 
-  // Pattern match following cases:
-  // maj_inv(x, x, y) -> x
-  // maj_inv(x, y, not y) -> x
```
uenoku, line 124: Could you keep the canonicalization since folder is essentially a bit weaker?

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -67,12 +67,60 @@ llvm::APInt MajorityInverterOp::evaluate(ArrayRef<APInt> inputs) {
...
+  // x x 1 -> x
+  // x 1 1 -> 1
+  // x ~x 1 -> 1
+  // x 0 0 -> 0
+  // x 1 1 -> 1
+
+  bool isOpConstant = true;
+  // for all constant inputs
   SmallVector<APInt, 3> inputValues;
+  size_t i = 0, index;
```
uenoku: Could you do slightly change the code structure to store non-constant index and check based on the number constants? 
```suggestion
  SmallVector<APInt, 3> inputValues;
  SmallVector<unsigned, 3> nonConstantIndex;
  
 for (auto [i, input] : llvm::enumerate(adaptor.getInputs())) {
    auto attr = llvm::dyn_cast_or_null<IntegerAttr>(input);
    if (attr) 
        inputValues.push_back(attr);
     else 
       nonConstantIndex.push_back(i);
  }
  
  if (inputValues.size() == getNumINputs())
    return IntegerAttr::get(getType(), evaluate(inputValues));  
    
  if (getNumOperands() != 3) 
    return {};
  
 if (nonConstantIndex.size() == 1) {
    // Get two constants. 
    auto idx = nonConstantIndex.front();
    auto constIndex1 = (idx+1)%3;
    auto constIndex2 = (idx+2)%3;
    auto c1 = adaptor.getInputs()[constIndex1];
    auto c2 = ...
    auto ...;
 }
 
 
```

markram1729:
Hi @uenoku  , 
I have pushed the new changes , I  will  optimise it  more 

Can you suggest how can i handle ~x in fold 

any way i can return ~x as i can't negate a operand 
will canonicalise handle it ? I am returning as `return {}`

uenoku:
Really thank you for working hard on this :bow::bow:!

> Can you suggest how can i handle ~x in fold

Unfortunately we cannot handle it since it's not allowed to create a new operation: https://mlir.llvm.org/docs/Canonicalization/#canonicalizing-with-the-fold-method
So it's necessary to just give up the folding for such cases.

markram1729:
Hi @uenoku , can i handle in fold ?
```
 x x 1 -> x
 x ~x 1 -> 1
 ~x ~x 1 -> ~x
```
I have commented in the code ,as canoicalisation ishandling it

uenoku:
The first two can be implemented in a straightforward way. 

For the 3rd one, i realized maybe MLIR folder allows in-place mutation for the op itself, so maybe you can do something like
```
            (*this)->setOperands({getOperand(k)});
            (*this).setInverted({true});
            return getResult();
```

markram1729:
my apologies @uenoku , I was caught up with some work in past few weeks 
I will try to implement them  and  push the commits

markram1729:
Hey @uenoku  ,
With non constant values ==2 there is slight issue when handling 
```
x x const
~x ~x const 
```
at lower-word-to-bits.mlir test ,it  expects to generate statement  
```
    %5 = synth.mig.maj_inv %false, %2, %4 : i1
```
when i handle at fold implementation , it doesn't generate as shown in the testcase fail 
it would be good if it can be handled at canoicalisation itself ?

markram1729:
Hi @uenoku , 
for nonconstant values ==2  
I am not able to debug the root cause , maybe i should change test files  or code as explained in the previous comment

lib/Dialect/Synth/SynthOps.cpp
markram1729: it is handled now

markram1729:
Hey @uenoku  , I am closing this PR since this might have deprecated by now 
This PR Might be bloated and not solving all the cases in issue

uenoku:
I'm seriously sorry, I've missed notification somehow... Let me see if I can cherry-pick subset of folder.

markram1729:
Hey @uenoku , I will rebase and resolve if there are any conflicts 
i will push the changes 
Thank you for your response

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -68,15 +68,96 @@ OpFoldResult MajorityInverterOp::fold(FoldAdaptor adaptor) {
...
+    auto c1 = getConstant(i);
+    auto c2 = getConstant(j);
+    // x 0 0 -> 0
+    // x 1 1 -> 1
+    // x 0 ~0 -> x
+    // x 1 ~1 -> x
+    // x ~1 ~1 -> ~1 -> 0
+    // ~x 0 0 -> ~x  no fold
+    // ~x 0 ~1 -> 0
+    if (c1 == c2) {
```
uenoku, line 107: Inversion is resolved in constants, so you can simply return the constant

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -68,15 +68,96 @@ OpFoldResult MajorityInverterOp::fold(FoldAdaptor adaptor) {
...
+      if (isInverted(k)) {
+        (*this)->setOperands({getOperand(i)});
+        (*this).setInverted({true});
+        return getResult();
+      } else
+        return getOperand(k);
+    }
+  }
+  // else if (nonConstantValues.size() == 2) {
+  //   // x x 1 -> x
```
uenoku: Please remove these comments.

markram1729:
Hi @uenoku , resolved the comments

lib/Dialect/Synth/SynthOps.cpp
```diff
@@ -75,15 +75,53 @@ OpFoldResult MajorityInverterOp::fold(FoldAdaptor adaptor) {
...
+    auto c1 = getConstant(i);
+    auto c2 = getConstant(j);
+    // x 0 0 -> 0
+    // x 1 1 -> 1
+    // x 0 ~0 -> x
+    // x 1 ~1 -> x
+    // x ~1 ~1 -> ~1 -> 0
+    // ~x 0 0 -> ~x  no fold
+    // ~x 0 ~1 -> 0
+    if (c1 == c2) {
```
uenoku: Could you update the comment?

markram1729: just want to confirm , should i remove the comments ?

uenoku: I meant
```
// ~x 0 0 -> ~x 
```
is not correct transformation. Also we started doing following folding. 
```
// ~x 0 1 -> ~x
```

I think 3 line comment is more than enough. 
```
x c !c -> x
!x c !c -> !x
x c  c -> c
```

markram1729: Hey @uenoku , Resolved 
Thank you for your time

PR #9040, opened 2025-09-30
[FIRRTL] Gate class deduplication behind an option

fabianschuiki:
Add a `dedupClasses` option to the Dedup pass, enabled by default. When the option is enabled, classes are deduplicated like modules, which makes their names unstable. Some current flows that use firtool require this behavior. If the option is disabled, classes are considered nominally typed (which they are intended to be), and are not deduplicated at all.

Add a corresponding `--dedup-classes` option to firtool, also enabled by default. This allows flows to continue relying on the current class deduplication behavior, but they can opt into proper nominally typed classes as they see fit.

Eventually we'll want to remove these options and make classes *not* dedup unconditionally. @dtzSiFive calls this a compiler chicken bit :rofl:.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/Dedup.cpp
@@ -1674,6 +1674,8 @@ static void fixupSymbolSensitiveOps(
 
 namespace {
 class DedupPass : public circt::firrtl::impl::DedupBase<DedupPass> {
+  using DedupBase::DedupBase;
+
   void runOnOperation() override {
     auto *context = &getContext();
     auto circuit = getOperation();
@@ -1755,6 +1757,10 @@ class DedupPass : public circt::firrtl::impl::DedupBase<DedupPass> {
           // Only dedup extmodule's with defname.
           if (auto ext = dyn_cast<FExtModuleOp>(*module);
               ext && !ext.getDefname().has_value())
+            return success();
+
+          // Only dedup classes if enabled.
+          if (isa<ClassOp>(*module) && !dedupClasses)
             return success();
 
           StructuralHasher hasher(hasherConstants);
--- a/lib/Firtool/Firtool.cpp
+++ b/lib/Firtool/Firtool.cpp
@@ -100,8 +100,11 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
 
   pm.nest<firrtl::CircuitOp>().addPass(firrtl::createDropConst());
 
-  if (opt.shouldDedup())
-    pm.nest<firrtl::CircuitOp>().addPass(firrtl::createDedup());
+  if (opt.shouldDedup()) {
+    firrtl::DedupOptions opts;
+    opts.dedupClasses = opt.shouldDedupClasses();
+    pm.nest<firrtl::CircuitOp>().addPass(firrtl::createDedup(opts));
+  }
 
   if (opt.shouldConvertVecOfBundle()) {
     pm.addNestedPass<firrtl::CircuitOp>(firrtl::createLowerFIRRTLTypes(
@@ -551,6 +554,12 @@ struct FirtoolCmdOptions {
       llvm::cl::desc("Disable deduplication of structurally identical modules"),
       llvm::cl::init(false)};
 
+  llvm::cl::opt<bool> dedupClasses{
+      "dedup-classes",
+      llvm::cl::desc(
+          "Deduplicate FIRRTL classes, violating their nominal typing"),
+      llvm::cl::init(true)};
+
   llvm::cl::opt<firrtl::CompanionMode> companionMode{
       "grand-central-companion-mode",
       llvm::cl::desc("Specifies the handling of Grand Central companions"),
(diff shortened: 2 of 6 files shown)


dtzSiFive:
LGTM, but need to run so didn't vet all-the-things (sorry).  Woohoo!

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -279,7 +284,7 @@ struct StructuralHasher {
       // Skip names and annotations, except in certain cases.
       // Names of ports are load bearing for classes, so we do hash those.
       bool isClassPortNames =
-          isa<ClassLike>(op) && name == constants.portNamesAttr;
+          dedupClasses && isa<ClassLike>(op) && name == constants.portNamesAttr;
```
dtzSiFive: I'm wondering how/why we might visit a class operation here anyway?

Is there an idea we might have a class operation at some point that isn't top-level (or what the StructuralHasher is instantiated on)?  (I don't think so, but that's one argument for this...)

Otherwise, maybe the StructuralHasher changes aren't needed, just keep the part that doesn't bother hashing class operations.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
```diff
@@ -279,7 +284,7 @@ struct StructuralHasher {
       // Skip names and annotations, except in certain cases.
```
dtzSiFive, line 279: Stray comment while looking:

```
      value.walk([&](FlatSymbolRefAttr) { hasSeenSymbol = true; });
```

This attribute walking constructs a fresh AttrTypewalker which uses a DenseMap to cache the Attribute -> WalkResult which can be slow if done every time instead of reusing a walker created once.

The re-use might save re-walking the same attributes in order to determine if they have a symbol?

Anyway something to keep in mind/consider, not sure it's problematic as-is and forgive the comment unrelated to this PR.

lib/Firtool/Firtool.cpp
```diff
@@ -100,8 +100,11 @@ LogicalResult firtool::populateCHIRRTLToLowFIRRTL(mlir::PassManager &pm,
 
   pm.nest<firrtl::CircuitOp>().addPass(firrtl::createDropConst());
 
-  if (opt.shouldDedup())
-    pm.nest<firrtl::CircuitOp>().addPass(firrtl::createDedup());
+  if (opt.shouldDedup()) {
+    firrtl::DedupOptions opts;
+    opts.dedupClasses = opt.shouldDedupClasses();
+    pm.nest<firrtl::CircuitOp>().addPass(firrtl::createDedup(opts));
```
dtzSiFive, line 106: This could be `createDedup({opt.shouldDedupClasses()})`.  We handle this a few ways throughout, this way has benefit of making it super clear what's going on.  If we had field designators (c++20) we could keep that readability benefit.

Anyway, whatever you like better :+1:.

dtzSiFive (approved):
LGTM!

It seems we tend to prefer the opposite polarity for these options (false as default), but I think the idea is that we can make this default `false` later....?

Anyway, left some feedback please take a look! Thanks for pushing on this!

ItsHappening.gif

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
fabianschuiki: Hmmm that's a good point. There's probably a trade-off between creating a size-1 DenseMap very frequently, or having a DenseMap grow up to a size on the order of the number of distinct attributes in the IR, and continuously performing lookups into that. I have no idea which of these is faster. My intuition would be the size-1 DenseMap, but that's just a guess.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
fabianschuiki: But I think you're actually right, this doesn't need to be disabled. If the classes themselves don't dedup, their class names will always produce distinct hashes and prevent anything involving that class from deduping. Very insightful :1st_place_medal: :100: I'll tweak this.

lib/Firtool/Firtool.cpp
fabianschuiki: Yeah field designators would be nice. I like the explicit syntax, because it gives you a name to error out on in case you change the layout of `DedupOptions`, instead of having C++ hare-brained implicit casts try to cast your options to make them fit into the rearranged layout.

fabianschuiki:
> It seems we tend to prefer the opposite polarity for these options (false as default), but I think the idea is that we can make this default `false` later....?

Yeah exactly! The idea is to set this to false in a follow-up PR, maybe after a week or two, and then a bit later remove the option and class dedup altogether.

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
dtzSiFive: It definitely has been slow to the point of needing to be fixed before, but quite possibly different situation.
I don't think it's a size-1 DenseMap, as it recurses each attribute it finds goes into the cache.  But it's probably small, and your point remains.

lib/Firtool/Firtool.cpp
dtzSiFive: Absolutely.  Ship it!

lib/Dialect/FIRRTL/Transforms/Dedup.cpp
fabianschuiki: Oh yeah, right, I forgot that this actually visits all nested attributes. Good point.

PR #9044, opened 2025-10-02
[Comb][circt-synth] Implement BalanceMux pass for optimizing mux chains

uenoku:
This pass performs two main optimizations on mux chains: enhanced mux
chain folding that converts chains of muxes with index comparisons into
array operations, and priority encoder rebalancing that transforms linear
chains into balanced tree structures reducing depth from O(n) to O(log n).

Unlike the FIRRTL pipeline which avoids mux balancing to preserve design 
intent (https://github.com/chipsalliance/chisel/issues/1199), the synthesis 
pipeline implements these optimizations since it's focused on circuit 
optimization rather than maintaining the original design structure.

The pass handles both false-side and true-side chain patterns and includes
comprehensive test cases covering priority encoding, duplicate conditions,
and index comparison folding.


include/circt/Dialect/Comb/CombOps.h
```diff
@@ -86,6 +86,17 @@ Value createDynamicInject(OpBuilder &builder, Location loc, Value value,
...
+enum MuxChainWithComparisonFoldingStyle { None, BalancedMuxTree, ArrayGet };
+/// Mux chain folding that converts chains of muxes with index
+/// comparisons into array operations or balanced mux trees. `styleFn` is a
+/// callback that returns the desired folding style based on the index
+/// width and number of entries.
+bool foldMuxChainWithComparison(
+    PatternRewriter &rewriter, MuxOp rootMux, bool isFalseSide,
+    llvm::function_ref<MuxChainWithComparisonFoldingStyle(size_t indexWidth,
+                                                          size_t numEntries)>
+        styleFn);
```
fabianschuiki, line 99: Nice idea to have this as a utility on the Comb dialect 💯

lib/Dialect/Comb/Transforms/BalanceMux.cpp
```diff
@@ -0,0 +1,253 @@
...
+    for (auto &cond : conditions) {
+      cond = rewriter.createOrFold<comb::XorOp>(
+          op.getLoc(), cond,
+          rewriter.create<hw::ConstantOp>(op.getLoc(), APInt(1, 1)), true);
+    }
+  }
+
+  LDBG() << "Rebalanced priority mux with " << conditions.size()
+         << " conditions, using " << (useFalseSideChain ? "false" : "true")
+         << "-side chain.\n";
```
fabianschuiki, line 128: Never seen `LDBG()` before. Very handy! 😄

fabianschuiki (approved):
LGTM! Very useful pass to have 💯

cowardsa:
@uenoku - didn't see any timing guidance after a brief scan - did you plan to integrate the timing analysis into this pass? As the discussion you pointed to highlighted the dependence on arrival times for an optimal mux tree structure?

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -1973,13 +1974,17 @@ getMuxChainCondConstant(Value cond, Value indexValue, bool isInverted,
 /// Given a mux, check to see if the "on true" value (or "on false" value if
 /// isFalseSide=true) is a mux tree with the same condition.  This allows us
 /// to turn things like `mux(VAL == 0, A, (mux (VAL == 1), B, C))` into
-/// `array_get (array_create(A, B, C), VAL)` which is far more compact and
-/// allows synthesis tools to do more interesting optimizations.
+/// `array_get (array_create(A, B, C), VAL)` or a balanced mux tree which is far
+/// more compact and allows synthesis tools to do more interesting
+/// optimizations.
```
cowardsa, line 1979: In case of interest - Synopsys documentation favours priority encoders: [see slide 88](https://picture.iczhiku.com/resource/eetop/wyiEELySpYPudnxV.pdf)

lib/Dialect/Comb/CombFolds.cpp
```diff
@@ -2376,6 +2386,22 @@ struct MuxRewriter : public mlir::OpRewritePattern<MuxOp> {
...
+  // If the array is greater that 9 bits, it will take over 512 elements and
+  // it will be too large for a single expression.
+  if (indexWidth >= 9 || numEntries < 3)
+    return MuxChainWithComparisonFoldingStyle::None;
+
+  // Next we need to see if the values are dense-ish.  We don't want to have
+  // a tremendous number of replicated entries in the array.  Some sparsity is
+  // ok though, so we require the table to be at least 5/8 utilized.
+  uint64_t tableSize = 1ULL << indexWidth;
+  if (numEntries >= tableSize * 5 / 8)
```
cowardsa, line 2399: Nit: Interesting - is there any data to support this 5/8 ths? If not perhaps worth adding a todo to benchmark - someone might pick it up?

uenoku:
Yes the current code structure would not be too hard to integrate timing analysis. Will do that in a follow-up. For priority mux we can choose ideal separation points based on timing, for index comparison it's a bit trickier since it's necessary to know don't care bits of index.

lib/Dialect/Synth/Transforms/SynthesisPipeline.cpp
```diff
@@ -63,6 +64,10 @@ void circt::synth::buildCombLoweringPipeline(
   }
   pm.addPass(createCSEPass());
   pm.addPass(createSimpleCanonicalizerPass());
+  // Balance mux chains. For area oriented flow, we want to keep the mux chains
+  // unless they are very deep.
+  comb::BalanceMuxOptions balanceOptions{OptimizationStrategyTiming ? 16 : 64};
+  pm.addPass(comb::createBalanceMux(balanceOptions));
```
cowardsa, line 70: Nice!

test/Dialect/Comb/balance-mux.mlir
```diff
@@ -0,0 +1,141 @@
...
+  %c3_i4 = hw.constant 3 : i4
+  %c4_i4 = hw.constant 4 : i4
+  %c0_i32 = hw.constant 0 : i32
+  %c1_i32 = hw.constant 1 : i32
+  %c2_i32 = hw.constant 2 : i32
+  %c3_i32 = hw.constant 3 : i32
+  %c4_i32 = hw.constant 4 : i32
+
+  // Create a mux chain that compares index against constants
+  // This represents: if (index == 0) 0 else if (index == 1) 1 else if (index == 2) 2 else if (index == 3) 3 else 4
```
cowardsa: Nit: Perhaps could break over multiple lines to improve readability?

cowardsa (approved):
LGTM - nice work and certainly seems like a valuable addition to the synthesis pipeline.

Only potential more major suggestion - could add circt-lec tests and longest-path-analysis to check the reduction in levels is as expected i.e. O(n) -> O(log n)?

lib/Dialect/Comb/CombFolds.cpp
uenoku: This parameter was introduced here https://github.com/llvm/circt/commit/df37fa11c776879080b681b38653617c9f91e9c2 a while ago. Not sure the rational behind this and I think it's fine to change the behaivor. This is general Comb folder where creating a balance tree may not be always ideal (especially when emitting SystemVerilog, it would be easier for commertial tools to pattern match as well).

lib/Dialect/Comb/CombFolds.cpp
uenoku: Thanks, yeah. I wish there is a SV builtin for priority mux so that we can capture design intents. Maybe eventually we also want to introduce an operation for priority mux.

PR #9045, opened 2025-10-02
[circt-verilog-lsp] Move to Slang's src mngr, drop LLVM src mngr

Scheremo:
- **Use Slang buffers as source of truth**
  - Replace `llvm::SourceMgr`-centric code with direct use of `slang::SourceManager`.
  - Add `slang::BufferID mainBufferId` and expose `getMainBufferID()`.

- **Interval map now indexes Slang buffer pointers**
  - New typedef: `using SlangBufferPointer = char const *;`.
  - Change from `IntervalMap<const char*, …>` to `IntervalMap<SlangBufferPointer, …>` (half-open).
  - Lookups are computed from `slang::SourceManager::getSourceText(buf).data() + offset`.

- **Remove LLVM `SourceMgr` plumbing**
  - Drop members/APIs: `sourceMgr`, `bufferIDMap`, `getSMLoc(...)`.
  - Update `filePathMap` to store `(slang::BufferID, path)` pairs.
  - `getOrOpenFile(...)` now assigns text via Slang (`driver.sourceManager.assignText(...)`) and caches the returned `BufferID`.

- **Implement LSP UTF-16 to Slang UTF-8 conversion**
  - Add `decodeUtf8(...)` and `lspPositionToOffset(...)` helpers.
  - Use these helpers in `getLocationsOf(...)` and `findReferencesOf(...)` before interval lookups.
  - Clamp positions past end-of-line to buffer end; avoid OOB pointer math.

- **Definition & reference resolution updated**
  - `getLspLocation(slang::SourceLocation/Range)` now uses Slang line/column and compares `loc.buffer()` directly to `mainBufferId`.
  - `findReferencesOf(...)` / `getLocationsOf(...)` locate the interval by pointer into the **main** buffer.

- **Top module discovery via syntax trees**
  - Derive `driver.options.topModules` from `CompilationUnitSyntax` members by reading `ModuleDeclarationSyntax` headers (covers modules and packages).
  - Avoid creating a separate `Compilation` just to enumerate tops.

### DIFF
--- a/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
+++ b/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
@@ -33,11 +33,13 @@
 #include "slang/syntax/SyntaxTree.h"
 #include "slang/text/SourceLocation.h"
 #include "slang/text/SourceManager.h"
+#include "slang/util/Enum.h"
 #include "llvm/ADT/IntervalMap.h"
 #include "llvm/ADT/PointerUnion.h"
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringMap.h"
+#include "llvm/Support/ConvertUTF.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/LSP/Logging.h"
 #include "llvm/Support/LSP/Protocol.h"
@@ -104,13 +106,13 @@ public:
 
   VerilogDocument &getDocument() { return document; }
 
-  /// The type of interval map used to store source references. SMRange is
-  /// half-open, so we also need to use a half-open interval map.
-  using MapT =
-      llvm::IntervalMap<const char *, VerilogIndexSymbol,
-                        llvm::IntervalMapImpl::NodeSizer<
-                            const char *, VerilogIndexSymbol>::LeafSize,
-                        llvm::IntervalMapHalfOpenInfo<const char *>>;
+  // Half open interval map containing paired pointers into Slang's buffer
+  using SlangBufferPointer = char const *;
+  using MapT = llvm::IntervalMap<
+      SlangBufferPointer, VerilogIndexSymbol,
+      llvm::IntervalMapImpl::NodeSizer<SlangBufferPointer,
+                                       VerilogIndexSymbol>::LeafSize,
+      llvm::IntervalMapHalfOpenInfo<const SlangBufferPointer>>;
 
   MapT &getIntervalMap() { return intervalMap; }
 
@@ -158,11 +160,6 @@ public:
 
   const llvm::lsp::URIForFile &getURI() const { return uri; }
 
-  llvm::SourceMgr &getSourceMgr() { return sourceMgr; }
-  llvm::SmallDenseMap<uint32_t, uint32_t> &getBufferIDMap() {
-    return bufferIDMap;
-  }
-
   const slang::SourceManager &getSlangSourceManager() const {
     return driver.sourceManager;
   }
@@ -171,8 +168,7 @@ public:
   llvm::lsp::Location getLspLocation(slang::SourceLocation loc) const;
   llvm::lsp::Location getLspLocation(slang::SourceRange range) const;
 
-  // Return SMLoc from slang location.
-  llvm::SMLoc getSMLoc(slang::SourceLocation loc);
+  slang::BufferID getMainBufferID() const { return mainBufferId; }
 
   //===--------------------------------------------------------------------===//
   // Definitions and References
@@ -186,19 +182,20 @@ public:
                         const llvm::lsp::Position &pos,
                         std::vector<llvm::lsp::Location> &references);
 
+  std::optional<uint32_t> lspPositionToOffset(const llvm::lsp::Position &pos);
+  const char *getPointerFor(const llvm::lsp::Position &pos);
+
 private:
-  std::optional<std::pair<uint32_t, SmallString<128>>>
+  std::optional<std::pair<slang::BufferID, SmallString<128>>>
   getOrOpenFile(StringRef filePath);
 
   VerilogServerContext &globalContext;
 
-  // A map from slang buffer ID to the corresponding buffer ID in the LLVM
-  // source manager.
-  llvm::SmallDenseMap<uint32_t, uint32_t> bufferIDMap;
+  slang::BufferID mainBufferId;
 
   // A map from a file name to the corresponding buffer ID in the LLVM
   // source manager.
-  llvm::StringMap<std::pair<uint32_t, SmallString<128>>> filePathMap;
+  llvm::StringMap<std::pair<slang::BufferID, SmallString<128>>> filePathMap;
 
   // The compilation result.
   FailureOr<std::unique_ptr<slang::ast::Compilation>> compilation;
@@ -206,11 +203,12 @@ private:
   // The slang driver.
   slang::driver::Driver driver;
 
-  // The LLVM source manager.
-  llvm::SourceMgr sourceMgr;
-
   /// The index of the parsed module.
   VerilogIndex index;
+
+  /// The precomputed line offsets for faster lookups
+  std::vector<uint32_t> lineOffsets;
+  void computeLineOffsets(std::string_view text);
 
   // The URI of the document.
   llvm::lsp::URIForFile uri;
@@ -324,17 +322,6 @@ VerilogDocument::VerilogDocument(
         mainBufferFileInCommandFileList(cmdFile, canonPath.str().str());
   }
 
-  auto memBufferOwn =
-      llvm::MemoryBuffer::getMemBufferCopy(contents, uri.file());
-  if (!memBufferOwn) {
-    circt::lsp::Logger::error(
-        Twine("Failed to create memory buffer for file ") + uri.file());
-    return;
-  }
-
-  const unsigned int mainBufferId =
-      sourceMgr.AddNewSourceBuffer(std::move(memBufferOwn), SMLoc());
-
   // Build the set of include directories for this file.
   llvm::SmallString<32> uriDirectory(uri.file());
   llvm::sys::path::remove_filename(uriDirectory);
@@ -344,10 +331,14 @@ VerilogDocument::VerilogDocument(
   libDirs.insert(libDirs.end(), context.options.libDirs.begin(),
                  context.options.libDirs.end());
 
-  // Populate source managers.
-  const llvm::MemoryBuffer *memBuffer = sourceMgr.getMemoryBuffer(mainBufferId);
+  auto memBuffer = llvm::MemoryBuffer::getMemBufferCopy(contents, uri.file());
+  if (!memBuffer) {
+    circt::lsp::Logger::error(
+        Twine("Failed to create memory buffer for file ") + uri.file());
+    return;
+  }
 
-  // This block compiles the top file to determine all definitions
+  // This block parses the top file to determine all definitions
   // This is used in a second pass to declare all those definitions
   // as top modules, so they are elaborated and subsequently indexed.
   {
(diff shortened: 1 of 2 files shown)


lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -1030,15 +975,94 @@ static llvm::lsp::Range getRange(const mlir::FileLineColRange &fileLoc) {
       llvm::lsp::Position(fileLoc.getEndLine(), fileLoc.getEndColumn()));
 }
 
+// Decode one UTF-8 code point from [p, p+rem), advancing 'adv' bytes.
+// On malformed input, consumes 1 byte and returns U+FFFD.
+static inline uint32_t decodeUtf8(const char *p, size_t rem, size_t &adv) {
+  const unsigned char b0 = (unsigned char)p[0];
+  if (b0 < 0x80) {
+    adv = 1;
```
uenoku: I'm not familiar with this area much but can we reuse utilities in https://llvm.org/doxygen/ConvertUTF_8h.html ?

uenoku (commented):
This looks fantastic, thank you for unifying them!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Good point, I wasn't aware of these helper functions. I refactored `lspPositionToOffset` to use those functions instead.

Scheremo:
> This looks fantastic, thank you for unifying them!

Thanks! I'll open this for review / merging now, do let me know if you have any points that stick out to you :smile:
I'm happy with the current state, so feel free to merge if you don't find any issues.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -1030,15 +976,77 @@ static llvm::lsp::Range getRange(const mlir::FileLineColRange &fileLoc) {
...
+  // 1) Find start-of-line (clamp to EOF if line is past end)
+  size_t lineStart = 0;
+  for (int i = 0; i < pos.line && lineStart < text.size(); ++i) {
+    size_t nl = text.find('\n', lineStart);
+    if (nl == std::string_view::npos) {
+      lineStart = text.size();
+      break;
+    }
+    lineStart = nl + 1;
+  }
```
uenoku: Not blocking but I think this will cause performance regression since it's necessary to walk entire text in the worst case (chisel ~often~ sometimes emit a big file so ...).  LLVM sourceManager caches offsets and do the binary search here: 
https://github.com/llvm/llvm-project/blob/19cd5bd350b730da35629de5095764861f70ecee/llvm/lib/Support/SourceMgr.cpp#L104-L118

Slang source manager actually caches offsets as well,  https://github.com/MikePopoloski/slang/blob/d9ac86a723653474548737f55384b0d99f447d92/source/text/SourceManager.cpp#L661 so perhaps we can do binary search on location to determine line? It's totally fine for me to continue with the current implementation if there a tracking issue :)

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -1030,15 +976,77 @@ static llvm::lsp::Range getRange(const mlir::FileLineColRange &fileLoc) {
...
+        src - reinterpret_cast<const llvm::UTF8 *>(text.data()));
+
+  // We don't care about the actual UTF-16 text; we just need to count units.
+  // Allocate exactly 'target' UTF-16 slots so conversion halts when filled.
+  std::vector<llvm::UTF16> sink(target);
+  llvm::UTF16 *out = sink.data();
+  llvm::UTF16 *outEnd = out + sink.size();
+
+  // lenientConversion: malformed sequences become U+FFFD and still advance.
+  (void)llvm::ConvertUTF8toUTF16(&src, srcEnd, &out, outEnd,
```
uenoku: Could you add a test for UTF16?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -519,58 +490,29 @@ VerilogDocument::getLspLocation(slang::SourceRange range) const {
       start.uri, llvm::lsp::Range(start.range.start, end.range.end));
 }
 
-llvm::SMLoc VerilogDocument::getSMLoc(slang::SourceLocation loc) {
```
uenoku, line 522: Great clean up!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -1076,10 +1083,21 @@ void VerilogDocument::getLocationsOf(
...
+  auto &sm = getSlangSourceManager();
+  std::optional<uint32_t> slangBufferOffset =
+      lspPositionToOffset(sm, getMainBufferID(), pos);
+
+  if (!slangBufferOffset.has_value())
+    return;
+
+  uint32_t offset = slangBufferOffset.value();
+  const auto *slangBufferPointer =
+      sm.getSourceText(mainBufferId).data() + offset;
```
uenoku: Could you create a helper function for this and use it at above as well? 
```cpp
const char* getPointerFor(llvm::lsp::Position&pos) {
    auto &sm = getSlangSourceManager();
    auto slangBufferOffset = lspPositionToOffset(sm, getMainBufferID(), pos);

    if (!slangBufferOffset.has_value())
      return nullptr;

    uint32_t offset = slangBufferOffset.value();
    return sm.getSourceText(mainBufferId).data() + offset;
}

findReferencesOf(..) {
  const auto* slangBufferPointer = getPointerFor(pos);
  if(!slangBufferPointer)
    return;
}
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: That's a valid point. I suggest we just cache all line offsets per document and then just add the character offsets in-flight. This should be fine especially if we plan to debounce updating the document state anyway.

I implemented a draft of a version of this.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
```diff
@@ -976,53 +984,52 @@ static llvm::lsp::Range getRange(const mlir::FileLineColRange &fileLoc) {
...
-
-  // 1) Find start-of-line (clamp to EOF if line is past end)
-  size_t lineStart = 0;
-  for (int i = 0; i < pos.line && lineStart < text.size(); ++i) {
-    size_t nl = text.find('\n', lineStart);
-    if (nl == std::string_view::npos) {
-      lineStart = text.size();
-      break;
+/// Build a vector of line start offsets (0-based).
+void VerilogDocument::computeLineOffsets(std::string_view text) {
```
uenoku, line 989: Is it different from SourceManager::computeLineOffsets? https://github.com/MikePopoloski/slang/blob/d9ac86a723653474548737f55384b0d99f447d92/source/text/SourceManager.cpp#L758C6-L758C39

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
uenoku: Cool, thanks!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Done!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Not really :laughing:. `SourceManager::computeLineOffsets` is just a private member of `SourceManager` so I don't see a straight-forward way to call it in this context. Maybe I'm missing something?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Sorry, not sure what you mean by test for UTF16?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
uenoku: Oh, sorry. I was missing it was a private.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Thank you 😄

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
uenoku: Sorry my description was not sufficient. Ideally tests like https://github.com/llvm/llvm-project/blob/436f9f39cb259d28381aad87b777236c7cb94d02/clang-tools-extra/clangd/unittests/SourceCodeTests.cpp#L101-L120 would be great. Though introducing unittests are heavy so we can do that later. 

Alternatively can we insert few emoji with comments `/*😀*/` before the declaration of out https://github.com/llvm/circt/blob/ba41ee0b8ff7a783a0b2b20ef7c889c9b01ae934/test/Tools/circt-verilog-lsp-server/find-definition.test#L10 and check the offset is correct? I believe offset wouldn't be correct without code here.

uenoku (approved):
LGTM, thank you for improving buffer handling!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogServer.cpp
Scheremo: Yes, sure. I think unit tests make a lot of sense (easier to test things that way), but I would suggest to first break out `VerilogServer`, `VerilogServerContext`, `VerilogTextFile` and `VerilogDocument` into separate `.cc` files with corresponding header files. We could then add a unittest directory in the circt-verilog-lsp-server folder, e.g. like this:

VerilogServerImpl/
├─ CMakeLists.txt
├─ include/
│  ├─ VerilogTextFile.h
│  ├─ VerilogDocument.h
│  ├─ VerilogServerContext.h
│  └─ VerilogServer.h
├─ src/
│  ├─ VerilogTextFile.cc
│  ├─ VerilogDocument.cc
│  ├─ VerilogServerContext.cc
│  └─ VerilogServer.cc
└─ unittests/
   └─  VerilogDocumentTests.cpp

If something like this is fine for you, I can open a separate PR to implement it. Let me know what you think!

Scheremo: Also, I modified `find-defintions` to include some emojis before the module definition. Test still seems to pass 😄

Scheremo:
> LGTM, thank you for improving buffer handling!

Thank you once more for the helpful and targetted feedback. You really helped to point out some areas to improve further!
I'm okay with merging at this point, but would be interested in your opinion on this:

I think unit tests make a lot of sense (easier to test things that way), but I would suggest to first break out VerilogServer, VerilogServerContext, VerilogTextFile and VerilogDocument into separate .cc files with corresponding header files. We could then add a unittest directory in the VerilogServerImpl directory, e.g. like this:

VerilogServerImpl/
├─ CMakeLists.txt
├─ include/
│ ├─ VerilogTextFile.h
│ ├─ VerilogDocument.h
│ ├─ VerilogServerContext.h
│ └─ VerilogServer.h
├─ src/
│ ├─ VerilogTextFile.cc
│ ├─ VerilogDocument.cc
│ ├─ VerilogServerContext.cc
│ └─ VerilogServer.cc
└─ unittests/
└─ VerilogDocumentTests.cpp

If something like this is fine for you, I can open a separate PR to implement it. Let me know what you think!

uenoku:
> I'm okay with merging at this point, but would be interested in your opinion on this:

Thanks!  The current PR looks also great me! 

>  with corresponding header files. We could then add a unittest directory in the VerilogServerImpl directory, e.g. like this:

I agree, I think this looks awesome! I also think separating headers is great. One nitpick is we probably want to use flat structures without `include` and `src` directories. LLVM somehow prefers this kind of flat directories for header files used under `lib/` directories. 
```
VerilogServerImpl/
|- CMakeLists.txt
|- VerilogTextFile.h
|- VerilogTextFile.cpp
|- VerilogDocument.h
|- VerilogDocument.cpp
...
```

PR #9046, opened 2025-10-02
[circt-lsp-verilog] "Debounce" onDidChange calls; update in worker

Scheremo:
This commit introduces onChange debouncing and thread-safe diagnostics handling for the Verilog LSP server. The main motivation for this is project-scale LSP work, where many files are needed per compilation, and many file buffers are open at the same time; this PR limits the recompile frequency from per-keystroke to per time bin. It also runs the re-elaboration in a separate thread, which avoids locking down everything during elaboration.

- **Debouncer utility**
  - Added a background worker with configurable `delay` and `maxWait`.
  - Ensures trailing-edge execution of rebuilds after document changes.
  - Offloads callbacks onto detached threads to avoid blocking the timer loop.

- **Change buffering**
  - Added `ChangeBuffer` to collect incremental `didChange` edits atomically.
  - Supports full-file replacement by discarding previous incremental changes.
  - Provides `drain()` to flush buffered edits into a single batch update.

- **FileBucket abstraction**
  - Each open document is associated with a `FileBucket`.
  - Maintains change buffer, debounce instance, and `alive` state.
  - Uses `shared_ptr` for safe lifetime management across detached workers.

- **LSPServer integration**
  - `buckets` map is now private with accessor methods:
    - `getOrCreateBucket`, `findBucket`, `eraseBucket`, `cancelAllBuckets`.
  - Buckets are cleaned up on document close and shutdown.
  - `alive` flag prevents stale callbacks from running after a file is closed.
  - Shutdown cancels all buckets and marks them dead.

- **Diagnostics handling**
  - Introduced `sendDiagnostics()` with a dedicated `diagMu` mutex.
  - Ensures serialized writes to the underlying LSP transport.
  - Diagnostics publishing is safe from background worker threads.

- **Concurrency improvements**
  - `shutdownRequestReceived` promoted to `std::atomic<bool>`.
  - `VerilogTextFile::document` switched from `unique_ptr` to `shared_ptr` with atomic load/store for safe concurrent access.
  - Readers (`getLocationsOf`, `findReferencesOf`) use acquire loads.

- Smooths out rapid change events into a single rebuild pass.
- Prevents transport corruption by serializing outbound notifications.
- Eliminates race conditions on document lifetime during concurrent access.
- Provides clean shutdown with all debounced tasks canceled safely.

Currently the `VerilogTextfile` resolves definition/reference calls on the "current" VerilogDocument without taking into account changes that have been committed in the meantime. It might lead to a nicer user experience to heuristically update the in-flight definition/reference requests with the changebuffer queue.


uenoku:
Super cool, I didn't think about debounce but definitely right direction for scalability! I'll try taking a look at more closely. 

FYI clangd has similar mechanism so maybe we can learn the architecture https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/TUScheduler.h

Scheremo:
> Super cool, I didn't think about debounce but definitely right direction for scalability! I'll try taking a look at more closely.
> 
> FYI clangd has similar mechanism so maybe we can learn the architecture https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/TUScheduler.h

Nice! I think we are already somewhat close; clangd uses similar debouncing semantics (minimum delay, maximum wait) with adaptive resizing of those delays. It also uses a centralized thread pool to control concurrency. I think we can either push all of this into this PR, or separate it out into individual commits. I think separating it might be nicer, since this PR is already quite large and cerebral 😅 WDYT?

Scheremo:
> Thank you for taking a look! Yes let's separate commits.

@uenoku, If you have some more time, please have another look 😄 I factored out the actual scheduling so we can later plug this into a threadpool and refine debouncing semantics further; the LSPServer-side interface should only change marginally. Also added some unit tests. I'd be very grateful about any and all pointers about GTest integration; I gave it a go, but maybe there is a better way.

uenoku:
Cool! Will take a look!

https://github.com/llvm/circt/tree/main/unittests has gtests unittests so writing unittests under that directories would be easier (maybe `unittests/Tools/circt-verilog-lsp-server`). One tricky thing is it's probably necessary to move header files under top-level `include` so they are visible from the existing unittests directory. I think we can put these headers under either `circt/Support` or `circt/Tools/circt-verilog-lsp-server/` (if the files are generic enough we can put to `circt/Support` but `circt/Tools/circt-verilog-lsp-server/` might be better if they are specific to LSP).

lib/Tools/circt-verilog-lsp-server/Utils/ChangeBuffer.h
```diff
@@ -0,0 +1,48 @@
+
```
uenoku:
```suggestion
//===----------------------------------------------------------------------===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
```

Scheremo:
> Cool! Will take a look!
> 
> https://github.com/llvm/circt/tree/main/unittests has gtests unittests so writing unittests under that directories would be easier (maybe `unittests/Tools/circt-verilog-lsp-server`). One tricky thing is it's probably necessary to move header files under top-level `include` so they are visible from the existing unittests directory. I think we can put these headers under either `circt/Support` or `circt/Tools/circt-verilog-lsp-server/` (if the files are generic enough we can put to `circt/Support` but `circt/Tools/circt-verilog-lsp-server/` might be better if they are specific to LSP).

I moved the tests to `unittests/Tools/circt-verilog-lsp-server/Utils` and (until we figure out a better solution) added an INTERFACE library for unit testing that points to the `Utils` directory. The setup seems to work okay.

In my opinion I would leave the header files in `Utils` unless any other use-case comes to mind; I think we can always "promote" them to something more broadly visible if we find applications. WDYT?

lib/Tools/circt-verilog-lsp-server/LSPServer.cpp
```diff
@@ -23,9 +30,26 @@ using namespace llvm::lsp;
 //===----------------------------------------------------------------------===//
 
 namespace {
+
+using ChangeEvent = llvm::lsp::TextDocumentContentChangeEvent;
+using Bucket = circt::lsp::DocChangeBucket<ChangeEvent>;
```
uenoku: Unnecessary? 
```suggestion
```

lib/Tools/circt-verilog-lsp-server/Utils/DocChangeBucket.h
```diff
@@ -0,0 +1,219 @@
...
+
+private:
+  static void defaultCancel(BucketT &b) {
+    if constexpr (has_alive_member<BucketT>::value) {
+      b.alive.store(false, std::memory_order_release);
+    }
+    if constexpr (has_deb_member<BucketT>::value) {
+      if (b.deb)
+        b.deb->cancel();
+    }
```
uenoku: Are there potential use cases for classes without alive/debounce? If not could you remove conditions and check alive/debounce with static assertion for simplicity?

lib/Tools/circt-verilog-lsp-server/Utils/DocChangeBucket.h
```diff
@@ -0,0 +1,219 @@
...
+  template <typename Fn>
+  void scheduleUpdate(Fn &&step) {
+    auto weak = this->weak_from_this();
+    auto stepFn = StepFn(std::forward<Fn>(step));
+
+    // Fallback-friendly worker: runs even if not owned by shared_ptr
+    auto work = [weak, raw = this, stepFn = std::move(stepFn)]() mutable {
+      // Prefer shared ownership; fall back to raw if not managed.
+      auto fb = weak.lock();
+      auto *p = fb ? fb.get() : raw;
```
uenoku: I'm not sure I'm following this logic. Is it safe to refer `raw` if not managed by shared pointer?

lib/Tools/circt-verilog-lsp-server/Utils/DocChangeBucket.h
```diff
@@ -0,0 +1,219 @@
...
+  /// Thread-safe function to enqueue new changes
+  void enqueueChanges(const std::vector<T> &update, int64_t version) {
+    changeBuf.add(update, version);
+  }
+
+  /// Thread-safe function to cancel any outstanding updates
+  void cancelUpdate() {
+    if (deb)
+      deb->cancel();
+  }
```
uenoku: Are we using `cancelUpdate` and `deb->cancel` in non-test code?

uenoku:
> added an INTERFACE library for unit testing that points to the Utils

I think this makes sense. 

Thank you for excellent work and I think this looks awesome! I'm still trying to understand the logic but I'm wondering if it's possible to simplify the code structure and concurrency. I'm a bit concerned about the complexity and maintainability ;) I have a few questions:

* Is template necessary for ChangeBuffer and BucketRegistry? I'm not sure if `EventT` could be other than `TextDocumentContentChangeEvent` in LSP context. It helps to reduce cognitive load :)

* Any chance a thread pool simplifies the concurrency and resource management quite a bit(like `alive` and `cancelAll` etc)? Basically something like:

```cpp
threadPool.async([this, uri, timestamp] {
  std::this_thread::sleep_for(std::chrono::milliseconds(debounceMs));

  // Check if this change is still the latest
  if (isStillLatest(uri, timestamp)) {
    processDocument(uri);
  }
});
```

I think this would eliminate concerns related to detached thread managements and `shared_ptr`/`weak_ptr` lifetime because detached threads don't outlive the objects they reference. With a thread pool, the lifetime is managed since the thread pool can be properly shut down, waiting for all tasks to complete. 

I think the debouncing can be achieved simply with timestamps as well: each change gets a timestamp, workers sleep for the debounce delay, then check if the timestamp is latest. This is similar to clangd's TUScheduler. 


I think the core code would be something like:
```cpp
struct LSPServer {
struct PendingChanges { // probably equal to DocChangeBucket 
  std::vector<TextDocumentContentChangeEvent> changes;
  int64_t version = 0;
  std::chrono::steady_clock::time_point firstChangeTime;
  std::chrono::steady_clock::time_point lastChangeTime;
};

StringMap<PendingChanges> pending;
std::mutex pendingMutex;

void LSPServer::onDocumentDidChange(const DidChangeTextDocumentParams &params) {
  auto now = std::chrono::steady_clock::now();
  // Record this change with a timestamp
  {
    std::lock_guard lock(pendingMutex);
    auto &pending = pendingChanges[params.textDocument.uri.file().str()];
    pending.changes.insert(pending.changes.end(),
                          params.contentChanges.begin(),
                          params.contentChanges.end());
    pending.version = params.textDocument.version;
    pending.lastChangeTime = now;  // Track when this change happened
    if (pending.changes.size() == params.contentChanges.size()) {
      pending.firstChangeTime = now;  // For max wait feature
    }
  }

  // Schedule debounced processing
  threadPool.async([this, uri=params.textDocument.uri, scheduleTime=now, options] {
    // Wait for the debounce delay
    if (!options.disableDebounce)
        std::this_thread::sleep_for(std::chrono::milliseconds(options.debounceMinMs));

    // Check if this is still the latest change
    std::vector<TextDocumentContentChangeEvent> allChanges;
    int64_t version;
    bool shouldProcess = false;

    {
      std::lock_guard lock(pendingMutex);
      auto it = pendingChanges.find(uri.file().str());
      if (it == pendingChanges.end()) return;  // Document closed

      auto &pending = it->second;
      auto now = std::chrono::steady_clock::now();

      // Check if this is still the latest change
      bool isLatest = (pending.lastChangeTime == scheduleTime);

      // Check max wait timeout
      bool maxWaitExpired = false;
      if (options.debounceMaxMs > 0) {
        auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
            now - pending.firstChangeTime);
        maxWaitExpired = elapsed.count() >= options.debounceMaxMs;
      }

      shouldProcess = options.disableDebounce || isLatest || maxWaitExpired;

      if (shouldProcess) {
        allChanges = std::move(pending.changes);
        version = pending.version;
        pending.changes.clear();
        // Reset for next batch
        if (maxWaitExpired) {
          pending.firstChangeTime = now;
        }
      }
    }

    if (!shouldProcess) return;  // Newer change came in, skip this one

    PublishDiagnosticsParams diagParams(uri, version);
    server.updateDocument(uri, allChanges, version, diagParams.diagnostics);

    std::lock_guard lock(diagnosticsMutex);
    publishDiagnostics(diagParams);
  }
}
```

I haven't checked this code on large designs with frequent modification so it's possible that more fine-grained parallelism might be necessary for scalability, but I'm wondering something simpler still works. This is just a suggestion and I'd like to hear your thoughts!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
```diff
@@ -46,8 +46,15 @@ LogicalResult VerilogTextFile::update(
...
     std::vector<llvm::lsp::Diagnostic> &diagnostics) {
-  if (failed(llvm::lsp::TextDocumentContentChangeEvent::applyTo(changes,
-                                                                contents))) {
+
+  LogicalResult applied = failure();
+  {
+    std::unique_lock lk(contentMutex);
+    applied =
+        llvm::lsp::TextDocumentContentChangeEvent::applyTo(changes, contents);
+  }
```
uenoku: Maybe it's necessary to keep the lock through `initialize` (contetMutex is locked anyway in the `initialize`)? Otherwise I think two threads could initialize in wrong order. 

```
A                 B
applyTo(..);
                  applyTo(...);
                  initialize();
initialize();

```

lib/Tools/circt-verilog-lsp-server/Utils/DocChangeBucket.h
Scheremo: Fair enough, let's remove cancel overrides.

lib/Tools/circt-verilog-lsp-server/Utils/DocChangeBucket.h
Scheremo: This looks like a remainder of "trying something" for another issue... Removed it.

lib/Tools/circt-verilog-lsp-server/Utils/DocChangeBucket.h
Scheremo: Yes, it's used `onShutdown` to cancel running updates and through `erase` of `BucketRegistry` in `onDocumentDidClose` for the same reason :smile:

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
Scheremo: I think the case you describe is possible but uncritical as only the latest version of contents will be used to initialize VerilogDocument. On the other hand, it's probably also not an issue to keep the lock over initialization as well, so I changed it to do that instead.

Scheremo:
> > added an INTERFACE library for unit testing that points to the Utils
> 
> I think this makes sense.
> 
> Thank you for excellent work and I think this looks awesome! I'm still trying to understand the logic but I'm wondering if it's possible to simplify the code structure and concurrency. I'm a bit concerned about the complexity and maintainability ;) I have a few questions:
> 
> * Is template necessary for ChangeBuffer and BucketRegistry? I'm not sure if `EventT` could be other than `TextDocumentContentChangeEvent` in LSP context. It helps to reduce cognitive load :)
> * Any chance a thread pool simplifies the concurrency and resource management quite a bit(like `alive` and `cancelAll` etc)? Basically something like:
> 
> ```c++
> threadPool.async([this, uri, timestamp] {
>   std::this_thread::sleep_for(std::chrono::milliseconds(debounceMs));
> 
>   // Check if this change is still the latest
>   if (isStillLatest(uri, timestamp)) {
>     processDocument(uri);
>   }
> });
> ```
> 
> I think this would eliminate concerns related to detached thread managements and `shared_ptr`/`weak_ptr` lifetime because detached threads don't outlive the objects they reference. With a thread pool, the lifetime is managed since the thread pool can be properly shut down, waiting for all tasks to complete.
> 
> I think the debouncing can be achieved simply with timestamps as well: each change gets a timestamp, workers sleep for the debounce delay, then check if the timestamp is latest. This is similar to clangd's TUScheduler.
> 
> I think the core code would be something like:
> 
> ```c++
> struct LSPServer {
> struct PendingChanges { // probably equal to DocChangeBucket 
>   std::vector<TextDocumentContentChangeEvent> changes;
>   int64_t version = 0;
>   std::chrono::steady_clock::time_point firstChangeTime;
>   std::chrono::steady_clock::time_point lastChangeTime;
> };
> 
> StringMap<PendingChanges> pending;
> std::mutex pendingMutex;
> 
> void LSPServer::onDocumentDidChange(const DidChangeTextDocumentParams &params) {
>   auto now = std::chrono::steady_clock::now();
>   // Record this change with a timestamp
>   {
>     std::lock_guard lock(pendingMutex);
>     auto &pending = pendingChanges[params.textDocument.uri.file().str()];
>     pending.changes.insert(pending.changes.end(),
>                           params.contentChanges.begin(),
>                           params.contentChanges.end());
>     pending.version = params.textDocument.version;
>     pending.lastChangeTime = now;  // Track when this change happened
>     if (pending.changes.size() == params.contentChanges.size()) {
>       pending.firstChangeTime = now;  // For max wait feature
>     }
>   }
> 
>   // Schedule debounced processing
>   threadPool.async([this, uri=params.textDocument.uri, scheduleTime=now, options] {
>     // Wait for the debounce delay
>     if (!options.disableDebounce)
>         std::this_thread::sleep_for(std::chrono::milliseconds(options.debounceMinMs));
> 
>     // Check if this is still the latest change
>     std::vector<TextDocumentContentChangeEvent> allChanges;
>     int64_t version;
>     bool shouldProcess = false;
> 
>     {
>       std::lock_guard lock(pendingMutex);
>       auto it = pendingChanges.find(uri.file().str());
>       if (it == pendingChanges.end()) return;  // Document closed
> 
>       auto &pending = it->second;
>       auto now = std::chrono::steady_clock::now();
> 
>       // Check if this is still the latest change
>       bool isLatest = (pending.lastChangeTime == scheduleTime);
> 
>       // Check max wait timeout
>       bool maxWaitExpired = false;
>       if (options.debounceMaxMs > 0) {
>         auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
>             now - pending.firstChangeTime);
>         maxWaitExpired = elapsed.count() >= options.debounceMaxMs;
>       }
> 
>       shouldProcess = options.disableDebounce || isLatest || maxWaitExpired;
> 
>       if (shouldProcess) {
>         allChanges = std::move(pending.changes);
>         version = pending.version;
>         pending.changes.clear();
>         // Reset for next batch
>         if (maxWaitExpired) {
>           pending.firstChangeTime = now;
>         }
>       }
>     }
> 
>     if (!shouldProcess) return;  // Newer change came in, skip this one
> 
>     PublishDiagnosticsParams diagParams(uri, version);
>     server.updateDocument(uri, allChanges, version, diagParams.diagnostics);
> 
>     std::lock_guard lock(diagnosticsMutex);
>     publishDiagnostics(diagParams);
>   }
> }
> ```
> 
> I haven't checked this code on large designs with frequent modification so it's possible that more fine-grained parallelism might be necessary for scalability, but I'm wondering something simpler still works. This is just a suggestion and I'd like to hear your thoughts!

Thanks for giving this a closer look! I appreciate the feedback. I think your concern is valid and it's possible we don't need fine-grained concurrency / multiple docs updating at the same time, I didn't really profile this 😅. What you sketched seems reasonable to me; the main concern I have is that multiple (more or less useless) re-compilations might happen in parallel (e.g. if min delay / max wait are set too optimistically). What we can do is go full-on clangd `TUSchedule` and also implement the scaling of min delay and max wait according to a fixed ratio between update frequency and (measured) compile time, so such problems stay transient. We can also live with it for the moment and tackle it in a follow-up. I don't have major intuitive performance concerns with the solution you sketched.

One way or another I would suggest to extract the asynchronous scheduling parts of this so we can unit test them on their own - I'll give it a go, and we can see which version we like best :smile:

Scheremo:
> > > added an INTERFACE library for unit testing that points to the Utils
> > 
> > 
> > I think this makes sense.
> > Thank you for excellent work and I think this looks awesome! I'm still trying to understand the logic but I'm wondering if it's possible to simplify the code structure and concurrency. I'm a bit concerned about the complexity and maintainability ;) I have a few questions:
> > 
> > * Is template necessary for ChangeBuffer and BucketRegistry? I'm not sure if `EventT` could be other than `TextDocumentContentChangeEvent` in LSP context. It helps to reduce cognitive load :)
> > * Any chance a thread pool simplifies the concurrency and resource management quite a bit(like `alive` and `cancelAll` etc)? Basically something like:
> > 
> > ```c++
> > threadPool.async([this, uri, timestamp] {
> >   std::this_thread::sleep_for(std::chrono::milliseconds(debounceMs));
> > 
> >   // Check if this change is still the latest
> >   if (isStillLatest(uri, timestamp)) {
> >     processDocument(uri);
> >   }
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this would eliminate concerns related to detached thread managements and `shared_ptr`/`weak_ptr` lifetime because detached threads don't outlive the objects they reference. With a thread pool, the lifetime is managed since the thread pool can be properly shut down, waiting for all tasks to complete.
> > I think the debouncing can be achieved simply with timestamps as well: each change gets a timestamp, workers sleep for the debounce delay, then check if the timestamp is latest. This is similar to clangd's TUScheduler.
> > I think the core code would be something like:
> > ```c++
> > struct LSPServer {
> > struct PendingChanges { // probably equal to DocChangeBucket 
> >   std::vector<TextDocumentContentChangeEvent> changes;
> >   int64_t version = 0;
> >   std::chrono::steady_clock::time_point firstChangeTime;
> >   std::chrono::steady_clock::time_point lastChangeTime;
> > };
> > 
> > StringMap<PendingChanges> pending;
> > std::mutex pendingMutex;
> > 
> > void LSPServer::onDocumentDidChange(const DidChangeTextDocumentParams &params) {
> >   auto now = std::chrono::steady_clock::now();
> >   // Record this change with a timestamp
> >   {
> >     std::lock_guard lock(pendingMutex);
> >     auto &pending = pendingChanges[params.textDocument.uri.file().str()];
> >     pending.changes.insert(pending.changes.end(),
> >                           params.contentChanges.begin(),
> >                           params.contentChanges.end());
> >     pending.version = params.textDocument.version;
> >     pending.lastChangeTime = now;  // Track when this change happened
> >     if (pending.changes.size() == params.contentChanges.size()) {
> >       pending.firstChangeTime = now;  // For max wait feature
> >     }
> >   }
> > 
> >   // Schedule debounced processing
> >   threadPool.async([this, uri=params.textDocument.uri, scheduleTime=now, options] {
> >     // Wait for the debounce delay
> >     if (!options.disableDebounce)
> >         std::this_thread::sleep_for(std::chrono::milliseconds(options.debounceMinMs));
> > 
> >     // Check if this is still the latest change
> >     std::vector<TextDocumentContentChangeEvent> allChanges;
> >     int64_t version;
> >     bool shouldProcess = false;
> > 
> >     {
> >       std::lock_guard lock(pendingMutex);
> >       auto it = pendingChanges.find(uri.file().str());
> >       if (it == pendingChanges.end()) return;  // Document closed
> > 
> >       auto &pending = it->second;
> >       auto now = std::chrono::steady_clock::now();
> > 
> >       // Check if this is still the latest change
> >       bool isLatest = (pending.lastChangeTime == scheduleTime);
> > 
> >       // Check max wait timeout
> >       bool maxWaitExpired = false;
> >       if (options.debounceMaxMs > 0) {
> >         auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
> >             now - pending.firstChangeTime);
> >         maxWaitExpired = elapsed.count() >= options.debounceMaxMs;
> >       }
> > 
> >       shouldProcess = options.disableDebounce || isLatest || maxWaitExpired;
> > 
> >       if (shouldProcess) {
> >         allChanges = std::move(pending.changes);
> >         version = pending.version;
> >         pending.changes.clear();
> >         // Reset for next batch
> >         if (maxWaitExpired) {
> >           pending.firstChangeTime = now;
> >         }
> >       }
> >     }
> > 
> >     if (!shouldProcess) return;  // Newer change came in, skip this one
> > 
> >     PublishDiagnosticsParams diagParams(uri, version);
> >     server.updateDocument(uri, allChanges, version, diagParams.diagnostics);
> > 
> >     std::lock_guard lock(diagnosticsMutex);
> >     publishDiagnostics(diagParams);
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I haven't checked this code on large designs with frequent modification so it's possible that more fine-grained parallelism might be necessary for scalability, but I'm wondering something simpler still works. This is just a suggestion and I'd like to hear your thoughts!
> 
> Thanks for giving this a closer look! I appreciate the feedback. I think your concern is valid and it's possible we don't need fine-grained concurrency / multiple docs updating at the same time, I didn't really profile this 😅. What you sketched seems reasonable to me; the main concern I have is that multiple (more or less useless) re-compilations might happen in parallel (e.g. if min delay / max wait are set too optimistically). What we can do is go full-on clangd `TUSchedule` and also implement the scaling of min delay and max wait according to a fixed ratio between update frequency and (measured) compile time, so such problems stay transient. We can also live with it for the moment and tackle it in a follow-up. I don't have major intuitive performance concerns with the solution you sketched.
> 
> One way or another I would suggest to extract the asynchronous scheduling parts of this so we can unit test them on their own - I'll give it a go, and we can see which version we like best 😄

I built a version of the structure you suggested; I took the liberty to pack it into a data structure that lives out `LSPServer` (PendingChanges.h) and keep the semantics of "enqueuing a change" and "debounce and then run my callback".
I did that because it makes unit testing much easier; I added some tests to make sure the algorithm correctly debounces.

I also took the liberty of extracting construction of `DebounceOptions`, so we can extend them in a follow-up without having to change the LSPServer.

I also moved the handling of un-debounced calls back to the main thread; this is needed for the `textdocument-didchange.test` test to succeed, since it doesn't allow for the task switch to take any amount of time 😄 

I like this version better; I think you were right, it has less surface area for things to go wrong.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
```diff
@@ -39,41 +39,77 @@ VerilogTextFile::VerilogTextFile(
...
+
+    if (failed(applied)) {
+      circt::lsp::Logger::error(Twine("Failed to update contents of ") +
+                                uri.file());
+      return failure();
+    }
+
+    // If the file contents were properly changed, reinitialize the text file.
+    initialize(uri, newVersion, diagnostics);
+  }
```
uenoku, line 117: Maybe we can just add `std::unique_lock lk(contentMutex)` to original code instead of adding indent and temporary variable here?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
```diff
@@ -39,41 +39,77 @@ VerilogTextFile::VerilogTextFile(
...
   return success();
 }
 
 void VerilogTextFile::initialize(
     const llvm::lsp::URIForFile &uri, int64_t newVersion,
     std::vector<llvm::lsp::Diagnostic> &diagnostics) {
+  std::shared_ptr<VerilogDocument> newDocument;
+  newDocument =
+      std::make_shared<VerilogDocument>(context, uri, contents, diagnostics);
+  setDocument(newDocument);
```
uenoku:
```suggestion
  setDocument(std::make_shared<VerilogDocument>(context, uri, contents, diagnostics));
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
```diff
@@ -39,41 +39,77 @@ VerilogTextFile::VerilogTextFile(
...
-  document =
-      std::make_unique<VerilogDocument>(context, uri, contents, diagnostics);
 }
 
 void VerilogTextFile::getLocationsOf(
     const llvm::lsp::URIForFile &uri, llvm::lsp::Position defPos,
     std::vector<llvm::lsp::Location> &locations) {
-  document->getLocationsOf(uri, defPos, locations);
+  auto doc = getDocument();
+  if (doc)
```
uenoku: Is it possible that doc is null?

lib/Tools/circt-verilog-lsp-server/LSPServer.cpp
```diff
@@ -7,10 +7,16 @@
...
 #include "LSPServer.h"
+#include "Utils/PendingChanges.h"
 #include "VerilogServerImpl/VerilogServer.h"
+#include "circt/Tools/circt-verilog-lsp-server/CirctVerilogLspServerMain.h"
 #include "llvm/Support/JSON.h"
 #include "llvm/Support/LSP/Protocol.h"
 #include "llvm/Support/LSP/Transport.h"
+
+#include <atomic>
+#include <cstdint>
```
uenoku, line 17: <cstdint> may not be necessary?

lib/Tools/circt-verilog-lsp-server/LSPServer.cpp
```diff
@@ -60,16 +70,34 @@ struct LSPServer {
   circt::lsp::VerilogServer &server;
   JSONTransport &transport;
 
-  /// An outgoing notification used to send diagnostics to the client when they
-  /// are ready to be processed.
-  OutgoingNotification<PublishDiagnosticsParams> publishDiagnostics;
+  /// A thread-safe version of `publishDiagnostics`
+  void sendDiagnostics(const PublishDiagnosticsParams &p) {
+    std::lock_guard<std::mutex> lk(diagnosticsMutex);
```
uenoku: Maybe `std::scoped_lock lk(diagnosticsMutex);` is preferred. 
https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-scoped-lock.html

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
```diff
@@ -39,41 +39,77 @@ VerilogTextFile::VerilogTextFile(
     StringRef fileContents, int64_t version,
     std::vector<llvm::lsp::Diagnostic> &diagnostics)
     : context(context), contents(fileContents.str()) {
-  initialize(uri, version, diagnostics);
+  {
+    std::unique_lock lk(contentMutex);
```
uenoku: This could be also std::scoped_lock or lock_guard?

lib/Tools/circt-verilog-lsp-server/Utils/PendingChanges.h
```diff
@@ -0,0 +1,220 @@
...
+/// Accumulated edits + timing for a single document key.
+struct PendingChanges {
+  std::vector<llvm::lsp::TextDocumentContentChangeEvent> changes;
+  int64_t version = 0;
+  std::chrono::steady_clock::time_point firstChangeTime{};
+  std::chrono::steady_clock::time_point lastChangeTime{};
+};
+
+/// Thread-safe accumulator + debouncer for text document changes.
+class PendingChangesMap {
```
uenoku, line 55: Non-blocking but maybe we don't want to put entire implementation to the header. I understand this is for the ease of the unittesting so I think we can fix later. Not ideal but it's not that bad to just put declaration to the public header and link the library in unittests.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
```diff
@@ -39,41 +39,77 @@ VerilogTextFile::VerilogTextFile(
...
+  if (doc)
+    doc->findReferencesOf(uri, pos, references);
+}
+
+std::shared_ptr<VerilogDocument> VerilogTextFile::getDocument() {
+  std::shared_ptr<circt::lsp::VerilogDocument> doc;
+  {
+    std::unique_lock lk(docMutex);
+    doc = document;
+  }
```
uenoku: Could you leave a comment why it uses a temporary variable?

lib/Tools/circt-verilog-lsp-server/Utils/PendingChanges.h
```diff
@@ -0,0 +1,220 @@
...
+public:
+  explicit PendingChangesMap(
+      unsigned maxThreads = std::thread::hardware_concurrency())
+      : pool(makeStrategy(maxThreads)), tasks(pool) {}
+
+  /// Call during server shutdown; Erase all file changes, then clear file map.
+  /// Thread-safe.
+  void abort() {
+    std::lock_guard<std::mutex> lock(mu);
+    pending.clear();
```
uenoku: I think it's necessary to wait workers to avoid UAF. 
```suggestion
pool.wait();
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
Scheremo: Yep, done!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
Scheremo: Sure, done!

lib/Tools/circt-verilog-lsp-server/LSPServer.cpp
Scheremo: True, removed it.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
Scheremo: Shouldn't be, it gets constructed with `initialize`. Will remove the check.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
Scheremo: We can also just return document; I'm just used to the {lock; set;} return; pattern :smile:

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
Scheremo: Sure!

lib/Tools/circt-verilog-lsp-server/LSPServer.cpp
Scheremo: Okay, will adapt lock_guards.

lib/Tools/circt-verilog-lsp-server/Utils/PendingChanges.h
Scheremo: Yep, done!

lib/Tools/circt-verilog-lsp-server/Utils/PendingChanges.h
Scheremo: Extracted implementations into their own `PendingChanges.cpp`

uenoku:
Cool!

PR #9048, opened 2025-10-03
[CombToSynth] Use parallel-prefix tree for unsigned comparisons

uenoku:
This commit extends the unsigned comparison lowering to support multiple parallel-prefix architectures (Sklanskey, Kogge-Stone, Brent-Kung) in addition to the existing ripple-carry implementation. No functional change in prefix-tree/adder lowering. 

Previously, all comparisons used a ripple-carry style implementation that processed bits sequentially from LSB to MSB, resulting in O(n) depth for n-bit comparisons. This was a significant performance bottleneck for wide comparisons. The comparison lowering is now refactored to use the same parallel-prefix tree algorithms as the adder, reducing depth to O(log n).

The comparison logic is formulated as a prefix computation where equal bits are computed as ~(a_i ^ b_i) and greater bits as ~a_i & b_i, with propagate and generate signals based on equality and greater-than conditions.

Integration tests for all architectures verify logical equivalence via circt-lec. 

before/after for 64-bit unsigned comparision. 
```
uenoku@pop-os ~/d/circt-synth (dev/hidetou/icmp-parallel) [1]> ../circt/build/bin/circt-synth bar.mlir -output-longest-path=- -o /dev/null
# Longest Path Analysis result for "icmp_unsigned"
Found 128 paths
Found 1 unique fanout points
Maximum path delay: 128
## Showing Levels
Level = 128       . Count = 1         . 100.00    %
## Top 0 (out of 1) fan-out points

uenoku@pop-os ~/d/circt-synth (dev/hidetou/icmp-parallel)> ./build/bin/circt-synth bar.mlir -output-longest-path=- -o /dev/null 
# Longest Path Analysis result for "icmp_unsigned"
Found 128 paths
Found 1 unique end points 
Maximum path delay: 14
## Showing Levels
Level = 14        . Count = 1         . 100.00    %
## Top 0 (out of 1) end points
```

### DIFF
--- a/integration_test/circt-synth/comb-lowering-compare.mlir
+++ b/integration_test/circt-synth/comb-lowering-compare.mlir
@@ -1,7 +1,7 @@
 // REQUIRES: libz3
 // REQUIRES: circt-lec-jit
 
-// RUN: circt-opt %s --hw-aggregate-to-comb --convert-comb-to-synth --convert-synth-to-comb -o %t.mlir
+// RUN: circt-opt %s --convert-comb-to-synth --convert-synth-to-comb -o %t.mlir
 
 // RUN: circt-lec %t.mlir %s -c1=icmp_unsigned_ripple_carry -c2=icmp_unsigned_ripple_carry --shared-libs=%libz3 | FileCheck %s --check-prefix=COMB_ICMP_UNSIGNED_RIPPLE_CARRY
 // COMB_ICMP_UNSIGNED_RIPPLE_CARRY: c1 == c2
--- a/integration_test/circt-synth/comb-lowering-compare.mlir
+++ b/integration_test/circt-synth/comb-lowering-compare.mlir
@@ -1,0 +1,56 @@
+// REQUIRES: libz3
+// REQUIRES: circt-lec-jit
+
+// RUN: circt-opt %s --hw-aggregate-to-comb --convert-comb-to-synth --convert-synth-to-comb -o %t.mlir
+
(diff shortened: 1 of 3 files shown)


uenoku:
FileCheck test is missing in non-integration test and I'll add a test for parallel prefix tree.

integration_test/circt-synth/comb-lowering-compare.mlir
```diff
@@ -0,0 +1,56 @@
+// REQUIRES: libz3
+// REQUIRES: circt-lec-jit
+
+// RUN: circt-opt %s --hw-aggregate-to-comb --convert-comb-to-synth --convert-synth-to-comb -o %t.mlir
```
cowardsa: Do you need hw-aggregate-to-comb? circt-lec usually works without this for me?

uenoku:
> could pass a flag to prevent the prefix computation generating a lot of gates that will then need to be removed by DCE? Namely, we are only interested in generating the carry-out and propagate out? 

Good points, but I'm not sure there is a non-complicated way to prune prefix computation since it's hard to know beforehand which index is actually used by carry-out and propagate out in the last stage. I think we can change the prefix computation for comparison to use recursive function with memoization that by nature lazily computes ony necessary prefixes, but it requires quite a bit of changes for prefix tree generation functions we have now. So please let me stick with the current implementation for this PR.

cowardsa:
> > could pass a flag to prevent the prefix computation generating a lot of gates that will then need to be removed by DCE? Namely, we are only interested in generating the carry-out and propagate out?
> 
> Good points, but I'm not sure there is a non-complicated way to prune prefix computation since it's hard to know beforehand which index is actually used by carry-out and propagate out in the last stage. I think we can change the prefix computation for comparison to use recursive function with memoization that by nature lazily computes ony necessary prefixes, but it requires quite a bit of changes for prefix tree generation functions we have now. So please let me stick with the current implementation for this PR.

Absolutely non-blocking for sure - just a thought for future improvements if we hit against performance issues.

PR #9049, opened 2025-10-04
No custom printer for CompReg - leading to loss of quotation marks Resolves #8833

### LINKED ISSUE #8833: [Seq] Register symbol with special characters does not round-trip
It looks like a `sym @...` on a `seq.compreg` (and likely any other op that can have an inner symbol) gets printed without `"..."` even if it contains special characters:
```
// foo.mlir
hw.module @Foo(in %a: i42, in %clk: !seq.clock) {
  %0 = seq.compreg sym @"foo/bar" %a, %clk : i42
  hw.output
}

// after circt-opt foo.mlir
hw.module @Foo(in %a : i42, in %clk : !seq.clock) {
  %0 = seq.compreg sym @foo/bar %a, %clk : i42
  hw.output
}

// after circt-opt foo.mlir | circt-opt
<stdin>:3:30: error: expected SSA operand
    %0 = seq.compreg sym @foo/bar %a, %clk : i42
                             ^
```

KaiS987:
I discovered that compreg didn’t have a custom printer, so it was using the default format. That format wasn’t handling quoting correctly. The solution was to add a small custom parser and printer for the optional inner symbol attribute. Resolves #8833


fabianschuiki:
Hey @KaiS987, sorry for taking so long to fire off CI. Thanks for taking a stab at fixing this :partying_face:! It looks like a few of your changes might have gotten lost along the way, e.g. actually invoking the custom parser/printer from `SeqOps.td`. CI will probably find that.

PR #9050, opened 2025-10-04
[CombToSynth] Compute Kogge-Stone prefix tree lazily in unsigned comparison lowering

uenoku:
Add LazyKoggeStonePrefixTree class for on-demand computation of prefix values, use lazy evaluation in ICmpOp conversion to avoid computing all intermediate prefix values.


integration_test/circt-synth/comb-lowering-compare.mlir
```diff
@@ -25,11 +25,12 @@ hw.module @icmp_unsigned_sklanskey(in %lhs: i3, in %rhs: i3, out out_ugt: i1, ou
 
 // RUN: circt-lec %t.mlir %s -c1=icmp_unsigned_kogge_stone -c2=icmp_unsigned_kogge_stone --shared-libs=%libz3 | FileCheck %s --check-prefix=COMB_ICMP_UNSIGNED_KOGGE_STONE
 // COMB_ICMP_UNSIGNED_KOGGE_STONE: c1 == c2
-hw.module @icmp_unsigned_kogge_stone(in %lhs: i3, in %rhs: i3, out out_ugt: i1, out out_uge: i1, out out_ult: i1, out out_ule: i1) {
-  %ugt = comb.icmp ugt %lhs, %rhs {synth.test.arch = "KOGGE-STONE"} : i3
-  %uge = comb.icmp uge %lhs, %rhs {synth.test.arch = "KOGGE-STONE"} : i3
-  %ult = comb.icmp ult %lhs, %rhs {synth.test.arch = "KOGGE-STONE"} : i3
-  %ule = comb.icmp ule %lhs, %rhs {synth.test.arch = "KOGGE-STONE"} : i3
+// Use slightly larger with to verify the lazy prefix tree logic
```
uenoku:
```suggestion
// Use slightly larger width to verify the lazy prefix tree logic
```

lib/Conversion/CombToSynth/CombToSynth.cpp
```diff
@@ -700,6 +700,67 @@ void lowerBrentKungPrefixTree(OpBuilder &builder, Location loc,
...
+  std::pair<Value, Value> getGroupAndPropagate(int64_t level, int64_t i);
+  OpBuilder &builder;
+  Location loc;
+  int64_t width;
+  DenseMap<std::pair<int64_t, int64_t>, std::pair<Value, Value>> prefixCache;
+};
+
+std::pair<Value, Value>
+LazyKoggeStonePrefixTree::getGroupAndPropagate(int64_t level, int64_t i) {
+  assert(i < static_cast<int64_t>(width) && "i out of bounds");
```
cowardsa: Probably don't understand - but do you need this cast as width is already int64_t?

lib/Conversion/CombToSynth/CombToSynth.cpp
```diff
@@ -700,6 +700,67 @@ void lowerBrentKungPrefixTree(OpBuilder &builder, Location loc,
...
+  auto key = std::make_pair(level, i);
+  auto it = prefixCache.find(key);
+  if (it != prefixCache.end())
+    return it->second;
+
+  assert(level > 0 && "level must be positive");
```
cowardsa: Nit: perhaps a comment to justify that we're now moving to compute the values and we can only do that after level 0 would help - took me a minute to get that

lib/Conversion/CombToSynth/CombToSynth.cpp
```diff
@@ -1080,37 +1141,49 @@ struct CombICmpOpConversion : OpConversionPattern<ICmpOp> {
...
     }
+    }
 
     // Final result: gPrefix[width-1] gives us "a < b"
     if (includeEq) {
       // a <= b iff (a < b) OR (a == b)
       // a == b iff pPrefix[width-1] (all bits are equal)
-      return comb::OrOp::create(rewriter, loc, gPrefix[width - 1],
-                                pPrefix[width - 1]);
+      return comb::OrOp::create(rewriter, loc, finalGroup, finalPropagate);
```
cowardsa, line 1183: Nit: perhaps update the comment to reflect the variable name changes?

lib/Conversion/CombToSynth/CombToSynth.cpp
```diff
@@ -1080,37 +1141,49 @@ struct CombICmpOpConversion : OpConversionPattern<ICmpOp> {
...
     }
     // a < b iff gPrefix[width-1]
-    return gPrefix[width - 1];
+    return finalGroup;
```
cowardsa, line 1186: Nit: again could update the comment?

lib/Conversion/CombToSynth/CombToSynth.cpp
uenoku: Absolutely, sorry for the confusion.

uenoku:
> Brent-Kung

Yeah Brent-Kung would be really non-trivial and not sure if there is a clean way :) I'll probably prioritize Sklanskey as it's used in 32-bit comparison (which creates ~500 dead operations right now).

PR #9053, opened 2025-10-05
[CombToDatapath] Lower comb::SubOp

uenoku:
Adds support for lowering comb::SubOp in the CombToDatapath conversion pass by refactoring the SubOp-to-AddOp conversion into a shared utility function that can be reused across multiple conversion passes. This change makes SubOp it possible to leverage datapath optimizations

This commit extracts the subtraction lowering logic from CombToSynth into a dialect-agnostic utility function `comb::convertSubToAdd` that transforms `sub(lhs, rhs)` into `add(lhs, ~rhs, 1)` using two's complement representation.

AIG logic depth improved 29 -> 24 for i128 subtraction

### DIFF
--- a/lib/Dialect/Comb/CombOps.cpp
+++ b/lib/Dialect/Comb/CombOps.cpp
@@ -12,6 +12,7 @@
 
 #include "circt/Dialect/Comb/CombOps.h"
 #include "circt/Dialect/HW/HWOps.h"
+#include "circt/Support/Naming.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/ImplicitLocOpBuilder.h"
 #include "mlir/IR/Matchers.h"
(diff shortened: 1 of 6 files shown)


lib/Dialect/Comb/CombOps.cpp
```diff
@@ -217,6 +218,22 @@ Value comb::createInject(OpBuilder &builder, Location loc, Value value,
...
 
+llvm::LogicalResult comb::convertSubToAdd(comb::SubOp subOp,
+                                          mlir::PatternRewriter &rewriter) {
+  auto lhs = subOp.getLhs();
+  auto rhs = subOp.getRhs();
+  // Since `-rhs = ~rhs + 1` holds, rewrite `sub(lhs, rhs)` to:
+  // sub(lhs, rhs) => add(lhs, -rhs) => add(lhs, add(~rhs, 1))
+  // => add(lhs, ~rhs, 1)
+  auto notRhs =
+      comb::createOrFoldNot(subOp.getLoc(), rhs, rewriter, subOp.getTwoState());
```
cowardsa, line 229: Neat - did no realise there was a createOrFoldNot() function - will stop writing create constant one and Xor... Even though it presumably maps to this

cowardsa (approved):
LGTM - thanks for embracing the compressors! I think this is a sensible step. But not entirely sure on the best long term strategy.

I've been dwelling on this for some time as currently the compress operator is unaware of constant ones. The problem is that each adder can accommodate a carry-in (at no logic depth cost) so we don't really need a compress operator at all here - but what we do need is an adder that can represent the carry-in - this is a bigger discussion (and certainly there may be other ways) so non-blocking as this step is clearly sensible!

uenoku:
> I've been dwelling on this for some time as currently the compress operator is unaware of constant ones. The problem is that each adder can accommodate a carry-in (at no logic depth cost) so we don't really need a compress operator at all here 

Yeah that makes sense.  That sounds like clever optimization in DatapathToComb.

PR #9055, opened 2025-10-06
[circt-verilog-lsp] Avoid recursing into other buffers when indexing

Scheremo:
Profiling showed `visitDefault` dominated indexing time (~99.99% of runtime in larger projects) due to deep traversal into non-relevant buffers. This patch keeps correctness while greatly reducing unnecessary recursion. In a SoC-scale System verilog project, this change reduces indexing time from 6.6 s to 4 ms.

- Added `definitelyOutsideMainBuffer()` helper to detect whether an AST node’s source range, syntax range, or location belongs to the main buffer.
- Integrated this check into the templated `visit<T>` function:
  - Skips recursion (`visitDefault`) when a node is clearly outside the main file.
  - Uses a new `ASTBase` alias for clarity and correctness when calling `visitDefault<T>()`.
- Removed redundant `visitDefault()` calls from individual symbol visitors (`NetSymbol`, `VariableSymbol`, `ImportSymbol`, etc.) that no longer need to recurse manually.
- Result: traversal remains correct but prunes large off-buffer branches early, improving indexing time for multi-file compilations.

### DIFF
--- a/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogIndex.cpp
+++ b/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogIndex.cpp
@@ -29,8 +29,39 @@ using namespace llvm;
 
 namespace {
 
+/// Helper function to determine whether an AST node is outside of the main
+/// buffer. If this can't be safely determined, return false.
+template <typename T>
+inline bool definitelyOutsideMainBuffer(const T &t, slang::BufferID bufferId) {
+  if constexpr (requires {
+                  t.location;
+                }) { // AST nodes with location, e.g. Symbols
+    auto r = t.location;
+    if (r.valid())
+      return r.buffer() != bufferId;
+  }
+
+  if constexpr (requires {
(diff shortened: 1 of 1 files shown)


lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogIndex.cpp
```diff
@@ -29,8 +29,40 @@ using namespace llvm;
 
 namespace {
 
+/// Helper function to determine whether an AST node is outside of the main
+/// buffer. If this can't be safely determined, return false.
+template <typename T>
+static inline bool definitelyOutsideMainBuffer(const T &t,
```
uenoku: nit: static is not necessary under anonymous namespace. 
```suggestion
inline bool definitelyOutsideMainBuffer(const T &t,
```

Scheremo: Addressed! Thanks for the quick review 😄

PR #9056, opened 2025-10-06
[circt-verilog-lsp] Minimize project-scope File IO

Scheremo:
Reduce filesystem I/O overhead caused by repeated re-opening and reparsing of project files in the Verilog LSP server when using the `-C` option. For a large, SoC-scale project this approach reduces parsing time from ~1s to 0.75s.
This implementation also correctly indexes the editor-state rather than the file system state of the main buffer.

- **Introduced shared `projectDriver`:** A persistent `slang::driver::Driver` is now created once per project and passed to all `VerilogDocument` instances to reuse loaded files and avoid redundant disk access.
- **Refactored `VerilogDocument` initialization:**
  - Removed per-buffer command file parsing and main-file filtering logic.
  - Added `setTopModules()` and `copyBuffers()` helpers to extract top-level modules and reuse in-memory buffers from `projectDriver` instead of reading from disk.
- **Inheritance of project context:**
  - Copied defines/undefines directly from the shared driver.
  - Library search directories are now applied once at construction.
- **VerilogTextFile integration:**
  - Added `initializeProjectDriver()` to parse all command files once and hold a shared driver reference for reuse.

### DIFF
--- a/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogDocument.cpp
+++ b/lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogDocument.cpp
@@ -55,75 +55,73 @@ using namespace circt::lsp;
 using namespace llvm;
 using namespace llvm::lsp;
 
-static std::filesystem::path
-canonicalizeFileName(const std::filesystem::path &file) {
-  std::error_code ec;
-  std::filesystem::path path = std::filesystem::weakly_canonical(file, ec);
-  if (ec)
-    path = std::filesystem::absolute(file).lexically_normal();
-  return path;
+static inline void setTopModules(slang::driver::Driver &driver) {
+  // Parse the main buffer
+  if (!driver.parseAllSources()) {
+    circt::lsp::Logger::error(Twine("Failed to parse main buffer "));
+    return;
+  }
+  // Extract all the top modules in the file directly from the syntax tree
+  std::vector<std::string> topModules;
+  for (auto &t : driver.syntaxTrees) {
+    if (auto *compUnit =
+            t->root().as_if<slang::syntax::CompilationUnitSyntax>()) {
+      for (auto *member : compUnit->members) {
+        // While it's called "ModuleDeclarationSyntax", it also covers
+        // packages
+        if (auto *moduleDecl =
+                member->as_if<slang::syntax::ModuleDeclarationSyntax>()) {
+          topModules.emplace_back(moduleDecl->header->name.valueText());
+        }
+      }
+    }
+  }
+  driver.options.topModules = std::move(topModules);
 }
 
-// Filter out the main buffer file from the command file list, if it is in
-// there.
-static inline bool
-mainBufferFileInCommandFileList(const std::string &cmdfileStr,
-                                const std::string &targetAbsStr) {
-  const std::filesystem::path targetAbs =
-      canonicalizeFileName(std::filesystem::path(targetAbsStr));
+static inline void
+copyBuffers(slang::driver::Driver &driver,
+            const slang::driver::Driver *const projectDriver,
+            const llvm::SmallString<256> &mainBufferFileName) {
+  for (auto bId : projectDriver->sourceManager.getAllBuffers()) {
+    std::string_view slangRawPath =
+        projectDriver->sourceManager.getRawFileName(bId);
 
-  std::string error;
-  auto cmdFile = mlir::openInputFile(cmdfileStr, &error);
-  if (!cmdFile) {
-    circt::lsp::Logger::error(Twine("Failed to open command file ") +
-                              cmdfileStr + ": " + error);
-    return false;
-  }
-
-  const std::filesystem::path base =
-      std::filesystem::path(cmdFile->getBufferIdentifier().str()).parent_path();
-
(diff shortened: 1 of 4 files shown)


lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogDocument.cpp
```diff
@@ -202,12 +156,33 @@ VerilogDocument::VerilogDocument(
...
+  if (projectDriver) {
+    setTopModules(driver);
+    // Copy defines & undefines from project driver
+    driver.options.defines.insert(driver.options.defines.end(),
+                                  projectDriver->options.defines.begin(),
+                                  projectDriver->options.defines.end());
+    driver.options.undefines.insert(driver.options.undefines.end(),
+                                    projectDriver->options.undefines.begin(),
+                                    projectDriver->options.undefines.end());
+    // Copy dependency buffers from project driver
```
uenoku: Maybe we should just copy options other than top modules?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
```diff
@@ -58,12 +63,80 @@ LogicalResult VerilogTextFile::update(
...
+  if (context.options.commandFiles.empty())
+    return;
+
+  projectDriver = std::make_shared<slang::driver::Driver>();
+
+  // --- Apply project command files (the “-C”s) to this per-buffer driver ---
+  for (const std::string &cmdFile : context.options.commandFiles) {
+    if (!projectDriver->processCommandFiles(cmdFile, false, true)) {
+      circt::lsp::Logger::error(Twine("Failed to open command file ") +
+                                cmdFile);
```
uenoku, line 76: Should it continue?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.h
```diff
@@ -64,11 +64,17 @@ class VerilogTextFile {
...
 
+  void initializeProjectDriver();
+
   VerilogServerContext &context;
 
   /// The full string contents of the file.
   std::string contents;
 
+  /// The project-scale driver
+  std::shared_ptr<slang::driver::Driver> projectDriver;
```
uenoku: Why is shared_ptr necessary? I thought it's owned by VerilogTextFile?

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogDocument.cpp
```diff
@@ -134,65 +132,21 @@ VerilogDocument::VerilogDocument(
   libDirs.insert(libDirs.end(), context.options.libDirs.begin(),
                  context.options.libDirs.end());
 
+  for (const auto &libDir : libDirs) {
+    driver.sourceLoader.addSearchDirectories(libDir);
+  }
```
uenoku: nit:
```suggestion
  for (const auto &libDir : libDirs) 
    driver.sourceLoader.addSearchDirectories(libDir);
```

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogDocument.cpp
```diff
@@ -202,12 +156,33 @@ VerilogDocument::VerilogDocument(
   driver.options.compilationFlags.emplace(
       slang::ast::CompilationFlags::DisableInstanceCaching, false);
 
+  for (auto &dir : projectIncludeDirectories) {
+    if (!driver.sourceManager.addUserDirectories(dir)) {
+      continue;
+    }
+  }
```
uenoku: nit:
```suggestion
  for (auto &dir : projectIncludeDirectories)
    (void)driver.sourceManager.addUserDirectories(dir);
```

uenoku (approved):
Question on shared pointer but otherwise LGTM

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.h
Scheremo: Fair point. I wasn't 100% sure that no `VerilogDocument` could outlive `VerilogTextFile`, but that is clearly not the case here. I made it a unique pointer and changed `VerilogDocument` to take a raw const pointer.

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogTextFile.cpp
Scheremo: No, fixed it!

lib/Tools/circt-verilog-lsp-server/VerilogServerImpl/VerilogDocument.cpp
Scheremo: Fair point. I copy all options now and then override top modules.

Scheremo:
> Question on shared pointer but otherwise LGTM

Thanks for the review! I think I addressed the points you raised, but I'll keep this open for a day or so before merging, just in case you'd like to have another look.

PR #9064, opened 2025-10-07
[ESI] Various tweaks to cosim scripts

mortbopet:
A collection of small tweaks to the cosim scripts:

- Make questa write a waveform if `debug` is enabled for the `Simulator`
- Allow for providing a list of error codes for Questa to suppress during compilation
- Allow for providing a dictionary of macro definitions that the simulator(s) should include when compiling the simulation.

### DIFF
--- a/lib/Dialect/ESI/runtime/python/esiaccel/cosim/questa.py
+++ b/lib/Dialect/ESI/runtime/python/esiaccel/cosim/questa.py
@@ -4,15 +4,41 @@
 
 import os
 from pathlib import Path
-from typing import List
+from typing import List, Optional, Callable, Dict
 
-from .simulator import CosimCollateralDir, Simulator
+from .simulator import CosimCollateralDir, Simulator, SourceFiles
 
 
 class Questa(Simulator):
   """Run and compile funcs for Questasim."""
 
   DefaultDriver = CosimCollateralDir / "driver.sv"
+
+  def __init__(
+      self,
+      sources: SourceFiles,
+      run_dir: Path,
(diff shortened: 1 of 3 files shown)


mortbopet:
Publish pipeline https://github.com/llvm/circt/actions/runs/18306516308

lib/Dialect/ESI/runtime/python/esiaccel/cosim/questa.py
```diff
@@ -53,9 +97,28 @@ def run_command(self, gui: bool) -> List[str]:
           vsim,
           "driver_opt",
           "-batch",
+      ]
+
+      if self.debug:
```
teqdruid, line 102: FYI: I also added the capability to dump a VCD using the SAVE_WAVE functionality just like in Verilator.

lib/Dialect/ESI/runtime/python/esiaccel/cosim/questa.py
```diff
@@ -53,9 +97,28 @@ def run_command(self, gui: bool) -> List[str]:
...
+            "-do",
+            str(wave_file.resolve()),
+        ]
+        # Questa will by default log to a vsim.wlf file in the current
+        # directory.
+        print(
+            f"Debug mode enabled - waveform will be in {wave_file.resolve().parent / 'vsim.wlf'}"
+        )
+
+      cmd += [
```
teqdruid, line 117: Have you tested this? I'm surprised that vsim accepts two `-do` arguments.

PR #9067, opened 2025-10-07
[FIRRTL] Add DomainDefineOp

rwy7:
Add a "domain define op" which can be used to drive output domain ports of modules, or input domain ports of instances.


include/circt/Dialect/FIRRTL/FIRRTLStatements.td
```diff
@@ -108,6 +108,32 @@ def RefDefineOp : FIRRTLOp<"ref.define", [SameTypeOperands, FConnectLike]> {
...
+    Source and destination must resolve statically.
+  }];
+  let arguments = (ins DomainType:$dest, DomainType:$src);
+  let hasVerifier = 1;
+  let assemblyFormat = "$dest `,` $src  attr-dict";
+  let extraClassDeclaration = [{
+    static ConnectBehaviorKind getConnectBehaviorKind() {
+      return ConnectBehaviorKind::StaticSingleConnect;
+    }
+  }];
```
seldridge, line 135: Thanks for _not_ adding a custom MLIR printer/parser. 👍

lib/Dialect/FIRRTL/Export/FIREmitter.cpp
```diff
@@ -745,9 +746,9 @@ void Emitter::emitStatementsInBlock(Block &block) {
               ConnectOp, MatchingConnectOp, PropAssignOp, InstanceOp,
               InstanceChoiceOp, AttachOp, MemOp, InvalidValueOp, SeqMemOp,
               CombMemOp, MemoryPortOp, MemoryDebugPortOp, MemoryPortAccessOp,
-              RefDefineOp, RefForceOp, RefForceInitialOp, RefReleaseOp,
-              RefReleaseInitialOp, LayerBlockOp, GenericIntrinsicOp>(
-            [&](auto op) { emitStatement(op); })
+              DomainDefineOp, RefDefineOp, RefForceOp, RefForceInitialOp,
+              RefReleaseOp, RefReleaseInitialOp, LayerBlockOp,
+              GenericIntrinsicOp>([&](auto op) { emitStatement(op); })
```
seldridge, line 758: Rant: we should alpha-sort this in a follow-on.

lib/Dialect/FIRRTL/Export/FIREmitter.cpp
seldridge, line 1: Thank you for keeping the emitter up to date!

test/Dialect/FIRRTL/parse-errors.fir
seldridge, line 1: Very nice, exhaustive tests!

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -3713,6 +3718,41 @@ ParseResult FIRStmtParser::parseMatch(unsigned matchIndent) {
...
+    return emitError(loc) << "expected domain-type expression, got " << type;
+
+  return success();
+}
+
+/// domain_exp ::= static_domain_exp
+ParseResult FIRStmtParser::parseDomainExp(Value &result, const Twine &message) {
+  // In the future, we will have other sorts of domain expressions.
+  return parseStaticDomainExp(result, message);
+}
```
seldridge: I'd probably drop this given that it is just wrapping static. We can add it back when we have a user.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -3713,6 +3718,41 @@ ParseResult FIRStmtParser::parseMatch(unsigned matchIndent) {
...
+  if (moduleContext.resolveSymbolEntry(result, entry, loc, false)) {
+    StringRef field;
+    if (parseToken(FIRToken::period, "expected '.' in field reference") ||
+        parseFieldId(field, "expected field name") ||
+        moduleContext.resolveSymbolEntry(result, entry, field, loc))
+      return failure();
+  }
+
+  if (parseOptionalExpPostscript(result, false))
+    return failure();
```
seldridge: I think this is the code case that is allowing for `[ ... ]` parsing (as well as a lot of other stuff. Would it be better to just _not_ handle this? A test would be good, but I think a test is impossible to write.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
```diff
@@ -4027,6 +4067,23 @@ ParseResult FIRStmtParser::parsePathExp(Value &result) {
...
 }
 
+/// domain_define ::= 'domain_define' static_domain_exp '=' domain_exp info?
+ParseResult FIRStmtParser::parseDomainDefine() {
+  auto startTok = consumeToken(FIRToken::kw_domain_define);
+  auto startLoc = startTok.getLoc();
+  locationProcessor.setLoc(startLoc);
+
+  Value dest, src;
+  if (requireFeature(missingSpecFIRVersion, "domains", startLoc) ||
```
seldridge, line 4070: Thank you for doing version gating!

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -4168,6 +4168,20 @@ LogicalResult PropAssignOp::verify() {
...
+LogicalResult DomainDefineOp::verify() {
+  if (failed(checkConnectFlow(*this)))
+    return failure();
+
+  for (auto *user : getDest().getUsers()) {
+    auto connection = dyn_cast<FConnectLike>(user);
+    if (connection && connection != *this && connection.getDest() == getDest())
+      return emitError("destination domain cannot be reused by multiple "
+                       "operations, it can only capture a unique dataflow");
+  }
```
seldridge: This verification shows up for `RefDefineOp`. Consider raising into a static function or some other utility.

seldridge (approved):
LGTM modulo a few nits.

Consider adding a test to the `round-trip.mlir`.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
rwy7: Honestly I think it's better to parse more than necessary here. That way, when we present an error to the user, it can be semantic, rather than syntactic, eg `expected domain-type expression, got !firrtl.uint<1>` vs `unexpected [`.  Besides, I was pretty sure we wanted to store domains in aggregates.

lib/Dialect/FIRRTL/Import/FIRParser.cpp
rwy7: dropped.

test/Dialect/FIRRTL/errors.mlir
```diff
@@ -3073,3 +3077,77 @@ firrtl.circuit "WrongNonDomainPortInfo" {
...
+// -----
+
+// Unable to determine domain type of destination.
+
+firrtl.circuit "Top" {
+  firrtl.domain @ClockDomain {}
+  firrtl.domain @PowerDomain {}
+  firrtl.module @Top(in %i: !firrtl.domain of @ClockDomain) {
+    %o = firrtl.wire : !firrtl.domain
+    // expected-error @below {{could not determine domain-type of destination}}
```
dtzSiFive, line 3136: Came to check for these tests!

I was going to suggest rejecting, e.g., unsafe_domain_cast as define target.
But probably that's source flow and would fail flow checking anyway, and if not then since it's not blockarg/inst port, we'd reject same as wires are rejected here.

PR #9068, opened 2025-10-07
[python][Synth] Provide a fine-grained APIs for path queries

uenoku:
This commit refactors the LongestPathAnalysis API to improve clarity and expose external path query methods. The main change renames "closed paths" to "internal paths" to better reflect their meaning as register-to-register paths within a module.

The C API and Python bindings now expose getOpenPathsFromInputPortsToInternal() and getOpenPathsFromInternalToOutputPorts(), allowing users to query input-to-register and register-to-output paths separately. The Python getAllPaths() method now composes results from these three path query methods rather than calling a single C API function.


integration_test/Bindings/Python/dialects/synth.py
```diff
@@ -147,10 +147,51 @@ def build_top(module):
...
     c1.merge(c2)
     # CHECK-NEXT: merge: True
     print("merge:", len(c1) == s1 + s2)
+
+with Context() as ctx, Location.unknown():
+  circt.register_dialects(ctx)
+
+  m = Module.create()
+  with InsertionPoint(m.body):
+    i4 = IntegerType.get_signless(2)
```
uenoku:
```suggestion
    i2 = IntegerType.get_signless(2)
```

PR #9073, opened 2025-10-09
[MooreToCore] Support slicing of nested arrays

SimonEbner:
Added a testcase that is the result of parsing the following sv code:

```sv
module bug (
  input logic [1:0][1:0][1:0] in,
  output logic [1:0] out
);
assign out = in[0][0][1:0];

endmodule
```

The current MooreToCore pass translates the sv ode incorrectly to
```
hw.module @bug(in %in : !hw.array<2xarray<2xi2>>, out out : i2) {
  %false = hw.constant false
  %0 = builtin.unrealized_conversion_cast %in : !hw.array<2xarray<2xi2>> to !hw.array<2xi2>
  %1 = hw.array_get %0[%false] : !hw.array<2xi2>, i1
  hw.output %1 : i2
}
```

Adding a check if operand and result array have the same level of nesting in the ExtractOpConversion fixes the bug.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -746,7 +746,8 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
       int32_t width = llvm::Log2_64_Ceil(arrTy.getNumElements());
       int32_t inputWidth = arrTy.getNumElements();
 
-      if (auto resArrTy = dyn_cast<hw::ArrayType>(resultType)) {
+      if (auto resArrTy = dyn_cast<hw::ArrayType>(resultType);
+          resArrTy && resArrTy != arrTy.getElementType()) {
         int32_t elementWidth = hw::getBitWidth(arrTy.getElementType());
         if (elementWidth < 0)
           return failure();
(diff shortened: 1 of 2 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -731,7 +743,8 @@ struct ExtractOpConversion : public OpConversionPattern<ExtractOp> {
       int32_t width = llvm::Log2_64_Ceil(arrTy.getNumElements());
       int32_t inputWidth = arrTy.getNumElements();
 
-      if (auto resArrTy = dyn_cast<hw::ArrayType>(resultType)) {
+      if (auto resArrTy = dyn_cast<hw::ArrayType>(resultType);
+          resArrTy && arrayNestingLevel(resArrTy) == arrayNestingLevel(arrTy)) {
```
fabianschuiki: I remember @jpienaar fixing a similar issue for `ExtractRefOp` a few weeks ago: 

https://github.com/llvm/circt/pull/8942/files#diff-69352128fb82df4324ed3a5ea9e15dda3bd643c2c183d66e8e1bcd15d01246f9R838-R841

Do you think we could use a similar way of checking for slicing vs. element access, by checking whether the `arrType.getElementType() != resultType`? That would avoid having to reason about nesting levels of arrays, and would turn this into a quick pointer equality check.

SimonEbner: That seems way better. I updated the code. Initially, I wasn't sure if there may be edge-cases I hadn't thought about. All tests pass though.

SimonEbner:
Thanks! I rebased to head to resolve the conflict and pushed my branch again

PR #9074, opened 2025-10-09
[FIRRTL] Add domain verification to instance-like ops

rwy7:
Add checking to ensure that the domain information on instances lines up with what is recorded on the referenced module.


dtzSiFive (approved):
Verifier and test changes LGTM.

The builder changes -- they seem part of a related/necessary (but not in PR scope per title?) change: adding Domain support for InstanceChoiceOp's.

Please either mention this change in the title/commit or split out :+1:.

(and consider positive test for instance choice op domain support?)

rwy7:
pulled instance choice domain support out to its own PR: https://github.com/llvm/circt/pull/9197

PR #9077, opened 2025-10-10
[ImportVerilog][MooreToCore] Implement CHandle import and lowering

Scheremo:
Add end-to-end support for SystemVerilog `chandle`:
- Import from Slang as `moore.chandle`.
- Lower to `!llvm.ptr` in MooreToCore.
- Treat `ref<chandle>` as a *by-ref SSA* value (no `hw.inout`).
- Default-initialize to `llvm.mlir.zero : !llvm.ptr`.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -2040,20 +2040,12 @@ namespace {
 struct MooreToCorePass
     : public circt::impl::ConvertMooreToCoreBase<MooreToCorePass> {
   void runOnOperation() override;
-  void getDependentDialects(mlir::DialectRegistry &r) const override;
 };
 } // namespace
 
 /// Create a Moore to core dialects conversion pass.
 std::unique_ptr<OperationPass<ModuleOp>> circt::createConvertMooreToCorePass() {
   return std::make_unique<MooreToCorePass>();
-}
-
-void MooreToCorePass::getDependentDialects(mlir::DialectRegistry &r) const {
-  r.insert<mlir::LLVM::LLVMDialect, comb::CombDialect, hw::HWDialect,
-           seq::SeqDialect, llhd::LLHDDialect, ltl::LTLDialect,
-           mlir::BuiltinDialect, mlir::math::MathDialect, sim::SimDialect,
-           verif::VerifDialect, scf::SCFDialect>();
 }
 
 /// This is the main entrypoint for the Moore to Core conversion pass.
--- a/include/circt/Conversion/Passes.td
+++ b/include/circt/Conversion/Passes.td
@@ -494,6 +494,7 @@ def ConvertMooreToCore : Pass<"convert-moore-to-core", "mlir::ModuleOp"> {
     "mlir::cf::ControlFlowDialect",
     "mlir::scf::SCFDialect",
     "mlir::math::MathDialect",
+    "mlir::LLVM::LLVMDialect",
     "sim::SimDialect",
     "verif::VerifDialect",
   ];


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -540,6 +543,20 @@ struct VariableOpConversion : public OpConversionPattern<VariableOp> {
...
 
+    // Handle CHandle types
+    if (isa<mlir::LLVM::LLVMPointerType>(resultType)) {
+      // Use converted initializer if present; otherwise synthesize null ptr.
+      Value init = adaptor.getInitial();
+      if (!init)
+        init = rewriter.create<mlir::LLVM::ZeroOp>(loc, resultType);
+      else if (init.getType() != resultType)
+        init = typeConverter->materializeTargetConversion(rewriter, loc,
+                                                          resultType, init);
```
fabianschuiki: The `adaptor.getInitial()` should already give you the lowered version of the `init` operand. That will already be of `resultType`. If it isn't, that's a bug 😁 and the dialect conversion framework would complain, too.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1798,10 +1816,25 @@ static void populateTypeConversion(TypeConverter &typeConverter) {
...
+
   typeConverter.addConversion([&](RefType type) -> std::optional<Type> {
-    if (auto innerType = typeConverter.convertType(type.getNestedType()))
+    if (auto innerType = typeConverter.convertType(type.getNestedType())) {
       if (hw::isHWValueType(innerType))
         return hw::InOutType::get(innerType);
+      // Otherwise, if we intentionally model this "by-ref as SSA", pass it
+      // through, e.g. CHandleType.
+      if (isByRefAsSSA(innerType))
+        return innerType;
```
fabianschuiki: Yikes, I forgot about the fact that we use `hw.inout` as a stand-in for a pointer at the core dialect level. I've run into similar problems when trying to support `!llhd.time` values in a variable. Does just returning the inner type here work? For example, if you assign to the variable, or pass it around in a function, `MooreToCore` will generate `llhd.prb` and `llhd.drv` operations that expect an `!hw.inout` type. Might be worth just not supporting `ref<chandle>` for now, WDYT? Things like function arguments should still work for a few first tests, since the Moore dialect tries to replace their `ref<T>` with the `T` value if they are only driven once.

In the future we should probably just create an `!llhd.ref<T>` type, and replace all uses of `!hw.inout<T>` with that. That'll also fix the `time` variable issue.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -2014,6 +2048,13 @@ std::unique_ptr<OperationPass<ModuleOp>> circt::createConvertMooreToCorePass() {
   return std::make_unique<MooreToCorePass>();
 }
 
+void MooreToCorePass::getDependentDialects(mlir::DialectRegistry &r) const {
+  r.insert<mlir::LLVM::LLVMDialect, comb::CombDialect, hw::HWDialect,
+           seq::SeqDialect, llhd::LLHDDialect, ltl::LTLDialect,
+           mlir::BuiltinDialect, mlir::math::MathDialect, sim::SimDialect,
+           verif::VerifDialect, scf::SCFDialect>();
+}
+
```
fabianschuiki: I think you can just extend the list of `dependentDialects` in the `Passes.td` file that lists all the conversions. Then you don't need to plumb this through manually.

test/Conversion/MooreToCore/basic.mlir
```diff
@@ -1278,4 +1278,9 @@ func.func @SeverityToPrint() {
 
   return
 }
-
+// CHECK-LABEL: func.func @CHandle
+func.func @CHandle(){
+  // CHECK: [[TEST:%.*]] = llvm.mlir.zero : !llvm.ptr
+  %test = moore.variable : <chandle>
+  return
+}
```
fabianschuiki: I think you can tweak this to 
```mlir
func.func @CHandle(%arg0: !moore.chandle) {
  return
}
```
to see if the conversion works without triggering that problem around `hw.inout` not accepting `!llvm.ptr` or `!llhd.time` types at the moment.

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Fair point, removed it.

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Thanks, wasn't aware this is another tablegen-magic thing 😄

test/Conversion/MooreToCore/basic.mlir
Scheremo: Yep, switched the test to this format.

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: I think not supporting `ref<chandle>` is definitely a safe way of handling this.
I added an explicit error case for RefType with underlying (because converted CHandle) `!llvm.ptr` so we don't need to rediscover where the footgun is 😅

PR #9078, opened 2025-10-10
[LLHD] Remove unused passes and unused memory types and ops

fabianschuiki:
The circt-verilog pipeline uses only a subset of the passes in the LLHD dialect. A few of the passes have been superseded by a reworked version which currently coexists with the original pass. Remove the old passes.

We also don't use the LLHD `PtrType` and the related operations anywhere in CIRCT. It looks like SV and VHDL are going to be easier to model if we treat everything as a signal, but have blocking assignments to model the behavior of variables. Remove these ops and the pointer type.


fabianschuiki:
Good point! I guess we still have the Deseq pass, but it might make sense to adjust the name in the diagram to match that!

Issue #9075, opened 2025-10-09
[circt-verilog-lsp] MaxCapForcesFlushDuringContinuousTyping fails with valgrind


uenoku:
CI run: https://github.com/llvm/circt/actions/runs/18375617219

The test spawned a background [typer](https://github.com/llvm/circt/blob/f585d9a842844d6b750bbda5e022fd5d328db326/unittests/Tools/circt-verilog-lsp-server/Utils/PendingChangesTest.cpp#L169-L175) thread to simulate continuous typing, but immediately scheduled a debounce check without waiting for the typer to actually run. Due to non-deterministic thread scheduling, the typer might not have enqueued any changes yet, causing the test to fail unpredictably.

uenoku:
CC: @Scheremo

PR #9080, opened 2025-10-10
[circt-verilog-lsp-server] Simplify MaxCapForcesFlushDuringContinuousTyping by removing bakground thread

uenoku:
This commit refactors the MaxCapForcesFlushDuringContinuousTyping test to eliminate the background thread used for simulating continuous typing. The cap is increased to pass valgrind (at least on CI environment...). Not proud of the change and we should seek a better way to reduce flaky parts. 

Fix https://github.com/llvm/circt/issues/9075. 

CI: https://github.com/llvm/circt/actions/runs/18416091701/job/52480002926

### DIFF
--- a/unittests/Tools/circt-verilog-lsp-server/Utils/PendingChangesTest.cpp
+++ b/unittests/Tools/circt-verilog-lsp-server/Utils/PendingChangesTest.cpp
@@ -12,7 +12,6 @@
 
 #include "PendingChanges.h"
 
-#include <atomic>
 #include <chrono>
 #include <condition_variable>
 #include <mutex>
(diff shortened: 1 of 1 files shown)


Scheremo:
> This commit refactors the MaxCapForcesFlushDuringContinuousTyping test to eliminate the background thread used for simulating continuous typing. The cap is increased to pass valgrind (at least on CI environment...). Not proud of the change and we should seek a better way to reduce flaky parts.
> 
> Fix #9075.
> 
> CI: https://github.com/llvm/circt/actions/runs/18416091701/job/52480002926

Thanks for pointing this out! In hindsight this test looks particularly fragile, but probably most of these unit tests are too optimistic with assuming scheduling stability.
I think we can avoid time-based races by factoring out `std::chrono::steady_clock::now` and replace it with a callable that we can manually override for testing. That way we can get rid of threading in unittesting altogether. E.g.:

```
using Clock = std::chrono::steady_clock;
using NowFn = std::function<Clock::time_point()>;
class PendingChangesMap {
public:
  explicit PendingChangesMap(NowFn now = []{ return Clock::now(); })
      : nowFn(std::move(now)) {}
...
private:
  NowFn nowFn;


struct ManualClock {
  using Clock = std::chrono::steady_clock;
  Clock::time_point t = Clock::time_point{};
  Clock::time_point now() const { return t; }
  void advanceMs(int ms) { t += std::chrono::milliseconds(ms); }
};

TEST(PendingChangesMapTest, MaxCapForcesFlushDuringContinuousTyping) {
  ManualClock mc;
  PendingChangesMap pcm(2, [&]{ return mc.now(); });

  DebounceOptions opt;
  opt.disableDebounce = false;
  opt.debounceMinMs = 100;
  opt.debounceMaxMs = 200;

  const auto *key = "d.sv";
  auto p = makeChangeParams(key, 1);
  pcm.enqueueChange(p);

  for (int v = 2; v < 20; ++v) {
    mc.advanceMs(50);                            // time passes
    pcm.enqueueChange(makeChangeParams(key, v)); // “user typed”
  }

  CallbackCapture first;
  pcm.debounceAndThen(p, opt, [&](std::unique_ptr<PendingChanges> r) {
    first.set(std::move(r));
  });
  mc.advanceMs(100);
  ASSERT_TRUE(first.waitFor(std::chrono::milliseconds(50)));
  EXPECT_EQ(first.got, nullptr);

  mc.advanceMs(250); // > 200ms cap since last schedule/key activity
  CallbackCapture second;
  auto p2 = makeChangeParams(key, /*version=*/999);
  pcm.debounceAndThen(p2, opt, [&](std::unique_ptr<PendingChanges> r) {
    second.set(std::move(r));
  });

  ASSERT_TRUE(second.waitFor(std::chrono::milliseconds(50)));
  ASSERT_TRUE(second.got != nullptr);
  EXPECT_GE(second.got->version, 2);
  EXPECT_GE(second.got->changes.size(), 1u);
}
```

I can take a stab at that; if you prefer we can also revert the LSP commit and re-land once we have a better solution.

uenoku:
Thank you for the review! Overriding clock in tests makes sense. Actually I explored a bit of that direction but it was challenging to model sleep in the async thread https://github.com/llvm/circt/blob/0dd8d0daea7b23958b59d0650121585190a7bab4/lib/Tools/circt-verilog-lsp-server/Utils/PendingChanges.cpp#L89-L91 with a manual clock. Maybe some inconsistency is fine in tests though.

PR #9081, opened 2025-10-10
[LLHD] Add new RefType to replace hw::InOutType

fabianschuiki:
Add a `RefType` to the LLHD dialect and replace all uses of HW's `InOutType` across the LLHD dialect and the ImportVerilog pipeline. We have been abusing the inout type to mean "signal/variable reference", which becomes problematic if we are trying to place `!llhd.time` or `!llvm.ptr` values behind such a reference. (The HW inout type requires the inner type to be a plain hardware value.)

This change now allows the circt-verilog frontend to generate variables with the `time` type and treat them just like any other type.

Most of the changes here are just systematic name replacements. I have also removed a few `qualified(...)` wrappers around types and made sure that when an op is known to accept or return a ref type, the printer would just print `<T>` instead of `!llhd.ref<T>`. This is the default behavior for unambiguous types, but sometimes this gets lost when explicitly askign for qualified types, or using `functional-type`.

### DIFF
--- a/include/circt/Dialect/LLHD/IR/LLHDTypes.td
+++ b/include/circt/Dialect/LLHD/IR/LLHDTypes.td
@@ -35,6 +35,46 @@ def LLHDTimeType : LLHDType<"Time"> {
   let mnemonic = "time";
 }
 
+def RefType : LLHDType<"Ref"> {
+  let summary = "a reference to a signal or variable";
+  let description = [{
+    Represents a reference to a value. Signals return a reference to the
+    underlying value, which allows them to be probed or driven.
+  }];
+  let mnemonic = "ref";
+  let parameters = (ins "Type":$nestedType);
+  let assemblyFormat = "`<` $nestedType `>`";
+  let builders = [
+    AttrBuilderWithInferredContext<(ins "Type":$nestedType), [{
+      return $_get(nestedType.getContext(), nestedType);
+    }]>
+  ];
+}
+
+//===----------------------------------------------------------------------===//
+// Type Constraints
+//===----------------------------------------------------------------------===//
+
+
+/// A ref type with a specific nested type.
+class RefTypeOf<Type type> : ConfinedType<RefType,
+  [SubstLeaves<"$_self", "cast<llhd::RefType>($_self).getNestedType()",
+               type.predicate>],
+  "ref of " # type.summary, "llhd::RefType"
+> {
+  Type nestedType = type;
+}
+
+class TypeMatchesRefNestedType<string type, string refType> : TypesMatchWith<
+  type # " type must match " # refType # " ref element type",
+  type, refType, "RefType::get($_self)"
+>;
+
+class RefNestedTypeMatchesType<string refType, string type> : TypesMatchWith<
+  refType # " ref element type must match " # type # " type",
+  refType, type, "cast<RefType>($_self).getNestedType()"
+>;
+
 //===----------------------------------------------------------------------===//
 // Attribute declarations
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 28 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1824,18 +1823,8 @@ static void populateTypeConversion(TypeConverter &typeConverter) {
...
-      // reference of a CHandle; return an error for now.
-      if (isa<mlir::LLVM::LLVMPointerType>(innerType)) {
-        mlir::emitError(mlir::UnknownLoc::get(type.getContext()))
-            << "Emission of references of LLVMPointerType is currently "
-               "unsupported!";
-        return {};
-      }
-    }
+    if (auto innerType = typeConverter.convertType(type.getNestedType()))
+      return llhd::RefType::get(innerType);
```
Scheremo, line 1828: Much nicer!

include/circt/Dialect/LLHD/IR/LLHDTypes.td
```diff
@@ -35,6 +35,46 @@ def LLHDTimeType : LLHDType<"Time"> {
...
 }
 
+def RefType : LLHDType<"Ref"> {
+  let summary = "a reference to a signal or variable";
+  let description = [{
+    Represents a reference to a value. Signals return a reference to the
+    underlying value, which allows them to be probed or driven.
+  }];
+  let mnemonic = "ref";
+  let parameters = (ins "::mlir::Type":$elementType);
```
Scheremo: Nit: In moore we call the `elementType` `nestedType`. It might be clearer to use the same name for both.

include/circt/Dialect/LLHD/IR/LLHDTypes.td
```diff
@@ -35,6 +35,46 @@ def LLHDTimeType : LLHDType<"Time"> {
...
+}
+
+class TypeMatchesRefElementType<string type, string refType> : TypesMatchWith<
+  type # " type must match " # refType # " ref element type",
+  type, refType, "RefType::get($_self)"
+>;
+
+class RefElementTypeMatchesType<string refType, string type> : TypesMatchWith<
+  refType # " ref element type must match " # type # " type",
+  refType, type, "cast<RefType>($_self).getElementType()"
```
Scheremo: Should this not be `llvm::cast`?

test/Conversion/MooreToCore/basic.mlir
```diff
@@ -479,9 +479,22 @@ moore.module @Variable() {
...
+
+  // CHECK: [[TMP:%.+]] = llhd.constant_time <0
+  // CHECK: llhd.sig [[TMP]] : !llhd.time
+  moore.variable : <!moore.time>
+
+  // CHECK: [[TMP:%.+]] = llhd.constant_time
+  // CHECK: llhd.sig [[TMP]] : !llhd.time
+  %c42_fs = moore.constant_time 42 fs
+  moore.variable %c42_fs : <!moore.time>
+
```
Scheremo, line 497: Love to see this!

include/circt/Dialect/LLHD/IR/LLHDTypes.td
fabianschuiki: This inherits the namespacing and includes from `LLHDTypes.h`, which pulls in the header that TableGen produces. That reduces the verbosity a bit.

include/circt/Dialect/LLHD/IR/LLHDTypes.td
fabianschuiki: Done

PR #9082, opened 2025-10-10
[circt-lsp-server] Make time source injectable

Scheremo:
Introduce an injectable clock and test-only wait shim to remove real sleeps
from `PendingChanges` logic in tests, while preserving production behavior.

### PendingChangesMap Changes

- Add an **injectable time source**:
  - `using SteadyClock = std::chrono::steady_clock;`
  - `using NowFn = std::function<SteadyClock::time_point()>;`
  - Store `NowFn nowFn` and a `bool useManualClock` in `PendingChangesMap`.
- Replace direct calls to `std::chrono::steady_clock::now()` with `nowFn()`.
- Add `waitForMinMs(uint64_t ms, SteadyClock::time_point start)`:
  - **Prod path**: `std::this_thread::sleep_for(ms)`.
  - **Test path** (`useManualClock`): busy-wait with `std::this_thread::yield()` until `nowFn()` reaches target.
- New ctor overload for tests:
  - `PendingChangesMap(unsigned maxThreads, NowFn now)` -> enables manual clock mode.
- Dtor now calls `abort()` to ensure all pending work is cleared on teardown.
- Replace sleep in debounce worker:
  - `sleep_for(ms)` -> `waitForMinMs(ms, scheduleTime)`.

### PendingChangesMap Unit Test Changes

- Add `ManualClock` with manual advancement (`advanceMs`) and `now()` accessor.
- Utility `advanceTestTime(clock, ms)` that advances test time and yields.
- `CallbackCapture::waitFor()` now uses a tight fixed 1ms window (no per-call timeout arg).
- Refactor tests to:
  - Construct `PendingChangesMap` with manual clock (`PendingChangesMap(2, [&]{ return clock.now(); })`).
  - Advance time deterministically instead of sleeping.
  - Validate debounce min/obsolete/max-cap behaviors without flakiness.
- Adjust debounce cap in `MaxCapForcesFlushDuringContinuousTyping` to 100ms to tighten assertions.
- Sprinkle end-of-test large time advances to flush any queued workers.

### DIFF
--- a/unittests/Tools/circt-verilog-lsp-server/Utils/PendingChangesTest.cpp
+++ b/unittests/Tools/circt-verilog-lsp-server/Utils/PendingChangesTest.cpp
@@ -22,6 +22,20 @@ using namespace std::chrono_literals;
 
 namespace {
 
+/// Manual, monotonic clock for tests. Time is advanced manually to avoid race
+/// conditions in threads
+struct ManualClock {
+  SteadyClock::time_point t = SteadyClock::time_point{}; // start at epoch
+  SteadyClock::time_point now() const { return t; }
+  void advanceMs(std::chrono::milliseconds ms) { t += ms; }
+};
+
+static inline void advanceTestTime(ManualClock &clock,
+                                   std::chrono::milliseconds ms) {
+  clock.advanceMs(ms);
+  std::this_thread::yield(); // let the worker run
+}
+
 /// Tiny helper to await a single callback result in tests.
 struct CallbackCapture {
   std::mutex m;
@@ -39,9 +53,11 @@ struct CallbackCapture {
   }
 
   // Wait up to timeout for a callback; returns true if called.
-  bool waitFor(std::chrono::milliseconds timeout) {
+  bool waitFor() {
     std::unique_lock<std::mutex> lock(m);
-    return cv.wait_for(lock, timeout, [&] { return called; });
+    // Keep the timeout very small, as all time references are anyway driven
+    // from the unit tests
+    return cv.wait_for(lock, 1ms, [&] { return called; });
   }
 };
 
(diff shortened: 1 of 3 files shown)


uenoku:
Cool, thank you for improving this! We need to make sure not to use raw now()/sleep_for() but that sounds reasonable

Scheremo:
> Cool, thank you for improving this! We need to make sure not to use raw now()/sleep_for() but that sounds reasonable

Thanks for spotting the issue and fixing it on short notice! 
My current approach is to centralize time control in the unit test, so each thread "runs on unit test time". This should remove any race conditions on time alone. I think the draft is almost complete, so please feel free to let me know if you have any feedback 😄

unittests/Tools/circt-verilog-lsp-server/Utils/PendingChangesTest.cpp
```diff
@@ -209,12 +246,15 @@ TEST(PendingChangesMapTest, MissingKeyYieldsNullptr) {
     cap.set(std::move(r));
   });
 
-  ASSERT_TRUE(cap.waitFor(400ms));
+  advanceTestTime(clock, 30ms);
+
+  ASSERT_TRUE(cap.waitFor());
   EXPECT_EQ(cap.got, nullptr);
```
uenoku, line 257: abort missiong?

Scheremo:
> LGTM, thank you for improving this! Maybe we might want to check abort is called in deconstructor of PendingChangesMap (RAII also works, but I personally prefer requiring explicit `abort` ).

That's a good point; I added a destructor that calls abort. This should help avoid accidental thread orphaning, and if there are no threads to wait on, this should also not cost anything if `abort` has been called prior to destruction.

PR #9085, opened 2025-10-12
[ImportVerilog] Add delayed assignment support

fabianschuiki:
Add the `moore.delayed_assign` and `moore.delayed_nonblocking_assign` ops to represent `assign #1ns a = b` and `a <= #1ns b` in SystemVerilog, respectively.

Lower the new delay ops to `llhd.drv` with the appropriate delay value.

Add support for blocking assignments with intra-assignment timing control to the ImportVerilog conversion. These are pretty trivial, since the blocking effects of the timing control are simply inserted in between computing the right-hand side and assigning to the left-hand side.

Also add support for non-blocking assignments with intra-assignment delays. These require the new `moore.delayed_nonblocking_assign`, since the operation cannot suspend execution of the surrounding process. Instead, the assignment has to be added to the event queue and executed at a later point in time. Theoretically, the user could type wild things here, like `a <= repeat(5) @(posedge b or negedge c) d`, which would require us to spawn a separate "thread" to determine when all the events have occurred and the assignment can take place. We don't support any of that for now, because this is just utterly deranged.

Also add support for delayed continuous assignments. This is pretty trivial since SystemVerilog only allows for simple delay values, such as `assign #1ns a = b`. This requires the new `moore.delayed_assign` op.

### DIFF
--- a/lib/Conversion/ImportVerilog/TimingControls.cpp
+++ b/lib/Conversion/ImportVerilog/TimingControls.cpp
@@ -155,7 +155,7 @@ struct LTLClockControlVisitor {
 // be handled by `handleDelayControl`.
 static LogicalResult handleRoot(Context &context,
                                 const slang::ast::TimingControl &ctrl,
-                                moore::WaitEventOp &implicitWaitOp) {
+                                moore::WaitEventOp *implicitWaitOp) {
   auto &builder = context.builder;
   auto loc = context.convertLocation(ctrl.sourceRange);
 
@@ -178,7 +178,9 @@ static LogicalResult handleRoot(Context &context,
     // empty wait op and let `Context::convertTimingControl` populate it once
     // the statement has been lowered.
   case TimingControlKind::ImplicitEvent:
-    implicitWaitOp = moore::WaitEventOp::create(builder, loc);
+    if (!implicitWaitOp)
+      return mlir::emitError(loc) << "implicit events cannot be used here";
+    *implicitWaitOp = moore::WaitEventOp::create(builder, loc);
     return success();
 
     // Handle event control.
@@ -207,6 +209,11 @@ static LogicalResult handleRoot(Context &context,
 }
 
 LogicalResult
+Context::convertTimingControl(const slang::ast::TimingControl &ctrl) {
+  return handleRoot(*this, ctrl, nullptr);
+}
+
+LogicalResult
 Context::convertTimingControl(const slang::ast::TimingControl &ctrl,
                               const slang::ast::Statement &stmt) {
   // Convert the timing control. Implicit event control will create a new empty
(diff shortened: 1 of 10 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1470,21 +1470,31 @@ struct AssignedVariableOpConversion
...
+    Value delay;
+    if constexpr (std::is_same_v<OpTy, ContinuousAssignOp> ||
+                  std::is_same_v<OpTy, BlockingAssignOp>) {
+      delay = llhd::ConstantTimeOp::create(
+          rewriter, op->getLoc(),
+          llhd::TimeAttr::get(op->getContext(), 0U, "ns", 0, 1));
+    } else if constexpr (std::is_same_v<OpTy, NonBlockingAssignOp>) {
+      delay = llhd::ConstantTimeOp::create(
+          rewriter, op->getLoc(),
+          llhd::TimeAttr::get(op->getContext(), 0U, "ns", 1, 0));
```
uenoku, line 1493: Is it correct to assume 1ns delay?

lib/Conversion/ImportVerilog/TimingControls.cpp
```diff
@@ -178,7 +178,9 @@ static LogicalResult handleRoot(Context &context,
     // empty wait op and let `Context::convertTimingControl` populate it once
     // the statement has been lowered.
   case TimingControlKind::ImplicitEvent:
-    implicitWaitOp = moore::WaitEventOp::create(builder, loc);
+    if (!implicitWaitOp)
+      return mlir::emitError(loc) << "implicit events cannot be used here";
```
uenoku, line 182: Can we add a test for this?

lib/Conversion/MooreToCore/MooreToCore.cpp
fabianschuiki: The LLHD time values are a bit quirky (and in need of some rework). This creates a `0ns 1d 0e` delay :smiling_face_with_tear:. I'll add a comment!

Originally, LLHD tried to explicitly model delta delays for VHDL interoperability, and it had an epsilon delay to accomodate the different scheduling regions of Verilog. But since moving everything to MLIR, I'm pretty sure we can make time values _just_ be actual seconds, and deal with delta delays and the scheduling regions differently.

lib/Conversion/ImportVerilog/TimingControls.cpp
fabianschuiki: Good point. Done!

lib/Conversion/MooreToCore/MooreToCore.cpp
uenoku: Ah I misread` that :) Make sense!

PR #9087, opened 2025-10-13
[FIRRTL] Add fields to domains

seldridge:
Extend domains to allow for them to contain fields.  These fields provide
a schema of information that all domains of certain kind must have.  This
models information that a user must provide for an input domain or is
provided for an output domain.

Opt for a "struct-like" representation as opposed to using input/output
ports like FIRRTL classes.  That said, these are intended to be lowered to
classes in a future patch.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
@@ -209,4 +209,19 @@ def InternalPathArrayAttr
 
 def LayerArrayAttr : TypedArrayRefAttrBase<SymbolRefAttr, "an array of layers">;
 
+def DomainFieldAttr : AttrDef<FIRRTLDialect, "DomainField"> {
+  let mnemonic = "domain.field";
+  let summary = "A single field of a domain";
+  let parameters = (
+    ins "::mlir::StringAttr":$name,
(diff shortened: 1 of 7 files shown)


dtzSiFive:
LGTM!
It might make sense to limit the field to a Property-type TypeAttr, ODS or verifier, since looks like at the FIRRTL text level that's an error to not be a property type.

If the field names must be unique (they ofc  "must" if indexing is string-based?) consider verifying that as well.

Nice focused incremental addition!

seldridge:
> It might make sense to limit the field to a Property-type TypeAttr, ODS or verifier, since looks like at the FIRRTL text level that's an error to not be a property type.

Good point. I started down this path and couldn't sort through the ODS / header inclusion to get it to compile.

> If the field names must be unique (they ofc "must" if indexing is string-based?) consider verifying that as well.

Great idea. I'll add this in a follow-up.

seldridge:
@dtzSiFive wrote: 
> It might make sense to limit the field to a Property-type TypeAttr, ODS or verifier, since looks like at the FIRRTL text level that's an error to not be a property type.

PR here: https://github.com/llvm/circt/pull/9094 The primary review request on #9094 is on the reorganization of the ODS files and headers.

PR #9088, opened 2025-10-13
[ImportVerilog][Moore] Add real-to-int & int-to-real operators

Scheremo:
This commit introduces explicit conversion operations between real and integer types in the Moore dialect, following IEEE 1800-2023 §6.24.1 *Cast operator* and §6.12 *Real, shortreal and realtime data types*. It also extends VerilogImport to recognize and materialize these conversions during AST translation.

- **Dialect (`MooreOps.td`)**
  - Added `RealToIntOp` for rounding real/shortreal values to two-valued integers.
  - Added `IntToRealOp` for widening integer values to real/shortreal.
  - Both ops marked `Pure` with explicit assembly formats.

- **Import logic (`Expressions.cpp`)**
  - Added conversion handling in `Context::materializeConversion`:
    - `RealType` -> `IntType` -> `RealToIntOp`
    - `IntType` -> `RealType` -> `IntToRealOp`

- **Tests**
  - **SystemVerilog import**
    - Added `basic.sv` tests verifying correct lowering of `real'(expr)` and `int'(expr)` casts for all real/int width combinations (`f32/i32`, `f64/i64`).
    - Updated `builtins.sv` to expect `moore.int_to_real` instead of a generic `conversion` op.
  - **Dialect MLIR**
    - Added `basic.mlir` tests ensuring proper parsing/printing and type mapping for 32- and 64-bit conversions.

### DIFF
--- a/test/Conversion/ImportVerilog/basic.sv
+++ b/test/Conversion/ImportVerilog/basic.sv
@@ -3308,3 +3308,39 @@ task NonBlockingAssignment(
   // CHECK-NEXT: moore.delayed_nonblocking_assign [[A]], [[NOTB]], [[TIME]]
   a <= #1ns ~b;
 endtask
+
+// CHECK-LABEL: func.func private @RealConversion(
+// CHECK-SAME: [[SR:%[^,]+]]: !moore.f32
+// CHECK-SAME: [[LR:%[^,]+]]: !moore.f64
+// CHECK-SAME: [[INT:%[^,]+]]: !moore.i42
+// CHECK-SAME: [[LINT:%[^,]+]]: !moore.i64
+function automatic void RealConversion(shortreal sr, real r, bit[41:0] i, longint longi);
+   // CHECK: [[INTTEST:%.+]] = moore.real_to_int [[SR]] : f32 -> i32
+   int intTest = int'(sr);
+    // CHECK: [[INTTEST2:%.+]] = moore.real_to_int [[LR]] : f64 -> i32
+   int intTest2 = int'(r);
+   // CHECK: [[LINTTEST:%.+]] = moore.real_to_int [[SR]] : f32 -> i64
+   longint longIntTest = longint'(sr);
+   // CHECK: [[LINTTEST2:%.+]] = moore.real_to_int [[LR]] : f64 -> i64
+   longint longIntTest2 = longint'(r);
+
+
+   // CHECK: [[SRTEST:%.+]] = moore.int_to_real [[INT]] : i42 -> f32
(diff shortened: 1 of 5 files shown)


test/Dialect/Moore/basic.mlir
```diff
@@ -449,3 +449,21 @@ func.func @TimeConversion(%arg0: !moore.time, %arg1: !moore.l64) {
...
+}
+
+// CHECK-LABEL: func.func @RealConversion64(%arg0: !moore.f64, %arg1: !moore.i64)
+func.func @RealConversion64(%arg0: !moore.f64, %arg1: !moore.i64) {
+  // CHECK: %0 = moore.real_to_int %arg0 : f64 -> i64
+  %0 = moore.real_to_int %arg0 : f64 -> i64
+  // CHECK: %1 = moore.int_to_real %arg1 : i64 -> f64
+  %1 = moore.int_to_real %arg1 : i64 -> f64
+  return
+}
```
fabianschuiki: Would it make sense to turn the `i32` and `i64` in this example into an `i42` to cover the fact that these are conversions to and from arbitrary integer bit widths?

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -3245,3 +3245,29 @@ function automatic void ConcatSformatf(string testStr, string otherString, ref s
...
+
+   // CHECK: [[SRTEST:%.+]] = moore.int_to_real [[INT]] : i32 -> f32
+   shortreal srTest = shortreal'(i);
+   // CHECK: [[SRTEST2:%.+]] = moore.int_to_real [[LINT]] : i64 -> f32
+   shortreal srTest2 = shortreal'(longi);
+   // CHECK: [[RTEST:%.+]] = moore.int_to_real [[INT]] : i32 -> f64
+   real rTest = real'(i);
+   // CHECK: [[RTEST2:%.+]] = moore.int_to_real [[LINT]] : i64 -> f64
+   real rTest2 = real'(longi);
+endfunction
```
fabianschuiki, line 3346: Would it make sense to turn the `int` and `longint` in this example into a `bit [41:0]` to cover the fact that these are conversions to and from arbitrary integer bit widths?

test/Dialect/Moore/basic.mlir
Scheremo: Sure, why not? Changed them to `i42`.

test/Conversion/ImportVerilog/basic.sv
Scheremo: Sure, why not? Changed `i` to `i42`.

PR #9091, opened 2025-10-14
[ESI] Add serde support for wide integers

mortbopet:
It is *very* frustrating that ser/de through ESI `(U)Int` types in C++ just _doesn't work_. I think that until (if ever) ESI provides an arbitrary-width C++ integer type, it'll be acceptable to fall-back to ser/de through bytearrays for >64-bit int-like types.


mortbopet:
publish run https://github.com/llvm/circt/actions/runs/18490843602

PR #9092, opened 2025-10-14
[ImportVerilog] Generalize materialization of constant real values

Scheremo:
This patch teaches the ImportVerilog conversion to correctly lower SystemVerilog
real and shortreal constant values into `moore.real_constant` operations.

* **MooreOps.td**
  - Relax `RealLiteralOp` result type from fixed `RealF64` to generic
    `RealType` to support both 32-bit and 64-bit real kinds.

* **ImportVerilog/Expressions.cpp**
  - Add `Context::materializeSVReal()` helper to construct `RealLiteralOp`s from
    `slang::ConstantValue` reals.
  - Replace direct creation of an F64 literal with a call to the new helper.
  - Hook `materializeConstant()` to handle `ConstantValue::isReal()` and
    `isShortReal()`.

* **ImportVerilogInternals.h**
  - Declare `materializeSVReal()` alongside `materializeSVInt()`.

* **Tests**
  - Add `testRealLiteral` and `testShortrealLiteral` functions to verify that
    real and shortreal literals produce the expected `moore.real_constant`
    operations with `f64` and `f32` types.

Overall, this enables correct handling of floating-point literals in the
ImportVerilog pipeline.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -1084,8 +1084,7 @@ struct RvalueExprVisitor : public ExprVisitor {
 
   /// Handle real literals.
   Value visit(const slang::ast::RealLiteral &expr) {
-    return moore::RealLiteralOp::create(
-        builder, loc, builder.getF64FloatAttr(expr.getValue()));
+    return context.materializeSVReal(*expr.getConstant(), *expr.type, loc);
   }
 
   /// Helper function to convert RValues at creation of a new Struct, Array or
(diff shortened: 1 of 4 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1499,6 +1499,13 @@ Value Context::convertToBool(Value value) {
   return {};
 }
 
+/// Materialize a Slang real literal as a constant op.
+Value Context::materializeSVReal(const slang::real_t &svreal, Location loc) {
+  auto f64Ty = mlir::Float64Type::get(getContext());
+  mlir::FloatAttr attr = mlir::FloatAttr::get(f64Ty, llvm::APFloat(svreal.v));
+  return moore::RealLiteralOp::create(builder, loc, attr);
+}
+
```
fabianschuiki, line 1526: Should this accept a `slang::ast::Type` and call `materializeConversion` like `materializeSVInt` does? How are `shortreal` constants materialized?

Scheremo: All good questions; the "usual" path to materialization is already implemented when visiting `slang::ast::RealLiteral` in the `ExprVisitor`. 
This materialization is only called when the LHS does not have a type, as in 
```
parameter test = 1.2345;
```
, since the type cannot be inferred in this case - I don't *expect* Slang to ever emit a `shortreal` when the LHS type cannot be determined, because the spec, in Section 5.7.2 says it shouldn't; materializing a `real` should be correct. Constructs like
```
parameter real test = 1.2345;
parameter shortreal test = 1.2345;
```
both already work correctly without this patch 😄

fabianschuiki: Nice, sounds great!

Scheremo: Actually, I just tried with an explicit cast, e.g.
```
parameter test = shortreal'(1.2345);
```
which actually emits a `shortreal`, so we do need a case distinction 😃

fabianschuiki: Maybe the existing support for  `parameter shortreal test = 1.2345;` could use the same `materializeSVReal` code path with the `shortreal` provided as a type?

Scheremo: Now that's a good idea! Refactored to do exactly that.

PR #9093, opened 2025-10-14
[FIRRTL] Update port insertion/erasure API for instance/instance-choice ops 

rwy7:
Align and flush out the functionality for port insertion and erasure on instance and instance-choice op.

Overview of new API for Port Insertions:
  - `cloneWithInsertedPorts`: create a clone of the op with additional inserted ports. Returns the clone. This API _must_ clone, even if no ports are inserted.
  - `cloneWithInsertedPortsAndReplaceUses`: replace the uses of this op with a clone created via the above API. This API _must_ clone, even if no ports are inserted.
  - ~`insertPorts`: the "do what I mean" API: clone with port insertions, replace uses, and then erase the old instance, but only if there are any ports being inserted.~

Then for port erasure, we have a parallel set of functions:
  - `cloneWithErasedPorts`
  - `cloneWithErasedPortsAndReplaceUses`
  - ~`erasePorts`~
  
All ~six~ four functions are implemented for both the InstanceOp and InstanceChoiceOp.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
@@ -139,14 +139,20 @@ def InstanceOp : HardwareDeclOp<"instance", [
     /// Hooks for domain information.
     ArrayAttr getPortDomain(unsigned portIdx);
 
-    /// Builds a new instance with the ports listed in `portIndices` erased, and
-    /// updates any users of the remaining ports to point at the new instance.
-    InstanceOp erasePorts(OpBuilder &builder, const llvm::BitVector &portIndices);
+    /// Clone this instance with inserted ports.
+    InstanceOp cloneWithInsertedPorts(
+      ArrayRef<std::pair<unsigned, PortInfo>> insertions);
 
-    /// Clone the instance op and add ports.  This is usually used in
-    /// conjuction with adding ports to the referenced module. This will emit
-    /// the new InstanceOp to the same location.
-    InstanceOp cloneAndInsertPorts(ArrayRef<std::pair<unsigned, PortInfo>> ports);
+    /// Clone this instance with inserted ports, then replaces uses.
+    InstanceOp cloneWithInsertedPortsAndReplaceUses(
+      ArrayRef<std::pair<unsigned, PortInfo>> insertions);
+
+    /// Clone this instance with erased ports.
+    InstanceOp cloneWithErasedPorts(const llvm::BitVector &erasures);
+
+    /// Clone this instance with erased ports, then replaces uses.
+    InstanceOp cloneWithErasedPortsAndReplaceUses(
+      const llvm::BitVector &erasures);
 
     //===------------------------------------------------------------------===//
     // Instance graph methods
@@ -211,12 +217,6 @@ def InstanceChoiceOp : HardwareDeclOp<"instance_choice", [
     StringAttr getPortName(size_t resultNo) {
       return cast<StringAttr>(getPortNames()[resultNo]);
     }
-
-    /// Builds a new instance with the ports listed in `portIndices` erased, and
-    /// updates any users of the remaining ports to point at the new instance.
-    InstanceChoiceOp erasePorts(OpBuilder &builder,
-                                const llvm::BitVector &portIndices);
-
     /// Return the default target attribute.
     FlatSymbolRefAttr getDefaultTargetAttr() {
       return llvm::cast<FlatSymbolRefAttr>(getModuleNamesAttr()[0]);
@@ -233,6 +233,21 @@ def InstanceChoiceOp : HardwareDeclOp<"instance_choice", [
     /// Return the list of case-module mappings.
     SmallVector<std::pair<SymbolRefAttr, FlatSymbolRefAttr>, 1>
     getTargetChoices();
+
+    /// Clone this instance with inserted ports.
+    InstanceChoiceOp cloneWithInsertedPorts(
+      ArrayRef<std::pair<unsigned, PortInfo>> insertions);
+
+    /// Clone this instance with inserted ports, then replaces uses.
+    InstanceChoiceOp cloneWithInsertedPortsAndReplaceUses(
+      ArrayRef<std::pair<unsigned, PortInfo>> insertions);
+
+    /// Clone this instance with erased ports.
+    InstanceChoiceOp cloneWithErasedPorts(const llvm::BitVector &erasures);
+
+    /// Clone this instance with erased ports, then replaces uses.
+    InstanceChoiceOp cloneWithErasedPortsAndReplaceUses(
+      const llvm::BitVector &erasures);
   }];
 
   let builders = [
--- a/lib/Dialect/FIRRTL/FIRRTLOps.cpp
+++ b/lib/Dialect/FIRRTL/FIRRTLOps.cpp
@@ -753,18 +753,19 @@ BlockArgument FModuleOp::getArgument(size_t portNumber) {
 
 /// Return an updated domain info Attribute with domain indices updated based on
 /// port insertions.
-static Attribute
-fixDomainInfoInsertions(MLIRContext *context, Attribute domainInfoAttr,
-                        const SmallVectorImpl<unsigned> &numInserted) {
+static Attribute fixDomainInfoInsertions(MLIRContext *context,
+                                         Attribute domainInfoAttr,
+                                         ArrayRef<unsigned> indexMap) {
   // This is a domain type port.  Return the original domain info unmodified.
   auto di = dyn_cast_or_null<ArrayAttr>(domainInfoAttr);
-  if (!di)
+  if (!di || di.empty())
     return domainInfoAttr;
+
   // This is a non-domain type port.  Update any indeices referenced.
   SmallVector<Attribute> domainInfo;
   for (auto attr : di) {
     auto oldIdx = cast<IntegerAttr>(attr).getUInt();
-    auto newIdx = oldIdx + numInserted[oldIdx];
+    auto newIdx = indexMap[oldIdx];
     if (oldIdx == newIdx)
       domainInfo.push_back(attr);
     else
@@ -807,7 +808,6 @@ static void insertPorts(FModuleLike op,
   SmallVector<bool> newDirections;
   SmallVector<Attribute> newNames, newTypes, newDomains, newAnnos, newSyms,
       newLocs, newInternalPaths;
-  SmallVector<unsigned> numInserted;
   newDirections.reserve(newNumArgs);
   newNames.reserve(newNumArgs);
   newTypes.reserve(newNumArgs);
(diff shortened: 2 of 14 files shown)


include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
```diff
@@ -139,14 +139,28 @@ def InstanceOp : HardwareDeclOp<"instance", [
     /// Hooks for domain information.
     ArrayAttr getPortDomain(unsigned portIdx);
 
-    /// Builds a new instance with the ports listed in `portIndices` erased, and
-    /// updates any users of the remaining ports to point at the new instance.
-    InstanceOp erasePorts(OpBuilder &builder, const llvm::BitVector &portIndices);
+    /// Clone this instance with inserted ports.
+    [[nodiscard]] InstanceOp cloneWithInsertedPorts(
+      ArrayRef<std::pair<unsigned, PortInfo>> insertions);
```
seldridge, line 144: The `[[nodiscard]]` is a great idea!

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
```diff
@@ -139,14 +139,28 @@ def InstanceOp : HardwareDeclOp<"instance", [
...
-    /// conjuction with adding ports to the referenced module. This will emit
-    /// the new InstanceOp to the same location.
-    InstanceOp cloneAndInsertPorts(ArrayRef<std::pair<unsigned, PortInfo>> ports);
+    /// Clones with inserted ports, then replaces uses.
+    [[nodiscard]] InstanceOp cloneWithInsertedPortsAndReplaceUses(
+      ArrayRef<std::pair<unsigned, PortInfo>> insertions);
+
+    /// Clone and replace with inserted ports, then drop this instance.
+    [[nodiscard]] InstanceOp insertPorts(
+      ArrayRef<std::pair<unsigned, PortInfo>> insertions);
```
seldridge: Nit: it seems slightly inconsistent that as the methods do more, their names get longer, except for `insertPorts` and `erasePorts`. It's also weird that the shortest methods do the most work. It may be clearer to either use a long name or to drop the variants that also erase (given that this is easy enough to do in a second function call).

This is just commentary as I'm not sure what makes the most sense here. Also, am I reading this correctly in that `erasePorts` will now delete the op whereas previously it wouldn't? It seems better to just drop `::erasePorts` entirely as it _should_ be fully replaced with `cloneWithInsertedPortsAndReplaceUses`.

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
```diff
@@ -233,6 +241,29 @@ def InstanceChoiceOp : HardwareDeclOp<"instance_choice", [
...
+    /// Clone this instance with erased ports.
+    [[nodiscard]] InstanceChoiceOp
+    cloneWithErasedPorts(const llvm::BitVector &erasures);
+
+    /// Clone this instance with erased ports, then replaces uses.
+    [[nodiscard]] InstanceChoiceOp
+    cloneWithErasedPortsAndReplaceUses(const llvm::BitVector &erasures);
+
+    /// Clone and replace with erased ports, then drop this instance.
+    [[nodiscard]] InstanceChoiceOp erasePorts(const llvm::BitVector &erasures);
```
seldridge: Great! Would these make sense to raise into the `FInstanceLike` op interface (which would then necessitate them being implemented for `ObjectOp` as well)?

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
seldridge: After thinking more about this, I'm now questioning if this makes sense. None of the upstream operations like this, e.g., `clone` are restricted like this. Also, the couple of times that `(void)` casts to avoid the unused shows up indicates this may be an anti-pattern.

lib/Dialect/FIRRTL/Transforms/SpecializeLayers.cpp
```diff
@@ -379,12 +379,8 @@ struct SpecializeLayers {
     for (auto result : instance->getResults())
       if (!specializeValue(result))
         disabledPorts.set(result.getResultNumber());
-    if (disabledPorts.any()) {
-      OpBuilder builder(instance);
-      auto newInstance = instance.erasePorts(builder, disabledPorts);
-      instance->erase();
-      instance = newInstance;
-    }
```
seldridge, line 387: Nice that this avoids the conditional now.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -753,18 +753,22 @@ BlockArgument FModuleOp::getArgument(size_t portNumber) {
 
 /// Return an updated domain info Attribute with domain indices updated based on
 /// port insertions.
-static Attribute
-fixDomainInfoInsertions(MLIRContext *context, Attribute domainInfoAttr,
-                        const SmallVectorImpl<unsigned> &numInserted) {
+static Attribute fixDomainInfoInsertions(MLIRContext *context,
+                                         Attribute domainInfoAttr,
+                                         ArrayRef<unsigned> indexMap) {
```
seldridge, line 758: Nite: I like the change to calling this a "map" which it is. Just "index" may be too non-descriptive. Consider `cumulativeInsertionsMap` or something like that. Only a thought.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -753,18 +753,22 @@ BlockArgument FModuleOp::getArgument(size_t portNumber) {
...
   // This is a domain type port.  Return the original domain info unmodified.
   auto di = dyn_cast_or_null<ArrayAttr>(domainInfoAttr);
   if (!di)
     return domainInfoAttr;
+
+  if (di.empty())
+    return di;
```
seldridge: This can be combined with the previous check.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -2607,52 +2608,83 @@ LogicalResult InstanceOp::verify() {
...
-                                  const llvm::BitVector &portIndices) {
-  assert(portIndices.size() >= getNumResults() &&
-         "portIndices is not at least as large as getNumResults()");
+static void replaceUsesRespectingInsertedPorts(
+    Operation *op1, Operation *op2,
+    ArrayRef<std::pair<unsigned, PortInfo>> insertions) {
+  assert(op1 != op2);
+  size_t n = insertions.size();
+  size_t inserted = 0;
+  for (size_t i = 0, e = op1->getNumResults(); i < e; ++i) {
```
seldridge, line 2614: Consider combining into the initialization of the `for` loop. This usually requires some shenanigans like choosing shorter variable names so it doesn't look heinous. Barring that, consider initialization on one line.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -2607,52 +2608,83 @@ LogicalResult InstanceOp::verify() {
...
+    while (inserted < n) {
+      auto &insertion = insertions[inserted];
+      if (insertion.first > i)
+        break;
```
seldridge: Nit: inline single uses that don't assist readability:

```suggestion
      if (insertions[inserted].first > i)
        break;
```

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -2607,52 +2608,83 @@ LogicalResult InstanceOp::verify() {
...
+  }
+}
 
-  if (portIndices.none())
-    return *this;
+static void replaceUsesRespectingErasedPorts(Operation *op1, Operation *op2,
+                                             const llvm::BitVector &erasures) {
+  assert(op1 != op2);
+  size_t erased = 0;
+  for (size_t i = 0, e = op1->getNumResults(); i < e; ++i) {
```
seldridge, line 2631: Same nit as above: consider moving `erased` into the initialization of the `for` loop as it has no users outside the loop.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -2607,52 +2608,83 @@ LogicalResult InstanceOp::verify() {
...
+    auto r1 = op1->getResult(i);
+    if (erasures[i]) {
+      assert(r1.use_empty() && "removed instance port has uses");
```
seldridge, line 2634: This assertion feels slightly misplaced as it creates an unexpected precondition that isn't related to the correct operation of this method. Probably remove and then reorder the `auto r1` to be closer to its (now) sole use or inline the single use if still readable.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -2607,52 +2608,83 @@ LogicalResult InstanceOp::verify() {
   return failure();
 }
 
-/// Builds a new `InstanceOp` with the ports listed in `portIndices` erased, and
-/// updates any users of the remaining ports to point at the new instance.
-InstanceOp InstanceOp::erasePorts(OpBuilder &builder,
-                                  const llvm::BitVector &portIndices) {
-  assert(portIndices.size() >= getNumResults() &&
-         "portIndices is not at least as large as getNumResults()");
+static void replaceUsesRespectingInsertedPorts(
```
seldridge, line 2608: This and the subsequent method are _fantastic_ to have in static functions. 💯

lib/Dialect/FIRRTL/FIRRTLOps.cpp
```diff
@@ -2674,56 +2706,66 @@ ArrayAttr InstanceOp::getPortDomain(unsigned portIdx) {
...
+      ++inserted;
     }
+
+    newPortDirections.push_back(getPortDirection(i));
+    newPortNames.push_back(getPortName(i));
+    newPortTypes.push_back(getType(i));
+    newPortAnnos.push_back(getPortAnnotation(i));
+    newDomainInfo.push_back(getDomainInfo()[i]);
+    indexMap[i] = i + inserted;
+  }
```
seldridge, line 2744: Is there a way to share this code with that in the instance choice version?

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
rwy7: It seems like a good idea, but not something I want to do in this PR.

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
rwy7: I really only wanted to put `nodiscard` on the functions `insertPorts`/`erasePorts`, because it is an easy mistake to think that these functions are updating the op in-place. I don't really think it's necessary on `cloneWithInsertedPorts` or `cloneWithInsertedPortsAndReplaceUses`, so I can remove `nodiscard` from these functions. I'm partial to keeping `nodiscard` on `insertPorts`/`erasePorts`... but now I wonder if we should just remove these two functions instead.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: FYI the meaning of this parameter has changed, it is no longer a map from old-index to displacement, it is now a map from old-index to new-index.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: The check was present before, so I decided to preserve it. I agree it is misplaced but I think I would rather have it than not.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
rwy7: I think so, but maybe we can do that in a follow up. I've got all the ducks in a row now so it should be an easy PR.

lib/Dialect/FIRRTL/FIRRTLOps.cpp
seldridge: SGTM.

include/circt/Dialect/FIRRTL/FIRRTLDeclarations.td
rwy7: OK dropped the nodiscard, and dropped insertPorts/erasePorts.

PR #9094, opened 2025-10-14
[FIRRTL] DomainFieldAttr types are PropertyTypes

seldridge:
Split the FIRRTL ODS description of enums from attributes.  This was all
combined and creates problems if more upstream ODS files that describe
attributes (FIRRTLAttributes.td) need to include from more downstream ODS
files (FIRRTLTypes.ts).

Add verification (via ODS) that the types of DomainFieldAttrs are
PropertyTypes as opposed to generic MLIR types.

### DIFF
--- a/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
+++ b/include/circt/Dialect/FIRRTL/FIRRTLAttributes.td
@@ -14,6 +14,7 @@
 #define CIRCT_DIALECT_FIRRTL_FIRRTLATTRIBUTES_TD
 
 include "FIRRTLDialect.td"
+include "FIRRTLTypes.td"
 include "mlir/IR/BuiltinAttributeInterfaces.td"
 include "circt/Types.td"
 
@@ -214,7 +215,7 @@ def DomainFieldAttr : AttrDef<FIRRTLDialect, "DomainField"> {
   let summary = "A single field of a domain";
   let parameters = (
     ins "::mlir::StringAttr":$name,
-        "::mlir::TypeAttr":$type
+        "::circt::firrtl::PropertyType":$type
   );
   let assemblyFormat = [{
     `<` $name `,` $type `>`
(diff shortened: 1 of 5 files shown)


include/circt/Dialect/FIRRTL/FIRRTLDialect.h
```diff
@@ -22,7 +22,4 @@
 // Pull in the dialect definition.
 #include "circt/Dialect/FIRRTL/FIRRTLDialect.h.inc"
 
-// Pull in all enum type definitions and utility function declarations.
-#include "circt/Dialect/FIRRTL/FIRRTLEnums.h.inc"
```
dtzSiFive, line 26: Was including this a problem?
Looking at upstream MLIR dialect headers, they seem to.. do things differently.

I think requiring including FIRRTLEnums.h separately is reasonable to match how we've organized attributes/operations/types similarly.

Good call, thanks!

dtzSiFive (approved):
LGTM.

I think the reorg is right, or at least nothing seems wrong about it to me and the changes make sense (and I think are a bit tidier than things were).

dtzSiFive:
And of course the change to PropertyType's is great and LGTM!  Thanks for the cleanup!

include/circt/Dialect/FIRRTL/FIRRTLDialect.h
seldridge: It was a problem if this was included by `FIRRTLEnums.h` (which the PR needed) as the `.h.inc` don't have any include guards. Some have the `GET_<kind>_CLASSES` guard which could help, but enums and attributes are totally unguarded. Hence, I needed to get this under something that we could include guard.

I was following along with how SPIRV dialect was defined upstream and modeling it off of that.

Yes, we are doing things differently from upstream and that makes me want to go redo our dialects to align with upstream...

PR #9097, opened 2025-10-15
Bump Slang to v9.1

fabianschuiki:
Bump our Slang dependency to v9.1. We have been pointing at a commit after v9.0 to pick up a few bug fixes. These have landed in v9.1, such that we can go back to pointing at the released version again.

This release subtly changes the way generate blocks are named, which requires a few tweaks in ImportVerilog's handling of generate block name prefixes.


seldridge (approved):
LGTM!

dtzSiFive:
Nice, thanks!

PR #9098, opened 2025-10-16
[Synth] Fix race condition and memory corruption in longest path analysis caching

uenoku:
This commit addresses a race condition and memory corruption issue in the LongestPathAnalysis where concurrent access to cached results could lead to iterator invalidation and memory corruption. The issue occurred when the cachedResults DenseMap was modified during iteration, causing existing iterators and references to become invalid and point to corrupted memory.

The fix changes the cachedResults map to store unique_ptr<SmallVector<OpenPath>> instead of SmallVector<OpenPath> directly. This ensures that the underlying SmallVector objects remain at stable memory locations even when the DenseMap is rehashed or modified, preventing iterator invalidation and memory corruption.

Additionally, in markRegEndPoint, we now preemptively call getOrComputePaths for the endpoint before recording paths. This ensures all necessary paths are computed and cached before any operations that might trigger map modifications, eliminating the race condition.

Additional unique_ptr is not ideal so eventually we might want to use BumpAllocator

### DIFF
--- a/lib/Dialect/Synth/Analysis/LongestPathAnalysis.cpp
+++ b/lib/Dialect/Synth/Analysis/LongestPathAnalysis.cpp
@@ -1132,8 +1132,8 @@ FailureOr<ArrayRef<OpenPath>> LocalVisitor::getOrComputePaths(Value value,
   // Unique the results.
   filterPaths(*results, ctx->doKeepOnlyMaxDelayPaths(), ctx->isLocalScope());
   LLVM_DEBUG({
-    llvm::dbgs() << value << "[" << bitPos << "] " << "Found "
-                 << results->size() << " paths\n";
+    llvm::dbgs() << value << "[" << bitPos << "] "
+                 << "Found " << results->size() << " paths\n";
     llvm::dbgs() << "====Paths:\n";
     for (auto &path : *results) {
       path.print(llvm::dbgs());


lib/Dialect/Synth/Analysis/LongestPathAnalysis.cpp
```diff
@@ -730,7 +730,8 @@ class LocalVisitor {
   std::unique_ptr<llvm::ImmutableListFactory<DebugPoint>> debugPointFactory;
 
   // A map from the value point to the longest paths.
-  DenseMap<std::pair<Value, size_t>, SmallVector<OpenPath>> cachedResults;
+  DenseMap<std::pair<Value, size_t>, std::unique_ptr<SmallVector<OpenPath>>>
+      cachedResults;
```
seldridge, line 734: Unrelated, consider a `struct` over `std::pair` in a follow-on. This has the benefit of readability as well as possibly getting this onto one line. 😉

seldridge (approved):
Nice solution. Is there any possibility of parallel insertion into the `cachedResults` that has to be considered? (Or is that entirely taken care of by the pre-emptive call to `getOrComputePath`?)

uenoku:
> Is there any possibility of parallel insertion into the cachedResults that 

cacheResults is intended to cache results under the current module scope and the cache must be frozen if accessed from other threads (which are basically computing paths from parent scope).  The issue was that cache was not pre-computed properly in the module scope.

PR #9099, opened 2025-10-16
[FIRRTL] Add LowerDomains to firtool pipeline

seldridge:
Add the LowerDomains to the low FIRRTL to HW pipeline inside firtool.
This is the latest position in which it can run as it has a hard
dependency on needing to run before LowerClasses.

This is not conditionalized with a chicken switch option as this pass will
eventually be required for correctness and it is a no-op if no domain
information exists in the circuit.

### DIFF
--- a/test/firtool/domains.fir
+++ b/test/firtool/domains.fir
@@ -1,0 +1,34 @@
+; RUN: firtool %s -output-final-mlir %t.mlir && FileCheck %s < %t.mlir
+
+FIRRTL version 6.0.0
+circuit Foo :
+  domain ClockDomain :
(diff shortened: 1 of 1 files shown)


seldridge:
One note: don't try to do anything interesting involving multiple modules and `domain_define` as that doesn't get through `ExpandWhens`. I'm working on fixing this...

_Edit: trivial fix for `ExpandWhens`: https://github.com/llvm/circt/commit/afc808430c5b60b3e12e8d6c124e506490e6ba39. There are later errors in the pipeline I'm dealing with now._

PR #9100, opened 2025-10-16
[MooreToCore] Lower `moore.real_constant` to `arith.constant`

Scheremo:
This commit introduces explicit support for `shortreal` literals in the Moore dialect and unifies the lowering and import handling of real constants. It ensures correct mapping between SystemVerilog `real`/`shortreal` and MLIR `f64`/`f32` types across all relevant conversions.

Changes:

Dialect (MooreOps.td):
- Added `ShortrealLiteralOp`:
  - Accepts `F32Attr` and produces `RealF32`.
  - Represents IEEE-754 single-precision (SystemVerilog `shortreal`).
- Updated `RealLiteralOp`:
  - Now explicitly uses `F64Attr` and produces `RealF64`.
  - Simplified `assemblyFormat` by removing redundant result type.
  - Clarified description to indicate IEEE-754 double precision.

ImportVerilog (Expressions.cpp):
- Updated `materializeSVReal`:
  - Emits `ShortrealLiteralOp` for `shortreal` values.
  - Emits `RealLiteralOp` for `real` values.
  - Uses appropriate `FloatAttr` construction and result type selection.

MooreToCore Conversion (MooreToCore.cpp):
- Added conversion patterns:
  - `RealConstantOpConv` → lowers to `arith.constant` with `f64` type.
  - `ShortrealConstantOpConv` → lowers to `arith.constant` with `f32` type.
- Marked `arith::ArithDialect` as legal.
- Extended `TypeConverter` to map:
  - `moore::RealWidth::f32` → `mlir::Float32Type`
  - `moore::RealWidth::f64` → `mlir::Float64Type`

Tests:
- Updated `ImportVerilog` tests to expect `moore.shortreal_constant` and `moore.real_constant` without explicit result type suffixes.
- Added `MooreToCore` tests verifying correct lowering:
  - `moore.real_constant` → `arith.constant f64`
  - `moore.shortreal_constant` → `arith.constant f32`

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -633,6 +633,35 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
   }
 };
 
+struct RealConstantOpConv : public OpConversionPattern<RealLiteralOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(RealLiteralOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    llvm::APFloat apf = op.getValue();
+    Type outTy = rewriter.getF64Type();
+    auto attr = FloatAttr::get(outTy, apf);
+    rewriter.replaceOpWithNewOp<arith::ConstantOp>(op, attr);
+    return success();
+  }
+};
+
+struct ShortrealConstantOpConv
+    : public OpConversionPattern<ShortrealLiteralOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ShortrealLiteralOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    llvm::APFloat apf = op.getValue();
+    Type outTy = rewriter.getF32Type();
+    auto attr = FloatAttr::get(outTy, apf);
+    rewriter.replaceOpWithNewOp<arith::ConstantOp>(op, attr);
+    return success();
+  }
+};
+
 struct ConstantTimeOpConv : public OpConversionPattern<ConstantTimeOp> {
   using OpConversionPattern::OpConversionPattern;
 
@@ -1742,6 +1771,7 @@ static void populateLegality(ConversionTarget &target,
   target.addLegalDialect<sim::SimDialect>();
   target.addLegalDialect<mlir::LLVM::LLVMDialect>();
   target.addLegalDialect<verif::VerifDialect>();
+  target.addLegalDialect<arith::ArithDialect>();
 
   target.addLegalOp<debug::ScopeOp>();
 
@@ -1770,6 +1800,16 @@ static void populateLegality(ConversionTarget &target,
 static void populateTypeConversion(TypeConverter &typeConverter) {
   typeConverter.addConversion([&](IntType type) {
     return IntegerType::get(type.getContext(), type.getWidth());
+  });
+
+  typeConverter.addConversion([&](RealType type) -> mlir::Type {
+    MLIRContext *ctx = type.getContext();
+    switch (type.getWidth()) {
+    case moore::RealWidth::f32:
+      return mlir::Float32Type::get(ctx);
+    case moore::RealWidth::f64:
+      return mlir::Float64Type::get(ctx);
+    }
   });
 
   typeConverter.addConversion(
@@ -1916,6 +1956,8 @@ static void populateOpConversion(ConversionPatternSet &patterns,
 
     // Patterns of miscellaneous operations.
     ConstantOpConv,
+    RealConstantOpConv,
+    ShortrealConstantOpConv,
     ConcatOpConversion,
     ReplicateOpConversion,
     ConstantTimeOpConv,
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -571,6 +571,23 @@ def StringConstantOp : MooreOp<"string_constant", [Pure]> {
   let assemblyFormat = "$value attr-dict `:` type($result)";
 }
 
+def ShortrealLiteralOp : MooreOp<"shortreal_constant", [Pure]> {
+  let summary = "Produce a constant shortreal value";
+  let description = [{
+    Produces a constant value of shortreal type.
+
+    Example:
+    ```mlir
+    %0 = moore.shortreal_constant 1.23
+    ```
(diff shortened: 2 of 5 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -633,6 +633,39 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
+      outTy = rewriter.getF64Type();
+    }
+
+    if (!outTy)
+      return failure();
+
+    val = op.getValue().convertToDouble();
+
+    // Build a FloatAttr with the chosen float type and the numeric value.
+    auto attr = rewriter.getFloatAttr(outTy, val);
```
fabianschuiki: I think the value might already be a float attribute that you can pass to `arith::ConstantOp` directly. (Haven't checked exactly how `RealLiteralOp` is defined. But that would avoid the potentially lossy conversion to `double`.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -633,6 +633,39 @@ struct ConstantOpConv : public OpConversionPattern<ConstantOp> {
...
+    Type outTy; // target float type (f32/f64)
+    double val; // use double to build the FloatAttr
+
+    if (inTy == moore::RealType::getShortReal(getContext())) {
+      // shortreal -> f32
+      outTy = rewriter.getF32Type();
+    } else if (inTy == moore::RealType::getReal(getContext())) {
+      // real -> f64
+      outTy = rewriter.getF64Type();
+    }
```
fabianschuiki: Would it be possible to check the width of `inTy` directly, like `inTy.getWidth() == 32`? The `getShortReal(...)` will re-intern the type into the context, which can be costly because it needs to be thread-safe. Checking the width should be practically free in comparison 🙂

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -1772,6 +1806,20 @@ static void populateTypeConversion(TypeConverter &typeConverter) {
...
+    mlir::FloatType fTy;
+    switch (type.getWidth()) {
+    case moore::RealWidth::f32:
+      fTy = mlir::Float32Type::get(ctx);
+      break;
+    case moore::RealWidth::f64:
+      fTy = mlir::Float64Type::get(ctx);
+      break;
+    }
+    return fTy;
```
fabianschuiki: I think you can return the types directly, and if nothing matches in the switch statement, returning a null type will indicate conversion failure to the conversion framework:

```suggestion
    switch (type.getWidth()) {
    case moore::RealWidth::f32:
      return mlir::Float32Type::get(ctx);
    case moore::RealWidth::f64:
      return mlir::Float64Type::get(ctx);
    default:
      return mlir::Type{};
    }
```

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Sure! Done.

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: I get a stack trace if I do that 😅 But I can put in some detective work to see what that stems from.

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: If we add a `-> mlir::Type` to the lambda signature this works. Done!

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: I also can't really come up with case where a cast to f64 is lossy; if we start from a real, it's already a double, if we start from a shortreal it's a single. Shouldn't casting to double always be fine? 🤔

fabianschuiki: I'm not sure how this exactly works in the background, for example if you're on x86, this might cast to intel's 80-bit float format. I'm not sure if that is guaranteed to be a lossless conversion :thinking: It might be?

You might also be able to use the underlying `APFloat` instead of a double. Just to avoid conversion.

Scheremo: We can stay in `APFloat` and convert to `IEEE` float types which seems to be the best I can figure out, i.e.
```
    if (inTy.getWidth() == moore::RealWidth::f32) {
      // shortreal -> f32
      outTy = rewriter.getF32Type();
      apf.convert(llvm::APFloat::IEEEsingle(),
                  llvm::APFloat::rmNearestTiesToEven, &losesInfo);
    } else if (inTy.getWidth() == moore::RealWidth::f64) {
      // real -> f64
      outTy = rewriter.getF64Type();
      apf.convert(llvm::APFloat::IEEEdouble(),
                  llvm::APFloat::rmNearestTiesToEven, &losesInfo);
    }
```
WDYT?

fabianschuiki: I think you can just take the `APFloat` and shove it into the constructor of the `FloatAttr`. Or does that fail? We might be missing something in `RealLiteralOp` that allows this to happen.

Scheremo: We store a F64Attr in RealLiteralOp (which makes sense imo, there are contexts where it's unclear whether we are dealing with a `real` or a `shortreal`). 
The problem then turns up when we pass the 64Bit `APFloat` to a 32Bit `FloatAttr:get`. 
We can also add an assertion on `losesInfo` to confirm, but I think it's actually possible we store at a `shortreal` at a higher precision than we cast it down to here.

fabianschuiki: No worries! We can always make another pass over real consts later 👍.

Scheremo: > We store a F64Attr in RealLiteralOp (which makes sense imo, there are contexts where it's unclear whether we are dealing with a `real` or a `shortreal`). The problem then turns up when we pass the 64Bit `APFloat` to a 32Bit `FloatAttr:get`. We can also add an assertion on `losesInfo` to confirm, but I think it's actually possible we store at a `shortreal` at a higher precision than we cast it down to here.

Actually I take this back, we should probably refactor `RealLiteralOp` in `moore`. 
The situation of indeterminate `real` type I was thinking of is
```
parameter zx = 3.1452
```
But this doesn't get emitted as a RealLiteralOp, we materialize it as a constant in `Expressions`. `RealLiteralOp` should always know whether the input is f32 or f64.

Scheremo: Fixed it by introducing a `moore.shortreal_constant` :smile:

PR #9105, opened 2025-10-17
[Conversion] Add comprehensive CombToLLVM conversion pass with func.func-only scope

prithayan:
This PR adds a complete implementation of the `ConvertCombToLLVM` pass that converts Comb and related dialect operations to LLVM IR. The pass is designed to only convert operations inside `func.func` operations, leaving all other operations (such as `hw.module`) unconverted.

Previously, the CombToLLVM conversion infrastructure only provided pattern population functions (`populateCombToLLVMConversionPatterns`) but no complete pass implementation.

This PR adds:
1. A conversion pass that orchestrates the complete lowering pipeline from Comb/HW to LLVM
2. Selective conversion scope that only processes `func.func` operations, allowing mixed representations where:
   - Software-like functions are lowered to LLVM IR for execution/simulation
   - Hardware modules remain in their original form for other passes or backends
3. Comprehensive test coverage that adds tests covering all major Comb operations

**Example  Input:**
```mlir
func.func @my_function(%a: i32, %b: i32) -> i32 {
  %result = comb.add %a, %b : i32
  return %result : i32
}

hw.module @my_hardware_module(in %x: i32, in %y: i32, out z: i32) {
  %sum = comb.add %x, %y : i32
  hw.output %sum : i32
}
```

**Output after `--convert-comb-to-llvm`:**
```mlir
llvm.func @my_function(%arg0: i32, %arg1: i32) -> i32 {
  %0 = llvm.add %arg0, %arg1 : i32
  llvm.return %0 : i32
}

hw.module @my_hardware_module(in %x : i32, in %y : i32, out z : i32) {
  %0 = comb.add %x, %y : i32
  hw.output %0 : i32
}
```

**Implementation detail:** 

The pass applies conversion patterns in a specific order to ensure correct lowering:

```
Input (Comb/HW/Func/SCF dialects)
    ↓
1. SCF → ControlFlow (if/while → branches)
    ↓
2. Func → LLVM (func.func → llvm.func)
    ↓
3. ControlFlow → LLVM (br/cond_br → llvm.br/llvm.cond_br)
    ↓
4. Comb → Arith (comb.add → arith.addi, etc.)
    ↓
5. Arith → LLVM (arith.addi → llvm.add, etc.)
    ↓
6. Index → LLVM (index operations)
    ↓
7. Function interface type conversions
    ↓
8. Comb → LLVM (comb.parity → llvm.ctpop, hw.output → llvm.return)
    ↓
Output (LLVM dialect only, inside func.func)
```

This change complements the existing conversion infrastructure:
- **`CombToArith`**: Converts most Comb operations to Arith dialect (used by this pass)
- **`HWToLLVM`**: Provides type conversions for HW types (used by this pass)
- **`ArithToLLVM`**: Converts Arith operations to LLVM dialect (used by this pass)
- **`FuncToLLVM`**: Converts Func operations to LLVM dialect (used by this pass)
- **`SCFToControlFlow`**: Converts SCF to ControlFlow dialect (used by this pass)

The **CombToLLVM pass orchestrates** all these conversions into a single, complete lowering pipeline, but only applies them to `func.func` operations.

Note: This PR has been developed with help from Calude Sonnet 4.5

### DIFF
--- a/lib/CAPI/Conversion/CMakeLists.txt
+++ b/lib/CAPI/Conversion/CMakeLists.txt
@@ -14,6 +14,7 @@ add_circt_public_c_api_library(CIRCTCAPIConversion
   CIRCTCombToLLVM
   CIRCTCombToSMT
   CIRCTConvertToArcs
+  CIRCTConvertToLLVM
   CIRCTDatapathToComb
   CIRCTDatapathToSMT
   CIRCTDCToHW


include/circt/Conversion/Passes.td
```diff
@@ -513,6 +513,24 @@ def ConvertHWToLLVM : Pass<"convert-hw-to-llvm", "mlir::ModuleOp"> {
...
+//===----------------------------------------------------------------------===//
+
+def ConvertCombToLLVM : Pass<"convert-comb-to-llvm", "mlir::ModuleOp"> {
+  let summary = "Convert Comb and HW to LLVM";
+  let description = [{
+    This pass translates Comb and HW operations to LLVM IR. It combines
+    both HW and Comb conversion patterns to provide a complete lowering
+    from hardware description to LLVM IR.
+  }];
+  let constructor = "circt::createConvertCombToLLVMPass()";
```
maerhart: Nit: why not use the auto-generated constructor?

lib/Conversion/CombToLLVM/CombToLLVM.cpp
```diff
@@ -43,9 +63,125 @@ struct CombParityOpConversion : public ConvertToLLVMPattern {
...
+//===----------------------------------------------------------------------===//
+// HW Structural Operation Conversion Patterns
+//===----------------------------------------------------------------------===//
+
+/// Convert a hw::OutputOp to an llvm.return.
+struct HWOutputOpConversion : public ConvertToLLVMPattern {
+  explicit HWOutputOpConversion(MLIRContext *ctx,
+                                LLVMTypeConverter &typeConverter)
+      : ConvertToLLVMPattern(hw::OutputOp::getOperationName(), ctx,
+                             typeConverter) {}
```
maerhart: We can use the more modern `ConvertOpToLLVMPattern` as base class. This will then use a proper `OpAdaptor` instead of a list of operands.

lib/Conversion/CombToLLVM/CombToLLVM.cpp
```diff
@@ -43,9 +63,125 @@ struct CombParityOpConversion : public ConvertToLLVMPattern {
...
+          rewriter.create<LLVM::UndefOp>(op->getLoc(), structType);
+      for (auto [i, operand] : llvm::enumerate(operands)) {
+        structVal = rewriter.create<LLVM::InsertValueOp>(op->getLoc(),
+                                                         structVal, operand, i);
+      }
+      rewriter.replaceOpWithNewOp<LLVM::ReturnOp>(op, structVal);
+    }
+    return success();
+  }
+};
```
maerhart: Maybe that pattern makes more sense in HWToLLVM.cpp?

lib/Conversion/CombToLLVM/CombToLLVM.cpp
```diff
@@ -43,9 +63,125 @@ struct CombParityOpConversion : public ConvertToLLVMPattern {
...
+  // pipeline
   patterns.add<CombParityOpConversion>(patterns.getContext(), converter);
+
+  // Add HW structural operation patterns
+  patterns.add<HWOutputOpConversion>(patterns.getContext(), converter);
+}
+
+void CombToLLVMLoweringPass::runOnOperation() {
+  // Iterate over all func.func operations in the module and convert them
+  getOperation().walk([&](func::FuncOp funcOp) { convertFuncOp(funcOp); });
```
maerhart: Nit: We don't need a full walk here since FuncOp can only be a child of ModuleOp. So maybe just `getOps<FuncOp>()`?

lib/Conversion/CombToLLVM/CombToLLVM.cpp
```diff
@@ -43,9 +63,125 @@ struct CombParityOpConversion : public ConvertToLLVMPattern {
...
+  // Add HW to LLVM type conversions
+  populateHWToLLVMTypeConversions(converter);
+
+  LLVMConversionTarget target(*context);
+  target.addIllegalDialect<comb::CombDialect>();
+  target.addIllegalDialect<arith::ArithDialect>();
+
+  // Mark HW and SV dialects as legal - we only convert operations inside
+  // func.func
+  target.addLegalDialect<hw::HWDialect>();
```
maerhart: That would mark any remaining `hw.output` inside a `func.func` legal. There is also no test for that case. Is the `hw.output` conversion pattern not necessary after all or should this legality constraint be changed?

lib/Conversion/CombToLLVM/CombToLLVM.cpp
```diff
@@ -43,9 +63,125 @@ struct CombParityOpConversion : public ConvertToLLVMPattern {
...
+  target.addLegalDialect<sv::SVDialect>();
```
maerhart: There shouldn't be any comb operation lowering to an SV operation. Why is this necessary?

test/Conversion/CombToLLVM/comb-to-llvm.mlir
```diff
@@ -0,0 +1,215 @@
...
+  hw.instance "child_decoder" @test_child_module(scan_data: %2: i1) -> ()
+  hw.output
+}
+
+hw.module @test_child_module(in %scan_data : i1) {
+  %0 = comb.extract %scan_data from 0 : (i1) -> i1
+  %SimpleChild.child_reg_decoded = sv.wire : !hw.inout<i1>
+  sv.assign %SimpleChild.child_reg_decoded, %0 : i1
+  hw.output
+}
```
maerhart: Super-Nit: we can probably reduce that test a bit

include/circt/Conversion/Passes.td
```diff
@@ -513,6 +513,24 @@ def ConvertHWToLLVM : Pass<"convert-hw-to-llvm", "mlir::ModuleOp"> {
...
+//===----------------------------------------------------------------------===//
+// CombToLLVM
+//===----------------------------------------------------------------------===//
+
+def ConvertCombToLLVM : Pass<"convert-comb-to-llvm", "mlir::ModuleOp"> {
+  let summary = "Convert Comb and HW to LLVM";
+  let description = [{
+    This pass translates Comb and HW operations to LLVM IR. It combines
+    both HW and Comb conversion patterns to provide a complete lowering
+    from hardware description to LLVM IR.
```
maerhart: Nit: maybe mention that it only considers `func.func` operations in this description as well.

maerhart (approved):
This is already getting close to a pass that does a full conversion to LLVM including HW aggregates etc. Been thinking of implementing one based on dialect interfaces for quite a while now.

Given that it also converts SCF and index operations and in the future probably also HW aggregates, we might want to call it something like `ConvertToLLVM` instead of `ConvertCombToLLVM` and put it in a separate file.
Otherwise LGTM

lib/Conversion/CombToLLVM/CombToLLVM.cpp
prithayan: I realized, this conversion pattern is not required. I was earlier trying to convert HWModule to func, but  dropped the idea. Thanks for pointing it out.

lib/Conversion/CombToLLVM/CombToLLVM.cpp
prithayan: Dropped the conversion pattern.

lib/Conversion/CombToLLVM/CombToLLVM.cpp
prithayan: Good catch, updated to use getOps instead.

lib/Conversion/CombToLLVM/CombToLLVM.cpp
prithayan: You are right, I dropped the hw.output conversion.

test/Conversion/CombToLLVM/comb-to-llvm.mlir
prithayan: Thanks for noticing, this is not required. Reduced it.

include/circt/Conversion/Passes.td
prithayan: Yes, updated the description.

prithayan:
> This is already getting close to a pass that does a full conversion to LLVM including HW aggregates etc. Been thinking of implementing one based on dialect interfaces for quite a while now.
> 
> Given that it also converts SCF and index operations and in the future probably also HW aggregates, we might want to call it something like `ConvertToLLVM` instead of `ConvertCombToLLVM` and put it in a separate file. Otherwise LGTM

Thanks @maerhart for the detailed review. I agree, this is not just CombToLLVM. So, created the new pass ConvertToLLVM instead. You can add future extensions and patterns as necessary there.

PR #9107, opened 2025-10-17
[ImportVerilog] Allow functions to capture values from parent scope

Scheremo:
This patch teaches the ImportVerilog conversion to detect and plumb captured references from enclosing scopes into lowered MLIR functions.

- **FunctionLowering**
  - Added `captures` and `captureIndex` to record refs captured from outer scopes.
- **Context**
  - Added `finalizeFunctionBodyCaptures` to extend the function signature with captured `moore::RefType` inputs and replace their in-body uses with the new block arguments.
  - `convertFunction` now installs a temporary `rvalueReadCallback` that collects refs read from outside the function’s body.
- **Expressions**
  - When lowering subroutine calls, the visitor now appends any captured operands to the call argument list. Captures are verified to be `moore::RefType` values originating from the enclosing (module) region.

Together these changes allow nested or free functions to reference module-level signals (`logic`, etc.) correctly in the generated MLIR, with explicit function parameters modeling captured variables.

### DIFF
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -1100,7 +1100,31 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
     valueSymbols.insert(subroutine.returnValVar, returnVar);
   }
 
+  // Save previous callback
+  auto prevCb = rvalueReadCallback;
+  auto prevCbGuard =
+      llvm::make_scope_exit([&] { rvalueReadCallback = prevCb; });
+
+  // Capture this function's captured context directly
+  rvalueReadCallback = [lowering, prevCb](moore::ReadOp rop) {
+    if (prevCb)
+      prevCb(rop); // chain previous callback
+
+    mlir::Value ref = rop.getInput();
+    // Only capture refs defined outside this function’s region
+    if (!lowering->op.getBody().isAncestor(ref.getParentRegion())) {
+      auto [it, inserted] =
+          lowering->captureIndex.try_emplace(ref, lowering->captures.size());
+      if (inserted)
+        lowering->captures.push_back(ref);
+    }
+  };
+
   if (failed(convertStatement(subroutine.getBody())))
+    return failure();
+
+  // Plumb captures into the function as extra block arguments
+  if (failed(finalizeFunctionBodyCaptures(*lowering)))
     return failure();
 
   // If there was no explicit return statement provided by the user, insert a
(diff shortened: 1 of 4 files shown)


lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -1100,9 +1100,34 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
     valueSymbols.insert(subroutine.returnValVar, returnVar);
   }
 
+  // Save previous callback
+  auto prevCb = rvalueReadCallback;
```
fabianschuiki, line 1104: Since the code below can early exit upon failure, you could use a scope exit guard to switch the read callback back to the original:
```suggestion
  auto prevCb = rvalueReadCallback;
  auto prevCbGuard =
      llvm::make_scope_exit([&] { rvalueReadCallback = prevCb; });
```
Probably not really needed in case of a failure, but it's one thing less that might get forgotten.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -1130,3 +1155,46 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
   }
   return success();
 }
+
+LogicalResult Context::finalizeFunctionBodyCaptures(FunctionLowering &FL) {
```
fabianschuiki: Maybe change `FL` to `lowering` to maintain MLIR camel case naming for arguments and variables.

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -3347,16 +3347,37 @@ endfunction
...
+
+// CHECK: moore.module @testFunctionCapture() {
+module testFunctionCapture();
+
+    // CHECK: [[A:%.+]] = moore.variable : <l1>
+    logic a;
+
+    function logic testCapture;
+        return a;
+    endfunction
```
fabianschuiki, line 3372: Can functions be nested? If yes, it might make sense to check if you can nest another function inside of `testCapture` that captures `a`. That would then need to be plumbed through both calls, which might be interesting 😁

fabianschuiki (approved):
Really cool! LGTM!

lib/Conversion/ImportVerilog/Structure.cpp
Scheremo: Nice! Done.

lib/Conversion/ImportVerilog/Structure.cpp
Scheremo: Done!

test/Conversion/ImportVerilog/basic.sv
Scheremo: If I parse IEEE 1800-2023 Section 13.4 correctly, function body declarations may not contain further function declarations, only statements.

PR #9108, opened 2025-10-18
[FIRRTL] Support domain.define in LowerDomains

seldridge:
Add support for lowering `firrtl.domain.define` operations to
`firrtl.propassign`.  This uses a technique similar to how the dialect
conversion framework works where unrealized conversion casts are used to
deal with the progressive lowering of ports by introducing "stubs" (0-1
placeholders) to allow ports to be erased and then "spliced" (1-1
conversions) after the ports are inserted with now-incompatible types
before finally converting defines to propassigns.

- [x] Corner cases with U-turns are sorted out.

### DIFF
--- a/lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
+++ b/lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
@@ -59,6 +59,7 @@ namespace firrtl {
 
 using namespace circt;
 using namespace firrtl;
+using mlir::UnrealizedConversionCastOp;
 
 class LowerDomainsPass : public impl::LowerDomainsBase<LowerDomainsPass> {
   using Base::Base;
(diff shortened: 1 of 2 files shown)


lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -292,6 +342,7 @@ LogicalResult LowerModule::lowerModule() {
   //   2. iDel tracks the port index after deletion.
   //   3. iIns tracks the port index after insertion.
   auto ports = op.getPorts();
+  OpBuilder::InsertPoint insertPoint;
```
rwy7, line 344: Suggestion: can you initialize the insertion point here, rather than on the first iteration? Maybe like this?

```suggestion
  OpBuilder::InsertPoint insertPoint;
  if (body) {
    insertPoint = {body, body->begin()};
  }
```

Or maybe even hoist the builder up and out of the loop?

seldridge: Nice idea. Switched in: b56d1136c

lib/Dialect/FIRRTL/Transforms/LowerDomains.cpp
```diff
@@ -380,22 +444,28 @@ LogicalResult LowerModule::lowerModule() {
...
-      // Assign input domain info if needed.
-      //
-      // TODO: Change this to hook up to its connection once domain connects are
-      // available.
+      // Hook up domain ports.  If this was an input domain port, then drive the
+      // object's "domain in" port with the new input class port.  Splice
+      // references to the old port with the new port---users (e.g., domain
+      // defines) will be updated later.  If an output, then splice to the
+      // object's "domain in" port.  If this participates in domain defines,
+      // this will be hoked up later.
```
dtzSiFive:
```suggestion
      // this will be hooked up later.
```

PR #9111, opened 2025-10-18
[ImportVerilog] [2/2] Add post-pass to rewrite recursive call sites

Scheremo:
This change makes recursive subroutines with captured refs lower correctly by:
1) **Deferring** capture operands at call sites while a callee is still being converted, and
2) **Rewriting** all call sites after capture finalization to pass the right operands for both
   internal (recursive) and external calls.

- **Call emission guard**
  - In `Expressions.cpp`, only append `lowering->captures` at call sites when
    `!lowering->isConverting`. This prevents attempting to “materialize” captures
    during a self-call while the callee’s body is still under construction.

- **Post-pass call-site rewrite**
  - Add `rewriteCallSitesToPassCaptures(func::FuncOp, ArrayRef<Value>)` in `Structure.cpp`.
  - Finds all `func.call` uses of `callee` within its `ModuleOp`, snapshots them, and rebuilds:
    - **Recursive (in-body) calls:** append the function’s **capture block arguments**
      (the tail args of the entry block) to the call.
    - **External calls:** append the **captured SSA values** collected during conversion.
  - Uses `FlatSymbolRefAttr::get(callee)` for the rebuilt call.
  - Invoked immediately after `finalizeFunctionBodyCaptures`.

- **Tests**
  - Extend `basic.sv` with `testRecursiveCaptureFunction`:
    - Verifies external call passes `[[CAPTUREME]]` to `@fact`.
    - Ensures `@fact`’s recursive in-body call appends `%arg1` (the capture block arg).
    - Checks base case reads from `%arg1` and returns it.

### DIFF
--- a/test/Conversion/ImportVerilog/basic.sv
+++ b/test/Conversion/ImportVerilog/basic.sv
@@ -3404,9 +3404,32 @@ module testLHSTaskCapture();
 
 endmodule
 
-// CHECK: func.func private @testRecursive(%arg0: !moore.i32) -> !moore.i32 {
+// CHECK-LABEL: @testRecursive
+// CHECK-SAME: %arg0: !moore.i32
+// CHECK-SAME: -> !moore.i32
 function int testRecursive(input int n);
     if (n <= 1) return 1;
     // CHECK: [[REC:%.+]] = call @testRecursive({{.*}}) : (!moore.i32) -> !moore.i32
     return n * testRecursive(n - 1);
 endfunction
+
+// CHECK-LABEL: moore.module @testRecursiveCaptureFunction() {
+module testRecursiveCaptureFunction();
+  // CHECK: [[CAPTUREME:%.+]] = moore.variable : <i32>
+  int captureMe;
+  int r;
+  initial begin
+    // CHECK: func.call @fact({{.*}}, [[CAPTUREME]]) : (!moore.i32, !moore.ref<i32>) -> !moore.i32
+    r = fact(5);
+  end
+
+  // CHECK: func.func private @fact(%arg0: !moore.i32, %arg1: !moore.ref<i32>) -> !moore.i32 {
+  function int fact(input int n);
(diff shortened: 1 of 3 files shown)


test/Conversion/ImportVerilog/basic.sv
```diff
@@ -3381,3 +3381,31 @@ module testFunctionCapture();
     // CHECK: [[CAPTUREDA:%.+]] = moore.read %arg0 : <l1>
     // CHECK: return [[CAPTUREDA]] : !moore.l1
 endmodule
+
+// CHECK: func.func private @testRecursive(%arg0: !moore.i32) -> !moore.i32 {
```
fabianschuiki: It often pays to have a `CHECK-LABEL` match on something trivial like the function name. FileCheck will limit all other searches/matches to within adjacent `CHECK-LABEL`s, which can produce better error messages since `CHECK` lines don't accidentally match on the next function.
```suggestion
// CHECK-LABEL: @testRecursive
// CHECK-SAME: %arg0: !moore.i32
// CHECK-SAME: -> !moore.i32
```

test/Conversion/ImportVerilog/basic.sv
```diff
@@ -3381,3 +3381,31 @@ module testFunctionCapture();
...
+function int testRecursive(input int n);
+    if (n <= 1) return 1;
+    // CHECK: [[REC:%.+]] = call @testRecursive({{.*}}) : (!moore.i32) -> !moore.i32
+    return n * testRecursive(n - 1);
+endfunction
+
+// CHECK: moore.module @testRecursiveCaptureFunction() {
```
fabianschuiki:
```suggestion
// CHECK-LABEL: moore.module @testRecursiveCaptureFunction() {
```

test/Conversion/ImportVerilog/basic.sv
Scheremo: Thanks for the hint! Will incorporate.

PR #9112, opened 2025-10-19
[ImportVerilog] Add support for string materialization

Scheremo:
Extend ImportVerilog to lower SystemVerilog string constants by:
- Introducing `moore.int_to_string` to convert integer-backed literals to `!moore.string`.
- Materializing Slang string constants via a new `Context::materializeString` path.
- Integrating string handling into the general constant materialization flow.

- **Dialect**
  - Add `IntToStringOp` (`moore.int_to_string`) to `MooreOps.td` with:
    - `TwoValuedIntType` input -> `StringType` result

- **ImportVerilog conversion**
  - Implement `Context::materializeString(constant, type, loc)`:
    - Converts Slang `ConstantValue` string to an integer representation and creates a `moore.string_constant`, then wraps it with `moore.int_to_string`.
  - Update `materializeConstant` to recognize `constant.isString()` and route to the new helper.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -751,6 +751,18 @@ def IntToRealOp : MooreOp<"int_to_real", [
   }];
 }
 
+def IntToStringOp : MooreOp<"int_to_string", [
+  Pure
+]> {
+  let summary = "Convert an integer to a string";
+  let arguments = (ins TwoValuedIntType:$input);
+  let results = (outs StringType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input)
+  }];
+}
+
+
 //===----------------------------------------------------------------------===//
 // Resizing
 //===----------------------------------------------------------------------===//
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -1622,6 +1622,26 @@ Value Context::materializeSVReal(const slang::ConstantValue &svreal,
   return {};
 }
 
+/// Materialize a Slang string literal as a literal string constant op.
+Value Context::materializeString(const slang::ConstantValue &stringLiteral,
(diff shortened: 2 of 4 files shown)


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -571,6 +571,21 @@ def StringConstantOp : MooreOp<"string_constant", [Pure]> {
...
+    Example:
+    ```mlir
+    %0 = moore.string_literal "hello world"
+    ```
+  }];
+  let arguments = (ins StrAttr:$value);
+  let results = (outs StringType:$result);
+  let assemblyFormat = "$value attr-dict";
+}
+
```
fabianschuiki: Can we use the existing `moore.string_constant` for this, and then use a new `moore.int_to_string` cast to go from an integer to a `!moore.string` type? It would be good if we had just one way of materializing a string literal in the IR, which SV really wants to be an integer 😭. That can then be converted to a `string`, but use as an integer is also possible.

Strings are very annoying… on the one hand, I'd love to just have a `moore.string_constant` op that returns a `!moore.string`. But that makes it harder to convert the string to an integer when needed, since `!moore.string` is a dynamic string. So maybe sticking close to the SV spec with a string constant that produces and integer makes sense?

Scheremo: I think no matter which way we pick it will feel bad 🙈 
`moore.int_to_string` is a workable way; It just hurts to think we take a perfectly well-formed string, cast it to an int, and cast it back to a string 😭 
Let's go for that solution for the moment though - I understand and agree with the concern of having a single way to handle `StringLiteral` expressions.

Scheremo: Switched the lowering path to `StringLiteral` -> `int` -> `StringType`, even though it makes my eyes bleed a little bit 😅

fabianschuiki: Yeah... this is pretty rough :rofl:

PR #9114, opened 2025-10-19
[ImportVerilog] Fix dynamic part-select index typing and add test

Scheremo:
Fixes incorrect zext/sext widths and mismatched constant signs in dynamic part-select extraction.

- Correct `getSelectIndex` to size and sign-extend indices based on full range bounds:
  - Signed if any bound is negative.
  - Bit-width = max(log2(max(|lo|, |hi|)+1), incoming width), with sign bit if signed.
- Apply consistent signedness to offset constants.
- Add regression tests:
  - `partselect_index_pos_be`: big-endian `[1:4]`, positive offset.
  - `partselect_index_neg_le`: little-endian `[0:-3]`, negative offset.

Further fixes third_party/tools/icarus/ivtest/ivltests/pr3054101[e-h].v in the chips alliance tests.


fabianschuiki (approved):
LGTM!

PR #9116, opened 2025-10-20
[FIRRTL] Add support for domain-connect driving instance-choice ports

rwy7:
The domain define op needs to verify that the destination has the same domain-type as the source. To do this, it needs to be able to read the domain type off of either operand. There is no generic way to do this yet so we have to support specific ops on a case-by-case basis. This PR adds support for sources and destinations defined by the instance-choice op.

At some point in the future we'll likely standardize an interface for reading domain information for results and arguments, but I'll save that for a later PR.


dtzSiFive:
Nice!

PR #9117, opened 2025-10-20
[FIRRTL] getPortNameStr to getPortName, getPortName to getPortNameAttr

rwy7:
Give these helpers more conventional names.


seldridge (approved):
Thanks for doing this cleanup! The inconsistency with the ODS-generated getters has been a consistent point of confusion.

PR #9119, opened 2025-10-20
[LLHD] Allow multiple layers of inlined functions to be followed by the same op

georgerennie:
This occurs when a function with empty body is inlined. Previously only one call would be popped off the call stack when the instruction following an inlined call was reached, meaning further uses of the unpopped function would be detected incorrectly as recursive.

This is fixed by repeatedly popping the call stack if possible.

### DIFF
--- a/lib/Dialect/LLHD/Transforms/InlineCalls.cpp
+++ b/lib/Dialect/LLHD/Transforms/InlineCalls.cpp
@@ -100,8 +100,9 @@ LogicalResult InlineCallsPass::runOnRegion(Region &region,
   // the inlined operations immediately after visiting the call.
   for (auto &block : region) {
     for (auto &op : block) {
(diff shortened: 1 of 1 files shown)


lib/Dialect/LLHD/Transforms/InlineCalls.cpp
```diff
@@ -101,7 +101,7 @@ LogicalResult InlineCallsPass::runOnRegion(Region &region,
   for (auto &block : region) {
     for (auto &op : block) {
       // If this is an op after a call, pop that call off the call stack.
```
tobiasgrosser:
```suggestion
      // While this is an op after a call, pop that call off the call stack.
```

georgerennie: I've reworded this to "Pop all calls that are followed by this op off the call stack."

PR #9120, opened 2025-10-21
[Arcilator] Print full values rather than truncating them.

SimonEbner:
Implemented by slicing the value into size_t chunks and passing them as variadic args to printf.

Example:
```
    // CHECK: v8 = fde6741e3a44d997e80d393ae643ee0fd9
    %v8 = arith.constant 0xfde6741e3a44d997e80d393ae643ee0fd9 : i137
    arc.sim.emit "v8", %v8 : i137
```

### DIFF
--- a/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
+++ b/lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
@@ -34,6 +34,7 @@
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/FormatVariadic.h"
 
 #define DEBUG_TYPE "lower-arc-to-llvm"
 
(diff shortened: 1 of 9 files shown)


SimonEbner:
Surprisingly this fails on windows, despite:
> The shrui operation shifts an integer value of the first operand to the right by the value of the second operand. The first operand is interpreted as unsigned, and the second operand is interpreted as unsigned. The high order bits are always filled with zeros. If the value of the second operand is greater or equal than the bitwidth of the first operand, then the operation returns poison.

Will debug this more in detail if this is useful to merge

test/arcilator/emit-values.mlir
```diff
@@ -0,0 +1,40 @@
+// RUN: arcilator --run %s | FileCheck %s
+module attributes { dlti.dl_spec = #dlti.dl_spec<
+  #dlti.dl_entry<index, 16>
+> } {
```
fzi-hielscher: Sadly, none of our CI runners use a 16 bit architecture. 😉 I expect the Windows test to pass without the `dl_spec`.

Personally, I would recommend getting rid of the dynamic `index`/`size_t` width and use `%llx` instead of `%zx`. It should be pretty safe to assume that `long long` is 64 bit.

test/arcilator/emit-values.mlir
```diff
@@ -0,0 +1,40 @@
...
+
+    // CHECK: v2 = abcdef
+    %v2 = arith.constant 0xABCDEF : i24
+    arc.sim.emit "v2", %v2 : i24
+
+    // CHECK: v3 = 123456789abcdef
+    %v3 = arith.constant 0x0123456789ABCDEF : i64
+    arc.sim.emit "v3", %v3 : i64
+
+    // CHECK: v4 = 0000000000001
```
fzi-hielscher: The zero padding behavior is a bit weird. Here we've got 13 characters for a `i64`. We should consistently either trim all leading zeros or pad to the full width of the type. I'd prefer the latter option.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -537,22 +537,59 @@ struct SimEmitValueOpLowering
...
-          IntegerType::get(getContext(), sizeOfSizeT.getFixedValue()), toPrint);
+
+    SmallString<16> printfFormatStr;
+
+    int remainingBits = valueType.getWidth();
+
+    Value value = adaptor.getValue();
+    int sizeTBits = sizeOfSizeT.getFixedValue();
+    auto sizeTType = IntegerType::get(getContext(), sizeTBits);
+    Value shiftValue = arith::ConstantOp::create(
```
fzi-hielscher: Nit: If there is a direct equivalent for the operation in the LLVM dialect I wouldn't take the indirection through `arith`.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -537,22 +537,59 @@ struct SimEmitValueOpLowering
...
+        // order.
+        // The leftmost highest-order chunk is printed without leading 0. All
+        // but the first element must be zero-padded to ensure the correct
+        // number printed.
+        printfVariadicArgs.push_back(
+            LLVM::TruncOp::create(rewriter, loc, sizeTType, value));
+        printfFormatStr.append(i == 0 ? "%zx" : sizeTPrintfFormatStr);
+
+        // Shift the value to the right by the size of size_t.
+        value = arith::ShRUIOp::create(rewriter, loc, value, shiftValue)
```
fzi-hielscher: As above.

test/arcilator/emit-values.mlir
SimonEbner: i was wondering the same. Updated to use %llx

test/arcilator/emit-values.mlir
SimonEbner: Makes sense. I was trying not to modify the default behaviour. Updated now to always show all leading zeros.

SimonEbner:
Thanks!

SimonEbner:
I see that by printing leading 0, some of the integration tests fail. Please advice what's the best solution. We could also use a custom printf function that trims all leading zeros correctly (also for >64bit values) which we link into the jit.

lib/Conversion/ArcToLLVM/LowerArcToLLVM.cpp
```diff
@@ -537,22 +538,53 @@ struct SimEmitValueOpLowering
     // Cast the value to a size_t.
     // FIXME: like the rest of MLIR, this assumes sizeof(intptr_t) ==
     // sizeof(size_t) on the target architecture.
-    Value toPrint = adaptor.getValue();
+    SmallVector<Value> printfVariadicArgs;
     DataLayout layout = DataLayout::closest(op);
     llvm::TypeSize sizeOfSizeT =
         layout.getTypeSizeInBits(rewriter.getIndexType());
     assert(!sizeOfSizeT.isScalable() &&
            sizeOfSizeT.getFixedValue() <= std::numeric_limits<unsigned>::max());
```
fzi-hielscher: Do we still need `sizeOfSizeT`?

fzi-hielscher (commented):
We'll definitely have to add custom print/formatting functions as part of a runtime library in the future. I had a go at that last year and learned the hard way that, to make it work consistently across platforms, there are a bunch of pitfalls regarding ABIs, libc implementations and linkage. It's doable, but not as straightforward as one would hope. 

Until then simply adapting the expected output of the integration tests to the new format should be fine. You could use a regex to ignore leading zeroes or just explicitly write them in.

SimonEbner:
Updated to add either a single leading zero or a 0* regex to the integration tests. I removed the size_t part too and changed slightly the implementation to distinguish between first and other chunks rather than only remainingBits.

OOC, regarding your point on ABI compatible prints: Wouldn't it be possible to create a vararg custom print that takes in int64 values, skips all leading ones that are zero - and starts printing from the first non-null value and rest? Probably not the most efficient but wouldn't that work irrespective of ABI ?

test/Dialect/Arc/lower-sim.mlir
```diff
@@ -52,14 +50,17 @@ module attributes { dlti.dl_spec = #dlti.dl_spec<
...
+
+  // CHECK-LABEL: llvm.func @print_i112
+  // CHECK-SAME:    %[[arg0:.*]]: i112
+  func.func @print_i112(%v: i112) {
+    // CHECK: %[[format_str3_ptr:.*]] = llvm.mlir.addressof @[[format_str3]] : !llvm.ptr
+    // CHECK: %[[shift:.*]] = llvm.mlir.constant(64 : i112)
+    // CHECK: %[[lower:.*]] = llvm.trunc %[[arg0]]
+    // CHECK: %[[shifted:.*]] = llvm.lshr %[[arg0]], %[[shift]]
+    // CHECK: %[[higher:.*]] = llvm.trunc %[[shifted]]
+    // %CHECK: llvm.call @printf(%[[format_str3_ptr]], %[[higher]], %[[lower]])
```
fzi-hielscher:
```suggestion
    // CHECK: llvm.call @printf(%[[format_str3_ptr]], %[[higher]], %[[lower]])
```

fzi-hielscher (approved):
Great, thanks!

> OOC, regarding your point on ABI compatible prints: Wouldn't it be possible to create a vararg custom print that takes in int64 values, skips all leading ones that are zero - and starts printing from the first non-null value and rest? Probably not the most efficient but wouldn't that work irrespective of ABI ?

Yeah, that should work. I'd still be hesitant to use a variadic function. The arcane ABI rules make it hard to interface with anything that isn't C, e.g. rust's `c_variadic` [is still unstable.](https://doc.rust-lang.org/beta/unstable-book/language-features/c-variadic.html#c_variadic)  I think the best way to deal with this, which should allow us to use types beyond simple integers, is to allocate a "marshaling buffer", manually arrange the raw argument data within it and then pass a pointer and some sort of type descriptor to the external function.

test/Dialect/Arc/lower-sim.mlir
SimonEbner: thanks, fixed

PR #9122, opened 2025-10-21
[Moore][ImportVerilog] Add real-valued ops + real-aware import; refactor logical handling

Scheremo:
Introduce full **real (f32/f64)** operator support in the Moore dialect and the ImportVerilog conversion, including unary, arithmetic, equality, relational, logical, and ++/-- forms. Refactor logical operator lowering to a reusable helper and add comprehensive tests.

- **Dialect (`MooreOps.td`):**
  - New unary real op: `NegRealOp` -> `moore.fneg`.
  - Real binary op base: `BinaryRealOpBase` for typed f32/f64 operands/results.
  - Arithmetic: `AddRealOp` (`fadd`), `SubRealOp` (`fsub`), `MulRealOp` (`fmul`), `DivRealOp` (`fdiv`), `PowRealOp` (`fpow`).
  - Equality: `EqRealOp` (`feq`), `NeRealOp` (`fne`).
  - Relational: `FltOp` (`flt`), `FleOp` (`fle`), `FgtOp` (`fgt`), `FgeOp` (`fge`).
  - All ops follow IEEE 1800-2023 §11.3.1, §11.4.3–§11.4.5 semantics.

- **Import (`Expressions.cpp`):**
  - Add `visitRealUOp` and `visitRealBOp` to lower real-typed unary/binary expressions.
  - Implement pre/post **increment/decrement** for real variables via `fadd`/`fsub` with an exact `APFloat` **"1.0"** (IEEEsingle/IEEEdouble) constructor.
  - Add `moore.fneg` for unary minus; unary plus becomes a no-op.
  - Map real comparisons to `feq/fne/flt/fle/fgt/fge`.
  - **Refactor logical ops** into `buildLogicalBOp(...)` (optional `Domain` param), shared by int/real paths. (Short-circuit remains TODO.)
  - Route mixed/real expressions early to the real path; keep existing integral flow otherwise.

- **Tests (`basic.sv`):**
  - New end-to-end checks for f64/f32 arithmetic, comparisons, logicals, unary ops, and ++/--.
  - Assert exact opcode emission (`moore.fadd/fsub/fmul/fdiv/fpow/feq/fne/flt/fle/fgt/fge/fneg`) and constant forms for `1.0`.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -580,7 +580,8 @@ def ShortrealLiteralOp : MooreOp<"shortreal_constant", [Pure]> {
     ```mlir
     %0 = moore.shortreal_constant 1.23
     ```
-    The shortreal type has fixed-point(1.2) and exponent(2.0e10) formats and corresponds to IEEE 754 f32.
+    The shortreal type has fixed-point(1.2) and exponent(2.0e10) formats and
+    corresponds to IEEE 754 f32.
     See IEEE 1800-2017 § 5.7.2 "Real literal constants".
   }];
   let arguments = (ins F32Attr:$value);
@@ -592,12 +593,13 @@ def RealLiteralOp : MooreOp<"real_constant", [Pure]> {
   let summary = "Produce a constant real value";
   let description = [{
     Produces a constant value of real type.
-    
+
     Example:
     ```mlir
     %0 = moore.real_constant 1.23
     ```
-    The real type has fixed-point(1.2) and exponent(2.0e10) formats and corresponds to IEEE 754 f64.
+    The real type has fixed-point(1.2) and exponent(2.0e10) formats and
+    corresponds to IEEE 754 f64.
     See IEEE 1800-2017 § 5.7.2 "Real literal constants".
   }];
   let arguments = (ins F64Attr:$value);
@@ -856,6 +858,21 @@ def NegOp : MooreOp<"neg", [Pure, SameOperandsAndResultType]> {
   }];
   let arguments = (ins SimpleBitVectorType:$input);
   let results = (outs SimpleBitVectorType:$result);
+  let assemblyFormat = [{
+    $input attr-dict `:` type($input)
+  }];
+}
+
+def NegRealOp : MooreOp<"fneg", [Pure, SameOperandsAndResultType]> {
+  let summary = "Arithmetic negation";
+  let description = [{
+    Negate a real value.
+
+    See IEEE 1800-2017 § 11.4.3 "Arithmetic operators" and § 11.3.1 "Operators
+    with real operands"
+  }];
+  let arguments = (ins RealType:$input);
+  let results = (outs RealType:$result);
   let assemblyFormat = [{
     $input attr-dict `:` type($input)
   }];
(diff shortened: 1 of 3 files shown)


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -861,6 +861,20 @@ def NegOp : MooreOp<"neg", [Pure, SameOperandsAndResultType]> {
   }];
 }
 
+def NegRealOp : MooreOp<"fneg", [Pure, SameOperandsAndResultType]> {
+  let summary = "Arithmetic negation";
+  let description = [{
+    Negate a real value.
+
+    See IEEE 1800-2017 § 11.4.3 "Arithmetic operators" and § 11.3.1 "Operators with real operands"
```
fabianschuiki: Nit: line break at 80 columns

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1386,6 +1400,124 @@ def ReplicateOp : MooreOp<"replicate", [
...
+    MooreOp<mnemonic, traits # [Pure, SameOperandsAndResultType]> {
+  let arguments = (ins RealType:$lhs, RealType:$rhs);
+  let results = (outs RealType:$result);
+  let assemblyFormat = [{
+    $lhs `,` $rhs attr-dict `:` type($result)
+  }];
+}
+
+def AddRealOp : BinaryRealOpBase<"fadd"> {
+  let summary = "Addition";
```
fabianschuiki: Nit: maybe "Add two real values"

fabianschuiki: You might be able to factor the `See IEEE ...` description into the `BinaryRealOpBase` class, and then use the summary for the "Add two real values", "Subtract the second real value from the first", etc.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1386,6 +1400,124 @@ def ReplicateOp : MooreOp<"replicate", [
...
+    Compares the bits in the left- and right-hand side operand and returns a
+    single bit 0, 1, or X result. If all corresponding bits in the left- and
+    right-hand side are equal, and all are 0 or 1 (not X or Z), the two operands
+    are considered equal (`eq` returns 1, `ne` returns 0). If any bits are not
+    equal, but all are 0 or 1, the two operands are considered not equal (`eq`
+    returns 0, `ne` returns 1). If any bit in the two operands is Z or X,
+    returns X. `eq` corresponds to the `==` operator and `ne` to the `!=`
+    operator.
+
+    See IEEE 1800-2017 § 11.4.5 "Equality operators" and § 11.3.1 "Operators with real operands".
```
fabianschuiki: Are there any X values in real comparisons? I thought these were limited to four-state integer comparisons, and comparison ops on reals would just produce a boolean.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -1386,6 +1400,124 @@ def ReplicateOp : MooreOp<"replicate", [
...
+  }];
+  let arguments = (ins RealType:$lhs, RealType:$rhs);
+  let results = (outs AnySingleBitType:$result);
```
fabianschuiki: Assuming that all equality/relational operators on reals just return a boolean, making all the comparisons uniform, you could use a comparison predicate argument in this op, similar to `comb.icmp` or `arith.cmpf`, and have the result be a a 1-bit `moore::IntType`. (I think we might already have a type constraint for that.)

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -596,6 +689,130 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+    else if (rhsFTy == slang::ast::FloatingType::Real)
+      bOpFTy = context.convertType(*rhsSlangFloatType, loc);
+    else if (lhsFTy == slang::ast::FloatingType::ShortReal)
+      bOpFTy = context.convertType(*lhsSlangFloatType, loc);
+    else if (rhsFTy == slang::ast::FloatingType::ShortReal)
+      bOpFTy = context.convertType(*rhsSlangFloatType, loc);
+
+    if (bOpFTy != moore::RealType::getReal(context.getContext()) &&
+        bOpFTy != moore::RealType::getShortReal(context.getContext()))
+      return {};
```
fabianschuiki: I would just check if the conversion was successful. The `convertType` function will emit an error and return `{}` if it failed. I would only return `{}` in that specific case. Otherwise there might be cases where no error gets emitted but the conversion still fails.
```suggestion
    if (!bOpFTy)
      return {};
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -596,6 +689,130 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+      rhsFTy = lhsSlangFloatType->floatKind;
+
+    if (lhsFTy == slang::ast::FloatingType::Real)
+      bOpFTy = context.convertType(*lhsSlangFloatType, loc);
+    else if (rhsFTy == slang::ast::FloatingType::Real)
+      bOpFTy = context.convertType(*rhsSlangFloatType, loc);
+    else if (lhsFTy == slang::ast::FloatingType::ShortReal)
+      bOpFTy = context.convertType(*lhsSlangFloatType, loc);
+    else if (rhsFTy == slang::ast::FloatingType::ShortReal)
+      bOpFTy = context.convertType(*rhsSlangFloatType, loc);
```
fabianschuiki: This feels redundant. Maytbe AI getting sloppy? How about something like:
```c++
if (!type)
  return {};
```
Like most other expression conversions do?

Scheremo: Well, the point of doing it this way is that either operand might be some other type, e.g. an integral or a string type or something else. Rather than descerning all of those, I decided to compare directly on the AST types.  
If you prefer some else we can switch it out, but in this case this was me being sloppy - I do it the same way for `UnaryExpression`, too :shrug:

include/circt/Dialect/Moore/MooreOps.td
Scheremo: This was left over from essentially copying 😅

include/circt/Dialect/Moore/MooreOps.td
Scheremo: Done! Good point.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Ah I see, sorry! The repeated ifs tricked me :grinning:. Slang should already insert the necessary conversion operations in most cases. And in all other cases, it should be sufficient to pick the expression's result type in the AST, convert that to MLIR, and then pass that to the `convertRvalueExpression` which you already do. In case we encounter any situations where slang doesn't insert conversion nodes (I've only seen this with packed structs in some rare cases), we can add a real-to-real cast (or int-to-real cast) to the `materializeConversion` function.

I'd go with something like:
```c++
if (!type)
  return {};
```
Basically pick the result type of the expression as Slang has computed it and run with it. That will then cause the `convertRvalueExpression` to fail with a proper error message in case we miss some conversion, and we go and add that :smiley:

Scheremo: Ah, I see, very valid point in this case! 
I got a bit code blind to the fact Slang should have conversion operators already. 

Trimmed it down and just added an assertion that rhsTy and lhsTy are already the same.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -596,6 +689,112 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+  Value visitRealBOp(const slang::ast::BinaryExpression &expr) {
+    Type lhsTy = context.convertType(*expr.left().type, loc);
+    Type rhsTy = context.convertType(*expr.right().type, loc);
+
+    if (!lhsTy || !rhsTy)
+      return {};
+
+    assert(lhsTy == rhsTy &&
+           "Slang should have converted both operands to the same type!");
+    Type targetTy = lhsTy;
```
fabianschuiki: I would use the type of the binary expression directly: `expr.type`. Slang computes that using the rules for operands outlined in the SV spec, including finding the widest operand width, signness matching, etc. I think it does the same thing for reals, where a `shortreal + real` would be determined to have type `real` for the `+` result.

That also means you don't need to assert equality on the LHS and RHS type. The assert will abort the compiler. If you don't assert, the conversions further down fail with a proper diagnostic pointing at the failing input, which is going to be much easier to debug and track down if it happens.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -611,6 +810,16 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+    // First check whether we need real or integral BOps
+    const auto *rhsFloatType =
+        expr.right().type->as_if<slang::ast::FloatingType>();
+    const auto *lhsFloatType =
+        expr.left().type->as_if<slang::ast::FloatingType>();
+
+    // If either op is real-typed, treat as real BOp. Bail here.
+    if (rhsFloatType || lhsFloatType)
+      return visitRealBOp(expr);
+
```
fabianschuiki, line 803: Same here: I think you can just check `expr.type` and see if it is a float. If it is, call out to `visitRealBOp`.

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: Yep, makes sense. Wasn't aware of `expr.type` but that simplifies this a lot!

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: Yep! Fixed it.

fabianschuiki (approved):
Really really cool! Thanks a lot for adding all of this 🥳 💯!

Scheremo:
> Really really cool! Thanks a lot for adding all of this 🥳 💯!

Thanks for the thorough review! Let's hope this fixes some more sv-tests :smile:

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: At least in here, we can't make the shortcut: a BinaryOperator::Equality will have return type `i1`, so this currently fails. This might also be a bug for the "normal" sbv pipeline if operands are cast to i1!

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: A bit too much as it turns out! `expr.type` holds the return type, which sometimes but not always is equal to the RHS types.

PR #9123, opened 2025-10-21
[ImportVerilog] Support real-to-integer conversion in format string

Scheremo:
Extend `FormatStringParser::emitInteger` to correctly handle **real-typed (`shortreal`/`real`) arguments** when emitting `%d`, `%h`, etc. format specifiers during import.

- Detect if the expression’s type is `moore::RealType`.
- For real operands:
  - Compute required integer width based on IEEE-754 format:
    - `shortreal` (binary32) → **129-bit signed integer**
    - `real` (binary64) → **1025-bit signed integer**
  - Insert a `moore::RealToIntOp` cast before integer formatting.
- Non-real (integral) values continue to use existing conversion logic.
- Ensures real values print correctly via implicit real-to-integer conversion.

### DIFF
--- a/lib/Conversion/ImportVerilog/FormatStrings.cpp
+++ b/lib/Conversion/ImportVerilog/FormatStrings.cpp
@@ -149,8 +149,36 @@ struct FormatStringParser {
   /// Emit an integer value with the given format.
   LogicalResult emitInteger(const slang::ast::Expression &arg,
                             const FormatOptions &options, IntFormat format) {
-    auto value =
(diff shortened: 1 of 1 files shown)


fabianschuiki (approved):
Holy cow, you can print reals with `%d` in Verilog 🤯? LGTM!

Scheremo:
> Holy cow, you can print reals with `%d` in Verilog 🤯? LGTM!

You should've seen my face when I first figured that out, and then again when I figured out how wide that integer would have to be 🙈

PR #9125, opened 2025-10-21
[FIRRTL] Add more port-related functionality to instance ops

rwy7:
- getNumPorts
- getPortLocation


seldridge (approved):
LGTM! These were definitely missing.

Should/could these be part of FInstanceLike?

rwy7:
> FInstanceLike

Yes I think that makes sense but I have been holding off on making changes to interfaces.

PR #9129, opened 2025-10-23
[ESI] Add bitvector and arbitrary-width integer classes

mortbopet:
To not constrain ESI C++ serialization to integer types that fit in `cstdint` types (<64 bits) as well as to allow non-byte aligned fields (e.g. structs with boolean values), this PR introduces the ESI `BitVector` type. It is a view on top of a `std::span<const uint8_t>` which tracks a byte- and bit index, and providing a bitvector view onto that bytearray.  
Deserialization has then been refactored to use such a `BitVector` view to pop values off of an incoming `MessageData`.

For serialization, we introduce `MutableBitVector`, which inherits from `BitVector`, but also owns the underlying byte array. This then allows for mutation operators to be implemented, facilitating serialization (e.g. shift-and-or style).

Finally, to support arbitrary-width integers, `Int` and `UInt` classes have been introduces. These contain constructors for their respective `cstdint` value, as well as conversion operators to `cstdint` classes. They do **not** contain arithmetic operators. Partly motivated by keeping this PR small, partly motivated by not being needed in the immediate usecase that this PR is motivated by.

### DIFF
--- a/lib/Dialect/ESI/runtime/cpp/lib/Values.cpp
+++ b/lib/Dialect/ESI/runtime/cpp/lib/Values.cpp
@@ -11,7 +11,6 @@
 #include <cstring>
 #include <format>
 
-
 using namespace esi;
 
 BitVector::BitVector(std::span<const byte> bytes, std::optional<size_t> width,


mortbopet:
Publish pipeline: https://github.com/llvm/circt/actions/workflows/esiRuntimePublish.yml

PR #9130, opened 2025-10-23
[Comb] Integer Overflow Analysis

cowardsa:
Building on the Integer Range Analysis for the Comb dialect. We now implement an overflow detection to determine when a particular operation truncates a carry-out bit. For example:

```Verilog
x[3:0] = a[3:0] + b[3:0]; // Carry-out truncated
y[4:0] = a[3:0] + b[3:0]; // Carry-out retained
````
This analysis will prove valuable for proving the correctness of downstream transformations to follow in a subsequent PR. The analysis inserts the result as an attribute that can be carried forward through a lowering pass to the datapath dialect, which will decompose basic operations, making them difficult to analyze locally.

### DIFF
--- a/lib/Dialect/Comb/Transforms/IntRangeAnnotations.cpp
+++ b/lib/Dialect/Comb/Transforms/IntRangeAnnotations.cpp
@@ -79,7 +79,9 @@ struct CombOpAnnotate : public OpRewritePattern<CombOpTy> {
     if (isa<comb::MulOp>(op))
       (void)a.umul_ov(b, overflowed);
 
-    op->setAttr("comb.nuw", BoolAttr::get(op->getContext(), overflowed));
+    if (!overflowed)
+      op->setAttr("comb.nuw", UnitAttr::get(op->getContext()));
+
     return success();
   }
 
--- a/include/circt/Dialect/Comb/Passes.td
+++ b/include/circt/Dialect/Comb/Passes.td
@@ -35,12 +35,11 @@ def CombIntRangeNarrowing : Pass<"comb-int-range-narrowing"> {
 }
 
 def CombOverflowAnnotating : Pass<"comb-overflow-annotating"> {
-  let summary = "Annotate comb ops with overflow signal.";
+  let summary = "Annotate comb ops with no overflow signal (nuw).";
   let description = [{
-    Using integer range analysis, annotate comb ops which may overflow with
-    LLVMs no unsigned wrap (nuw) signal that indicates whether the output
-    truncates carry-out bits. This is useful for downstream passes which need to
-    handle overflow conditions.
+    Using integer range analysis, annotate comb ops which cannot overflow with
+    LLVMs no unsigned wrap (nuw) signal. This is useful for downstream passes 
+    which need to handle overflow conditions.
   }];
 }
 
--- a/test/Dialect/Comb/comb-overflow-annotating.mlir
+++ b/test/Dialect/Comb/comb-overflow-annotating.mlir
@@ -3,10 +3,10 @@
 // CHECK-LABEL: @basic_add
 hw.module @basic_add(in %a : i4, in %b : i4, out add4 : i4, out add5 : i5) {
   // CHECK-NEXT: %false = hw.constant false
-  // CHECK-NEXT: comb.add %a, %b {comb.nuw = true} : i4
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: lib/Dialect/Comb/Transforms/IntRangeAnnotations.cpp [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+  solver.load<IntegerRangeAnalysis>();
+  if (failed(solver.initializeAndRun(op)))
+    return signalPassFailure();
+
+  RewritePatternSet patterns(ctx);
+  // TODO: determine how to support subtraction
+  patterns.add<CombOpAnnotate<comb::AddOp>, CombOpAnnotate<comb::MulOp>>(
+      patterns.getContext(), solver);
+
+  if (failed(applyPatternsGreedily(op, std::move(patterns))))
```
uenoku: Please use [WalkPatternRewriteDriver](https://github.com/llvm/llvm-project/blob/9681705fbc3e16810ed031ca9bdd4b78654a3058/mlir/include/mlir/Transforms/WalkPatternRewriteDriver.h#L33) since this is one-shot.

lib/Dialect/Comb/Transforms/IntRangeAnnotations.cpp
```diff
@@ -0,0 +1,116 @@
...
+                                PatternRewriter &rewriter) const override {
+    // Use upper_bit_trunc attribute to avoid confusion with IEEE overflow flags
+    if (op->hasAttr("comb.upper_bit_trunc"))
+      return failure();
+
+    if (op->getNumOperands() != 2 || op->getNumResults() != 1)
+      return rewriter.notifyMatchFailure(
+          op, "Only support binary operations with one result");
+
+    if (!isa<comb::AddOp>(op) && !isa<comb::MulOp>(op) && !isa<comb::SubOp>(op))
```
uenoku: Probably assertion is ok (maybe static_assert  is better) since CombOpTy is a template type.
```suggestion
    assert(isa<comb::AddOp, comb::MulOp, comb::SubOp>(op));
```

cowardsa:
Thanks for the comments @uenoku - have implemented the non-blocking changes.

With regards to making overflow flags part of the operator - I think we need to agree on what the equivalent statement is for comb (from the arith description):
"If the `nuw` and/or `nsw` flags are present, and an unsigned/signed overflow occurs (respectively), the result is poison."

Presumably the same is also sort of true - but I guess one would not want your synthesis engine to just optimise away hardware which it thinks is poison?

lib/Dialect/Comb/Transforms/IntRangeAnnotations.cpp
```diff
@@ -0,0 +1,113 @@
...
+    // Only use the overflow flag - need to explicitly discard the result.
+    // Cannot use operators intervals since they will be truncated to the
+    // result's bitwidth.
+    if (isa<comb::AddOp>(op))
+      (void)a.uadd_ov(b, overflowed);
+
+    if (isa<comb::MulOp>(op))
+      (void)a.umul_ov(b, overflowed);
+
+    op->setAttr("comb.nuw", BoolAttr::get(op->getContext(), overflowed));
```
uenoku: Maybe polarity is opposite? Also can we use UnitAttr instead of BoolAttr? 
```suggestion
    if(!overflowed)
      op->setAttr("comb.nuw", UnitAttr::get(op->getContext()));
```

uenoku:
> "If the nuw and/or nsw flags are present, and an unsigned/signed overflow occurs (respectively), the result is poison."
> Presumably the same is also sort of truex

I think this is true. 

> but I guess one would not want your synthesis engine to just optimise away hardware which it thinks is poison?

I agree we might not exploit poison to optimize aggregissively like LLVM does, but however every transformation that relies on nuw flag will assume that the operation never overflow, so the transformation will be entirely invalid if the operation actually overflows. From that perspective it's actually valid to consider overflow with nuw flag to be poison and replace them with arbitary value.

lib/Dialect/Comb/Transforms/IntRangeAnnotations.cpp
cowardsa: Apologies - yes my mistake - will update the polarity

PR #9131, opened 2025-10-23
Add sv.module.verbatim op and use for firrtl.extmodule with BlackBoxInlineAnno/BlackBoxPathAnno

tmckay-sifive:
This PR adds an `sv.verbatim.module` op and lowers inline black boxes (`firrtl.extmodule` with `BlackBoxInlineAnno` or `BlackBoxPathAnno`) to the new op.

Emitted verilog is semantically the same (there can be a difference in emission order)--so, this is mostly an IR-only change.

Before:

- BlackBoxReader
  - For each `firrtl.extmodule` with `BlackBoxInlineAnno` or `BlackBoxPathAnno`, create `emit.file` and `emit.verbatim` operations for the inline content
- LowerToHW pass: For each `firrtl.extmodule`, creates a `hw.module.extern` operation

After:

- BlackBoxReader
  - For each `firrtl.extmodule` with verbatim annotations:
    - Process each `BlackBoxInlineAnno` and `BlackBoxPathAnno` to calculate the verbatim content and determine the output file path
    - Store this information in a new annotation type designed for consumption by the LowerToHW pass
    - Remove the original `BlackBoxInlineAnno`/`BlackBoxPathAnno` annotations
- LowerToHW:
  - For each `firrtl.extmodule` with the new verbatim annotation:
    - On the first verbatim annotation encountered, create a `sv.verbatim.module` operation
    - For additional verbatim annotations on the same module, add them to the `additional_files` attribute of the existing `sv.verbatim.module` operation
    - Do not create `hw.module.extern` for modules with verbatim content

### BlackBoxReader

It is a bit strange to partially lower the `BlackBoxInlineAnno` `BlackBoxPathAnno`. The intention here was to preserve the fragile logic around calculating output directories for black boxes.

Eventually, I think we can remove BlackBoxReader entirely and rely on AssignOutputDirs for this. Then, by the time we get to LowerToHW, we don't need to special case any of this stuff for black boxes.

### Supporting multiple files

firtool currently accepts multiple instances of `BlackBoxInlineAnno` and `BlackBoxPathAnno` annotations. This is often used to include additional SV libraries, C files, or headers. In the future, we should disallow this and ensure that one module corresponds to one SV file. Before we do this however, we need to provide another mechanism for specifying external dependencies.

The `addition_files` attribute is a way to support the existing behavior in the meantime. Once we have addressed this tech-debt, we can just remove that attribute.

### DIFF
--- a/integration_test/Bindings/Python/dialects/sv.py
+++ b/integration_test/Bindings/Python/dialects/sv.py
@@ -37,11 +37,12 @@ with ir.Context() as ctx, ir.Location.unknown() as loc:
     verbatim = sv.VerbatimModuleOp(
         name="MyVerbatim",
         content=verbatim_content,
-        output_file=hw.OutputFileAttr.get_from_filename(ir.StringAttr.get("MyVerbatim.v"), False, False),
+        output_file=hw.OutputFileAttr.get_from_filename(
+            ir.StringAttr.get("MyVerbatim.v"), False, False),
         input_ports=[("clk", i1)],
         output_ports=[("out", i1)],
-        additional_files=ir.ArrayAttr.get([ir.FlatSymbolRefAttr.get("test_header")])
-    )
+        additional_files=ir.ArrayAttr.get(
+            [ir.FlatSymbolRefAttr.get("test_header")]))
 
     print(verbatim)
     # CHECK: "sv.verbatim.module"() <{additional_files = [@test_header], content = "module MyVerbatim(input clk, output reg out); always @(posedge clk) out <= ~out; endmodule", module_type = !hw.modty<input clk : i1, output out : i1>, output_file = #hw.output_file<"MyVerbatim.v">, parameters = [], per_port_attrs = [], sym_name = "MyVerbatim"}> : () -> ()


tmckay-sifive:
> The implementation direction makes sense, could you include tests for ExportVerilog, lower-to-hw, and sv dialect?

Yep, will do.

> It's a bit unclear to me how additional_files are used.

Sorry, I meant to talk about that in the PR description. I've added it but I'll paste it here too:

> firtool currently accepts multiple instances of BlackBoxInlineAnno and BlackBoxPathAnno annotations. This is often used to include additional SV libraries, C files, or headers. In the future, we should disallow this and ensure that one module corresponds to one SV file. Before we do this however, we need to provide another mechanism for specifying external dependencies.

> The addition_files attribute is a way to support the existing behavior in the meantime. Once we have addressed this tech-debt, we can just remove that attribute.

tmckay-sifive:
> The implementation direction makes sense, could you include tests for ExportVerilog, lower-to-hw, and sv dialect?

I added some tests for LowerToHW and ExportVerilog. I wasn't sure what SV dialect tests should exercise that isn't covered by the LowerToHW, ExportVerilog, or BlackBox tests. If you have something in mind please let me know.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1155,12 +1155,126 @@ FIRRTLModuleLowering::lowerExtModule(FExtModuleOp oldModule,
...
-  // Build the new hw.module op.
   auto builder = OpBuilder::atBlockEnd(topLevelModule);
   auto nameAttr = builder.getStringAttr(oldModule.getName());
-  // Map over parameters if present.  Drop all values as we do so, so there are
-  // no known default values in the extmodule.  This ensures that the
-  // hw.instance will print all the parameters when generating verilog.
+
+  // Check for verbatim black box annotation
+  AnnotationSet annos(oldModule);
+  Annotation verbatimAnno;
```
uenoku:
```suggestion
  Annotation verbatimAnno = annos.getAnnotation(verbatimBlackBoxAnnoClass);
```

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1155,12 +1155,126 @@ FIRRTLModuleLowering::lowerExtModule(FExtModuleOp oldModule,
...
+    if (anno.isClass(verbatimBlackBoxAnnoClass)) {
+      verbatimAnno = anno;
+      isVerbatimSv = true;
+      break;
+    }
+  }
+
+  // Lower to sv.verbatim.module instead of hw.module.extern when the blackbox
+  // has verbatim verilog content.
+  if (isVerbatimSv) {
```
uenoku:
```suggestion
  if (verbatimAnno) {
```

include/circt/Dialect/FIRRTL/AnnotationDetails.h
```diff
@@ -58,6 +58,7 @@ constexpr const char *blackBoxTargetDirAnnoClass =
     "firrtl.transforms.BlackBoxTargetDirAnno";
 constexpr const char *blackBoxAnnoClass =
     "firrtl.transforms.BlackBox"; // Not in SFC
+constexpr const char *verbatimBlackBoxAnnoClass = "circt.VerbatimBlackBoxAnno";
```
uenoku, line 61: Could you update FIRRTLAnnotations.md?

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1155,12 +1155,126 @@ FIRRTLModuleLowering::lowerExtModule(FExtModuleOp oldModule,
...
+      auto fileDict = cast<DictionaryAttr>(filesAttr[i]);
+      auto fileName = fileDict.getAs<StringAttr>("name");
+      auto fileContent = fileDict.getAs<StringAttr>("content");
+      auto fileOutputFile = fileDict.getAs<StringAttr>("output_file");
+
+      if (fileName && fileContent && fileOutputFile &&
+          !seenFiles.contains(fileName.getValue())) {
+        seenFiles.insert(fileName.getValue());
+
+        auto fileSymbolName = "blackbox_" + fileName.getValue().str();
```
uenoku: Please use circuitNamespace to get an unique symbol.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1155,12 +1155,126 @@ FIRRTLModuleLowering::lowerExtModule(FExtModuleOp oldModule,
...
+    }
+
+    // Create emit.file operations for additional files (these are usually
+    // additional collateral such as headers or DPI files).
+    for (size_t i = 1; i < filesAttr.size(); ++i) {
+      auto fileDict = cast<DictionaryAttr>(filesAttr[i]);
+      auto fileName = fileDict.getAs<StringAttr>("name");
+      auto fileContent = fileDict.getAs<StringAttr>("content");
+      auto fileOutputFile = fileDict.getAs<StringAttr>("output_file");
+
```
uenoku, line 1219: Could you return null if there is malformed annotations?

include/circt/Dialect/SV/SVTypeDecl.td
```diff
@@ -206,6 +208,50 @@ def ModportType : SVType<"Modport"> {
...
+    To support verbatim content split accross multiple files, there is an optional
+    additional_files field which can reference additional emitted files.
+  }];
+
+  let arguments = (ins SymbolNameAttr:$sym_name,
+      TypeAttrOf<ModuleType>:$module_type, StrAttr:$content,
+      OutputFileAttr:$output_file, OptionalAttr<DictArrayAttr>:$per_port_attrs,
+      OptionalAttr<LocationArrayAttr>:$port_locs,
+      ParamDeclArrayAttr:$parameters,
+      OptionalAttr<SymbolRefArrayAttr>:$additional_files);
```
uenoku: Could you add (or leave TODO comment) verilogName attribute https://github.com/llvm/circt/blob/b7821984dc43e206d8499c5cff85bbf3904f8d0e/include/circt/Dialect/HW/HWStructure.td#L256 because symbol may not match with verilog name?

uenoku: Please add SymbolUserOpInterface to SVVerbatimModuleOp since additional_files refer to symbols.

lib/Conversion/FIRRTLToHW/LowerToHW.cpp
```diff
@@ -1155,12 +1155,126 @@ FIRRTLModuleLowering::lowerExtModule(FExtModuleOp oldModule,
...
+        parameters,
+        additionalFiles.empty() ? nullptr
+                                : builder.getArrayAttr(additionalFiles));
+
+    SymbolTable::setSymbolVisibility(
+        verbatimModule, SymbolTable::getSymbolVisibility(oldModule));
+
+    annos.removeAnnotations([](Annotation anno) {
+      return anno.isClass(verbatimBlackBoxAnnoClass);
+    });
```
uenoku: I think this will work
```suggestion
    annos.removeAnnotation(verbatimBlackBoxAnnoClass);
```

uenoku (commented):
Generally looks great! Could you add round-trip test for SV dialect?

lib/Bindings/Python/dialects/sv.py
```diff
@@ -107,3 +107,56 @@ def create(value):
...
+                       input_ports,
+                       output_ports,
+                       parameters=parameters,
+                       attributes=attributes,
+                       body_builder=None,
+                       loc=loc,
+                       ip=ip)
+
+  def instantiate(self, *args, **kwargs):
+    return hw.ModuleLike.instantiate(self, *args, **kwargs)
```
uenoku, line 146: Could you add python binding tests under integration tests?

lib/Conversion/ExportVerilog/ExportVerilog.cpp
```diff
@@ -196,6 +196,8 @@ StringRef ExportVerilog::getSymOpName(Operation *symOp) {
   return TypeSwitch<Operation *, StringRef>(symOp)
       .Case<HWModuleOp, HWModuleExternOp, HWModuleGeneratedOp, FuncOp>(
           [](Operation *op) { return getVerilogModuleName(op); })
+      .Case<sv::SVVerbatimModuleOp>(
+          [](sv::SVVerbatimModuleOp op) { return op.getVerilogModuleName(); })
```
uenoku: super nit: maybe we don't`sv::` for the consistency with surrounding code.

test/Dialect/SV/sv-verbatim-module.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: circt-opt %s | FileCheck %s
```
uenoku: super nit: pleas move this test to https://github.com/llvm/circt/blob/main/test/Dialect/SV/basic.mlir as it's very small.

lib/Dialect/SV/SVOps.cpp
```diff
@@ -1944,6 +1946,318 @@ void BindOp::build(OpBuilder &builder, OperationState &odsState, StringAttr mod,
...
+SVVerbatimModuleOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
+  if (auto additionalFiles = getAdditionalFiles()) {
+    auto *symtable = SymbolTable::getNearestSymbolTable(*this);
+    if (!symtable)
+      return emitError("sv.verbatim.module must exist within a region "
+                       "which has a symbol table.");
+
+    for (auto fileRef : *additionalFiles) {
+      auto symbolRef = llvm::cast<SymbolRefAttr>(fileRef);
+      auto *referencedOp = symbolTable.lookupSymbolIn(symtable, symbolRef);
```
uenoku: please remove `getnearestSymbolTable(*this)` above since it walks entire IR.
```suggestion
symbolTable.lookupNearestSymbolFrom(getOperation(), symbolRef);
```

lib/Dialect/SV/SVOps.cpp
```diff
@@ -1944,6 +1946,318 @@ void BindOp::build(OpBuilder &builder, OperationState &odsState, StringAttr mod,
...
+  }
+
+  auto newModuleType = hw::ModuleType::get(getContext(), newPorts);
+  setModuleTypeAttr(TypeAttr::get(newModuleType));
+}
+
+LogicalResult SVVerbatimModuleOp::verify() {
+  auto moduleType = getModuleType();
+  if (!moduleType)
+    return emitOpError("missing module type");
```
uenoku: I think this is verified by ODS already so I don't think it's necessary to re-verify here.

lib/Dialect/SV/SVOps.cpp
```diff
@@ -1944,6 +1946,318 @@ void BindOp::build(OpBuilder &builder, OperationState &odsState, StringAttr mod,
...
+  if (!outputFileAttr) {
+    return parser.emitError(
+        parser.getCurrentLocation(),
+        "sv.verbatim.module requires 'output_file' attribute");
+  }
+  if (!isa<hw::OutputFileAttr>(outputFileAttr)) {
+    return parser.emitError(
+        parser.getCurrentLocation(),
+        "sv.verbatim.module 'output_file' attribute must be an OutputFileAttr");
+  }
```
uenoku: nit: Please use a single brace. 
```suggestion
  if (!result.attributes.get("content")) 
    return parser.emitError(parser.getCurrentLocation(),
                            "sv.verbatim.module requires 'content' attribute");
  

  // parse output file
  auto outputFileAttr = result.attributes.get("output_file");
  if (!outputFileAttr) 
    return parser.emitError(
        parser.getCurrentLocation(),
        "sv.verbatim.module requires 'output_file' attribute");
  
  if (!isa<hw::OutputFileAttr>(outputFileAttr)) 
    return parser.emitError(
        parser.getCurrentLocation(),
        "sv.verbatim.module 'output_file' attribute must be an OutputFileAttr");
  
```

lib/Dialect/SV/SVOps.cpp
```diff
@@ -1944,6 +1946,318 @@ void BindOp::build(OpBuilder &builder, OperationState &odsState, StringAttr mod,
...
+    for (auto fileRef : *additionalFiles) {
+      auto symbolRef = llvm::cast<SymbolRefAttr>(fileRef);
+      auto *referencedOp = symbolTable.lookupSymbolIn(symtable, symbolRef);
+      if (!referencedOp)
+        return emitError("Symbol not found: ") << symbolRef << ".";
+
+      if (!isa<circt::emit::FileOp>(referencedOp))
+        return emitError("Symbol ")
+               << symbolRef << " is not an emit.file operation.";
+    }
```
uenoku, line 2235: Could you add a test in https://github.com/llvm/circt/blob/main/test/Dialect/SV/errors.mlir?

lib/Dialect/FIRRTL/Transforms/BlackBoxReader.cpp
```diff
@@ -185,6 +177,9 @@ void BlackBoxReaderPass::runOnOperation() {
   auto bboxAnno =
       builder.getDictionaryAttr({{builder.getStringAttr("class"),
                                   builder.getStringAttr(blackBoxAnnoClass)}});
+  // Track which modules reference which files
+  DenseMap<Operation *, SmallVector<StringAttr>> verbatimExtmoduleToFile;
```
uenoku: Please use MapVector to prevent non-determinism since verbatimExtmoduleToFile is iterated.

PR #9132, opened 2025-10-23
[FIRRTL] Various reduction pattern tweaks

fabianschuiki:
This is a set of tweaks and improvements to reduction patterns as I've been debugging a few FIRRTL dedup and class lowering issues. I plan to land this using a rebase merge, not a squash merge, such that the individual independent commits stay separated.


fabianschuiki:
Landing this since it only touches circt-reduce.

PR #9133, opened 2025-10-24
[Moore] [1/4] Implement ClassHandleType, ClassDeclarationOp

Scheremo:
This change introduces first-class **class** support to the Moore dialect and teaches the Verilog importer to **recognize and lower SV classes** into that IR. It also adds an opaque **class handle type** for values of class objects.

- New **IR ops** for classes: declarations, properties, and a class-body terminator.
- New **type**: `!moore.class.object<@Sym>` (with `ref` variant).
- Importer now **creates symbols for classes**, wires up `extends`/`implements`, and lowers **class-typed values** to `!moore.class.object<…>`.
- Extensive **tests** for inheritance, interfaces, properties, scoping, and handle declarations.

### DIFF
--- a/include/circt/Dialect/Moore/MooreTypes.td
+++ b/include/circt/Dialect/Moore/MooreTypes.td
@@ -52,6 +52,19 @@ def EventType : MooreTypeDef<"Event", [], "moore::UnpackedType"> {
   let summary = "the SystemVerilog `event` type";
 }
 
+//===----------------------------------------------------------------------===//
+// ClassHandleType
+//===----------------------------------------------------------------------===//
+
+def ClassHandleType : MooreTypeDef<"ClassHandle", [], "moore::UnpackedType"> {
+  let mnemonic = "class";
+  let summary = "Class object handle type, pointing to an object on the heap.";
+  let description = [{
+      The `!moore.class<@C>` type represents a handle of a class instance of class `@C` living on the program's heap.
+  }];
+  let parameters = (ins AttrParameter<"::mlir::SymbolRefAttr", "class symbol">:$classSym);
+  let assemblyFormat = "`<` $classSym `>`";
+}
 
 //===----------------------------------------------------------------------===//
 // TimeType
@@ -559,4 +572,6 @@ def UnionRefType : SpecificRefType<UnionType>;
 def UnpackedUnionRefType : SpecificRefType<UnpackedUnionType>;
 def AnyUnionRefType : SpecificRefType<AnyUnionType>;
 
+/// Class references.
+def ClassHandleRefType : SpecificRefType<ClassHandleType>;
 #endif // CIRCT_DIALECT_MOORE_MOORETYPES
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -8,6 +8,7 @@
 
 #include "ImportVerilogInternals.h"
 #include "slang/ast/Compilation.h"
+#include "slang/ast/symbols/ClassSymbols.h"
 #include "llvm/ADT/ScopeExit.h"
 
 using namespace circt;
@@ -52,6 +53,11 @@ struct BaseVisitor {
   // sake of name resolution, such as enum variant names.
   LogicalResult visit(const slang::ast::TransparentMemberSymbol &) {
     return success();
+  }
+
+  // Handle classes without parameters
+  LogicalResult visit(const slang::ast::ClassType &classdecl) {
+    return context.convertClassDeclaration(classdecl);
   }
 
   // Skip typedefs.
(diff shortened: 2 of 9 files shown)


include/circt/Dialect/Moore/MooreClasses.td
```diff
@@ -0,0 +1,58 @@
...
+    // Convenience typed accessor for the stored type.
+    ::mlir::Type getPropertyType() { return getTypeAttr().getValue(); }
+  }];
+}
+
+def ClassBodyEndOp
+    : MooreOp<"class.declend", [Terminator, HasParent<"ClassDeclOp">]> {
+  let summary = "end of classdecl region";
+  let assemblyFormat = "attr-dict";
+}
```
fabianschuiki: I think you can mark the `ClassDeclOp` with the `NoTerminator` trait. That way we don't need an explicit terminator op.

include/circt/Dialect/Moore/MooreClasses.td
```diff
@@ -0,0 +1,58 @@
...
+  // Printed as:  moore.class.propertydecl @foo : !moore.ref<!moore.i32> { ... }
+  let assemblyFormat = [{
+     $sym_name `:` $type  attr-dict
+  }];
+
+  let extraClassDeclaration = [{
+    // Convenience typed accessor for the stored type.
+    ::mlir::Type getPropertyType() { return getTypeAttr().getValue(); }
+  }];
+}
```
fabianschuiki: Since there are probably only a handful of ops for classes, it might make sense to just add them to `MooreOps.td` where all the other ops live. WDYT?

lib/Dialect/Moore/MooreClasses.cpp
```diff
@@ -0,0 +1,35 @@
...
+    // allow only property and method decls and terminator
+    if (llvm::isa<circt::moore::ClassBodyEndOp,
+                  circt::moore::ClassPropertyDeclOp>(&op))
+      continue;
+
+    return emitOpError()
+           << "body may only contain 'moore.class.propertydecl' operations";
+  }
+  return mlir::success();
+}
```
fabianschuiki: I would add this to `MooreOps.cpp`. That will give us an opportunity to think about splitting things apart later if we chose to.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -52,6 +52,34 @@ def EventType : MooreTypeDef<"Event", [], "moore::UnpackedType"> {
...
+  let assemblyFormat = "`<` $classSym `>`";
+  let extraClassDeclaration = [{
+    /// True if this handle’s class is the same as, or derived from, `baseName`.
+    /// `baseName` must be the root symbol name (e.g. @"pkg::Class").
+    bool isSameOrDerivedFrom(::mlir::Operation *anchor,
+                             ::mlir::SymbolRefAttr baseName);
+    /// Find the lowest ancestor of this class which has a property `fieldName`.
+    ClassHandleType ancestorWithProperty(Operation *anchor,
+                                         StringRef fieldName);
+  }];
```
fabianschuiki: Are these ever used? It's usually a good idea to only need the functionality that is immediately needed and useful for the PR.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -52,6 +52,34 @@ def EventType : MooreTypeDef<"Event", [], "moore::UnpackedType"> {
...
+
+def ClassHandleType : MooreTypeDef<"ClassHandle", [], "moore::UnpackedType"> {
+  let mnemonic = "class.object";
+  let summary = "Class handle type";
+  let description = [{
+    Class handle type representing `!moore.class.object<@C>`. The parameter is a
+    SymbolRef to the class declaration symbol. The handle is opaque here.
+  }];
+  let parameters = (ins AttrParameter<"::mlir::SymbolRefAttr", "class symbol">:$classSym);
+  let assemblyFormat = "`<` $classSym `>`";
```
fabianschuiki, line 66: If I recall correctly from some earlier discussions, you're going for something like `!moore.ref<!moore.class.object>` as the general way of passing handles to classes around, right? In that case, I would drop the "Handle" from the type name, since the `ref<...>` is the handle, and the `class.object` is the actual class value, probably a struct aggregate of its field. I'd go for `ClassType`, and `!moore.class`, just to keep it short and indicate that these are actual values, not references. The reference comes in through `ref<class<@C>>`.

include/circt/Dialect/Moore/MooreTypes.td
```diff
@@ -52,6 +52,34 @@ def EventType : MooreTypeDef<"Event", [], "moore::UnpackedType"> {
...
 }
 
+//===----------------------------------------------------------------------===//
+// ClassHandleType
+//===----------------------------------------------------------------------===//
+
+// !moore.class.object<@C>
+// Encodes a handle of class instance living on the program's heap,
+// whose dynamic layout begins with a vptr.
+// The parameter is the class symbol (e.g., @C).
```
fabianschuiki: I'd drop these comments in favor of the nice "description" in the class type. That one gets baked into the dialect documentation and makes it easier for people to discover 😃

```suggestion
```

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -6,8 +6,11 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include <ranges>
+
```
fabianschuiki: Possible stray include?

include/circt/Dialect/Moore/MooreClasses.td
```diff
@@ -0,0 +1,58 @@
...
+include "mlir/Interfaces/ControlFlowInterfaces.td"
+
+def ClassPropertyDeclOp
+    : MooreOp<"class.propertydecl", [Symbol, HasParent<"ClassDeclOp">]> {
+  let summary = "Declare a class property (variable)";
+
+  let arguments = (ins SymbolNameAttr:$sym_name, TypeAttr:$type);
+
+  let results = (outs);
+  // Printed as:  moore.class.propertydecl @foo : !moore.ref<!moore.i32> { ... }
```
fabianschuiki:
```suggestion
```

include/circt/Dialect/Moore/MooreClasses.td
```diff
@@ -0,0 +1,58 @@
...
+  let assemblyFormat = [{
+     $sym_name `:` $type  attr-dict
+  }];
+
+  let extraClassDeclaration = [{
+    // Convenience typed accessor for the stored type.
```
fabianschuiki:
```suggestion
```

include/circt/Dialect/Moore/MooreClasses.td
```diff
@@ -0,0 +1,58 @@
...
+  let arguments = (ins SymbolNameAttr:$sym_name,
+      OptionalAttr<SymbolRefAttr>:$base,
+      OptionalAttr<SymbolRefArrayAttr>:$implementedInterfaces);
+  let results = (outs);
+  let regions = (region AnyRegion:$body);
+  let assemblyFormat = [{
+    $sym_name
+    (`extends` $base^)?
+    (`implements` $implementedInterfaces^)?
+    `:` $body attr-dict-with-keyword
```
fabianschuiki: I'd drop the `:` which is commonly only used to separate return or operand types in the assembly. And pull the attributes dictionary in front of the body, to not have the attributes dangling after a potentially large class body.

```suggestion
    attr-dict-with-keyword $body
```

include/circt/Dialect/Moore/MooreClasses.td
Scheremo: Good idea, thank you!

include/circt/Dialect/Moore/MooreClasses.td
Scheremo: There will be a few more (new, delete, call, vcall, property_ref), but in general I don't have a strong opinion. Shall I move this to `MooreOps.td`?

lib/Dialect/Moore/MooreClasses.cpp
Scheremo: Can do!

include/circt/Dialect/Moore/MooreTypes.td
Scheremo: Fair enough; they are a remnant from splitting commits 😄

include/circt/Dialect/Moore/MooreTypes.td
Scheremo: I was thinking about using the same logic as for `CHandleType`, i.e. a type that translates directly to an opaque pointer (e.g. `!llvm.ptr`), since by spec alone objects should (always?) be heap-allocated and handling of class handles is "very opaque". Since we don't directly deal with the heap in System Verilog (and moore as far as I can tell), I think keeping these heap-allocated values opaque makes sense. In that sense `!moore.ref<!moore.class<@C>>` is a reference to a handle. Does that make sense to you?

include/circt/Dialect/Moore/MooreTypes.td
Scheremo: Fair enough. Folded them in.

include/circt/Dialect/Moore/MooreClasses.td
Scheremo: Sure!

include/circt/Dialect/Moore/MooreTypes.td
Scheremo: The relevant Section is 1800-2023 Section 8.4 "Objects (class instances)" 🙂

fabianschuiki: Ah I see, that makes sense! The `ClassHandle` itself is essentially a pointer or fat pointer to the class, and a `!moore.ref<!moore.class<@C>>` is something like a reference to a variable holding that pointer. That makes total sense! So all class operations work on `!moore.class<@C>` pointers?

Scheremo: Yes exactly! 👍

PR #9134, opened 2025-10-24
[Moore] [2/4] Implement new operator

Scheremo:
Adds heap allocation support for class objects.

- **New op:** `moore.class.new`
  - Allocates a new instance of a class symbol (`!moore.class.object<@C>`).
  - Implements `MemoryEffectsOpInterface` (reports `Allocate`).
  - Verifies that the referenced class symbol exists and is a valid `ClassDeclOp`.

- **Importer:**
  - `ImportVerilog/Expressions.cpp` now lowers `new` expressions to `moore.class.new`.
  - Emits an error if a constructor call is present (not yet supported).

- **Tests:**
  - Added `testModule3` in `classes.sv` to verify `new` lowering and expected IR.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -2395,5 +2395,22 @@ def ClassDeclOp
   let hasVerifier = 1;
 }
 
+def ClassNewOp
+    : MooreOp<
+          "class.new", [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
+  let summary = "Allocate a new class instance";
+  let description = [{
+    Allocates a new instance of class @C. This op does not call the constructor.
+    The result is a non-null `!moore.class.object<@C>` handle.
+  }];
+
+  let arguments = (ins);
+  let results = (outs ClassHandleType:$result);
+
+  let assemblyFormat = [{
+     `:` type($result) attr-dict
+  }];
+  let hasVerifier = 1;
+}
 
 #endif // CIRCT_DIALECT_MOORE_MOOREOPS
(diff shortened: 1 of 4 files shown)


include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -2395,5 +2395,22 @@ def ClassDeclOp
   let hasVerifier = 1;
 }
 
+def ClassNewOp
+    : MooreOp<
+          "class.new", [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
+  let summary = "Allocate a new class instance";
+  let description = [{
+    Allocates a new instance of class @C. This op does not call the constructor.
```
TaoBi22, line 2403: I'm guessing from the verifier that this is meant to eventually call the constructor once that's supported? If so we can probably drop the 'This op does not call the constructor.' since there's an unsupported error that gets thrown anyway (and it saves on some documentation that might get missed and end up outdated)

TaoBi22 (approved):
LGTM! :rocket:

include/circt/Dialect/Moore/MooreOps.td
Scheremo: I would keep the call to the constructor an explicit Op, so `new` does not invoke it, since lots of the same class semantic rules that go for methods also apply for the constructor, including implicit upcasts :smile: I would, however, emit a call to the constructor anytime `new` is called in System Verilog.

TaoBi22: Ah cool, sounds sensible!

PR #9135, opened 2025-10-24
[Moore] [3/4] Implement class member access, upcast

Scheremo:
- **New IR ops**
  - `moore.class.property_ref`: yields a `!moore.ref<T>` to a class property.
  - `moore.class.upcast`: converts a derived class handle to one of its base class handles (zero-cost).

- **Class handle utilities**
  - Added `Context::isSameOrDerivedFrom()` and `getAncestorClassWithProperty()` for inheritance and field lookup.
  - Added verifiers for both new ops ensuring valid symbols & types.

- `ImportVerilog/Expressions.cpp`:
  - Added implicit handle upcasting (`maybeUpcastHandle`).
  - Added lowering for `t.a` reads/writes via `moore.class.property_ref`.
  - Emits `moore.class.upcast` when accessing inherited fields.

- Added coverage for:
  - Property read/write (`t.a`, `t.a = x`).
  - Implicit upcast from derived -> base before member access.

### DIFF
--- a/include/circt/Dialect/Moore/MooreOps.td
+++ b/include/circt/Dialect/Moore/MooreOps.td
@@ -2413,4 +2413,35 @@ def ClassNewOp
   let hasVerifier = 1;
 }
 
+
+def ClassPropertyRefOp
+    : MooreOp<"class.property_ref", [Pure,
+    DeclareOpInterfaceMethods<SymbolUserOpInterface>,]> {
+  let summary = "Get a !moore.ref to a class property";
+  let description = [{
+    Construct a reference to a class instance's property field. Translates
+    to a fixed offset from the the class handle into its data struct.
+  }];
+  let arguments = (ins ClassHandleType:$instance,
+      FlatSymbolRefAttr:$property);
+  let results = (outs RefType:$propertyRef);
+  let assemblyFormat = "$instance`[`$property`]` `:` type($instance) `->` "
+                       "type($propertyRef) attr-dict";
+}
+
+def ClassUpcastOp
+    : MooreOp<"class.upcast", [Pure,
+    DeclareOpInterfaceMethods<SymbolUserOpInterface>,]> {
+  let summary = "Upcast a derived handle to a base handle (zero-cost)";
+  let description = [{
+    Casts an instance of a class to one of its base classes.
+    The created instance points to the same underlying data struct as the original
+    instance, and modifying it will modify the spawning instance's data as well.
+  }];
+  let arguments = (ins ClassHandleType:$instance);
+  let results = (outs ClassHandleType:$result);
+  let assemblyFormat =
+      "$instance `:` type($instance) `to` type($result) attr-dict";
+}
+
 #endif // CIRCT_DIALECT_MOORE_MOOREOPS
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -328,37 +328,81 @@ struct ExprVisitor {
   /// Handle member accesses.
   Value visit(const slang::ast::MemberAccessExpression &expr) {
     auto type = context.convertType(*expr.type);
-    auto valueType = expr.value().type;
-    auto value = convertLvalueOrRvalueExpression(expr.value());
-    if (!type || !value)
+    if (!type)
       return {};
 
-    auto resultType =
-        isLvalue ? moore::RefType::get(cast<moore::UnpackedType>(type)) : type;
+    auto *valueType = expr.value().type.get();
     auto memberName = builder.getStringAttr(expr.member.name);
 
     // Handle structs.
     if (valueType->isStruct()) {
+      auto resultType =
+          isLvalue ? moore::RefType::get(cast<moore::UnpackedType>(type))
+                   : type;
+      auto value = convertLvalueOrRvalueExpression(expr.value());
+      if (!value)
+        return {};
+
       if (isLvalue)
         return moore::StructExtractRefOp::create(builder, loc, resultType,
                                                  memberName, value);
-      else
-        return moore::StructExtractOp::create(builder, loc, resultType,
-                                              memberName, value);
+      return moore::StructExtractOp::create(builder, loc, resultType,
+                                            memberName, value);
     }
 
     // Handle unions.
     if (valueType->isPackedUnion() || valueType->isUnpackedUnion()) {
+      auto resultType =
+          isLvalue ? moore::RefType::get(cast<moore::UnpackedType>(type))
+                   : type;
+      auto value = convertLvalueOrRvalueExpression(expr.value());
+      if (!value)
+        return {};
+
       if (isLvalue)
         return moore::UnionExtractRefOp::create(builder, loc, resultType,
                                                 memberName, value);
-      else
-        return moore::UnionExtractOp::create(builder, loc, type, memberName,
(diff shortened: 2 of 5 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -2243,3 +2324,70 @@ Context::convertSystemCallArity1(const slang::ast::SystemSubroutine &subroutine,
           .Default([&]() -> Value { return {}; });
   return systemCallRes();
 }
+
+// Resolve any (possibly nested) SymbolRefAttr to an op from the root.
+static mlir::Operation *resolve(Context &context, mlir::SymbolRefAttr sym) {
+  return mlir::SymbolTable::lookupNearestSymbolFrom(context.intoModuleOp, sym);
+}
```
fabianschuiki, line 2331: Can you pass in a concrete `SymbolTable &` reference here? The call to `lookupNearestSymbolFrom` will re-scan the entire IR to build up the symbol table, which turns into O(N²) behavior very quickly 😞

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -1417,6 +1417,110 @@ void ClassNewOp::getEffects(
...
+  auto *op = getOperation();
+
+  auto *srcDeclOp =
+      mlir::SymbolTable::lookupNearestSymbolFrom(op, srcTy.getClassSym());
+  auto *dstDeclOp =
+      mlir::SymbolTable::lookupNearestSymbolFrom(op, dstTy.getClassSym());
+  if (!srcDeclOp || !dstDeclOp)
+    return emitOpError() << "failed to resolve class symbol(s): src="
+                         << srcTy.getClassSym()
+                         << ", dst=" << dstTy.getClassSym();
```
fabianschuiki, line 1445: These have O(N²) scaling in practice unfortunately. I think you can add the `SymbolUserOpInterface` to the `ClassUpcastOp`. That interface defines an additional verification method that gets a `SymbolTable` reference passed in which you can use to resolve symbols. The instance ops in the Moore dialect probably already use that 🙂

lib/Dialect/Moore/MooreOps.cpp
```diff
@@ -1417,6 +1417,110 @@ void ClassNewOp::getEffects(
...
+  SymbolRefAttr classSym = handleTy.getClassSym();
+  if (!classSym)
+    return emitOpError("instance type is missing a class symbol");
+
+  // Resolve the class symbol starting from the nearest symbol table.
+  Operation *clsSym =
+      mlir::SymbolTable::lookupNearestSymbolFrom(getOperation(), classSym);
+  if (!clsSym)
+    return emitOpError("referenced class symbol `")
+           << classSym << "` was not found";
```
fabianschuiki, line 1490: The `ClassPropertyRefOp` probably also needs to be a symbol user such that you can verify that the referenced field actually exists in the class.

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -2413,4 +2413,33 @@ def ClassNewOp
   let hasVerifier = 1;
 }
 
+
+def ClassPropertyRefOp : MooreOp<"class.property_ref", [Pure]> {
+  let summary = "Get a !moore.ref to a class property";
+  let description = [{
+  Construct a reference to a class instance's property field. Translates
+  to a fixed offset from the the class handle into its data struct.
+  }];
```
fabianschuiki, line 2424: Nit: indentation
```suggestion
  let description = [{
    Construct a reference to a class instance's property field. Translates
    to a fixed offset from the the class handle into its data struct.
  }];
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -2413,4 +2413,33 @@ def ClassNewOp
...
+  let hasVerifier = 1;
+}
+
+def ClassUpcastOp : MooreOp<"class.upcast", [Pure]> {
+  let summary = "Upcast a derived handle to a base handle (zero-cost)";
+  let description = [{
+  Casts an instance of a class to one of its base classes.
+  The created instance points to the same underlying data struct as the original instance,
+  and modifying it will modify the spawning instance's data as well.
+  }];
```
fabianschuiki, line 2440: Nit: indentation
```suggestion
  let description = [{
    Casts an instance of a class to one of its base classes.
    The created instance points to the same underlying data struct as the original instance,
    and modifying it will modify the spawning instance's data as well.
  }];
```

include/circt/Dialect/Moore/MooreOps.td
```diff
@@ -2413,4 +2413,33 @@ def ClassNewOp
...
+  let arguments = (ins ClassHandleType:$instance);
+  let results = (outs ClassHandleType:$result);
+  let assemblyFormat =
+      "$instance `:` type($instance) `to` type($result) attr-dict";
```
fabianschuiki, line 2444: I'm wondering if it would make sense to constrain the upcast op to only perform a single layer of upcasting; basically a class to one of its immediate base classes. This would potentially simplify lowering of the op, since you can just perform a single projection on the class pointer and the vtable pointer. The frontend could then insert multiple upcast ops if needed. WDYT?

fabianschuiki (approved):
LGTM, modulo the O(N²) behavior thanks to MLIR's `SymbolTable`s.

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: I switched to using `context.symbolTable`, is there a narrower one to use?

lib/Dialect/Moore/MooreOps.cpp
Scheremo: Oh lord 🙈 Thanks for pointing this out! Refactored using `SymbolUserOpInterface`.

lib/Dialect/Moore/MooreOps.cpp
Scheremo: Yep. Refactored using `SymbolUserOpInterface`.

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: That should be perfect!

include/circt/Dialect/Moore/MooreOps.td
Scheremo: So my original thought was to have "multi-step" upcasts for a few reasons:
1) I think it looks neater 😄 
2) System Verilog only allows single-inheritance from base classes, (but multiple "inheritance" for interfaces - however all of their fields must be virtual and there is no shadowing, so vtable resolution shouldn't produce the diamond inheritance problem) so the upcast path is unique, there's no diamond inheritance problem. 
3) My idea was to translate upcasts into literal no-ops; we practically only need the instance's RTTI to decide on the vtable pointer for dispatch, and since we can rely on objects only having single inheritance, the class pointer upcast is straight-forward. We can just construct the object layout to append properties to its data structure for each level of inheritance.

Maybe I'm missing something but I believe multi-step upcasting is safe. I think your solution works equally well, but since we don't have multiple inheritance we should be able to skip it; thoughts?

fabianschuiki: Makes sense! I think the approach you have in this PR is great!

PR #9136, opened 2025-10-24
[Moore] [4/4] Implement direct class method calls

Scheremo:
Adds **class method declarations**, wires it into the importer, and supports implicit `this`, inherited property access, and upcasting.

- **New ops**
  - `moore.class.methoddecl`: declare methods inside `classdecl` (FunctionType).

- **Importer**
  - Lowers `t.m(...)` and unqualified `m(...)` (inside methods) to `func.call`.
  - Adds implicit `%this : !moore.class<@C>` as first block arg for methods.
  - Performs **implicit upcast** of the receiver when needed.
  - Resolves class properties via `class.property_ref` (now also for implicit `this`).
  - Clearer diag: prints SV type for unsupported member-access bases.

- **Lowering plumbing**
  - `declareFunction` split into `declareCallableImpl` to share logic; supports methods (qualified name `@"Pkg::Class"::method`).
  - Tracks `currentThisRef` in `Context` for implicit `this` in expressions.
  - `ClassDeclVisitor` emits `class.methoddecl` in source order after lowering the method body.

- **Verifiers / build**
  - `ClassDeclOp` verifier accepts `class.methoddecl`.
  - Includes/linkage for call interfaces (`MooreOps.h`, CMake).

- **Tests**
  - `classes.sv`: calls on concrete classes, inherited property access through upcast, and method decls reflected in IR.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -438,6 +438,37 @@ struct RvalueExprVisitor : public ExprVisitor {
       return value;
     }
 
+    // We're reading a class property.
+    if (auto *const property =
+            expr.symbol.as_if<slang::ast::ClassPropertySymbol>()) {
+      auto type = context.convertType(*expr.type);
+
+      // Get the scope's implicit this variable
+      mlir::Value instRef = context.getImplicitThisRef();
+      if (!instRef) {
+        mlir::emitError(loc) << "class property '" << property->name
+                             << "' referenced without an implicit 'this'";
+        return {};
+      }
+
+      auto fieldSym =
+          mlir::FlatSymbolRefAttr::get(builder.getContext(), property->name);
+      auto fieldTy = cast<moore::UnpackedType>(type);
+      auto fieldRefTy = moore::RefType::get(fieldTy);
+
+      moore::ClassHandleType classTy =
+          cast<moore::ClassHandleType>(instRef.getType());
+
+      auto targetClassHandle =
+          context.getAncestorClassWithProperty(classTy, property->name);
+      auto upcastRef = context.materializeConversion(targetClassHandle, instRef,
+                                                     false, instRef.getLoc());
+
+      Value fieldRef = moore::ClassPropertyRefOp::create(
+          builder, loc, fieldRefTy, upcastRef, fieldSym);
+      return moore::ReadOp::create(builder, loc, fieldRef).getResult();
+    }
+
     // Try to materialize constant values directly.
     auto constant = context.evaluateConstant(expr);
     if (auto value = context.materializeConstant(constant, *expr.type, loc))
@@ -1155,12 +1186,6 @@ struct RvalueExprVisitor : public ExprVisitor {
 
   /// Handle calls.
   Value visit(const slang::ast::CallExpression &expr) {
-    // Class method calls are currently not supported.
-    if (expr.thisClass()) {
-      mlir::emitError(loc, "unsupported class method call");
-      return {};
-    }
-
     // Try to materialize constant values directly.
     auto constant = context.evaluateConstant(expr);
     if (auto value = context.materializeConstant(constant, *expr.type, loc))
@@ -1171,9 +1196,77 @@ struct RvalueExprVisitor : public ExprVisitor {
         expr.subroutine);
   }
 
+  /// Get both the actual `this` argument of a method call and the required
+  /// class type.
+  std::pair<Value, moore::ClassHandleType>
+  getMethodReceiverTypeHandle(const slang::ast::CallExpression &expr) {
+
+    moore::ClassHandleType handleTy;
+    Value thisRef;
+
+    // Qualified call: t.m(...), extract from thisClass.
+    if (const slang::ast::Expression *recvExpr = expr.thisClass()) {
+      thisRef = context.convertRvalueExpression(*recvExpr);
+      if (!thisRef)
+        return {};
+    } else {
+      // Unqualified call inside a method body: try using implicit %this.
+      thisRef = context.getImplicitThisRef();
+      if (!thisRef) {
+        mlir::emitError(loc) << "method '" << expr.getSubroutineName()
+                             << "' called without an object";
+        return {};
+      }
+    }
+    handleTy = cast<moore::ClassHandleType>(thisRef.getType());
+    return {thisRef, handleTy};
+  }
+
+  /// Build a method call including implicit this argument.
+  mlir::func::CallOp
+  buildMethodCall(const slang::ast::SubroutineSymbol *subroutine,
+                  FunctionLowering *lowering,
+                  moore::ClassHandleType actualHandleTy, Value actualThisRef,
+                  SmallVector<Value> &arguments,
+                  SmallVector<Type> &resultTypes) {
+
+    // Get the expected receiver type from the lowered method
+    auto funcTy = lowering->op.getFunctionType();
+    auto expected0 = funcTy.getInput(0);
+    auto expectedHdlTy = cast<moore::ClassHandleType>(expected0);
+
+    // Upcast the handle as necessary.
+    auto implicitThisRef = context.materializeConversion(
+        expectedHdlTy, actualThisRef, false, actualThisRef.getLoc());
+
+    // Build an argument list where the this reference is the first argument.
+    SmallVector<Value> explicitArguments;
+    explicitArguments.reserve(arguments.size() + 1);
+    explicitArguments.push_back(implicitThisRef);
+    explicitArguments.append(arguments.begin(), arguments.end());
+
+    // Method call: choose direct vs virtual.
+    const bool isVirtual =
+        (subroutine->flags & slang::ast::MethodFlags::Virtual) != 0;
+
+    if (!isVirtual) {
+      // Direct (non-virtual) call -> moore.class.call
+      auto calleeSym =
+          SymbolRefAttr::get(context.getContext(), lowering->op.getSymName());
+      return mlir::func::CallOp::create(builder, loc, resultTypes, calleeSym,
+                                        explicitArguments);
+    }
+
+    mlir::emitError(loc) << "virtual method calls not supported";
+    return {};
+  }
+
   /// Handle subroutine calls.
   Value visitCall(const slang::ast::CallExpression &expr,
                   const slang::ast::SubroutineSymbol *subroutine) {
+
+    const bool isMethod = (subroutine->thisVar != nullptr);
+
     auto *lowering = context.declareFunction(*subroutine);
     if (!lowering)
       return {};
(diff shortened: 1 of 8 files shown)


lib/Conversion/ImportVerilog/ImportVerilogInternals.h
```diff
@@ -294,9 +307,19 @@ struct Context {
...
+                      mlir::StringRef qualifiedName,
+                      llvm::SmallVectorImpl<Type> &extraParams);
+  /// Variable to track the value of the current functions implicit this
+  /// reference
+  Value currentThisRef = {};
 };
 
 } // namespace ImportVerilog
 } // namespace circt
-
```
tobiasgrosser: This seems unrelated.

Scheremo: It was! Thanks for pointing it out.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+    auto funcTy = lowering->op.getFunctionType();
+    auto expected0 = funcTy.getInput(0);
+    auto expectedHdlTy = cast<moore::ClassHandleType>(expected0);
+
+    // Upcast the handle as necessary.
+    mlir::Value implicitThisRef = context.materializeConversion(
+        expectedHdlTy, actualThisRef, false, actualThisRef.getLoc());
+
+    // Build an argument list where the this reference is the first argument.
+    SmallVector<Value> explicitArguments;
```
fabianschuiki, line 1243: Nit: pre-allocate the needed storage:

```suggestion
    SmallVector<Value> explicitArguments;
    explicitArguments.reserve(arguments.size() + 1);
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+    return {thisRef, handleTy};
+  }
+
+  /// Build either a class.vcall or class.call for a method call
+  mlir::Value buildMethodCall(const slang::ast::SubroutineSymbol *subroutine,
+                              FunctionLowering *lowering,
+                              moore::ClassHandleType actualHandleTy,
+                              mlir::Value actualThisRef,
+                              SmallVector<Value> &arguments,
+                              //! This does NOT include the implicit handle!
```
fabianschuiki:
```suggestion
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+      }
+    }
+    return {thisRef, handleTy};
+  }
+
+  /// Build either a class.vcall or class.call for a method call
+  mlir::Value buildMethodCall(const slang::ast::SubroutineSymbol *subroutine,
+                              FunctionLowering *lowering,
+                              moore::ClassHandleType actualHandleTy,
+                              mlir::Value actualThisRef,
```
fabianschuiki:
```suggestion
                              Value actualThisRef,
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+                              //! This does NOT include the implicit handle!
+                              SmallVector<Type> &resultTypes) {
+
+    // Get the expected receiver type from the lowered method
+    auto funcTy = lowering->op.getFunctionType();
+    auto expected0 = funcTy.getInput(0);
+    auto expectedHdlTy = cast<moore::ClassHandleType>(expected0);
+
+    // Upcast the handle as necessary.
+    mlir::Value implicitThisRef = context.materializeConversion(
```
fabianschuiki:
```suggestion
    auto implicitThisRef = context.materializeConversion(
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+    explicitArguments.push_back(implicitThisRef);
+    explicitArguments.append(arguments.begin(), arguments.end());
+
+    // Method call: choose direct vs virtual.
+    const bool isVirtual =
+        (subroutine->flags & slang::ast::MethodFlags::Virtual) != 0;
+
+    if (!isVirtual) {
+      // Direct (non-virtual) call -> moore.class.call
+      auto calleeSym = mlir::SymbolRefAttr::get(context.getContext(),
```
fabianschuiki:
```suggestion
      auto calleeSym = SymbolRefAttr::get(context.getContext(),
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+                                                lowering->op.getSymName());
+      auto callOp = mlir::func::CallOp::create(builder, loc, resultTypes,
+                                               calleeSym, explicitArguments);
+      return resultTypes.size() > 0 ? callOp.getResult(0) : Value{};
```
fabianschuiki: You might want to make this function return `callOp`, such that you can distinguish between `return {}` in case of an error and a call that does not return any result.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+    }
+
+    mlir::emitError(loc) << "Virtual methods are not yet supported!";
```
fabianschuiki: Nit: the frontend tries to stick with lowercase error messages:
```suggestion
    mlir::emitError(loc) << "virtual method calls not supported";
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+      if (!handleTy) {
+        mlir::emitError(loc)
+            << "implicit 'this' must be class<...>, got " << thisRef.getType();
+        return {};
+      }
+    }
+    return {thisRef, handleTy};
+  }
+
+  /// Build either a class.vcall or class.call for a method call
```
fabianschuiki: Outdated comment?

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1171,9 +1196,91 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+        mlir::emitError(loc) << "method '" << expr.getSubroutineName()
+                             << "' called without an object";
+        return {};
+      }
+      handleTy = dyn_cast<moore::ClassHandleType>(thisRef.getType());
+      if (!handleTy) {
+        mlir::emitError(loc)
+            << "implicit 'this' must be class<...>, got " << thisRef.getType();
+        return {};
+      }
```
fabianschuiki: Can you add error tests cases that trigger each of these errors? I suspect that most of these cannot be triggered because Slang prevents such an AST from being formed, in which case I'd turn these into asserts.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1254,20 +1361,38 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+        lowering->op.getFunctionType().getResults().end());
+
+    mlir::Value result;
+    if (isMethod) {
+      // Class functions -> moore.class.call or moore.class.vcall
+      mlir::Value thisRef;
+      moore::ClassHandleType tyHandle;
+      auto retP = getMethodReceiverTypeHandle(expr);
+      thisRef = retP.first;
+      tyHandle = retP.second;
```
fabianschuiki:
```suggestion
      auto [thisRef, tyHandle] = getMethodReceiverTypeHandle(expr);
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1254,20 +1361,38 @@ struct RvalueExprVisitor : public ExprVisitor {
...
 
-    // Create the call.
-    auto callOp =
-        mlir::func::CallOp::create(builder, loc, lowering->op, arguments);
+    // Determine result types from the declared/converted func op.
+    SmallVector<Type> resultTypes(
+        lowering->op.getFunctionType().getResults().begin(),
+        lowering->op.getFunctionType().getResults().end());
+
+    mlir::Value result;
```
fabianschuiki: Since all code paths produce a `CallOp`, you could pull `CallOp callOp` up here instead of `result` and use the old logic of returning an unrealized conversion cast to build a void typed value.

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -988,7 +987,65 @@ Context::declareFunction(const slang::ast::SubroutineSymbol &subroutine) {
...
+    auto &ownerLowering = classes[classTy];
+    ownerDecl = ownerLowering->op;
+  } else {
+    mlir::emitError(loc) << "expected 'this' to be a class type, got "
+                         << thisTy.toString();
+    return {};
+  }
+
+  if (subroutine.flags.has(slang::ast::MethodFlags::Virtual)) {
+    mlir::emitError(loc) << "Virtual methods are not yet supported!";
```
fabianschuiki, line 1017: Nit:
```suggestion
    mlir::emitError(loc) << "virtual methods not supported";
```

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -1031,23 +1083,19 @@ Context::declareFunction(const slang::ast::SubroutineSymbol &subroutine) {
 
   auto funcType = FunctionType::get(getContext(), inputTypes, outputTypes);
 
-  // Prefix the function name with the surrounding namespace to create somewhat
-  // sane names in the IR.
-  SmallString<64> funcName;
-  guessNamespacePrefix(subroutine.getParentScope()->asSymbol(), funcName);
-  funcName += subroutine.name;
-
```
fabianschuiki, line 1039: Just out of curiosity: does the namespace prefix guessing not work if there are classes or modules around the function/subroutine?

lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -1131,11 +1179,27 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
...
+  // If this is a class method, the first input is %this :
+  // !moore.class<@C>
+  unsigned inputIdx = 0;
+  if (isMethod) {
+    auto thisLoc = convertLocation(subroutine.location);
+    auto thisType = lowering->op.getFunctionType().getInput(inputIdx++);
+    auto thisArg = block.addArgument(thisType, thisLoc);
+
+    // Bind `this` so NamedValue/MemberAccess can find it.
+    valueSymbols.insert(subroutine.thisVar, thisArg);
```
fabianschuiki, line 1199: Nice 😎

lib/Conversion/ImportVerilog/Structure.cpp
Scheremo: Yeah, it seems to not resolve package scope names or module scope names.

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: I'd expect the same thing. Maybe we can simplify this a bit.

Scheremo: I wasn't able to trigger the error for `dyn_cast`s to ClassHandleType going wrong; so I just made those `cast`s instead. 
The error on `getImplicitThisRef` wouldn't be a Slang error, that'd be a me error for wrong this reference tracking so I'd keep it. I again wasn't able to trigger this through valid system verilog, but you never know 🙂

lib/Conversion/ImportVerilog/Structure.cpp
fabianschuiki: Makes sense!

fabianschuiki (approved):
LGTM! Thanks for adding  this 🙌

PR #9141, opened 2025-10-27
[Synth] Refactor and expose dropNonCriticalPaths to Python bindings

uenoku:
Refactored LongestPathCollection::sortAndDropNonCriticalPathsPerEndPoint to dropNonCriticalPaths with a parameter to filter by end point or start point. Separated sorting from dropping for better control. Exposed the functionality through C API and Python bindings.


uenoku:
I'm going to merge to include in the next circt release. Post-commit review is appreciated.

PR #9146, opened 2025-10-28
[FIRRTL] Remove dead annotations relating to taps and blackboxes.

dtzSiFive:
Most of these were internal-only and removed with previous changes years ago.


seldridge (approved):
LGTM

Are these dead internally?

Anything to update in docs about this?

dtzSiFive:
I need to triple-check the MemTap-based annotations aren't reachable internally, thanks folks!

DataTaps will be removed in follow-on PR, after also removing them internally :+1:.  Docs removed there, the annotations removed here were not documented or tested in CIRCT (at least not anymore/FWIW).

dtzSiFive:
I believe we're good!

source/sink/port are members of the MemTap annotation but not used as the `class` -- that's something that we did within CIRCT: https://github.com/llvm/circt/commit/539e148b1d0ae0c00cff81d0dbf655c8ad102daa .
That is to say -- these annotation classes were never from the input.

🏅 for the wonderful explanatory commit message, made the source code archaeology go a lot easier!

The Deleted/Literal annotation classes were removed internally Feb 2023.

PR #9149, opened 2025-10-28
[FIRRTL] Remove GC DataTaps, unused upstream.

dtzSiFive:
Probes (+BoringUtils) replace this.


seldridge (approved):
LGTM as long as this is gone internally. 💯

dtzSiFive:
Thanks for the review!!

The data taps support I expect to remove from our internal code tonight so
I'm feeling very good about this being safe once that lands.

Still need to do a bit of due diligence about some memtap annotation
derivatives in the morning though -- this PR should be good regardless just
depends what it's based on 👍.

Follow-on PR removes internal paths from IR and passes, so there's a little
bit more cleanup to come!  Including dropping ref statement support, FWIW.
I felt previously that as long as we supported that feature via
(undocumented?) annotations, supporting via FIRRTL text even if we didn't
use it, was a small cost.  That is to say, deleting it helped very little
and weakly arguably was a little nicer (and part of 4.0.0 even if we went
different way thereafter).

However, after this PR that'll be the only user of the internal paths bits
and so plan to remove them.

Woohoo!

PR #9151, opened 2025-10-29
[MooreToCore] [2/4] Add support to lower ClassNewOp to llvm.malloc

Scheremo:
- Add `getOrCreateMalloc` to declare `llvm.func @malloc(i64) -> !llvm.ptr` (opaque ptr).
- Implement `ClassNewOpConversion`:
  - Resolve class struct body, compute size via `DataLayout`, call `malloc`.
  - Replace `moore.class.new` with result of `malloc` (no cast in opaque-ptr mode).
- Make base-class struct resolution lazy:
  - If a base’s struct isn’t finalized, resolve it on demand.
  - Add overload `resolveClassStructBody(ModuleOp, SymbolRefAttr, ...)`.
- Register new conversion pattern in `populateOpConversion`.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -60,7 +60,7 @@ struct ClassTypeCache {
     /// Record/overwrite the field path to a single property for a class.
     void setFieldPath(StringRef propertyName, ArrayRef<unsigned> path) {
       this->propertyPath[propertyName] =
-          llvm::SmallVector<unsigned, 2>(path.begin(), path.end());
+          SmallVector<unsigned, 2>(path.begin(), path.end());
     }
 
     /// Lookup the full GEP path for a (class, field).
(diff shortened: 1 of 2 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -120,21 +138,30 @@ static LogicalResult resolveClassStructBody(ClassDeclOp op,
...
-    // Inherit base field paths with a leading 0.
-    for (auto &kv : baseClassStruct->propertyPath) {
-      SmallVector<unsigned, 2> path;
-      path.push_back(0); // into base subobject
-      path.append(kv.second.begin(), kv.second.end());
-      structBody.setFieldPath(kv.first, path);
+    ModuleOp mod = op->getParentOfType<ModuleOp>();
+    // It's possible the base class does not resolve since it materialzes
+    // no methods or properties. In that case we can simply ignore inheritance.
+    if (auto *opSym = mod.lookupSymbol(baseClass)) {
```
fabianschuiki: Shouldn't this be an error? If the class mentions a base class symbol, we should enforce that symbol to exist. The symbol user verifier should already do this. That means you can just `assert(opSym)` here :smiley:

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -120,21 +138,30 @@ static LogicalResult resolveClassStructBody(ClassDeclOp op,
...
+      auto classDeclOp = dyn_cast<ClassDeclOp>(opSym);
+      if (!classDeclOp)
+        return op.emitError() << "Found symbol corresponding to " << baseClass
+                              << " but it wasn't a classdeclop!";
```
fabianschuiki: This is something we should do in the symbol user op verifier of the ClassDeclOp, and then simply use a `cast<...>` here. That way we get the verifier to enforce invariants which we then can assume/assert to hold here.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -678,6 +713,52 @@ static Value createZeroValue(Type type, Location loc,
...
+    auto sym = handleTy.getClassSym();
+
+    ModuleOp mod = op->getParentOfType<ModuleOp>();
+
+    if (failed(resolveClassStructBody(mod, sym, *typeConverter, cache)))
+      return op.emitError() << "Could not resolve class struct for " << sym;
+
+    auto structTy = cache.getStructInfo(sym)->classBody;
+
+    DataLayout dl(mod);
```
fabianschuiki, line 732: Thanks for thinking of the data layout and different hosts :1st_place_medal:

fabianschuiki (approved):
LGTM! Just a minor nit on being too defensive with symbol resolution -- not sure if that is part of a previous PR or this one here.

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Fair enough, we can always be more defensive if needed.

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Done!

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: 🫡

PR #9152, opened 2025-10-29
[MooreToCore] [3/4] Add support for lowering moore.class.upcast

Scheremo:
- Added `ClassUpcastOpConversion` lowering:
  - Converts `moore.class.upcast` to a direct value forward when both source and
    destination are LLVM opaque pointers (`!llvm.ptr`).
  - Emits a diagnostic if applied to non-opaque pointer types.

- Registered the new pattern in `populateOpConversion` alongside other class
  conversion patterns.

### DIFF
--- a/test/Conversion/MooreToCore/classes.mlir
+++ b/test/Conversion/MooreToCore/classes.mlir
@@ -80,3 +80,23 @@ moore.class.classdecl @E extends @C {
   moore.class.propertydecl @b : !moore.l32
   moore.class.propertydecl @c : !moore.l32
 }
+
+/// Check that upcast lowers to no-op
+
+// CHECK-LABEL: func.func private @test_new5
+// CHECK-SAME: (%arg0: !llvm.ptr) -> !llvm.ptr {
+// CHECK:   return %arg0 : !llvm.ptr
+
+// CHECK-NOT: moore.class.new
+// CHECK-NOT: moore.class.upcast
+// CHECK-NOT: moore.class.classdecl
+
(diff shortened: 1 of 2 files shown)


test/Conversion/MooreToCore/classes.mlir
```diff
@@ -80,3 +80,25 @@ moore.class.classdecl @E extends @C {
...
+
+// CHECK-NOT: moore.class.new
+// CHECK-NOT: moore.class.upcast
+// CHECK-NOT: moore.class.classdecl
+
+func.func private @test_new5() {
+  %h = moore.class.new : <@F>
+  %upcast = moore.class.upcast %h : <@F> to <@C>
+  return
+}
```
fabianschuiki, line 97: How about passing class `%h` in as a func arg, such that the test doesn't cover `moore.class.new`, but only `moore.class.upcast`? Something like:
```mlir
// CHECK-LABEL: func.func private @test_new5
// CHECK-SAME: (%arg0: !llvm.ptr) -> !llvm.ptr
// CHECK-NEXT:   return %arg0 : !llvm.ptr

// CHECK-NOT: moore.class.upcast
// CHECK-NOT: moore.class.classdecl

func.func private @test_new5(%arg0: !moore.class<@F>) -> !moore.class<@C> {
  %0 = moore.class.upcast %arg0 : <@F> to <@C>
  return %0 : !moore.class<@C>
}
```

Scheremo: Sure, done!

PR #9153, opened 2025-10-29
[MooreToCore] [4/4] Add support for class.property_ref

Scheremo:
Add a conversion pattern to translate `moore.class.property_ref` into an LLVM `gep` over the identified class struct, bridging the resulting pointer back to `!llhd.ref<…>` via `unrealized_conversion_cast`.

- Implement `ClassPropertyRefOpConversion`:
  - Validates result lowers to `!llhd.ref<T>`.
  - Ensures instance is `!llvm.ptr`.
  - Resolves the class’s identified struct and per-field GEP path via `ClassTypeCache`.
  - Emits `llvm.getelementptr` with cached indices and wraps back to ref.
- Register the pattern in `populateOpConversion`.
- Extend `classes.mlir` to verify that `property_ref` lowers to GEP.

### DIFF
--- a/lib/Conversion/MooreToCore/MooreToCore.cpp
+++ b/lib/Conversion/MooreToCore/MooreToCore.cpp
@@ -707,6 +707,68 @@ static Value createZeroValue(Type type, Location loc,
   return rewriter.createOrFold<hw::BitcastOp>(loc, type, constZero);
 }
 
+struct ClassPropertyRefOpConversion
+    : public OpConversionPattern<circt::moore::ClassPropertyRefOp> {
+  ClassPropertyRefOpConversion(TypeConverter &tc, MLIRContext *ctx,
+                               ClassTypeCache &cache)
+      : OpConversionPattern(tc, ctx), cache(cache) {}
+
+  LogicalResult
+  matchAndRewrite(circt::moore::ClassPropertyRefOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    MLIRContext *ctx = rewriter.getContext();
+
+    // Convert result type; we expect !llhd.ref<someT>.
+    Type dstTy = getTypeConverter()->convertType(op.getPropertyRef().getType());
+    // Operand is a !llvm.ptr
+    Value instRef = adaptor.getInstance();
+
+    // Resolve identified struct from cache.
+    auto classRefTy =
+        cast<circt::moore::ClassHandleType>(op.getInstance().getType());
+    SymbolRefAttr classSym = classRefTy.getClassSym();
+    ModuleOp mod = op->getParentOfType<ModuleOp>();
+    if (failed(resolveClassStructBody(mod, classSym, *typeConverter, cache)))
+      return rewriter.notifyMatchFailure(op,
+                                         "Could not resolve class struct for " +
+                                             classSym.getRootReference().str());
+
+    auto structInfo = cache.getStructInfo(classSym);
+    assert(structInfo && "class struct info must exist");
+    auto structTy = structInfo->classBody;
+
+    // Look up cached GEP path for the property.
+    auto propSym = op.getProperty();
+    auto pathOpt = structInfo->getFieldPath(propSym);
+    if (!pathOpt)
+      return rewriter.notifyMatchFailure(op,
+                                         "no GEP path for property " + propSym);
+
+    auto i32Ty = IntegerType::get(ctx, 32);
+    SmallVector<Value> idxVals;
+    for (unsigned idx : *pathOpt)
+      idxVals.push_back(LLVM::ConstantOp::create(
+          rewriter, loc, i32Ty, rewriter.getI32IntegerAttr(idx)));
+
+    // GEP to the field (opaque ptr mode requires element type).
+    auto ptrTy = LLVM::LLVMPointerType::get(ctx);
+    auto gep =
+        LLVM::GEPOp::create(rewriter, loc, ptrTy, structTy, instRef, idxVals);
+
+    // Wrap pointer back to !llhd.ref<someT>.
+    Value fieldRef = UnrealizedConversionCastOp::create(rewriter, loc, dstTy,
+                                                        gep.getResult())
+                         .getResult(0);
+
+    rewriter.replaceOp(op, fieldRef);
+    return success();
+  }
+
+private:
+  ClassTypeCache &cache;
+};
+
 struct ClassUpcastOpConversion : public OpConversionPattern<ClassUpcastOp> {
   using OpConversionPattern::OpConversionPattern;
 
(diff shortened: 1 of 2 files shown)


lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -713,6 +713,73 @@ static Value createZeroValue(Type type, Location loc,
...
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    MLIRContext *ctx = rewriter.getContext();
+
+    // Convert result type; we expect !llhd.ref<someT>.
+    Type dstTy = getTypeConverter()->convertType(op.getPropertyRef().getType());
+    auto dstRefTy = dyn_cast<llhd::RefType>(dstTy);
+    if (!dstRefTy)
+      return rewriter.notifyMatchFailure(op,
+                                         "result must lower to !llhd.ref<T>");
```
fabianschuiki: I think this is enforced by the op verifier and op constraints. You also don't use `dstRefTy`, so you should be fine by just keeping the `Type dstTy = ...` line.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -713,6 +713,73 @@ static Value createZeroValue(Type type, Location loc,
...
+    // Operand is a !llvm.ptr
+    Value instRef = adaptor.getInstance();
+    auto instRefTy = dyn_cast<LLVM::LLVMPointerType>(instRef.getType());
+    if (!instRefTy)
+      return rewriter.notifyMatchFailure(op, "instance must be !llvm.ptr");
```
fabianschuiki: This should also be covered by the op verifier. You can probably drop `instRefTy`.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -713,6 +713,73 @@ static Value createZeroValue(Type type, Location loc,
...
+
+    // Resolve identified struct from cache.
+    auto classRefTy =
+        cast<circt::moore::ClassHandleType>(op.getInstance().getType());
+    SymbolRefAttr classSym = classRefTy.getClassSym();
+    ModuleOp mod = op->getParentOfType<ModuleOp>();
+    if (failed(resolveClassStructBody(mod, classSym, *typeConverter, cache)))
+      return op.emitError()
+             << "Could not resolve class struct for " << classSym;
```
fabianschuiki: You probably need a `return rewriter.notifyMatchFailure(...);` here.

lib/Conversion/MooreToCore/MooreToCore.cpp
```diff
@@ -713,6 +713,73 @@ static Value createZeroValue(Type type, Location loc,
...
+
+    auto structInfo = cache.getStructInfo(classSym);
+    assert(structInfo && "class struct info must exist");
+    auto structTy = structInfo->classBody;
+
+    // Look up cached GEP path for the property.
+    auto propSym = op.getProperty();
+    auto pathOpt = structInfo->getFieldPath(propSym);
+    if (!pathOpt)
+      return op.emitError() << "no GEP path for property " << propSym;
```
fabianschuiki: You probably need a `return rewriter.notifyMatchFailure(...);` here, too.

test/Conversion/MooreToCore/classes.mlir
```diff
@@ -102,3 +102,29 @@ moore.class.classdecl @F extends @C {
...
+// CHECK-NOT: moore.class.new
+// CHECK-NOT: moore.class.upcast
+// CHECK-NOT: moore.class.property_ref
+// CHECK-NOT: moore.class.classdecl
+
+func.func private @test_new6() {
+  %h = moore.class.new : <@G>
+  %gep = moore.class.property_ref %h[@d] : <@G> -> <i32>
+  return
+}
```
fabianschuiki, line 119: Same as in the previous PR: maybe make the function accept and return a class, such that you don't need to test `moore.class.new` again:
```mlir
// CHECK-LABEL: func.func private @test_new6
// CHECK-SAME: (%arg0: !llhd.ptr) -> !llhd.ref<i32>
// CHECK:   [[CONSTIDX:%.+]] = llvm.mlir.constant(1 : i32) : i32
// CHECK:   [[GEP:%.+]] = llvm.getelementptr %arg0[[[CONSTIDX]]] : (!llvm.ptr, i32) -> !llvm.ptr, !llvm.struct<"G", (struct<"C", (i32, i32, i32)>, i32, i32, i32)>
// CHECK:   [[CONV:%.+]] = builtin.unrealized_conversion_cast [[GEP]] : !llvm.ptr to !llhd.ref<i32>
// CHECK:   return [[CONV]]

// CHECK-NOT: moore.class.property_ref
// CHECK-NOT: moore.class.classdecl

func.func private @test_new6(%arg0: !moore.class<@G>) -> !moore.ref<i32> {
  %0 = moore.class.property_ref %arg0[@d] : <@G> -> <i32>
  return %0 : !moore.ref<i32>
}
```

Scheremo:
> LGTM! We'll need to think about how we are going to deal with that `!llvm.ptr` vs `!llhd.ref<T>` type mismatch. The `!llhd.ref<T>` expects to point to a signal that can then be used with `llhd.prb` and `llhd.drv`. Since classes are just bits of memory on the heap, that doesn't really work. Not sure how to reconcile this. Maybe we need a special `!moore.ref` type that is a pointer and can't be used with delayed and non-blocking assigns, which we could lower to `!llvm.ptr` 🤔

I think the problem is mainly difficult for LValue accesses. I think we can add another operator for RValues in Moore (property_value?) to just return `!T` rather than `!llhd.ref<T>`. 
On the other hand, what's the semantic issue with probing and driving on the heap?

test/Conversion/MooreToCore/classes.mlir
Scheremo: Done!

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Fair enough. Changed!

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Sure, check is gone

lib/Conversion/MooreToCore/MooreToCore.cpp
Scheremo: Yep, fair point. Changed it.

fabianschuiki:
I think the problem is that `llhd.prb` and `llhd.drv` expect to be interacting with an event queue. When the `llhd.drv` has a delay on it, it wants to have the `!llhd.ref<T>` represent something that has a mechanism to deal with the delay. When we lower to Arcilator for example, we may want to turn those into an `!arc.event_queue<T>`, with ops to push/pop entries into the queue. But classes only point at memory and want load and stores, so lowering that to `!arc.event_queue<T>` doesn't make a lot of sense. And we can't just make those `!llvm.ptr`, because that `!moore.ref` may be travelling through many layers of module hierarchy, each of which cannot "see" whether the ref was a signal where event queue mechanisms are allowed, or whether it was a class where they are disallowed.

PR #9155, opened 2025-10-29
[HWToBTOR2] Add support for (most) variadic ops

TaoBi22:
This adds support for the variadic forms of comb.(add|sub|xor|or|and). I'll add support for unary instantiations and concat in follow-up PRs as they both require special handling.

There's some unrelated diff on some of the BTOR2 printing which I guess is just due to a clang-format change...

### DIFF
--- a/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
+++ b/lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
@@ -274,10 +274,9 @@ private:
   // and a result width.
   void genBinOp(StringRef inst, Operation *binop, Value op1, Value op2,
                 size_t width) {
-    // TODO: adding support for most variadic ops shouldn't be too hard
     if (binop->getNumOperands() != 2) {
-      binop->emitError("only the binary form of this operation is currently "
-                       "supported");
+      binop->emitError(
+          "only the binary form of this operation is currently supported");
       return;
     }
 


TaoBi22:
Looks like my local clang-format and the version in CI disagree :thinking:

TaoBi22:
Downgrading my clang-format to 17 fixed the issue (current CI failures look unrelated)

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -276,7 +276,8 @@ struct ConvertHWToBTOR2Pass
                 size_t width) {
     // TODO: adding support for most variadic ops shouldn't be too hard
     if (binop->getNumOperands() != 2) {
-      binop->emitError("variadic operations not are not currently supported");
+      binop->emitError("only the binary form of this operation is currently "
+                       "supported");
```
cowardsa: Nit: this looks super funky - why are there two strings?

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -276,7 +276,8 @@ struct ConvertHWToBTOR2Pass
                 size_t width) {
     // TODO: adding support for most variadic ops shouldn't be too hard
     if (binop->getNumOperands() != 2) {
```
cowardsa: Nit: does this comment need updating?

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
```diff
@@ -296,6 +297,43 @@ struct ConvertHWToBTOR2Pass
...
+    // sorts
+    if (!op->hasTrait<mlir::OpTrait::SameOperandsAndResultType>()) {
+      op->emitError(
+          "variadic operations with differing operand and result types are "
+          "not currently supported");
+      return;
+    }
+
+    // Unroll variadic op into series of binary ops
+    // This will represent the previous operand in the chain:
```
cowardsa, line 320: Comment: in CombToSynth variadic operators are lowered as binary trees - not sure if this would help with performance of downstream tools that consume the BTOR?

cowardsa (approved):
LGTM - but not super familiar with this part of CIRCT...

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
TaoBi22: I think this was a clang-format spill but I managed to shunt it to something else (on a machine with a more recent clang-format), I'll see if the CI clang-format agrees

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
TaoBi22: Oops, good catch, thanks!

lib/Conversion/HWToBTOR2/HWToBTOR2.cpp
TaoBi22: Ah interesting, I can imagine that would help! I might test that and follow up with a PR if downstream tools appreciate it

PR #9158, opened 2025-10-30
[Bug][ImportVerilog] [1/2] Add support for class method forward decls

Scheremo:
This commit adds the plumbing to convert `slang::ast::MethodPrototypeSymbol`s corresponding to class method forward declarations. This is done by looking up the implementation and converting it when encountering a forward declaration. This makes sure field order within the class declaration is respected and the emitted `methoddecl`'s signature is correct.

It also fixes a bug where the implicit `this` reference argument of method calls would get emitted twice in call sites.

### DIFF
--- a/lib/Conversion/ImportVerilog/Structure.cpp
+++ b/lib/Conversion/ImportVerilog/Structure.cpp
@@ -1189,10 +1189,9 @@ Context::convertFunction(const slang::ast::SubroutineSymbol &subroutine) {
 
   // If this is a class method, the first input is %this :
   // !moore.class<@C>
-  unsigned inputIdx = 0;
   if (isMethod) {
     auto thisLoc = convertLocation(subroutine.location);
-    auto thisType = lowering->op.getFunctionType().getInput(inputIdx++);
+    auto thisType = lowering->op.getFunctionType().getInput(0);
     auto thisArg = block.addArgument(thisType, thisLoc);
 
     // Bind `this` so NamedValue/MemberAccess can find it.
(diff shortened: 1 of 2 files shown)


lib/Conversion/ImportVerilog/Structure.cpp
```diff
@@ -1518,6 +1518,26 @@ struct ClassDeclVisitor {
...
+  // The latter case requires support for virtual interface methods, which is
+  // currently not implemented. Since forward declarations of non-interface
+  // methods must be followed by an implementation within the same compilation
+  // unit, we can simply return a failure if we can't find a unique
+  // implementation until we implement support for interface methods.
+  LogicalResult visit(const slang::ast::MethodPrototypeSymbol &fn) {
+    const auto *externImpl = fn.getSubroutine();
+    // We needn't convert a forward declaration without a unique implementation.
+    if (!externImpl)
+      return failure();
```
fabianschuiki: This should probably emit an error. Either Slang already catches this and emits an error, in which case we can just `assert(externImpl)`; or Slang doesn't complain at all, in which case we have to produce an error for the user here. Otherwise the tool returns with a non-zero exit code but no indication what failed.

test/Conversion/ImportVerilog/classes.sv
```diff
@@ -421,3 +421,27 @@ module testModule9;
...
+
+    class testModuleClass;
+        extern function int testFunction(int a);
+    endclass
+
+    function int testModuleClass::testFunction(int a);
+        return a;
+    endfunction
+
+endmodule
```
fabianschuiki: Nit: Since this is just checking forward decls, would it make sense to remove the surrounding `module`? We only need a single test that checks whether classes nested in modules work. Besides that, we should keep the tests as simple and focused on the specific issue they're testing as possible. In this case we're interested in forward decl, and not nesting in a class, so the surrounding class seems redundant.

lib/Conversion/ImportVerilog/Structure.cpp
Scheremo: It's technically valid system verilog that occurs on `pure virtual` methods in interface classes (see long comment before the visitor 🙂 ). We can emit an error and revisit once we get to interface virtual methods.

test/Conversion/ImportVerilog/classes.sv
Scheremo: Fair enough! Just boilerplated this up, but we can remove the module framing.

lib/Conversion/ImportVerilog/Structure.cpp
Scheremo: Done, now throws an error.

test/Conversion/ImportVerilog/classes.sv
Scheremo: Removed the framing module.

PR #9159, opened 2025-10-30
[ImportVerilog] [2/2] Add support to materialize ctor call on new call

Scheremo:
This commit adds full lowering support for class constructor (`new`) and superclass constructor (`super.new`) calls.

- Enhanced lowering of `slang::ast::NewClassExpression`:
  - Detects whether the `new` call originates from within a class constructor.
  - Differentiates between normal object allocation (`moore.class.new`) and `super.new` calls, which reuse the current `this` handle instead of allocating a new object.
  - Tracks the implicit `this` reference via a newly public `Context::currentThisRef` field, ensuring correct behavior in nested constructor and superclass calls.
  - Automatically invokes the appropriate class constructor after allocation, binding the newly created handle as `%this` for the call.

### DIFF
--- a/lib/Conversion/ImportVerilog/Expressions.cpp
+++ b/lib/Conversion/ImportVerilog/Expressions.cpp
@@ -10,6 +10,7 @@
 #include "slang/ast/EvalContext.h"
 #include "slang/ast/SystemSubroutine.h"
 #include "slang/syntax/AllSyntax.h"
+#include "llvm/ADT/ScopeExit.h"
 
 using namespace circt;
 using namespace ImportVerilog;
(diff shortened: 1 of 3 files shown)


lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1712,19 +1713,76 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+  // super is always unambiguous), but no allocation, as the child class' new
+  // invocation already allocated space for both its own and its parent's
+  // properties.
   Value visit(const slang::ast::NewClassExpression &expr) {
     auto type = context.convertType(*expr.type);
-    auto classTy = dyn_cast<moore::ClassHandleType>(type);
-    if (!classTy)
-      return {};
+    circt::moore::ClassHandleType classTy =
+        dyn_cast<moore::ClassHandleType>(type);
```
fabianschuiki:
```suggestion
    auto classTy =
        dyn_cast<moore::ClassHandleType>(type);
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1712,19 +1713,76 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+
+    // We are calling new from within a new function, and it's pointing to
+    // super. Check the implicit this ref to figure out the super class type.
+    // Do not allocate a new object.
+    if (!classTy && expr.isSuperClass) {
+      newObj = context.getImplicitThisRef();
+      if (!newObj || !newObj.getType() ||
+          !isa<moore::ClassHandleType>(newObj.getType())) {
+        mlir::emitError(loc) << "Implicit this ref was not set while "
+                                "converting new class function";
```
fabianschuiki:
```suggestion
        mlir::emitError(loc) << "implicit this ref was not set while "
                                "converting new class function";
```

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1712,19 +1713,76 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+
+    if (const auto *callConstructor =
+            constructor->as_if<slang::ast::CallExpression>())
+      if (const auto *subroutine =
+              std::get_if<const slang::ast::SubroutineSymbol *>(
+                  &callConstructor->subroutine)) {
+        // Bit paranoid, but virtually free checks that new is a class method
+        // and the subroutine has already been converted.
+        if (!(*subroutine)->thisVar)
+          return {};
```
fabianschuiki, line 1771: This should either be an assert, or you need to print an error message for the user. When returning `{}` to indicate failure, the "contract" is that an error message has already been produced. So if you call `convertFunction` and it returns `{}`, you can assume that the error was reported. But if you do your own checks like with that `thisVar`, you need to print that error. These compiler invariants are often easier to encode as an assert.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1712,19 +1713,76 @@ struct RvalueExprVisitor : public ExprVisitor {
...
-    mlir::emitError(loc) << "New expression requires constructor call, which "
-                            "is not yet implemented!";
+        // Pass the newObj as the implicit this argument of the ctor.
+        auto savedThis = context.currentThisRef;
+        context.currentThisRef = newObj;
+        auto restoreThis =
+            llvm::make_scope_exit([&] { context.currentThisRef = savedThis; });
+
+        // Emit a call to ctor
+        (void)visitCall(*callConstructor, *subroutine);
```
fabianschuiki: If this returns failure or `{}`, we should forward that failure upwards.

lib/Conversion/ImportVerilog/Expressions.cpp
```diff
@@ -1712,19 +1713,76 @@ struct RvalueExprVisitor : public ExprVisitor {
...
+      auto classDecl =
+          cast<moore::ClassDeclOp>(*context.symbolTable.lookupNearestSymbolFrom(
+              context.intoModuleOp, thisType.getClassSym()));
+      auto baseClassSym = classDecl.getBase();
+      classTy = circt::moore::ClassHandleType::get(context.getContext(),
+                                                   baseClassSym.value());
+    } else {
+      // We are calling from outside a class; allocate space for the object.
+      auto newOp = moore::ClassNewOp::create(builder, loc, classTy, {});
+      newObj = newOp.getResult();
```
fabianschuiki:
```suggestion
      newObj = moore::ClassNewOp::create(builder, loc, classTy, {});
```

lib/Conversion/ImportVerilog/Expressions.cpp
fabianschuiki: Maybe this should be an assert, if it's a compiler invariant? Although, making it an error will be easier for debugging 😁

test/Conversion/ImportVerilog/classes.sv
```diff
@@ -421,3 +421,105 @@ module testModule9;
...
+        endfunction
+    endclass // testModuleClass
+
+    class testModuleClass2 extends testModuleClass;
+        function new(int a);
+            super.new(a);
+        endfunction
+    endclass // testModuleClass
+
+endmodule
```
fabianschuiki: Similar to the other PR: we only need one check for nesting under modules. I'd just move these classes into the top-level. That makes the test more focused (it doesn't break if we break module nesting) and there's less text involved.

fabianschuiki (approved):
LGTM, modulo a few cases where we need to emit errors or propagate failures

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: We can also do both for good measure 😄

test/Conversion/ImportVerilog/classes.sv
Scheremo: Done!

lib/Conversion/ImportVerilog/Expressions.cpp
Scheremo: I'll keep it as an error for now though. I find it nicer for debugging as well 🙂

